PR #5, opened 2022-05-21
Fork oracle module

codchen:
Fork the oracle module from terra with the following modifications:
- Removed the reward structure
- Use `sei` as the slashing denomination
- Use `usdc` as the base denomination for exchange rates

Next steps:
- Add utils to derive exchange rate between non-base denoms
- Integrate `dex` with oracle
- Testing


udpatil (approved):
:shipit:

PR #11, opened 2022-05-24
[oracle] Update miss calculation for abstaining and unmet quorum

udpatil:
This updates the oracle logic to differ from the Terra behavior. We want the oracle to penalize abstaining from votes by counting them as misses, and to also count unmet quorum as miss for validator/oracles that don't provide in-range values to prevent collusion to avoid meeting quorum.


udpatil:
There are a couple TODOs, which pertain to a measure of staleness, which will indicate whether an exchange rate is a stale rate or new. I plan on addressing this in a next PR.

PR #12, opened 2022-05-24
Add Ledger support to binary and rename sei-chaind -> seid

dylanschultzie:
By using this Makefile, you can do: `make install` which makes `seid` and sticks it in `~/go/bin/`.

This also renames the root dir, `sei-chaind` to `seid` to make it align with the binary name.


dylanschultzie:
Really only the `makefile` is necessary, but the rest of the work streamlines the repo a bit.

philipsu522 (approved):
this is great @dylanschultzie thanks for doing this! We'll update our documentation to use the makefile so it's cleaner

PR #14, opened 2022-05-25
Upgrade 1.0.0

philipsu522:
### Description
Implements upgrade code from 1.0.0beta to 1.0.0. The following changes are included:
- sequence numbers added to support duplicate commands
- small fixes to seid name

### Test

1. Started chain with sequence numbers disabled
2. Submit the following proposal and voted on it ` ./build/seid tx gov submit-proposal software-upgrade "upgrade-1.0.0" --title "upgrade from 1.0.0beta to 1.0.0" --description "upgrade"  --from $ACCOUNT_NAME --upgrade-height 30 --deposit 10000000stake --chain-id sei-chain`
3. Verified that upgrade was needed `9:26PM ERR UPGRADE "upgrade-1.0.0" NEEDED at height: 30:`
4. Built new binary, restarted chain, verified blocks were built
5. Sent duplicate transactions, succeeded. Also verified tool shows changes: 
```
> ./build/seid -h
Stargate Sei App

Usage:
  seid [command]
```

### DIFF
--- a/app/const.go
+++ b/app/const.go
@@ -1,0 +1,8 @@
+package app
+
+const (
+	// Prefix of bech32 encoded address
+	AccountAddressPrefix = "sei"
+	// Application name
+	AppName = "sei-chain"
+)
--- a/app/const.go
+++ b/app/const.go
@@ -2,7 +2,7 @@ package app
 
 const (
 	// Prefix of bech32 encoded address
-	AccountAddressPrefix = "sei"
+	AccountAddressPrefix = "cosmos"
 	// Application name
 	AppName = "sei-chain"
 )
--- a/app/app.go
+++ b/app/app.go
@@ -113,11 +113,6 @@ import (
 	"go.opentelemetry.io/otel"
 )
 
-const (
-	AccountAddressPrefix = "cosmos"
-	Name                 = "sei-chain"
-)
-
 // this line is used by starport scaffolding # stargate/wasm/app/enabledProposals
 
 func getGovProposalHandlers() []govclient.ProposalHandler {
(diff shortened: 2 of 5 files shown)


LCyson:
just a quick q, why we're still using stake here `--deposit 10000000stake`

app/const.go
```diff
@@ -0,0 +1,8 @@
+package app
+
+const (
+	// Prefix of bech32 encoded address
+	AccountAddressPrefix = "cosmos"
+	// Application name
+	AppName = "sei-chain"
```
LCyson, line 7: we probably want to change this to "sei" instead of "sei-chain", so that the cmd dir will be the same name as the seid. See ticket 0030 and https://github.com/sei-protocol/sei-chain/pull/12

LCyson (approved):

mezhcoder:
Developers, I saw that in your official documentation, the `./build/seid` command is used to work with the project https://docs.seinetwork.io/nodes-and-validators/joining-testnets

How do you look at adding to the documentation or in the process of building a command of this kind?

`alias seid=$HOME/sei-chain/./build/seid`
Then it will be possible to write command `seid`

philipsu522:
> Developers, I saw that in your official documentation, the `./build/seid` command is used to work with the project https://docs.seinetwork.io/nodes-and-validators/joining-testnets
> 
> How do you look at adding to the documentation or in the process of building a command of this kind?
> 
> `alias seid=$HOME/sei-chain/./build/seid` Then it will be possible to write command `seid`

@dylanschultzie actually contributed a Makefile https://github.com/sei-protocol/sei-chain/pull/12 which adds the binary to go/bin. We're going to create a new release soon and update the docs to use seid automatically

PR #16, opened 2022-05-25
Fix default chain id

philipsu522:
App Name is "sei", chain is usually "sei-chain"


LCyson:
What will this change on the cmd side?

philipsu522:
it defaults to "sei-chain". Thinking about this more let's discard and keep it the same as other chains (default chain created is "sei" not "sei-chain"

PR #17, opened 2022-05-26
[oracle] Add staleness indicator to exchange rate store

udpatil:
Still a work in progress, most of the business logic is complete though.


x/oracle/keeper/keeper.go
```diff
@@ -67,25 +67,27 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 //-----------------------------------
 // ExchangeRate logic
 
-func (k Keeper) GetBaseExchangeRate(ctx sdk.Context, denom string) (sdk.Dec, error) {
+func (k Keeper) GetBaseExchangeRate(ctx sdk.Context, denom string) (sdk.Dec, sdk.Int, error) {
 	if denom == utils.MicroBaseDenom {
-		return sdk.OneDec(), nil
+		return sdk.OneDec(), sdk.ZeroInt(), nil
```
codchen: it might be easier for this to be set as the latest period-end block so that callers can handle base vs. non-base denom exchange rates consistently. Something like `ctx.BlockHeight() / votePeriod * votePeriod`. (would also make the eventual switch of base denom to `sei` easier)

codchen (approved):

codchen (approved):

x/oracle/keeper/migrations_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/sei-protocol/sei-chain/x/oracle/types"
+	"github.com/sei-protocol/sei-chain/x/oracle/utils"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func TestMigrate2to3(t *testing.T) {
```
philipsu522, line 14: +1 - this is an awesome test

philipsu522 (approved):

PR #19, opened 2022-05-28
Update DEX TWAP endpoint

codchen:
The previous twap endpoint was stateful and hardcoded to Vortex's use case.

The new twap endpoint itself is stateless, and performs calculation on the fly based on price snapshots. An exchange price snapshot is generated roughly every 60s and has a retention of one hour. Even though the snapshotting cadence is "roughly" 60s, the price snapshot itself will store the exact timestamp, which would make the twap calculation accurate.

Tested through both unit tests and local chain manual tests.

PR is large mainly due to proto-generated files.

The following steps will be performed in subsequent PRs:
- migration handler for this change
- parameterize the snapshotting cadence (right now it depends on epoch's cadence which is hardcoded)


x/dex/keeper/grpc_query_get_twaps.go
```diff
@@ -0,0 +1,57 @@
...
+
+func calculateTwap(ctx sdk.Context, prices []*types.Price, lookback uint64) sdk.Dec {
+	// sort prices in descending order to start iteration from the latest
+	sort.Slice(prices, func(p1, p2 int) bool {
+		return prices[p1].SnapshotTimestampInSeconds > prices[p2].SnapshotTimestampInSeconds
+	})
+	var timeTraversed uint64 = 0
+	var weightedPriceSum sdk.Dec = sdk.ZeroDec()
+	for _, price := range prices {
+		newTimeTraversed := uint64(ctx.BlockTime().Unix()) - price.SnapshotTimestampInSeconds
```
philipsu522, line 43: just double check ctx.BlockTime() is static (i.e. block start time) right?

x/dex/keeper/price.go
```diff
@@ -0,0 +1,61 @@
...
+	if !store.Has(key) {
+		res.Pair = &pair
+		return res, false
+	}
+	b := store.Get(key)
+	k.Cdc.MustUnmarshal(b, &res)
+	return res, true
+}
+
+func (k Keeper) GetAllPrices(ctx sdk.Context, contractAddr string, pair types.Pair) (list []*types.Price) {
```
philipsu522, line 35: when do we purge prices we no longer need?

x/dex/keeper/grpc_query_get_twaps.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	return &types.QueryGetTwapsResponse{
+		Twaps: twaps,
+	}, nil
+}
+
+func calculateTwap(ctx sdk.Context, prices []*types.Price, lookback uint64) sdk.Dec {
+	// sort prices in descending order to start iteration from the latest
+	sort.Slice(prices, func(p1, p2 int) bool {
```
philipsu522, line 37: since the ordering won't change, we could consider using a priority queue to reduce number of sorts . If number of prices is pretty small then probably doesn't matter

x/dex/keeper/grpc_query_get_twaps.go
codchen: yeah it's set during BeginBlock based on block proposal (so consistent across all nodes)

x/dex/keeper/price.go
codchen: https://github.com/sei-protocol/sei-chain/pull/19/files#diff-f32c70484252fadec70cda9a6ad7888fd568f6afffb3750bb9621090e2ac88bbR249

philipsu522 (approved):

x/dex/module.go
```diff
@@ -228,51 +230,25 @@ func (am AppModule) beginBlockForContract(ctx sdk.Context, contractAddr string)
...
-			ctx.Logger().Info("Setting funding payment rate")
-			am.callClearingHouseContractSudo(ctx, wasmMsg, contractAddr)
-
-			var newPrices []uint64
-			if len(twap.Prices) == 24 { // replace with config
-				newPrices = append(twap.Prices[1:], twap.Prices[len(twap.Prices)-1])
-			} else {
-				newPrices = append(twap.Prices, twap.Prices[len(twap.Prices)-1])
+			// condition to prevent unsigned integer overflow
+			if currentEpoch >= priceRetention {
```
udpatil, line 247: is it always guaranteed that an epoch length is 60s? wouldnt it be good to add a check here that we only delete the epoch data if the current BlockTime - epoch BlockTime > 60 min? however one issue with that is that there is no mechanism to go back to an epoch if you skip over it on that condition :/ WDYT @codchen

x/dex/keeper/grpc_query_get_twaps.go
```diff
@@ -0,0 +1,57 @@
...
+		return prices[p1].SnapshotTimestampInSeconds > prices[p2].SnapshotTimestampInSeconds
+	})
+	var timeTraversed uint64 = 0
+	var weightedPriceSum sdk.Dec = sdk.ZeroDec()
+	for _, price := range prices {
+		newTimeTraversed := uint64(ctx.BlockTime().Unix()) - price.SnapshotTimestampInSeconds
+		if newTimeTraversed > lookback {
```
udpatil, line 44: in this case, wouldn't we exclude any data from the first time interval? 

Lets say for example we had data at T-61 and then T-59 ... T-0, when calculating the TWAP, it would technically only represent the weighted price sum for T-59 - T-0, right? because we would break on T-61 and never include data for the first interval?

x/dex/module.go
codchen: actually for `dex` I'll extend the retention to be way longer than 60min since its price snapshots are needed for historical price queries as well. In that case we probably don't have this issue with twap I think?

x/dex/keeper/grpc_query_get_twaps.go
codchen: that's true. We should align on the behavior in such case between dex twap and oracle twap. Still factoring in T-61 but only assign a time weight of `1` sounds reasonable to me, if that's also how oracle twap is calculated

udpatil: yeah that makes sense to me, currently my implementation also has the missing first interval issue, but I can modify it to keep the first out of bounds data point for calculating to the start of the TWAP window

udpatil: will approve once this change is in

codchen: pushed the fix as well as a unit test for this case

PR #21, opened 2022-05-29
Set gas limit in parameter for Begin/EndBlock in dex

codchen:
Set a gas limit for both Begin/EndBlock. If the gas limit is hit, all store writes need to be aborted, which is achieved via the usage of cache store.


x/dex/types/params.go
```diff
@@ -11,10 +11,14 @@ import (
 var (
 	KeyPriceSnapshotRetention = []byte("PriceSnapshotRetention") // number of epochs to retain price snapshots for
 	KeySudoCallGasPrice       = []byte("KeySudoCallGasPrice")    // gas price for sudo calls from endblock
+	KeyBeginBlockGasLimit     = []byte("KeyBeginBlockGasLimit")
+	KeyEndBlockGasLimit       = []byte("KeyEndBlockGasLimit")
```
LCyson, line 15: are we using this in the endblock rn?

LCyson (approved):

PR #24, opened 2022-05-31
Clean up data types in `dex` module

codchen:
- Change all `price`/`quantity`/`leverage` data type to `sdk.Dec` which is a fixed 18-decimal-place numeric representation.
- Change all denom data type to an enum
- Remove `nonce`
- Add migration logics for data type changes of the following states: long book, short book, settlements, registered pairs, twap

Tested in local chain with the following steps:
- Started chain of the current master version (i.e. old version)
- Populated states by sending orders to `dex` (at this point the states are in old data type)
- Sent and passed proposal to upgrade
- Resumed chain of this commit's version (i.e. new version)
- Queried states and observed that the states are all in new data type correctly


x/dex/migrations/v2_to_v3.go
```diff
@@ -0,0 +1,212 @@
...
+	longBookKey := types.KeyPrefix(types.LongBookKey)
+	store := prefix.NewStore(ctx.KVStore(storeKey), longBookKey)
+	iterator := sdk.KVStorePrefixIterator(store, []byte{})
+
+	defer iterator.Close()
+	for ; iterator.Valid(); iterator.Next() {
+		var val legacytypes.LongBook
+		cdc.MustUnmarshal(iterator.Value(), &val)
+		key := iterator.Key()
+		store.Delete(key)
```
philipsu522, line 33: should we wait until the end to delete?

x/dex/migrations/v2_to_v3.go
```diff
@@ -0,0 +1,212 @@
...
+		priceDenom, err := types.GetDenomFromStr(val.Entry.PriceDenom)
+		if err != nil {
```
philipsu522, line 35: rather than just continuing, should we log it so that we can revisit and try to fix this later? If we delete at the end the value is still there

codchen: together with the `.Delete` comment above, my thought was to prevent any bad state from lingering in the blockchain after the migration since a bad state could potentially crash the chain during logics in Begin/EndBlock, and I don't know of a good way to fix bad states after migrations. Maybe we can just panic here? So that migration will fail, and we can go back and add in any missed edge cases and rerun migration

philipsu522 (approved):

PR #33, opened 2022-06-05
Use slice based loop instead of map based loop in Begin/EndBlock

codchen:
Instead of looping over a map, we will fetch a slice of registered pairs first, and then iterate over the slice and get map's value keyed by each pair, so that the iteration is deterministic.


LCyson (approved):
lgtm, can we add some test sets to dex endBlock so that if there is any non-determinism in the future the CI will fail

codchen:
@LCyson added a test for idempotency

PR #36, opened 2022-06-07
buf in ./proto

vuong177:
I added buf in proto for easier generate protobuf.


philipsu522 (commented):
@vuong177 thanks for doing this! To clarify, what would the correct command be to generate proto going forward?

vuong177:
> @vuong177 thanks for doing this! To clarify, what would the correct command be to generate proto going forward?

My command is :
```
cd proto
buf generate --template buf.gen.gogo.yaml
```

philipsu522 (approved):

PR #40, opened 2022-06-11
Fix loadtest scripts

codchen:
Fix scripts used by loadtest to work with sequence numbers properly. Tested on local chain with 1000 accounts.

### DIFF
--- a/x/dex/keeper/end_block_place_orders.go
+++ b/x/dex/keeper/end_block_place_orders.go
@@ -18,7 +18,7 @@ func (k *Keeper) HandleEBPlaceOrders(ctx context.Context, sdkCtx sdk.Context, tr
 	_, span := (*tracer).Start(ctx, "SudoPlaceOrders")
 	span.SetAttributes(attribute.String("contractAddr", contractAddr))
 
-	msgs := k.getPlaceSudoMsg(contractAddr, registeredPairs)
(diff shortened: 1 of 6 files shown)


philipsu522 (approved):

x/dex/keeper/end_block_place_orders.go
```diff
@@ -52,25 +62,37 @@ func (k *Keeper) HandleEBPlaceOrders(ctx context.Context, sdkCtx sdk.Context, tr
...
+				Deposits: contractDepositInfo,
+			},
+		},
+	}
 	for _, pair := range registeredPairs {
 		if orderPlacements, ok := k.OrderPlacements[contractAddr][pair.String()]; ok {
 			for _, orderPlacement := range orderPlacements.Orders {
 				if !orderPlacement.Liquidation {
 					contractOrderPlacements = append(contractOrderPlacements, dexcache.ToContractOrderPlacement(orderPlacement))
+					if len(contractOrderPlacements) == MAX_ORDERS_PER_SUDO_CALL {
```
LCyson, line 84: here if the number of contractOrderPlacements doesn't reach 50000 here, is the contractOrderPlacements discarded then?

codchen: oh good catch

LCyson: can we add a unit test for this one^

PR #41, opened 2022-06-12
pass password in when running oracle script

codchen:
Keyring password is enabled on remote boxes

### DIFF
--- a/scripts/oracle/mock_price_feed.py
+++ b/scripts/oracle/mock_price_feed.py
@@ -4,7 +4,8 @@ import requests
 from pathlib import Path
 import sys
 
-CMD = str(Path.home()) + "/go/bin/seid"
+# for internal loadtest only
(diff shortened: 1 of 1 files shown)


scripts/oracle/mock_price_feed.py
```diff
@@ -4,7 +4,7 @@
 from pathlib import Path
 import sys
 
-CMD = str(Path.home()) + "/go/bin/seid"
+CMD = "printf '12345678\n' | " + str(Path.home()) + "/go/bin/seid"
```
LCyson, line 8: Can we leave a comment here about this is for internal testing only

LCyson (approved):

scripts/oracle/mock_price_feed.py
codchen: :+1:

PR #43, opened 2022-06-15
[oracle] Add price snapshot history and associated queries

udpatil:
This adds the logic for maintaining a price snapshot history and querying the history. Will put out a separate PR shortly for the TWAP specific logic.


x/oracle/keeper/keeper.go
```diff
@@ -361,3 +361,72 @@ func (k Keeper) ValidateFeeder(ctx sdk.Context, feederAddr sdk.AccAddress, valid
...
+			indexToSliceBefore = i - 1
+			if indexToSliceBefore < 0 {
+				// if there is no datapoint outside of range, we simply keep all of data
+				indexToSliceBefore = 0
+			}
+			break
+		}
+	}
+	snapshots = snapshots[indexToSliceBefore:]
+	k.SetPriceSnapshots(ctx, snapshots)
```
codchen: we would be writing the entire price history to disk each time we add a new snapshot, which might be a bit too heavy. Maybe we can use timestamp as a subkey so that each snapshot can be added/deleted individually?

udpatil: ah yeah, this is a good point, I'll make this change

philipsu522 (approved):
nice work

codchen (approved):

PR #44, opened 2022-06-15
[oracle] Add twap calculation and endpoints

udpatil:
This is a PR for the oracle TWAP stuff on top of the price history snapshots. Putting it out for review, but will modify PR to merge to master after the prior PR is merged.


codchen (approved):

x/oracle/keeper/keeper.go
```diff
@@ -417,7 +417,102 @@ func (k Keeper) IteratePriceSnapshots(ctx sdk.Context, handler func(snapshot typ
...
+
+			// set the denom TWAP data
+			denomToTimeWeightedMap[denom] = denomTimeWeightedSum
+			denomDurationMap[denom] = timeTraversed
+		}
+		return
+	})
+
+	// iterate over all denoms with TWAP data
+	for denomKey := range denomToTimeWeightedMap {
```
philipsu522: let's be careful here - we're iterating over a map which can be nondeterministic. I think it should be ok since this isn't saving to state, but it might still be safer to somehow store the keys in a list, sort then iterate based on that in case we decide to use this in the future for saving to state

x/oracle/keeper/keeper.go
```diff
@@ -417,7 +417,102 @@ func (k Keeper) IteratePriceSnapshots(ctx sdk.Context, handler func(snapshot typ
...
+			if !exists {
+				// set up the TWAP info for a denom
+				denomToTimeWeightedMap[denom] = sdk.ZeroDec()
+				denomDurationMap[denom] = 0
+			}
+			// get the denom specific TWAP data
+			denomTimeWeightedSum := denomToTimeWeightedMap[denom]
+			denomDuration := denomDurationMap[denom]
+
+			// calculate the new Time Weighted Sum for the denom exchange rate
```
philipsu522, line 478: do you mind adding a comment here including the twap formula?

x/oracle/keeper/keeper.go
udpatil: sure yeah, I'll collect denoms in a sorted list prior to iterating and outputting twaps 👍

x/oracle/keeper/keeper.go
udpatil: yup will do

udpatil:
@philipsu522 I addressed the feedback 👍

philipsu522 (approved):

PR #45, opened 2022-06-15
Gasless transaction fixes

codchen:
This PR does two things:
1. Add a whitelist (currently hardcoded but will be parameterized in a subsequent PR) for addresses that can cancel orders gaslessly.
2. Ensure order placements and order cancellations (for whitelisted accounts) are truly gasless by overriding the gas meter in TX's context to an infinite gas meter

Tested on local chain


philipsu522 (approved):
nice nice

PR #47, opened 2022-06-17
[oracle] Modify vote penalty to not jail if validator abstains instead of misses

udpatil:
still need to resolve issues with `simulation/decoder` but the rest of the change is ready for review.


app/upgrades.go
```diff
@@ -9,22 +9,37 @@ import (
...
+// this will introduce the oracle module as well
+const Upgrade104 = "1.0.4beta"
+
 func (app App) RegisterUpgradeHandlers() {
-	app.UpgradeKeeper.SetUpgradeHandler(UpgradeNameOracleModule, func(ctx sdk.Context, plan upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
+	app.UpgradeKeeper.SetUpgradeHandler(IgniteCLIRemovalUpgradeHandler, func(ctx sdk.Context, plan upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
 		return app.mm.RunMigrations(ctx, app.configurator, fromVM)
 	})
-	app.UpgradeKeeper.SetUpgradeHandler(UpgradeOracleStaleIndicator, func(ctx sdk.Context, plan upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
+	app.UpgradeKeeper.SetUpgradeHandler(UpgradeDexDataType, func(ctx sdk.Context, plan upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
```
LCyson: we can safely remove this and `IgniteCLIRemovalUpgradeHandler` right (as long as we keep the N-1 upgrade handler) @philipsu522, if so let's remove them in this PR

x/oracle/abci.go
```diff
@@ -96,15 +96,20 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper) {
...
 		// Do miss counting & slashing
 		for _, claim := range validatorClaimMap {
-			// Skip abstain & valid voters
 			// we require validator to have submitted in-range data
 			// for all assets to not be counted as a miss
 			if int(claim.WinCount) == totalTargets {
 				continue
 			}
 
+			if int(claim.WinCount+claim.AbstainCount) == totalTargets {
```
LCyson, line 105: quick q, what's claim.WinCount here? any vote count or just correct vote count?

x/oracle/tally.go
```diff
@@ -22,8 +22,15 @@ func Tally(ctx sdk.Context, pb types.ExchangeRateBallot, rewardBand sdk.Dec, val
 
 	for _, vote := range pb {
 		// Filter ballot winners
-		// abstaining counts as out of range and will be eventually penalized
-		if vote.ExchangeRate.GTE(weightedMedian.Sub(rewardSpread)) &&
+		// abstaining counts as out of range and will be eventually penalized but not jailed
+		if !vote.ExchangeRate.IsPositive() {
```
LCyson, line 26: `IsPositive()` here doesn't include 0 right?

LCyson (approved):

LCyson:
Can we add the command to test pre-vote and vote in this PR description

x/oracle/abci.go
udpatil: WinCount is the count of votes that the validator made that are within the 1% range of the weighted median. So, if they win count is equal to total targets, that means the validator was in range on all of the targets, if they specifically abstained on some, we expect them to win the ones they didnt abstain to qualify as `abstaining` instead of `miss`

x/oracle/tally.go
udpatil: no it doesn't, there is a specific `sign` corresponding to zero:
```
func (d Dec) IsZero() bool      { return (d.i).Sign() == 0 }          // is equal to zero
func (d Dec) IsNegative() bool  { return (d.i).Sign() == -1 }         // is negative
func (d Dec) IsPositive() bool  { return (d.i).Sign() == 1 }          // is positive
```

udpatil:
prevote command example:
`seid tx oracle aggregate-prevote $SALT 1uusdc,2uatom SEIVALOPER_ADDR --from $ACCOUNT_NAME --fees 2000usei`

(the vote has to be in the next vote window)
vote command example:
`seid tx oracle aggregate-vote $SALT 1uusdc,2uatom SEIVALOPER_ADDR --from $ACCOUNT_NAME --fees 2000usei`

PR #52, opened 2022-06-22
Pass order type to SettlementEntry

codchen:
Contracts would need order type information for each settlement entry for things like order fee calculation, so sei needs to pass order type during settlement.

Tested with unit tests as well as local integration test with Vortex contract.


x/dex/exchange/settlement.go
```diff
@@ -16,6 +16,7 @@ func Settle(
 	order types.OrderBook,
 	takerDirection types.PositionDirection,
 	worstPrice sdk.Dec,
+	isTakerLiquidation bool,
```
LCyson: can we just pass in the takerOrderType (e.g. `types.OrderType_LIQUIDATION`) here? so that we don't need to do the check below

LCyson: also can we add a general comment of the settlement logic in this function

LCyson (approved):

x/dex/exchange/settlement.go
codchen: added

PR #53, opened 2022-06-23
Add spawn chain script

0xeren:
Add some quick scripts that work on current top commit


x/dex/client/rest/tx.go
```diff
@@ -0,0 +1,11 @@
+package rest
+
+import (
+	"net/http"
+
+	"github.com/cosmos/cosmos-sdk/types/rest"
+)
+
+func registerTxRoutes() {
```
codchen: these aren't intended right?

codchen (approved):

PR #56, opened 2022-06-24
[oracle] Add combined vote and prevote tx to reduce block usage

udpatil:
I tested this via CLI with localsei to verify behavior. Not sure what the best way to verify REST endpoint behavior is.

### DIFF
--- a/x/oracle/client/cli/tx.go
+++ b/x/oracle/client/cli/tx.go
@@ -30,6 +30,7 @@ func GetTxCmd() *cobra.Command {
 		GetCmdDelegateFeederPermission(),
 		GetCmdAggregateExchangeRatePrevote(),
 		GetCmdAggregateExchangeRateVote(),
+		GetCmdAggregateExchangeRateCombinedVote(),
 	)
 
 	return oracleTxCmd
@@ -92,11 +93,11 @@ func GetCmdAggregateExchangeRatePrevote() *cobra.Command {
 		Short: "Submit an oracle aggregate prevote for the exchange rates of Luna",
 		Long: strings.TrimSpace(`
 Submit an oracle aggregate prevote for the exchange rates of Luna denominated in multiple denoms.
-The purpose of aggregate prevote is to hide aggregate exchange rate vote with hash which is formatted 
+The purpose of aggregate prevote is to hide aggregate exchange rate vote with hash which is formatted
 as hex string in SHA256("{salt}:{exchange_rate}{denom},...,{exchange_rate}{denom}:{voter}")
 
 # Aggregate Prevote
-$ terrad tx oracle aggregate-prevote 1234 8888.0ukrw,1.243uusd,0.99usdr 
+$ terrad tx oracle aggregate-prevote 1234 8888.0ukrw,1.243uusd,0.99usdr
 
 where "ukrw,uusd,usdr" is the denominating currencies, and "8888.0,1.243,0.99" is the exchange rates of micro Luna in micro denoms from the voter's point of view.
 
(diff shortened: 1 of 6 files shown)


x/oracle/keeper/msg_server.go
```diff
@@ -128,6 +128,52 @@ func (ms msgServer) AggregateExchangeRateVote(goCtx context.Context, msg *types.
...
+		return nil, err
+	}
+
+	params := ms.GetParams(ctx)
+
+	var voteErr error
+	aggregatePrevote, err := ms.GetAggregateExchangeRatePrevote(ctx, valAddr)
+	// if there isn't a prevote, we want to no-op the vote so we don't get an error
+	// this way, it is safe to use combined vote regardless of a missed vote window
+	if err == nil && (uint64(ctx.BlockHeight())/params.VotePeriod)-(aggregatePrevote.SubmitBlock/params.VotePeriod) == 1 {
```
codchen: nit we can put `(uint64(ctx.BlockHeight())/params.VotePeriod)-(aggregatePrevote.SubmitBlock/params.VotePeriod) == 1` to a helper function called something like `IsPrevoteCurrent` for readability

codchen (approved):

LCyson:
after removing prevote, how do we guarantee validators won't simply copy others' price feeds by looking at their combinedVote here

LCyson (approved):

x/oracle/client/cli/tx.go
```diff
@@ -209,3 +210,75 @@ $ terrad tx oracle aggregate-vote 1234 8888.0ukrw,1.243uusd,0.99usdr terravalope
...
 }
+
+// GetCmdAggregateExchangeRatePrevote will create a aggregateExchangeRatePrevote tx and sign it with the given key.
+func GetCmdAggregateExchangeRateCombinedVote() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "aggregate-combined-vote [vote-salt] [vote-exchange-rates] [prevote-salt] [prevote-exchange-rates] [validator]",
+		Args:  cobra.RangeArgs(4, 5),
+		Short: "Submit an oracle aggregate vote AND prevote for the exchange rates",
+		Long: strings.TrimSpace(`
+Submit an oracle aggregate vote and prevote for the exchange rates of the base denom denominated in multiple denoms. The vote is a companian to a prevote from the previous vote window.
```
philipsu522, line 221: nit - "while the prevote is for the current window"

philipsu522 (commented):
let's also add tests.
We'd also want to check the behavior regarding providing a hash and leaving it out

LCyson:
> @LCyson we're not removing prevote, the combined vote performs a vote for the current window, and a prevote for the next window. the prevote for the next window protects against other validators copying the exchange rates.

what's the workflow for combinedVote in this case? 
previously it's 1) val sends prevote tx 2) val sends vote tx 3) compare prevote and vote in the endblock
now is it 1) val sends a combinedVote tx 2) compare prevote and vote included in the combinedVote in the endblock

x/oracle/client/cli/tx.go
udpatil: the prevote is performed in the current window, but is for the vote in the next window

x/oracle/keeper/msg_server.go
udpatil: done

LCyson:
> now it's, val sends combined vote, if there isn't a prevote from previous block, the vote is a no-op, and then prevote for the next block is performed.

oh ic, so the combinedVote consists of (n - 1) prevote + n vote, that makes sense

philipsu522 (approved):

LCyson (approved):

PR #57, opened 2022-06-25
[oracle] Add implicit abstaining behavior and remove explicit abstain

udpatil:
This still has a miss counter and abstain counter so that later one we can calculate what the wincount was based on vote window and miss/abstain counts. Additionally, I added events at the end of the slashing window with wincount, miss count, and abstain count. Not super familiar with the proper way to handle events, but presumably the mint module could consume those events and reward the validator at the end of vote windows when it receives the events.


philipsu522 (approved):

LCyson:
not sure which it's easier 1) retrieve miss counter and abstain counter from oracle keeper 2) retrieve from emitted event, but up to @philipsu522 to decide

codchen (approved):

PR #58, opened 2022-06-25
Generate JSON tag for proto types

codchen:
Since we use `gogoproto` for protobuf, the default JSON tag doesn't work and we have to explicitly specify `gogoproto.jsontag`.

All changes in this PR are in `proto` files, and `.pb.go` changes are all auto-generated. Since this change doesn't modify marshal/unmarshal behavior (i.e. we didn't add new fields, change data type, or reorder fields), no migration is needed.


0xeren (commented):
this mainly for readability not for correctness? looks good

codchen:
@0xeren yes. Can you approve it if it lgty?

0xeren (approved):

PR #59, opened 2022-06-26
[oracle][dex] Refactor dex ante handler into top level app, and add oracle gasless tx for first tx in voting window

udpatil:
This refactors the gasless ante handler, and we want to include oracle gasless voting too.
Additionally, I refactored the gasless wrapper decorator to be build into the gasless wrapper since otherwise we'd incur two checks for gasless TX

Oracle gasless behavior summary
Prevote: gasless allowed if the validator hasn't already submitted a prevote in the voting window. This means feeder or validator will be able to make one gasless prevote tx
Vote: gasless allowed if validator hasnt already submitted a vote in the voting window. ALSO there needs to be a prevote from the previous window present in store (implying the vote can potentially be valid)

Combined: TODO, will add in future PR


app/antedecorators/gasless.go
```diff
@@ -0,0 +1,169 @@
...
+	"bytes"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	dextypes "github.com/sei-protocol/sei-chain/x/dex/types"
+	oraclekeeper "github.com/sei-protocol/sei-chain/x/oracle/keeper"
+	oracletypes "github.com/sei-protocol/sei-chain/x/oracle/types"
+)
+
+type GaslessDecorator struct {
+	wrapped      sdk.AnteDecorator
```
codchen: we might want to make this a list, since it's possible that some future cosmos version adds a new builtin ante handler that we want to skip for gasless transactions

philipsu522 (approved):

codchen (approved):

PR #61, opened 2022-06-27
[sei] Governance proposal for new token pair on the dex

Kbhat1:
https://linear.app/passportlabs/issue/PAS-812/[sei]-require-governance-proposal-for-new-token-pair-on-the-dex

- removes ability to submit transaction to create a dex whitelisted token pairs
- Adds governance proposal type to add dex whirelisted token pairs
- Does not add ability to arbitrarily add any denom - only restricted to current enum set 

Sample Register Pair proposal.json

```
{
  "title": "Add usdc,sei + usdc,atom pair to contract 1 and usdc,btc pair to contract 2",
  "description": "gov proposal to update contract 1 and contract 2 registered pairs",
  "batch_contract_pair": [
    {
      "contract_addr":  "contract_1_address",
      "pairs": [
        {
          "price_denom": "usdc",
          "asset_denom": "sei"
        },
        {
          "price_denom": "usdc",
          "asset_denom": "atom"
        }
      ],
    },
    {
      "contract_addr": "contract_2_address",
      "pairs": [
        {
          "price_denom": "usdc",
          "asset_denom": "btc"
        }
      ],
    },
  ],
  "deposit": "10000000usei"
}
```


x/dex/handler.go
```diff
@@ -40,3 +39,18 @@ func NewHandler(k keeper.Keeper, tracingInfo *tracing.TracingInfo) sdk.Handler {
...
+		switch c := content.(type) {
+		case *types.RegisterPairsProposal:
+			return handleRegisterPairsProposal(ctx, k, c)
+		default:
+			return sdkerrors.Wrapf(sdkerrors.ErrUnknownRequest, "unrecognized dex proposal content type: %T", c)
+		}
+	}
+}
+
+func handleRegisterPairsProposal(ctx sdk.Context, k keeper.Keeper, p *types.RegisterPairsProposal) error {
```
udpatil: why do we need this function? why don't we just call
`k.HandleRegisterPairsProposal(ctx, p)` in the switch-case?

x/dex/keeper/pair.go
```diff
@@ -30,13 +30,21 @@ func (k Keeper) GetPairCount(ctx sdk.Context, contractAddr string) uint64 {
 	return binary.BigEndian.Uint64(cnt)
 }
 
-func (k Keeper) AddRegisteredPair(ctx sdk.Context, contractAddr string, pair types.Pair) {
+func (k Keeper) AddRegisteredPair(ctx sdk.Context, contractAddr string, pair types.Pair) bool {
+	// Only add pairs that have not been added before
+	for _, prevPair := range k.GetAllRegisteredPairs(ctx, contractAddr) {
```
udpatil, line 35: instead of iterating here, I think we should make the `AddRegisteredPair` idempotent since if we are trying to add multiple pairs, itll be super poor performance since we'd iterate over all registered pairs  for each one.

x/dex/types/gov.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// TODO: String support for register pair type
+func (p RegisterPairsProposal) String() string {
+	batchContractPairRecords := ""
+	for _, contractPair := range p.Batchcontractpair {
+		batchContractPairRecords += contractPair.String()
+	}
+	var b strings.Builder
+	b.WriteString(fmt.Sprintf(`Update Fee Token Proposal:
```
udpatil: this is incorrect, right?

x/dex/keeper/pair.go
udpatil: Discussed offline, we don't need to address this at this moment, we can optimize this when we move to a string representation of denom. along with that, we'd want to do a store migration for the registered pair store and have the registered pair key be build by string literal prefix + contract address + pair literal. This we we can check for existence without iteration.
Here is the location of the key we would need to modify: https://github.com/sei-protocol/sei-chain/blob/a15402115078923698cafb36295def41761d8754/x/dex/types/keys.go#L73

When we do the modification, we would need to do a store migration to ensure the old data format converted.

CC @Kbhat1 @codchen

Kbhat1: Yup, will do this when doing the larger PR to update the price type from enum to string (for arbitrary token denoms)

x/dex/types/gov.go
Kbhat1: Updated

x/dex/handler.go
Kbhat1: Updated

PR #63, opened 2022-06-28
[oracle][wasm] Add support for wasm custom querier

udpatil:
Added unit test, and also tested with `sei-cosmwasm` sei-tester contract to verify the query works end to end


wasmbinding/query_plugin.go
```diff
@@ -0,0 +1,61 @@
...
+	Route string `json:"route,omitempty"`
+	// The query data that should be parsed into the module query
+	QueryData json.RawMessage `json:"query_data,omitempty"`
+}
+
+func CustomQuerier(qp *QueryPlugin) func(ctx sdk.Context, request json.RawMessage) ([]byte, error) {
+	return func(ctx sdk.Context, request json.RawMessage) ([]byte, error) {
+		var contractQuery SeiQueryWrapper
+		// TODO: debugging, remove this
+		ctx.Logger().Error("received a custom query")
```
philipsu522: should we change this to debug level?

wasmbinding/query_plugin.go
```diff
@@ -0,0 +1,61 @@
...
+		if err := json.Unmarshal(request, &contractQuery); err != nil {
+			return nil, sdkerrors.Wrap(err, "sei query")
+		}
+		switch contractQuery.Route {
+		case OracleRoute:
+			// TODO: debugging, remove this
+			ctx.Logger().Error("parse route: oracle query")
```
philipsu522: same here

wasmbinding/message_plugin.go
```diff
@@ -0,0 +1,282 @@
+package wasmbinding
+
```
philipsu522: do we plan to keep this file?

wasmbinding/queries.go
```diff
@@ -0,0 +1,44 @@
...
+func (qp QueryPlugin) HandleOracleQuery(ctx sdk.Context, queryData json.RawMessage) ([]byte, error) {
+	var parsedQuery oraclebindings.SeiOracleQuery
+	if err := json.Unmarshal(queryData, &parsedQuery); err != nil {
+		return nil, sdkerrors.Wrap(err, "sei oracle query")
+	}
+	switch {
+	case parsedQuery.ExchangeRates != nil:
+		res, err := qp.oracleHandler.GetExchangeRates(ctx)
+		if err != nil {
+			return nil, sdkerrors.Wrap(err, "sei oracle GetExchangeRates query")
```
philipsu522: Given that these errors will be user facing, let's make these error messages more descriptive. I.e. 
- Unable to send GetExchangeRates query
- Unable to decode GetExchangeRates response

wasmbinding/wasm.go
```diff
@@ -0,0 +1,27 @@
...
+func RegisterCustomPlugins(
+	oracle *oraclekeeper.Keeper,
+) []wasmkeeper.Option {
+	oracleHandler := oraclewasm.NewOracleWasmQueryHandler(oracle)
+	wasmQueryPlugin := NewQueryPlugin(oracleHandler)
+
+	queryPluginOpt := wasmkeeper.WithQueryPlugins(&wasmkeeper.QueryPlugins{
+		Custom: CustomQuerier(wasmQueryPlugin),
+	})
+	// messengerDecoratorOpt := wasmkeeper.WithMessageHandlerDecorator(
```
philipsu522: rm if not needed

x/oracle/client/wasm/bindings/queries.go
```diff
@@ -0,0 +1,25 @@
...
+import "github.com/sei-protocol/sei-chain/x/oracle/types"
+
+type SeiOracleQuery struct {
+	// queries the oracle exchange rates
+	ExchangeRates *WasmQueryExchangeRatesRequest `json:"exchange_rates,omitempty"`
+}
+
+type WasmQueryExchangeRatesRequest struct{}
+
+// type WasmQueryOracleExchangeRate struct {
```
philipsu522: rm if not needed

app/app.go
```diff
@@ -455,6 +456,7 @@ func New(
 	// The last arguments can contain custom message handlers, and custom query handlers,
 	// if we want to allow any custom callbacks
 	supportedFeatures := "iterator,staking,stargate"
+	wasmOpts = append(wasmbinding.RegisterCustomPlugins(&app.OracleKeeper), wasmOpts...)
```
codchen, line 459: we can set wasmopts here https://github.com/sei-protocol/sei-chain/blob/master/cmd/seid/cmd/root.go#L233 just in case  some future logics in `New` need to make use of wasmOpts but happens before this line

wasmbinding/query_plugin.go
udpatil: oof sorry, I needed to remove this I forgot to remove from all places 🤦‍♂️

wasmbinding/message_plugin.go
udpatil: yeah, we'll use this when integrating txs from cosmwasm. we'll handle messages (txs) separately from queries so this one will be the plugin for message handling

wasmbinding/queries.go
udpatil: yup, will fix 👍

wasmbinding/wasm.go
udpatil: we'll need this for when we do integrate message handling (which would be soon presumably for adding in things like `PlaceOrder` or `CancelOrder` from smart contract to dex

x/oracle/client/wasm/bindings/queries.go
udpatil: yup, will do

app/app.go
udpatil: I don't think we can do that there because we need to already have an OracleKeeper for this option to be added, and the keepers are all initialized in New() not in root, right?

udpatil: @codchen wdyt?

codchen: oh yeah you're right

wasmbinding/message_plugin.go
codchen: Is there a reason why we want a custom message dispatcher, rather than use the default message dispatcher with a custom CustomMsg encoder?

codchen (approved):

wasmbinding/message_plugin.go
udpatil: I believe the default message dispatcher won't allow messages to the custom modules, right? or is that not the case?

codchen: It allows custom message but the default behavior is to no-op. If you add something like this to wasmopts you can override the no-op behavior:
```
	wasmopts := []wasm.Option{wasmkeeper.WithMessageEncoders(&wasmkeeper.MessageEncoders{
		Custom: xwasm.DexCustomEncoder,
	})}
```
Of course we can change `DexCustomEncoder` to be something more generic

udpatil: ah ok, in that case i'll remove these files

PR #69, opened 2022-06-30
[oracle] add sei to whitelist

codchen:
We need `sei`'s price for both integration test as well as incentivized testnet.


LCyson (approved):

LCyson:
whitelist SEI price is only for integ test purposes correct? note that before we launch in mainnet the SEI token won't have real market price

codchen:
@LCyson yes. Validators can abstain on sei's price now I think, so should be okay to include in the whitelist

PR #73, opened 2022-07-01
Fix register pair proposal

codchen:
- Register `gov` router after `dex` keeper is instantiated, instead of before
- Add common flags (e.g. `--from`) to `NewRegisterPairsProposalTxCmd`
- Minor fixes to local chain script


Kbhat1 (approved):
This makes sense, thanks for this

PR #80, opened 2022-07-03
Add tick size 

0xeren:
check whether intended price of tx is multiple of pre-configured ticksize
add governance proposal to change tick size
allow pass a ticksize to override default one when proposing a new pair, `the problem is ticksize is to underlying orderbook while pair is for contract, do we want ticksize different for different contract?` now change to per pair per contract
another irrelevant echo is we would need an asset-list for dynamically-added assets instead of hard-coding denoms as enum which @Kbhat1 has been mentioning and working on, which requires add another field of ticksize when register a new pair to define ticksize (since currently when init dex module, the default ticksize is only set for hardcoded asset in the enum)


x/dex/keeper/tick.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"github.com/cosmos/cosmos-sdk/store/prefix"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/sei-protocol/sei-chain/x/dex/types"
+)
+
+// pair -> tick size
+func (k Keeper) SetTickSizeForPair(ctx sdk.Context, pair types.Pair, ticksize sdk.Dec) {
```
codchen: I think we should allow different tick sizes for the same pair across different contracts here just in case we need to in the future, which might involve messy migrations.

codchen (approved):

PR #82, opened 2022-07-04
Sei-chain testnet node setup with cosmovisor.

snipeTR:
This script uses the sei daemon program from version 1.0.2beta to version 1.0.6beta; downloads, builds, installs.
It downloads and installs all necessary tools and programs for the zero linux operating system.
In addition, you can learn some useful commands for seid with the helpsei command.

### DIFF
--- a/sei_for_cosmovisor
+++ b/sei_for_cosmovisor
@@ -1,5 +1,5 @@
 #!/bin/bash
-#v0.1alpha
+#v0.2alpha
 #snipeTR 2022 dscrd:snipeTR#8374
 #karboran 2022 dscrd:karboran#2719
 #Apache License Version 2.0
@@ -112,7 +112,7 @@ echo -e "\e[1m\e[32m2. installing tools and libs... \e[0m" && sleep 1
 sudo apt install curl tar wget clang pkg-config libssl-dev jq build-essential bsdmainutils git make ncdu gcc git jq chrony tmux net-tools liblz4-tool -y
 
 sleep 1
-cd "$HOME" && echo "$HOME" dizinine girilemiyor && sleep 10 && exit 13
+cd "$HOME" || echo "$HOME" dizinine girilemiyor && sleep 10 && exit 13
 wget -O go1.18.1.linux-amd64.tar.gz https://golang.org/dl/go1.18.1.linux-amd64.tar.gz
 rm -rf /usr/local/go && tar -C /usr/local -xzf go1.18.1.linux-amd64.tar.gz && rm go1.18.1.linux-amd64.tar.gz
 echo 'export GOROOT=/usr/local/go' >> "$HOME"/.bash_profile
@@ -125,7 +125,7 @@ source "$HOME"/.bash_profile
 echo -e "\e[1m\e[32m3. sei-chain library download and install... \e[0m"
 
 # download binary
-cd "$HOME" && echo "$HOME" dizinine girilemiyor && sleep 10 && exit 13
+cd "$HOME" || echo "$HOME" dizinine girilemiyor && sleep 10 && exit 13
 echo git clone --depth 1 --branch "$SEIDVER" https://github.com/sei-protocol/sei-chain.git
 git clone --depth 1 --branch "$SEIDVER" https://github.com/sei-protocol/sei-chain.git
 cd sei-chain && make install
@@ -286,7 +286,7 @@ sudo systemctl daemon-reload
 #sudo systemctl restart seid
 
 #cosmovisor build
-cd "$HOME" && echo "$HOME" dizinine girilemiyor && sleep 10 && exit 13
+cd "$HOME" || echo "$HOME" dizinine girilemiyor && sleep 10 && exit 13
 rm -rf ./cosmos-sdk
 #git clone git@github.com:cosmos/cosmos-sdk
 git clone --depth 1 --branch main https://github.com/cosmos/cosmos-sdk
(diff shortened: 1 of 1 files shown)


snipeTR:
Could you please help me test it on an empty linux server and make the necessary improvements?

sei_for_cosmovisor
```diff
@@ -0,0 +1,537 @@
...
+source "$HOME"/.bash_profile
+
+echo -e "\e[1m\e[32m3. sei-chain library download and install... \e[0m"
+
+# download binary
+cd $HOME || { echo "$HOME dizinine girilemiyor"; sleep 10; exit 13;}
+echo git clone --depth 1 --branch "$SEIDVER" https://github.com/sei-protocol/sei-chain.git
+git clone --depth 1 --branch "$SEIDVER" https://github.com/sei-protocol/sei-chain.git
+cd sei-chain && make install
+go build -o build/seid ./cmd/seid
```
philipsu522: make install also works

sei_for_cosmovisor
```diff
@@ -0,0 +1,537 @@
...
+
+#unnecessary maybe necessary
+#chmod +x ./build/seid && sudo cp ./build/seid /usr/local/bin/seid
+
+sleep 1
```
philipsu522: make install & go build are both blocking calls, don't think we need to sleep?

sei_for_cosmovisor
```diff
@@ -0,0 +1,537 @@
...
+
+#impornant
+cp "$HOME"/go/bin/seid /usr/local/bin/
+
+mv "$HOME"/.sei-chain "$HOME"/.sei
```
philipsu522: there should not be a ~/.sei-chain folder anymore, should all be in ~/.sei

sei_for_cosmovisor
```diff
@@ -0,0 +1,537 @@
...
+
+# config
+seid config chain-id "$CHAIN_ID"
+seid config keyring-backend file
+
+# init
+seid init "$NODENAME" --chain-id "$CHAIN_ID"
+
+# download genesis and addrbook
+wget -qO "$HOME"/.sei/config/genesis.json "https://raw.githubusercontent.com/sei-protocol/testnet/main/sei-testnet-2/genesis.json"
```
philipsu522: we may want to add a CHAIN_ID variable so we can reuse this script.
E.g. 
CHAIN_ID=sei-testnet-2
`"https://raw.githubusercontent.com/sei-protocol/testnet/main/$CHAIN_ID/genesis.json"`

sei_for_cosmovisor
```diff
@@ -0,0 +1,537 @@
...
+wget -qO "$HOME"/.sei/config/addrbook.json "https://raw.githubusercontent.com/sei-protocol/testnet/main/sei-testnet-2/addrbook.json"
+
+#Tuning config.toml
```
philipsu522, line 171: we probably want to remove these config changes, since we'll want to let users tune this themselves

sei_for_cosmovisor
```diff
@@ -0,0 +1,537 @@
...
+  printf "\t timeout_prevote = \"100ms\"\n"
+  printf "\t timeout_precommit = \"100ms\"\n"
+  printf "\t timeout_commit = \"100ms\"\n"
+  printf "\t skip_timeout_commit = true\n"
+  read -r -s -n 1 -p "Press any key to EXIT . . ."
+  exit 13
+fi
+
+# set minimum gas price
+sed -i -e "s/^minimum-gas-prices *=.*/minimum-gas-prices = \"0usei\"/" "$HOME"/.sei/config/app.toml
```
philipsu522: we do not want to set min-gas-price to 0 to prevent spamming

sei_for_cosmovisor
```diff
@@ -0,0 +1,532 @@
...
+  SEIDBUILDVER=$("$DAEMON_HOME"/cosmovisor/upgrades/"$SEIDVER"/bin/seid version)
+  if [ "$SEIDBUILDVER" == "$SEIDVER" ]; then
+  	echo "$DAEMON_HOME"/cosmovisor/upgrades/"$SEIDVER"/bin/seid version "$SEIDVER"
+  else
+    	echo -e "\e[1m\e[31m2. Error version not match $SEIDVER \e[0m"
+    	echo -e "\e[1m\e[31m2. please check $DAEMON_HOME/cosmovisor/upgrades/$SEIDVER/bin version file \e[0m"
+    	read -r -s -n 1 -p "Press any key to EXIT . . ."
+  	exit 13
+  fi
+fi
```
philipsu522, line 420: all of this logic seems to be repeated for each version. Let's make a bash function and pass in the SEIDVER so that we can reduce duplicate code

sei_for_cosmovisor
```diff
@@ -0,0 +1,532 @@
...
+sed -i -e "s/^pruning-keep-recent *=.*/pruning-keep-recent = \"$pruning_keep_recent\"/" "$HOME"/.sei/config/app.toml
+sed -i -e "s/^pruning-keep-every *=.*/pruning-keep-every = \"$pruning_keep_every\"/" "$HOME"/.sei/config/app.toml
+sed -i -e "s/^pruning-interval *=.*/pruning-interval = \"$pruning_interval\"/" "$HOME"/.sei/config/app.toml
+
+
+#When you are going to run multiple NODEs on the same server, please change the ports using the commands below to avoid port conflict.
+#Change port 37 
+#------------------------------
+#26658 to 36378
+port_a=26658
```
philipsu522, line 227: can we have more descriptive names for these ports? like port_a -> proxy_app

sei_for_cosmovisor
philipsu522: bump on this - don't htink we want to make config changes

philipsu522 (commented):
let's also move this script into https://github.com/sei-protocol/sei-chain/tree/master/scripts

PR #83, opened 2022-07-05
Create cosmos_sei_installer.sh

snipeTR:
This script uses the sei daemon program from version 1.0.2beta to version 1.0.6beta; downloads, builds, installs.
It downloads and installs all necessary tools and programs for the zero linux operating system.
In addition, you can learn some useful commands for seid with the helpsei command.


faddat:
I don't recommend using scripts like these.

PR #86, opened 2022-07-07
Add Dex Module AssetList

Kbhat1:


proto/dex/asset_list.proto
```diff
@@ -0,0 +1,27 @@
...
+message AssetIBCInfo {
+    string sourceChannel = 1 [(gogoproto.jsontag) = "source_channel"];
+    string dstChannel = 2 [(gogoproto.jsontag) = "dst_channel"];
+    string sourceDenom = 3 [(gogoproto.jsontag) = "source_denom"];
+    string sourceChainID = 4 [(gogoproto.jsontag) = "source_chain_id"];
+}
+
+message AssetMetadata {
+    AssetIBCInfo ibcInfo = 1 [
+        (gogoproto.nullable) = false,
```
codchen: would native asset like `sei` require an ibcinfo too?

x/dex/keeper/asset_list.go
```diff
@@ -0,0 +1,41 @@
+package keeper
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/sei-protocol/sei-chain/x/dex/types"
+)
+
+func (k Keeper) SetAssetMetadata(ctx sdk.Context, assetMetadata types.AssetMetadata) {
+	store := ctx.KVStore(k.storeKey)
+	// Even if asset exists already, overwrite the store with new metadata
```
codchen, line 11: Maybe add a comment saying that we only allow one asset per base for now

proto/dex/asset_list.proto
Kbhat1: It wouldn't require the ibcinfo. I just updated it and allowed `ibcInfo` to be nullable for this case

x/dex/keeper/asset_list.go
Kbhat1: Added on latest commit

PR #87, opened 2022-07-07
[Sei] refactor order representation by order ID

codchen:
PR is large because references to orders are ubiquitous in `dex` module. There are no changes to the order matching core logic though, and most changes are just refactor of data structures.

- Store individual orders by order ID (uint64). Specifically, all orders and cancels will be stored immutably, and each account has an index to a list of active order IDs. New orders will have IDs added to the index, whereas full executions and cancellation will have IDs removed from the index.
- Unified order representation as `order.pb.go` and deprecated `OrderPlacement`, `dexcache.Order`, `LiquidationOrder`, and `ContractOrderPlacement`.
- Store order IDs instead of regex strings as allocation identifiers on the order book.
- Hide product-specific details like `leverage` and `position effect` in an opaque `Data` field that will be passed through to the contract. In other words, `dex` doesn't need to know about `leverage` etc. anymore.

A corresponding change will be made on Vortex's side and verified via integration tests. This PR should not be merged until that happens. [EDIT: verified in https://github.com/sei-protocol/vortex-contract/pull/78]


proto/dex/order.proto
```diff
@@ -0,0 +1,68 @@
...
+    ];
+    OrderStatus status = 2 [
+        (gogoproto.jsontag)    = "status",
+        (gogoproto.nullable)   = true
+    ];
+    string account = 3 [
+        (gogoproto.jsontag)    = "account",
+        (gogoproto.nullable)   = true
+    ];
+    string contractAddr = 4 [
```
LCyson, line 22: how does other projects interact with order placement rn

x/dex/keeper/msg_server_place_orders.go
```diff
@@ -60,32 +60,23 @@ func (k msgServer) PlaceOrders(goCtx context.Context, msg *types.MsgPlaceOrders)
 		return nil, err
 	}
 
-	pairToOrderPlacements := k.OrderPlacements[msg.GetContractAddr()]
+	contractBlockOrders := k.BlockOrders[types.ContractAddress(msg.GetContractAddr())]
 
-	
 	nextId := k.GetNextOrderId(ctx)
```
LCyson, line 51: is this a uuid?

x/dex/keeper/order.go
```diff
@@ -0,0 +1,100 @@
+package keeper
```
LCyson, line 1: can we add a test for the keeper store

proto/dex/order_entry.proto
```diff
@@ -20,19 +19,28 @@ message OrderEntry {
...
-      (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
-      (gogoproto.nullable)   = false,
-      (gogoproto.jsontag) = "allocation"
-  ];
-  Denom priceDenom = 5 [
+  string priceDenom = 4 [
     (gogoproto.jsontag) = "price_denom"
   ];
-  Denom assetDenom = 6 [
+  string assetDenom = 5 [
```
LCyson, line 28: ?

x/dex/module.go
```diff
@@ -270,59 +271,68 @@ func (am AppModule) endBlockForContract(ctx sdk.Context, contractAddr string) {
...
-		exchange.CancelOrders(ctx, orders.CancelBuys, allExistingBuys, types.PositionDirection_LONG, &longDirtyPrices)
-		exchange.CancelOrders(ctx, orders.CancelBuys, allExistingSells, types.PositionDirection_SHORT, &shortDirtyPrices)
 
-		settlements := []*types.Settlement{}
+		originalOrdersToCancel := am.keeper.GetOrdersByIds(ctx, contractAddr, cancels.GetIdsToCancel())
+		exchange.CancelOrders(ctx, *cancels, allExistingBuys, originalOrdersToCancel, &longDirtyPrices)
+		exchange.CancelOrders(ctx, *cancels, allExistingSells, originalOrdersToCancel, &shortDirtyPrices)
+
+		settlements := []*types.SettlementEntry{}
+		zeroOrders := []exchange.AccountOrderId{}
```
LCyson, line 292: what's zeroorders here

x/dex/keeper/msg_server_place_orders.go
LCyson: ever-increasing uint64

LCyson:
TLDR, options for upgrading:
1. In migration handler, clear all previous states [Best option]
2. Upgrade before anyone starts using the dex module on incentivized testnet
3. Complete migration handler logic
4. Start a new chain

x/dex/module.go
codchen: added a comment

x/dex/keeper/order.go
codchen: added

proto/dex/order.proto
```diff
@@ -0,0 +1,68 @@
...
+    string assetDenom = 8 [
+        (gogoproto.jsontag) = "asset_denom"
+    ];
+    OrderType orderType = 9 [
+        (gogoproto.jsontag) = "order_type"
+    ];
+    PositionDirection positionDirection = 10 [
+        (gogoproto.jsontag) = "position_direction"
+    ];
+    string data = 11 [
```
LCyson, line 50: what's the data here

utils/slice.go
```diff
@@ -0,0 +1,11 @@
+package utils
```
LCyson, line 1: can we add a test for this

x/dex/types/keys.go
```diff
@@ -96,15 +111,19 @@ const (
 	ShortBookCountKey = "ShortBook-count-"
 )
 
-
+const (
```
LCyson: nit: why not merge the these two const closure as both of they hold constants for key

x/dex/cache/cache.go
```diff
@@ -1,7 +1,7 @@
 package dex
```
LCyson, line 1: can we add a test for this file

x/dex/cache/cache.go
```diff
@@ -18,225 +18,158 @@ const (
...
-	Creator    string
-	OrderType  types.OrderType
-	Direction  types.PositionDirection
-	Effect     types.PositionEffect
-	PriceDenom string
-	AssetDenom string
-	Leverage   sdk.Dec
-}
+// All new orders attempted to be placed in the current block
+type BlockOrders []types.Order
```
LCyson: can we move this to the top of the `MemState` definition

x/dex/module.go
```diff
@@ -270,59 +258,69 @@ func (am AppModule) endBlockForContract(ctx sdk.Context, contractAddr string) {
...
 	for _, pair := range registeredPairs {
-		am.keeper.OrderCancellations[contractAddr][pair.String()] = dexcache.NewOrderCancellations()
-		orders := am.keeper.Orders[contractAddr][pair.String()]
-		ctx.Logger().Info(pair.String())
-		ctx.Logger().Info(fmt.Sprintf("Orders %s", am.keeper.Orders))
-		ctx.Logger().Info(fmt.Sprintf("Number of LB: %d, LS: %d, MB: %d, MS: %d", len(orders.LimitBuys), len(orders.LimitSells), len(orders.MarketBuys), len(orders.MarketSells)))
-		allExistingBuys := am.keeper.GetAllLongBookForPair(ctx, contractAddr, pair.PriceDenom, pair.AssetDenom)
-		allExistingSells := am.keeper.GetAllShortBookForPair(ctx, contractAddr, pair.PriceDenom, pair.AssetDenom)
+		typedPairStr := types.GetPairString(&pair)
+		orders := am.keeper.MemState.GetBlockOrders(typedContractAddr, typedPairStr)
```
LCyson, line 271: random q, theoretically this memstate is still redundant as the keeper also doesn't store directly to disk right? what's the reason we're keeping the MemState layer rn?

x/dex/types/alias.go
```diff
@@ -0,0 +1,12 @@
+package types
```
LCyson, line 1: can we add a test for this file

x/dex/types/keys.go
```diff
@@ -73,6 +76,18 @@ func TickSizeKeyPrefix(contractAddr string) []byte {
 	return append(KeyPrefix(TickSizeKey), KeyPrefix(contractAddr)...)
 }
 
+func OrderPrefix(contractAddr string) []byte {
```
LCyson, line 79: nit: add tests

LCyson (changes requested):
lgtm pending small tests

proto/dex/order.proto
codchen: any product-specific data that a contract needs but dex doesn't need to know

x/dex/module.go
codchen: we need memstate just to retain information from `DeliverTx` to `EndBlock`. It can be discarded after `EndBlock` runs

x/dex/types/alias.go
codchen: added

x/dex/types/keys.go
codchen: added

x/dex/cache/cache.go
codchen: :+1:

x/dex/cache/cache.go
codchen: added

x/dex/types/keys.go
codchen: :+1:

utils/slice.go
codchen: added

proto/dex/order.proto
LCyson: why this is a string in that case, shouldn't this be an object with needed fields? or the idea here is to parse the data string into a json, does it guarantee backwards/forwards compatibility in cosmwasm?

LCyson (approved):

proto/dex/order.proto
LCyson: oh saw it in the other pr, will ask offline if i still have q after that

PR #91, opened 2022-07-10
Update docs.go

karborantr:
add comment


BrandonWeng (approved):

BrandonWeng:
The comment's pretty redundant, I don't think we need this

PR #95, opened 2022-07-11
Add atomic handling to `dex` EndBlock

codchen:
Per feedback from external partners who wish to build on top of `sei`, it's a significant problem if contract developers need to worry about ensuring atomicity. Specifically, there were two scenarios where atomicity may not hold:
1. between `dex` transactions and `dex` EndBlock
2. operations within `dex` EndBlock (e.g. across different `sudo` calls).

This PR proposes a solution to the aforementioned challenges:
1. `dex` transactions will no longer cause direct state changes. Instead, all potential state changes are cached into a `MemState` data structure within `dex` keeper and will not be materialized into KV store until `EndBlock`. `MemState` has lifetime of a block.
2. `dex` EndBlock now ensures per-contract atomicity. Specifically, it will attempt to process everything on top of a cached KV store (note that this has nothing to do with `MemState` mentioned above) and a copy of `MemState`, and only flush to persistent KV store if everything is successful. If not, it will filter out contracts that failed as well as any data in `MemState` associated with the failed contracts, use a fresh cached KV store and `MemState` copy (without the failed contracts' data) to reprocess. This process will continue until either success or all contracts are filtered out, in which case nothing will be written to persistent KV during this block.

Tested via both automated tests as well as local chain


codchen:
Commenting out loadtest files since it's causing linter to fail. Will fix loadtest in a separate PR

loadtest/main.go
```diff
@@ -1,226 +1,226 @@
...
-		numberOfBlocks,
-		longPriceFloor,
-		longPriceCeiling,
-		shortPriceFloor,
-		shortPriceCeiling,
-		quantityFloor,
-		quantityCeiling,
-	)
-}
+// import (
```
philipsu522: do we want to uncomment this? may be related to what @faddat was seeing being unable to build this code

x/dex/module.go
```diff
@@ -246,25 +245,93 @@ func (am AppModule) beginBlockForContract(ctx sdk.Context, contractAddr string)
...
-		am.endBlockForContract(ctx, contractAddr)
+	validContractAddresses := map[string]types.ContractInfo{}
+	for _, contractInfo := range am.getAllContractInfo(ctx) {
+		validContractAddresses[contractInfo.ContractAddr] = contractInfo
+	}
+	// Each iteration is atomic. If an iteration finishes without any error, it will return,
+	// otherwise it will rollback any state change, filter out contracts that cause the error,
+	// and proceed to the next iteration. The loop is guaranteed to finish since
+	// `validContractAddresses` will always decrease in size every iteration.
+	for len(validContractAddresses) > 0 {
```
philipsu522, line 257: why do we need to loop here? seems like nothing is added to validContractAddress in the body, so for each contract wouldn't there only be 2 outcomes:
1. it fails -> deleted in ln302
2. succeeded. We'd probalby run some other logic to commit it. Do we need to loop again?

loadtest/main.go
codchen: this was commented to unblock linter CI for this PR, since the loadtest files aren't updated to match the new data structures. I plan to do that in a followup PR to not bloat up this PR too much

x/dex/module.go
codchen: @philipsu522 so if it fails for *some contracts*, we will delete those contracts from `validContractAddress` and rollback the store (not really rollback but more like discard the corrupt cached store and use a fresh one), and rerun the logics for contracts that are not failing. If there are no inter-dependencies among contracts, this loop would iterate at most twice (the first time it would weed out all contracts that are failing, and the second time it would succeed for all remaining). But if there is any inter-dependency (which we don't have a way to prevent), some contracts that succeeded in the first iter might fail in the second iter, so we keep the loop going until either all remaining succeed or everyone fails.

x/dex/keeper/msg_server_register_contract.go
```diff
@@ -9,13 +9,15 @@ import (
...
+	for _, contract := range k.GetAllContractInfo(ctx) {
+		if msg.Contract.ContractAddr == contract.ContractAddr {
 			return &types.MsgRegisterContractResponse{}, nil
 		}
 	}
-	contractAddr := msg.Contract.ContractAddr
-	k.SetContractAddress(ctx, contractAddr, msg.Contract.CodeId)
+	k.SetContract(ctx, msg.Contract)
+	if msg.Contract.HookOnly {
+		return &types.MsgRegisterContractResponse{}, nil
```
LCyson: this is the same return as line 22, do we intend to return a different response here?

x/dex/keeper/end_block_place_orders.go
```diff
@@ -16,19 +16,28 @@ import (
...
+func (k *Keeper) HandleEBPlaceOrders(ctx context.Context, sdkCtx sdk.Context, tracer *otrace.Tracer, contractAddr string, registeredPairs []types.Pair) error {
 	_, span := (*tracer).Start(ctx, "SudoPlaceOrders")
 	span.SetAttributes(attribute.String("contractAddr", contractAddr))
 
 	typedContractAddr := types.ContractAddress(contractAddr)
 	msgs := k.GetPlaceSudoMsg(sdkCtx, typedContractAddr, registeredPairs)
-	k.CallContractSudo(sdkCtx, contractAddr, msgs[0]) // deposit
+	_, err := k.CallContractSudo(sdkCtx, contractAddr, msgs[0]) // deposit
+	if err != nil {
+		return nil
```
LCyson: return err?

x/dex/keeper/end_block_place_orders.go
```diff
@@ -16,19 +16,28 @@ import (
...
+	}
 
 	responses := []types.SudoOrderPlacementResponse{}
 	for _, msg := range msgs[1:] {
-		data := k.CallContractSudo(sdkCtx, contractAddr, msg)
+		data, err := k.CallContractSudo(sdkCtx, contractAddr, msg)
+		if err != nil {
+			return err
```
LCyson, line 36: may be better to add a logger here?

x/dex/keeper/end_block_liquidation.go
```diff
@@ -11,15 +11,21 @@ import (
...
+func (k *Keeper) HandleEBLiquidation(ctx context.Context, sdkCtx sdk.Context, tracer *otrace.Tracer, contractAddr string, registeredPairs []types.Pair) error {
 	_, liquidationSpan := (*tracer).Start(ctx, "SudoLiquidation")
 	liquidationSpan.SetAttributes(attribute.String("contractAddr", contractAddr))
 
 	typedContractAddr := types.ContractAddress(contractAddr)
 	msg := k.getLiquidationSudoMsg(typedContractAddr)
-	data := k.CallContractSudo(sdkCtx, contractAddr, msg)
+	data, err := k.CallContractSudo(sdkCtx, contractAddr, msg)
+	if err != nil {
+		return err
```
LCyson, line 22: ditto: add logger

x/dex/keeper/end_block_cancel_orders.go
```diff
@@ -9,14 +9,17 @@ import (
...
-func (k *Keeper) HandleEBCancelOrders(ctx context.Context, sdkCtx sdk.Context, tracer *otrace.Tracer, contractAddr string, registeredPairs []types.Pair) {
+func (k *Keeper) HandleEBCancelOrders(ctx context.Context, sdkCtx sdk.Context, tracer *otrace.Tracer, contractAddr string, registeredPairs []types.Pair) error {
 	_, span := (*tracer).Start(ctx, "SudoCancelOrders")
 	span.SetAttributes(attribute.String("contractAddr", contractAddr))
 
 	typedContractAddr := types.ContractAddress(contractAddr)
 	msg := k.getCancelSudoMsg(typedContractAddr, registeredPairs)
-	_ = k.CallContractSudo(sdkCtx, contractAddr, msg)
+	if _, err := k.CallContractSudo(sdkCtx, contractAddr, msg); err != nil {
+		return err
```
LCyson, line 21: ditto: add logger

x/dex/keeper/contract.go
```diff
@@ -1,35 +1,40 @@
...
 		ctx.KVStore(k.storeKey),
 		[]byte(PrefixKey),
 	)
-	bz := make([]byte, 8)
-	binary.BigEndian.PutUint64(bz, codeID)
-	ctx.Logger().Info(fmt.Sprintf("Setting contract address %s", contractAddr))
-	store.Set(contractKey(contractAddr), bz)
+	bz, err := contract.Marshal()
+	if err != nil {
+		panic(err)
```
LCyson: why don't we return err instead of panic here

x/dex/module.go
philipsu522: Gotcha. I'm just worried some edge case happens that we don't catch where this loop ends up going forever. A cleaner way might be to explicitly recall the method with a flag that's
```
second_call_remove_invalid=False
```
and when it's the second time, pass it true. That way we guarantee it's only called twice. I"ll leave that up to you thought!

philipsu522 (approved):

x/dex/keeper/begin_block_new_block_test.go
```diff
@@ -0,0 +1,17 @@
...
+	"github.com/sei-protocol/sei-chain/x/dex/keeper"
+)
+
+const SupportedFeatures = "iterator,staking,stargate"
+const TestContract = "sei14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9sh9m79m"
+
+func TestHandleBBNewBlock(t *testing.T) {
+	ctx, wasmkeepers := wasmkeeper.CreateTestInput(t, false, SupportedFeatures)
+	dexKeeper := keeper.Keeper{WasmKeeper: *wasmkeepers.WasmKeeper}
+	dexKeeper.HandleBBNewBlock(ctx, TestContract, 1)
```
LCyson, line 18: what we're testing here?

x/dex/module.go
```diff
@@ -246,25 +245,93 @@ func (am AppModule) beginBlockForContract(ctx sdk.Context, contractAddr string)
...
+		cachedCtx, msCached := store.GetCachedContext(ctx)
+		// cache keeper in-memory state
+		memStateCopy := am.keeper.MemState.DeepCopy()
+		finalizeBlockMessages := map[string]*types.SudoFinalizeBlockMsg{}
+		for contractAddr := range validContractAddresses {
+			finalizeBlockMessages[contractAddr] = types.NewSudoFinalizeBlockMsg()
+		}
+
+		for contractAddr, contractInfo := range validContractAddresses {
+			if contractInfo.HookOnly {
```
LCyson: the variable name hookOnly may be a little bit confusing, cuz every contract will be able to return a hook (callback), but not each of them needs to process limit/market order. So maybe it's better to create two variables, one as `NeedCallBack` or ` NeedHook`, another one as `NeedOrderMatching` (default to true)

LCyson (changes requested):

x/dex/module.go
codchen: oh i meant that it's possible for have inter-dependency between contracts and thus possible for it to run more than two times. But i get what you are saying and I can probably add a counter such that it runs at most len(registeredContracts) times to guarantee termination

LCyson:
Probably we also need to think of whether this will expose a reentry exploits: https://docs.cosmwasm.com/docs/1.0/architecture/smart-contracts/#avoiding-reentrancy-attacks

codchen:
@LCyson think it should be fine since we still don't allow one contract to call another contract

x/dex/keeper/msg_server_register_contract.go
codchen: ah good catch. Not really the response should be the same

x/dex/keeper/end_block_place_orders.go
codchen: :+1:

x/dex/keeper/end_block_place_orders.go
codchen: yeah added

x/dex/keeper/contract.go
codchen: changed to returning error

x/dex/keeper/begin_block_new_block_test.go
codchen: added a comment

x/dex/module.go
codchen: sounds good. Updated

x/dex/keeper/contract.go
```diff
@@ -10,17 +11,18 @@ import (
...
 
-func (k Keeper) SetContract(ctx sdk.Context, contract *types.ContractInfo) {
+func (k Keeper) SetContract(ctx sdk.Context, contract *types.ContractInfo) error {
 	store := prefix.NewStore(
 		ctx.KVStore(k.storeKey),
 		[]byte(PrefixKey),
 	)
 	bz, err := contract.Marshal()
 	if err != nil {
-		panic(err)
```
LCyson: nit: add a logger here

x/dex/module.go
```diff
@@ -301,6 +305,12 @@ func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.Val
 			am.keeper.MemState.DeepFilterAccount(failedContractAddress)
 			delete(validContractAddresses, failedContractAddress)
 		}
+
+		iterCounter--
+		if iterCounter == 0 {
```
LCyson, line 310: why did we change len(validContractAddresses) to the iterCounter here?

LCyson (approved):

x/dex/module.go
codchen: this is more like a safeguard due to the discussion here https://github.com/sei-protocol/sei-chain/pull/95#discussion_r919624869

PR #96, opened 2022-07-11
Add getOrder endpoints

codchen:
Add endpoints to get orders:
- By ID
- By account

TODO: add json schemas in PR description


proto/dex/query.proto
```diff
@@ -53,6 +54,14 @@ service Query {
 		option (google.api.http).get = "/sei-protocol/seichain/dex/get_twaps/{contractAddr}/{lookbackSeconds}";
 	}
 
+	rpc GetOrders(QueryGetOrdersRequest) returns (QueryGetOrdersResponse) {
```
udpatil: should we name this `GetOrdersByAccount` for clarity?

proto/dex/query.proto
```diff
@@ -112,23 +121,25 @@ message QueryAllShortBookResponse {
 }
 
 message QueryGetSettlementsRequest {
-	string contractAddr = 1;
```
udpatil, line 132: Nit: Why don't we name the message fields in snake case so the derived Json stuff is snake case too? That's what I've been doing for other PRs

x/dex/client/cli/query_orders.go
```diff
@@ -0,0 +1,89 @@
...
+
+var _ = strconv.Itoa(0)
+
+func CmdGetOrders() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "get-orders [contract-address] [account]",
+		Short: "Query get orders for account",
+		Args:  cobra.ExactArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			reqContractAddr := args[0]
```
udpatil, line 20: Why do we require contract address for querying by account? Isn't it possible that there is a use case that would want to show all of a users orders? Or do we intend to keep contract specific orders clearly separated always?

x/dex/keeper/grpc_query_order.go
```diff
@@ -0,0 +1,59 @@
...
+	if found {
+		totalSettledQuantity := sdk.ZeroDec()
+		for _, settlement := range settlements.Entries {
+			totalSettledQuantity = totalSettledQuantity.Add(settlement.Quantity)
+		}
+		if totalSettledQuantity.GTE(order.Quantity) {
+			order.Status = types.OrderStatus_FULFILLED
+		}
+	}
+}
```
udpatil: If the order hasn't been cancelled or settled, what do we need to decorate order status with?

x/dex/keeper/grpc_query_order.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+	settlements, found := k.GetSettlementsState(c, order.ContractAddr, order.PriceDenom, order.AssetDenom, order.Id)
+	if found {
+		totalSettledQuantity := sdk.ZeroDec()
+		for _, settlement := range settlements.Entries {
+			totalSettledQuantity = totalSettledQuantity.Add(settlement.Quantity)
+		}
+		if totalSettledQuantity.GTE(order.Quantity) {
+			order.Status = types.OrderStatus_FULFILLED
+		}
```
udpatil: Do we need an order status for partial fulfillment? Probably not relevant for this PR, but just wondering for future order status behavior

x/dex/types/keys.go
```diff
@@ -55,10 +55,10 @@ func PricePrefix(contractAddr string) []byte {
 	return append(KeyPrefix(PriceKey), KeyPrefix(contractAddr)...)
 }
 
-func SettlementEntryPrefix(contractAddr string, blockHeight uint64) []byte {
+func SettlementEntryPrefix(contractAddr string, priceDenom string, assetDenom string) []byte {
 	return append(
 		append(KeyPrefix(SettlementEntryKey), KeyPrefix(contractAddr)...),
-		GetKeyForHeight(blockHeight)...,
+		PairPrefix(priceDenom, assetDenom)...,
```
udpatil, line 61: Do we need a store migration for this change to the key behavior?

udpatil: Also, why do we want to key on price + asset denom? is it just an effective way to partition the store keys further?

x/dex/keeper/grpc_query_order.go
```diff
@@ -0,0 +1,59 @@
...
+func (k Keeper) decorateOrderStatus(c sdk.Context, order *types.Order) {
+	cancels := k.GetCancelsByIds(c, order.ContractAddr, []uint64{order.Id})
+	if _, ok := cancels[order.Id]; ok {
+		order.Status = types.OrderStatus_CANCELLED
+		return
+	}
+	settlements, found := k.GetSettlementsState(c, order.ContractAddr, order.PriceDenom, order.AssetDenom, order.Id)
+	if found {
+		totalSettledQuantity := sdk.ZeroDec()
+		for _, settlement := range settlements.Entries {
```
udpatil: won't this become a costly process? would it be better to track whether the order is FULFILLED or not fully filled on the `Settlements` protobuf? maybe it doesnt matter if not many use cases need to perform this calculation

proto/dex/query.proto
codchen: we can, but even if it's in snakecase we still need to add `[(gogoproto.jsontag) = "twaps"]` for it to be encoded into json properly, since most of our structs use gogoproto

x/dex/client/cli/query_orders.go
codchen: the current use cases we have are all per-contract (e.g. user should only see their vortex orders when they are using vortex). if we want to add a hub-like product for users to view all activities across products we can add a global view endpoint per account then

x/dex/keeper/grpc_query_order.go
codchen: it's either placed or failed-to-place, which is set when the order is placed

x/dex/keeper/grpc_query_order.go
codchen: we can add one if product needs it

x/dex/types/keys.go
codchen: we will have a catch-all migration for all recent `dex` changes where the migration handler nukes everything under the `dex` namespace, since we don't have any real dex use cases yet

x/dex/keeper/grpc_query_order.go
codchen: yeah we can set it during write. I'll update the PR

proto/dex/query.proto
codchen: :+1:

x/dex/keeper/grpc_query_order.go
codchen: updated

proto/dex/query.proto
udpatil: hm, some of the oracle protos don't need that and are still generated properly 🤔

udpatil: oh wait, some of those do have `omitempty`

x/dex/keeper/grpc_query_order.go
```diff
@@ -34,26 +33,8 @@ func (k Keeper) GetOrders(c context.Context, req *types.QueryGetOrdersRequest) (
 	orders := k.GetOrdersByIds(ctx, req.ContractAddr, activeOrderIds.Ids)
 	response := &types.QueryGetOrdersResponse{Orders: []*types.Order{}}
 	for _, order := range orders {
-		k.decorateOrderStatus(ctx, &order)
+		order := order
```
udpatil, line 36: this is redundant, right?

udpatil (approved):
LGTM, left a nit

x/dex/keeper/grpc_query_order.go
codchen: @udpatil not really this is another thing that linter was complaining about. See https://stackoverflow.com/questions/62446118/implicit-memory-aliasing-in-for-loop for details

PR #99, opened 2022-07-12
456

faddat:
Non-state breaking PR that only updates cosmos sdk


BrandonWeng:
In master already: https://github.com/sei-protocol/sei-chain/blob/master/go.mod#L8C1-L11

PR #100, opened 2022-07-12
DNM: 46 (need your help to get it over the line)

faddat:
Hi, so this is almost entirely complete. There is one place where it is not and that is the staking keeper. What we would need to do is changed a little decimal math to use integers and I did not feel totally all right changing that for you I figured it's probably best to let you make those changes. Also I'm pretty sure that 46 will make the decentralized exchange features more maintainable over time.

Probably the most notable feature change though is the version of tendermint used in 46 supports mempool prioritization and if I had to guess that's going to be helpful as well.

cosmos sdk v46 upgrade

* state breaking
* no upgrade handlers for 46
* Uses latest ibc-go (v4)
* enforces the use of the grpc gateway because 46 doesn't support rest anymore


philipsu522:

PR #102, opened 2022-07-12
ibc-go 3.1.0

faddat:
This PR contains state-breaking changes.

#99 has been updated so that it is not state-breaking.


BrandonWeng:
Should already be in master https://github.com/sei-protocol/sei-chain/blob/master/go.mod#L8C1-L11

PR #103, opened 2022-07-12
use tweaked tm-db

faddat:
I think that if each validator on the testnet applies this version of tm-db, block times will be reduced further.

Little information about that build of tm-db:

It is much faster. We currently use it on our osmosis validator, our Juno validator and our evmos validator so it is quite fair to say that it is tested, but because it is a giant PR, it has not been merged into tm-db and can't really blame them for that.  The majority of the work comes from Notional and Terra.

In addition to that, this is what we used to host our RPCs. The feedback that we get from RPC users is that ours perform extremely well and I think that this is why.  So that's about another 100 nodes that use this successfully.

Likely of particular interest, is the fact that this made it possible for a number of evmos validators and evmos rpc providers to actually stay in sync up to the tip of the chain.


philipsu522:
closing this out in favor of seidb project we're releasing

PR #104, opened 2022-07-12
Add more checks `ValidateBasic` message method

lumtis:
`ValidateBasic` of `sdk.Msg` allows performing static checks for an sdk message.
These checks are performed before the message is broadcast to the network, therefore it is more optimized to have these checks in `ValidateBasic` rather that in the message handler to save gas for txs bound to fail.

This PR adds some more checks in message validate basic like contract address check or coins validity checks


faddat:
Super awesome! I just approved the CI for this and thanks so much for this PR.

PR #106, opened 2022-07-12
[epoch][wasmbindings] Add epoch query wasmbinding

udpatil:
Tested with localsei, I'll update with another commit adding some tests shortly


philipsu522 (approved):

philipsu522:
oops, accidentally closed (reopened it)

PR #112, opened 2022-07-13
Fix loadtest

codchen:
Fix loadtest to use the new data structure

### DIFF
--- a/loadtest/main.go
+++ b/loadtest/main.go
@@ -39,6 +39,7 @@ var (
 )
 
 const BATCH_SIZE = 100
+const VORTEX_DATA = "{\"position_effect\":\"Open\",\"leverage\":\"1\"}"
 
 var FROM_MILI = sdk.NewDec(1000000)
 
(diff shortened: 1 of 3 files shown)


loadtest/main.go
```diff
@@ -158,9 +163,7 @@ func run(
 		}
 		sendersList = append(sendersList, senders)
 
-		tmp := inactiveAccounts
-		inactiveAccounts = activeAccounts
-		activeAccounts = tmp
+		inactiveAccounts, activeAccounts = activeAccounts, inactiveAccounts
```
udpatil, line 166: does this actually work? wouldn't it assign `inactiveAccounts = activeAccounts` first and then the variable for active accounts is the same as a result? if not, kinda cool

udpatil (approved):

loadtest/main.go
philipsu522: this is like python, you can do multiple variable assignments in 1 line

philipsu522 (approved):

loadtest/main.go
codchen: yeah i didn't know it works but this is what the linter asked me to do lol

PR #117, opened 2022-07-15
calculate amount of collateral automatically when an order is placed

just-in-chang:


x/dex/keeper/msg_server_place_orders.go
```diff
@@ -44,6 +45,20 @@ func (k msgServer) PlaceOrders(goCtx context.Context, msg *types.MsgPlaceOrders)
...
+		calculatedCollateral = calculatedCollateral.Add(order.Price.Mul(order.Quantity))
+	}
+
+	if msg.AutoCalculateDeposit {
+		// throw error if current funds amount is less than calculatedCollateral
+		if msg.Funds[0].Amount.LT(calculatedCollateral.Mul(sdk.NewDecFromInt(sdk.NewInt(1000000))).RoundInt()) {
+			return nil, sdkerrors.Wrapf(sdkerrors.ErrInsufficientFunds, "insufficient funds to place order")
+		}
+		msg.Funds[0].Amount = calculatedCollateral.Mul(sdk.NewDecFromInt(sdk.NewInt(1000000))).RoundInt()
+		msg.Funds[0].Denom = msg.Orders[0].PriceDenom
```
codchen: normally the fund denom would be micro unit whereas price denom would be whole unit. We can add a helper like this in some util files:
```
func ConvertMicroToWholeDenom(amount sdk.Dec, denom string) (sdk.Dec, string) {
    assert denom starts with a `u`
    wholeDenom := denom[1:]
    assert wholeDenom is a valid denom (whitelisted in AssetList)
    return amount * 1000000, wholeDenom
}
```

x/dex/keeper/msg_server_place_orders.go
```diff
@@ -44,6 +45,20 @@ func (k msgServer) PlaceOrders(goCtx context.Context, msg *types.MsgPlaceOrders)
...
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
+	calculatedCollateral := sdk.NewDecFromBigInt(big.NewInt(0))
+	for _, order := range msg.Orders {
+		calculatedCollateral = calculatedCollateral.Add(order.Price.Mul(order.Quantity))
+	}
+
+	if msg.AutoCalculateDeposit {
+		// throw error if current funds amount is less than calculatedCollateral
+		if msg.Funds[0].Amount.LT(calculatedCollateral.Mul(sdk.NewDecFromInt(sdk.NewInt(1000000))).RoundInt()) {
```
codchen: I probably didn't describe this clearly enough, but we don't really care what's passed in `msg.Funds` by the user. As long as `msg. AutoCalculateDeposit` is true, and there are enough funds in the user's bank, we will execute line 58-59. To check user's bank balance you can do `k.Keeper.BankKeeper.GetBalance(ctx sdk.Context, addr sdk.AccAddress, denom string)`

x/dex/keeper/msg_server_place_orders.go
```diff
@@ -44,6 +47,27 @@ func (k msgServer) PlaceOrders(goCtx context.Context, msg *types.MsgPlaceOrders)
 
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
+	calculatedCollateral := sdk.NewDecFromBigInt(big.NewInt(0))
```
codchen: move inside the if block below

x/dex/keeper/msg_server_place_orders.go
```diff
@@ -44,6 +47,27 @@ func (k msgServer) PlaceOrders(goCtx context.Context, msg *types.MsgPlaceOrders)
...
+	for _, order := range msg.Orders {
+		calculatedCollateral = calculatedCollateral.Add(order.Price.Mul(order.Quantity))
+	}
+
+	if msg.AutoCalculateDeposit {
+		// throw error if current funds amount is less than calculatedCollateral
+		if msg.Funds[0].Amount.LT(k.Keeper.BankKeeper.GetBalance(ctx, sdk.AccAddress(msg.GetCreator()), msg.Orders[0].PriceDenom).Amount) {
```
codchen: if calculatedCollateral.GT(k.Keeper.BankKeeper.GetBalance(...))

codchen (approved):

PR #118, opened 2022-07-16
Add endpoint to query historical prices in candlestick form

codchen:
Request format: `/sei-protocol/seichain/dex/get_historical_prices/{contractAddr}/{priceDenom}/{assetDenom}/{periodLengthInSeconds}/{numOfPeriods}`
For example, if you want to get a price bar of size 30s and you want 120 bars (for an hour), `periodLengthInSeconds` would be 30 and `numOfPeriods` would be 120.

Response format:
```
{
  "prices": [
    {
      "begin_timestamp": 123450,
      "end_timestamp": 123480,
      "open": 100,
      "high": 102,
      "low": 99,
      "close": 101,
      "volume": 1,
    },
    {
      "begin_timestamp": 123420,
      "end_timestamp": 123450,
      "open": 99,
      "high": 100,
      "low": 99,
      "close": 1010
      "volume": 2,
    },
    ....
  ]
}
```

Prices are sorted in descending order based on time (i.e. the most recent price bar is at the top). In this PR we haven't populated `volume` yet so this field will always be 0 for the time being.

Also changed price snapshotting to be tick-based rather than epoch-based, and extended price snapshot retention to 24 hours.


x/dex/types/keys.go
```diff
@@ -51,8 +51,14 @@ func TwapPrefix(contractAddr string) []byte {
 	return append(KeyPrefix(TwapKey), KeyPrefix(contractAddr)...)
 }
 
-func PricePrefix(contractAddr string) []byte {
-	return append(KeyPrefix(PriceKey), KeyPrefix(contractAddr)...)
+func PricePrefix(contractAddr string, priceDenom string, assetDenom string) []byte {
```
LCyson, line 55: can we add an example for how the prefix is added in a comment, and also a test for this

x/dex/keeper/grpc_query_get_historical_prices.go
```diff
@@ -0,0 +1,120 @@
...
+	for i := range candlesticks {
+		candlesticks[i] = &types.PriceCandlestick{}
+		candlesticks[i].EndTimestamp = currentTimeStamp - uint64(i)*req.PeriodLengthInSeconds
+		candlesticks[i].BeginTimestamp = candlesticks[i].EndTimestamp - req.PeriodLengthInSeconds
+	}
+
+	// set open
+	pricePtr := 0
+	for i := range candlesticks {
+		for pricePtr < len(validPrices) && validPrices[pricePtr].SnapshotTimestampInSeconds > candlesticks[i].BeginTimestamp {
```
LCyson, line 54: nit: is it better to sort the validPrices in time ascending order so it can reduce some code duplication here? guess the computation is not so important as it's a query endpoint, i'm also good with changing this in a separate PR

LCyson (changes requested):
lgtm pending a small test and comment
Also are we going to revert this change after we have an off-chain data store?

x/dex/keeper/grpc_query_get_historical_prices.go
codchen: i intentionally kept open/high/low/close logic separate just to make logic clear for each. If needed I can also consolidate all four computations, but it might be less readable

x/dex/types/keys.go
codchen: added a comment and a test

LCyson (approved):

PR #124, opened 2022-07-18
Custom mint module impl

philipsu522:
### Description
Implement custom minting logic so that token release can be controlled by unix time (via x/epoch module). This is pretty similar to how Osmosis does it. Mainly, we will specify the following:
- `genesis_epoch_provisions` - total sei to add from genesis
- `reduction_period_in_epochs` - number of epochs before reducing rewards
- `reduction_factor` - fraction to reduce subsequent rewards by

It also removes old params like inflation, inflation_max...etc

### Testing
- Unit tests
- Verified genesis file:
```
238     "mint": {
239       "minter": {
240         "epoch_provisions": "0.000000000000000000"
241       },
242       "params": {
243         "mint_denom": "usei",
244         "genesis_epoch_provisions": "5000000.000000000000000000",
245         "reduction_period_in_epochs": "2",
246         "reduction_factor": "0.500000000000000000"
247       },
248       "halven_started_epoch": "0"
249     },
```
- Ran local chain with above parameters:
```
> cat ../mint_logs.txt| grep "minted\|epoch" | awk '{print $2, $3, $4, $5, $6, $7, $8, $9}' | uniq -c
   1 INF Current epoch 0
  12 INF Current epoch 1
  11 INF Current epoch 2
   1 INF minted coins from module account amount=2500000usei from=mint
  12 INF Current epoch 3
  12 INF Current epoch 4
   1 INF minted coins from module account amount=1250000usei from=mint
  12 INF Current epoch 5
  12 INF Current epoch 6
   1 INF minted coins from module account amount=625000usei from=mint
  12 INF Current epoch 7
  12 INF Current epoch 8
   1 INF minted coins from module account amount=312500usei from=mint
  12 INF Current epoch 9
  12 INF Current epoch 10
   1 INF minted coins from module account amount=156250usei from=mint
  12 INF Current epoch 11
  12 INF Current epoch 12
   1 INF minted coins from module account amount=78125usei from=mint
   3 INF Current epoch 13
```
- Queried total supply, ensured they added up to start + epoch provisions (property of geometric series)
```
> seid q bank total
pagination:
  next_key: null
  total: "0"
supply:
- amount: "100000000000004687500"
  denom: usei
> seid q mint epoch-provisions
312500.000000000000000000
```
- Ran with different configs incl:
  - different epoch provisions
  - different reduction rate
  - different epoch intervals


x/epoch/types/hooks.go
```diff
@@ -0,0 +1,39 @@
...
+func NewMultiEpochHooks(hooks ...EpochHooks) MultiEpochHooks {
+	return hooks
+}
+
+// AfterEpochEnd is called when epoch is going to be ended, epochNumber is the number of epoch that is ending.
+func (h MultiEpochHooks) AfterEpochEnd(ctx sdk.Context, epoch Epoch) {
+	for i := range h {
+		cacheCtx, write := ctx.CacheContext()
+		h[i].AfterEpochEnd(cacheCtx, epoch)
+		write()
```
codchen: cache would not be flushed only if `AfterEpochEnd` panicks, which would halt the chain anyway before `Commit`, so we might not need CacheContext here?

x/mint/keeper/keeper.go
```diff
@@ -16,14 +16,16 @@ type Keeper struct {
 	paramSpace       paramtypes.Subspace
 	stakingKeeper    types.StakingKeeper
 	bankKeeper       types.BankKeeper
+	epochKeeper      types.EpochKeeper
```
codchen: is this used?

codchen (approved):

codchen:
One caveat with the current approach is that it shifts the actual logic of minting from `mint` module to `epoch` module, so we need to be careful about `BeginBlock` ordering in app constructor. `epoch` now needs to go before modules that depend on minting results (distribution? not sure if there is any)

x/epoch/keeper/keeper.go
```diff
@@ -39,6 +40,17 @@ func NewKeeper(
 	}
 }
 
+// Set the gamm hooks.
```
LCyson: nit: wrong comment?

x/epoch/types/hooks.go
```diff
@@ -0,0 +1,39 @@
...
+type EpochHooks interface {
+	// the first block whose timestamp is after the duration is counted as the end of the epoch
+	AfterEpochEnd(ctx sdk.Context, epoch Epoch)
+	// new epoch is next block of epoch end block
+	BeforeEpochStart(ctx sdk.Context, epoch Epoch)
+}
+
+var _ EpochHooks = MultiEpochHooks{}
+
+// combine multiple gamm hooks, all hook functions are run in array sequence.
```
LCyson: ditto: gamm is a osmosis specific pool type

x/epoch/types/hooks.go
```diff
@@ -0,0 +1,39 @@
...
+type MultiEpochHooks []EpochHooks
+
+func NewMultiEpochHooks(hooks ...EpochHooks) MultiEpochHooks {
+	return hooks
+}
+
+// AfterEpochEnd is called when epoch is going to be ended, epochNumber is the number of epoch that is ending.
+func (h MultiEpochHooks) AfterEpochEnd(ctx sdk.Context, epoch Epoch) {
+	for i := range h {
+		cacheCtx, write := ctx.CacheContext()
```
LCyson: are the `AfterEpochEnd` and `BeforeEpochStart` supposed to break in any circumstance? i.e. when we do cacheContext and write() if the mint module's afterEpochEnd call fails somehow will the chain runs into a bad state indefinitely?

LCyson (approved):

x/epoch/types/hooks.go
philipsu522: good point, since we just panic i think it's fine to use original ctx

x/mint/keeper/keeper.go
philipsu522: nope, removed

x/epoch/types/hooks.go
philipsu522: right now, only AfterEpochEnd calls anything. The logic will panic which will halt the chain so we can go in and fix things: https://github.com/sei-protocol/sei-chain/pull/124/files#diff-5d06e3a8167fa5a91ac1040e8e5da549aaa66942648a558e517c38ab01cf71a2R32

philipsu522: \cc @codchen we were discussing this in his comment as well

philipsu522:
> One caveat with the current approach is that it shifts the actual logic of minting from `mint` module to `epoch` module, so we need to be careful about `BeginBlock` ordering in app constructor. `epoch` now needs to go before modules that depend on minting results (distribution? not sure if there is any)

good point. I don't think epoch has any dependencies so I put it first (since our x/upgrades module will depend on it too \cc @BrandonWeng)

PR #134, opened 2022-07-19
Remove register pair restriction

Kbhat1:
- remove register pair restriction to be in asset list


x/dex/keeper/msg_server_place_orders.go
```diff
@@ -58,6 +58,7 @@ func (k msgServer) PlaceOrders(goCtx context.Context, msg *types.MsgPlaceOrders)
 			return nil, sdkerrors.Wrapf(sdkerrors.ErrInsufficientFunds, "insufficient funds to place order")
 		}
 
+		// TODO: Should we still do this check? We still want price denom (deposits) to be in asset list
```
codchen: yeah I think we still want this check

codchen (approved):

x/dex/keeper/msg_server_place_orders.go
Kbhat1: Got you, leaving this in

PR #136, opened 2022-07-20
Add endpoint to simulate market order liquidity

codchen:
Add an endpoint that if provided order details will return how many units can be executed based on the current order book state.


LCyson (commented):
two questions:
1. how does the simulation make sure multiple quotes won't run into race condition? for example, contract A and contract B both quote 1 LONG order USDC<>SOL pair (spot market) at price 100,  and there are only 1 satisfied liquidity, both A and B seem to get an info that there are enough liquidity 
2. the algorithm seems to be expensive here, but given this won't be used frequently for now, think it's acceptable for us

codchen:
@LCyson since we don't have parallelism yet, there must be an ordering between contract A and B. If A runs first, it will place the order before B runs. That placed order will cause this liquidity check to return 0 when B runs

LCyson (approved):

PR #140, opened 2022-07-21
Update example proposals display type to micro denom

Kbhat1:
- Dex Asset List and Register Pair proposals should be the same denom as the price denom (accepted as deposits)
- Updates all display types (key for asset list) to micro denom e.g. "usei"

### DIFF
--- a/x/dex/example/register-pair-proposal.json
+++ b/x/dex/example/register-pair-proposal.json
@@ -6,18 +6,18 @@
         "contract_addr": "sei14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9sh9m79m",
         "pairs": [
           {
-            "price_denom": "usdc",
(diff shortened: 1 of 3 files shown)


x/dex/example/register-pair-proposal.json
```diff
@@ -6,18 +6,18 @@
         "contract_addr": "sei14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9sh9m79m",
         "pairs": [
           {
-            "price_denom": "usdc",
-            "asset_denom": "sei",
+            "price_denom": "uusdc",
```
codchen, line 9: kind of sad that json doesn't allow comments, but these denoms don't have to correspond to asset list denoms; they are denoms that the registering contract recognize internally

codchen (approved):

PR #142, opened 2022-07-21
Add registered pairs query to cmds

Kbhat1:
- Small PR fix to add registered pair queries to commands


philipsu522 (approved):
short n sweet

PR #143, opened 2022-07-21
Fix issues found during bug bash

codchen:
- Explicitly sort order book entries before matching. Long term we should avoid this sorting by making sure the entries are stored and iterated in order in the store.
- Mark order objects as cancelled if a market order is not filled

### DIFF
--- a/x/dex/module.go
+++ b/x/dex/module.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"sort"
 
 	"github.com/CosmWasm/wasmd/x/wasm"
 	"github.com/gorilla/mux"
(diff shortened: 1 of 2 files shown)


x/dex/module.go
```diff
@@ -346,6 +347,12 @@ func (am AppModule) endBlockForContract(ctx sdk.Context, contract types.Contract
 		assetDenomStr := pair.AssetDenom
 		allExistingBuys := am.keeper.GetAllLongBookForPair(ctx, contractAddr, priceDenomStr, assetDenomStr)
 		allExistingSells := am.keeper.GetAllShortBookForPair(ctx, contractAddr, priceDenomStr, assetDenomStr)
+		sort.Slice(allExistingBuys, func(i, j int) bool {
+			return allExistingBuys[i].GetPrice().LT(allExistingBuys[j].GetPrice())
+		})
+		sort.Slice(allExistingSells, func(i, j int) bool {
+			return allExistingSells[i].GetPrice().LT(allExistingSells[j].GetPrice())
```
LCyson, line 354: just to confirm, both bid and ask book should be in ascending order correct

LCyson (approved):

x/dex/module.go
codchen: yeah they should both be in ascending order

PR #145, opened 2022-07-22
Set status to be fulfilled if market order is fully executed

codchen:
Right now market order status would stay as `PLACED` even after it's fully executed. It makes more sense for its status to be updated to `FULFILLED` in such cases.


LCyson (approved):
qq, for cancelled orders, do we still persist them on the chain for 12hrs rn or we purge them once they're cancelled

codchen:
@LCyson right now we still persist them

PR #146, opened 2022-07-22
Add endpoint to get all settlements for an account

codchen:
Frontend needs an endpoint to get all settlements for an account under a certain pair. Settlements also need to have timestamp/height as part of it.

Frontend JS files will be generated in a followup PR to reduce this one's size


x/dex/keeper/settlement.go
```diff
@@ -11,29 +11,67 @@ import (
...
+		res = append(res, val)
+	}
+
+	return res
+}
+
+func (k Keeper) GetAllSettlementsState(ctx sdk.Context, contractAddr string, priceDenom string, assetDenom string, limit int) []types.Settlements {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.SettlementEntryPrefix(contractAddr, priceDenom, assetDenom))
+	res := []types.Settlements{}
+	iterator := sdk.KVStorePrefixIterator(store, []byte{})
```
LCyson, line 56: is there other access way rather than iterating through the KVStore? like getting a slice with length n instead?

LCyson (changes requested):
can we add tests for x/dex/keeper/settlement.go

x/dex/keeper/settlement.go
codchen: i think the point of iterator is so that we don't have to fetch everything at once. Since it's a key-value store there isn't much of an N+1 concern

codchen:
@LCyson added settlements tests

LCyson (approved):

PR #147, opened 2022-07-23
Assert that upgrade names are sorted

BrandonWeng:
Adds a check that the upgrade names are in alphabetical order before the handlers are registered. 

This will only prevent users from appending non-alphabetical ordering names. I think if we wanted to catch the case where users are inserting a "valid" upgrade name in between two past upgrades, we would need to know additional metadata (when it was added) or if an upgrade was already applied in the past. 

Note, I had to remove `"ignite-cli-removal-upgrade"` as it was conflicting with the alphabetical ordering assertion. It was causing the test to fail. 

### Testing
1. make install 
2. fresh instance of the chain 
3. propose 1.0.8beta at 25
4. restart chain with 1.0.8beta handler added  and proceeds fine 
5. propose 1.0.0beta and add handler to code after it stops
6. try to start and see that it errors out 
```
 ~/sei/brando  seid start                                                                                                             ✓  11266  01:25:46
2022/07/26 01:25:54 New upgrades must be appended to 'upgradesList' in alphabetical order
```

PRs are blocked if its not sorting: https://github.com/sei-protocol/sei-chain/pull/155

### Verify 
1. verify missing n-1 upgrade handler causes issues --> True
     - Reproduced the issue from this git issue: https://github.com/cosmos/cosmos-sdk/issues/11707 <-- Same class of issues
     - Saw the same error message reported 
5. verify unsorted causes issues  
     - Continue from above -- proposed `test-v1.0` at block 90
     - Stopped chain, removed `test-v0.9`  and added `test-v1.0`. See that it's still fine 
image
      - proposed test-v1.1 at height 145
      - stopped chain, removed all past handlers, and add v1.1 
      - make install and run seid start and see that it ran fine and proceeded 
      
image

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -21,7 +21,6 @@ import (
 	"github.com/cosmos/cosmos-sdk/server/config"
 	servertypes "github.com/cosmos/cosmos-sdk/server/types"
 	"github.com/cosmos/cosmos-sdk/simapp"
-	storetypes "github.com/cosmos/cosmos-sdk/store/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/module"
 	"github.com/cosmos/cosmos-sdk/version"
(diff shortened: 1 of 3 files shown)


app/app.go
```diff
@@ -657,7 +657,7 @@ func New(
 		panic(err)
 	}
 
-	if upgradeInfo.Name == Upgrade104 && !app.UpgradeKeeper.IsSkipHeight(upgradeInfo.Height) {
+	if upgradeInfo.Name == UpgradeNameOracleModule && !app.UpgradeKeeper.IsSkipHeight(upgradeInfo.Height) {
```
philipsu522: should we revert this?

philipsu522 (approved):

app/upgrades.go
```diff
@@ -9,57 +12,26 @@ import (
...
-// 1.0.4beta
-// this will introduce the oracle module as well
-const Upgrade104 = "1.0.4beta"
-
-// 1.0.5beta
-const Upgrade105 = "1.0.5beta upgrade"
-
-// 1.0.6beta
-const Upgrade106 = "1.0.6beta"
-
```
BrandonWeng, line 30: Should be fine to remove as shown in the test that we can remove historical handlers

philipsu522 (approved):

PR #148, opened 2022-07-25
Add endpoint to get market summary

codchen:
Frontend needs an endpoint to fetch last 24hr volume/high/low as well as last price. Ideally this should live offchain but before we have that we will put it onchain.

Volume fields will be populated once we start tracking volumes in `dex` (shouldn't be a blocker for act 2)


suujia (approved):
Looks good! From my side, only need codegen whenever you get to it :) But feel free to do that in a separate PR.

PR #153, opened 2022-07-26
topologically sort contract info in dex module

codchen:
- Allow contracts to specify dependencies during contract registration
- Allow contract to update its registered information
- Reject registration if the contract would introduce a circular dependency
- Topologically sort contracts to decide processing order in `dex` module Begin/EndBlock


x/dex/contract/dag.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/sei-protocol/sei-chain/utils"
+	"github.com/sei-protocol/sei-chain/x/dex/types"
+)
+
+type node struct {
+	contractAddr  string
+	incomingNodes *utils.StringSet
+}
+
+// Kahn's algorithm
```
philipsu522, line 13: wonder if this can be generalized for module sorting. Can be addressed in a new pr though

philipsu522 (commented):
🔥

philipsu522 (approved):
🔥

x/dex/contract/dag.go
codchen: yeah that should be possible. A bit sad that golang doesn't support generics so it might get a little verbose.

LCyson (approved):

PR #174, opened 2022-07-30
Refactor `dex` `Endblock` logic per contract

codchen:
Pure refactor with no intended business logic changes. Specifically, the old big `endBlockForContract` is divided into the following pieces for better readability, composability, and testability:
- PreExecutionHooks (e.g. `sudo` calls to contracts for order placements/cancellations and liquidation requests)
- Execution and orderbook update
- Order state update
- Cancel unfilled market orders
- Settlements


x/dex/keeper/abci/end_block_liquidation.go
```diff
@@ -66,7 +66,8 @@ func (w KeeperWrapper) PlaceLiquidationOrders(ctx sdk.Context, contractAddr stri
 		pair := types.Pair{PriceDenom: order.PriceDenom, AssetDenom: order.AssetDenom}
 		orders := w.MemState.GetBlockOrders(typesutils.ContractAddress(contractAddr), typesutils.GetPairString(&pair))
 		order.Id = nextID
-		orders.AddOrder(order)
+		orderCopy := order
```
LCyson, line 69: why do we do a deep copy here

LCyson (approved):

x/dex/keeper/abci/end_block_liquidation.go
codchen: yeah it's because we changed order cache from struct to a pointer to the struct in this PR

PR #175, opened 2022-07-31
Add more validation to order placement

philipsu522:
Chain crashes if orders include negative coins:
```
Jul 31 07:25:17 ip-172-31-25-17 seid[1168442]: 7:25AM ERR CONSENSUS FAILURE!!! err="negative coin amount: -1010000" module=consensus stack="goroutine 731 [running]:```
Add more checks on order. Verified on devnet-1 sei-node-1

### CODE REVIEW

File: x/dex/keeper/msgserver/msg_server_place_orders_test.go [hunk 1]
```diff
@@ -26,6 +26,8 @@ func TestPlaceOrder(t *testing.T) {
 				OrderType:         types.OrderType_LIMIT,
 				PriceDenom:        keepertest.TestPriceDenom,
 				AssetDenom:        keepertest.TestAssetDenom,
+				ContractAddr:      TestContract,
```
philipsu522: @codchen looks like contract addr and account is in 2 places here:
1. MsgPlaceOrders: https://github.com/sei-protocol/sei-chain/blob/master/x/dex/keeper/msgserver/msg_server_place_orders_test.go#L17
2. Orders (this change)
Is it possible to remove this from Orders and infer it from MsgPlaceOrders (or remove from MsgPlaceOrders)? I assume they should be the same. Would like to remove duplicate data if possible.

codchen: Account and contract address are already inferred here: https://github.com/sei-protocol/sei-chain/pull/175/files#diff-555639b7dfe25d62ab5dc8f8746b1588a98740aef6e622539386fcaec4508c04R89 and so we don’t need to validate those two fields in the validation method

codchen (approved):

codchen (approved):

LCyson (approved):
lgtm, we may need to validate more `nullable=false` case and also note that this one doesn't resolve the deposit is negative bug

PR #176, opened 2022-07-31
Catch all panics in DEX EndBlock

codchen:
For now we will catch all panics in EndBlock so that bugs won't cause the chain to be halted. Each panic will result in a metric being incremented with the error description as the label.

We will need to remove this catch before mainnet launch to prevent silent failing

### DIFF
--- a/x/dex/module.go
+++ b/x/dex/module.go
@@ -259,7 +259,7 @@ func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) (ret []abc
 		if err := recover(); err != nil {
 			ctx.Logger().Error(fmt.Sprintf("panic occurred in dex EndBlock: %s", err))
 			telemetry.IncrCounterWithLabels(
-				[]string{types.ModuleName, "endblockpanic"},
+				[]string{fmt.Sprintf("%s%s", types.ModuleName, "endblockpanic")},
 				1,
 				[]metrics.Label{
 					telemetry.NewLabel("error", fmt.Sprintf("%s", err)),
--- a/x/dex/module.go
+++ b/x/dex/module.go
@@ -254,10 +254,10 @@ func (am AppModule) beginBlockForContract(ctx sdk.Context, contract types.Contra
 // EndBlock executes all ABCI EndBlock logic respective to the capability module. It
 // returns no validator updates.
 func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) (ret []abci.ValidatorUpdate) {
-	cachedCtx, msCached := store.GetCachedContext(ctx)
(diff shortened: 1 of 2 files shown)


x/dex/module.go
```diff
@@ -251,9 +253,24 @@ func (am AppModule) beginBlockForContract(ctx sdk.Context, contract types.Contra
 
 // EndBlock executes all ABCI EndBlock logic respective to the capability module. It
 // returns no validator updates.
-func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.ValidatorUpdate {
+func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) (ret []abci.ValidatorUpdate) {
+	cachedCtx, msCached := store.GetCachedContext(ctx)
+	defer func() {
+		if err := recover(); err != nil {
+			ctx.Logger().Error(fmt.Sprintf("panic occurred in dex EndBlock: %s", err))
```
philipsu522: can we add module name here?

x/dex/module.go
```diff
@@ -251,9 +253,24 @@ func (am AppModule) beginBlockForContract(ctx sdk.Context, contract types.Contra
 
 // EndBlock executes all ABCI EndBlock logic respective to the capability module. It
 // returns no validator updates.
-func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.ValidatorUpdate {
+func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) (ret []abci.ValidatorUpdate) {
```
philipsu522, line 256: let's also add a todo here:
// TODO (codchen): Revert https://github.com/sei-protocol/sei-chain/pull/176/files before mainnet so we don't silently fail on errors

x/dex/module.go
```diff
@@ -293,14 +310,15 @@ func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.Val
...
-			if _, err := dexkeeperutils.CallContractSudo(cachedCtx, &am.keeper, contractAddr, finalizeBlockMsg); err != nil {
+			if _, err := dexkeeperutils.CallContractSudo(cachedCachedCtx, &am.keeper, contractAddr, finalizeBlockMsg); err != nil {
 				ctx.Logger().Error(fmt.Sprintf("Error calling FinalizeBlock of %s", contractAddr))
 				failedContractAddresses.Add(contractAddr)
 			}
 		}
 
 		// No error is thrown for any contract. This should happen most of the time.
 		if failedContractAddresses.Size() == 0 {
+			msCachedCached.Write()
```
philipsu522: curious why we need cache and cachcached

philipsu522 (approved):

x/dex/module.go
codchen: we don't actually need both. I removed the double cache

philipsu522 (approved):

PR #178, opened 2022-07-31
Add catch-all for all custom modules

codchen:
Since we have custom logics for:
- `epoch` BeginBlock
- `dex` BeginBlock and EndBlock
- `mint` BeginBlock
- `oracle` EndBlock

We will also be adding catch-all logics just for the testing phase so that the chain isn't halted frequently. These need to be removed before mainnet launch.

### DIFF
--- a/x/oracle/module.go
+++ b/x/oracle/module.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math/rand"
 
+	"github.com/armon/go-metrics"
 	"github.com/gorilla/mux"
 	"github.com/grpc-ecosystem/grpc-gateway/runtime"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


x/oracle/module.go
```diff
@@ -168,7 +170,22 @@ func (AppModule) ConsensusVersion() uint64 { return 4 }
...
+func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) (ret []abci.ValidatorUpdate) {
+	// TODO (codchen): Revert before mainnet so we don't silently fail on errors
+	defer func() {
+		if err := recover(); err != nil {
+			ctx.Logger().Error(fmt.Sprintf("panic occurred in %s EndBlock: %s", types.ModuleName, err))
+			telemetry.IncrCounterWithLabels(
+				[]string{fmt.Sprintf("%s%s", types.ModuleName, "endblockpanic")},
+				1,
+				[]metrics.Label{
+					telemetry.NewLabel("error", fmt.Sprintf("%s", err)),
```
LCyson, line 182: btw we probably wanna set up a bz hour alert on both endblock panic and begin block panic @philipsu522

LCyson (approved):

PR #180, opened 2022-08-02
Fixes for loadtest

philipsu522:
Fixes error on `/root/outputs` dir already existing when creating a file. Also uses "test" keyring backend so no password is required, and fixes homedir to be .sei


BrandonWeng (approved):
nice!

PR #181, opened 2022-08-02
Token Factory Module

Kbhat1:
- Adds in tokenFactory Module
- Note: Adding simulation + updated tests

For reviewers: Please take a look at ordering of tokenfactory module in app.go (in begin/end block, etc)

1. Please take a look at app.go changes especially and lmk if there is issues with ordering etc. I ordered it as osmosis did -which is to say they did not care about tokenfactory ordering
2. It charges a create denom fee which then goes to community pool - I'm not sure if we want to send it automatically to community pool but it is included (as part of distr module)
3. There is a TODO `https://github.com/sei-protocol/sei-chain/pull/181/files#:~:text=app.-,BankKeeper,-%2C%20//%20TODO(kartik` at this line where for some reason BankKeeper did not have a property `MintingRestrictionFn` which osmosis uses. This sets a minting restriction for the tokenfactory, but even though our cosmos-sdk version includes that custom restriction fn as part BankKeeper , I'm getting errors when including it. Here is the part of the cosmos sdk which defines it `https://sourcegraph.com/github.com/cosmos/cosmos-sdk@v0.45.6/-/blob/x/bank/keeper/keeper.go?L60:6#tab=references`. This will be added soon / in follow up PR's


wasmbinding/encoder.go
```diff
@@ -24,6 +29,14 @@ func CustomEncoder(sender sdk.AccAddress, msg json.RawMessage) ([]sdk.Msg, error
 		return dexwasm.EncodeDexPlaceOrders(parsedMessage.PlaceOrders)
 	case parsedMessage.CancelOrders != nil:
 		return dexwasm.EncodeDexCancelOrders(parsedMessage.CancelOrders)
+	case parsedMessage.CreateDenom != nil:
```
udpatil, line 32: LGTM

do we need to add in logic for tokenfactory queries as well?

udpatil (commented):
The wasmbinding stuff looks good to me, I'm not super familiar with where all modifications need to be made when adding a new module though, maybe @LCyson or @codchen can review that part of the code?


Just to clarify, is minting restricted to the admin of that token, whereas burning can be performed by anyone with a token balance? or does burning also need to be performed by that admin?

proto/tokenfactory/query.proto
```diff
@@ -0,0 +1,71 @@
...
+import "tokenfactory/params.proto"; 
+
+option go_package = "github.com/sei-protocol/sei-chain/x/tokenfactory/types";
+
+// Query defines the gRPC querier service.
+service Query {
+  // Params defines a gRPC query method that returns the tokenfactory module's
+  // parameters.
+  rpc Params(QueryParamsRequest) returns (QueryParamsResponse) {
+    option (google.api.http).get = "/osmosis/tokenfactory/v1beta1/params";
```
udpatil: btw these paths have `osmosis` in them which we should fix, probably best to just find all "osmosis" and correct them to sei or something

Kbhat1: Updated

wasmbinding/encoder.go
Kbhat1: Discussed this on discord, it seems like we may want to the query of all the denoms created by an admin? Can include that in a separate PR if not high priority

proto/tokenfactory/query.proto
```diff
@@ -14,23 +14,23 @@ service Query {
   // Params defines a gRPC query method that returns the tokenfactory module's
   // parameters.
   rpc Params(QueryParamsRequest) returns (QueryParamsResponse) {
-    option (google.api.http).get = "/osmosis/tokenfactory/v1beta1/params";
+    option (google.api.http).get = "/sei-protocol/seichain/tokenfactory/params";
```
udpatil, line 17: I noticed most modules have `seichain` in the path but oracle has `sei-chain`, we should probably fix that for oracle, right? (not actionable for you @Kbhat1, asking @LCyson)

wasmbinding/bindings/msg.go
```diff
@@ -0,0 +1,43 @@
+package bindings
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+type SeiMsg struct {
```
udpatil: where do we use this struct? I saw that we use the other structs for encoding from the wasm struct to the proto generated struct for the message, but not sure why we need this one? @Kbhat1

Kbhat1: Yea sorry this should be removed

x/tokenfactory/keeper/bankactions.go
```diff
@@ -0,0 +1,73 @@
...
+		sdk.NewCoins(amount))
+	if err != nil {
+		return err
+	}
+
+	return k.bankKeeper.BurnCoins(ctx, types.ModuleName, sdk.NewCoins(amount))
+}
+
+// nolint: unused
+func (k Keeper) forceTransfer(ctx sdk.Context, amount sdk.Coin, fromAddr string, toAddr string) error {
```
codchen: do we want to comment out/remove this?

codchen (approved):

x/tokenfactory/keeper/bankactions.go
Kbhat1: Will comment out for now

PR #183, opened 2022-08-02
Use generic and add common data structures

codchen:
This PR is a refactor without business logic changes. Specifically it:
- added common, **synchronous**, data structures like set and typed map, using generics to reduce code duplication
- refactored `dex/cache` using the generic data structures and split it into multiple files for better readability
- improved unit test coverage for `dex/cache`

The synchronous data structures added here are prerequisites for `dex` `EndBlock` parallel processing.


utils/datastructures/set.go
```diff
@@ -0,0 +1,54 @@
...
+		s.Remove(val)
+	}
+}
+
+func (s *SyncSet[T]) Contains(val T) bool {
+	_, ok := s.dict[val]
+	return ok
+}
+
+func (s *SyncSet[T]) ToSlice() []T {
```
philipsu522: Will this be nondeterministic? if so, maybe we should sort when appending to res, and call this "ToOrderedSlice" or something

utils/datastructures/set.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func (s *SyncSet[T]) Remove(val T) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	delete(s.dict, val)
+}
+
+func (s *SyncSet[T]) RemoveAll(vals []T) {
+	for _, val := range vals {
```
philipsu522, line 42: do we need to lock on this? otherwise there could be a race condition while iterating that a new object gets added in

utils/datastructures/typed_sync_map.go
```diff
@@ -0,0 +1,132 @@
...
+	if !ok {
+		return
+	}
+	nestedMap.Delete(key2)
+	if nestedMap.Len() == 0 {
+		m.TypedSyncMap.Delete(key1)
+	}
+}
+
+func (m *TypedNestedSyncMap[K1, K2, V]) DeepCopy(copier func(V) V) *TypedNestedSyncMap[K1, K2, V] {
```
philipsu522, line 127: do we not need to lock this func and the next func?

utils/datastructures/set.go
codchen: made it sorted. This was like this because the only place that calls it doesn't care about ordering, but agreed that we should sort it to prevent future use cases from breaking

utils/datastructures/set.go
codchen: ah good call

utils/datastructures/typed_sync_map.go
codchen: `sync.Map.Range` would ensure each layer of ranging in itself is synchronous. Without mutex it's possible for nested layer of upcoming iterations to change during the outer layer iteration, but non-determinism caused by such a scenario can't really be helped even with a mutex (e.g. on one node the entire copy logic happened before some nested elements are added, whereas on another node it happened after). That said, it's probably easier to reason about the logic with a mutex in place here. Added one

utils/datastructures/set.go
```diff
@@ -0,0 +1,67 @@
+package datastructures
+
+import (
+	"sort"
+	"sync"
+)
+
+type SyncSet[T comparable] struct {
```
LCyson, line 11: can we add comments including:
- basic introduction to the data struct
- functionalities
- simple example

utils/datastructures/set.go
```diff
@@ -0,0 +1,67 @@
...
+func (s *SyncSet[T]) RemoveAll(vals []T) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	for _, val := range vals {
+		delete(s.dict, val)
+	}
+}
+
+func (s *SyncSet[T]) Contains(val T) bool {
+	_, ok := s.dict[val]
```
LCyson, line 48: do we need to worry about the result of contains can be non-deterministic here? if another thread remove the val between 45-46

utils/datastructures/typed_sync_map.go
```diff
@@ -0,0 +1,136 @@
+package datastructures
+
+import (
+	"sync"
+)
+
+type TypedSyncMap[K any, V any] struct {
```
LCyson, line 10: ditto

utils/datastructures/typed_sync_map.go
```diff
@@ -0,0 +1,136 @@
...
+
+func (m *TypedSyncMap[K, V]) DeepApply(toApply func(V)) {
+	m.Range(func(_ K, val V) bool {
+		toApply(val)
+		return true
+	})
+}
+
+type TypedNestedSyncMap[K1 any, K2 any, V any] struct {
+	*TypedSyncMap[K1, *TypedSyncMap[K2, V]]
```
LCyson, line 79: nit: what's the reason we don't directly use sync.Map here

LCyson (approved):

utils/datastructures/set.go
codchen: chatted about this offline. The data structure here only guarantees data race free, not non-determinism-free

utils/datastructures/typed_sync_map.go
codchen: `sync.Map` is type-less, so the same map may have keys in different types if the code writer is not careful

utils/datastructures/set.go
codchen: added

PR #184, opened 2022-08-03
Update ContractInfo data structure

codchen:
According to parallelization design discussed in https://www.notion.so/3pgv/RFC-dex-EndBlock-Parallelization-96188f5904f94172a7fdd7d2bac196f3#5937b9bab86e46dc97b8cc7d8b516007 , we will now maintain the immediate sibling (both younger and elder) for each inter-contract dependency. This PR:
- replaces `ContractInfo`'s existing field of `DependentContractAddrs` with a new `Dependencies` field which is an array of struct `ContractDependencyInfo` that keeps track of siblings. The old `ContractInfo` is renamed to `LegavcyContractInfo` for migration purposes.
- adds a `numIncomingDependencies` field to record the number of contracts that depend on the current contract.
- adds logics in `msg_server_register_contract` to keep the new fields updated whenever a contract is added/updated
- adds `dex` migration handler (v5 -> v6) to transform legacy contract info into the new format

### DIFF
--- a/x/dex/keeper/msgserver/msg_server_register_contract.go
+++ b/x/dex/keeper/msgserver/msg_server_register_contract.go
@@ -2,8 +2,10 @@ package msgserver
 
 import (
 	"context"
+	"errors"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/sei-protocol/sei-chain/utils"
 	"github.com/sei-protocol/sei-chain/x/dex/contract"
 	"github.com/sei-protocol/sei-chain/x/dex/types"
 )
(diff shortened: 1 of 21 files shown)


x/dex/keeper/msgserver/msg_server_register_contract.go
```diff
@@ -2,8 +2,10 @@ package msgserver
 
 import (
 	"context"
+	"errors"
```
philipsu522, line 5: seems to be a lot of complex logic in here. Given that I don't expect this to run suuuuper frequently, I'd suggest adding a lot of logging so we can debug when there are issues.

philipsu522 (approved):

x/dex/keeper/msgserver/msg_server_register_contract.go
codchen: sounds good. Added a bunch of logging

PR #188, opened 2022-08-04
Loadtest client fixes

philipsu522:
This PR changes it so that the client doesn't limit the number of txns / block since that's ultimately what we're trying to benchmark. Instead, we now send as many txns as possible and specify `rounds` - which is the number of times the accounts should loop and send txns. Includes other optimizations (see inline comments).

### DIFF
--- a/app/params/config.go
+++ b/app/params/config.go
@@ -82,6 +82,7 @@ func SetTendermintConfigs(config *tmcfg.Config) {
 	config.P2P.SendRate = 20480000
 	config.P2P.RecvRate = 20480000
 	config.P2P.MaxPacketMsgPayloadSize = 10240
+	config.P2P.FlushThrottleTimeout = 10 * time.Millisecond
 	// Mempool configs
 	config.Mempool.Size = 5000
 	config.Mempool.MaxTxsBytes = 10737418240
--- a/loadtest/main.go
+++ b/loadtest/main.go
@@ -126,7 +126,7 @@ func run(config Config) {
 	}
 	wgs := []*sync.WaitGroup{}
 	sendersList := [][]func(){}
-	fmt.Printf("%s - Starting block prepare\n", time.Now().String())
+	fmt.Printf("%s - Starting block prepare\n", time.Now().Format("2006-01-02T15:04:05"))
 	for i := 0; i < int(config.Rounds); i++ {
 		fmt.Printf("Preparing %d-th block\n", i)
 		wg := &sync.WaitGroup{}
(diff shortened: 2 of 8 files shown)


loadtest/main.go
```diff
@@ -178,11 +179,7 @@ func run(config Config) {
 			txBuilder := TestConfig.TxConfig.NewTxBuilder()
 			_ = txBuilder.SetMsgs(msg)
 			seqDelta := uint64(i / 2)
-			SignTx(&txBuilder, key, seqDelta)
```
philipsu522, line 181: This is duplicated in `SendTx`. Can remove

loadtest/main.go
```diff
@@ -178,11 +179,7 @@ func run(config Config) {
...
 			mode := typestx.BroadcastMode_BROADCAST_MODE_SYNC
-			if j == len(activeAccounts)-1 {
```
philipsu522, line 183: We don't want to block on the last account, since that will limit throughput. Rather, we should send everything as SYNC (see differences here https://github.com/cosmos/cosmos-sdk/issues/4186#issuecomment-486664720). Otherwise, throughput is limited at the transaction with broadcast-mode=block

x/dex/ante.go
```diff
@@ -52,7 +52,6 @@ func (tsmd TickSizeMultipleDecorator) CheckTickSizeMultiple(ctx sdk.Context, msg
 						PriceDenom: order.PriceDenom,
 						AssetDenom: order.AssetDenom,
 					})
-				fmt.Println(contractAddr)
```
philipsu522, line 55: Remove unnecessary IO, this could contribute a bit to perf.

loadtest/main.go
```diff
@@ -195,42 +190,17 @@ func run(config Config) {
 		inactiveAccounts, activeAccounts = activeAccounts, inactiveAccounts
 	}
 
-	lastHeight := getLastHeight()
```
philipsu522, line 198: We don't restrict number of txns per block now so this can be removed. Additionally, this was a tcp call so it could've caused some perf issues.

app/params/config.go
```diff
@@ -82,6 +82,7 @@ func SetTendermintConfigs(config *tmcfg.Config) {
 	config.P2P.SendRate = 20480000
 	config.P2P.RecvRate = 20480000
 	config.P2P.MaxPacketMsgPayloadSize = 10240
+	config.P2P.FlushThrottleTimeout = 10 * time.Millisecond
```
philipsu522, line 85: further decreases blocktimes

codchen (approved):

PR #200, opened 2022-08-10
Update initialize local sei node script

LCyson:


scripts/initialize_local_test_node.sh
```diff
@@ -45,4 +43,33 @@ cat ~/.sei/config/genesis.json | jq '.app_state["staking"]["params"]["bond_denom
...
+if [ ! -z "$1" ]; then
+  CONFIG_PATH="$1"
+else
+  CONFIG_PATH="$HOME/.sei/config/config.toml"
+fi
+
+if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+  sed -i 's/timeout_prevote =.*/timeout_prevote = "2000ms"/g' $CONFIG_PATH
+  sed -i 's/timeout_precommit =.*/timeout_precommit = "2000ms"/g' $CONFIG_PATH
+  sed -i 's/timeout_commit =.*/timeout_commit = "2000ms"/g' $CONFIG_PATH
```
philipsu522, line 57: is this just going to be a single node?

philipsu522 (approved):

scripts/initialize_local_test_node.sh
LCyson: yes, this is for local testing only

PR #201, opened 2022-08-10
Fix spans and endblockpanic alert

philipsu522:
Fix spans so they use the same context (otherwise they don't show up under the same request).
Additionally, fix the endblockpanic alert since we cannot glob over titles (only labels).
Span: image
Page: https://seinetwork.pagerduty.com/incidents/Q3UV2GHWDHQWQR

### DIFF
--- a/x/dex/module.go
+++ b/x/dex/module.go
@@ -273,10 +273,11 @@ func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) (ret []abc
 		if err := recover(); err != nil {
 			ctx.Logger().Error(fmt.Sprintf("panic occurred in %s EndBlock: %s", types.ModuleName, err))
 			telemetry.IncrCounterWithLabels(
-				[]string{fmt.Sprintf("%s%s", types.ModuleName, "endblockpanic")},
+				[]string{fmt.Sprintf("endblockpanic")},
 				1,
 				[]metrics.Label{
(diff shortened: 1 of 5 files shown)


x/dex/module.go
```diff
@@ -264,15 +264,20 @@ func (am AppModule) beginBlockForContract(ctx sdk.Context, contract types.Contra
...
+	defer span.End()
 	// TODO (codchen): Revert https://github.com/sei-protocol/sei-chain/pull/176/files before mainnet so we don't silently fail on errors
 	defer func() {
+		_, span := (*am.tracingInfo.Tracer).Start(am.tracingInfo.TracerContext, "DexEndBlockRollback")
+		defer span.End()
 		if err := recover(); err != nil {
 			ctx.Logger().Error(fmt.Sprintf("panic occurred in %s EndBlock: %s", types.ModuleName, err))
 			telemetry.IncrCounterWithLabels(
-				[]string{fmt.Sprintf("%s%s", types.ModuleName, "endblockpanic")},
+				[]string{fmt.Sprintf("endblockpanic")},
```
philipsu522: fixes metric so that alert name doesn't need to be globbed. See alert: https://github.com/sei-protocol/sei-node-monitoring/pull/11/files#diff-3c740453dd758f4cf3d23745832b46ea89f80a5c033852b91c88746efbe29c6aR25

codchen (approved):

PR #203, opened 2022-08-10
Port Upgrade module from Cosmos SDK

BrandonWeng:
Checked out `v0.45.4` cosmos-sdk and copied it directly to our repo and did a full replace of all `github.com/cosmos/cosmos-sdk/x/upgrade` --> `github.com/sei-protocol/sei-chain/x/upgrade` 

This is the first pr for adding epoch to the upgrade module, I decided to split it up to make it easier to review


go.mod
```diff
@@ -135,6 +136,5 @@ require (
 
 replace (
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
-	github.com/keybase/go-keychain => github.com/99designs/go-keychain v0.0.0-20191008050251-8e49817e8af4
```
BrandonWeng, line 138: This was causing issues with building the upgrade module that was ported in. I don't think this is needed, it was added in the init commit: https://github.com/sei-protocol/sei-chain/blame/master/go.mod#L138 

Osmosis and Juno both don't have this 
https://github.com/osmosis-labs/osmosis/blob/main/go.mod#L286-L297 
https://github.com/CosmosContracts/juno/blob/main/go.mod#L127-L132

philipsu522 (dismissed):
have we tested this by running through an upgrade?

PR #205, opened 2022-08-12
Fix race conditions detected by --race

codchen:
Fixed all race conditions detected when running load test with `--race` flag on. Note that the race detector doesn't necessarily capture non-determinism, so testing on multiple nodes is still necessary even after this change.

### DIFF
--- a/loadtest/contracts/deploy_ten_contracts.sh
+++ b/loadtest/contracts/deploy_ten_contracts.sh
@@ -69,8 +69,8 @@ marsaddr4=$(python3 parser.py contract_address $marsinsres3)
 # register
 echo "Registering..."
 
-valaddr=$(printf "12345678\n" | $seidbin keys show $(printf "12345678\n" | $seidbin keys show node_admin --output json | jq -r .address) --bech=val --output json | jq -r '.address')
-printf "12345678\n" | $seidbin tx staking delegate $valaddr 1000000000usei --from=$keyname --chain-id=$chainid -b block -y
+# valaddr=$(printf "12345678\n" | $seidbin keys show $(printf "12345678\n" | $seidbin keys show node_admin --output json | jq -r .address) --bech=val --output json | jq -r '.address')
+# printf "12345678\n" | $seidbin tx staking delegate $valaddr 1000000000usei --from=$keyname --chain-id=$chainid -b block -y
 
 printf "12345678\n" | $seidbin tx dex register-contract $marsaddr $marsid false true  -y --from=$keyname --chain-id=$chainid --fees=10000000usei --gas=500000 --broadcast-mode=block
 printf "12345678\n" | $seidbin tx dex register-contract $saturnaddr $saturnid false true -y --from=$keyname --chain-id=$chainid --fees=10000000usei --gas=500000 --broadcast-mode=block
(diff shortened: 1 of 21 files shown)


loadtest/contracts/deploy_ten_contracts.sh
```diff
@@ -69,8 +69,8 @@ marsaddr4=$(python3 parser.py contract_address $marsinsres3)
 # register
 echo "Registering..."
 
-valaddr=$(printf "12345678\n" | $seidbin keys show $(printf "12345678\n" | $seidbin keys show node_admin --output json | jq -r .address) --bech=val --output json | jq -r '.address')
-printf "12345678\n" | $seidbin tx staking delegate $valaddr 1000000000usei --from=$keyname --chain-id=$chainid -b block -y
+# valaddr=$(printf "12345678\n" | $seidbin keys show $(printf "12345678\n" | $seidbin keys show node_admin --output json | jq -r .address) --bech=val --output json | jq -r '.address')
+# printf "12345678\n" | $seidbin tx staking delegate $valaddr 1000000000usei --from=$keyname --chain-id=$chainid -b block -y
```
philipsu522: shoudl we uncomment?

loadtest/contracts/deploy_ten_contracts.sh
```diff
@@ -143,7 +143,7 @@ marsproposalid4=$(python3 parser.py proposal_id $marspair4)
 printf "12345678\n" | $seidbin tx gov deposit $marsproposalid4 10000000usei -y --from=$keyname --chain-id=$chainid --fees=10000000usei --gas=500000 --broadcast-mode=block
 printf "12345678\n" | $seidbin tx gov vote $marsproposalid4 yes -y --from=$keyname --chain-id=$chainid --fees=10000000usei --gas=500000 --broadcast-mode=block
 
-printf "12345678\n" | $seidbin tx staking unbond $valaddr 1000000000usei --from=$keyname --chain-id=$chainid -b block -y
+# printf "12345678\n" | $seidbin tx staking unbond $valaddr 1000000000usei --from=$keyname --chain-id=$chainid -b block -y
```
philipsu522: can prob remove this and add sleep 90 above

x/dex/contract/abci.go
```diff
@@ -154,22 +194,36 @@ func orderMatchingRunnable(ctx context.Context, sdkContext sdk.Context, env *env
...
 				msg.AddContractResult(orderResults)
 				env.finalizeMsgMutex.Unlock()
 			}
 		}
 		env.settlementsByContract.Store(contractInfo.ContractAddr, settlements)
 	}
+
+	// ordering of events doesn't matter since events aren't part of consensus
+	env.eventManagerMutex.Lock()
+	parentSdkContext.EventManager().EmitEvents(sdkContext.EventManager().Events())
```
philipsu522, line 224: nit - is it possible for this to throw error? should we use defer for the mutex?

philipsu522 (approved):
fuego 🔥

loadtest/contracts/deploy_ten_contracts.sh
codchen: :+1:

loadtest/contracts/deploy_ten_contracts.sh
codchen: :+1:

x/dex/contract/abci.go
codchen: yeah makes sense

PR #207, opened 2022-08-12
Remove liquidation sudo logic

codchen:
Liquidation is a product-specific feature, and most contracts that will be deployed on `Sei` will not need liquidation capabilities. So contracts should instead implement liquidation as `execute` endpoints. This PR removes Sei-level logics that handle liquidation via `sudo`.


LCyson (approved):
note this will break the current integ test of vortex contracts, we can comment out the liquidation related scenarios

codchen:
@LCyson integration tests were already updated in https://github.com/sei-protocol/vortex-contract/pull/120

PR #208, opened 2022-08-13
Aggregate individual order/settlement state into a single entry

codchen:
Storing orders and settlements individually as KV entries in the AVL tree based store has proven to be too expensive. We are going to experiment with having all orders/settlements aggregated into a single KV store entry per block/contract and see how much performance improvements it brings. If this gives us satisfactory performance results, we will go ahead and build offchain indexer that indexes the single entries to be more query-able format.


x/dex/contract/execution.go
```diff
@@ -212,8 +201,9 @@ func ExecutePairsInParallel(ctx sdk.Context, contractAddr string, dexkeeper *kee
 		// need to re-throw panic to the top level goroutine
 		panic("panicked during pair execution")
 	}
+	dexkeeper.SetMatchResult(ctx, contractAddr, types.NewMatchResult(orderResults, cancelResults, settlements))
```
philipsu522, line 204: once the offchain solution is built, can this be removed?

philipsu522 (approved):

x/dex/contract/execution.go
codchen: @philipsu522 this will actually be what the offchain indexer reads from for now. We can change it to be event-based later but I want to first see what's the performance impact of emitting lots of events from the load test kartik's about to do

PR #211, opened 2022-08-16
Make settlement FIFO

codchen:
Use FIFO for settlement to avoid fragmented settlement records. MEV prevention would need to be done with other methods (e.g. threshold encryption)


x/dex/exchange/settlement.go
```diff
@@ -94,24 +111,40 @@ func SettleFromBook(
...
+			} else {
+				newLongToSettle = append(newLongToSettle, ToSettle{
+					amount:  allocation.Quantity,
+					orderID: allocation.OrderId,
+					account: allocation.Account,
+				})
+				longSettledQuantity = longSettledQuantity.Add(allocation.Quantity)
+			}
+		} else {
 			nonZeroNewLongAllocations = append(nonZeroNewLongAllocations, &types.Allocation{
```
LCyson, line 145: do we need to break here? it seems that we are still going over the entire long book allocations even if we have fulfilled the amount of `executedQuantity` here

LCyson: actually is it on purpose? why do we create a new long allocations instead of modifying the existing one here

codchen: it's intentional. I don't think a go slice can be modified and we would need to create a new slice if we want to delete items from a slice

LCyson (approved):

PR #212, opened 2022-08-16
Optimistic block processing RFC

codchen:
- adds template for RFC in Sei
- adds RFC for optimistic block processing


docs/rfc/README.md
```diff
@@ -0,0 +1,39 @@
+---
+order: 1
+parent:
+  order: false
+---
+
+# Requests for Comments
+
+A Request for Comments (RFC) is a record of discussion on an open-ended topic
+related to the design and implementation of Tendermint Core, for which no
```
philipsu522: nit - doesn't have to just be Tendermint

docs/rfc/rfc-000-optimistic-proposal-processing.md
```diff
@@ -0,0 +1,94 @@
...
+
+The OP goroutine would operate on top of a cache branch of the Cosmos store, and a branch
+equivalent for any state that is not managed by the Cosmos store.
+
+The OP goroutine would periodically (e.g. after every 10 txs) check if a termination signal is sent
+to it, and stops if so. If not, the OP goroutine would set the completion signal when it finishes
+processing.
+
+Upon receiving a `ProcessProposal` call, the application would adopt the following procedure:
+> if round == 0
```
philipsu522: might be good to call out why we only do OP for round 0 and not subsequent rounds

philipsu522 (approved):

docs/rfc/README.md
codchen: nice catch

docs/rfc/rfc-000-optimistic-proposal-processing.md
codchen: :+1:

PR #221, opened 2022-08-23
Add ethermint/evm module to app

philipsu522:


ratankaliani:
This is pretty cool! Does adding EVM & WASM support in the same app significantly increase strain on existing validator set-ups/add complexity at the app-layer by needing to have support for ERC-20s & cw20s?

philipsu522:
> This is pretty cool! Does adding EVM & WASM support in the same app significantly increase strain on existing validator set-ups/add complexity at the app-layer by needing to have support for ERC-20s & cw20s?

Hey! Good question. We're still looking into this, since we ultimately would like to support EVM < > Solidity compatibility. That will likely be addressed in a separate PR

ratankaliani:
> Hey! Good question. We're still looking into this, since we ultimately would like to support Solidity < > Cosmwasm compatibility. That will likely be addressed in a separate PR

Yeah makes sense, integrating cw & erc-20 in tokenfactory feels like a great place to start

When the PR comes out would be great if you could link bc would love to take a look

app/app.go
```diff
@@ -290,6 +293,10 @@ type App struct {
 	EpochKeeper epochmodulekeeper.Keeper
 
 	TokenFactoryKeeper tokenfactorykeeper.Keeper
+
+	// Ethermint keepers
+	EvmKeeper       *evmkeeper.Keeper
+	FeeMarketKeeper feemarketkeeper.Keeper
```
LCyson, line 300: how does the feemarket module interact with the original gas fee system here?

cmd/seid/cmd/root.go
```diff
@@ -342,15 +357,75 @@ func initAppConfig() (string, interface{}) {
...
 # This is the maximum sdk gas (wasm and storage) that we allow for any x/wasm "smart" queries
 query_gas_limit = 300000
 # This is the number of wasm vm instances we keep cached in memory for speed-up
 # Warning: this is currently unstable and may lead to crashes, best to keep for 0 unless testing locally
-lru_size = 0`
+lru_size = 0
+
+###############################################################################
+###                           JSON RPC Configuration                        ###
+###############################################################################
```
LCyson, line 398: why do we put this in the root.go instead of the params store (from genesis.json)

cmd/seid/cmd/root.go
```diff
@@ -126,20 +133,27 @@ func initRootCmd(
 	)
 
 	// add server commands
-	server.AddCommands(
```
LCyson, line 129: do we also need the `sdkserver.AddCommands` here to keep seid command working

app/ante/ante.go
```diff
@@ -0,0 +1,110 @@
...
+	if err := options.validate(); err != nil {
+		return nil, err
+	}
+
+	return func(
+		ctx sdk.Context, tx sdk.Tx, sim bool,
+	) (newCtx sdk.Context, err error) {
+		var anteHandler sdk.AnteHandler
+
+		defer Recover(ctx.Logger(), &err)
```
LCyson, line 45: do we want to remove the panic-recover here? we properly want to avoid recover panic as that may hide more severe bugs (and we may not be able to monitor this each time)

app/ante/ante.go
```diff
@@ -0,0 +1,110 @@
...
+// NewAnteHandler returns an ante handler responsible for attempting to route an
+// Ethereum or SDK transaction to an internal ante handler for performing
+// transaction-level processing (e.g. fee payment, signature verification) before
+// being passed onto it's respective handler.
+func NewAnteHandler(options HandlerOptions) (sdk.AnteHandler, error) {
+	if err := options.validate(); err != nil {
+		return nil, err
+	}
+
+	return func(
```
LCyson, line 40: is it possible to return built EthAnteHandler & CosmosAnteHandler map here? or register ethAnteHandler to the cosmos ante decorator? returning a func here seems to make each tx creates a new AnteHandler here

codchen (approved):

app/ante/ante.go
philipsu522: good point. Done!

app/ante/ante.go
philipsu522: this is a good question. I think it doesn't make sense for a single txn to bring down the whole chain though. Ill think it makes sense to add monitoring though

app/app.go
philipsu522: This is for allowing for variable fee for Cosmos txns so that it behaves similarly to EIP-1559: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md
Do you think it makes sense to remove it and keep the original gas fee mechanism? \cc @codchen

cmd/seid/cmd/root.go
philipsu522: nope, basically ethermintserver.AddCommands is a superset of server.AddCommands, but includes necessary ones for evm. See https://github.com/evmos/ethermint/blob/main/server/util.go#L24
Note that it has `sdkserver` which is what we have here as `server`

cmd/seid/cmd/root.go
philipsu522: I'm not sure what you mean? Do you mind clarifying?

app/ante/ante.go
LCyson: yea the thing i'm worried about is this may hide err and make the chain get into a corrupted state. I'm aligned with having this on testnet + monitor, and remove before mainnet

app/app.go
LCyson: we can discuss this in a separate ticket^

LCyson (approved):

BrandonWeng:
Leave open, we will revist soon

PR #222, opened 2022-08-24
Implement optimistic processing

codchen:
Implementation of [RFC-000](https://github.com/sei-protocol/sei-chain/blob/master/docs/rfc/rfc-000-optimistic-proposal-processing.md)


LCyson (approved):

app/app.go
```diff
@@ -809,21 +815,71 @@ func (app *App) PrepareProposalHandler(ctx sdk.Context, req *abci.RequestPrepare
...
+	} else if !bytes.Equal(app.optimisticProcessingInfo.Hash, req.Hash) {
+		app.optimisticProcessingInfo.Aborted = true
+	}
 	return &abci.ResponseProcessProposal{
 		Status: abci.ResponseProcessProposal_ACCEPT,
 	}, nil
 }
 
 func (app *App) FinalizeBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlock) (*abci.ResponseFinalizeBlock, error) {
+	startTime := time.Now()
```
philipsu522, line 842: i might need a bit more context, but when is this called? Is this called in a separate thread?

codchen: Both ProcessProposal and FinalizeBlock are called synchronously by tendermint. ProcessProposal spawns an async goroutine which FinalizeBlock may wait on

PR #224, opened 2022-08-27
Add Metrics to measure sudo calls 

BrandonWeng:
Adds custom metrics to measure sei-specific logic, similar to measuring request duration in Web2. In this case, we need to measure the E2E latency on the client side as we are unable to export time-based metrics from WASM 

This is a more repeatable way of measuring the latency without manually adding print statements and parsing the query in an excel sheet

From http://grafana.sei-loadtest-testnet.seinetwork.io/d/ellgRxZ4k/sei-custom-metrics?orgId=1&var-instance=18.219.201.246:26660: 

Spikes line up with when I ran the load test 
image
* One thing to call out here is that there are a lot more bulk_cancel order calls here (almost 1-1 with the bulk order placements), which I believe is why there's no settlements

### DIFF
--- a/app/export.go
+++ b/app/export.go
@@ -49,7 +49,8 @@ func (app *App) ExportAppStateAndValidators(
 
 // prepare for fresh start at zero height
 // NOTE zero height genesis is a temporary feature which will be deprecated
-//      in favour of export at a block height
+//
+//	in favour of export at a block height
 func (app *App) prepForZeroHeightGenesis(ctx sdk.Context, jailAllowedAddrs []string) {
 	applyAllowedAddrs := false
 
--- a/x/dex/keeper/utils/wasm.go
+++ b/x/dex/keeper/utils/wasm.go
@@ -4,8 +4,36 @@ import (
 	"encoding/json"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/sei-protocol/sei-chain/utils/metrics"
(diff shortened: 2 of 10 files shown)


app/export.go
```diff
@@ -49,7 +49,8 @@ func (app *App) ExportAppStateAndValidators(
 
 // prepare for fresh start at zero height
 // NOTE zero height genesis is a temporary feature which will be deprecated
-//      in favour of export at a block height
+//
+//	in favour of export at a block height
```
BrandonWeng, line 53: ran `golangci-lint run --config .golangci.yml --fix ` on repo

x/dex/keeper/utils/wasm.go
```diff
@@ -2,11 +2,44 @@ package utils
...
+	case dextypeswasm.SudoSettlementMsg:
+		return "settlement"
+	case dextypeswasm.SudoOrderPlacementMsg:
+		return "bulk_order_placements"
+	case dextypeswasm.SudoOrderCancellationMsg:
+		return "bulk_order_cancellations"
+	case dextypeswasm.SudoLiquidationMsg:
+		return "liquidation"
+	default:
+		return "unknown"
```
BrandonWeng, line 30: should cover all the cases, but if we add more msg types in the future we'd need to update this 

image

codchen (approved):

LCyson (approved):

philipsu522 (approved):

PR #226, opened 2022-08-31
Create denom fee whitelist + proposal

Kbhat1:
- adds whitelist capability to exclude certain addresses from paying the create denom fee in the tokenfactory
- adds proposal to add creator addresses to this whitelist


x/tokenfactory/keeper/createdenom.go
```diff
@@ -11,9 +11,12 @@ import (
 
 // ConvertToBaseToken converts a fee amount in a whitelisted fee token to the base fee token amount
 func (k Keeper) CreateDenom(ctx sdk.Context, creatorAddr string, subdenom string) (newTokenDenom string, err error) {
-	err = k.chargeForCreateDenom(ctx, creatorAddr)
-	if err != nil {
-		return "", err
+	// Don't charge denom creation fee for those in whitelist
+	if !(k.IsCreatorInDenomFeeWhitelist(ctx, creatorAddr)) {
+		err = k.chargeForCreateDenom(ctx, creatorAddr)
```
udpatil, line 16: we don't have to address this in this PR, but what do you think we should do about the charge for create funds if we're modifying the community fund?

x/tokenfactory/keeper/createdenomfeewhitelist.go
```diff
@@ -0,0 +1,38 @@
+package keeper
```
udpatil: nit: can we add underscores in the file name

x/tokenfactory/keeper/grpc_query.go
```diff
@@ -33,3 +33,17 @@ func (k Keeper) DenomsFromCreator(ctx context.Context, req *types.QueryDenomsFro
 	denoms := k.getDenomsFromCreator(sdkCtx, req.GetCreator())
 	return &types.QueryDenomsFromCreatorResponse{Denoms: denoms}, nil
 }
+
+// Checks whether creator address is in create denom fee whitelist
+func (k Keeper) CreatorInDenomFeeWhitelist(ctx context.Context, req *types.QueryCreatorInDenomFeeWhitelistRequest) (*types.QueryCreatorInDenomFeeWhitelistResponse, error) {
```
udpatil, line 38: should we expose this via sei-cosmwasm / sei query wasm handler as well? that way contract could also implement a hard check for inclusion in this whitelist before doing some action (otherwise they would get charged the fee, so they wouldnt be able to gate the ability to create denom by whitelist membership)

udpatil: doesn't have to be in this PR btw

x/tokenfactory/types/gov.go
```diff
@@ -0,0 +1,61 @@
...
+func (p AddCreatorsToDenomFeeWhitelistProposal) String() string {
+	creators := ""
+	for _, creator := range p.CreatorList {
+		creators += creator
+	}
+	var b strings.Builder
+	b.WriteString(fmt.Sprintf(`Add Creators to Denom Fee Whitelist Proposal:
+  Title:       %s
+  Description: %s
+  Records:     %s
```
udpatil: nit: shouldnt this say CreatorList instead of Records

x/tokenfactory/types/keys.go
```diff
@@ -47,3 +48,15 @@ func GetCreatorPrefix(creator string) []byte {
 func GetCreatorsPrefix() []byte {
 	return []byte(strings.Join([]string{CreatorPrefixKey, ""}, KeySeparator))
 }
+
+// GetDenomFeeWhitelistPrefix returns the store prefix of addresses which are
+// whitelisted to avoid a denom creation fee
+func GetCreateDenomFeeWhitelistPrefix() []byte {
```
udpatil, line 54: do we need a function for this? can't we just use `CreateDenomFeeWhitelistKey` since this function doesn't apend anything specific (I believe we can do prefix iteration even without joining the separator)

udpatil (commented):
Left a couple comments, mostly nits, some for discussion and some actionable

x/tokenfactory/keeper/createdenom.go
Kbhat1: Yea this is something I'll discuss with tony on. Not really sure if we eliminate the community fund

x/tokenfactory/keeper/grpc_query.go
Kbhat1: Yes will address this in follow-up PR to sei-cosmwasm

x/tokenfactory/types/keys.go
Kbhat1: I added this for consistency with the rest of the file, particularly `GetCreatorsPrefix` which is a similar scenario which was already in tokenfactory beforehand

udpatil (approved):

codchen (approved):

PR #245, opened 2022-09-12
Refactor wasmbinding to use message decorator

LCyson:
This PR removes the wasm encoder and uses the wasm message decorator instead. This change will make sei-chain able to use more state variables in wasm msg processing (i.e. sdk context and other modules' keeper in this case).

Tested tokenfactory create & mint from vortex contract e2e locally.


Kbhat1 (approved):
LGTM thanks

PR #248, opened 2022-09-13
feat: add docker support

achilleas-kal:
This PR adds docker support for both aarch64 & x86/64

Steps to test:
1. git checkout 1.1.4beta
2. docker build . -t seid:prod
3. docker run seid:prod seid version (this should output 1.1.4beta)


BrandonWeng:
we have docker support, `make build-docker-node`

PR #249, opened 2022-09-13
[oracle] Remove oracle prevote

udpatil:
Still testing with localsei but  here's the PR that removes the prevote functionality


x/oracle/types/genesis.go
```diff
@@ -10,28 +10,25 @@ import (
...
 }
 
 // DefaultGenesisState - default GenesisState used by columbus-2
 func DefaultGenesisState() *GenesisState {
 	return &GenesisState{
-		Params:                        DefaultParams(),
-		ExchangeRates:                 []ExchangeRateTuple{},
-		FeederDelegations:             []FeederDelegation{},
-		PenaltyCounters:               []PenaltyCounter{},
-		AggregateExchangeRatePrevotes: []AggregateExchangeRatePrevote{},
```
LCyson, line 33: do we need to remove this from the state in the upgrade handler?

LCyson (approved):

x/oracle/types/genesis.go
udpatil: yes good point, I'll make that change

x/oracle/keeper/migrations.go
```diff
@@ -56,3 +56,17 @@ func (m Migrator) Migrate3to4(ctx sdk.Context) error {
...
 	return nil
 }
+
+// Migrate3to4 migrates from version 4 to 5
+func (m Migrator) Migrate4to5(ctx sdk.Context) error {
+	// we remove the prevotes from store in this migration
+	store := ctx.KVStore(m.keeper.storeKey)
+
+	oldPrevoteKey := []byte{0x04}
+	iter := sdk.KVStorePrefixIterator(store, oldPrevoteKey)
```
LCyson, line 66: this removes every in the oracle store right

LCyson (approved):

x/oracle/keeper/migrations.go
udpatil: yeah it iterates over the store and deletes anything thats prefixed by the prevote key

PR #246, opened 2022-09-12
RFC: parallelize transaction processing

codchen:

PR #253, opened 2022-09-19
Add accesscontrol module

BrandonWeng:
Adds a new accesscontrol module as defined in this RFC:
https://github.com/sei-protocol/sei-chain/pull/246/files 

This implementation is just the basic skeleton and additional functionality will be added as needed. Right now it should just register the module and not impact any existing features. 

Tested by running `scripts/initialize_local_test_node.sh` to make sure that the chain can still startup fine


proto/accesscontrol/genesis.proto
```diff
@@ -0,0 +1,30 @@
...
+message MessageDependencyMapping {
+    MessageType message_type = 1;
+    repeated string resource_dependencies = 2;
+}
+
+message Params {
+    option (gogoproto.goproto_stringer) = false;
+    
+    // mapping between every message type and its predetermined resource read/write sequence
+    repeated MessageDependencyMapping message_dependency_mapping = 1 [
```
udpatil, line 23: instead of having this be part of params, I think we should have it be its own item managed by accesscontrolkeeper. This way, we can have better key prefixing for each message type so we can get it individually instead of searching for it in a list.

x/accesscontrol/types/access_operation.go
```diff
@@ -0,0 +1,16 @@
+package types
+
+import (
+	fmt "fmt"
+)
+
+type AcessOperation struct {
```
udpatil: typo

x/accesscontrol/types/keys.go
```diff
@@ -0,0 +1,10 @@
+package types
+
+const (
+	// ModuleName defines the module name
+	ModuleName = "accesscontrol"
+
+	QuerierRoute = ModuleName
+
+	StoreKey = ModuleName
+)
```
udpatil, line 13: this is where we would add in the message dependency mapping key builders so that we can store it independently in the keeper

x/epoch/types/query.pb.go
```diff
@@ -150,7 +150,7 @@ func (m *QueryEpochRequest) XXX_DiscardUnknown() {
 var xxx_messageInfo_QueryEpochRequest proto.InternalMessageInfo
 
 type QueryEpochResponse struct {
-	Epoch Epoch `protobuf:"bytes,11,opt,name=epoch,proto3" json:"epoch"`
+	Epoch Epoch `protobuf:"bytes,1,opt,name=epoch,proto3" json:"epoch"`
```
udpatil, line 153: why did this proto gen file have changes if there were no associated proto def changes for epoch?

app/app.go
```diff
@@ -290,6 +297,8 @@ type App struct {
 	ScopedTransferKeeper capabilitykeeper.ScopedKeeper
 	ScopedWasmKeeper     capabilitykeeper.ScopedKeeper
 
+	AccessControlKeeper aclmodulekeeper.Keeper
```
philipsu522, line 300: Since we're introducing a new keeper, we likely need to add a migration handler for the store too: https://github.com/sei-protocol/sei-chain/blob/master/app/app.go#L750

app/app.go
```diff
@@ -99,6 +100,10 @@ import (
 
 	"github.com/sei-protocol/sei-chain/utils/tracing"
 
+	aclmodule "github.com/sei-protocol/sei-chain/x/accesscontrol"
```
philipsu522, line 106: we should set the order of genesis / begin / end block too https://github.com/sei-protocol/sei-chain/blob/master/app/app.go#L571

x/epoch/types/query.pb.go
BrandonWeng: Good question, I'm I don't think we're currently checking if generated proto files are updated -- this was the output when I ran it on the whole repo

proto/accesscontrol/genesis.proto
BrandonWeng: Gotcha, thanks for bringing this up.

So I kept it in the params (as we get the governance proposal changes part for free) but then in the keeper added the ability for us to store it based on the key (ModuleName-MessageType), this way the the other modules can still just use the keeper to get the resource dependency individually 

We'll also need to update the gov proposal handler to update each index here

app/app.go
BrandonWeng: Yessir, should be set down here  https://github.com/sei-protocol/sei-chain/pull/253/files#diff-0f1d2976054440336a576d47a44a37b80cdf6701dd9113012bce0e3c425819b7R675

app/export.go
```diff
@@ -49,7 +49,8 @@ func (app *App) ExportAppStateAndValidators(
 
 // prepare for fresh start at zero height
 // NOTE zero height genesis is a temporary feature which will be deprecated
-//      in favour of export at a block height
+//
```
BrandonWeng, line 52: autoformat

proto/accesscontrol/genesis.proto
```diff
@@ -0,0 +1,28 @@
...
+
+message GenesisState {
+    Params params = 1 [(gogoproto.nullable) = false];
+}
+
+message MessageDependencyMapping {
+    MessageType message_type = 1;
+    Module module_name = 2;
+    repeated string resource_dependencies = 3;
+}
```
BrandonWeng, line 17: I was going to use a map of <string, list>  for this but it seems like it's not very well supported with protos

https://stackoverflow.com/questions/38886789/protobuf3-how-to-describe-map-of-repeated-string

wasmbinding/bindings/msg.go
```diff
@@ -5,18 +5,18 @@ import (
...
-/// factory/{creating contract address}/{Subdenom}
-/// Subdenom can be of length at most 44 characters, in [0-9a-zA-Z./]
-/// The (creating contract address, subdenom) pair must be unique.
-/// The created denom's admin is the creating contract address,
-/// but this admin can be changed using the ChangeAdmin binding.
+// / CreateDenom creates a new factory denom, of denomination:
+// / factory/{creating contract address}/{Subdenom}
+// / Subdenom can be of length at most 44 characters, in [0-9a-zA-Z./]
+// / The (creating contract address, subdenom) pair must be unique.
+// / The created denom's admin is the creating contract address,
```
BrandonWeng, line 12: autoformat

x/accesscontrol/keeper/genesis.go
```diff
@@ -0,0 +1,20 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/sei-protocol/sei-chain/x/accesscontrol/types"
+)
+
+func (k Keeper) InitGenesis(ctx sdk.Context, genState types.GenesisState) {
+	k.SetParams(ctx, genState.Params)
+	for _, resourceDepedencyMapping := range genState.Params.GetMessageDependencyMapping() {
+		k.SetResourceDepedencyMapping(ctx, resourceDepedencyMapping)
+	}
```
BrandonWeng, line 13: @udpatil setting the keystore for each resource dep here

x/accesscontrol/keeper/params.go
```diff
@@ -0,0 +1,21 @@
...
+	k.paramSpace.GetParamSet(ctx, &params)
+	return params
+}
+
+// SetParams sets the total set of params.
+func (k Keeper) SetParams(ctx sdk.Context, params types.Params) {
+	k.paramSpace.SetParamSet(ctx, &params)
+	for _, resourceDepedencyMapping := range params.GetMessageDependencyMapping() {
+		k.SetResourceDepedencyMapping(ctx, resourceDepedencyMapping)
+	}
```
BrandonWeng, line 20: ditto for updating the keys when updating Params

x/accesscontrol/module.go
```diff
@@ -0,0 +1,169 @@
+package accesscontrol
```
BrandonWeng, line 1: most of this file was just copy paste from other modules

PR #255, opened 2022-09-22
Add Nitro manager module

codchen:
Add a new custom module in Sei. Right now this module has only one transaction endpoint `RecordTransactionData` that's permissioned by a whitelisted account list in this module's params.

This transaction endpoint simply writes the data it receives (transaction data, state root, and sender) into its store, and only one call is allowed for each slot (an L2 concept). Transaction data and state root are all in hex-encoded form. For transaction data specifically, since we don't want to write one entry per transaction but still want to be able to read individual transaction data when we implement fraud proof, a custom encoding scheme is applied to flatten the transactions into a 1D byte array:
- First 8 bytes: count (C) of transactions
- Next 8 * C bytes: transaction sizes
- The rest: transaction bodies


proto/nitro/genesis.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+package seiprotocol.seichain.nitro;
+
+import "gogoproto/gogo.proto";
+import "nitro/params.proto";
+
+option go_package = "github.com/sei-protocol/sei-chain/x/nitro/types";
+
+message GenesisState {
+  Params                     params             = 1 [(gogoproto.nullable) = false];
```
philipsu522: nit - spacing

x/nitro/keeper/msg_server.go
```diff
@@ -0,0 +1,51 @@
...
+	Keeper
+}
+
+func NewMsgServerImpl(keeper Keeper) types.MsgServer {
+	return &msgServer{Keeper: keeper}
+}
+
+var _ types.MsgServer = msgServer{}
+
+func (server msgServer) RecordTransactionData(goCtx context.Context, msg *types.MsgRecordTransactionData) (*types.MsgRecordTransactionDataResponse, error) {
```
LCyson, line 25: nit: can we add comments to the function

philipsu522 (approved):

x/nitro/keeper/data.go
```diff
@@ -0,0 +1,80 @@
...
+	key := make([]byte, 8)
+	binary.BigEndian.PutUint64(key, slot)
+	encoded := store.Get(key)
+	if encoded == nil {
+		return nil, errors.New("not found")
+	}
+	return DecodeTransactionData(encoded)
+}
+
+func EncodeTransactionData(data [][]byte) []byte {
```
LCyson, line 41: nit: can we add comment about the encoding logic here

LCyson (approved):

PR #258, opened 2022-09-23
[accesscontrol] Add dependency dag

udpatil:


go.mod
```diff
@@ -129,7 +130,7 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.1.66
+	github.com/cosmos/cosmos-sdk => ../sei-cosmos // github.com/sei-protocol/sei-cosmos v0.1.20
```
BrandonWeng: Reminder to revert this before merging

app/app.go
```diff
@@ -375,7 +376,7 @@ func New(
 			Tracer:        &tr,
 			TracerContext: context.Background(),
 		},
-		// txDecoder: encodingConfig.TxConfig.TxDecoder(),
+		txDecoder: encodingConfig.TxConfig.TxDecoder(),
```
BrandonWeng, line 379: for my understanding, why was this commented out previously?

app/app.go
```diff
@@ -941,23 +970,34 @@ func (app *App) ProcessBlock(ctx sdk.Context, txs [][]byte, req BlockProcessRequ
...
-			Code:      deliverTxResp.Code,
-			Data:      deliverTxResp.Data,
-			Log:       deliverTxResp.Log,
-			Info:      deliverTxResp.Info,
-			GasWanted: deliverTxResp.GasWanted,
-			GasUsed:   deliverTxResp.GasUsed,
-			Events:    deliverTxResp.Events,
-			Codespace: deliverTxResp.Codespace,
-		})
+	if err != nil {
```
BrandonWeng, line 972: why not just check for `ErrCycleInDAG` or maybe we could have `if err != nil` or `if err != nil || err == ErrCycleInDAG`?  I think it'll be a bit easier to read

app/graph.go
```diff
@@ -0,0 +1,193 @@
...
+type DagNode struct {
+	NodeID          int
+	TxIndex         int
+	AccessOperation acltypes.AccessOperation
+}
+
+type DagEdge struct {
+	FromNodeID      int
+	ToNodeID        int
+	AccessOperation *acltypes.AccessOperation
```
BrandonWeng: Why do we need to store the access operation of the edge here? If we have A -> B, will the access operation in the edge refer to the operation in A or B?

app/graph.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	if _, ok := dag.NodeMap[toIndex]; !ok {
+		return nil
+	}
+	newEdge := DagEdge{fromIndex, toIndex, accessOp}
+	dag.EdgesMap[fromIndex] = append(dag.EdgesMap[fromIndex], newEdge)
+	return &newEdge
+}
+
+func (dag *Dag) AddNodeBuildDependency(ctx sdk.Context, txIndex int, accessOp acltypes.AccessOperation) {
```
BrandonWeng: nit - could we get some high level comment on how this works

app/graph.go
```diff
@@ -0,0 +1,193 @@
...
+		if writeNodeIDs, ok := dag.AccessOpsMap[writeAccessOp]; ok {
+			for _, wn := range writeNodeIDs {
+				writeNode := dag.NodeMap[wn]
+				// if accessOp exists already (and from a previous transaction), we need to define a dependency on the previous message (and make a edge between the two)
+				// if from a previous transaction, we need to create an edge
+				if writeNode.TxIndex < dagNode.TxIndex {
+					lastTxNode := dag.NodeMap[dag.TxIndexMap[writeNode.TxIndex]]
+					dag.AddEdge(lastTxNode.NodeID, dagNode.NodeID, &writeAccessOp)
+				}
+			}
```
BrandonWeng: thoughts on making this a helper? Seems like the logic is repeated a couple times

app/graph.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+	case acltypes.AccessType_WRITE:
```
BrandonWeng: I believe there's also `AccessType_ANY`, though I think that might be similar to Accesstype_WRITE since it should just block on previous reads and writes

app/graph.go
BrandonWeng: Since we're adding the edge to the previous accessOperation (either read or write), would this be an issue when that transaction is not the last operation in a given message 

(i.e don't we need to wait for write operations to commit in a message before reading it?)

BrandonWeng (commented):
thanks for the change, just have some questions

app/app.go
udpatil: I think this was introduced in a previous change Tony made and then commented out, I needed this field for the dependency DAG so we can decode TXs to be able to get their message information to get dependencies.

app/app.go
udpatil: if it is a different error, we might need to handle the tx processing differently (currently not present). I can leave a TODO so that if its a different error we can have special handling for that.

app/graph.go
udpatil: the reasoning is because of cases where we have an edge for a write operation that occurs earlier in a Message, we want it to actually be linked to the very last access operation instead so that the completion signal can be sent when the msg processing is over. heres an example:
In this case, we have message 1 writing to resource A, but the actual edge would be from the read of resource B since thats the last access op of the message, so once that is complete we can know that the message is committed. 
m1wA -> m1rB ----> m2rA

 maybe it might be better instead of implicitly linking from the last access op of a message, each message can have an accessOp specifying the end of the message (This becomes more relevant if the last access op (eg m1rB in this case) is prior to end of message and there can still be a panic in between that and the end of message handling. This way, if we have a write in the message, we send its completion signal from the end of message access operation instead of the last specified access operation. WDYT?

app/graph.go
udpatil: for this specifically, if we are dependent on a previous write, we actually are tracking the last operation in a message and linking edge from that instead of directly linked to the write access op. I'll go back and add some more comments to discuss in more detail though so its clearer.

app/graph.go
udpatil: yeah, I already have a TODO above to replace with a helper that also takes into account resource hierarchy so instead of only linking resources of the same type, it'll also link from resources of superseding types as well.

app/graph.go
udpatil: Good point, I'll modify it to handle this as well, I was initially assuming the `0` proto values were being treated as basically invalid but this is a more graceful way of handling it.

udpatil: wait I double checked and it has access type UNKNOWN. Do we want to handle it as if it's ANY (and/or refactor to ANY?)

BrandonWeng: Ah, I think they should be the same thing. If you feel that ANY is more clear then I can make a PR to change it

app/graph.go
BrandonWeng: Gotcha, thanks for clarifying!

app/app.go
```diff
@@ -871,6 +872,34 @@ func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcess
...
 }
 
+func (app *App) BuildDependencyDag(ctx sdk.Context, txs [][]byte) (*Dag, error) {
+	// contains the latest msg index for a specific Access Operation
+	dependencyDag := Dag{
+		NodeMap:  make(map[int]DagNode),
+		EdgesMap: make(map[int][]DagEdge),
+	}
+	for txIndex, txBytes := range txs {
+		tx, err := app.txDecoder(txBytes)
```
codchen: with this there will be two repetitive decoding for each tx (once here, once in `runMsg`). We don't need to optimize it in this PR but let's add a TODO here in case we forget

app/graph.go
```diff
@@ -0,0 +1,193 @@
...
+	default:
+		// unknown or something else, raise error
+		ctx.Logger().Error("Invalid AccessControlType Received")
+	}
+	// update access ops map with the latest node id using a specific access op
+	dag.AccessOpsMap[accessOp] = append(dag.AccessOpsMap[accessOp], dagNode.NodeID)
+}
+
+// returns completion signaling map and blocking signals map
+func (dag *Dag) BuildCompletionSignalMaps() (completionSignalingMap map[int]map[acltypes.AccessOperation][]CompletionSignal, blockingSignalsMap map[int]map[acltypes.AccessOperation][]CompletionSignal) {
```
codchen, line 189: nit: it's not very easy to tell what `int` represents. Maybe we can define an explicit type `type DagNodeID int` for readability

BrandonWeng (approved):
lgtm, thanks for the change

PR #259, opened 2022-09-25
Codify wasmbinding errors

LCyson:
This PR registers the wasmbinding errors into sdk errors. 

Right now the wasmd module redacts any error threw by `msg_dispatcher` and `query_plugins` ( https://github.com/CosmWasm/wasmd/blob/main/x/wasm/keeper/msg_dispatcher.go#L175), which causes all sei-cosmwasm errors unable to surface to the top. An example of get_order_by_id err is: ```Generic error: Querier contract error: codespace: undefined, code: 1: execute wasm contract failed```

As a workaround, this PR codifies the errors so that people can trace which type of errors they encounter by the codespace and code id.


x/oracle/keeper/querier.go
```diff
@@ -113,17 +115,26 @@ func (q querier) Twaps(c context.Context, req *types.QueryTwapsRequest) (*types.
 	}
 
 	ctx := sdk.UnwrapSDKContext(c)
+	fmt.Println("breakpoint 1")
```
udpatil: nit: cleanup print statements

udpatil (approved):
LGTM, lests remove debugging print statements first before merging

PR #261, opened 2022-09-26
Create Channels from DAG

BrandonWeng:
Creates channels based on the DAG 

Pointing the PR to https://github.com/sei-protocol/sei-chain/pull/258 

Tested by running: ./scripts/old_initialize_local.sh

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 
 	storetypes "github.com/cosmos/cosmos-sdk/store/types"
@@ -30,6 +31,7 @@ import (
 	servertypes "github.com/cosmos/cosmos-sdk/server/types"
 	"github.com/cosmos/cosmos-sdk/simapp"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkacltypes "github.com/cosmos/cosmos-sdk/types/accesscontrol"
 	"github.com/cosmos/cosmos-sdk/types/module"
 	"github.com/cosmos/cosmos-sdk/version"
 	"github.com/cosmos/cosmos-sdk/x/auth"
@@ -38,6 +40,7 @@ import (
 	aclmodule "github.com/cosmos/cosmos-sdk/x/accesscontrol"
 	aclkeeper "github.com/cosmos/cosmos-sdk/x/accesscontrol/keeper"
 	acltypes "github.com/cosmos/cosmos-sdk/x/accesscontrol/types"
+
 	authrest "github.com/cosmos/cosmos-sdk/x/auth/client/rest"
 	authkeeper "github.com/cosmos/cosmos-sdk/x/auth/keeper"
 	authsims "github.com/cosmos/cosmos-sdk/x/auth/simulation"
@@ -881,11 +884,11 @@ func (app *App) BuildDependencyDag(ctx sdk.Context, txs [][]byte) (*Dag, error) 
 			return nil, err
 		}
 		msgs := tx.GetMsgs()
-		for _, msg := range msgs {
+		for messageIndex, msg := range msgs {
 			msgDependencies := app.AccessControlKeeper.GetResourceDependencyMapping(ctx, acltypes.GenerateMessageKey(msg))
-			for _, accessOp := range msgDependencies.AccessOps {
+			for _, accessOp := range msgDependencies.GetAccessOps() {
 				// make a new node in the dependency dag
-				dependencyDag.AddNodeBuildDependency(txIndex, accessOp)
+				dependencyDag.AddNodeBuildDependency(messageIndex, txIndex, accessOp)
 			}
 		}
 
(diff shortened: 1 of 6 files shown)


app/app.go
```diff
@@ -929,6 +930,98 @@ func (app *App) FinalizeBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlock)
...
+func (app *App) ProcessBlockSynchronous(ctx sdk.Context, txs [][]byte) []*abci.ExecTxResult {
+	txResults := []*abci.ExecTxResult{}
+	for _, tx := range txs {
+		txResults = append(txResults, app.DeliverTxWithResult(ctx, tx))
+	}
+	return txResults
+}
+
+// Returns a mapping of the access index to the channels
+func getChannelsFromSignalMapping(signalMapping map[acltypes.AccessOperation][]CompletionSignal) (channelsMapping [][](chan interface{})){
```
BrandonWeng: PR to update Context object https://github.com/sei-protocol/sei-cosmos/pull/27

app/graph.go
```diff
@@ -34,6 +34,7 @@ func (edge *DagEdge) GetCompletionSignal() *CompletionSignal {
 			FromNodeID:      edge.FromNodeID,
 			ToNodeID:        edge.ToNodeID,
 			AccessOperation: *edge.AccessOperation,
+			Channel:		 make(chan interface{}),
```
BrandonWeng: Thought it would make sense to store the channel here as it's set for both to and from nodes

udpatil: makes sense to me 👍

app/app.go
```diff
@@ -929,6 +930,98 @@ func (app *App) FinalizeBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlock)
...
+
+	// Waits for all the transactions to complete
+	go func() {
+		waitGroup.Wait()
+		close(resultChan)
+	}()
+
+	// Gather Results
+	txResults := []*abci.ExecTxResult{}
+	for result := range resultChan {
```
udpatil, line 1029: this isnt necessarily going to be in the same order as txs that were provided, right? is that safe, or should we also include the tx index with the results so we can sort into the tx index order prior to returning Tx results?

app/app.go
```diff
@@ -929,6 +930,98 @@ func (app *App) FinalizeBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlock)
...
+	wg *sync.WaitGroup,
+	resultChan chan<- *abci.ExecTxResult,
+	txCompletionSignalingMap map[acltypes.AccessOperation][]CompletionSignal,
+	txBlockingSignalsMap map[acltypes.AccessOperation][]CompletionSignal,
+) {
+	defer wg.Done()
+	// Store the Channels in the Context Object for each transaction
+	ctx.WithTxBlockingChannels(getChannelsFromSignalMapping(txBlockingSignalsMap))
+	ctx.WithTxCompletionChannels(getChannelsFromSignalMapping(txCompletionSignalingMap))
+
```
udpatil, line 989: can we add TODOs to block on blocking channels and writing to completion channels to properly block and release resources

BrandonWeng: added here: https://github.com/sei-protocol/sei-chain/pull/261/files#diff-0f1d2976054440336a576d47a44a37b80cdf6701dd9113012bce0e3c425819b7R968

go.mod
```diff
@@ -130,7 +130,7 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.1.67
+	github.com/cosmos/cosmos-sdk => ../sei-cosmos // github.com/sei-protocol/sei-cosmos v0.1.20
```
BrandonWeng: Will revert this once the sei-cosmos PR merges and I can bump up the version

app/graph.go
```diff
@@ -208,11 +230,5 @@ func (dag *Dag) BuildCompletionSignalMaps() (completionSignalingMap map[int]map[
 	return
 }
 
-type CompletionSignal struct {
```
BrandonWeng, line 211: Moving this up with all the other types

app/app.go
```diff
@@ -967,30 +1064,15 @@ func (app *App) ProcessBlock(ctx sdk.Context, txs [][]byte, req BlockProcessRequ
...
-				Code:      deliverTxResp.Code,
-				Data:      deliverTxResp.Data,
-				Log:       deliverTxResp.Log,
-				Info:      deliverTxResp.Info,
-				GasWanted: deliverTxResp.GasWanted,
-				GasUsed:   deliverTxResp.GasUsed,
-				Events:    deliverTxResp.Events,
-				Codespace: deliverTxResp.Codespace,
-			})
+		if err == ErrCycleInDAG {
```
philipsu522, line 1088: would be good to add metrics over how many times we process synchronously vs concurrently

app/app.go
```diff
@@ -926,6 +927,102 @@ func (app *App) FinalizeBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlock)
...
+			&waitGroup,
+			resultChan,
+			completionSignalingMap[txIndex],
+			blockingSignalsMap[txIndex],
+		)
+	}
+
+	// Waits for all the transactions to complete
+	go func() {
+		waitGroup.Wait()
```
philipsu522: we're waiting for the WG in a separate goroutine? doesn't that mean it won't block?

app/app.go
```diff
@@ -926,6 +927,102 @@ func (app *App) FinalizeBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlock)
...
+	completionSignalingMap map[int]MessageCompletionSignalMapping,
+	blockingSignalsMap map[int]MessageCompletionSignalMapping,
+) []*abci.ExecTxResult {
+	var waitGroup sync.WaitGroup
+	resultChan := make(chan *abci.ExecTxResult)
+
+	// For each transaction, start goroutine and deliver TX
+	for txIndex, txBytes := range txs {
+		waitGroup.Add(1)
+		app.ProcessTxConcurrent(
```
philipsu522: dont we want this in a separate goroutine?

udpatil (approved):

app/app.go
BrandonWeng: I'll add a todo here, looks like we don't have the telemetry lib from master. It would be easier to do once we merge this into master

app/app.go
BrandonWeng: ah shoot, yeah good catch. We'll need another `waitGroup.Wait()` in the method too

app/app.go
BrandonWeng: good catch 😵‍💫

app/app.go
philipsu522: ah what i meant is do we need this Wait()?  From what I can tell, you essentially want to launch a bunch of goroutines in ln 1000 from this main thread, and block when it's finally all done. If that's the case, then the waitGroup.Wati() in ln 1018 is sufficient to achieve this

app/app.go
```diff
@@ -926,6 +927,105 @@ func (app *App) FinalizeBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlock)
...
+			resultChan,
+			completionSignalingMap[txIndex],
+			blockingSignalsMap[txIndex],
+		)
+	}
+
+	// Spin off another gorountine to close channel once it's done
+	go func() {
+		waitGroup.Wait()
+		close(resultChan)
```
philipsu522: why do we close here and not in the goroutine in `ProcessTxConcurrent` ?

app/app.go
BrandonWeng: Ahh yeah, it's technically not really needed but more so it's just closing the channel. I guess in our case since we don't really need need to communicate that the data is done then we could just remove this

app/app.go
BrandonWeng: Yes, this will be in the order that it's completed in - let me double check if the ordering matters

philipsu522 (approved):

PR #263, opened 2022-09-28
Dag optimization

udpatil:
In this, I made some fixes to the completion signal maps builder, and added more testing for it. Additionally, we eliminate edges between access operations for the same tx index, helping make the graph a bit sparser.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -130,7 +130,7 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.1.69
+	github.com/cosmos/cosmos-sdk => ../sei-cosmos //github.com/sei-protocol/sei-cosmos v0.1.68
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 	github.com/keybase/go-keychain => github.com/99designs/go-keychain v0.0.0-20191008050251-8e49817e8af4
 	github.com/tendermint/tendermint => github.com/sei-protocol/sei-tendermint v0.1.54
--- a/go.mod
+++ b/go.mod
@@ -130,7 +130,7 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => ../sei-cosmos //github.com/sei-protocol/sei-cosmos v0.1.68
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -130,7 +130,7 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.1.67
+	github.com/cosmos/cosmos-sdk => ../sei-cosmos //github.com/sei-protocol/sei-cosmos v0.1.68
```
udpatil: self reminder to revert this prior to merging

app/app.go
```diff
@@ -967,30 +1067,17 @@ func (app *App) ProcessBlock(ctx sdk.Context, txs [][]byte, req BlockProcessRequ
...
-				Code:      deliverTxResp.Code,
-				Data:      deliverTxResp.Data,
-				Log:       deliverTxResp.Log,
-				Info:      deliverTxResp.Info,
-				GasWanted: deliverTxResp.GasWanted,
-				GasUsed:   deliverTxResp.GasUsed,
-				Events:    deliverTxResp.Events,
-				Codespace: deliverTxResp.Codespace,
-			})
+		if err == ErrCycleInDAG {
```
philipsu522: we probably want to handle other errors too. i.e. 
```
tx, err := app.txDecoder(txBytes) // TODO: results in repetitive decoding for txs with runtx decode (potential optimization)
if err != nil {
	return nil, err
}
```
 in BuildDependencyDag

philipsu522 (approved):

codchen (approved):

app/app.go
udpatil: we actually have addressed this in a previous PR that is already merged into 2.0.0beta I believe. 
https://github.com/sei-protocol/sei-chain/commit/e573b4ab96da1c1cceb461134cf2b3e7bd92ca4c#diff-0f1d2976054440336a576d47a44a37b80cdf6701dd9113012bce0e3c425819b7R972
This is currently on top of brandon's changes, and those need to be rebased on top of the merged commit that fixed this error branching.

go.mod
BrandonWeng: fyi, i bumped it to `v0.1.69`, should be in 2.0.0

BrandonWeng (approved):

PR #268, opened 2022-09-29
[graph] Add resource hierarchy to dependency graph

udpatil:
This adds in the ability to create dependencies between resources of different types based on predefined resource hierarchies. Currently, for dependencies with the same resource, we are only performing exact matching on the identifier template, but we can modify that in the future to allow for partial matching and fuzzy matching for more control over dependency granularity.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/cosmos/cosmos-sdk v0.45.4
 	github.com/cosmos/go-bip39 v1.0.0
 	github.com/cosmos/ibc-go/v3 v3.0.0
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -130,7 +133,7 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.1.69
+	github.com/cosmos/cosmos-sdk => ../sei-cosmos //github.com/sei-protocol/sei-cosmos v0.1.69
```
udpatil: note to self, revert this.

BrandonWeng (approved):

PR #270, opened 2022-09-30
Parallel wasm

udpatil:


docs/rfc/rfc-002-wasm-contract-parallelization.md
```diff
@@ -0,0 +1,37 @@
...
+One approach we can take is a gradual parallelization of a wasm contract. When the wasm contract is first uplaoded, we can have it behave as if it uses all resources, essentially making its execution sequential. After the contract has been uploaded (or even as part of upload proposal), we can then allow for a dependency mapping for the contract parallelization which would map between the different execute and query messages for the contract to the resource dependencies that the contract would require during the execution of that execute / query. These dependency mappings can be associated with a contract code ID so that the parallelization can be consistently applied to multiple instances of that contract.
+
+> OPEN QUESTION: If we require additional granularity, should we allow mapping specific to the contract instance (this would be relevant for other contracts calling a specific contract (eg. code ID 1), if there are multiple instances of thcontracts with code ID 1, we could allow them to run in parallel since the other calling contract can specify exactly what resources would be affected that are specific to the contract instance that it would be calling).
```
codchen, line 20: I think we should and can with the template approach similar to obtaining an account for bank send during runtime

docs/rfc/rfc-002-wasm-contract-parallelization.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Discussion
+
+### Parallelization Approach
+One approach we can take is a gradual parallelization of a wasm contract. When the wasm contract is first uplaoded, we can have it behave as if it uses all resources, essentially making its execution sequential. After the contract has been uploaded (or even as part of upload proposal), we can then allow for a dependency mapping for the contract parallelization which would map between the different execute and query messages for the contract to the resource dependencies that the contract would require during the execution of that execute / query. These dependency mappings can be associated with a contract code ID so that the parallelization can be consistently applied to multiple instances of that contract.
```
codchen, line 18: It might be tricky for contract developers to get this mapping uploaded right. As an intermediate step, I wonder if it's possible to simplify this for them by just asking 1. resources the execute handler itself will query from, 2. resources the execute handler will send messages to update. Then we can construct for them a rough but possibly good enough dependency mapping:
1a. Read from ANY query resource
1b. Write to `wasm/{contract address}`
2. Write to ANY message resource
Note that this might get recursive since a contract can send message to another contract, which is okay as long as there is no cycle (to think about it I'm curious if wasm itself has any guard against it, since even without our parallelization effort, cycles in contract calling can still be a problem as in infinite loop)

docs/rfc/rfc-002-wasm-contract-parallelization.md
```diff
@@ -0,0 +1,37 @@
...
+
+We need to fail the current transaction so that it doesn't improperly modify any resources that haven't been properly defined in the contract mapping. Additionally, we would also need to fail any other transactions with that contract in the current block, since otherwise we would need to rebuild the dag with the sequential processing for the contract and restart the block's execution. Then, in the following blocks, those contract transactions would be able to process sequentially as expected.
+
+> OPEN QUESTION: What are the potential side effects of failing contract transactions this way?
+>
+> OPEN QUESTION: What about the specific transactions that failed? Is it ok to just leave them as failed and assume that the user can resend them later? Or do we need to some other handling to ensure that we process those TXs sequentially as well (This would negatively impact block time)?
```
codchen, line 34: i think we can get away with just leaving them as failed since technically it's on contract developers who specified the dependency wrong. How strong this argument is though depends on how easy we make dependency registration

docs/rfc/rfc-002-wasm-contract-parallelization.md
```diff
@@ -0,0 +1,37 @@
...
+
+> OPEN QUESTION: If we require additional granularity, should we allow mapping specific to the contract instance (this would be relevant for other contracts calling a specific contract (eg. code ID 1), if there are multiple instances of thcontracts with code ID 1, we could allow them to run in parallel since the other calling contract can specify exactly what resources would be affected that are specific to the contract instance that it would be calling).
+
+This would follow a similar pattern for the message dependency mapping for sdk messages, but instead would be blocked prior to contract execution and would release the resources at the end of the contract execution. The reason for this is because it would be much more effort to construct a system to granularly release resources within contract execution.
```
codchen, line 22: We should be able to release reads fairly easily since reads within a contract needs to go through the wasmbinding querier where we can manage resources.

docs/rfc/rfc-002-wasm-contract-parallelization.md
codchen: For contracts with `reply` though I can't think of a good way at the moment. I don't think too many contracts make use of `reply` anyway so maybe we can just make all execute with reply sequential

docs/rfc/rfc-002-wasm-contract-parallelization.md
udpatil: currently for that we're using identifier templates, which will be used to partition mutually exclusive resources that aren't static (eg. account addresses vs KV store prefixes)

docs/rfc/rfc-002-wasm-contract-parallelization.md
udpatil: The thing about the proposed approach above is that most contracts will interact with the chain modules outside of the contract itself, with things such as bank transfers, which means that in that case the contract would take control of READ ANY and WRITE ANY, so it would result in basically all contract becoming blocking / sequential inherently. Haven't thought too much about reply, but my thought was that the resource wouldn't be released until AFTER the TX is done executing, which should include the reply processing.

docs/rfc/rfc-002-wasm-contract-parallelization.md
udpatil: the issue is, lets say it has multiple bank reads, but we don't have a good way of knowing when the last read is, and we would need to know that to confidently release a read resource, right?

docs/rfc/rfc-002-wasm-contract-parallelization.md
udpatil: The behavior would only fail on that singular block, since the registration would then be disabled and future contract executions would be processed sequentially, so its not super consequential.

docs/rfc/rfc-002-wasm-contract-parallelization.md
codchen: It just needs to block on the resource list of the message type it could possibly send, which should be a smaller set than ANY i think

docs/rfc/rfc-002-wasm-contract-parallelization.md
codchen: ideally multiple accesses to the same resource within a single message would also be treated as separate nodes on the dependency graph, but i guess we wouldn't have that kind of granularity initially

udpatil: this would be easier to manage as part of normal message handlers, but its a little more difficult when controlling the wasm contract execution. That's why I think blocking on the contract execution as a whole is much simpler and can allow for different contracts to be processed concurrently (although not as optimally as being able to release resources during contract execution)

docs/rfc/rfc-002-wasm-contract-parallelization.md
udpatil: yeah thats true, and thats where the accuracy of the mapped dependencies is important, I'd imagine we could work with contract developers to identify the best parallelization dependencies for their contract if theyre having trouble identifying the best approach

docs/rfc/rfc-002-wasm-contract-parallelization.md
```diff
@@ -0,0 +1,37 @@
...
+
+One key difference would be that we would also store an enabled flag for the parallelization mapping, and if the wasm contract fails validation, we would disable the parallelization mappings and only process that contract sequentially until the parallellization mappings are updated.
```
BrandonWeng, line 24: Would this mean that for a given contract, if it fails validation it will keep building out the resource dep, process the contract concurrently, fail, and process sequentially every time it's executed?

If this flag is in a persistent store, would users also need to flip the flag to re-enable parallelization after the mappings are updated?

docs/rfc/rfc-002-wasm-contract-parallelization.md
BrandonWeng, line 22: Based on what we've learned from the previous parallelization task, is it possible that there are resources accessed in the wasm contract that might be accessed before or after the contract execution in the same block that we should also be aware of?

PR #271, opened 2022-09-30
surface query errors to wasmbinding

eric-zhu-uw:
There were some errors that were being obfuscated since we would be returning `gRPC.status` codes instead of `cosmos-sdk/types/errors`. As a result, the end-user wouldn't get a meaningful error message. Added some tests to verify that the errors are getting surfaced through wasmbinding. In addition, added a test for the `default` case where the `req=nil` since it was missing a test case.

Removed some un-necessary error checks [Example](https://github.com/sei-protocol/sei-chain/compare/ezhu/surface_wasmbinding_errors?expand=1#diff-e69effac3fe86bcc0faf7288cffdef80f2f0d062c01a2f384997c4a187982554L14-L17) since this is already being checked [here](https://github.com/sei-protocol/sei-chain/compare/ezhu/surface_wasmbinding_errors?expand=1#diff-84abdccc6bcefcdb026d549c9ccf24a612648edffaabefbe8df24c4bfbcf6eceR122).

Overall, I think some future improvements we can make are:
- better error-checking for inputs for some of these queries and having more descriptive errors
- better documentation for the user so they don't have to dig through the code
- better naming convention for errors (Eg. error_code=2 always means InvalidRequestArguments) across all modules to provide some consistency for the user.


LCyson (approved):

LCyson:
lgtm, can we add the error (or at least a github link) to the gitbook https://app.gitbook.com/o/YiBih4jOIh8lif9Z44jw/s/vVOoEaSQGRIbgTgSvoEo/introduction/overview

eric-zhu-uw:
Sounds good, I create a [gitbook PR](https://app.gitbook.com/o/YiBih4jOIh8lif9Z44jw/s/vVOoEaSQGRIbgTgSvoEo/~/diff/~/changes/jV2TSQvKO0KaV5a5Yjyy/smart-contracts-and-local-development/errors) for an errors page. I put it under smart contract & local development heading but I'm not sure if that's the best place... I don't know if there's anywhere better.

If it looks good, I can merge it.

x/dex/keeper/query/grpc_query_get_twaps.go
```diff
@@ -6,15 +6,9 @@ import (
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/sei-protocol/sei-chain/x/dex/types"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 )
 
 func (k KeeperWrapper) GetTwaps(goCtx context.Context, req *types.QueryGetTwapsRequest) (*types.QueryGetTwapsResponse, error) {
-	if req == nil {
```
udpatil, line 14: why are we removing the nil request checks? Is it guaranteed to be non-nil?

udpatil (approved):

x/dex/keeper/query/grpc_query_get_twaps.go
eric-zhu-uw: It's guaranteed to be non-nil because the wasmbinding router will check if it's `nil` here: https://github.com/sei-protocol/sei-chain/compare/ezhu/surface_wasmbinding_errors?expand=1#diff-84abdccc6bcefcdb026d549c9ccf24a612648edffaabefbe8df24c4bfbcf6eceR122 so it would be caught higher up before the `GetTwaps` function.

PR #272, opened 2022-10-02
Fixes for parallel TX and metrics

BrandonWeng:
Fix for waitGroup and adding metrics to record the number of concurrent TXs processed vs synchronous 

Ran load test script -- see that number of total txs reported == concurrent txs processed 
image


loadtest/contracts/deploy_ten_contracts.sh
```diff
@@ -15,17 +15,17 @@ echo
 cd $seihome/loadtest/contracts/mars && cargo build && docker run --rm -v "$(pwd)":/code \
   --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
   --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
-  cosmwasm/rust-optimizer:0.12.5
+  cosmwasm/rust-optimizer:0.12.6
```
BrandonWeng, line 18: votex README says to use 0.12.6

x/dex/keeper/epoch.go
```diff
@@ -14,13 +14,13 @@ func (k Keeper) SetEpoch(ctx sdk.Context, epoch uint64) {
...
 	store.Set([]byte(EpochKey), bz)
+	ctx.Logger().Info(fmt.Sprintf("Current epoch %d", epoch))
 }
 
 func (k Keeper) IsNewEpoch(ctx sdk.Context) (bool, uint64) {
 	store := ctx.KVStore(k.storeKey)
 	b := store.Get([]byte(EpochKey))
 	lastEpoch := binary.BigEndian.Uint64(b)
 	currentEpoch := k.EpochKeeper.GetEpoch(ctx).CurrentEpoch
-	ctx.Logger().Info(fmt.Sprintf("Current epoch %d", currentEpoch))
```
BrandonWeng, line 24: This is super noisy since it gets called once for each settlement check
image

udpatil (approved):

philipsu522 (commented):
can we also have a metric on how long it takes to build the DAG?

app/graph.go
```diff
@@ -133,6 +135,7 @@ func (dag *Dag) AddEdge(fromIndex DagNodeID, toIndex DagNodeID) *DagEdge {
 //
 // It will also register the new node with AccessOpsMap so that future nodes that amy be dependent on this one can properly identify the dependency.
 func (dag *Dag) AddNodeBuildDependency(messageIndex int, txIndex int, accessOp acltypes.AccessOperation) {
+	defer metrics.MeasureBuildDagDuration(time.Now(), "AddNodeBuildDependency")
```
BrandonWeng, line 138: adding it to the two high level calls that are not constant time cc @philipsu522

app/graph.go
```diff
@@ -283,6 +287,7 @@ func (dag *Dag) BuildCompletionSignalMaps() (
 	completionSignalingMap map[int]MessageCompletionSignalMapping,
 	blockingSignalsMap map[int]MessageCompletionSignalMapping,
 ) {
+	defer metrics.MeasureBuildDagDuration(time.Now(), "BuildCompletionSignalMaps")
```
BrandonWeng, line 289: This should be the slowest call called once per block

philipsu522 (approved):

PR #276, opened 2022-10-03
[app] Add behavior to process blocks with gov txs sync

udpatil:
Tested locally with sei chain and gov txs

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -876,6 +876,15 @@ func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcess
 	}, nil
 }
 
+func isGovMessage(msg sdk.Msg) bool {
+	switch msg.(type) {
+	case *govtypes.MsgVoteWeighted, *govtypes.MsgVote, *govtypes.MsgSubmitProposal, *govtypes.MsgDeposit:
+		return true
+	default:
+		return false
+	}
+}
+
 func (app *App) BuildDependencyDag(ctx sdk.Context, txs [][]byte) (*Dag, error) {
 	// contains the latest msg index for a specific Access Operation
 	dependencyDag := NewDag()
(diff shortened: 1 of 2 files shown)


app/app.go
```diff
@@ -886,6 +895,9 @@ func (app *App) BuildDependencyDag(ctx sdk.Context, txs [][]byte) (*Dag, error)
 		}
 		msgs := tx.GetMsgs()
 		for messageIndex, msg := range msgs {
+			if isGovMessage(msg) {
+				return nil, ErrGovMsgInBlock
+			}
```
BrandonWeng, line 900: Hmm, it's not completely an error since this is expected right?

udpatil: yeah true, but this seemed like the best way to interrupt the dependency graph creation and have the sequential behavior. any idea on a better pattern to use?

scripts/old_initialize_local.sh
```diff
@@ -1,43 +1,35 @@
...
-  -p 6831:6831/udp \
-  -p 6832:6832/udp \
-  -p 5778:5778 \
-  -p 16686:16686 \
-  -p 14250:14250 \
-  -p 14268:14268 \
-  -p 14269:14269 \
-  -p 9411:9411 \
-  jaegertracing/all-in-one:1.33
+#!/bin/bash
```
codchen: are the changes below intended as part of the commit?

codchen (approved):

BrandonWeng (approved):
lgtm, reminder to undo changes for old init script

app/app.go
BrandonWeng: Yeah, that's fair -- perhaps we could just leave a comment here noting why gov messages are handled this way

scripts/old_initialize_local.sh
udpatil: oops no

PR #278, opened 2022-10-04
Remove Community Tax

Kbhat1:


app/upgrades.go
```diff
@@ -57,6 +59,20 @@ func (app App) RegisterUpgradeHandlers() {
 	}
 	for _, upgradeName := range upgradesList {
 		app.UpgradeKeeper.SetUpgradeHandler(upgradeName, func(ctx sdk.Context, plan upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
+			// Set params to Distribution here when migrating
+			if upgradeName == "1.2.3beta" {
+				newVM, err := app.mm.RunMigrations(ctx, app.configurator, fromVM)
+				if err != nil {
+					return newVM, err
```
philipsu522, line 66: should we return nil, err here?

Kbhat1: Yea I wasn't sure here. I defaulted to what osmosis did which was still returning the `newVM` with the err: https://github.com/osmosis-labs/osmosis/blob/c960e4a6d443d8d0a434179312e86c99600faae9/app/upgrades/v7/upgrades.go#L32 . But if this is better practice can change to `nil`

philipsu522: ah shoudl be fine

philipsu522 (approved):

PR #283, opened 2022-10-04
Update market order status at the end of block execution

LCyson:
This PR updates market order status at the end of block execution.

Verified in local E2E test.


Kbhat1 (approved):
LGTM

PR #287, opened 2022-10-06
Add aclmapping options and mapping folder

udpatil:
Add in the dynamic dependency generators and implement the wasm dependency generator to handle generating dependencies for wasm contracts.
TODO: still need to write tests for the mappings.

### DIFF
--- a/aclmapping/dex/mappings.go
+++ b/aclmapping/dex/mappings.go
@@ -1,0 +1,36 @@
+package acldexmapping
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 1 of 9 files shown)


aclmapping/dex/mappings.go
```diff
@@ -0,0 +1,36 @@
+package acldexmapping
```
BrandonWeng, line 1: just curious, whats the reasoning for adding it here instead of where the msg types are stored?

udpatil: we could do that too, but for modules such as wasm we'd have to fork the modules. This way it's kept separate from the module codebase for easier mantainence

philipsu522 (approved):

BrandonWeng (approved):

PR #294, opened 2022-10-07
[ante] Add ante dep generator default behavior

udpatil:
This uses the ante dep generator introduced in sei-cosmos and defines a wrapper for compatibility with existing antedecorators to have conservative dependency generation.
Needs testing

### DIFF
--- a/app/ante.go
+++ b/app/ante.go
@@ -67,27 +67,27 @@ func NewAnteHandlerAndDepGenerator(options HandlerOptions) (sdk.AnteHandler, sdk
 	// }
 
 	anteDecorators := []sdk.AnteFullDecorator{
-		DefaultWrappedAnteDecorator(ante.NewSetUpContextDecorator()), // outermost AnteDecorator. SetUpContext must be called first
+		sdk.DefaultWrappedAnteDecorator(ante.NewSetUpContextDecorator()), // outermost AnteDecorator. SetUpContext must be called first
 		// TODO: have dex antehandler separate, and then call the individual antehandlers FROM the gasless antehandler decorator wrapper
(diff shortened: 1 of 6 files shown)


philipsu522 (approved):

BrandonWeng (approved):

app/ante.go
```diff
@@ -66,29 +66,31 @@ func NewAnteHandler(options HandlerOptions) (sdk.AnteHandler, error) {
 	// 	sigVerifyDecorator = ante.NewBatchSigVerificationDecorator(options.BatchVerifier, sequentialVerifyDecorator)
 	// }
 
-	anteDecorators := []sdk.AnteDecorator{
-		ante.NewSetUpContextDecorator(), // outermost AnteDecorator. SetUpContext must be called first
+	anteDecorators := []sdk.AnteFullDecorator{
+		sdk.DefaultWrappedAnteDecorator(ante.NewSetUpContextDecorator()), // outermost AnteDecorator. SetUpContext must be called first
```
philipsu522, line 70: nit - could probably create a function for this i.e. `wrapDecorator(decs []Decorator) ([]WrappedDecorator, err)`

go.mod
```diff
@@ -131,7 +131,7 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.1.105
+	github.com/cosmos/cosmos-sdk => ../sei-cosmos //github.com/sei-protocol/sei-cosmos v0.1.105
```
philipsu522: let's revert this

philipsu522 (approved):

app/ante.go
udpatil: I think its better to explicitly wrap each ante handler, because otherwise future changes may change order of the antehandlers without properly changing the order of dependency generators, and that could potentially result in unintended behavior

PR #296, opened 2022-10-10
Add more parallel TX metrics

BrandonWeng:
Adding more metrics to keep monitor parallel TX, as it's difficult to debug, it would be useful to have more metrics for observability. 

image

### DIFF
--- a/utils/metrics/metrics_util.go
+++ b/utils/metrics/metrics_util.go
@@ -19,6 +19,32 @@ func IncrTxProcessTypeCounter(processType string) {
 	)
 }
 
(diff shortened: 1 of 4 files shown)


utils/metrics/metrics_util.go
```diff
@@ -32,16 +58,3 @@ func MeasureDeliverTxDuration(start time.Time) {
 	)
 }
 
-// Measures the time taken to execute a sudo msg
```
BrandonWeng, line 35: This was moved to sei-cosmos

udpatil (approved):

PR #298, opened 2022-10-10
Fix invalid contract sig fault

LCyson:
When a contract is removed from the `validContractsInfo` its dependent contracts cannot find it in the runner, causing a nil pointer exception. This PR fixes this issue by ignoring any contract that is removed from the contract address info map.

### DIFF
--- a/x/dex/contract/runner.go
+++ b/x/dex/contract/runner.go
@@ -120,7 +120,11 @@ func (r *ParallelRunner) wrapRunnable(contractAddr utils.ContractAddress) {
 		for _, dependency := range contractInfo.Dependencies {
 			dependentContract := dependency.Dependency
 			typedDependentContract := utils.ContractAddress(dependentContract)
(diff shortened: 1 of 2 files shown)


x/dex/contract/runner.go
```diff
@@ -120,7 +120,11 @@ func (r *ParallelRunner) wrapRunnable(contractAddr utils.ContractAddress) {
 		for _, dependency := range contractInfo.Dependencies {
 			dependentContract := dependency.Dependency
 			typedDependentContract := utils.ContractAddress(dependentContract)
-			dependentInfo, _ := r.contractAddrToInfo.Load(typedDependentContract)
+			dependentInfo, ok := r.contractAddrToInfo.Load(typedDependentContract)
+			if !ok {
```
philipsu522, line 128: let's add logging + metrics

philipsu522 (approved):

PR #301, opened 2022-10-11
add stop loss and stop limit orders

eric-zhu-uw:
- Added a new TriggerBook that stores triggered orders (stop-loss and stop-limit)
- Before matching starts, move any triggered orders from previous blocks from TriggerBook into the normal orderbook for matching
- _matching occurs like normal_
- At the end of matching, based on the min/max execution price, you would check the new triggered orders and old triggered orders to see if any of their conditions were triggered. If so, mark them so in the next block, they will be put into the order book.

- unit testing


x/dex/exchange/limit_order.go
```diff
@@ -48,6 +53,8 @@ func MatchLimitOrders(
 		TotalNotional: totalPrice,
 		TotalQuantity: totalExecuted,
 		Settlements:   settlements,
+		MinPrice:      minPrice,
+		MaxPrice:      maxPrice,
```
LCyson, line 57: can we add tests for the minPrice and maxPrice change

x/dex/exchange/market_order.go
```diff
@@ -33,12 +36,15 @@ func MatchMarketOrders(
...
 		}
+		minPrice, maxPrice = clearingPrice, clearingPrice
 		settlements = append(settlements, allTakerSettlements...)
 	}
 	return ExecutionOutcome{
 		TotalNotional: totalPrice,
 		TotalQuantity: totalExecuted,
 		Settlements:   settlements,
+		MinPrice:      minPrice,
+		MaxPrice:      maxPrice,
```
LCyson, line 47: ditto

LCyson (approved):
lgtm

PR #302, opened 2022-10-11
Change Order Processing

Kbhat1:
Update the order of the dex module processing to add limit orders to order book before matching. Order is now:
```
1) cancel trades

2) add all limit orders to the orderbook

3) fill market orders with the more liquid orderbook

4) fill matching buys and sell limit orders
```

- Updated `limit_order_test` and `exchange_test`

### DIFF
--- a/x/dex/contract/execution.go
+++ b/x/dex/contract/execution.go
@@ -59,7 +59,10 @@ func ExecutePair(
 	// First cancel orders
 	cancelForPair(ctx, typedContractAddr, typedPairStr, dexkeeper, orderbook)
 	// Add all limit orders to the orderbook
-	exchange.AddOutstandingLimitOrdersToOrderbook(ctx, typedContractAddr, typedPairStr, dexkeeper, orderbook)
+	orders := dexkeeper.MemState.GetBlockOrders(ctx, typedContractAddr, typedPairStr)
+	limitBuys := orders.GetLimitOrders(types.PositionDirection_LONG)
+	limitSells := orders.GetLimitOrders(types.PositionDirection_SHORT)
+	exchange.AddOutstandingLimitOrdersToOrderbook(orderbook, limitBuys, limitSells)
 	// Fill market orders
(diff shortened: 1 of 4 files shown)


x/dex/contract/execution.go
```diff
@@ -56,9 +56,17 @@ func ExecutePair(
 	typedContractAddr := dextypesutils.ContractAddress(contractAddr)
 	typedPairStr := dextypesutils.GetPairString(&pair)
 
+	// First cancel orders
 	cancelForPair(ctx, typedContractAddr, typedPairStr, dexkeeper, orderbook)
+	// Add all limit orders to the orderbook
+	orders := dexkeeper.MemState.GetBlockOrders(ctx, typedContractAddr, typedPairStr)
+	limitBuys := orders.GetLimitOrders(types.PositionDirection_LONG)
+	limitSells := orders.GetLimitOrders(types.PositionDirection_SHORT)
+	exchange.AddOutstandingLimitOrdersToOrderbook(orderbook, limitBuys, limitSells)
```
LCyson, line 65: nit: can you move line 62-64 to the `AddOutstandingLimitOrdersToOrderbook` function

LCyson (approved):
lgtm, we need to add a limit order test in the dex module test after this

x/dex/contract/execution.go
Kbhat1: @LCyson this was done on purpose. It was more modular to pass in `limitBuys` and `limitSells` into `AddOutstandingLimitOrdersToOrderbook` rather than having `AddOutstandingLimitOrdersToOrderbook` take in the dex keeper and call that itself. More importantly, it makes it significantly easier to test `AddOutstandingLimitOrdersToOrderbook` and the limit order matching functionality.

codchen (approved):

x/dex/contract/execution.go
Kbhat1: Resolved offline on discord

PR #303, opened 2022-10-12
Add msg send dynamic access ops

BrandonWeng:
Adding the dynamic access op mapping for testing purposes - this is most likely the end state, and adding this won't affect the actual processing logic but it does make changes easier as I don't need to branch off my personal branches 

make install works fine, and dynamic access ops are disabled by default

### DIFF
--- a/aclmapping/bank/mappings.go
+++ b/aclmapping/bank/mappings.go
@@ -23,6 +23,7 @@ func GetBankDepedencyGenerator() aclkeeper.DependencyGeneratorMap {
 	return dependencyGeneratorMap
 }
 
+// TODO:: we can make resource types more granular  (e.g KV_PARAM or KV_BANK_BALANCE)
 func MsgSendDependencyGenerator(keeper aclkeeper.Keeper, ctx sdk.Context, msg sdk.Msg) ([]sdkacltypes.AccessOperation, error) {
 	msgSend, ok := msg.(*banktypes.MsgSend)
 	if !ok {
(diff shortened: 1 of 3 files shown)


aclmapping/bank/mappings.go
```diff
@@ -0,0 +1,81 @@
...
+
+	accessOperations := []sdkacltypes.AccessOperation{
+		// MsgSend also checks if the coin denom is enabled, but the information is from the params.
+		// Changing the param would require a gov proposal, which is synchrounos by default
+
+		// Checks balance of sender
+		{
+			AccessType:         sdkacltypes.AccessType_READ,
+			ResourceType:       sdkacltypes.ResourceType_KV,
+			IdentifierTemplate: utils.GetIdentifierTemplatePerModule(utils.BANK, msgSend.FromAddress),
```
udpatil, line 41: instead of having the identifier template be per module, we should create more resource types to further segment the population. For example, under KV, we can have resource type `KV_BANK`, and then `KV_BANK_BALANCE` and `KV_BANK_PARAMS`, and then for the specific account we could have another resource type called `KV_BANK_BALANCE_ACCOUNT`. This way its much easier to have static dependencies that are granular enough to still allow for parallelization, and then the static read for this would be resource `KV_BANK_BALANCE` whereas the dynamic resource would be `KV_BANK_BALANCE_ACCOUNT` with the account identifier. We can represent this hierarchy in the resource hierarchy tree in resource.go in `sei-cosmos`.

udpatil: @BrandonWeng lets leave a TODO at the top of the generator function to make the resources more granular down the road

udpatil (approved):

aclmapping/bank/mappings.go
BrandonWeng: Yessir, thanks! As discussed offline - we want to unblock the work first and this can be something we address later once we get the concurrency working

PR #305, opened 2022-10-13
Wasm signal

udpatil:
This introduces the validation layer for verifying dependencies are declared properly for wasm dependencies during wasm contract execution. More specifically, this is used whenever messages are originating from the wasm contract and outgoing to other modules, we'll first validate that the wasm dependency properly declared dependencies that cover the dependencies used by the message, and only then allow it. If there are issues with the declared dependency, we error on the message, and then disable the wasm dependency mapping (basically meaning that the wasm contract will be a bottleneck by utilizing ALL resources during execution)
Added unit test for decorator behavior and the different cases in which we may want to disable the wasm dependency mapping,

TODO: Will put out an additional PR that further validates wasm queries as well.


go.mod
```diff
@@ -131,7 +131,7 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.1.114
+	github.com/cosmos/cosmos-sdk => ../sei-cosmos //github.com/sei-protocol/sei-cosmos v0.1.114
```
philipsu522: need to revert this prior to merging

app/app.go
```diff
@@ -499,10 +499,33 @@ func New(
...
+		appCodec,
+		app.keys[acltypes.StoreKey],
+		app.GetSubspace(acltypes.ModuleName),
+		aclOpts...,
+	)
 	// The last arguments can contain custom message handlers, and custom query handlers,
 	// if we want to allow any custom callbacks
 	supportedFeatures := "iterator,staking,stargate,sei"
-	wasmOpts = append(wasmbinding.RegisterCustomPlugins(&app.OracleKeeper, &app.DexKeeper, &app.EpochKeeper), wasmOpts...)
+	wasmOpts = append(
```
philipsu522, line 513: how come we need to register all plugins now?

philipsu522: is it b/c we any tx's may be dependent on any of the other modules?

udpatil: Because now we create the message handler itself, and we need the inputs to create the message handler as well

philipsu522 (approved):

PR #307, opened 2022-10-13
Get Latest Price Endpoint

Kbhat1:
- Fetches the latest price based on current context Block Time for a market

- Added in get latest price testing + get price tests


Kbhat1:
NOTE: One concern is whether there isn't a price written for a market based on the timestamp in `ctx.BlockTime().Unix()`. In the current code we just return `found` as `false`.

x/dex/keeper/query/grpc_query_get_latest_price.go
```diff
@@ -0,0 +1,26 @@
...
+	"google.golang.org/grpc/status"
+)
+
+func (k KeeperWrapper) GetLatestPrice(goCtx context.Context, req *types.QueryGetLatestPriceRequest) (*types.QueryGetLatestPriceResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	blockTimeStamp := uint64(ctx.BlockTime().Unix())
```
Kbhat1: @codchen @LCyson Right here is where we get the current block timestamp and use that for the current price. One issue is that we may not have a price written for this, but may exist for the previous block.

codchen: yeah we should use a iterator to find the closest price, otherwise this would return not found most of the time

Kbhat1: @codchen check latest commit, it loops through all prices and finds the latest given price

LCyson (approved):

PR #308, opened 2022-10-14
Fix market order by nominal type order status

LCyson:
This PR fixes the market order by nominal type order status. This order type's executed quantity follows the real time settlement price and nominal requested, which may not always be the same as the order.quantity. So in the cancellation check, we only need to cancel it when zero quantity is executed.

TODO:
- add unit test for the FOKMARKETBYVALUE type after the dex/module_test.go is updated

### DIFF
--- a/x/dex/contract/market_order.go
+++ b/x/dex/contract/market_order.go
@@ -41,8 +41,13 @@ func getUnfulfilledPlacedMarketOrderIds(
 			continue
 		}
 		if order.OrderType == types.OrderType_MARKET || order.OrderType == types.OrderType_LIQUIDATION ||
-			order.OrderType == types.OrderType_FOKMARKET || order.OrderType == types.OrderType_FOKMARKETBYVALUE {
+			order.OrderType == types.OrderType_FOKMARKET {
 			if settledQuantity, ok := orderIDToSettledQuantities[order.Id]; !ok || settledQuantity.LT(order.Quantity) {
+				res = append(res, order.Id)
+			}
+		} else if order.OrderType == types.OrderType_FOKMARKETBYVALUE {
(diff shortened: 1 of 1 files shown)


x/dex/contract/market_order.go
```diff
@@ -41,10 +41,15 @@ func getUnfulfilledPlacedMarketOrderIds(
 			continue
 		}
 		if order.OrderType == types.OrderType_MARKET || order.OrderType == types.OrderType_LIQUIDATION ||
-			order.OrderType == types.OrderType_FOKMARKET || order.OrderType == types.OrderType_FOKMARKETBYVALUE {
+			order.OrderType == types.OrderType_FOKMARKET {
 			if settledQuantity, ok := orderIDToSettledQuantities[order.Id]; !ok || settledQuantity.LT(order.Quantity) {
 				res = append(res, order.Id)
 			}
+		} else if order.OrderType == types.OrderType_FOKMARKETBYVALUE {
```
eric-zhu-uw, line 48: nit: maybe worth to add a test case where the `FOKMARKETBYVALUE` order has a non-zero settlement quantity but the settled quantity is < order quantity

eric-zhu-uw:
approved but I don't have perms to approve since I don't have write access.
image

x/dex/contract/market_order.go
LCyson: sg, will create a separate ticket to add the test for FOKMARKETBYVALUE and FOKMARKET after the dex module_test.go is updated

LCyson:
> lgtm, but I don't have perms to approve since I don't have write access. image

hm weird, updated your permission, you should be able to approve now

PR #313, opened 2022-10-17
Lazy Deposit All Module Accounts During EndBlock

BrandonWeng:
Main changes in this PR: 
https://github.com/sei-protocol/sei-cosmos/pull/58 

We are moving the module deposits to the end of the block and transferring them all at once. This is to prevent the gas fee deposit from being a single source of a bottleneck when it comes to concurrency. 

Ran 3 iterations of load tests with 5 rounds, and 400 orders per block 
image

image

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1114,7 +1114,7 @@ func (app *App) ProcessBlock(ctx sdk.Context, txs [][]byte, req BlockProcessRequ
 	}
 
 	// Finalize all Bank Module Transfers here so that events are included
-	lazyWriteEvents := app.BankKeeper.WriteLazyDepositsToModuleAccounts(ctx)
+	lazyWriteEvents := app.BankKeeper.WriteDeferredDepositsToModuleAccounts(ctx)
 	events = append(events, lazyWriteEvents...)
 
 	endBlockResp := app.EndBlock(ctx, abci.RequestEndBlock{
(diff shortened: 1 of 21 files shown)


codchen (approved):

app/app.go
```diff
@@ -1007,6 +1008,7 @@ func (app *App) ProcessBlockConcurrent(
 		return txResults
 	}
 
+	ctx.Logger().Info(fmt.Sprintf("Processing Tx Count=%d", len(txs)))
```
philipsu522: nit - might be good to make this debug & a metric instead

app/app.go
```diff
@@ -1032,6 +1034,7 @@ func (app *App) ProcessBlockConcurrent(
 	// Gather Results and store it based on txIndex and read results from channel
 	// Concurrent results may be in different order than the original txIndex
 	txResultsMap := map[int]*abci.ExecTxResult{}
+	ctx.Logger().Info("Waiting for TXs to return")
```
philipsu522: maybe set this to debug?

x/oracle/ante.go
```diff
@@ -47,6 +49,31 @@ func (spd SpammingPreventionDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, si
 	return next(ctx, tx, simulate)
 }
 
+func (spd SpammingPreventionDecorator) AnteDeps(txDeps []sdkacltypes.AccessOperation, tx sdk.Tx, next sdk.AnteDepGenerator) (newTxDeps []sdkacltypes.AccessOperation, err error) {
```
philipsu522, line 52: are these AnteDeps functions going to look similar across all antes (i.e. oracle, gasless)? If so, maybe it'll make sense to have this somewhere centralized. If not, ffti

philipsu522 (approved):

x/oracle/ante.go
BrandonWeng: Tony did that in his PR - I'm going to remove my changes and rebase on his

PR #314, opened 2022-10-18
Add dependencies for ante handlers that read/write accounts

codchen:
The following ante handlers access individual account KV states:
- `ConsumeGasForTxSizeDecorator` -> read only
- `SetPubKeyDecorator` -> read/write
- `SigGasConsumeDecorator` -> read only
- `SigVerificationDecorator` -> read only
- `IncrementSequenceDecorator` -> read/write
Since handlers above are applicable to all message types, it could potentially cause race condition for even bank send type (whose loadtest was only run with one msg per account per block). This PR adds dependencies on accounts for these handlers.

ran sei locally

### DIFF
--- a/aclmapping/utils/identifier_templates.go
+++ b/aclmapping/utils/identifier_templates.go
@@ -3,6 +3,7 @@ package util
 import "fmt"
 
 const (
+	ACCOUNT           = "acc"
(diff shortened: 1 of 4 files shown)


BrandonWeng (approved):
lgtm

PR #318, opened 2022-10-19
Oracle script fix

udpatil:
This updates the oracle script to follow the new pattern to provide prices (no prevote, just vote). 
Currently running on atlantic-1

### DIFF
--- a/scripts/oracle/price_feeder.py
+++ b/scripts/oracle/price_feeder.py
@@ -96,7 +96,7 @@ def main():
     parser.add_argument('--api-key', help='API Key for price fetcher', type=str)
     args=parser.parse_args()
 
-    pf = PriceFeeder(args.key, args.password, args.binary, args.chain_id, args.node, args.valoper, args.api_key)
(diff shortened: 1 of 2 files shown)


scripts/oracle/price_feeder.py
```diff
@@ -46,51 +39,17 @@ def init_val_addr(self):
     def get_current_vote_period(self):
         res = requests.get("{node}/blockchain".format(node=self.node))
         body = res.json()
-        return int(body["result"]["last_height"]) // 10
+        return int(body["result"]["last_height"])
```
philipsu522: we're also changing voting to happen 10x more frequently - is that intended?

philipsu522 (approved):

scripts/oracle/price_feeder.py
udpatil: yeah it would allow us to vote more frequently, although come to think of it, the atlantic-1 voting period is still 10 so this doesn;t really accomplish anything, I'll change this back

LCyson (approved):

PR #319, opened 2022-10-19
Add Nitro rust bindings

codchen:
Provide a `Replay` function that takes in a list of Nitro accounts, a list of Nitro program accounts, and a list of Nitro transactions, and invoke a rust library that replay the transactions and return the updated account data.

Because of the size of the input and output, inputs and outputs are communicated through files in `/tmp` folder that's cleaned up at the end of `Replay`.

Currently only Mac rust lib is included.  Will build lib on other platforms and include them in subsequent steps.

Also added proto definition for Nitro accounts and transactions.

unit tests that invoke the rust library


x/nitro/replay/rustbinding.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"fmt"
+
+	"github.com/sei-protocol/sei-chain/utils"
+)
+
+func callReplayer(accountFiles []string, programFiles []string, txFiles []string, outputDir string) (err error) {
+	defer func() {
+		if err := recover(); err != nil {
+			err = fmt.Errorf("received error when calling replayer: %s", err)
```
LCyson, line 18: can we avoid adding panic recover if most of the only predictable panic will be thrown?

if we have to add one, can we add metric (can reuse the `endblockpanic`) to this so that we won't be unaware of any panic happening during testnet period

x/nitro/replay/replay.go
```diff
@@ -0,0 +1,187 @@
...
+	"os"
+	"strconv"
+	"strings"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/sei-protocol/sei-chain/utils"
+	"github.com/sei-protocol/sei-chain/x/nitro/types"
+)
+
+func Replay(ctx sdk.Context, txs [][]byte, accounts []types.Account, programs []types.Account) ([]types.Account, error) {
```
LCyson, line 18: nit: can we add description to this fn

x/nitro/replay/replay.go
```diff
@@ -0,0 +1,187 @@
...
+	serialized += "|"
+	serialized += strings.Join(transactionData.Signatures, ",")
+	filepath := fmt.Sprintf("%s%s", directory, transactionData.Signature)
+	return filepath, os.WriteFile(filepath, []byte(serialized), 0644)
+}
+
+func writeAccountToFile(directory string, account types.Account) (string, error) {
+	serialized := ""
+	serialized += account.Pubkey
+	serialized += "|"
```
LCyson, line 169: nit: can we create constants for delimiter

x/nitro/replay/replay.go
```diff
@@ -0,0 +1,187 @@
...
+				"%d_%s_%s",
+				i.ProgramIdIndex,
+				strings.Join(utils.Map(i.Accounts, func(a uint32) string { return fmt.Sprintf("%d", a) }), ":"),
+				i.Data,
+			)
+		})
+		serialized += strings.Join(instructions, "-")
+		serialized += "|"
+	} else {
+		panic("unknown message type")
```
LCyson, line 151: may be better to return err instead of panic here?

x/nitro/replay/replay.go
LCyson: what's the `programs` here

x/nitro/replay/replay.go
```diff
@@ -0,0 +1,187 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/sei-protocol/sei-chain/utils"
+	"github.com/sei-protocol/sei-chain/x/nitro/types"
+)
+
+func Replay(ctx sdk.Context, txs [][]byte, accounts []types.Account, programs []types.Account) ([]types.Account, error) {
+	// there can be at most one replay per Sei block
+	inputDirectory := fmt.Sprintf("/tmp/replay_input_%d/", ctx.BlockHeight())
+	if err := os.Mkdir(inputDirectory, os.ModePerm); err != nil {
```
LCyson, line 21: will the replay cause the validator node to fall behind as it requires many os operations

LCyson (approved):

x/nitro/replay/rustbinding.go
codchen: will follow up in another PR. Merging now to unblock you

PR #320, opened 2022-10-19
Optimize genesis account creation script

BrandonWeng:
Taking inspiration from work around Parallel Tx,

The create account and adding balance commands don't actually write to any of the data store (AVL trees) (the data folder is empty after genesis accounts are created), the data is persisted in the genesis file.
* https://github.com/sei-protocol/sei-cosmos/blob/63ac1b222e281be5ff25623ade4b93bd61517c1f/simapp/simd/cmd/genaccounts.go#L115-L169

To deal with the race condition for the genesis file, we just store it in memory and flush it to the file at the end when all accounts are complete. 

For 20 nodes, 5000 accounts went from 30mins+ to < ~30 seconds

Created load test cluster and ran 5 rounds of 2k txs per block 

image

image

For 20 nodes, 10k accounts went from 2 hours + to 76 seconds
image

image

### DIFF
--- a/loadtest/scripts/populate_genesis_accounts.py
+++ b/loadtest/scripts/populate_genesis_accounts.py
@@ -7,7 +7,11 @@ import time
 
 PARALLEISM=32
 
-def add_genesis_account(account_name, lock, local=False):
+# Global Variable used for accounts
+# Does not need to be thread safe, each thread should only be writing to its own index
+global_accounts_mapping = {}
+
+def add_key(account_name, local=False):
     if local:
         add_key_cmd = f"yes | ~/go/bin/seid keys add {account_name} --keyring-backend test"
     else:
@@ -17,9 +21,14 @@ def add_genesis_account(account_name, lock, local=False):
         stderr=subprocess.STDOUT,
         shell=True,
     ).decode()
+
     splitted_outputs = add_key_output.split('\n')
     address = splitted_outputs[3].split(': ')[1]
     mnemonic = splitted_outputs[11]
+    return address, mnemonic
+
+
+def add_account(account_name, address, mnemonic, local=False):
     if local:
         add_account_cmd = f"~/go/bin/seid add-genesis-account {address} 1000000000usei --keyring-backend test"
     else:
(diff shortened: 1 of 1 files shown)


loadtest/scripts/populate_genesis_accounts.py
```diff
@@ -34,45 +43,106 @@ def add_genesis_account(account_name, lock, local=False):
...
         try:
-            with lock:
-                subprocess.check_call(
-                    [add_account_cmd],
-                    shell=True,
-                    timeout=20,
-                )
-                success = True
+            print(f'Running: ${add_account_cmd}')
+            subprocess.call(
```
BrandonWeng, line 60: We're essentially leaving the mutex to Seid command, it already does a bunch of checks against the genesis state when writing to it. 

This script will spam it until it passes, and we only care about the address that's returned

loadtest/scripts/populate_genesis_accounts.py
```diff
@@ -34,45 +43,106 @@ def add_genesis_account(account_name, lock, local=False):
...
         except subprocess.CalledProcessError as e:
-            print(f"Encountered error {e}, retrying {retry_counter - 1} times")
-            retry_counter -= 1
+            print(f"Encountered error {e}, retried {retry_counter} times")
+            retry_counter += 1
             sleep_time += 0.5
             time.sleep(sleep_time)
 
+    global_accounts_mapping[account_index] = {
+        "balance": {
```
BrandonWeng, line 76: took these two values directly from the output of a genesis.json file from one of the load test nodes, other than the address everything was the same

loadtest/scripts/populate_genesis_accounts.py
philipsu522: for safety, should we set a max of like 1k and if it still fails then error out?

loadtest/scripts/populate_genesis_accounts.py
```diff
@@ -34,45 +43,106 @@ def add_genesis_account(account_name, lock, local=False):
...
+        balances[key] = global_accounts_mapping[key]["balance"]
+        account_info[key] = global_accounts_mapping[key]["account"]
+
+    genesis_file["app_state"]["bank"]["balances"] = genesis_file["app_state"]["bank"]["balances"] + balances
+    genesis_file["app_state"]["auth"]["accounts"] = genesis_file["app_state"]["auth"]["accounts"] + account_info
+
+    num_accounts_created = len([account for account in account_info if account != 0])
+    print(f'Created {num_accounts_created} accounts')
+
+    assert num_accounts_created > number_of_accounts
```
philipsu522: should this be >=?

philipsu522 (approved):
this is fuego 🔥

loadtest/scripts/populate_genesis_accounts.py
```diff
@@ -34,45 +43,109 @@ def add_genesis_account(account_name, lock, local=False):
...
+    if retry_counter >= 1000:
+        exit(-1)
+    
+    global_accounts_mapping[account_index] = {
+        "balance": {
+            "address": address,
+            "coins": [
+                {
+                    "denom": "usei",
+                    "amount": "1000000000"
```
BrandonWeng, line 81: fine to hard code for now but should update when this runs
https://github.com/sei-protocol/sei-chain/pull/320/files#diff-e1c1ef35431f7e66f33f0a100a17660ccb93ed9a36e479b660fd1931faf905adR33

PR #323, opened 2022-10-20
Add failure load client

kwhuo68:
Adding in some msg types for failure cases (invalid/malformed) with corresponding config files

Running loadtest on instance and verifying script fails with failure flags


loadtest/failure_basic_invalid.json
```diff
@@ -0,0 +1,39 @@
+{
```
philipsu522: can we juts use the same config json for different tests? the only diff seems to be `message_type`

philipsu522 (approved):

PR #326, opened 2022-10-20
Constant load test

ziyincody:
Added a constant load test flag `-constant` so that it will keep on sending txs to the chain. For each tx, it randomly selects based on the available msg types.

After each txs are executed, it sleeps for 3s then verifies all txs are committed. 

After each test is done, it sleeps according to the param `const_load_interval` to determine how often it sends the txs.
The number of txs can be configured as the same as the existing loadtest set up.

Ran the constant load test and it's able to perform the above


loadtest/main.go
```diff
@@ -205,17 +209,43 @@ func run(config Config) {
...
+	if *constant {
+		// sleep 3 seconds wait for transaction to finish
+		time.Sleep(3 * time.Second)
+		for i := 0; i < len(txs); i++ {
+			txResponse := GetTxResponse(txs[i])
+			if txResponse.Tx == nil {
+				// Replace panic with metrics instead
+				panic("transaction not committed")
+			}
+			telemetry.IncrCounter(1, "load_test_committed_tx_counts")
```
ziyincody: telemetry doesn't really work since it's not inside the chain, any other recommended way to add the metrics?

loadtest/constant_load_config.json
```diff
@@ -0,0 +1,40 @@
+{
```
philipsu522: do we need a separate config file? can we just reuse config.json?

loadtest/main.go
```diff
@@ -127,7 +130,7 @@ func init() {
 	app.ModuleBasics.RegisterInterfaces(TestConfig.InterfaceRegistry)
 }
 
-func run(config Config) {
+func run(config Config, constant *bool) {
```
philipsu522: could this be grabbed from config.Constant?

loadtest/main.go
philipsu522: let's just remove telemetry fro now

loadtest/main.go
```diff
@@ -205,17 +209,43 @@ func run(config Config) {
...
+				panic("transaction not committed")
+			}
+			telemetry.IncrCounter(1, "load_test_committed_tx_counts")
+		}
+	}
 	fmt.Printf("%s - Finished\n", time.Now().Format("2006-01-02T15:04:05"))
 }
 
-func generateMessage(config Config, key cryptotypes.PrivKey, batchSize uint64) sdk.Msg {
+func generateMessage(config Config, key cryptotypes.PrivKey, batchSize uint64, constant *bool) sdk.Msg {
```
philipsu522: sort of strange to be passing a pointer to bool. Can we just pass bool?

loadtest/main.go
```diff
@@ -289,11 +319,27 @@ func getLastHeight() int {
 }
 
 func main() {
+	constant := flag.Bool("constant", false, "using constant client")
```
philipsu522: let's put this inside config

philipsu522 (approved):

PR #328, opened 2022-10-20
add staking dep gen mappings

eric-zhu-uw:
Added staking dep gen mappings.

Ran loadtesting on a distribution of `Delegate, Undelegate, BeginRedelegate` messages
image

Verified the following details:
1. Checked that the height continues to grow `seid q block | jq ".block.last_commit.height"` which implies a consensus is reached and the validators agree on an apphash
2. Verified that the `Delegation` message correctly submits the delegations for validator, account pairing `seid query staking delegations-to <validator_addr>`
3. Verified that `Undelegate` and `BeginRedelegate` messages are correctly submitted. This is a little more tricky to test since the delegation will only stay in the `Unbonding,Redelegate` state for a short period of time. However, you can verify this immediately after you run the loadtest script with `seid query staking redelegations-from <validator_addr>` and `seid query staking unbonding-delegations-from <validator_addr>`. 

If you want to try and test the changes locally, utilize my `sei-infra` [branch](https://github.com/sei-protocol/sei-infra/tree/ezhu-5-nodes). 

## To Do
- rebase off `2.0.0beta`
- Evaluate the performance impact of the parallelization's - will need to sync with @BrandonWeng on how to do this.
- Move the resource mappings to a more granularized/standardized configurations once @udpatil  pushes his change since my `sei-cosmos` [branch](https://github.com/sei-protocol/sei-cosmos/pull/69/files ) needs to merge into his [branch](https://github.com/sei-protocol/sei-cosmos/tree/oracle-resources).


BrandonWeng:
* Does delegation have any interactions with module accounts? We need to be careful with module accounts
* The hooks should be fine if we're not using any right now

eric-zhu-uw:
> * Does delegation have any interactions with module accounts? We need to be careful with module accounts

Yes, there is some interaction with the module account, For example, for `MsgDelegate`, it will call `DelegateCoinsFromAccountToModule`. However, all module accounts accessed are owned by the `validator` which should be covered by the this [ACL Operation](https://github.com/sei-protocol/sei-chain/pull/328/files#diff-672698a5be534fd675c70e0de09c6ee340e507485d681fd71cfe70c50b1b88b8R72-R83). I treat `staking/validator_addr` to represent validators and their corresponding module accounts.

Not certain if this is the best way, might be better to be more granular if there is ever a time where we might access the validator and not their corresponding module accounts or vice versa.

aclmapping/staking/mappings.go
```diff
@@ -0,0 +1,260 @@
...
+	if !ok {
+		return []sdkacltypes.AccessOperation{}, ErrorInvalidMsgType
+	}
+
+	accessOperations := []sdkacltypes.AccessOperation{
+		// Checks if the delegator exists
+		// Checks if there is a delegation object that already exists for (delegatorAddr, validatorAddr)
+		{
+			AccessType:         sdkacltypes.AccessType_READ,
+			ResourceType:       sdkacltypes.ResourceType_KV,
```
eric-zhu-uw, line 40: This and subsequent mappings will need to be updated to `sdkacltypes.ResourceType_KV_STAKING_*` once uday's PR gets merged.

loadtest/main.go
```diff
@@ -288,12 +333,17 @@ func getLastHeight() int {
 	return height
 }
 
+func setup() {
```
BrandonWeng: Do you still need this if we have `GetValidators` already?

loadtest/config.json
```diff
@@ -15,9 +15,12 @@
...
     "message_type_distribution": {
         "limit_order_percentage": "0.2",
-        "market_order_percentage": "0.8"
+        "market_order_percentage": "0.8",
+        "delegate_percentage": "0.5",
+        "undelegate_percentage": "0.25",
+        "begin_redelegate_percentage": "0.25"
     },
-    "message_type": "basic",
+    "message_type": "staking",
```
BrandonWeng: could we revert this to basic

loadtest/main.go
```diff
@@ -60,11 +61,14 @@ func (d *NumericDistribution) Sample() sdk.Dec {
 }
 
 type MsgTypeDistribution struct {
-	LimitOrderPct  sdk.Dec `json:"limit_order_percentage"`
-	MarketOrderPct sdk.Dec `json:"market_order_percentage"`
+	LimitOrderPct      sdk.Dec `json:"limit_order_percentage"`
+	MarketOrderPct     sdk.Dec `json:"market_order_percentage"`
+	DelegatePct        sdk.Dec `json:"delegate_percentage"`
+	UndelegatePct      sdk.Dec `json:"undelegate_percentage"`
+	BeginRedelegatePct sdk.Dec `json:"begin_redelegate_percentage"`
```
BrandonWeng: Might be useful to add some comments to separate it based on usage

loadtest/main.go
eric-zhu-uw: good catch, forgot to remove this

loadtest/main.go
eric-zhu-uw: sounds good, will add comments

loadtest/config.json
eric-zhu-uw: sg

BrandonWeng (approved):
🔥

PR #334, opened 2022-10-21
Oracle parallel

udpatil:
This contains the implementation for oracle vote parallelization via aclmapping dynamic generators. Additionally, this PR is stacked on top of a cherry picked PR that previously modified oracle so as to apply the parallelizatin changes to the latest oracle code. 
Also added a makefile target to build loadtest for faster development

Tested using loadtest cluster, and also added unit tests for the parallelization dynamic generators to validate the access operations to ensure that they can't be unknowingly changed to violate the validation steps.


go.mod
```diff
@@ -131,7 +131,7 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.1.176
+	github.com/cosmos/cosmos-sdk => ../sei-cosmos //github.com/sei-protocol/sei-cosmos v0.1.193
```
udpatil: reminder to remove local dep

loadtest/loadtest_client.go
```diff
@@ -89,32 +89,53 @@ func (c *LoadTestClient) BuildTxs() (workgroups []*sync.WaitGroup, sendersList [
...
 
+	valKeys := c.SignerClient.GetValKeys()
+
 	for i := 0; i < int(config.Rounds); i++ {
 		fmt.Printf("Preparing %d-th round\n", i)
 
 		wg := &sync.WaitGroup{}
 		var senders []func()
 		workgroups = append(workgroups, wg)
+		if config.MessageType != "none" {
```
BrandonWeng, line 101: whats this check for?

loadtest/loadtest_client.go
```diff
@@ -89,32 +89,53 @@ func (c *LoadTestClient) BuildTxs() (workgroups []*sync.WaitGroup, sendersList [
...
+				_ = txBuilder.SetMsgs(msg)
+				seqDelta := uint64(i)
+				mode := typestx.BroadcastMode_BROADCAST_MODE_SYNC
+				sender := SendTx(valKey, &txBuilder, mode, seqDelta, *c)
+				wg.Add(1)
+				senders = append(senders, func() {
+					defer wg.Done()
+					sender()
+				})
+			}
```
BrandonWeng: kind of a nit but it seems like we have multiple things happening in preparing round now, it would be nice if they're in separate methods

app/antedecorators/gasless.go
```diff
@@ -45,18 +45,11 @@ func (gd GaslessDecorator) AnteDeps(txDeps []sdkacltypes.AccessOperation, tx sdk
 	for _, msg := range tx.GetMsgs() {
 		// Error checking will be handled in AnteHandler
 		switch m := msg.(type) {
-		case *oracletypes.MsgAggregateExchangeRatePrevote:
-			feederAddr, _ := sdk.AccAddressFromBech32(m.Feeder)
-			valAddr, _ := sdk.ValAddressFromBech32(m.Validator)
-			deps = append(deps, aclutils.GetOracleReadAccessOpsForValAndFeeder(feederAddr, valAddr)...)
```
BrandonWeng, line 51: is this part of the cherry-picked changes?

BrandonWeng (approved):
lgtm with non-blocking nits

loadtest/loadtest_client.go
udpatil: I added a "none" type so that I could run ONLY oracle txs on the cluster, if we don't think we want to keep it I can remove it

loadtest/loadtest_client.go
udpatil: yeah good point, I'll make a helper fn

app/antedecorators/gasless.go
udpatil: yup, it removed the `prevote` and `combined vote` from oracle

udpatil: well, this exact change wasn't cherry picked, but I removed this to match the removal in the cherry picked section

PR #336, opened 2022-10-24
Refactor Loadtest script

BrandonWeng:
Concurrency didn't seem to be the bottleneck. 

* Refactored the script 
* Made a cache for the signer/account seqno responses
   * Instead of querying for each Seq No, we will just manage it locally
* For opening test account files, it will be the same, and read it from memory instead of from file each time 

Went from ~5 minutes -> 2 minutes

The first two rounds take the longest (active/inactive accounts fetching live info) then the remaining 3 rounds are super fast since it's just reading from the cache. 
image

Without this change:
image

### DIFF
--- a/loadtest/main.go
+++ b/loadtest/main.go
@@ -53,7 +53,7 @@ func run() {
 	defer client.Close()
 
 	if config.TxsPerBlock < config.MsgsPerTx {
-		panic("Must have more orders per block than batch size")
+		panic("Must have more TxsPerBlock than MsgsPerTx")
 	}
 
 	configString, _ := json.Marshal(config)
--- a/loadtest/main.go
+++ b/loadtest/main.go
@@ -6,102 +6,26 @@ import (
 	"math/rand"
 	"os"
 	"os/exec"
-	"path/filepath"
 	"strconv"
 	"sync"
 	"time"
 
 	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
+	"github.com/cosmos/cosmos-sdk/std"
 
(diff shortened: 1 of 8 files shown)


loadtest/main.go
```diff
@@ -127,93 +46,30 @@ func init() {
...
-	for i := 0; i < int(numberOfAccounts); i++ {
-		if i%2 == 0 {
-			activeAccounts = append(activeAccounts, i)
-		} else {
-			inactiveAccounts = append(inactiveAccounts, i)
-		}
+	defer client.Close()
+
+	if config.TxsPerBlock < config.MsgsPerTx {
+		panic("Must have more orders per block than batch size")
```
philipsu522: nit - update comment to "txs"

loadtest/sign.go
```diff
@@ -48,12 +85,19 @@ func GetKey(accountIdx uint64) cryptotypes.PrivKey {
 	algo, _ := keyring.NewSigningAlgoFromString(algoStr, keyringAlgos)
 	hdpath := hd.CreateHDPath(sdk.GetConfig().GetCoinType(), 0, 0).String()
 	derivedPriv, _ := algo.Derive()(accountInfo.Mnemonic, "", hdpath)
-	return algo.Generate()(derivedPriv)
+	privKey := algo.Generate()(derivedPriv)
+
+	// Cache this so we don't need to regenerate it
+	sc.CachedAccountKey.Store(accountIdx, privKey)
```
philipsu522, line 91: nice - didn't know this was a bottleneck

philipsu522 (approved):

PR #257, opened 2022-09-23
Register accesscontrol module 

BrandonWeng:
Adds ACL module to sei-chain 

Also had to make some fixes based on this PR:
https://github.com/sei-protocol/sei-cosmos/pull/23/files 

Tested this by starting a local chain with Phil script:
https://github.com/sei-protocol/sei-chain/pull/251/files#diff-eb97f7c509f5a70c4ca7de4773937b07c26a74d4d9142c5ec1c97896dece4318L1

PR #344, opened 2022-10-27
Register accesscontrol module (#257)

udpatil:


app/app.go
```diff
@@ -817,6 +864,15 @@ func (app *App) SetStoreUpgradeHandlers() {
...
 	}
+
+	if upgradeInfo.Name == "2.0.0beta" && !app.UpgradeKeeper.IsSkipHeight(upgradeInfo.Height) {
+		storeUpgrades := storetypes.StoreUpgrades{
+			Added: []string{acltypes.StoreKey},
+		}
+
+		// configure store loader that checks if version == upgradeHeight and applies store upgrades
+		app.SetStoreLoader(upgradetypes.UpgradeStoreLoader(upgradeInfo.Height, &storeUpgrades))
+	}
```
philipsu522, line 879: what's the plan for this? One idea: we leave atlantic-1 using 1.x.x upgrades and cherry-pick out 2.0.0beta PRs so we theroetically never hit this

loadtest/config.json
```diff
@@ -1,63 +1,71 @@
 {
```
philipsu522, line 1: we should revert this. The one on the left is the correct one

loadtest/main.go
```diff
@@ -301,23 +107,36 @@ func generateMessage(config Config, key cryptotypes.PrivKey, batchSize uint64) (
...
 		}
 	case Staking:
-		delegatorAddr := sdk.AccAddress(key.PubKey().Address()).String()
-		chosenValidator := Validators[rand.Intn(len(Validators))].OperatorAddress
-		// Randomly pick someone to redelegate / unbond from
-		srcAddr := ""
-		for k := range DelegationMap[delegatorAddr] {
-			if k == chosenValidator {
-				continue
+		msgType := config.MsgTypeDistr.SampleStakingMsgs()
```
philipsu522: are these changes intentional? Looks like we're throwing out some of the logic that's required to ensure we have existing delegations

loadtest/config.json
udpatil: the main difference is the batch size + orders per block vs msgs per tx / txs per block. I believe the latter was refactored by @eric-zhu-uw when making loadtest changes. Do we not want to use that pattern anymore?

loadtest/main.go
udpatil: Ah yeah this needs some refactoring. I had decided to use erics version of the staking msg sampling because it allowed for selecting what types of staking message distributions to use but I think the left side one is better, I can refactor

app/app.go
udpatil: does that mean we don't want to upgrade atlantic-1 anymore after the rebase? because a lot of the rebased changes are dependent on like accesscontrol and such which is part of 2.0.0

BrandonWeng (approved):
overall looks fine for the parallel tx work, and the load testing client seems good too but it would be good to get some testing in. 

Will leave the WASM changes for tony to double check

codchen (approved):

PR #345, opened 2022-10-27
Validate Concurrent Messages + Update BankSend 

BrandonWeng:

* Updates the banksend acl mapping to use the prefix/storekeys as the identifiers rather than arbitrary ones
* Adds logic for handling failovers in the case when concurrent executions fail 
* Adds metric to count failed TX counts, we should add alerting on this in the future 

Will need to update the mapping for all the other modules

Unit tests

Deployed to load test cluster and was able to run bank sends e2e and verified that the height increases
image

Also manually forced failure scenario
image

### DIFF
--- a/aclmapping/bank/mapping_test.go
+++ b/aclmapping/bank/mapping_test.go
@@ -1,0 +1,155 @@
+package aclbankmapping
+
+import (
+	"testing"
+
+	"github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1"
+	"github.com/cosmos/cosmos-sdk/simapp"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkacltypes "github.com/cosmos/cosmos-sdk/types/accesscontrol"
+	acltypes "github.com/cosmos/cosmos-sdk/x/accesscontrol/types"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	"github.com/cosmos/cosmos-sdk/x/bank"
+	"github.com/cosmos/cosmos-sdk/x/bank/types"
+	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
+	oracletypes "github.com/sei-protocol/sei-chain/x/oracle/types"
+	"github.com/stretchr/testify/require"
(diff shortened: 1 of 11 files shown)


aclmapping/bank/mapping_test.go
```diff
@@ -0,0 +1,154 @@
...
+			} else {
+				require.NoError(t, err)
+			}
+			missing := (sdkacltypes.ValidateAccessOperations(depdenencies, cms.GetEvents()))
+			require.Empty(t, missing)
+		})
+	}
+}
+
+func TestGeneratorInvalidMessageTypes(t *testing.T) {
```
BrandonWeng: This is the leaf node unit test, decided to just add it in this PR too

codchen (approved):

aclmapping/bank/mappings.go
```diff
@@ -37,46 +39,53 @@ func MsgSendDependencyGenerator(keeper aclkeeper.Keeper, ctx sdk.Context, msg sd
...
 			AccessType:         sdkacltypes.AccessType_WRITE,
-			ResourceType:       sdkacltypes.ResourceType_KV,
-			IdentifierTemplate: utils.GetIdentifierTemplatePerModule(utils.AUTH, msgSend.ToAddress),
+			ResourceType:       sdkacltypes.ResourceType_KV_AUTH_ADDRESS_STORE,
+			IdentifierTemplate: string(authtypes.CreateAddressStoreKeyFromBech32(msgSend.ToAddress)),
+		},
+
+		// Gets Account Info for the sender
+		{
+			AccessType:         sdkacltypes.AccessType_READ,
```
eric-zhu-uw, line 78: for the `ToAddress`, there is a write access but for the `FromAddress` there is no write access, only read. Just want to make sure it wasn't missed.

eric-zhu-uw (approved):

aclmapping/bank/mappings.go
BrandonWeng: Yeee, I believe it has to exist already based  (If it doesn't exist then there's no money to be sent)  on the logic we only create accounts for the receiver if it doesn't exist

PR #351, opened 2022-10-29
Update Oracle Parallel Tx Resource Identifiers 

BrandonWeng:
Updates the identifiers to use the actual keys and also fixes one of them

Mainly unit testing since the resource keys and the underlying value isn't changing


aclmapping/oracle/mappings.go
```diff
@@ -55,15 +58,12 @@ func MsgVoteDependencyGenerator(keeper aclkeeper.Keeper, ctx sdk.Context, msg sd
 		// set exchange rate vote - WRITE
 		{
 			ResourceType:       sdkacltypes.ResourceType_KV_ORACLE_AGGREGATE_VOTES,
-			AccessType:         sdkacltypes.AccessType_READ,
-			IdentifierTemplate: msgVote.Validator,
+			AccessType:         sdkacltypes.AccessType_WRITE,
```
BrandonWeng, line 61: This is a SET operation so it should have been a write (validation logic caught this)

philipsu522 (approved):

PR #356, opened 2022-11-01
Add ContractInfoV2 that supports rent deposit

codchen:
We want contracts that register with `dex` module to be able to deposit rent which will be spent for the contract's sudo endpoint invocations. This PR adds a new ContractInfoV2 type that holds a rent balance of `usei`. The actual `usei` is transferred from the contract registerer to `dex`'s module account. The contract registerer can deposit more rent to his contract through the new `ContractDepositRent` message type. The contract registerer can also unregister his contract and get his `usei` refunded.

unit tests

### DIFF
--- a/x/dex/keeper/msgserver/msg_server_contract_deposit_rent.go
+++ b/x/dex/keeper/msgserver/msg_server_contract_deposit_rent.go
@@ -22,6 +22,9 @@ func (k msgServer) ContractDepositRent(goCtx context.Context, msg *types.MsgCont
 	}
 	// deposit
 	creatorAddr, err := sdk.AccAddressFromBech32(contract.Creator)
+	if err != nil {
+		return nil, err
+	}
 	if err := k.BankKeeper.SendCoins(ctx, creatorAddr, k.AccountKeeper.GetModuleAddress(types.ModuleName), sdk.NewCoins(sdk.NewCoin(appparams.BaseCoinUnit, sdk.NewInt(int64(msg.Amount))))); err != nil {
 		return nil, err
 	}
--- a/x/dex/keeper/msgserver/msg_server_contract_deposit_rent.go
+++ b/x/dex/keeper/msgserver/msg_server_contract_deposit_rent.go
@@ -1,0 +1,34 @@
+package msgserver
+
+import (
+	"context"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	appparams "github.com/sei-protocol/sei-chain/app/params"
+	"github.com/sei-protocol/sei-chain/x/dex/types"
+)
+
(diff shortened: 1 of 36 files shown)


x/dex/keeper/msgserver/msg_server_contract_deposit_rent.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+func (k msgServer) ContractDepositRent(goCtx context.Context, msg *types.MsgContractDepositRent) (*types.MsgContractDepositRentResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	contract, err := k.GetContract(ctx, msg.ContractAddr)
+	if err != nil {
+		return nil, err
+	}
+	if contract.RentBalance > 0 && contract.Creator != msg.Sender {
```
LCyson, line 19: why don't we just use whether the creator == sender to check here, and let people to deposit whenever they want

x/dex/keeper/msgserver/msg_server_register_contract.go
```diff
@@ -166,9 +172,43 @@ func (k msgServer) UpdateNewDependencies(ctx sdk.Context, msg *types.MsgRegister
...
+		if msg.Contract.RentBalance < existingContract.RentBalance {
+			// refund
+			refundAmount := existingContract.RentBalance - msg.Contract.RentBalance
+			if err := k.BankKeeper.SendCoins(ctx, k.AccountKeeper.GetModuleAddress(types.ModuleName), creatorAddr, sdk.NewCoins(sdk.NewCoin(appparams.BaseCoinUnit, sdk.NewInt(int64(refundAmount))))); err != nil {
+				return err
+			}
+		} else if msg.Contract.RentBalance > existingContract.RentBalance {
+			// deposit
+			depositAmount := msg.Contract.RentBalance - existingContract.RentBalance
+			if err := k.BankKeeper.SendCoins(ctx, creatorAddr, k.AccountKeeper.GetModuleAddress(types.ModuleName), sdk.NewCoins(sdk.NewCoin(appparams.BaseCoinUnit, sdk.NewInt(int64(depositAmount))))); err != nil {
```
LCyson, line 200: nit: we probably want to make this clear in the doc that the creator's account will be charged if they specify a larger than existing balance rentBalance

LCyson (approved):

x/dex/keeper/msgserver/msg_server_contract_deposit_rent.go
codchen: just concerned about legacy contracts which doesn't have a creator set. We don't want multiple people to deposit to it because only one can withdraw

x/dex/keeper/msgserver/msg_server_register_contract.go
codchen: yeah good point

x/dex/keeper/msgserver/msg_server_contract_deposit_rent.go
LCyson: think there is a chance that an attack can take the creator role if the contract isn't registered with rent

PR #360, opened 2022-11-01
Add ACL Mapping for Dex Send and Cancel orders 

BrandonWeng:
* Adding the ACL mapping for dex send and cancel orders
* Also fix the completion signal mapping, we were getting lucky previously with the testing

From Eric's changes: Also updates load testing script to send cancellation orders

Mainly unit tests for now - trying to run e2e with loadtest cluster

image

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1018,10 +1018,8 @@ func (app *App) ProcessTxConcurrent(
 	)
 
 	// Deliver the transaction and store the result in the channel
-	ctx.Logger().Info(fmt.Sprintf("DEBUG: Processing txIndex=%d", txIndex))
 	resultChan <- ChannelResult{txIndex, app.DeliverTxWithResult(ctx, txBytes)}
 	metrics.IncrTxProcessTypeCounter(metrics.CONCURRENT)
-	ctx.Logger().Info(fmt.Sprintf("DEBUG: Finished txIndex=%d", txIndex))
 }
 
 func (app *App) ProcessBlockConcurrent(
@@ -1067,7 +1065,6 @@ func (app *App) ProcessBlockConcurrent(
 	// Gather Results and store it based on txIndex and read results from channel
 	// Concurrent results may be in different order than the original txIndex
 	txResultsMap := map[int]*abci.ExecTxResult{}
-	ctx.Logger().Info(fmt.Sprintf("DEBUG: Waiting for %d results", len(txs)))
 	for result := range resultChan {
 		txResultsMap[result.txIndex] = result.result
 	}
@@ -1078,7 +1075,6 @@ func (app *App) ProcessBlockConcurrent(
 	}
 
 	ok := true
-	ctx.Logger().Info(fmt.Sprintf("DEBUG: Parsing Results for %d results", len(txs)))
 	for i, result := range txResults {
 		if result.GetCode() == sdkerrors.ErrInvalidConcurrencyExecution.ABCICode() {
 			ctx.Logger().Error(fmt.Sprintf("Invalid concurrent execution of deliverTx index=%d", i))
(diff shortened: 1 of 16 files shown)


app/app.go
```diff
@@ -964,12 +965,17 @@ func (app *App) ProcessBlockSynchronous(ctx sdk.Context, txs [][]byte) []*abci.E
...
 	for messageIndex, accessOperationsToSignal := range signalMapping {
 		for accessOperation, completionSignals := range accessOperationsToSignal {
 			var channels []chan interface{}
-			channelsMapping[messageIndex] = make(sdkacltypes.AccessOpsChannelMapping)
+			if val, ok := channelsMapping[messageIndex]; ok {
+				channels = val[accessOperation]
+			} else {
+				channelsMapping[messageIndex] = make(sdkacltypes.AccessOpsChannelMapping)
+			}
+
```
BrandonWeng: This was a bug, added a unit test to catch this. This resets the channelMapping for each index every time so we really only have one accessOperation channel to signal off. Probably less of a problem when we had much simpler messaging dependencies previously it would happen periodically but now it's consistently happening.

codchen (approved):

app/app.go
```diff
@@ -964,12 +965,17 @@ func (app *App) ProcessBlockSynchronous(ctx sdk.Context, txs [][]byte) []*abci.E
...
 // Returns a mapping of the accessOperation to the channels
-func getChannelsFromSignalMapping(signalMapping acltypes.MessageCompletionSignalMapping) sdkacltypes.MessageAccessOpsChannelMapping {
+func GetChannelsFromSignalMapping(signalMapping acltypes.MessageCompletionSignalMapping) sdkacltypes.MessageAccessOpsChannelMapping {
 	channelsMapping := make(sdkacltypes.MessageAccessOpsChannelMapping)
 	for messageIndex, accessOperationsToSignal := range signalMapping {
 		for accessOperation, completionSignals := range accessOperationsToSignal {
 			var channels []chan interface{}
-			channelsMapping[messageIndex] = make(sdkacltypes.AccessOpsChannelMapping)
+			if val, ok := channelsMapping[messageIndex]; ok {
+				channels = val[accessOperation]
```
udpatil: why do we need to get the channels here? since we're iterating over accessOperation, shouldnt we be starting with a fresh slice each iteration? (aka there shouldnt be an entry for `val[accessOperation]` right?

loadtest/config.json
```diff
@@ -15,32 +15,36 @@
...
         }
     },
-    "run_oracle": false,
-    "message_type": "basic",
+    "order_type_distribution": {
+        "limit_order_percentage": "0.2",
+        "market_order_percentage": "0.8"
+    },
+    "message_type": "dex",
+    "run_oracle": true,
```
udpatil: nit: you might want to leave `run_oracle` false since idt your ansible script (or the one committed in master) has the steps for getting validator keys

loadtest/main.go
```diff
@@ -110,48 +111,89 @@ func generateMessage(config Config, key cryptotypes.PrivKey, msgPerTx uint64, va
...
-		amount, err := sdk.ParseCoinsNormalized(fmt.Sprintf("%d%s", price.Mul(quantity).Ceil().RoundInt64(), "usei"))
-		if err != nil {
-			panic(err)
-		}
-		msg = &dextypes.MsgPlaceOrders{
-			Creator:      sdk.AccAddress(key.PubKey().Address()).String(),
-			Orders:       orderPlacements,
-			ContractAddr: contract,
-			Funds:        amount,
+		msgType := c.LoadTestConfig.MsgTypeDistr.SampleDexMsgs()
```
udpatil: can you rebase your loadtest script onto `cherry-pick-parallel` instead of 2.0.0beta? theres several changes made there that I'd prefer get included by default instead of rebasing conflicts

BrandonWeng: Im getting a lot of merge conflicts when trying to rebase it on loadtest/main.go, maybe i can just revert the loadtest/changes to 2.0.0beta and make the loadtest/changes against your branch in another PR?

loadtest/config.json
BrandonWeng: Noted for next PR: https://github.com/sei-protocol/sei-chain/pull/360#discussion_r1011800396

app/app.go
BrandonWeng: Im gonna remove this if statement entirely, we can move  `channelsMapping[messageIndex] = make(sdkacltypes.AccessOpsChannelMapping)`to the outer loop so it's only instantiated once

udpatil (approved):

PR #363, opened 2022-11-02
Sort keys in TypedSyncMap before applying

philipsu522:
As a follow up to the 1.2.3beta outage, iterating over map keys may result in nondeterminism, especially since som places iterate over the map and apply the changes to keeper (e.g. https://github.com/sei-protocol/sei-chain/blob/6f2f39f912a3048d5f0ef6291c3eb1330c024dc3/x/dex/keeper/utils/order_book.go#L48)

Add unit tests (they failed, now they pass)


utils/datastructures/typed_sync_map_test.go
```diff
@@ -78,13 +78,20 @@ func TestTypedSyncMapDeepCopy(t *testing.T) {
 }
 
 func TestTypedSyncMapDeepApply(t *testing.T) {
-	m := datastructures.NewTypedSyncMap[int, int]()
-	m.Store(1, 1)
-	m.Store(2, 2)
+	m := datastructures.NewTypedSyncMap[string, int]()
+	m.Store("a", 1)
```
philipsu522, line 82: changed these to str since all uses of TypedMap are strings (or contractAddrs)

PR #375, opened 2022-11-04
Add address validity check in dex contract registration

LCyson:
This PR adds bech32 check in the dex contract registration process. Tested e2e and via unit tests.

TODO:
- Check if the contract has been instantiated by `wasmKeeper.HasContract(x)`, couldn't add this in bc we need to create a wasmKeeper mock object otherwise it will break the entire contract registration unit test.


Kbhat1 (approved):
LGTM, thanks

PR #380, opened 2022-11-08
Fix partial message handling

codchen:
We should not add things to mem store if there could be errors being thrown later. This PR changes things such that we only add to mem store when there is no more error. This is still not ideal since there could be future logics slipping in, so we need a more robust mem store rollback mechanism.

TODO unit test

### DIFF
--- a/x/dex/keeper/msgserver/msg_server_place_orders.go
+++ b/x/dex/keeper/msgserver/msg_server_place_orders.go
@@ -39,11 +39,6 @@ func (k msgServer) transferFunds(goCtx context.Context, msg *types.MsgPlaceOrder
 		if fund.Amount.IsNil() || fund.IsNegative() {
 			return errors.New("fund deposits cannot be nil or negative")
 		}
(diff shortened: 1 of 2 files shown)


x/dex/keeper/msgserver/msg_server_place_orders.go
```diff
@@ -73,11 +69,23 @@ func (k msgServer) PlaceOrders(goCtx context.Context, msg *types.MsgPlaceOrders)
 		order.Id = nextID
 		order.Account = msg.Creator
 		order.ContractAddr = msg.GetContractAddr()
-		dexutils.GetMemState(ctx.Context()).GetBlockOrders(ctx, typesutils.ContractAddress(msg.GetContractAddr()), pairStr).Add(order)
+		ordersToAdd[pairStr] = order
 		idsInResp = append(idsInResp, nextID)
 		nextID++
 	}
+	// setting bumped order id BEFORE adding orders to cache, so that if the latter fails, at least we wouldn't have
```
philipsu522, line 76: hm, thinking about this more, how does duplicated orderId happen again? like if the order fails, it theoretically should never show up so we should be able to reuse the order.
Or is the idea here just to make it more resilient b/c there's the unknown ongoing issue?

codchen:
We don't need this anymore now that we are moving from memstate to keeper. Closing

PR #381, opened 2022-11-08
[oracle] Add price feeder sidecar

udpatil:
This is largely forked from the Kujira Oracle Price Feeder sidecar, with changes made for compatibility with sei-chain. This will need to be further updated to update the provider integrations, and to improve the documentation for running this oracle process.

Specific Files of interest:
oracle/price_feeder/oracle.go
oracle/cmd/price_feeder.go

Most of the files have minor modifications specifically for import differences and compatibility differences as opposed to complete rewrites.

Tested on local chain with some denoms.


oracle/price-feeder/oracle/oracle.go
```diff
@@ -0,0 +1,581 @@
...
+		providerCandles,
+		providerPrices,
+		o.providerPairs,
+		o.deviations,
+	)
+	if err != nil {
+		return err
+	}
+
+	if len(computedPrices) != len(requiredRates) {
```
philipsu522, line 258: this logic seems like the price feeder is all-or-nothing. do we want to allow validators to submit prices for whatever they can get?

oracle/price-feeder/dockerfile/README.md
```diff
@@ -0,0 +1,102 @@
...
+keyring-backend = "file"
+output = "text"
+node = "tcp://rpc-kujira.synergynodes.com:80"
+broadcast-mode = "sync"
+EOF
+```
+
+## Recover oracle `keyring-file` to local file
+```bash
+kujirad keys add oracle --keyring-backend file --recover
```
philipsu522: should we change this (and all other references to kujira) to sei?

oracle/price-feeder/oracle/oracle.go
udpatil: we could if we want, but I'd prefer to first add this as a TODO and get the sidecar merged in. We can already tune the config toml to exclude certain denominations and thats no issue even if we dont support all whitelisted assets, whereas this relates to the presence of rates for the denoms we have in the config. If it sounds good I'll add in a TODO for more graceful exclusion of prices that are missing

oracle/price-feeder/dockerfile/README.md
udpatil: yup good call

oracle/price-feeder/oracle/oracle.go
udpatil: added todo

philipsu522 (approved):

PR #383, opened 2022-11-08
Add wrapper for load tests to run in different modes

BrandonWeng:
Adding a python script wrapper to be able to run load tests with different configurations, it uses the existing loadtest/config.json as the base and creates a copy of it with different 

Will make another PR to create service config files for these to run internally

Burst:
* Much larger tx per block but less cadence

Steady:
* Less TX per block but the higher cadence 

Steady Mode
image

Burst mode 
image

### DIFF
--- a/loadtest/scripts/run_tests.py
+++ b/loadtest/scripts/run_tests.py
@@ -1,0 +1,113 @@
+import argparse
+import json
+import os
+import subprocess
+import tempfile
+
+from dataclasses import dataclass
+
+
+# This strategy is used to simulate sudden spikes or bursts of load on the on the chain.
+# Burst load testing is done to ensure that the chain hub is able to handle such sudden
+# spikes in traffic.
+BURST = 'BURST'
+
+#  This is used to simulate a "steady-state" scenario where there is a constant load.
+STEADY = 'STEADY'
+
+
+@dataclass
+class LoadTestConfig:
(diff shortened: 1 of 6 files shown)


loadtest/scripts/run_tests.py
```diff
@@ -0,0 +1,113 @@
...
+
+def write_to_temp_json_file(data):
+    with tempfile.NamedTemporaryFile(mode='w', encoding='utf-8', delete=False) as tmp:
+        json.dump(data, tmp, ensure_ascii=False)
+        return tmp
+
+def create_burst_loadtest_config(base_config_json):
+    new_config = base_config_json.copy()
+    new_config["constant"] = True
+    new_config["metrics_port"] = 9697
```
udpatil, line 33: is there a specific reason we want a different metrics port for burst?

udpatil: also for clarity, do we plan to run the chain with Both configs? aka start with some baseline steady load and add burst load on top of that?

udpatil (approved):

loadtest/scripts/run_tests.py
BrandonWeng: yeah, its possible we may want to run both at the same time. These would be executed as different binaries so they would need to listen on different ports

PR #388, opened 2022-11-10
Update tick size change from proposal to transaction

Kbhat1:
- Update the update tick size from a proposal to a transaction
- Adds tx checks that msg is well formed, only original contract creator can update tick size
- Tested e2e and added in unit tests for tx


proto/dex/tx.proto
```diff
@@ -105,4 +107,17 @@ message MsgRegisterPairs {
 
 message MsgRegisterPairsResponse {}
 
+message MsgUpdateTickSize {
+  string creator = 1 [
+    (gogoproto.jsontag) = "creator"
+  ];
+  repeated TickSize tickSizeList = 3 [
```
LCyson: ` tickSizeList = 2`?

LCyson (approved):

PR #395, opened 2022-11-14
Set default configs that are more optimal

philipsu522:
Sets configs such that:
- metrics are automatically enabled
- pruning configs are optimal (discussed with some validators)
- use pebbledb as backend
- Tested locally, verified app toml configs are updated
app.toml:
```
 cat ~/.sei/config/app.toml| grep "prun\|prom" | grep -v "#"           ok | 01:50:04 PM
pruning = "custom"
pruning-keep-recent = "2000"
pruning-keep-every = "0"
pruning-interval = "3079"
prometheus-retention-time = 60
```
config.toml:
```
cat ~/.sei/config/config.toml | grep "prom\|backend" | grep -v "#"
db-backend = "pebbledb"
prometheus = true
prometheus-listen-addr = ":26660"
```


cmd/seid/cmd/root.go
```diff
@@ -335,10 +358,25 @@ func initAppConfig() (string, interface{}) {
 	srvCfg.MinGasPrices = "0.01usei,0.001ibc/6D45A5CD1AADE4B527E459025AC1A5AEF41AE99091EF3069F3FEAACAFCECCD21"
 	srvCfg.API.Enable = true
 
+	// Pruning configs
+	srvCfg.Pruning = "custom"
+	srvCfg.PruningKeepRecent = "2000"
+	// Randomly generate pruning interval. We want the following properties:
+	//   - random: if eveyrone has the same value, the block that everyone prunes will be slow
```
BrandonWeng:
```suggestion
	//   - random: if everyone has the same value, the block that everyone prunes will be slow
```

BrandonWeng: oh this is a good idea

BrandonWeng (approved):
lgtm

PR #397, opened 2022-11-15
Move order placement in MemState to store

codchen:
Change the underlying data of order placement in MemState to be backed by KV store. Some details to note:
- `GetBlockOrders` always instantiates a new `BlockOrders` which is now just a wrapper over the underlying KV store
- `DeepCopy` of `BlockOrders` don't really do anything
- these order placement KV entries are deleted at the end of each block because they don't need to get written to disk during `Commit()`

unit tests


x/dex/contract/abci.go
```diff
@@ -203,6 +203,7 @@ func orderMatchingRunnable(ctx context.Context, sdkContext sdk.Context, env *env
 	sdkContext.Logger().Info(fmt.Sprintf("End block for %s", contractInfo.ContractAddr))
 	pairs, pairFound := env.registeredPairs.Load(contractInfo.ContractAddr)
 	orderBooks, found := env.orderBooks.Load(contractInfo.ContractAddr)
+	fmt.Println(env.failedContractAddresses.Size())
```
LCyson: nit: remove println

LCyson (approved):

PR #400, opened 2022-11-15
[ante] Fix gasless ante and allow for specifying gas with gasless txs

udpatil:
This fixes the regression with gasless TXs, once again allowing them to be executed without spending any gas. One change here is that gasless TXs NEED to specify zero gas in the tx to be processed as gasless. Otherwise, we will treat them as non-gasless in order to allow for tx prioritization via higher gas prices.
Verified gasless and gas enabled oracle txs

### DIFF
--- a/oracle/price-feeder/oracle/client/tx.go
+++ b/oracle/price-feeder/oracle/client/tx.go
@@ -19,12 +19,8 @@ func BroadcastTx(clientCtx client.Context, txf tx.Factory, msgs ...sdk.Msg) (*sd
 		return nil, err
 	}
 
-	// _, adjusted, err := tx.CalculateGas(clientCtx, txf, msgs...)
-	// if err != nil {
-	// 	return nil, err
-	// }
-
-	// txf = txf.WithGas(adjusted)
+	txf = txf.WithGas(0)
+	// txf = txf.WithFees("0usei")
 
 	unsignedTx, err := tx.BuildUnsignedTx(txf, msgs...)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


philipsu522 (approved):
nice!

BrandonWeng (approved):

oracle/price-feeder/oracle/client/tx.go
```diff
@@ -25,15 +25,15 @@ func BroadcastTx(clientCtx client.Context, txf tx.Factory, msgs ...sdk.Msg) (*sd
...
+	// txf = txf.WithFees("0usei")
 
 	unsignedTx, err := tx.BuildUnsignedTx(txf, msgs...)
 	if err != nil {
 		return nil, err
 	}
 
-	// unsignedTx.SetFeeGranter(clientCtx.GetFeeGranterAddress())
-	unsignedTx.SetFeeAmount(sdk.NewCoins(sdk.NewCoin("usei", sdk.NewInt(0))))
-	// unsignedTx.SetFeePayer(clientCtx.GetFeePayerAddress())
```
BrandonWeng, line 36: how come we dont need to set the fee amount anymore?

oracle/price-feeder/oracle/client/tx.go
```diff
@@ -25,15 +25,15 @@ func BroadcastTx(clientCtx client.Context, txf tx.Factory, msgs ...sdk.Msg) (*sd
 	// }
 
 	// txf = txf.WithGas(adjusted)
+	txf = txf.WithGas(0)
+	// txf = txf.WithFees("0usei")
```
BrandonWeng, line 29: Whats the difference between WithFees vs SetFeeAmount?

udpatil: WithFees is on the factory level, and SetFeeAmount modifies the unsigned tx as a one off adjustment

oracle/price-feeder/oracle/client/tx.go
udpatil: because we don't need to send any fees with the transaction (I believe this behaves the same as explicitly setting 0?)

PR #410, opened 2022-11-22
Change default rent price to 0

codchen:
This PR does 3 things:
1. Fix a race condition in determining gas consumed by a sudo call. Previously it reads a shared gas meter before and after the `sudo` call, but since the reads and the sudo call are not atomic, race conditions across multiple threads using the same gas meter are possible and observed in loadtest.
2. Fix the rollback logic where the original context (hence the original store) instead of the cached store is returned if the processing fails.
3. We will not launch with sudo rent initially, so setting default sudo gas price to 0.

unit test


BrandonWeng (approved):
thanks for the fix. Going to merge it to unblock and test it out!

PR #412, opened 2022-11-22
Increase gas limit for loadtest TXs

BrandonWeng:
Seeing some `out of gas in location: WriteFlat; gasWanted: 200000, gasUsed: 201581: out of gas` in the latest client, looks like we need to bump up the gas for some of these transactions

Also updates contract addr based on recent changes, it might still be different between runs

### DIFF
--- a/loadtest/tx.go
+++ b/loadtest/tx.go
@@ -20,7 +20,7 @@ func SendTx(
 	failureExpected bool,
 	loadtestClient LoadTestClient,
 ) func() {
-	(*txBuilder).SetGasLimit(200000)
+	(*txBuilder).SetGasLimit(300000)
 	(*txBuilder).SetFeeAmount([]sdk.Coin{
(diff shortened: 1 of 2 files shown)


loadtest/tx.go
```diff
@@ -20,7 +20,7 @@ func SendTx(
 	failureExpected bool,
 	loadtestClient LoadTestClient,
 ) func() {
-	(*txBuilder).SetGasLimit(200000)
+	(*txBuilder).SetGasLimit(300000)
```
udpatil, line 23: is this enough? sometimes it will say out of gas but then incrementing it by a little bit will be insufficient

udpatil (approved):

loadtest/tx.go
BrandonWeng: Should be here, we're adding another 100k and it's saying that its missing 1.5k

PR #425, opened 2022-11-23
[oracle] Add price snapshot history to the genesis export / import

udpatil:
This adds the price snapshot history to regenesis process so we don't lose it upon chain reinitialization.

Updated genesis unit tests, and will perform e2e testing.


Kbhat1 (approved):
LGTM, thanks

PR #427, opened 2022-11-24
Automate deploy_ten_contracts.sh

BrandonWeng:
Instead of taking arguments, we can just get the info from the local instance 

This also replaces the need to manually editing the loadtest config file every time too 
Ran it on a LT cluster and see that the config was replaced and able torun loadtest binary after

### DIFF
--- a/loadtest/contracts/deploy_ten_contracts.sh
+++ b/loadtest/contracts/deploy_ten_contracts.sh
@@ -1,16 +1,8 @@
 #!/bin/bash
-echo -n Admin Key Name:
-read keyname
-echo
-echo -n Chain ID:
-read chainid
-echo
-echo -n seid binary:
-read seidbin
-echo
-echo -n sei-chain directory:
-read seihome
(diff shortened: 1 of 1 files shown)


loadtest/contracts/deploy_ten_contracts.sh
```diff
@@ -1,16 +1,8 @@
...
-echo -n seid binary:
-read seidbin
-echo
-echo -n sei-chain directory:
-read seihome
-echo
+keyname=$(printf "12345678\n" | seid keys list --output json | jq ".[0].name" | tr -d '"')
+seidbin=$(which seid | tr -d '"')
+chainid=$(seid status | jq ".NodeInfo.network" | tr -d '"')
+seihome=$(git rev-parse --show-toplevel | tr -d '"')
```
BrandonWeng, line 5: Should be able to infer this

udpatil (approved):

PR #428, opened 2022-11-24
Add ante handler to check if dex order/cancelleation provides enough fees + Add logics to only charge contract rent if user gas cannot cover

codchen:
Check if a tx that contains dex orders/cancellations provide enough fees during ante phase.

To keep things simple, we will apply a flat rate for order and cancellation respectively. The only accepted denom for such txs is `usei` for the time being. The gas price will also use the flat rate of 0.1 usei/gas.


x/dex/keeper/contract.go
```diff
@@ -67,7 +67,12 @@ func (k Keeper) GetAllContractInfo(ctx sdk.Context) []types.ContractInfoV2 {
 	return list
 }
 
-func (k Keeper) ChargeRentForGas(ctx sdk.Context, contractAddr string, gasUsed uint64) error {
+func (k Keeper) ChargeRentForGas(ctx sdk.Context, contractAddr string, gasUsed uint64, userProvidedGas uint64) error {
```
LCyson, line 71: nit: can we add an updated comment for this function?

x/dex/ante.go
```diff
@@ -89,3 +89,45 @@ func IsDecimalMultipleOf(a, b sdk.Dec) bool {
...
+	dexKeeper keeper.Keeper
+}
+
+func NewCheckDexGasDecorator(dexKeeper keeper.Keeper) CheckDexGasDecorator {
+	return CheckDexGasDecorator{
+		dexKeeper: dexKeeper,
+	}
+}
+
+func (d CheckDexGasDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (newCtx sdk.Context, err error) {
```
LCyson, line 106: nit: comment for the function

x/dex/contract/abci.go
```diff
@@ -179,7 +179,7 @@ func handleFinalizedBlocks(ctx context.Context, sdkCtx sdk.Context, env *environ
 		if !contractsNeedHook.Contains(contractAddr) {
 			return true
 		}
-		if _, err := dexkeeperutils.CallContractSudo(sdkCtx, keeper, contractAddr, finalizeBlockMsg); err != nil {
+		if _, err := dexkeeperutils.CallContractSudo(sdkCtx, keeper, contractAddr, finalizeBlockMsg, 0); err != nil {
```
LCyson: can we replace the zero here with a constant variable

LCyson (approved):

PR #435, opened 2022-11-28
[ante] Add ante deps for gas meter setter

udpatil:
This properly adds the READ resource dependency on the wasm dependency mapping in the ante decorators for the gas meter setter. This is needed when we see wasm execute messages because the ante handler then reads from accesscontrol to get the dependencies to determine whether to allow gas discounts.

This also refactors identifier templates to use hex encoded strings

Tested with local sei to ensure that no more concurrent execution errors due to this access op being excluded

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -266,7 +266,7 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.1.295
+	github.com/cosmos/cosmos-sdk => ../sei-cosmos //github.com/sei-protocol/sei-cosmos v0.1.295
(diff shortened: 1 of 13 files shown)


go.mod
```diff
@@ -266,7 +266,7 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.1.295
+	github.com/cosmos/cosmos-sdk => ../sei-cosmos //github.com/sei-protocol/sei-cosmos v0.1.295
```
udpatil: self reminder to fix this with new cosmos version before merging

codchen (approved):

PR #443, opened 2022-12-02
Update seid start script with nitro lib build

yzang2019:
**Problem:**
The current master was broken after building and running seid locally. It is complaining about not able to find some nitro libraries.

`dyld[78984]: Library not loaded: /Users/tonychen/repos/nitro-replayer/target/release/deps/libnitro_replayer.dylib
  Referenced from: /Users/kartikbhat/go/bin/seid
  Reason: tried: '/Users/tonychen/repos/nitro-replayer/target/release/deps/libnitro_replayer.dylib' (no such file), '/usr/local/lib/libnitro_replayer.dylib' (no such file), '/usr/lib/libnitro_replayer.dylib' (no such file)
zsh: abort      seid`

**Fix:**
The way to fix it is to build from the latest nitro-replayer repo, and copy the latest library over to sei-chain.
This change fix the issue by:
1. Adding a new bash script to automate the whole process
2. Fix the existing initialize scripts to use the new build.sh script
3. Add some instructions in make file

Manually run and tested build.sh, initialize_local.sh

This should fail:
`./scripts/build.sh seid && ./build/seid` 

This should succeed:
`./scripts/build.sh seid-nitro && ./build/seid`

`./scripts/build.sh `
```
This script support build or install seid.
Usage: ./build.sh [seid, seid-nitro]
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,10 @@ ldflags := $(strip $(ldflags))
 BUILD_FLAGS := -tags "$(build_tags)" -ldflags '$(ldflags)'
 
 #### Command List ####
+# Make sure you are using latest nitro replayer library before make install:
+# 1. git clone https://github.com/sei-protocol/nitro-replayer
+# 2. cd nitro-replayer && cargo build --release && cd ..
+# 3. cp nitro-replayer/target/release/libnitro_replayer.dylib sei-chain/x/nitro/replay
 
 all: lint install
 
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+
+SEI_ROOT=$(git rev-parse --show-toplevel)
+
+# seid: only build seid (might failed to run without nitro library)
+# seid-nitro: build seid and also nitro replayer library
+display_usage() {
+  echo "This script support build or install seid."
+  echo "Usage: ./build.sh [seid, seid-nitro]"
+}
+
+build_nitro_replayer() {
+  NITRO_ROOT="$SEI_ROOT/../nitro-replayer/"
+  if [ ! -d "${NITRO_ROOT}" ]; then
+    cd "${SEI_ROOT}/../" || exit
+    git clone https://github.com/sei-protocol/nitro-replayer
+  fi
+  cd "${NITRO_ROOT}" || exit
+  cargo build --release
+  cp ${NITRO_ROOT}/target/release/libnitro_replayer.dylib "${SEI_ROOT}/x/nitro/replay"
+}
+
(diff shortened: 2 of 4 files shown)


scripts/build.sh
```diff
@@ -0,0 +1,55 @@
...
+   echo "Unrecognized option: $OPTION"
+   display_usage
+   exit 1
+   ;;
+esac
+
+cd "$SEI_ROOT" || exit
+make clean
+go build -o ./build/seid ./cmd/seid
+make install
```
philipsu522, line 55: re. my previous comment, it looks like you intended people to run scripts/build.sh to invoke make install. I think we want it the other way, since make install is the more standard way of running it

Makefile
```diff
@@ -58,6 +58,10 @@ ldflags := $(strip $(ldflags))
 BUILD_FLAGS := -tags "$(build_tags)" -ldflags '$(ldflags)'
 
 #### Command List ####
+# Make sure you are using latest nitro replayer library before make install:
```
philipsu522, line 61: is it possible to call `scripts/build.sh` from here? From an external user perspective, i'd still like to be able to just run
```
make install
```
and have my seid binary created directly.
@codchen @LCyson is this somethign we can check in? otherwise every time we run make install there's going to be a git diff. Or do we want to add `*.dylib` to `.gitignore`?

scripts/initialize_local.sh
```diff
@@ -1,7 +1,6 @@
 # min go compiler version >=1.18.2
 # gvm use go1.18.2
-# build seid
-go build -o build/seid ./cmd/seid/
+./build.sh seid-nitro
```
philipsu522, line 3: should probably change this to make install

scripts/old_initialize_local.sh
```diff
@@ -20,7 +20,7 @@ keyname=admin
 
 rm -rf ~/.sei
 echo "Building..."
-make install
+./build.sh sei-nitro
```
philipsu522, line 23: keep as make install

Makefile
LCyson: +1 is it possible to invoke the shell script to automatically build the binary
and think we already check in a .dylib here: https://github.com/sei-protocol/sei-chain/tree/master/x/nitro/replay, I think the best practice here is to update the .dylib artifact with sei release so that we don't need to worry the replay dylib version (if people download and compile the nitro-replayer from scratch)

LCyson: can we try copying the checked in .dylib to the correct path and see if `make install` works in this case, lmk hdyt of this @yzang2019 @philipsu522

PR #444, opened 2022-12-02
Refund gas fees if concurrent TX fails

BrandonWeng:
We already reset the cacheKv (by not writing it) if the transactions fail, but we also need to reset the in-mem values to as those are different from the store 

Previously we didn't clear the state so the deferred sends from the previous call carried over 

Unit tests

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1057,6 +1057,14 @@ func (app *App) ProcessTxConcurrent(
 	metrics.IncrTxProcessTypeCounter(metrics.CONCURRENT)
 }
 
+type ProcessBlockConcurrentFunction func(
+		ctx sdk.Context,
+		txs [][]byte,
+		completionSignalingMap map[int]acltypes.MessageCompletionSignalMapping,
+		blockingSignalsMap map[int]acltypes.MessageCompletionSignalMapping,
+		txMsgAccessOpMapping map[int]acltypes.MsgIndexToAccessOpMapping,
+) ([]*abci.ExecTxResult, bool)
+
 func (app *App) ProcessBlockConcurrent(
 	ctx sdk.Context,
 	txs [][]byte,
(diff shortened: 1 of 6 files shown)


app/app.go
```diff
@@ -1121,9 +1129,43 @@ func (app *App) ProcessBlockConcurrent(
...
+	if ok {
+		// Write the results back to the concurrent contexts - if concurrent execution fails,
+		// this should not be called and the state is rolled back and retried with synchronous execution
+		processBlockCache.Write()
+		return concurrentResults
+	}
+
+	ctx.Logger().Error("Concurrent Execution failed, retrying with Synchronous")
+	// Clear the memcache context from the previous state as it failed, we no longer need to commit the data
+	ctx = ctx.WithContextMemCache(sdk.NewContextMemCache())
```
udpatil: just to confirm, is this the step that we need in order to reset the in memory state prior to rerunning txs synchronously?

app/app.go
```diff
@@ -1121,9 +1129,43 @@ func (app *App) ProcessBlockConcurrent(
 	return txResults, ok
 }
 
+func (app *App) ProcessTxs(
+	ctx sdk.Context,
+	txs [][]byte,
+	dependencyDag *acltypes.Dag,
+	processBlockConcurrentFunction ProcessBlockConcurrentFunction,
```
udpatil, line 1136: why do we pass the the function instead of directly calling the `ProcessBlockConcurrent`? is it primarily for the mocking purposes in testing? if so, isn't that an antipattern? (I don't have strong feelings one way or the other on whether we keep it this way, just looking to better understand this & golang best practices.

app/app.go
BrandonWeng: Yes, this resets the state for the rerun of TXs synchronously

app/app.go
BrandonWeng: Yeah, this is mainly for testing, I was following this: https://www.myhatchpad.com/insight/mocking-techniques-for-go/

udpatil (approved):

app/app_test.go
```diff
@@ -76,3 +77,87 @@ func TestGetChannelsFromSignalMapping(t *testing.T) {
...
+	// Set some test context mem cache values
+	testWrapper.Ctx.ContextMemCache().UpsertDeferredSends("Some Account", sdk.NewCoins(sdk.Coin{
+		Denom:  "test",
+		Amount: sdk.NewInt(1),
+	}))
+	testWrapper.Ctx.ContextMemCache().UpsertDeferredWithdrawals("Some Other Account", sdk.NewCoins(sdk.Coin{
+		Denom:  "test",
+		Amount: sdk.NewInt(1),
+	}))
+	// It should be reset if it fails to prevent any values from being written
```
philipsu522: is this comment correct? i think the function succeeds?

philipsu522 (approved):

PR #447, opened 2022-12-05
[wasm] Fix wasm bad dependency identification

udpatil:
This fixes the wasm bad dependency identification to properly extract contract addresses that need their dependencies reset.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1130,17 +1130,17 @@ func (app *App) ProcessBlockConcurrent(
 }
 
 func (app *App) ProcessTxs(
-	ctx *sdk.Context,
+	ctx sdk.Context,
 	txs [][]byte,
 	dependencyDag *acltypes.Dag,
 	processBlockConcurrentFunction ProcessBlockConcurrentFunction,
-) []*abci.ExecTxResult {
+) ([]*abci.ExecTxResult, sdk.Context) {
 	// Only run concurrently if no error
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -266,7 +266,7 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.1.319
+	github.com/cosmos/cosmos-sdk => ../sei-cosmos //github.com/sei-protocol/sei-cosmos v0.1.319
```
udpatil: Reminder: replace this with latest sei-cosmos version

app/app.go
```diff
@@ -1130,7 +1130,7 @@ func (app *App) ProcessBlockConcurrent(
 }
 
 func (app *App) ProcessTxs(
-	ctx sdk.Context,
+	ctx *sdk.Context,
```
BrandonWeng: why does this need to be a pointer? ctx should be immutable

udpatil: because we modify the context with the list of bad wasm dependencies (previously this was done inline but now that this is in a helper function we need to persist the context changes back so that the context is updated for endblocker

udpatil: https://github.com/sei-protocol/sei-chain/pull/447/files#diff-0f1d2976054440336a576d47a44a37b80cdf6701dd9113012bce0e3c425819b7R1158

BrandonWeng (approved):

PR #455, opened 2022-12-16
Add nitro-replayer/target to gitignore

yzang2019:
We should ignore nitro-replayer/target/ just as what we did in the nitro-replayer repo


philipsu522 (approved):

.gitignore
```diff
@@ -33,3 +33,6 @@ loadtest/contracts/**/target/
 loadtest/contracts/**.json
 
 loadtest_results.txt
+
+# nitro-replayer
+nitro-replayer/target
```
udpatil, line 38: we should also include standard cargo gitignore items, such as `Cargo.lock`

yzang2019: Added one more ignore for Cargo.lock

PR #456, opened 2022-12-16
Fix local initialization script

yzang2019:
Problem:
Initialize_local.sh is broken on Mac because sed command is not working on latest MacOS:
```
sed -i 's/mode = "full"/mode = "validator"/g' $HOME/.sei/config/config.toml
sed: 1: "/Users/yzang/.sei/confi ...": unterminated transform source string
```

Fix:
sed is little bit different on Mac compared to Linux GNU, we need to change the command to this
`sed -i'.bak' -e 's/something/else'g' filepath`

Tested on Mac and Ubuntu

### DIFF
--- a/scripts/initialize_local.sh
+++ b/scripts/initialize_local.sh
@@ -1,5 +1,19 @@
 # min go compiler version >=1.18.2
 # gvm use go1.18.2
+
+#########################################################
+## How to submitting a proposal after chain is running ##
+#########################################################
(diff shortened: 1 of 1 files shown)


yzang2019:
@philipsu522 Updated with the change to remove password requirement, and added some instructions for how to submit a proposal. Please take another look, thanks!

philipsu522 (approved):

PR #457, opened 2022-12-19
Fix Begin/EndBlock stuck at infinite loop in contracts

codchen:
The `wasm` module would allow contracts to run forever if `ctx.GasMeter()` doesn't have a limit. So instead of using an infinite meter as the temporary meter (which was originally added to prevent race conditions between concurrent calls), we will use one that inherits the limit depending on whether it's called within BeginBlock or EndBlock (inferred from `ctx.GasMeter().Limit()`.

Tested on local chain with contract that has infinite loop in hooks

### DIFF
--- a/x/dex/keeper/utils/wasm.go
+++ b/x/dex/keeper/utils/wasm.go
@@ -40,9 +40,7 @@ func sudo(sdkCtx sdk.Context, k *keeper.Keeper, contractAddress []byte, wasmMsg 
 	defer metrics.MeasureSudoExecutionDuration(time.Now(), msgType)
 	// set up a tmp context to prevent race condition in reading gas consumed
 	tmpCtx := sdkCtx.WithGasMeter(sdk.NewGasMeter(sdkCtx.GasMeter().Limit()))
-	data, err := k.WasmKeeper.Sudo(
-		tmpCtx, contractAddress, wasmMsg,
-	)
+	data, err := sudoWithoutOutOfGasPanic(tmpCtx, k, contractAddress, wasmMsg)
 	gasConsumed := tmpCtx.GasMeter().GasConsumed()
 	if gasConsumed > 0 {
 		sdkCtx.GasMeter().ConsumeGas(gasConsumed, "sudo")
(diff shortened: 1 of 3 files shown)


x/dex/keeper/utils/wasm.go
```diff
@@ -39,12 +39,14 @@ func sudo(sdkCtx sdk.Context, k *keeper.Keeper, contractAddress []byte, wasmMsg
 	// Measure the time it takes to execute the contract in WASM
 	defer metrics.MeasureSudoExecutionDuration(time.Now(), msgType)
 	// set up a tmp context to prevent race condition in reading gas consumed
-	tmpCtx := sdkCtx.WithGasMeter(sdk.NewInfiniteGasMeter())
+	tmpCtx := sdkCtx.WithGasMeter(sdk.NewGasMeter(sdkCtx.GasMeter().Limit()))
```
udpatil, line 45: Where is this gas limit determined for begin / end block?

codchen: it's set at the beginning of Begin/EndBlock (and it merely infers the value here): https://github.com/sei-protocol/sei-chain/blob/master/x/dex/module.go#L234 and https://github.com/sei-protocol/sei-chain/blob/master/x/dex/contract/abci.go#L45

udpatil: This doesn't account for the already spent gas though, right? because lets say gas limit is 1000, but the beginblock already spent 250, and the wasm contract needs 800. Then with this current PR, it would allow a total consumption of 1050 gas, right? or is that fine since we're only looking to mitigate the infinite gas issue?

codchen: that's intentional. We don't want to take into account already spent gas because otherwise it may cause non-determinism - contract A may be processed and consume most gas in val 1 and contract B would fail because it only has little gas remaining whereas it could be opposite in val 2.

BrandonWeng: So IUCC in some cases (example Uday listed above) may perform more computation than the original gas limit but it will still fail later when we check the gas consumed? 

```
	gasConsumed := tmpCtx.GasMeter().GasConsumed()
	if gasConsumed > 0 {
		sdkCtx.GasMeter().ConsumeGas(gasConsumed, "sudo")
	}
```

If so, it might be useful to add a comment here for future reference.

udpatil: I don't think it would fail later because of greater than expected gas consumption, because the gas being consumed by contracts is on a new meter, so doesn't actually affect the original gas meter. I was just wondering whether that was ok considering that gas is generally tied to compute cost but this seems fine.

udpatil (approved):

x/dex/keeper/utils/wasm.go
codchen: yeah the gas limit is more like "no individual call can exceed it", but a soft limit overall. I'll add a comment for this

PR #459, opened 2022-12-20
Dragonberry patch part 2

BrandonWeng:
As per the README we also need to replace this package

https://github.com/cosmos/cosmos-sdk/releases/tag/v0.45.9 

image

Started local chain

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -266,7 +266,8 @@ require (
 )
 
 replace (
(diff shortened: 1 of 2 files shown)


Kbhat1 (approved):
LGTM, thanks. While we were testing dragonberry p1 I included this replace in Sei-chain. But I think we have a ticket in 2.0.0 testing to test this further

PR #467, opened 2022-12-27
Use contract rent to decide sudo call gas limit

codchen:
Set sudo gas limit based on the remaining rent of a contract instead of overall block gas limit so that the processing may time out sooner without affecting processing outcome. There would also be no race condition since there could be at most one sudo call happening at a time for the same contract.

unit test
local chain test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,5 @@ loadtest_results.txt
 # nitro-replayer
 nitro-replayer/target
 nitro-replayer/Cargo.lock
+# nitro mac artifact
+x/nitro/replay/libnitro_replayer.dylib
--- a/x/dex/keeper/contract.go
+++ b/x/dex/keeper/contract.go
@@ -50,14 +50,17 @@ func (k Keeper) GetContract(ctx sdk.Context, contractAddr string) (types.Contrac
 	return res, nil
 }
 
-func (k Keeper) GetContractGasLimit(ctx sdk.Context, contractAddr []byte) (uint64, error) {
-	bech32ContractAddr := sdk.AccAddress(contractAddr).String()
+func (k Keeper) GetContractGasLimit(ctx sdk.Context, contractAddr sdk.AccAddress) (uint64, error) {
+	bech32ContractAddr := contractAddr.String()
 	contract, err := k.GetContract(ctx, bech32ContractAddr)
 	if err != nil {
 		return 0, err
 	}
 	rentBalance := contract.RentBalance
 	gasPrice := k.GetParams(ctx).SudoCallGasPrice
+	if gasPrice.LTE(sdk.ZeroDec()) {
+		return 0, errors.New("invalid gas price: must be positive")
+	}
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -37,3 +37,5 @@ loadtest_results.txt
 # nitro-replayer
 nitro-replayer/target
 nitro-replayer/Cargo.lock
+# nitro mac artifact
+x/nitro/replay/libnitro_replayer.dylib
```
udpatil, line 41: if we're adding this to gitignore, whys the file shown as modified below? shouldn't we remove it?

x/dex/keeper/contract.go
```diff
@@ -50,6 +50,18 @@ func (k Keeper) GetContract(ctx sdk.Context, contractAddr string) (types.Contrac
 	return res, nil
 }
 
+func (k Keeper) GetContractGasLimit(ctx sdk.Context, contractAddr []byte) (uint64, error) {
```
udpatil: why is this `[]byte` instead of `AccAddress`? Then we can know its already validated and simply call `.String()` on it, right?

x/dex/keeper/contract.go
```diff
@@ -50,6 +50,18 @@ func (k Keeper) GetContract(ctx sdk.Context, contractAddr string) (types.Contrac
...
+	bech32ContractAddr := sdk.AccAddress(contractAddr).String()
+	contract, err := k.GetContract(ctx, bech32ContractAddr)
+	if err != nil {
+		return 0, err
+	}
+	rentBalance := contract.RentBalance
+	gasPrice := k.GetParams(ctx).SudoCallGasPrice
+	gasDec := sdk.NewDec(int64(rentBalance)).Quo(gasPrice)
```
udpatil, line 64: is it possible for `gasPrice` to be zero? If so we should have zero div protection here.

x/dex/keeper/utils/wasm.go
```diff
@@ -42,7 +42,11 @@ func sudo(sdkCtx sdk.Context, k *keeper.Keeper, contractAddress []byte, wasmMsg
 	// Note that the limit will effectively serve as a soft limit since it's
 	// possible for the actual computation to go above the specified limit, but
 	// the associated contract would be charged corresponding rent.
-	tmpCtx := sdkCtx.WithGasMeter(sdk.NewGasMeter(sdkCtx.GasMeter().Limit()))
+	gasLimit, err := k.GetContractGasLimit(sdkCtx, contractAddress)
```
udpatil, line 45: why are we passing contractAddress here (and above in `sudo` args as `[]byte` instead of `AccAddress` which is a defined type for `[]byte` IIRC?

codchen: updated

x/dex/keeper/contract.go
codchen: good point, updated

udpatil (approved):

PR #470, opened 2022-12-28
parallelization test contracts & fixes

codchen:
Add in contracts used for testing parallelization.

As made some fixes to dependencies in bank & dex message types

tested by
1. deploying the contracts to local chain
2. register dependencies according to `deps.json` under each contract's directory
3. add in logging before and after `DeliverTx`
4. send multiple transactions in the same block (e.g. with `async` broadcast type) and observe the logs. If there are overlaps between logs for different transactions, we know that the transactions are running in parallel.


BrandonWeng (dismissed):

aclmapping/bank/mappings.go
```diff
@@ -81,6 +81,17 @@ func MsgSendDependencyGenerator(keeper aclkeeper.Keeper, ctx sdk.Context, msg sd
 			IdentifierTemplate: hex.EncodeToString(authtypes.CreateAddressStoreKeyFromBech32(msgSend.FromAddress)),
 		},
 
+		{
+			AccessType:         sdkacltypes.AccessType_READ,
+			ResourceType:       sdkacltypes.ResourceType_KV_AUTH_GLOBAL_ACCOUNT_NUMBER,
+			IdentifierTemplate: hex.EncodeToString(authtypes.GlobalAccountNumberKey),
+		},
+		{
+			AccessType:         sdkacltypes.AccessType_WRITE,
```
udpatil: doesn't this mean that NO bank sends can be parallelized? Is there some check we can do before including this dependency so that we can retain some level of performance in the normal case for bank sends?

BrandonWeng (approved):

PR #475, opened 2023-01-04
[aclmapping] Modify staking dependencies for max vote power change

udpatil:
This makes the relevant changes to the staking dependency generators to maintain parity with the new resources used as part of the max voting power check.

Updated unit tests

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -19,7 +19,7 @@ def runtest_exists(dir):
     return os.path.exists(f"{dir}/runtest.sh")
 
 def run():
-    cpu_count = multiprocessing.cpu_count()
+    cpu_count = max(multiprocessing.cpu_count() - 2, 1)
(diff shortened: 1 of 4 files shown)


runtests.py
```diff
@@ -19,7 +19,7 @@ def runtest_exists(dir):
     return os.path.exists(f"{dir}/runtest.sh")
 
 def run():
-    cpu_count = multiprocessing.cpu_count()
+    cpu_count = max(multiprocessing.cpu_count() - 2, 1)
```
BrandonWeng, line 22: whats the context behind this?

BrandonWeng (approved):

runtests.py
udpatil: my laptop was basically useless while the tests ran because no cpu cores for anything else lol. This frees up 2 cores for some usability of other functionality (eg. vscode, browser, etc)

PR #476, opened 2023-01-04
Fix local docker script for M1 laptops

BrandonWeng:
Need to specify the platform for docker containers as it may try to use ARM64 for mac m1 users 

Verified it worked locally (M1)
image

Verified it works for @yzang2019 (Intel based)
image


yzang2019 (approved):
Thanks for fixing this!

PR #483, opened 2023-01-12
Replace bach32 with AccAddress in dex module

LCyson:
Current x/dex module uses plain bech32 string instead of sdk.AccAddress bytes as the KVStore key, which is inconsistent with other cosmos native modules. This issue causes the parallelization access operators unable to identify dex module KVStore resource correctly and blocks users from defining the correct dex resource dependency.

This PR adds the following changes:
- Changed all KVStore key in dex/types/keys.go from bech32 string to sdk.AccAddress bytes
- Updated all affected unit tests
- Updated the local Sei deployment script to work with the latest tendermint version
- Refactor dex/msgservers to use the `ValidateBasic` in the message_ files to gate bech32 incompatible addresses
- Fix an ordering non-agnostic bug found in the contract registration (see details in the code)

Unit test, e2e test

## Next Steps
Discussed with @philipsu522 and @codchen, we would like to change all bech32 string used in any sei-chain custom module to sdk.AccAddress to be consistent:
- Refactor all sei-chain custom module to use AccAddress https://linear.app/seilabs/issue/SEI-2565/refactor-all-bech32-to-sdkaccaddress-in-sei-chain-custom-modules
- Add a migration function to this change


x/dex/keeper/msgserver/msg_server_register_contract.go
```diff
@@ -242,6 +228,9 @@ func (k msgServer) SetNewContract(ctx sdk.Context, msg *types.MsgRegisterContrac
 		dependency.ImmediateYoungerSibling = ""
 		found := false
 		for _, contractInfo := range allContractInfo {
+			if contractInfo.ContractAddr == newContract.ContractAddr {
```
LCyson, line 231: this was not order agnostic previously @codchen and it updated the newContract to its dependency in a wrong order

x/dex/types/keys.go
```diff
@@ -27,7 +31,8 @@ func KeyPrefix(p string) []byte {
 }
 
 func ContractKeyPrefix(p string, contractAddr string) []byte {
-	return append([]byte(p), []byte(contractAddr)...)
+	address, _ := sdk.AccAddressFromBech32(contractAddr)
+	return append([]byte(p), address.Bytes()...)
```
codchen: cosmos SDK uses `address.MustLengthPrefix` to pad the bytes so we probably want to do the same too https://github.com/sei-protocol/sei-cosmos/blob/main/x/bank/types/key.go#L58

LCyson: think only bank module uses that, other modules use plain accAddress bytes: https://github.com/sei-protocol/sei-cosmos/blob/main/x/auth/types/keys.go#L30-L37

actually err: other modules use `MustLengthPrefix`, only the auth module uses the plain accAddress as the prefix, I'll change this to MustLengthPrefix in this case

x/dex/keeper/msgserver/msg_server_unregister_contract.go
```diff
@@ -20,10 +26,7 @@ func (k msgServer) UnregisterContract(goCtx context.Context, msg *types.MsgUnreg
 		return nil, sdkerrors.ErrUnauthorized
 	}
 	// refund remaining rent to the creator
-	creatorAddr, err := sdk.AccAddressFromBech32(contract.Creator)
-	if err != nil {
-		return nil, err
-	}
+	creatorAddr, _ := sdk.AccAddressFromBech32(contract.Creator)
```
philipsu522, line 29: should we still handle err?

philipsu522 (approved):
lgtm, thanks for this! Do you mind updating the title (i think there are typos)

codchen (approved):

x/dex/keeper/msgserver/msg_server_unregister_contract.go
LCyson: no we can assume the creatorAddr must be a valid bech32 address (as all msgserver has validated any request's address). This is also aligned with how other cosmos native modules convert bech32 address

PR #484, opened 2023-01-12
Update consensus configs to be less strict

BrandonWeng:
Sei-devnet-2 seeing a lot of validators miss blocks, turn off skip `UnsafeBypassCommitTimeoutOverride` and increase commit timeout to 100ms 

Don't need to change TM as this directly overrides whatever is from the config.toml file

local chain with timeout set to 5s, can see that each block height was very slow

### DIFF
--- a/app/params/config.go
+++ b/app/params/config.go
@@ -21,8 +21,8 @@ const (
 	Bech32PrefixAccAddr = "sei"
 )
 
-// Commits block as soon as we reach consensus instead of waiting for timeout, this may
-// cause validators to not get their votes in time
+// UnsafeBypassCommitTimeoutOverride commits block as soon as we reach consensus instead of waiting
+// for timeout, this may cause validators to not get their votes in time
 var UnsafeBypassCommitTimeoutOverride = false
 
 var (
--- a/app/params/config.go
+++ b/app/params/config.go
@@ -93,11 +93,11 @@ func SetTendermintConfigs(config *tmcfg.Config) {
 	config.Mempool.MaxTxBytes = 2048576
 	// Consensus Configs
 	config.Consensus.GossipTransactionKeyOnly = true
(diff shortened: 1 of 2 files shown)


app/params/config.go
```diff
@@ -93,11 +95,9 @@ func SetTendermintConfigs(config *tmcfg.Config) {
 	config.Consensus.GossipTransactionKeyOnly = true
 	config.Consensus.UnsafeProposeTimeoutOverride = 1 * time.Second
 	config.Consensus.UnsafeProposeTimeoutDeltaOverride = 500 * time.Millisecond
-
 	config.Consensus.UnsafeVoteTimeoutOverride = 50 * time.Millisecond
 	config.Consensus.UnsafeVoteTimeoutDeltaOverride = 500 * time.Millisecond
-	config.Consensus.UnsafeCommitTimeoutOverride = 50 * time.Millisecond
-
+	config.Consensus.UnsafeCommitTimeoutOverride = 100 * time.Millisecond
```
philipsu522: can we keep this as 50ms? we can bump up if needed. This will add about 50ms to consensus if we make this change, and this is on top of waiting for 2/3 votes already

app/params/config.go
```diff
@@ -93,11 +95,9 @@ func SetTendermintConfigs(config *tmcfg.Config) {
 	config.Consensus.GossipTransactionKeyOnly = true
 	config.Consensus.UnsafeProposeTimeoutOverride = 1 * time.Second
```
philipsu522: can we change this to 5 seconds? i have a theory that a lower value leads to a lot more rounds (more prevotes of nil) which makes consensus slow when we barely have 67%

BrandonWeng: yessir

philipsu522 (approved):

PR #486, opened 2023-01-13
[ACL] integrate with inter-contract parallelization changes

codchen:
- Integrate with new inter-contract parallelization sei-cosmos change
- Conform with the new function signature in dependency generator
- Evaluate referenced contract's dependency identifier template to decide if gas discount should be granted. Gas discount is granted only if all referred messages' identifiers are non-wildcard. Note that if there are cycles in the reference graph, gas discount will also not be granted.

unit tests

### DIFF
--- a/app/antedecorators/gas.go
+++ b/app/antedecorators/gas.go
@@ -1,8 +1,6 @@
 package antedecorators
 
 import (
-	"fmt"
-
 	wasmtypes "github.com/CosmWasm/wasmd/x/wasm/types"
 	"github.com/cosmos/cosmos-sdk/store/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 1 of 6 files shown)


app/antedecorators/gas.go
```diff
@@ -32,22 +35,112 @@ func GetGasMeterSetter(aclkeeper aclkeeper.Keeper) func(bool, sdk.Context, uint6
 }
 
 func getMessageMultiplierDenominator(ctx sdk.Context, msg sdk.Msg, aclKeeper aclkeeper.Keeper) uint64 {
+	// TODO: reason through whether it's reasonable to require non-* identifier for all operations
```
udpatil, line 36: one approach we can do is to get the dependencies, and then have a helper function to detect whether a synchronous access op exists (eg. resource ALL with identifier *). If that isn't present AND the contract has a registered dependency, we could give a gas discount?

app/antedecorators/gas.go
```diff
@@ -32,22 +35,112 @@ func GetGasMeterSetter(aclkeeper aclkeeper.Keeper) func(bool, sdk.Context, uint6
...
+		// no need to raise an error here. Simply returning false since a contract-message
+		// with cycle will not be parallelized anyway
+		return false
+	}
+	cycleDetector[cycleIdentifier] = struct{}{}
+	addr, err := sdk.AccAddressFromBech32(contractAddrStr)
+	if err != nil {
+		return false
+	}
+	mapping, err := aclKeeper.GetRawWasmDependencyMapping(ctx, addr)
```
udpatil: why not simply call GetWasmDependencyMapping instead of getting the raw one and having branching logic for contract references? That way it would traverse the references for us using the existing logic and wouldnt require repeating logic here, right?

app/antedecorators/gas.go
codchen: Can you elaborate the diff between what you described and the current impl? also I think we want to discourage using * not just for type ALL but also for resource type like bank since that would make it blocking for most messages

app/antedecorators/gas.go
codchen: The logic here would return early as soon as it hits a *, but agreed that with the new reference logic the repetition is probably too much to be worthwhile the gain. Will update

app/antedecorators/gas.go
udpatil: Yeah but having * for bank balances for example still allows things like staking messages to be unblocked. If the gas discount is the main incentive for parallelization, there should be SOME incentive to parallelize even if your contract needs to iterate over some addresses or something resulting in a * on a KV prefix otherwise people wont do anything aka blocking EVERYTHING, right?

udpatil (approved):

PR #489, opened 2023-01-18
[price-feeder] Update grpc so insecure verify works properly

udpatil:
This fixes grpc client according to discussion around insecure skip verify

### DIFF
--- a/oracle/price-feeder/oracle/oracle.go
+++ b/oracle/price-feeder/oracle/oracle.go
@@ -2,6 +2,7 @@ package oracle
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 	"math"
 	"net/http"
@@ -12,6 +13,7 @@ import (
 	"github.com/rs/zerolog"
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
 
 	"github.com/sei-protocol/sei-chain/oracle/price-feeder/config"
 	"github.com/sei-protocol/sei-chain/oracle/price-feeder/oracle/client"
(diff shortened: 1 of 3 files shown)


oracle/price-feeder/oracle/oracle.go
```diff
@@ -391,8 +393,8 @@ func (o *Oracle) GetParamCache(ctx context.Context, currentBlockHeight int64) (o
 func (o *Oracle) GetParams(ctx context.Context) (oracletypes.Params, error) {
 	grpcConn, err := grpc.Dial(
 		o.oracleClient.GRPCEndpoint,
-		// the Cosmos SDK doesn't support any transport security mechanism
-		grpc.WithInsecure(),
+		grpc.WithTransportCredentials(credentials.NewTLS(&tls.Config{InsecureSkipVerify: true})),
+		grpc.WithBlock(),
```
BrandonWeng, line 397: Could you give context around the discussion in this PR so future readers can reference it?

philipsu522: does it make sense to block?

oracle/price-feeder/oracle/oracle.go
```diff
@@ -391,8 +393,8 @@ func (o *Oracle) GetParamCache(ctx context.Context, currentBlockHeight int64) (o
 func (o *Oracle) GetParams(ctx context.Context) (oracletypes.Params, error) {
 	grpcConn, err := grpc.Dial(
 		o.oracleClient.GRPCEndpoint,
-		// the Cosmos SDK doesn't support any transport security mechanism
-		grpc.WithInsecure(),
+		grpc.WithTransportCredentials(credentials.NewTLS(&tls.Config{InsecureSkipVerify: true})),
```
philipsu522, line 396: could we apply Kartik's patch https://discord.com/channels/973057323805311026/1062774084917796904/1065301857901223966 here? It might be hard though, since it'll require val nodes to have acecss to our ACM. Do you know if other price feeders are running TLS (i.e. Kujira's)?

Kbhat1: Yea here is the PR which added the patch https://github.com/sei-protocol/dex-indexer/pull/52 . But in that case, we have direct access to our ACM from Indexer. Not sure it makes sense here

oracle/price-feeder/oracle/oracle.go
Kbhat1: Block shouldn't be necessary

PR #491, opened 2023-01-19
Migrate old dex prefix data

LCyson:
This PR adds a migration handler for the dex prefixes change here: https://github.com/sei-protocol/sei-chain/pull/483. To be noted that here we will simply nuke the old data as this will only be needed on the devnet-2.

Other option is to replace the existing data like the snippet below, but given there are keys with suffix, either getting all pairs from contracts or assuming the contract address length is 62 in bytes is not as safe as nuking the old data on the devnet-2.
```

	for _, prefixKey := range DexPrefixes {
		store := prefix.NewStore(ctx.KVStore(dexkeeper.GetStoreKey()), []byte(prefixKey))
		iterator := sdk.KVStorePrefixIterator(store, []byte{})

		defer iterator.Close()
		for ; iterator.Valid(); iterator.Next() {
			key := iterator.Key()
			if !bytes.HasPrefix(key, []byte(prefixKey)) {
				panic(fmt.Sprintf("Key %x doesn't has prefix %s\n", key, prefixKey))
			}
			key = bytes.TrimPrefix(key, []byte(prefixKey))

			key = append([]byte(prefixKey), types.AddressKeyPrefix(string(key))...)

			store.Set(key, iterator.Value())
			store.Delete(iterator.Key())
		}
	}

```

- Unit test
- Upgraded the chain locally with the following steps
  - deployed vortex contract to V10
  - placed an order
  - upgraded to V11
  - confirmed the order is nuked and no panic/exception was thrown

### DIFF
--- a/x/dex/migrations/v10_to_v11.go
+++ b/x/dex/migrations/v10_to_v11.go
@@ -1,0 +1,45 @@
+package migrations
+
+import (
+	"github.com/cosmos/cosmos-sdk/store/prefix"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/sei-protocol/sei-chain/x/dex/keeper"
+	"github.com/sei-protocol/sei-chain/x/dex/types"
+)
+
+var DexPrefixes = []string{
+	types.LongBookKey,
+	types.ShortBookKey,
+	types.TriggerBookKey,
+	types.OrderKey,
+	types.TwapKey,
+	types.RegisteredPairKey,
+	types.OrderKey,
+	types.CancelKey,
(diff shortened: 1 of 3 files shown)


philipsu522 (approved):
lgtm, but let's get @codchen 's reviews as well.
in your local test, was vortex contract deployed with data on it (long book, orders...etc.)

LCyson:
> lgtm, but let's get @codchen 's reviews as well. in your local test, was vortex contract deployed with data on it (long book, orders...etc.)

Yes, tested by placing a LONG order and verified that order and also the trading pairs are nuked

PR #495, opened 2023-01-20
Refactor mintting logic

BrandonWeng:
Mint denom should just be usei as that's the base currency that the staking rewards will be paid out as, it shouldn't be any other denoms 

But we'll keep it in genesis to avoid confusion around the denom of the fields 

See inline comments for more detailed description 

unit tests 

image


x/mint/types/params.go
```diff
@@ -73,15 +73,19 @@ func (p *Params) ParamSetPairs() paramtypes.ParamSetPairs {
...
 	if !ok {
 		return fmt.Errorf("invalid parameter type: %T", i)
 	}
-
-	if strings.TrimSpace(v) == "" {
+	if denomTrimed == "" {
 		return errors.New("mint denom cannot be blank")
 	}
-	if err := sdk.ValidateDenom(v); err != nil {
+	if denomTrimed != "usei" {
```
philipsu522: nit - is there a variable we can use that specifies base denom? maybe
```
x/dex/types/keys.go:155:        DefaultPriceDenom = "usei"
```

BrandonWeng: gotcha, using the default bond denom

proto/mint/v1beta1/mint.proto
```diff
@@ -5,14 +5,22 @@ option go_package = "github.com/sei-protocol/sei-chain/x/mint/types";
 
 import "gogoproto/gogo.proto";
 
-// Minter represents the minting state.
+// Minter represents the most recent
 message Minter {
-  // current epoch provisions
-  string epoch_provisions = 1 [
```
BrandonWeng, line 11: not used at all, going to change minter to track the state of a mint instead

proto/mint/v1beta1/mint.proto
```diff
@@ -26,14 +34,8 @@ message Params {
 
   // type of coin to mint
   string mint_denom = 1;
-  // epoch provisions from the first epoch
-  string genesis_epoch_provisions = 2 [
```
BrandonWeng, line 30: not used at all

x/mint/client/cli/query.go
```diff
@@ -62,8 +61,8 @@ func GetCmdQueryParams() *cobra.Command {
 // epoch provisions value.
 func GetCmdQueryEpochProvisions() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "epoch-provisions",
-		Short: "Query the current minting epoch provisions value",
+		Use:   "minter",
```
BrandonWeng, line 64: query the last minter saved, i.e the last mint event

philipsu522 (approved):

PR #497, opened 2023-01-22
[OraclePriceFeeder] Improve feeder performance by making tx broadcast async

yzang2019:
**Problem:**
Currently, the oracle price feeder is not able to keep up with the block chain speed, it is only able to broadcast the transactions for every a few blocks. Ideally, we want the transaction to be broadcasted for every block, not to skippping any blocks.

**Solution:**
This PR propose the change to make the tick asynchronous and best effort. 

The previous behavior is that, we have a separate thread keep updating the height, and on block height changes, we will send a new tx broadcast. Whenever each time if there are some broadcast failures, we will retry the same tx for the next few blocks up till some max retry blocks  and will sleep for 1 second between each retry. This 1 second sleep takes too much time and doesn't make lot of sense any more. 

The new behavior will use a blockHeight channel to get new block updates, and for each new block, it will spin up a new goroutine to call tick() asynchronously. Within each ticket, we will get the latest prices and send the tx broadcast request without doing any retry. Retry is not needed here because as long as we are doing broadcast for every height, the future requests will succeed anyway.

Tested on sei-devenet-2

### DIFF
--- a/oracle/price-feeder/oracle/oracle.go
+++ b/oracle/price-feeder/oracle/oracle.go
@@ -122,24 +122,24 @@ func (o *Oracle) Start(ctx context.Context) error {
 
 			// Track how many pending goroutines are there
 			telemetry.IncrCounter(1, "num_pending_blocks", "tick")
-			go func() {
+			go func(blockHeight int64) {
 				startTime := time.Now()
 				defer telemetry.MeasureSince(startTime, "runtime", "tick")
 				defer telemetry.IncrCounter(1, "new", "tick")
 				defer telemetry.IncrCounter(-1, "num_pending_blocks", "tick")
 
-				if err := o.tick(ctx, clientCtx, txFactory, currBlockHeight); err != nil {
+				if err := o.tick(ctx, clientCtx, txFactory, blockHeight); err != nil {
 					telemetry.IncrCounter(1, "failure", "tick")
-					o.logger.Err(err).Msg(fmt.Sprintf("oracle tick failed for height %d", currBlockHeight))
+					o.logger.Err(err).Msg(fmt.Sprintf("oracle tick failed for height %d", blockHeight))
 				}
+			}(currBlockHeight)
 
-				// Catch any missing blocks (should never happen)
-				if currBlockHeight > (lastProcessedBlock+1) && lastProcessedBlock > 0 {
-					missedBlocks := currBlockHeight - (lastProcessedBlock + 1)
-					telemetry.IncrCounter(float32(missedBlocks), "num_missed_blocks", "tick")
-				}
-				lastProcessedBlock = currBlockHeight
-			}()
+			// Catch any missing blocks (should never happen)
+			if currBlockHeight > (lastProcessedBlock+1) && lastProcessedBlock > 0 {
+				missedBlocks := currBlockHeight - (lastProcessedBlock + 1)
+				telemetry.IncrCounter(float32(missedBlocks), "num_missed_blocks", "tick")
+			}
+			lastProcessedBlock = currBlockHeight
 
 		}
 	}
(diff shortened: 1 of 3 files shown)


oracle/price-feeder/oracle/oracle.go
```diff
@@ -116,19 +119,24 @@ func (o *Oracle) Start(ctx context.Context) error {
 		default:
 			o.logger.Debug().Msg("starting oracle tick")
 
-			startTime := time.Now()
+			// Wait for next block height to be available in the channel
+			nextBlockHeight := <-o.oracleClient.BlockHeightEvents
```
udpatil: good idea, helps us not need the arbitrary wait

oracle/price-feeder/oracle/oracle.go
```diff
@@ -116,19 +119,24 @@ func (o *Oracle) Start(ctx context.Context) error {
...
+				}
 
-			o.lastPriceSyncTS = time.Now()
+				o.lastPriceSyncTS = time.Now()
 
-			telemetry.MeasureSince(startTime, "runtime", "tick")
-			telemetry.IncrCounter(1, "new", "tick")
+				telemetry.MeasureSince(startTime, "runtime", "tick")
+				telemetry.IncrCounter(1, "new", "tick")
+
```
udpatil: can we also add metric for tick for chain height? then we can visualize missed ones  / eventualy alert on them better too

oracle/price-feeder/oracle/oracle.go
udpatil: what happens if we're processing height 100 and for some reason it takes too long such that we have height 101 and 102 in the channel? then the next height processed would be the old one, right? shouldn't we take EVERYTHING from the channel (if theres more than 1) and then process for latest block height?

udpatil: nit, shouldnt this be `currBlockHeight` not next block height?

oracle/price-feeder/oracle/oracle.go
yzang2019: what do you mean by missed ones?

oracle/price-feeder/oracle/oracle.go
yzang2019: Right now each blocks is going to be processed in a separate thread, each broadcast takes 20ms-30ms to complete, so there's no possibility of being blocked at all. In the worse case where each broadcast take a few seconds to complete, the side effect is that we will end up creating more goroutines, where the num of goroutines= num of pending blocks within last few seconds

yzang2019: will fix

oracle/price-feeder/oracle/oracle.go
yzang2019: Do you mean catch the event subscribe function missing to push some block height to the channel?

oracle/price-feeder/oracle/oracle.go
yzang2019: It might be worth to add a metric to monitor the total number of processing goroutines though, I'll add one for that

oracle/price-feeder/oracle/oracle.go
udpatil: i mean if the tick doesn't happen for a specific height because we had a newer height

udpatil: to go along with the comment for not processing older tick heights if a newer one is present

oracle/price-feeder/oracle/oracle.go
udpatil: this sounds good, I was just curious if there was some small chance of a backlog in the channel, but i think its reasonable that it would be processing fast enough

oracle/price-feeder/oracle/oracle.go
udpatil: I think this isnt actionable right now considering that theres no real expectation of backlog in the height channel

udpatil (approved):

PR #503, opened 2023-01-24
Add systemd and feeder instructions to price feeder readme

udpatil:
Added steps outlining systemd service setup and feeder setup.

### DIFF
--- a/oracle/price-feeder/README.md
+++ b/oracle/price-feeder/README.md
@@ -121,3 +121,31 @@ Ex :
 `export PRICE_FEEDER_PASS=keyringPassword`
 
 If this environment variable is not set, the price feeder will prompt the user for input.
+
+## Systemd Configuration
+
+In order to run the price feeder as a background process, you can set up a systemd service for it. Here is an example of the service that will run the price feeder process:
+
+```ini
+[Unit]
+Description=Oracle Price Feeder
+After=network.target
+
+[Service]
+User=root
+Type=simple
+Environment="PRICE_FEEDER_PASS={KEYRING_PASSWORD}"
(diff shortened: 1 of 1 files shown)


yzang2019 (approved):

oracle/price-feeder/README.md
```diff
@@ -121,3 +121,31 @@ Ex :
...
+```ini
+[Unit]
+Description=Oracle Price Feeder
+After=network.target
+
+[Service]
+User=root
+Type=simple
+Environment="PRICE_FEEDER_PASS={KEYRING_PASSWORD}"
+ExecStart=/root/go/bin/price-feeder {PATH-TO-CONFIG-TOML}
```
BrandonWeng, line 138: how do we get the config-toml?

oracle/price-feeder/README.md
```diff
@@ -121,3 +121,31 @@ Ex :
...
+
+[Install]
+WantedBy=multi-user.target
+```
+
+## Feeder Delegation
+
+It is suggested to run the oracle price feeder with an account that is configured as a feeder as opposed to the validator account. Here is the command to set up the feeder:
+```bash
+seid tx oracle set-feeder {feeder-bech32} --from {validator-account} --fees 2000usei -b block -y --chain-id {chain-id}
```
BrandonWeng, line 150: how do we get the feeder-bech32?

oracle/price-feeder/README.md
BrandonWeng: i think we should also add instructions for building/running the price-feeder? I'm looking at the sei-infra instructions and it seems more involved than just setting up the systemd process 

`cd /home/ubuntu/sei-chain/; make install-price-feeder`

PR #505, opened 2023-01-24
[Oracle Price Feeder] Improve Success Rate by managing account sequence locally

yzang2019:
Problem:
Currently even if we query the latest account sequence number just before we send out the broadcastTx request, we would still got rejected because of sequence mismatch. That means the sequence number we get from the query is reflecting a stale number, not the number is being expected.

Fix:
Sequence number is expected to increase by 1 for every new transaction. To match with this expectation, we can manually maintain the sequence number in local memory and increase it by one for each new request sending to the chain. 

However, the tricky part is how to handle failures. In some scenarios where even if we have the correct sequence number, but the TX still failed for some other reason, we will increase the sequence locally, but the chain will not increase the sequence number, then we will have a mismatch again, and going on, all future TX will fail.

Here the proposal is to manually increase the sequence number after each broadcastTx returns a response, and when we hit any error, we will reset our local sequence number to match with the remote side, so that we can re-align with each other


oracle/price-feeder/oracle/client/client.go
```diff
@@ -143,11 +143,18 @@ func (r *passReader) Read(p []byte) (n int, err error) {
...
 
-	resp, err := BroadcastTx(clientCtx, factory, msgs...)
+	clientCtx, err := oc.CreateClientContext()
+	if err != nil {
+		return err
+	}
+	txFactory, err := oc.CreateTxFactory()
+	if err != nil {
+		return err
+	}
```
philipsu522, line 155: this seems to be run every block. could we initialize these further up (at least oc.CreateClientContext()) so that we don't run it every time?

oracle/price-feeder/oracle/oracle.go
```diff
@@ -121,23 +111,20 @@ func (o *Oracle) Start(ctx context.Context) error {
 			currBlockHeight := <-o.oracleClient.BlockHeightEvents
 
 			// Track how many pending goroutines are there
```
philipsu522: is this comment still correct?

philipsu522 (approved):
had a few minor qs, but LGTM

oracle/price-feeder/oracle/client/client.go
yzang2019: I tried to reuse the same txfactory, but it seems to be causing issues where the sequence number doesn't really get updated

oracle/price-feeder/oracle/oracle.go
yzang2019: Good catch, it's not needed since we switch to single threaded mode

PR #510, opened 2023-01-27
Configurable loadtest grpc endpoint + wasm message

Kbhat1:
- Adds a new "continuous" load type that continuously sends load by setting the steady loadtest interval to 0. NOTE: Added to-dos to created a `runContinuously` to not close client each time. Ran into acc seq issues as described in the comment and will resolve
- Allows pointing loadtest to webapp elb
- Adds a wasm message minting message to loadtest

- Tested on loadtest cluster sei-devnet-pre-internal


loadtest/main.go
```diff
@@ -100,6 +101,42 @@ func runOnce(config Config) {
 	go client.ValidateTxs()
 }
 
+// TODO: runContinuously continously sends transactions without closing and reopening channel
+// Ran into seq acc mismatch issues
+// func runContinously(config Config) {
```
Kbhat1: `runContinuously` would send tx's without closing the channel. Included a `prevSeqDelta` that would count the number of rounds accounts were involved in so that we can increment seq number properly. However still ran into acc seq issues particularly the expected acc seq number was 1 larger than the actual acc seq number. Will investigate as a P1 and add. For now, we resort to just recreating the client

BrandonWeng: nit can we remove this? its not the cleanest to have a large chunk of commented out code, might be better to document this in the ticket or somewhere else

loadtest/scripts/run_tests.py
```diff
@@ -88,9 +105,9 @@ def run():
                         description = 'Wrapper for the golang client to run loadtests with different configs')
     parser.add_argument(
         'type',
-        help='Type of loadtest to run (e.g constant, burst)',
+        help='Type of loadtest to run (e.g constant, burst, breaking)',
```
BrandonWeng: What's "breaking" here?

BrandonWeng (commented):
just some nits

loadtest/main.go
Kbhat1: Removed

loadtest/scripts/run_tests.py
Kbhat1: Goal here was to "break" the node by sending it enough traffic that it falls behind / is unresponsive. We could rename this as "constant" but its confusing with having the steady (which still takes breaks).

LCyson (approved):

LCyson:
are we going to resolve the acc seq in a different PR (i suppose by creating more accounts)?

LCyson:
> are we going to resolve the acc seq in a different PR (i suppose by creating more accounts)?

^saw the account creation in the other PR

loadtest/scripts/run_tests.py
Kbhat1: As discussed offline, renamed to "continuous"

BrandonWeng (approved):

Kbhat1:
NOTE: This hasn't been merged - need to clean up and refactor some of the reused components + better surface acc seq errors.

yzang2019 (approved):

PR #515, opened 2023-01-30
[cleanup] Remove Terra and Luna references

BrandonWeng:
Did a mass replace for `Terra,terra,TERRA` and `Luna,luna,LUNA` for `Sei,sei,SEI` respectively 

Unit tests

### DIFF
--- a/oracle/.env
+++ b/oracle/.env
@@ -4,8 +4,8 @@ CHAIN_ID="sei-chain"
 # Cron tab for how frequently prices will be posted (ex: 10 seconds)
 CRONTAB="*/10 * * * * *"
 
-# List of markets the oracle will post prices for, against ust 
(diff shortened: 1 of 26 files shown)


oracle/.env
```diff
@@ -4,8 +4,8 @@ CHAIN_ID="sei-chain"
 # Cron tab for how frequently prices will be posted (ex: 10 seconds)
 CRONTAB="*/10 * * * * *"
 
-# List of markets the oracle will post prices for, against ust 
-MARKET_IDS="btc,luna,btc,eth,sol,osmo"
```
BrandonWeng, line 8: can probably leave here if we want to support luna?

codchen: yea i think either way is fine since this is just a var

codchen (approved):

PR #524, opened 2023-02-01
Bump max packet msg payload

BrandonWeng:
based on sei-tendermint, max packet mesg should just be used in the channel when reading msgs and errors out if the packet size is too long so this should be fine to bump up 

https://github.com/sei-protocol/sei-tendermint/blob/46d0a598a7f5c67cbdefea37c8da18df2c25d184/internal/p2p/conn/connection_test.go#L503-L544 

image

### DIFF
--- a/loadtest/contracts/deploy_ten_contracts.sh
+++ b/loadtest/contracts/deploy_ten_contracts.sh
@@ -66,8 +66,6 @@ marsaddr4=$(python3 parser.py contract_address $marsinsres4)
 # register
 echo "Registering..."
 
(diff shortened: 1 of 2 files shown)


philipsu522 (approved):

loadtest/contracts/deploy_ten_contracts.sh
```diff
@@ -67,7 +67,6 @@ marsaddr4=$(python3 parser.py contract_address $marsinsres4)
 echo "Registering..."
 
 valaddr=$(printf "12345678\n" | $seidbin keys show $(printf "12345678\n" | $seidbin keys show node_admin --output json | jq -r .address) --bech=val --output json | jq -r '.address')
```
udpatil: I believe we can remove this line too

udpatil (approved):

PR #533, opened 2023-02-03
promote dex rent log level

codchen:
Since we are seeing non-determinism with rent consumption, we need more visible log for rent gas consumed for future clusters.

n/a

### DIFF
--- a/x/dex/keeper/utils/wasm.go
+++ b/x/dex/keeper/utils/wasm.go
@@ -57,6 +57,7 @@ func sudo(sdkCtx sdk.Context, k *keeper.Keeper, contractAddress sdk.AccAddress, 
 			sdk.NewAttribute("consumed", fmt.Sprintf("%d", gasConsumed)),
 			sdk.NewAttribute("type", msgType),
 			sdk.NewAttribute("contract", contractAddress.String()),
+			sdk.NewAttribute("height", fmt.Sprintf("%d", sdkCtx.BlockHeight())),
 		),
 	)
 	if gasConsumed > 0 {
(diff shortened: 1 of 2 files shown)


philipsu522 (commented):
@codchen this ended up being super noisy for all nodes - wdyt of keeping this as debug and rolling back + runnign with debug logs when there are issues?

codchen:
> @codchen this ended up being super noisy for all nodes - wdyt of keeping this as debug and rolling back + runnign with debug logs when there are issues?

That's what I did today but it's so rare that rollbacks can't really reproduce it, so we probably want something that can catch it in action. I'm gonna change this to events which would at least keep logs clean

x/dex/keeper/utils/wasm.go
```diff
@@ -50,7 +51,14 @@ func sudo(sdkCtx sdk.Context, k *keeper.Keeper, contractAddress sdk.AccAddress,
 	tmpCtx := sdkCtx.WithGasMeter(sdk.NewGasMeter(gasLimit))
 	data, err := sudoWithoutOutOfGasPanic(tmpCtx, k, contractAddress, wasmMsg, msgType)
 	gasConsumed := tmpCtx.GasMeter().GasConsumed()
-	sdkCtx.Logger().Debug(fmt.Sprintf("%s %s consumed %d gas", contractAddress.String(), msgType, gasConsumed))
+	sdkCtx.EventManager().EmitEvent(
+		sdk.NewEvent(
```
BrandonWeng, line 55: should we add height here?

codchen: updated

BrandonWeng (approved):

PR #536, opened 2023-02-03
[go.mod] use new golang-lru version for generics support 

BrandonWeng:
https://github.com/sei-protocol/sei-cosmos/pull/154

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -33,7 +33,7 @@ jobs:
         run: |
           until [ $(cat build/generated/launch.complete |wc -l) = 4 ]
           do
-            sleep 10
+            sleep 60
           done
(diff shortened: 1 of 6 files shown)


philipsu522 (approved):

.github/workflows/integration-test.yml
```diff
@@ -33,9 +33,9 @@ jobs:
         run: |
           until [ $(cat build/generated/launch.complete |wc -l) = 4 ]
           do
-            sleep 10
+            sleep 60
```
yzang2019: why do we bump to 60 seconds? Isn't that gonna slow down the test?

BrandonWeng: not going to commit this, just seeing if it eventually builds

PR #546, opened 2023-02-07
[price-feeder] Add handling for jaield validator in oracle tick

udpatil:
This adds a check for whether the validator is jailed, and if so, doesn't allow submission of oracle votes, and instead logs error output

Verified with personal cluster

### DIFF
--- a/oracle/price-feeder/oracle/jail.go
+++ b/oracle/price-feeder/oracle/jail.go
@@ -1,7 +1,16 @@
 package oracle
 
+import (
+	"context"
+	"fmt"
+	"time"
+
+	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+	"google.golang.org/grpc"
+)
+
 const (
-	jailCacheInterval = int64(50)
+	jailCacheIntervalBlocks = int64(50)
 )
 
 type JailCache struct {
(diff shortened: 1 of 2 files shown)


oracle/price-feeder/oracle/jail.go
```diff
@@ -0,0 +1,31 @@
...
+		return false
+	}
+
+	// This is an edge case, which should never happen.
+	// The current blockchain height is lower
+	// than the last updated block, to fix we should
+	// just update the cached params again.
+	if currentBlockHeight < jailCache.lastUpdatedBlock {
+		return true
+	}
```
BrandonWeng: Hmm, if it should never happen then should this be a panic or at least log an error?

oracle/price-feeder/oracle/oracle.go
```diff
@@ -417,6 +420,46 @@ func (o *Oracle) GetParams(ctx context.Context) (oracletypes.Params, error) {
...
+	if err != nil {
+		return false, err
+	}
+
+	o.jailCache.Update(currentBlockHeight, isJailed)
+	return isJailed, nil
+}
+
+// GetJailedState returns the current on-chain jailing state of the validator
+func (o *Oracle) GetJailedState(ctx context.Context) (bool, error) {
```
yzang2019: Can we move this to jail.go?

oracle/price-feeder/oracle/oracle.go
```diff
@@ -417,6 +420,46 @@ func (o *Oracle) GetParams(ctx context.Context) (oracletypes.Params, error) {
 	return queryResponse.Params, nil
 }
 
+func (o *Oracle) GetJailedStateCache(ctx context.Context, currentBlockHeight int64) (bool, error) {
```
yzang2019: Let's move this to jail.go?

oracle/price-feeder/oracle/oracle.go
```diff
@@ -417,6 +420,46 @@ func (o *Oracle) GetParams(ctx context.Context) (oracletypes.Params, error) {
...
+		o.oracleClient.GRPCEndpoint,
+		// the Cosmos SDK doesn't support any transport security mechanism
+		grpc.WithInsecure(),
+		grpc.WithContextDialer(dialerFunc),
+	)
+	if err != nil {
+		return false, fmt.Errorf("failed to dial Cosmos gRPC service: %w", err)
+	}
+
+	defer grpcConn.Close()
```
yzang2019: Can we keep the grpcConn open?

oracle/price-feeder/oracle/jail.go
yzang2019: Yeah, we don't really need this and the above check. It should only become stale if currentBlockHeight > lastUpdate + interval

yzang2019:
Apart from adding block height based stale, we should probably add time based staleness check. This would cover the case where the node halted and become jailed, where the curr height doesn't move any more, but we should still stop sending any request right?

oracle/price-feeder/oracle/jail.go
```diff
@@ -0,0 +1,31 @@
+package oracle
+
+const (
+	jailCacheInterval = int64(50)
```
philipsu522: nit - can you add units to the name of this var?

oracle/price-feeder/oracle/oracle.go
philipsu522: fucntion seems to return jail state rather than the actual cache - should we rename?

philipsu522: or would `GetCachedJailState` make more sense? I thought we were returning some cache like
```

type JailCache struct {
	isJailed         bool
	lastUpdatedBlock int64
}
```

philipsu522 (approved):
more of a meta point, does it make sense to start having metrics here?

oracle/price-feeder/oracle/oracle.go
udpatil: I followed the behavior here:
https://github.com/sei-protocol/sei-chain/blob/7f5681547fd498c5fbdd9fc8af45e200caae7e33/oracle/price-feeder/oracle/oracle.go#L408
where it closes the connection after the GetParams, I guess we could open one at the start of oracle and leave it open and then only close it when we terminate the price feeder?

PR #552, opened 2023-02-08
Charge rent even if contract failed execution in EndBlock

codchen:
Dex EndBlock has a rollback mechanism such that any state change by a failed contract would be reverted. However this rollback should not include the rent charge applied to the failed contract; the rent should be always be charged. This PR adds an explicit operation for rent charge to be applied to the outer cache store.

unit test


yzang2019 (approved):
LGTM, let's fix the lint issue

philipsu522 (approved):

PR #554, opened 2023-02-10
Delete contract-related states unpon unregister

codchen:
Remove all states associated with a contract if it is unregistered

unit tests

### DIFF
--- a/x/dex/types/keys.go
+++ b/x/dex/types/keys.go
@@ -45,16 +45,20 @@ func PairPrefix(priceDenom string, assetDenom string) []byte {
 }
 
 func OrderBookPrefix(long bool, contractAddr string, priceDenom string, assetDenom string) []byte {
+	return append(
+		OrderBookContractPrefix(long, contractAddr),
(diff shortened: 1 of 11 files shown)


x/dex/types/keys.go
```diff
@@ -45,16 +45,20 @@ func PairPrefix(priceDenom string, assetDenom string) []byte {
 }
 
 func OrderBookPrefix(long bool, contractAddr string, priceDenom string, assetDenom string) []byte {
+	return append(
+		OrderBookContractPrefix(long, contractAddr),
+		PairPrefix(priceDenom, assetDenom)...,
+	)
```
BrandonWeng, line 51: Do we need to worry about where this is called?

codchen: this is just a refactor so the caller wouldn't see a difference

BrandonWeng: bbut in the old method it appends the `append(prefix` wheras this doesnt?

BrandonWeng (approved):

PR #556, opened 2023-02-10
Enable Profiling by default 

BrandonWeng:
https://stackoverflow.com/questions/26545159/how-big-is-pprof-import-overhead-in-go

### DIFF
--- a/cmd/seid/cmd/root.go
+++ b/cmd/seid/cmd/root.go
@@ -6,6 +6,8 @@ import (
 	"io"
 	"math"
 	"math/rand"
+	"net/http"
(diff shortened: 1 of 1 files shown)


cmd/seid/cmd/root.go
```diff
@@ -6,6 +6,8 @@ import (
 	"io"
 	"math"
 	"math/rand"
+	"net/http"
+	_ "net/http/pprof"
```
yzang2019, line 10: // nolint:gosec,G108

yzang2019 (approved):

codchen (approved):

PR #560, opened 2023-02-13
Add mint and dex module parsers

udpatil:
This change adds mint and dex module parsers for the iavl-dump script.

Tested on cluster with dex activity, example parsed line:
```
  4C6F6E67426F6F6B2D76616C75652D2048B7A40898C7CF2635910035709590D13FD1F323642BB7EB1EB3E6B14D49:zSEI|ATOM45000000000000 | LongBook-value- | AddrBech32: sei1fzm6gzyccl8jvdv3qq6hp9vs6ylaruervs4m06c7k0ntzn2f8faq8un0p6 | RemainingString: SEI|ATOM45000000000000
  ```

### DIFF
--- a/cmd/seid/cmd/iavl_parser.go
+++ b/cmd/seid/cmd/iavl_parser.go
@@ -63,69 +63,62 @@ func BankParser(key []byte) ([]string, error) {
 
 func DexParser(key []byte) ([]string, error) {
 	keyItems := []string{}
+	if matched, items, _, err := MatchAndExtractDexAddressPrefixKeys(key); matched {
+		if err != nil {
+			return keyItems, err
+		}
+		keyItems = append(keyItems, items...)
+		return keyItems, nil
+	}
 	switch {
-	case bytes.HasPrefix(key, dextypes.KeyPrefix(dextypes.LongBookKey)):
-		keyItems = append(keyItems, "LongBook")
-		remaining := bytes.TrimPrefix(key, dextypes.KeyPrefix(dextypes.LongBookKey))
-		items, remaining, err := parseLengthPrefixedAddress(remaining)
-		if err != nil {
-			return keyItems, err
-		}
-		keyItems = append(keyItems, items...)
-		// TODO: make this better
-		keyItems = append(keyItems, fmt.Sprintf("RemainingString: %s", string(remaining)))
-	case bytes.HasPrefix(key, dextypes.KeyPrefix(dextypes.ShortBookKey)):
-		keyItems = append(keyItems, "ShortBook")
-		remaining := bytes.TrimPrefix(key, dextypes.KeyPrefix(dextypes.ShortBookKey))
-		items, remaining, err := parseLengthPrefixedAddress(remaining)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cmd/seid/cmd/iavl_parser.go [hunk 1]
```diff
@@ -37,19 +49,86 @@ func BankParser(key []byte) ([]string, error) {
...
+		// do nothing since the key is a string and no other data to be parsed
+	default:
+		keyItems = append(keyItems, "Unrecognized prefix")
+	}
+	return keyItems, nil
+}
+
+func MatchAndExtractDexAddressPrefixKeys(key []byte) (bool, []string, []byte, error) {
+	keyItems := []string{}
+	keysToMatch := []string{
```
BrandonWeng, line 85: might be helpful to add a source of truth can be found if it can't be updated automatically

udpatil: Added comment discussing source of truth. The constants are all in the dex keys file

cmd/seid/cmd/iavl_parser.go
```diff
@@ -6,22 +6,34 @@ import (
...
+
+func MintParser(key []byte) ([]string, error) {
+	keyItems := []string{}
+	switch {
+	case bytes.HasPrefix(key, minttypes.MinterKey):
+		keyItems = append(keyItems, "MinterKey")
+	case bytes.HasPrefix(key, minttypes.LastTokenReleaseDate):
+		keyItems = append(keyItems, "LastTokenReleaseDate")
+	default:
+		keyItems = append(keyItems, "Unrecognized prefix")
```
philipsu522, line 30: typically you would use a tombstone value for this, and it can be used across all modules (i.e. default case forDex, Bank parser as well).
basically a static string that denotes unrecognized

cmd/seid/cmd/iavl_parser.go
```diff
@@ -37,19 +49,87 @@ func BankParser(key []byte) ([]string, error) {
...
+	}
+	return false, keyItems, key, nil
+}
+
+func parseLengthPrefixedAddress(remainingKey []byte) ([]string, []byte, error) {
+	keyItems := []string{}
+	lengthPrefix, remaining := int(remainingKey[0]), remainingKey[1:]
+	accountAddr := remaining[0:lengthPrefix]
+	remaining = remaining[lengthPrefix:]
+	bech32Addr, err := sdk.Bech32ifyAddressBytes("sei", accountAddr)
```
philipsu522, line 129: can we reference this https://github.com/sei-protocol/sei-chain/blob/master/app/params/config.go#L21 instead?

philipsu522 (approved):
some nits, but lgtm

cmd/seid/cmd/iavl_parser.go
udpatil: made this and the above fix in https://github.com/sei-protocol/sei-chain/pull/562, can we review that instead since it also contains the staking parser logic

PR #561, opened 2023-02-13
make chain-id optional for seid start

BrandonWeng:
On init, save chain-id to client.toml (used to other commands to run seid commands to bypass requirement for --chain-id already) 

Then on startup, just read from client.toml's value if chainId isn't passed

Ran the new `./scripts/old_initialize_local.sh` 

image

Then restarted chain with `seid start`

Also validates
image

### DIFF
--- a/cmd/seid/cmd/init.go
+++ b/cmd/seid/cmd/init.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/cosmos/go-bip39"
 	"github.com/pkg/errors"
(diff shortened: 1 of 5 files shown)


cmd/seid/cmd/init.go
```diff
@@ -138,6 +141,12 @@ func InitCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
 				return errors.Wrap(err, "Failed to export genesis file")
 			}
 
+			clientConfig, err := clientconfig.GetClientConfig(configPath, clientCtx.Viper)
+			if err != nil {
+				return err
+			}
+			clientconfig.SetClientConfig(flags.FlagChainID, chainID, configPath, clientConfig)
```
philipsu522: shoudl we handle any errs here

philipsu522 (approved):

PR #565, opened 2023-02-15
Add metrics for oracle vote penalties

BrandonWeng:
Better visibility around other validator's oracle votes and if they're missing 

Helps with trackign oracle related changes

### DIFF
--- a/utils/metrics/metrics_util.go
+++ b/utils/metrics/metrics_util.go
@@ -104,3 +104,12 @@ func MeasureDeliverTxDuration(start time.Time) {
 		start.UTC(),
 	)
 }
+
+// sei_oracle_vote_penalty_count
+func SetOracleVotePenaltyCount(count uint64, penaltyType string) {
+	metrics.SetGaugeWithLabels(
+		[]string{"sei", "oracle", "vote", "penalty", "count"},
+		float32(count),
+		[]metrics.Label{telemetry.NewLabel("type", penaltyType)},
+	)
+}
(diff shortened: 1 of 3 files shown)


x/oracle/keeper/keeper.go
```diff
@@ -181,6 +183,9 @@ func (k Keeper) GetVotePenaltyCounter(ctx sdk.Context, operator sdk.ValAddress)
 
 // SetVotePenaltyCounter updates the # of vote periods missed in this oracle slash window
 func (k Keeper) SetVotePenaltyCounter(ctx sdk.Context, operator sdk.ValAddress, missCount uint64, abstainCount uint64) {
+	defer metrics.SetOracleVotePenaltyCount(missCount, "miss")
```
udpatil: shouldn't we also be labeling on the validator address so the vote penalties are partitioned by val Addr

BrandonWeng: ah yeah forgot,t hanks!

utils/metrics/metrics_util.go
```diff
@@ -104,3 +104,15 @@ func MeasureDeliverTxDuration(start time.Time) {
 		start.UTC(),
 	)
 }
+
+// sei_oracle_vote_penalty_count
+func SetOracleVotePenaltyCount(count uint64, valAddr string, penaltyType string) {
+	metrics.SetGaugeWithLabels(
```
udpatil, line 110: btw, not sure exactly how this behaves vs `incr`, but are we able to get the combined vote count if we query with a `validator` but without a `type`? because the slashing is based on the combined value so would be helpful to have, or if we can't then maybe just add a `combined` label that just adds the two vals

BrandonWeng: as discussed offline, we can sum by the validator

udpatil (approved):

PR #571, opened 2023-02-17
Defer store read/write from iterator loop

codchen:
Performing store access before an iterator is closed may cause deadlocks (see https://github.com/sei-protocol/sei-chain/pull/567 for context).
This PR refactors existing logic to move store access that were previously within iterator's scope to be after iterator closure.

Note that there are still cases for this in oracle keeper, but all the methods there are only called in oracle's Mid/EndBlock, which are not parallelized in any way and thus not subject to the deadlock issue

unit tests + existing tests

### DIFF
--- a/x/dex/cache/cache.go
+++ b/x/dex/cache/cache.go
@@ -182,17 +182,24 @@ func DeepDelete(kvStore sdk.KVStore, storePrefix []byte, matcher func([]byte) bo
 		kvStore,
 		storePrefix,
 	)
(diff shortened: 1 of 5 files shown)


yzang2019 (approved):

philipsu522 (approved):
I wonder if it's possible to add some sort of unit test that ensures that panics to ensure that there is no store read/write within iterator loop? We can also have a whitelist of "stores" that are allowed to do this (i.e. oracle). Wdyt?
This can be a separate PR though

PR #572, opened 2023-02-17
[oracle] Sorts map keys prior to tallying for consistent ordering

udpatil:
This fixes map iteration to happen in sorted order for consistent event order, and also removes store READs within a store iterator

Existing tests

### DIFF
--- a/oracle/price-feeder/config.example.toml
+++ b/oracle/price-feeder/config.example.toml
@@ -55,7 +55,7 @@ enable_service_label = true
 enabled = true
 global_labels = [["chain-id", "sei-chain"]]
 service_name = "price-feeder"
-type = "prometheus"
+prometheus_retention = 60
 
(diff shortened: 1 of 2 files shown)


oracle/price-feeder/config.example.toml
```diff
@@ -55,7 +55,7 @@ enable_service_label = true
 enabled = true
 global_labels = [["chain-id", "sei-chain"]]
 service_name = "price-feeder"
-type = "prometheus"
+prometheus_retention = 60
```
philipsu522, line 58: is this intended?

philipsu522 (approved):

oracle/price-feeder/config.example.toml
udpatil: yeah since the `type` field doesn't actully do anything

PR #581, opened 2023-02-21
Fix documentation for default inflation rate

yzang2019:
**Problem**:
Update inflation documentation In x/mint/types/minter.go:26 is stated that the intended default initial inflation rate should be 13%. However, in L21 the initial inflation rate is set to sdk.NewDec(0). If this is not intended behavior, this error may cause severe errors on the initial token distribution of the Sei token.


yzang2019 (commented):
I'm a bit confused whether we should fix documentation or fix the default value

udpatil (approved):

philipsu522 (approved):

PR #582, opened 2023-02-21
[oak] Change oracle lookback seconds to uint64

Kbhat1:
- Changes oracle lookback seconds to `uint64` - consistent with dex twap lookback seconds

NOTE: PR contains fmt from unrelated files

- Tested oracle e2e


x/oracle/keeper/keeper.go
```diff
@@ -376,7 +376,7 @@ func (k Keeper) AddPriceSnapshot(ctx sdk.Context, snapshot types.PriceSnapshot)
 	var lastOutOfRangeSnapshotTimestamp int64 = -1
 	// we need to evict old snapshots (except for one that is out of range)
 	k.IteratePriceSnapshots(ctx, func(snapshot types.PriceSnapshot) (stop bool) {
-		if snapshot.SnapshotTimestamp+lookbackDuration >= ctx.BlockTime().Unix() {
+		if snapshot.SnapshotTimestamp+int64(lookbackDuration) >= ctx.BlockTime().Unix() {
```
udpatil: is it safe to just cast the uint64 into int64? would this panic is uint64 value is above the positive range of int64? (realistically the lookback duration will be magnitudes smaller, but worth taking proper steps here) is there a checked conversion that will raise error or something if appropriate?

udpatil (approved):

x/oracle/keeper/keeper.go
Kbhat1: This has been updated, as we discussed I added a panic if the lookbackDuration is larger than int64 max. In reality though the magnitude would be quite small so we should never hit this case

x/oracle/keeper/keeper.go
```diff
@@ -370,13 +371,21 @@ func (k Keeper) SetPriceSnapshot(ctx sdk.Context, snapshot types.PriceSnapshot)
 
 func (k Keeper) AddPriceSnapshot(ctx sdk.Context, snapshot types.PriceSnapshot) {
 	params := k.GetParams(ctx)
-	lookbackDuration := params.LookbackDuration
+
+	// Sanity check to make sure LookbackDuration can be converted to int64
+	// Lookback duration should never get this large
+	if params.LookbackDuration > uint64(math.MaxInt64) {
```
udpatil, line 377: is Uint(MaxInt64) and MaxInt64 semantically the same value? idr the way the representation of sign works so just wanted to make sure it would be fine and not like a max Uint64 value

Kbhat1: Yes its semantically the same. I also verified this.

PR #588, opened 2023-02-22
Send deposit funds in message server instead of EndBlock

codchen:
We originally placed coin sending logic in EndBlock to improve performance with batch operation (e.g. if an account placed a lot of micro deposits in the same currency, it will only be sent once in EndBlock). However this opened up an attack vector as described in https://linear.app/seilabs/issue/SEI-2678/[sei-chain]-msgplaceorders-should-hold-funds-in-escrow So we are moving the logic into transaction processing stage.

unit tests

### DIFF
--- a/aclmapping/dex/mappings.go
+++ b/aclmapping/dex/mappings.go
@@ -8,6 +8,9 @@ import (
 	sdkacltypes "github.com/cosmos/cosmos-sdk/types/accesscontrol"
 	aclkeeper "github.com/cosmos/cosmos-sdk/x/accesscontrol/keeper"
 	acltypes "github.com/cosmos/cosmos-sdk/x/accesscontrol/types"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
+	"github.com/sei-protocol/sei-chain/aclmapping/utils"
 	dextypes "github.com/sei-protocol/sei-chain/x/dex/types"
 )
 
@@ -47,6 +50,8 @@ func DexPlaceOrdersDependencyGenerator(keeper aclkeeper.Keeper, ctx sdk.Context,
 		return []sdkacltypes.AccessOperation{}, ErrPlaceOrdersGenerator
 	}
 
+	senderBankAddrIdentifier := hex.EncodeToString(banktypes.CreateAccountBalancesPrefixFromBech32(placeOrdersMsg.Creator))
+	contractBankAddrIdentifier := hex.EncodeToString(banktypes.CreateAccountBalancesPrefixFromBech32(placeOrdersMsg.ContractAddr))
 	contractAddr := placeOrdersMsg.ContractAddr
 
 	aclOps := []sdkacltypes.AccessOperation{
@@ -81,6 +86,7 @@ func DexPlaceOrdersDependencyGenerator(keeper aclkeeper.Keeper, ctx sdk.Context,
 				[]byte(placeOrdersMsg.Creator)...,
 			)),
 		},
+
 		{
 			AccessType:         sdkacltypes.AccessType_READ,
 			ResourceType:       sdkacltypes.ResourceType_KV_DEX_MEM_ORDER,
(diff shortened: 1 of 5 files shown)


aclmapping/dex/mappings.go
```diff
@@ -91,6 +97,74 @@ func DexPlaceOrdersDependencyGenerator(keeper aclkeeper.Keeper, ctx sdk.Context,
...
+			IdentifierTemplate: utils.DefaultIDTemplate,
+		})
+		return aclOps, nil
+	}
+	if !keeper.AccountKeeper.HasAccount(ctx, toAddr) {
+		aclOps = append(aclOps, sdkacltypes.AccessOperation{
+			AccessType:         sdkacltypes.AccessType_WRITE,
+			ResourceType:       sdkacltypes.ResourceType_KV_AUTH_GLOBAL_ACCOUNT_NUMBER,
+			IdentifierTemplate: hex.EncodeToString(authtypes.GlobalAccountNumberKey),
+		})
```
udpatil: deps looks good

x/dex/keeper/msgserver/msg_server_place_orders.go
```diff
@@ -44,6 +48,9 @@ func (k msgServer) transferFunds(goCtx context.Context, msg *types.MsgPlaceOrder
 			Amount:  sdk.NewDec(fund.Amount.Int64()),
 		})
 	}
+	if err := k.BankKeeper.SendCoins(ctx, sender, contractAddr, msg.Funds); err != nil {
```
udpatil: is this a problem for contracts if they receive the coins during block (eg. other wasm txs could be made against the contract as well), and then the deposit info is sent to the contract in endblock? its probably not our responsibility to prevent alll exploitable behavior in contracts, but does this method of account make it easier for contract developers to accidentally add exploits where users making withdrawals may be incorrectly getting a larger amount of funds due to the deposit that happened under the hood to the contract balance instead of via contract wasm call with accompanying funds?

udpatil: wouldn't the appropriate behavior be to instead transfer the coins to the module account as the escrow, and then the module account transfers to the contract via sudo in endblock?

x/dex/keeper/msgserver/msg_server_place_orders.go
```diff
@@ -34,6 +34,10 @@ func (k msgServer) transferFunds(goCtx context.Context, msg *types.MsgPlaceOrder
 		return sdkerrors.Wrapf(sdkerrors.ErrUnauthorized, "%s is not allowed to receive funds", contractAddr.String())
 	}
 
+	sender, err := sdk.AccAddressFromBech32(msg.Creator)
```
LCyson: we should do all msg validations in the ValidateBasics of the `message_place_order` so that we don't miss/over check

x/dex/keeper/msgserver/msg_server_place_orders.go
LCyson: i'd be inclined to keep the deposit logic as simple as it is so that both contract developers and us can understand and avoid possible exploits, which can be missed in an complicated design

codchen: > is this a problem for contracts if they receive the coins during block (eg. other wasm txs could be made against the contract as well), and then the deposit info is sent to the contract in endblock? its probably not our responsibility to prevent alll exploitable behavior in contracts, but does this method of account make it easier for contract developers to accidentally add exploits where users making withdrawals may be incorrectly getting a larger amount of funds due to the deposit that happened under the hood to the contract balance instead of via contract wasm call with accompanying funds?

`where users making withdrawals may be incorrectly getting a larger amount of funds` can you elaborate on this possibility maybe with an example?

udpatil: For example, if the contract is performing accounting based on overall balance and a users share of that balance, that overall balance being updated prior to end block order updates being updated in the contract may mean a withdrawal later in that block may result in the withdrawal being performed against an incorrect balance.

x/dex/keeper/msgserver/msg_server_place_orders.go
```diff
@@ -39,6 +37,10 @@ func (k msgServer) transferFunds(goCtx context.Context, msg *types.MsgPlaceOrder
 			Amount:  sdk.NewDec(fund.Amount.Int64()),
 		})
 	}
+	escrow := k.AccountKeeper.GetModuleAccount(ctx, types.ModuleName).GetAddress()
+	if err := k.BankKeeper.SendCoins(ctx, sender, escrow, msg.Funds); err != nil {
```
udpatil: @BrandonWeng @codchen should this be a deferred send instead of a normal send? otherwise wouldn't this message have a bottleneck on module account? when is the deferred send processed relative to when the funds are needed for endblock processing?

BrandonWeng (approved):

PR #598, opened 2023-02-23
[oak audit] Deposit Rent with new owner claim should not fail

yzang2019:
**Problem**:
In the ContractDepositRent function, there is logic for a new owner to claim a contract when the contract.RentBalance==0. However, in x/dex/keeper/msgserver/ msg_server_contract_deposit_rent.go:48, the SendCoins operation is executed from the old creatorAddr, not the msg.Sender. Thus, this function will always fail when called by any address which is not the current owner, rendering the claim functionality unusable.

**Fix**:
- Added a check and change the address to be msg.sender
- Added a unit test
Added unit test

### DIFF
--- a/x/dex/keeper/msgserver/msg_server_contract_deposit_rent.go
+++ b/x/dex/keeper/msgserver/msg_server_contract_deposit_rent.go
@@ -36,12 +36,18 @@ func (k msgServer) ContractDepositRent(goCtx context.Context, msg *types.MsgCont
 		// a sender can only "claim" the contract if the rent balance is 0
 		return nil, sdkerrors.ErrUnauthorized
 	}
-	// deposit
-	creatorAddr, err := sdk.AccAddressFromBech32(contract.Creator)
+
+	senderAddr, err := sdk.AccAddressFromBech32(contract.Creator)
+	// claim a contract when the contract.RentBalance==0 should be executed from msg.Sender
+	if contract.RentBalance == 0 {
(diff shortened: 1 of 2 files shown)


x/dex/keeper/msgserver/msg_server_contract_deposit_rent.go
```diff
@@ -36,12 +36,18 @@ func (k msgServer) ContractDepositRent(goCtx context.Context, msg *types.MsgCont
...
+	if contract.RentBalance == 0 {
+		senderAddr, err = sdk.AccAddressFromBech32(msg.Sender)
+	}
 	if err != nil {
 		return nil, err
 	}
-	if err := k.BankKeeper.SendCoins(ctx, creatorAddr, k.AccountKeeper.GetModuleAddress(types.ModuleName), sdk.NewCoins(sdk.NewCoin(appparams.BaseCoinUnit, sdk.NewInt(int64(msg.Amount))))); err != nil {
+
+	// deposit
+	if err := k.BankKeeper.SendCoins(ctx, senderAddr, k.AccountKeeper.GetModuleAddress(types.ModuleName), sdk.NewCoins(sdk.NewCoin(appparams.BaseCoinUnit, sdk.NewInt(int64(msg.Amount))))); err != nil {
```
udpatil, line 50: doesnt this make it such that anybody could force the creator account to transfer funds to the contract rent (if rent != 0). Isn't this an issue since rent can't be withdrawn so anyone could force the account to have its whole balance to be locked in the rent (almost like burning?)

PR #600, opened 2023-02-23
Add params to guard Nitro fraud challenge

codchen:
We will not offer fraud challenge from the get-go due to concerns for DDoS attacks. Adding a parameter to guard fraud challenge that's default to false

unit tests


LCyson (approved):

LCyson:
i assume we don't need a migration handler for this as this param is a boolean?

PR #605, opened 2023-02-24
Remove unsafe cors for api and grpc

BrandonWeng:
as per todo


besated (approved):
Just to confirm, removing this will prevent other domains from sending requests to the rest/rpc urls?

LCyson (approved):

PR #618, opened 2023-02-27
Modify loadtest metric script to output best block

philipsu522:
Adds "Best Block" output:
```
    "Best block": {
        "height": 56602,
        "num_txs": 87,
        "tx_mapping": {
            "bank": 30,
            "oracle": 28
        },
        "block_time_ms": 506
```
It simply picks the best block based on most number of txs and shows you the tx breakdown. Note that sum(tx_mapping) < num_txs b/c the block contains txs that failed checks as well (i.e. code != 0).
Ran and tested

### DIFF
--- a/loadtest/scripts/metrics.py
+++ b/loadtest/scripts/metrics.py
@@ -1,9 +1,10 @@
 import json
 import requests
 import os
-from datetime import datetime
+from datetime import datetime, timedelta
 
-def get_block_height_and_timestamp(tx_hash_str):
+DATE_TIME_FMT = "%Y-%m-%dT%H:%M:%S.%f"
+def get_block_height(tx_hash_str):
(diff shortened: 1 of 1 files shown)


loadtest/scripts/metrics.py
```diff
@@ -48,6 +75,16 @@ def get_metrics():
     total_txs_num = sum([block["number_of_txs"] for block in skip_edge_blocks])
     average_txs_num = total_txs_num / len(skip_edge_blocks)
 
+    # Best block stats:
```
codchen, line 81: the definition of "Best Block" I used was actually the block with the highest throughput, not necessarily the most transactions. I wonder if it'd be easier to just print a tuple (number of tx, tx per second, duration, tx mapping) for every block

codchen (approved):

PR #621, opened 2023-02-28
Add success count to the oracle query 

BrandonWeng:
Requested by one of the validators, right now we're just dividing the penalities by the window size to get the % error. 

We would need to calculate how many successes we have based on the 
`((uint64)(ctx.BlockHeight())+1)%blocksPerPeriod  - penalties` i thought it would just be easier to track it as a field and return it. This way we can emit metrics too  

image


udpatil (commented):
LGTM, make sure to add a migration as well that will infer the success count from the block height + miss / abstain counts and update the store appropriately

BrandonWeng:
> LGTM, make sure to add a migration as well that will infer the success count from the block height + miss / abstain counts and update the store appropriately

Hmm we would need to do it when we register the upgrade handler right? Ill add a TODO in the upgrade handler file

udpatil (approved):
lgtm

PR #627, opened 2023-03-03
[oracle] Add slash window progress query

udpatil:
At the request of validators, a slash window progress query would be useful to have. This query is very similar to the Umee implementation that was mentioned as a useful endpoint.

unit tests


BrandonWeng (approved):

proto/oracle/query.proto
```diff
@@ -196,6 +202,18 @@ message QueryAggregateVotesResponse {
   repeated AggregateExchangeRateVote aggregate_votes = 1 [(gogoproto.nullable) = false];
 }
 
+// QuerySlashWindow is the request type for the
+// Query/SlashWindow RPC method.
+message QuerySlashWindow {}
```
yzang2019: Based on the above naming pattern, we should name this as QuerySlashWindowRequest right?

udpatil: ah yes good point, I'll fix

PR #637, opened 2023-03-07
Fix race condition when optimisticProcessingInfo is not nil

LCyson:
This PR adds a fix for race condition in which `optimisticProcessingInfo` is not nil and previous channel was not reset.

- e2e with `optimisticProcessingInfo` is nil scenario
- unit tests

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -944,8 +944,8 @@ func (app *App) ClearOptimisticProcessingInfo() {
 }
 
 func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcessProposal) (*abci.ResponseProcessProposal, error) {
+	completionSignal := make(chan bool, 1)
 	if app.optimisticProcessingInfo == nil {
-		completionSignal := make(chan bool, 1)
 		optimisticProcessingInfo := &OptimisticProcessingInfo{
 			Height:     req.Height,
 			Hash:       req.Hash,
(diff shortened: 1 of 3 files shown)


app/app.go
```diff
@@ -966,9 +966,15 @@ func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcess
 				optimisticProcessingInfo.Completion <- true
 			}()
 		}
-	} else if !bytes.Equal(app.optimisticProcessingInfo.Hash, req.Hash) {
-		app.optimisticProcessingInfo.Completion <- false
+	} else {
+		app.optimisticProcessingInfo.Completion = completionSignal
```
philipsu522, line 970: should we just set `app.optimisticProcessingInfo.Completion = completionSignal` in all branches? i.e. should line 959 do this too?

app/app.go
```diff
@@ -966,9 +966,15 @@ func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcess
...
+		if !bytes.Equal(app.optimisticProcessingInfo.Hash, req.Hash) {
+			app.optimisticProcessingInfo.Completion <- false
+		} else {
+			app.optimisticProcessingInfo.Completion <- true
```
codchen, line 974: we do not want to set true here since the OP goroutine could still be running at this time. It would cause ProcessBlock in FinalizeBlock to run in parallel with the OP goroutine which would cause non determinism

codchen (changes requested):

LCyson:
Cancelled as this change may introduce some nondeterminism

PR #642, opened 2023-03-08
Fix deadlock when optimistic processing is skipped for the 1st proposal

codchen:
Optimistic processing may be skipped even for the first proposal of a height for reasons like upgrade. In those cases we want to mark completion immediately so that FinalizeBlocker won't be blocked.

The original fix was reverted due to concerns of regression on the non-determinism problem.


BrandonWeng (approved):

LCyson (approved):
gonna cancel #640 as it's duplicate: https://github.com/sei-protocol/sei-chain/pull/640

PR #654, opened 2023-03-14
Improve endblock performance and fix tracer

yzang2019:
**Performance Improvements**

Currently newEnv will have to create all orderbooks sequentially, this would increase the time complexity linearly as the number of contracts. This change makes the orderbooks population happening in parallel so that we could save a lot of time doing that.

The result of this improvement will reduce the latency from 40-50ms to 10ms, (75% reduction)

Tested for snapshotter node


philipsu522 (approved):
🔥

PR #663, opened 2023-03-18
Record and emit transaction, metrics and order counter 

BrandonWeng:
Emit metrics for chain level performance. 

We need to store it in the baseapp as a counter we only scrape every 30seconds and our block times are < 1s most of the time, so if it's emitted per block it's not entirely accurate. (As seen in my testing, it was severally underreporting) 

The reason why we add the final counter at the end of the ProcessBlock is also due to the lifecycle of the contextMemcache object, and that's when we're sure that the counters are the ones being committed to the block. 

See this PR:
https://github.com/sei-protocol/sei-cosmos/pull/200

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1015,6 +1015,8 @@ func (app *App) RecordAndEmitMetrics(ctx sdk.Context) {
 	for metricName, value := range *(app.metricCounter) {
 		metrics.IncrementThroughputMetrics(metricName, float32(value))
 	}
+
+	ctx.ContextMemCache().Clear()
 }
 
 func (app *App) DeliverTxWithResult(ctx sdk.Context, tx []byte) *abci.ExecTxResult {
--- a/app/app.go
+++ b/app/app.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	storetypes "github.com/cosmos/cosmos-sdk/store/types"
+	"github.com/cosmos/iavl/internal/logger"
 
 	"github.com/sei-protocol/sei-chain/aclmapping"
 	aclutils "github.com/sei-protocol/sei-chain/aclmapping/utils"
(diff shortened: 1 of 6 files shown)


go.sum
```diff
@@ -1067,8 +1067,8 @@ github.com/seccomp/libseccomp-golang v0.9.2-0.20210429002308-3879420cc921/go.mod
 github.com/securego/gosec/v2 v2.11.0 h1:+PDkpzR41OI2jrw1q6AdXZCbsNGNGT7pQjal0H0cArI=
 github.com/securego/gosec/v2 v2.11.0/go.mod h1:SX8bptShuG8reGC0XS09+a4H2BoWSJi+fscA+Pulbpo=
 github.com/segmentio/fasthash v1.0.3/go.mod h1:waKX8l2N8yckOgmSsXJi7x1ZfdKZ4x7KRMzBtS3oedY=
-github.com/sei-protocol/sei-cosmos v0.2.4 h1:t4lQYtoUcLWjoisQtwuppYgL1xKV7TLTnnNmoTvj5Ac=
-github.com/sei-protocol/sei-cosmos v0.2.4/go.mod h1:LUhhplOtRXliV1x17gbOptKPy90xSK0Sxv8zmgsMvTY=
+github.com/sei-protocol/sei-cosmos v0.2.5-0.20230318010240-8584b787f3e5 h1:eA/b9roIzs9lNHB/yqfS0qndiZ+amSolyeAWi7U8kx0=
+github.com/sei-protocol/sei-cosmos v0.2.5-0.20230318010240-8584b787f3e5/go.mod h1:LUhhplOtRXliV1x17gbOptKPy90xSK0Sxv8zmgsMvTY=
```
BrandonWeng: Will properly tag this once the sei-cosmos PR merges

app/app.go
```diff
@@ -341,6 +341,9 @@ type App struct {
 	txDecoder sdk.TxDecoder
 
 	versionInfo version.Info
+
+	// Stores mapping counter name to counter value
+	metricCounter *map[string]float32
```
BrandonWeng, line 346: metrics are float32

yzang2019 (approved):

philipsu522 (approved):

PR #667, opened 2023-03-21
Hot fix for max gas change

BrandonWeng:


yzang2019 (approved):

udpatil (changes requested):
We need a special hotfix that doesn't include slashing migration and things later than the 2.0.40beta version

PR #674, opened 2023-03-30
Reject proposal if max block gas is exceeded

codchen:
Implementation for https://github.com/sei-protocol/sei-chain/pull/new/tony-chen-max-block-gas-check

We want honest validators (i.e. validators who do not use a maliciously modified binary) to reject block proposals whose transactions' GasWanted add up to something greater than max block gas. Since transactions in block proposals are sent as raw bytes, the only place to decode them would be on application-level during ProcessProposal. For context, ProcessProposal is called during `prevote` phase by validators, and if it returns a status of `REJECT`, validators will vote nil.

unit test
integration test in a loadtest cluster with a modified node to verify such proposal will indeed be rejected

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -943,6 +943,13 @@ func (app *App) ClearOptimisticProcessingInfo() {
 }
 
 func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcessProposal) (*abci.ResponseProcessProposal, error) {
+	// TODO: this check decodes transactions which is redone in subsequent processing. We might be able to optimize performance
+	// by recording the decoding results and avoid decoding again later on.
+	if !app.checkTotalBlockGasWanted(ctx, req.Txs) {
+		return &abci.ResponseProcessProposal{
+			Status: abci.ResponseProcessProposal_REJECT,
+		}, nil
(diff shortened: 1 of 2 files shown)


app/app.go
```diff
@@ -1468,6 +1475,28 @@ func (app *App) RegisterTendermintService(clientCtx client.Context) {
 	tmservice.RegisterTendermintService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.interfaceRegistry)
 }
 
+func (app *App) checkTotalBlockGasWanted(ctx sdk.Context, txs [][]byte) bool {
+	totalGasWanted := uint64(0)
+	for _, tx := range txs {
+		decoded, err := app.txDecoder(tx)
```
yzang2019, line 1481: Nit: If we have perf concern, we can make decode run in parallel, and then use atomic integer for calculating totalGasWanted

yzang2019 (approved):

philipsu522 (approved):

app/app.go
codchen: yeah good point. I'll add it in in a later PR if we find this to be a bottleneck

PR #678, opened 2023-03-31
Periodically collect and distribute tokens to test accounts 

BrandonWeng:
Change to periodically collect rewards and also distribute them to test accounts 

Left is node-0, and right is node-1 we see that node-0 (where the LT client is running) has a lot less rewards as it was collected. For sei-internal we'll need to update distribution to collect rewards from the other validators 
image

Saw that the balance for the admin account was dropping 
image

Need to refactor the sei-infra script to test the reward collecting properly (too many acc seq issues caused by oracle not using its own acc)

### DIFF
--- a/loadtest/config.json
+++ b/loadtest/config.json
@@ -17,6 +17,20 @@
         "max": "21",
         "number_of_distinct_values": 20
     },
+    "message_configs": {
+      "default": {
+        "gas": "3000000",
+        "fees": "50000"
+      },
+      "collect_rewards": {
+        "gas": "10000000",
+        "fees": "1000000"
+      },
+      "distribute_rewards": {
+        "gas": "10000000",
+        "fees": "1000000"
+      }
+    },
     "message_type_distribution": {
         "dex": {
             "limit_order_percentage": "0.2",
--- a/loadtest/config.json
+++ b/loadtest/config.json
@@ -19,16 +19,16 @@
     },
     "message_configs": {
       "default": {
-        "gas": "3000000",
-        "fees": "50000"
+        "gas": 3000000,
+        "fees": 50000
       },
       "collect_rewards": {
-        "gas": "10000000",
-        "fees": "1000000"
+        "gas": 10000000,
+        "fees": 1000000
       },
       "distribute_rewards": {
-        "gas": "10000000",
-        "fees": "1000000"
+        "gas": 10000000,
+        "fees": 1000000
       }
     },
     "message_type_distribution": {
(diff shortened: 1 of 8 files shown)


loadtest/main.go
```diff
@@ -101,15 +101,31 @@ func runOnce(config Config) {
...
+	var msg sdk.Msg
+	messageTypes := strings.Split(config.MessageType, ",")
+	messageType := c.getRandomMessageType(messageTypes)
 	fmt.Printf("Message type: %s\n", messageType)
 
 	defer IncrTxMessageType(messageType)
 
+	signer := key
+	gas := 3000000
+	fee := 50000
```
philipsu522: could we have a mapping of:
messageType -> (gas,fee) 
and use that instead?

loadtest/main.go
```diff
@@ -135,6 +151,36 @@ func (c *LoadTestClient) generateMessage(config Config, key cryptotypes.PrivKey,
...
+	case DistributeRewards:
+		adminKey := c.SignerClient.GetAdminKey()
+		msg = &banktypes.MsgSend{
+			FromAddress: sdk.AccAddress(adminKey.PubKey().Address()).String(),
+			ToAddress:   sdk.AccAddress(key.PubKey().Address()).String(),
+			Amount: sdk.NewCoins(sdk.Coin{
+				Denom:  "usei",
+				Amount: sdk.NewInt(10000000),
+			}),
+		}
```
philipsu522, line 172: are we only distributing to 1 validator?

BrandonWeng: yeah, plan is to delegate more from the node_admin account from node-0 so that it can collect the rewards from other validators

loadtest/scripts/run_tests.py
```diff
@@ -36,20 +36,20 @@ def create_burst_loadtest_config(base_config_json):
...
     new_config = base_config_json.copy()
     new_config["constant"] = True
     new_config["metrics_port"] = 9696
-    new_config["txs_per_block"] = 300
-    new_config["msgs_per_tx"] = 10
+    new_config["txs_per_block"] = 500
+    new_config["msgs_per_tx"] = 30
     # Run every min
-    new_config["loadtest_interval"] = 60
+    new_config["loadtest_interval"] = 10
```
BrandonWeng, line 52: Was able to run this for more than 24 hours, think we can just bump up the load

loadtest/config.json
```diff
@@ -17,6 +17,20 @@
         "max": "21",
         "number_of_distinct_values": 20
     },
+    "message_configs": {
```
BrandonWeng, line 20: added it as config here

philipsu522 (approved):

PR #708, opened 2023-04-19
Modified minimum-gas-prices to 0.0001usei

meetrick:
In README.md, corrected the value of `minimum-gas-prices` from `0.01usei` to `0.0001usei` before the current recommended setting.


meetrick:
@udpatil please review it

PR #709, opened 2023-04-19
Update config.example.toml for price-feeder

meetrick:
price-feeder config is an early version. Updated `oracle/price-feeder/config.example.toml` based on the latest working version of price-feeder config.


meetrick:
@udpatil please review it

PR #717, opened 2023-04-21
Add metrics for gas limit and oracle slashing events

BrandonWeng:
Better observability for when the gas exceeds block limit, this should rarely happen 

Also easier to answer "Why a validator got slashed", sei-comsos emits metrics for voting misses and double signing, sei-chain handles oracle slashing: https://github.com/sei-protocol/sei-cosmos/pull/226 

I'll update the dashboard when we deploy these


x/oracle/types/params.go
```diff
@@ -25,8 +25,8 @@ var (
 
 // Default parameter values
 const (
-	DefaultVotePeriod  = utils.BlocksPerMinute / 2 // 30 seconds
-	DefaultSlashWindow = utils.BlocksPerWeek       // window for a week
+	DefaultVotePeriod  = 1                   // Voting every block
```
BrandonWeng, line 28: This is the default we're using in all the chains, I think we should just update it

philipsu522 (approved):

PR #719, opened 2023-04-24
Send consumed rent to fee collector

codchen:
We want to automatically send rent consumed in each EndBlock to FeeCollector.

Specifically we will take an in-mem snapshot of eligible contracts' rent before EndBlock execution starts, and take another only if all contracts execute successfully. Then we will send the difference between the snapshots from `dex` module to `fee_collector` module (note that all rents are already stored in `dex` module's bank account: https://github.com/sei-protocol/sei-chain/blob/master/x/dex/keeper/msgserver/msg_server_contract_deposit_rent.go#L41)

unit tests

### DIFF
--- a/x/dex/contract/abci.go
+++ b/x/dex/contract/abci.go
@@ -6,10 +6,13 @@ import (
 	"sync"
 	"time"
 
+	seiutils "github.com/sei-protocol/sei-chain/utils"
 	"github.com/sei-protocol/sei-chain/utils/logging"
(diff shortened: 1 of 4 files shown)


LCyson (approved):
why we chose not to store rent charged in memory which will be slightly more performant than this solution (still O(n) thou)

codchen:
> why we chose not to store rent charged in memory which will be slightly more performant than this solution (still O(n) thou)

just for simplicity -> less chance for bugs. The perf gain is prob not worth it since the store itself also caches things

PR #727, opened 2023-04-27
Bump max wasm file size to 2MB

LCyson:
This change bumps the maximal uploadable wasm file size from 1024 * 1024 to 1280 * 1024.

- unit test
- e2e contract deployment test

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1580,6 +1580,6 @@ func (app *App) decorateContextWithDexMemState(base context.Context) context.Con
 }
 
 func init() {
-	// override max wasm size to 1MB
-	wasmtypes.MaxWasmSize = 1280 * 1024
+	// override max wasm size to 2MB
+	wasmtypes.MaxWasmSize = 2 * 1024 * 1024
 }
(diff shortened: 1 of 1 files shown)


app/app.go
```diff
@@ -1578,5 +1578,5 @@ func (app *App) decorateContextWithDexMemState(base context.Context) context.Con
 
 func init() {
 	// override max wasm size to 1MB
-	wasmtypes.MaxWasmSize = 1024 * 1024
+	wasmtypes.MaxWasmSize = 1280 * 1024
```
codchen: any particular reason to choose this number vs. maybe just 2MB? also the comment above needs to be updated

codchen (approved):

app/app.go
philipsu522: +1

philipsu522 (approved):

app/app.go
LCyson: I'm inclined not to bump the uploadable wasm size too much as larger wasm file has negative impact on chain performance. Choosing this number to unblock Levana. We can also make a chain upgrade to adjust the param here, but let me know if you have a different thought

LCyson: will bump it to 2MB for now

PR #733, opened 2023-04-28
Remove liquidation logics from dex

codchen:
We've recognized liquidation as a vortex-specific feature rather than something generic to all exchange applications, so we removed the sudo call related to liquidation but didn't completely clean it up, which is fixed by this PR

unit tests


x/dex/client/cli/tx/tx_register_contract.go
```diff
@@ -16,10 +16,10 @@ var _ = strconv.Itoa(0)
 
 func CmdRegisterContract() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "register-contract [contract address] [code id] [need hook] [need order matching] [deposit] [dependency1,dependency2,...]",
+		Use:   "register-contract [contract address] [code id] [(deprecated)] [need order matching] [deposit] [dependency1,dependency2,...]",
```
LCyson, line 19: think we can just remove this flag here as only the spot-lite and vortex contract deployment scripts are using this cmd

LCyson (approved):
can we test this with vortex to verify it doesn't break the basic order placement flow

x/dex/client/cli/tx/tx_register_contract.go
codchen: i think some infra tooling (e.g. loadtest) also relies on this so want to change it after we are done with all loadtesting

PR #736, opened 2023-04-28
Wait for completion signal in unit test

BrandonWeng:
This is the logic in the app code already

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -340,7 +340,7 @@ type App struct {
 	versionInfo version.Info
 
 	// Stores mapping counter name to counter value
-	metricCounter *map[string]float32
+	metricCounter      *map[string]float32
 	metricCounterMutex sync.Mutex
 
 	mounter func()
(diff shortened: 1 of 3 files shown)


app/upgrade_test.go
```diff
@@ -56,6 +55,9 @@ func TestSkipOptimisticProcessingOnUpgrade(t *testing.T) {
 		Height: 1,
 	})
 
+	// Sleep for 1 second in case the ProcessProposalHandler goroutine is still running
+	time.Sleep(1 * time.Second)
+
```
BrandonWeng, line 60: Slightly difficult to wait on the completion channel since that's accessing the optimistic processing info object too and I don't want to make any changes to the handler at this point

app/app.go
```diff
@@ -940,8 +940,8 @@ func (app *App) PrepareProposalHandler(ctx sdk.Context, req *abci.RequestPrepare
 	}, nil
 }
 
-func (app *App) GetOptimisticProcessingInfo() OptimisticProcessingInfo {
-	return *app.optimisticProcessingInfo
+func (app *App) GetOptimisticProcessingInfo() *OptimisticProcessingInfo {
+	return app.optimisticProcessingInfo
 }
```
BrandonWeng, line 945: To check for nil

yzang2019 (approved):

PR #739, opened 2023-05-01
Automatically unregister failed contracts in Dex EndBlock

codchen:
If a contract fails execution in dex EndBlock, we want to automatically unregister it so that it doesn't slow down the chain for an indefinite amount of time

unit test

### DIFF
--- a/x/dex/module.go
+++ b/x/dex/module.go
@@ -301,10 +301,7 @@ func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) (ret []abc
 	}
 	for _, c := range validContractsInfoAtBeginning {
 		if _, ok := validContractAddrs[c.ContractAddr]; !ok {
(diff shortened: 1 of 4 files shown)


x/dex/module.go
```diff
@@ -294,6 +295,18 @@ func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) (ret []abc
 			break
 		}
 	}
+	validContractAddrs := map[string]struct{}{}
+	for _, c := range validContractsInfo {
+		validContractAddrs[c.ContractAddr] = struct{}{}
+	}
+	for _, c := range validContractsInfoAtBeginning {
+		if _, ok := validContractAddrs[c.ContractAddr]; !ok {
+			// unregister here also refunds rents, which might be a behavior we want to revisit if spamming becomes bad
```
LCyson: per discussion offchain: we don't want to refund here

PR #743, opened 2023-05-02
do not unregister if out of rent

codchen:


LCyson (changes requested):
is it better to also filter out contracts that have lower rent than a threshold (like 100k usei) in the `EndBlockerAtomic`? so that we don't need to process that contract tx at all. 

Also can we add unit tests and do e2e test for this pr

LCyson (approved):

PR #762, opened 2023-05-05
Update price feeder config to be same as atlantic-2 curr pairs

udpatil:
This updates the currency pairs to be the same as those in atlantic-2


oracle/price-feeder/config.example.toml
```diff
@@ -13,6 +13,36 @@ write_timeout = "20s"
 base = "USDT"
 threshold = "2"
 
+[[currency_pairs]]
```
BrandonWeng, line 16: is this the same cahnge as https://github.com/sei-protocol/sei-chain/pull/709/files? 

We fixed the permission issue btw

BrandonWeng: Do we need this?

 ```
[[provider_endpoints]]
name = "coinbase"
rest = "https://api.exchange.coinbase.com"
websocket = "ws-feed.exchange.coinbase.com"

[[provider_endpoints]]
name = "huobi"
rest = "https://api.huobi.pro"
websocket = "api-aws.huobi.pro"

[[provider_endpoints]]
name = "kraken"
rest = "https://api.kraken.com"
websocket = "ws.kraken.com"```

udpatil: yeah i think so except that one defines some provider endpoints as well?

philipsu522 (approved):

PR #763, opened 2023-05-08
Refactor dex EndBlock to optimize store access

codchen:
`dex` EndBlock would always load all order book entries before processing, even if there is no or very few incoming orders to match against. This PR changes it so that entries are only loaded as needed. The core logic changes are in:
- `types/orders.go`
- `exchange/limit_order.go`
- `exchange/market_order.go`

Migrations are needed due to order book entry key encoding changes, which are necessary to make order book entries sorted in store. There is also a new parameter added.

unit test


x/dex/exchange/limit_order.go
```diff
@@ -14,41 +15,35 @@ func MatchLimitOrders(
...
 		totalExecuted = totalExecuted.Add(executed).Add(executed)
 		totalPrice = totalPrice.Add(
 			executed.Mul(
-				orderbook.Longs.Entries[longPtr].GetPrice().Add(orderbook.Shorts.Entries[shortPtr].GetPrice()),
+				longEntry.GetPrice().Add(shortEntry.GetPrice()),
 			),
 		)
-		minPrice = sdk.MinDec(minPrice, orderbook.Longs.Entries[longPtr].GetPrice())
-		maxPrice = sdk.MaxDec(maxPrice, orderbook.Longs.Entries[longPtr].GetPrice())
+		minPrice = sdk.MinDec(minPrice, longEntry.GetPrice()) // tony: need to revisit whether it's intended to use long entry price here. Keeping it as-is so that test expectations are met
```
codchen, line 32: @LCyson is it intentional here to use long entry price to get minPrice? not sure how minPrice is used with TriggeredOrder

x/dex/exchange/limit_order.go
```diff
@@ -14,41 +15,35 @@ func MatchLimitOrders(
 	settlements := []*types.SettlementEntry{}
 	totalExecuted, totalPrice := sdk.ZeroDec(), sdk.ZeroDec()
 	minPrice, maxPrice := sdk.NewDecFromInt(sdk.NewIntFromUint64(math.MaxInt64)), sdk.OneDec().Neg()
-	longPtr, shortPtr := len(orderbook.Longs.Entries)-1, 0
 
-	for longPtr >= 0 && shortPtr < len(orderbook.Shorts.Entries) && orderbook.Longs.Entries[longPtr].GetPrice().GTE(orderbook.Shorts.Entries[shortPtr].GetPrice()) {
+	for longEntry, shortEntry := orderbook.Longs.Next(ctx), orderbook.Shorts.Next(ctx); longEntry != nil && shortEntry != nil && longEntry.GetPrice().GTE(shortEntry.GetPrice()); longEntry, shortEntry = orderbook.Longs.Next(ctx), orderbook.Shorts.Next(ctx) {
```
BrandonWeng, line 19: nit, can we split this into multiple lines or have a helper method? It's pretty hard to read

x/dex/keeper/long_book.go
```diff
@@ -75,6 +85,57 @@ func (k Keeper) GetAllLongBookForPair(ctx sdk.Context, contractAddr string, pric
...
+		list = append(list, &val)
+		if len(list) == n {
+			break
+		}
+	}
+
+	return
+}
+
+func (k Keeper) GetTopNLongBooksForPairStarting(ctx sdk.Context, contractAddr string, priceDenom string, assetDenom string, n int, startExclusive sdk.Dec) (list []types.OrderBookEntry) {
```
LCyson, line 115: can we add comment here - function description + parameters description

x/dex/keeper/long_book.go
```diff
@@ -75,6 +85,57 @@ func (k Keeper) GetAllLongBookForPair(ctx sdk.Context, contractAddr string, pric
...
+	if n == 0 {
+		return
+	}
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.OrderBookPrefix(true, contractAddr, priceDenom, assetDenom))
+	iterator := sdk.KVStoreReversePrefixIterator(store, []byte{})
+
+	defer iterator.Close()
+
+	// Fast-forward
+	// TODO: add iterator interface that allows starting at a certain subkey under prefix
```
LCyson, line 125: this todo is to add a O(logn) access method to replace the current O(n) way we're doing this correct

x/dex/utils/dec.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"encoding/binary"
+	"math/big"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func DecToBigEndian(d sdk.Dec) (res []byte) {
+	// if d.IsNegative() {
```
LCyson: remove comments

x/dex/utils/dec.go
```diff
@@ -0,0 +1,74 @@
+package utils
+
+import (
+	"encoding/binary"
+	"math/big"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func DecToBigEndian(d sdk.Dec) (res []byte) {
```
LCyson, line 10: is there a lib we can leverage, feel less error-prune in that way

x/dex/utils/dec_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"testing"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/require"
+)
+
+var DECS = []sdk.Dec{
+	sdk.MustNewDecFromStr("90.0"),
```
LCyson, line 11: can we add an overflow dec test case here

x/dex/exchange/limit_order.go
LCyson: hm yea likely an issue here, minPrice should be the lowest short

x/dex/types/orders.go
```diff
@@ -3,49 +3,145 @@ package types
 import (
```
LCyson, line 3: can we add unit tests for new functions in this file

x/dex/types/unit.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+var UnitToMultiplier = map[Unit]sdk.Dec{
+	Unit_STANDARD: sdk.OneDec(),
+	Unit_MILLI:    sdk.NewDec(1_000),
+	Unit_MICRO:    sdk.NewDec(1_000_000),
+	Unit_NANO:     sdk.NewDec(1_000_000_000),
+}
+
+func ConvertDecToStandard(unit Unit, dec sdk.Dec) sdk.Dec {
```
LCyson: if this is only used by test package can we move this to unit_test

x/dex/types/orders.go
```diff
@@ -3,49 +3,145 @@ package types
...
+	var loaded []OrderBookEntry
+	if len(c.CachedEntries) == 0 {
+		loaded = c.loader(ctx, sdk.ZeroDec(), false)
+	} else {
+		loaded = c.loader(ctx, c.CachedEntries[len(c.CachedEntries)-1].GetOrderEntry().Price, true)
+	}
+	c.CachedEntries = append(c.CachedEntries, loaded...)
+}
+
+func (c *CachedSortedOrderBookEntries) SettleQuantity(ctx sdk.Context, quantity sdk.Dec) (res []ToSettle, settled sdk.Dec) {
```
LCyson, line 55: ditto - add comment

x/dex/types/orders.go
```diff
@@ -3,49 +3,145 @@ package types
...
+	newFirstAllocationIdx := 0
+	for idx, a := range currentEntry.Allocations {
+		postSettle := settled.Add(a.Quantity)
+		if postSettle.LTE(quantity) {
+			settled = postSettle
+			res = append(res, AllocationToSettle(a))
+		} else {
+			newFirstAllocationIdx = idx
+			if settled.Equal(quantity) {
+				break
```
LCyson, line 80: we should remove allocations[newFirstAllocationIdx] when all its quantity gets settled here right

x/dex/types/orders.go
```diff
@@ -3,49 +3,145 @@ package types
...
+	Longs    *CachedSortedOrderBookEntries
+	Shorts   *CachedSortedOrderBookEntries
 }
 
 // entries are always sorted by prices in ascending order, regardless of side
 type CachedSortedOrderBookEntries struct {
-	Entries      []OrderBookEntry
-	DirtyEntries *datastructures.TypedSyncMap[string, OrderBookEntry]
+	CachedEntries []OrderBookEntry
+	currentPtr    int
```
LCyson: can we init this in the new() function

LCyson (changes requested):

x/dex/keeper/long_book.go
codchen: added

x/dex/keeper/long_book.go
codchen: yeah

x/dex/utils/dec.go
codchen: not really since `Dec` is a cosmos struct, and cosmos doesn't have an encoding for Dec that preserves numerical ordering

x/dex/utils/dec_test.go
codchen: how to define an overflow dec? I thought dec is supposed to be immune to overflow

x/dex/types/unit.go
codchen: i deleted it since it doesn't seem to be used anywhere

x/dex/types/orders.go
codchen: this is the case where the order is fully settled *before* applying the current allocation. In other words the order book entry isn't exhausted yet - it still at least has this current allocation left

x/dex/types/orders.go
codchen: added

x/dex/types/orders.go
codchen: added

x/dex/exchange/limit_order.go
LCyson: we can leave it as it is for now, we already remove support for TRIGGER ORDER correct

LCyson (approved):

BrandonWeng (approved):

PR #764, opened 2023-05-08
Fix Zero Price Order

Kbhat1:

- Allows Market Price of Zero
- Fixes issue with Orders where zero price (which is allowed) was giving error because it previously wasn't a multiple of tick size

- Tested e2e local chain placing orders + market orders with price 0


LCyson (changes requested):
can we remove the libnitro_replayer and add unit tests

Kbhat1:
> can we remove the libnitro_replayer and add unit tests

Updated

codchen (approved):

LCyson (approved):

PR #773, opened 2023-05-12
modify wasm gas multiplier to make wasm executes more in line with compute cost

udpatil:
This increases the gas used by wasm vm to better represent compute time used

Tested on loadtest clusters, verified that contract upload is not affected as well


cmd/seid/cmd/root.go
```diff
@@ -250,6 +251,9 @@ func newApp(
 		panic(err)
 	}
 
+	wasmGasRegisterConfig := wasmkeeper.DefaultGasRegisterConfig()
+	wasmGasRegisterConfig.GasMultiplier = 8750000
```
philipsu522: can we add a comment explaining what the default is and why we chose this number? 2-3 sentences shoudl be sufficient

philipsu522: also nit - could we do `8_750_000`

udpatil: yup, wil fix

philipsu522 (approved):

PR #781, opened 2023-05-16
Add order limit for price level and pair limit for contracts

codchen:
In order to prevent application-level DDoS attack against `dex` module, we introduce limits for both the number of open orders allowed at a price level for a given asset pair, and the number of asset pairs a contract can have.

The naive implementation would be to query order entries in `msg_server_order_placement` every time an order comes in and get the length of existing allocations, however this could become a performance drag since it would need to load all allocations. This PR implemented it in another way by introducing a new counter state `OrderCount`, which is updated whenever the number of allocations for a price level changes (i.e. when cancellation or placement are handled in EndBlock). Then, in `msg_server_order_placement` it only needs to load this counter which is trivial in terms of latency.

For the number of pair per contract we took the naive implementation since the limit of pairs is much smaller than the limit of orders, and transaction frequency of pair registration is a lot less than that of order placement.

Note that the dependency on sei-cosmos needs to be updated to a stable release once one is cut with the new ACL types.

unit tests as well as semi-integration test in module_test.go


x/dex/keeper/msgserver/msg_server_register_pairs.go
```diff
@@ -15,10 +15,14 @@ func (k msgServer) RegisterPairs(goCtx context.Context, msg *types.MsgRegisterPa
 
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
+	maxPairsPerContract := k.GetMaxPairsPerContract(ctx)
 	events := []sdk.Event{}
 	// Validation such that only the user who stored the code can register pairs
 	for _, batchPair := range msg.Batchcontractpair {
 		contractAddr := batchPair.ContractAddr
+		if uint64(len(k.GetAllRegisteredPairs(ctx, contractAddr))) >= maxPairsPerContract {
+			return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "contract %s already has %d pairs registered", contractAddr, maxPairsPerContract)
```
philipsu522: is this log message correct?

philipsu522 (commented):
do we need to migrate and add this? How do we set existing limits?

x/dex/keeper/order_count.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return err
+	}
+	value := make([]byte, 8)
+	binary.BigEndian.PutUint64(value, count)
+	store.Set(key, value)
+	return nil
+}
+
+func (k Keeper) GetOrderCount(ctx sdk.Context, contractAddr string, priceDenom string, assetDenom string, direction types.PositionDirection, price sdk.Dec) uint64 {
```
Kbhat1: Should we expose a query as well to expose the OrderCount given a contract, pair, direciton and price?

Kbhat1:
> do we need to migrate and add this? How do we set existing limits?

@philipsu522 Default limit values are set here https://github.com/sei-protocol/sei-chain/pull/781/files#diff-550a1ad5f41f704815d0f9b9a67ede8452fae725384c1f8c10291ea31a51bebaR65 (check changes in `x/dex/types/params.go`)

x/dex/contract/whitelist.go
```diff
@@ -20,6 +20,8 @@ var DexWhitelistedKeys = []string{
 	types.SettlementEntryKey,
 	types.NextOrderIDKey,
 	types.MatchResultKey,
+	types.LongOrderCountKey,
```
LCyson, line 23: why do we need OrderCount keys for both long and short if we only define one `max_order_per_price` in param?

go.mod
```diff
@@ -269,7 +269,7 @@ require (
 replace (
 	github.com/CosmWasm/wasmd => github.com/sei-protocol/sei-wasmd v0.0.1
 	github.com/confio/ics23/go => github.com/cosmos/cosmos-sdk/ics23/go v0.8.0
-	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.2.29
+	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.2.29-new-type
```
LCyson: reminder to update this before merge

x/dex/keeper/order_count.go
```diff
@@ -0,0 +1,56 @@
...
+		return 0
+	}
+	value := store.Get(key)
+	if value == nil {
+		return 0
+	}
+	return binary.BigEndian.Uint64(value)
+}
+
+func (k Keeper) DecreaseOrderCount(ctx sdk.Context, contractAddr string, priceDenom string, assetDenom string, direction types.PositionDirection, price sdk.Dec, count uint64) error {
```
LCyson, line 44: nit: why don't we use a generic `UpdateOrderCount` here

x/dex/keeper/order_count_test.go
```diff
@@ -0,0 +1,22 @@
...
+
+func TestGetSetOrderCount(t *testing.T) {
+	keeper, ctx := keepertest.DexKeeper(t)
+	for _, direction := range []types.PositionDirection{
+		types.PositionDirection_LONG,
+		types.PositionDirection_SHORT,
+	} {
+		require.Equal(t, uint64(0), keeper.GetOrderCount(ctx, keepertest.TestContract, keepertest.TestPair.PriceDenom, keepertest.TestPair.AssetDenom, direction, sdk.NewDec(1)))
+		require.Nil(t, keeper.SetOrderCount(ctx, keepertest.TestContract, keepertest.TestPair.PriceDenom, keepertest.TestPair.AssetDenom, direction, sdk.NewDec(1), 5))
+		require.Equal(t, uint64(5), keeper.GetOrderCount(ctx, keepertest.TestContract, keepertest.TestPair.PriceDenom, keepertest.TestPair.AssetDenom, direction, sdk.NewDec(1)))
```
LCyson: can we add tests for `Increase` and `Descrese` count

x/dex/contract/whitelist.go
codchen: since long orders and short orders are stored separately it's easier to just store their counts separately as well. We can add params to specify max orders for long and for short separately but i don't know if they would ever be set differently

x/dex/keeper/order_count.go
codchen: want to keep integer type consistent instead of having to use a signed int64

x/dex/keeper/msgserver/msg_server_register_pairs.go
codchen: fixed, and added migration for params. No need to migrate for the new order counter state since it will be overwritten once order book is changed, and empty keys are handled as 0

x/dex/keeper/order_count_test.go
codchen: added

philipsu522 (approved):

LCyson (approved):

PR #788, opened 2023-05-20
Clear dex memstate cache

BrandonWeng:
We need to clear the dex memcache

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1210,7 +1210,7 @@ func (app *App) ProcessTxs(
 
 	oldDexMemStateCtx := context.WithValue(ctx.Context(), dexutils.DexMemStateContextKey, oldDexMemState)
 	ctx = ctx.WithContext(oldDexMemStateCtx)
-		
+
 	dexMemState := dexutils.GetMemState(ctx.Context())
 	dexMemState.Clear(ctx)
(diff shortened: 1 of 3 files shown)


x/dex/cache/cache.go
```diff
@@ -23,7 +23,7 @@ type MemState struct {
 }
 
 func NewMemState(storeKey sdk.StoreKey) *MemState {
-	contractsToProcess := datastructures.NewSyncSet[string]([]string{})
```
BrandonWeng, line 26: not needed

app/app.go
```diff
@@ -1207,6 +1208,9 @@ func (app *App) ProcessTxs(
 	// Clear the memcache context from the previous state as it failed, we no longer need to commit the data
 	ctx.ContextMemCache().Clear()
 
+	oldDexMemStateCtx := context.WithValue(ctx.Context(), dexutils.DexMemStateContextKey, oldDexMemState)
```
codchen, line 1211: memstate has deep pointers so we need to explicitly clear it like
```
dexMemState := dexutils.GetMemState(ctx.Context())
dexMemState.Clear(ctx)
dexMemState.ClearContractToDependencies()
```

codchen (approved):

PR #827, opened 2023-06-01
Add tkf unit tests for two mints with deferred behavior

udpatil:
This adds unit tests for the scenario known to cause issues for the deferred operation prior to sei-cosmos fix.

Unit Tests

DO NOT MERGE - currently using a sei-cosmos commit hash, needs to be replaced with a tag after merging changes

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -269,7 +269,7 @@ require (
 replace (
 	github.com/CosmWasm/wasmd => github.com/sei-protocol/sei-wasmd v0.0.2
 	github.com/confio/ics23/go => github.com/cosmos/cosmos-sdk/ics23/go v0.8.0
-	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.2.39-0.20230601025259-2f2cf106f1b8
+	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.2.40
 	github.com/cosmos/iavl => github.com/sei-protocol/sei-iavl v0.1.4
 	github.com/cosmos/ibc-go/v3 => github.com/sei-protocol/sei-ibc-go/v3 v3.1.0
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
--- a/go.mod
+++ b/go.mod
@@ -269,7 +269,7 @@ require (
 replace (
 	github.com/CosmWasm/wasmd => github.com/sei-protocol/sei-wasmd v0.0.2
 	github.com/confio/ics23/go => github.com/cosmos/cosmos-sdk/ics23/go v0.8.0
-	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.2.38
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -269,7 +269,7 @@ require (
 replace (
 	github.com/CosmWasm/wasmd => github.com/sei-protocol/sei-wasmd v0.0.2
 	github.com/confio/ics23/go => github.com/cosmos/cosmos-sdk/ics23/go v0.8.0
-	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.2.38
+	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.2.39-0.20230601025259-2f2cf106f1b8
```
Kbhat1: Is this Sei-cosmos change intended?

x/tokenfactory/keeper/admins_test.go
```diff
@@ -8,6 +8,32 @@ import (
 	"github.com/sei-protocol/sei-chain/x/tokenfactory/types"
 )
 
+func (suite *KeeperTestSuite) TestBuggyScenario() {
```
BrandonWeng: Can we rename this to something more explicit around the bug?

x/tokenfactory/keeper/admins_test.go
```diff
@@ -8,6 +8,32 @@ import (
...
+	suite.Require().NoError(err)
+
+	suite.App.BankKeeper.WriteDeferredOperations(suite.Ctx)
+
+	addr0Bal := suite.App.BankKeeper.GetBalance(suite.Ctx, suite.TestAccs[0], suite.defaultDenom).Amount.Int64()
+	suite.Require().Equal(int64(55), addr0Bal, addr0Bal)
+
+	tkModuleBal := suite.App.BankKeeper.GetBalance(suite.Ctx, suite.App.AccountKeeper.GetModuleAddress(types.ModuleName), suite.defaultDenom).Amount.Int64()
+	suite.Require().Equal(int64(0), tkModuleBal, tkModuleBal)
+}
```
BrandonWeng, line 35: For my understanding, without the fix this would fail?

go.mod
udpatil: yes, we need the change from sei cosmos for the fix to deferred sends. I left a DO NOT MERGE note in the pr header to update this with an actual tag prior to merging once https://github.com/sei-protocol/sei-cosmos/pull/261 is merged.

x/tokenfactory/keeper/admins_test.go
udpatil: yeah sorry, will do, had just named it while trying to repro the issue and did it hastily

x/tokenfactory/keeper/admins_test.go
udpatil: correct, previously it would fail on the second mint statement (eg. never even make it to the 55 assertion)

philipsu522: thanks for adding this

philipsu522 (approved):

PR #829, opened 2023-06-01
Filter out dex contracts that do not need order matching

codchen:
https://linear.app/seilabs/issue/SEI-4326/mod-01-or-contracts-that-do-not-need-order-matching-could-be-filtered

no behavior change really since we removed hook support, all registered contracts would be for order matching


LCyson (changes requested):
can we update the unit test for this

codchen:
@LCyson added

LCyson (approved):

PR #836, opened 2023-06-02
Add integration test for place order and time-locked token contract

yzang2019:
**Goal:**
Use github workflow + docker for CI integration test

**Change:**
As the first step, we add a simple test for testing dex module including:
1. Register a contract
2. Place an order
3. Verify order is placed successfully

We also added the verification and integration test for timelocked token contract and split into 3 test files.

Run and tested locally and verified it works


philipsu522 (approved):

.github/workflows/integration-test.yml
```diff
@@ -38,4 +40,16 @@ jobs:
...
       - name: Verify Sei Chain is able to start up
-        run: ./docker/integration_test/startup_test.sh
+        run: docker exec -i sei-node-0 integration_test/startup/startup_test.sh
+
+      - name: Testing Dex Module
+        run: docker exec -i sei-node-0 integration_test/dex_module/place_order_test.sh
+
+      - name: Testing Wasm Module
+        run: |
+          docker exec -i sei-node-0 integration_test/contracts/deploy_timelocked_token_contract.sh
```
yzang2019, line 50: We will initiate new contracts before each test scenario

Eko101089 (commented):
Why this ?

codchen (approved):
are the mars etc. wasm builds intentionally included in the commit?

yzang2019:
> are the mars etc. wasm builds intentionally included in the commit?

The PR was initially for dex module testing, so mars were added for that. Then I repurpose it to also include wasm module

yzang2019:
> Why this ?

Why what?

philipsu522 (commented):
- Let's change https://github.com/sei-protocol/sei-chain/pull/836/files#diff-396c970b862a5f8668dd8016364b8c5b58b942dc47846ce01e250e5df7954908L19 to be 2 to be consistent
- do we want to check in the wasm files? or have them in s3 or something

yzang2019:
> * Let's change https://github.com/sei-protocol/sei-chain/pull/836/files#diff-396c970b862a5f8668dd8016364b8c5b58b942dc47846ce01e250e5df7954908L19 to be 2 to be consistent
> * do we want to check in the wasm files? or have them in s3 or something

It's already on s3, but put it here for convenience, is there any risk adding these files?

PR #844, opened 2023-06-05
Update Tokenfactory Readme

Kbhat1:
- Update Tokenfactory Readme
- Adds more descriptions to each message
- Adds state modifications for each message
- Adds example

### DIFF
--- a/x/dex/keeper/contract.go
+++ b/x/dex/keeper/contract.go
@@ -2,8 +2,10 @@ package keeper
 
 import (
 	"errors"
+	"time"
 
 	"github.com/cosmos/cosmos-sdk/store/prefix"
+	"github.com/cosmos/cosmos-sdk/telemetry"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	appparams "github.com/sei-protocol/sei-chain/app/params"
 	"github.com/sei-protocol/sei-chain/utils"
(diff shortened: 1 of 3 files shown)


x/tokenfactory/README.md
```diff
@@ -58,3 +95,77 @@ message MsgChangeAdmin {
...
+A token factory token's denom is: `factory/{creator address}/{subdenom}`
+
+Splitting up into sub-components, this has:
+
+- `len(factory) = 7`
+- `2 * len("/") = 2`
+- `len(longest_subdenom)`
+- `len(creator_address) = len(bech32(longest_addr_length, chain_addr_prefix))`.
+
+Longest addr length at the moment is `32 bytes`. Due to SDK error correction
```
philipsu522, line 161: is this enforced when trying to create a tokenfactory denom?

x/tokenfactory/README.md
```diff
@@ -58,3 +95,77 @@ message MsgChangeAdmin {
...
+So please don't make code rely on these max lengths for parsing.
+
+# Examples
+To create a new token, use the create-denom command from the tokenfactory module. The following example uses the address sei166vhptur29s3gw5qr6dm30s06gej6pr4n6qc4l from mylocalwallet as the default admin for the new token.
+
+## Creating a token
+To create a new token we can use the create-denom command.
+
+```sh
+seid tx tokenfactory create-denom ufoo --from mylocalwallet
```
philipsu522: do we need fees to do this?

Kbhat1: @philipsu522 Purposely left out the fees and gas (and other flags) to focus on just the components that matter here: the command + the admin who is creating the denom (mylocalwallet) whose address will be used in the full tokenfactory denom

x/tokenfactory/README.md
Kbhat1: Max subdenom length, and max creator length are enforced here https://github.com/sei-protocol/sei-chain/blob/master/x/tokenfactory/types/denoms.go . The longest addr length is used to to compute the max creator length. 

Btw, I updated this section such that it just lists out the length restrictions for subdenom and creator. I added an appendix where the detailed derivations are included.

philipsu522 (approved):

PR #850, opened 2023-06-07
Add testing harness for e2e application logic tests

codchen:
Add a testing harness to simplify writing application-level e2e test. The goal is to allow testing the full application write path that roots from `FinalizeBlock` and `Commit`, since more granular tests like ones on keeper level lack the capability to uncover bugs caused by improper interactions between modules or invalid assumptions across transactions and Begin/Mid/EndBlock logics.  Specifically the harness includes:
- A wrapper around the core Sei `App` object, which provides simple interfaces for processing a block and transaction signing. [`testutils/processblock/common.go`]
- A set of helpers to define the "genesis" state of a test. [`testutils/processblock/genesis*.go`]
- A set of helpers to create messages to test with. [`testutils/processblock/msgs/*.go`]
- A set of verifiers to assert test outcome. [`testutils/processblock/verify/*.go`]

A test written using this harness typically includes three phases:
1. Initialization - create the test app wrapper object and call the genesis helper to initialize the app state
2. Msg Creation - create messages to be tested, group messages into transactions, and sign the transactions
3. Run and Verify - group transactions into blocks and run the blocks through the app wrapper, and verify state using the verifiers.
`tests/template_test.go` provided a template for writing such tests, and `tests/dex_test.go` is a concrete examples.

Next steps:
- Add message helpers for other core functionalities
- Add verifiers for other core functionalities
- Add tests for other core functionalities

tests

### DIFF
--- a/app/test_helpers.go
+++ b/app/test_helpers.go
@@ -59,7 +59,7 @@ func NewTestWrapper(t *testing.T, tm time.Time, valPub crptotypes.PubKey) *TestW
 		Ctx: ctx,
 	}
 	wrapper.SetT(t)
-	wrapper.setupValidator(stakingtypes.Bonded, valPub)
+	wrapper.setupValidator(stakingtypes.Unbonded, valPub)
 	return wrapper
 }
 


testutil/processblock/verify/bank.go
```diff
@@ -0,0 +1,84 @@
+package verify
```
BrandonWeng, line 1: Some things that might help with verification are the implicit actions that are preformed each block like:

- epoch hooks (mint events)
- fee/delegation distribution
- slashing/jailing
-

tests/dex_test.go
```diff
@@ -0,0 +1,39 @@
...
+	aliceLimitOrder := market.LongLimitOrder(alice, "10.5", "5")
+	msgs := []sdk.Msg{aliceLimitOrder}
+	tx := app.Sign(alice, msgs, 10000)
+
+	block2 := []signing.Tx{tx}
+	blockRunner := func() []uint32 { return app.RunBlock(block2) } // 2nd block to place order
+	blockRunner = verify.DexOrders(t, app, blockRunner, block2)
+	blockRunner = verify.Balance(t, app, blockRunner, block2)
+
+	require.Equal(t, []uint32{0}, blockRunner())
```
BrandonWeng: So when we need to apply verify methods we need to create a blockRunner variable?

codchen: yeah it's basically a wrapper around `RunBlock` but with verifier chained together:
```
verifierA_load_pre_execution_state
    verifierB_load_pre_execution_state
        RunBlock
    verifierB_check_post_execution_state
verifierA_check_post_execution_state
```

testutil/processblock/verify/bank.go
codchen: Added verifiers and tests for epoch/mint/distribution. Slashing/jailing is a bit more complicated so I plan to add it in a separate PR

tests/template_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/auth/signing"
+	"github.com/sei-protocol/sei-chain/testutil/processblock"
+	"github.com/sei-protocol/sei-chain/testutil/processblock/msgs"
+	"github.com/sei-protocol/sei-chain/testutil/processblock/verify"
+	"github.com/stretchr/testify/require"
+)
+
+func TestTemplate(t *testing.T) {
```
yzang2019: Is there an easy way to modify the template so that it could support the best practice golang table driven testing norm such as:

 ```
type TestCase {
              description   string
              input              Input
              verifier           func
              expectedResult Result
}

for _, scenario := range []TestCase{
              {
                     description: "test case1",
                     input: Input{},
                     verifier: func(){},
                     expectedResult: []uint32{0,0},
              },
              {
                     description: "test case2",
                     input: Input{},
                     verifier: func(){},
                     expectedResult: []uint32{0,1},
              },
}
{
              t.Run(scenario.description, func(t *testing.T) {
                     ...
              })
}
```

File: tests/dex_test.go [hunk 4]
```diff
@@ -0,0 +1,72 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/auth/signing"
+	"github.com/sei-protocol/sei-chain/testutil/processblock"
+	"github.com/sei-protocol/sei-chain/testutil/processblock/msgs"
+	"github.com/sei-protocol/sei-chain/testutil/processblock/verify"
+	"github.com/stretchr/testify/require"
+)
+
+func TestOrders(t *testing.T) {
```
yzang2019: can we have a more specific name for the Test function instead of calling TestOrders? Basically I would like to know what are we testing for orders? Which specific logic are we verifying here, let's break it down into smaller pieces, same thing applied to other modules

testutil/processblock/verify/bank.go
BrandonWeng: Sounds good thank you!

tests/template_test.go
codchen: @yzang2019 it should be doable, but not sure if it'd be very readable given the amount of different setup needed for different tests. I'll give it a try

yzang2019 (approved):

PR #852, opened 2023-06-07
Decrease consensus timeout backoff values

philipsu522:
Tunes consensus configs so that we don't have high p95 block times on a slow validator:
- UnsafeProposeTimeoutOverride: let's validators vote "nil" if a proposal doesn't come in in time: 1s -> 300ms, shoudl be enough time for 1 network rtt
- UnsafeProposeTimeoutDeltaOverride: How much time to increase the propose timeout every time we enter a new round
- UnsafeVoteTimeoutDelta: How much time to incrase the vote timeout every time we enter a new round

Launched a cluster and slowed a node down.
Before:
image
image

After:
image
image


yzang2019 (approved):
nice

PR #853, opened 2023-06-07
ignore pb files for codecov

BrandonWeng:


codecov.yml
```diff
@@ -3,4 +3,5 @@ coverage:
     project:
       default:
         target: 80%
-        threshold: 3%
\ No newline at end of file
+ignore:
+  - "**/*_pb.go"         # Auto-generated files
```
yzang2019, line 7: Why do we ignore pb files? I thought there were needed for proto

BrandonWeng: This is just for the code coverage, they're auto generated and don't come with unit tests

yzang2019: https://github.com/cosmos/cosmos-sdk/tree/main/types I think open source also keep pb files

PR #859, opened 2023-06-08
Run parallel unit tests when possible and update codecov

BrandonWeng:

### DIFF
--- a/.github/workflows/unit-test-coverage.yml
+++ b/.github/workflows/unit-test-coverage.yml
@@ -1,9 +1,6 @@
-name: Go
+name: Unit Test Coverage
 
-
-on:
-  pull_request:
-    branches: [ master ]
+on: [push, pull_request]
 
 defaults:
  run:
@@ -29,7 +26,7 @@ jobs:
         go-version: 1.18
 
     - name: Run coverage
-      run: go test -v ./... -coverprofile=coverage.txt -covermode=atomic
-    
+      run: go test -v ./... -coverprofile=coverage.txt -covermode=atomic -parallel 4
+
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/unit_tests.yml
```diff
@@ -6,6 +6,7 @@ on:
       - "**.go"
     branches:
       - main
+      - master
```
BrandonWeng, line 9: We haven't been running the unit tests on master branch

Makefile
```diff
@@ -197,4 +197,4 @@ $(BUILDDIR)/packages.txt:$(GO_TEST_FILES) $(BUILDDIR)
 split-test-packages:$(BUILDDIR)/packages.txt
 	split -d -n l/$(NUM_SPLIT) $< $<.
 test-group-%:split-test-packages
-	cat $(BUILDDIR)/packages.txt.$* | xargs go test -mod=readonly -timeout=10m -race -coverprofile=$(BUILDDIR)/$*.profile.out
+	cat $(BUILDDIR)/packages.txt.$* | xargs go test -parallel 4 -mod=readonly -timeout=10m -race -coverprofile=$(BUILDDIR)/$*.profile.out
```
BrandonWeng, line 200: Some unit tests have parallel enabled, this is just so that they can run in parallel

BrandonWeng: Technically should speed up our test runtime too

.github/workflows/unit-test-coverage.yml
```diff
@@ -1,9 +1,12 @@
...
-
+name: Unit Test Coverage
 
 on:
   pull_request:
-    branches: [ master ]
+  push:
+    branches:
+      - master
+      - release/**
```
BrandonWeng, line 8: I think the issue is that we're not running on the master branch (base commit that PRs are forked from)

image

philipsu522 (approved):

yzang2019 (approved):

PR #860, opened 2023-06-08
Update unit-test-coverage.yml

BrandonWeng:

### DIFF
--- a/.github/workflows/unit-test-coverage.yml
+++ b/.github/workflows/unit-test-coverage.yml
@@ -21,9 +21,6 @@ jobs:
     permissions: write-all
     steps:
     - uses: actions/checkout@v3
-      with:
-        fetch-depth: 0
-        ref: ${{ github.event.pull_request && 'refs/pull/${{ github.event.pull_request.number }}/merge' }}
     - uses: actions/setup-python@v3
 
     - name: Set up Go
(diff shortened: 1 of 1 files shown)


.github/workflows/unit-test-coverage.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     - uses: actions/checkout@v3
       with:
         fetch-depth: 0
-        ref: "refs/pull/${{ github.event.number }}/merge"
+        ref: ${{ github.event.pull_request && 'refs/pull/${{ github.event.pull_request.number }}/merge' }}
```
BrandonWeng: Doon't need to check out if it's not a PR

.github/workflows/unit-test-coverage.yml
```diff
@@ -21,9 +21,6 @@ jobs:
     permissions: write-all
     steps:
     - uses: actions/checkout@v3
-      with:
-        fetch-depth: 0
-        ref: "refs/pull/${{ github.event.number }}/merge"
```
BrandonWeng, line 26: Copying what the other jobs are doing, idt we need to specify this. Causes more issues if we want to run on the master branch (push)

yzang2019 (approved):

PR #867, opened 2023-06-09
remove deferred send usages from dex and tk

udpatil:
This removes deferred send usages from tokenfactory and dex. This increases overall safety because now only feecollector uses deferred balances.

tested on localsei

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1308,6 +1308,10 @@ func (app *App) ProcessBlock(ctx sdk.Context, txs [][]byte, req BlockProcessRequ
 	prioritizedResults, ctx := app.BuildDependenciesAndRunTxs(ctx, prioritizedTxs)
 	txResults = append(txResults, prioritizedResults...)
 
+	// Finalize all Bank Module Transfers here so that events are included for prioritiezd txs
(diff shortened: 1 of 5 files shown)


app/app.go
```diff
@@ -1308,6 +1308,10 @@ func (app *App) ProcessBlock(ctx sdk.Context, txs [][]byte, req BlockProcessRequ
 	prioritizedResults, ctx := app.BuildDependenciesAndRunTxs(ctx, prioritizedTxs)
 	txResults = append(txResults, prioritizedResults...)
 
+	// Finalize all Bank Module Transfers here so that events are included for prioritiezd txs
```
udpatil, line 1311: note that we need this because SOME prioritized transactions can pay fees (oracle doesn't but some of the dex ones do) so we need to make sure to flush those BEFORE we create the new context memcache for the rest of the transactions cc @BrandonWeng

BrandonWeng (approved):

app/app.go
BrandonWeng: If possible can we add a test to enforce this?

PR #875, opened 2023-06-13
Parameterize gasless cancel whitelist

codchen:
We might want to whitelist certain LPs so that they can place cancellations without gas. Having it hardcoded would require an upgrade which is not ideal.

unit test


philipsu522 (commented):
Do you mind adding migration logic?

codchen:
@philipsu522 added

PR #876, opened 2023-06-14
Refactor deferred usages

udpatil:
This PR does the following:
- refactors ante dependency decorators to follow the new pattern of passing txIndex through
- Adds the new deferredcache resource type to the tree
- removes the ContextCache (because of incoming deprecation) setup performed in BuildDependenciesAndRunTxs
- updates instantiation of seid app to use the new BankKeeper with DeferredCache memstore created

local sei testing (and needs more tests added via test harness)

### DIFF
--- a/app/antedecorators/accesscontrol_wasm_dependency.go
+++ b/app/antedecorators/accesscontrol_wasm_dependency.go
@@ -54,7 +54,7 @@ func (ad ACLWasmDependencyDecorator) SenderMatchesContractAdmin(ctx sdk.Context,
 	return contractInfo.Admin == msg.FromAddress, nil
 }
 
-func (ad ACLWasmDependencyDecorator) AnteDeps(txDeps []sdkacltypes.AccessOperation, tx sdk.Tx, next sdk.AnteDepGenerator) (newTxDeps []sdkacltypes.AccessOperation, err error) {
+func (ad ACLWasmDependencyDecorator) AnteDeps(txDeps []sdkacltypes.AccessOperation, tx sdk.Tx, txIndex int, next sdk.AnteDepGenerator) (newTxDeps []sdkacltypes.AccessOperation, err error) {
 	deps := []sdkacltypes.AccessOperation{}
 
 	for _, msg := range tx.GetMsgs() {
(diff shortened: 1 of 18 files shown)


app/antedecorators/accesscontrol_wasm_dependency.go
```diff
@@ -54,7 +54,7 @@ func (ad ACLWasmDependencyDecorator) SenderMatchesContractAdmin(ctx sdk.Context,
 	return contractInfo.Admin == msg.FromAddress, nil
 }
 
-func (ad ACLWasmDependencyDecorator) AnteDeps(txDeps []sdkacltypes.AccessOperation, tx sdk.Tx, next sdk.AnteDepGenerator) (newTxDeps []sdkacltypes.AccessOperation, err error) {
+func (ad ACLWasmDependencyDecorator) AnteDeps(txDeps []sdkacltypes.AccessOperation, tx sdk.Tx, txIndex int, next sdk.AnteDepGenerator) (newTxDeps []sdkacltypes.AccessOperation, err error) {
```
alexanderbez, line 57: Why are these called `AnteDeps`?

udpatil: to be similarly named to `AnteHandler` which handles the stuff prior to running messages, `AnteDeps` is meant o represent `AnteDependencyGenerator` to generate dependencies prior to message dependencies

tests/dex_test.go
```diff
@@ -98,3 +98,22 @@ func TestPlaceOrders(t *testing.T) {
 		testCase.run(t, app)
 	}
 }
+
+func TestPlaceImposterOrders(t *testing.T) {
+	app := processblock.NewTestApp()
+	p := processblock.DexPreset(app, 3, 2)
+	p.DoRegisterMarkets(app)
+
+	for _, testCase := range []TestCase{
```
yzang2019, line 107: Can we add a unit test for sending multiple bank transactions in a same block to test out the defer behavior?

BrandonWeng: +1 ideally execute them in parallel on a shared cache if possible

BrandonWeng (approved):
overall seems fine but +1 on more testing, either with a LT cluster or with integration tests to verify the supply invariant

PR #878, opened 2023-06-14
Fix denom key collision

codchen:
We need to change how we serialize denom-based keys to prevent collisions as described in https://linear.app/seilabs/issue/SEI-4587/key-02-or-potential-key-collision-if-one-denom-is-a-prefix-of-other

unit test/load cluster


x/dex/migrations/v10_to_v11.go
```diff
@@ -10,7 +10,7 @@ import (
 var DexPrefixes = []string{
 	types.LongBookKey,
 	types.ShortBookKey,
-	types.TriggerBookKey,
+	"TriggerBook-value-",
```
philipsu522, line 13: are these changes necessary?

philipsu522 (approved):

Kbhat1 (approved):

x/dex/migrations/v10_to_v11.go
codchen: yeah we want to remove TriggerOrder related stuff since it's been causing confusion to the audit team (and future readers)

PR #883, opened 2023-06-14
3.0.4 release

philipsu522:


x/tokenfactory/keeper/migrations.go
```diff
@@ -43,16 +43,19 @@ func (m Migrator) Migrate2to3(ctx sdk.Context) error {
 
 func (m Migrator) Migrate3to4(ctx sdk.Context) error {
 	// Set denom metadata for all denoms
-	m.keeper.bankKeeper.IterateTotalSupply(ctx, func(coin sdk.Coin) bool {
-		fmt.Printf("Migration denom: %s\n", coin.Denom)
-		if denomMetadata, err := m.keeper.bankKeeper.GetDenomMetaData(ctx, coin.Denom); !err {
-			panic(fmt.Errorf("denom %s does not exist", coin.Denom))
+	iter := m.keeper.GetAllDenomsIterator(ctx)
```
philipsu522, line 46: the issue is that this does not cover denoms in which there is no supply for. So now instead of iterating over supply, we get all denoms and then set the data (regardless of if there is any supply)

CHANGELOG.md
```diff
@@ -27,14 +27,29 @@ Ref: https://keepachangelog.com/en/1.0.0/
 -->
 
 # Changelog
+## 3.0.4
+sei-chain:
+* [#874] (https://github.com/sei-protocol/sei-chain/pull/874) Charge rent after Sudo call
```
BrandonWeng:
```suggestion
* [#874] (https://github.com/sei-protocol/sei-chain/pull/874) Charge rent after failed Sudo call
```

BrandonWeng (approved):

PR #886, opened 2023-06-15
Add validation for negative price in cancellation

codchen:
We do not want negative price to be allowed even in cancellation

unit test


LCyson (changes requested):
can we fix the integration test before merge

PR #894, opened 2023-06-16
fix: handle panics in epoch hooks

alexanderbez:
Gracefully recover any x/epoch hook, both before and after, and log the recovered error.

ref: sei-4591


x/epoch/module.go
alexanderbez, line 1: added line breaks to improve readability of block

philipsu522 (approved):

PR #897, opened 2023-06-19
Environment spin up in tests using Kurtosis

victorcolombo:

This PR shows how Kurtosis can be used to spin up a SEI Chain that is later used in integration tests


.github/workflows/integration-test.yml
```diff
@@ -35,30 +35,32 @@ jobs:
...
-          do
-            sleep 10
-          done
-          sleep 10
+      - name: Build docker image
+        run: make clean && make build-docker-node
+      
+      - name: Run cluster with 4 nodes
+        run: >
+            kurtosis run github.com/kurtosis-tech/sei-package@b130bd658f9bf58e6410e93ace46398435df0d1c '{"cluster_size": 4, "num_accounts": 10, "invariant_check_interval": 10, "image": "sei-chain/localnode", "git_url": "${{ github.server_url }}/${{ github.repository }}/", "git_ref": "${{ github.head_ref}}"}' --enclave sei-integration-test
```
yzang2019: How do we update the image in the future?

victorcolombo: Hey @yzang2019, this is a Starlark script with generic code to spin up a SEI blockchain. The code currently lives here https://github.com/kurtosis-tech/sei-package. We are more than happy to give you and team maintainer rights to that repo

victorcolombo (commented):
I see the integration test is not happy that its not finding my branch name in this repo since I'm trying to merge from a fork. Taking a look!

victorcolombo:
Should also work on PRs coming from forks now

.github/workflows/integration-test.yml
yzang2019: Hmm that seems not very convenient, we might need to constantly change the config on our local machine and launch the local docker based on that. We might also frequently merge some config changes to build the docker cluster. Is there a way to make it easier for us to make changes without releasing a new image or making changes to Kurtosis repo?

victorcolombo: @yzang2019 There are two pieces of spinning an environment:
- A `Starlark package`, that describes how the environment should be spin up, connected together, description of its health pre-conditions and possibly integration tests. They are intended to be a general formula. They live in a repository, so a very big change in the logic on how chainlink nodes are spin up requires code changes and a repo push.
- Packages are parametrizable, that is, you can change the shape of the environment you spin up without touching any code, only by changing the parameters passed to the package. For example, in this PR:
`'{"cluster_size": 4, "num_accounts": 10, "invariant_check_interval": 10, "image": "sei-chain/localnode", "git_url": "${{ env.GIT_URL }}", "git_ref": "${{ github.head_ref}}"}'`
You can pick to change the cluster size, number of account, image, etc without changing the package code, because it takes those as inputs.

A great parallel here is our ETH package, that allows very big customization via parameters (choosing clients, topologies, fork/genesis, etc): https://github.com/kurtosis-tech/eth-network-package

Can you share some details on what is the config you are concerned about?

yzang2019: For example, we need to frequently change the genesis step and tune some parameters: https://github.com/sei-protocol/sei-chain/blob/master/docker/localnode/scripts/step2_genesis.sh#L16

And we might also add more tunable configs in the future

yzang2019: Is there an easy way to expose them as a config file or something so that we don't need to rebuild an image? The whole purpose of using docker is to make local testing easier

victorcolombo: It is definitely possibly to parametrize the genesis overrides so you can "inject" custom behavior when interacting with Kurtosis via command line or CI. Here is a concrete example to illustrate https://github.com/kurtosis-tech/sei-package/pull/8/files

yzang2019: I see, is there any instruction for how to build and release a new image if we want to make some changes to the logic of building up and genesis a chain

yzang2019:
@philipsu522 @BrandonWeng Please take a look at this as well when you have cycles, and let me know if there are any other concerns

.github/workflows/integration-test.yml
victorcolombo: A starlark script is not an image like Docker's where you need to build and push to a specific repository. A starlark script is interpreted by Kurtosis, and fetched directly from Github. There is no need for ahead of time compilation. So to update the logic it is as easy as pushing the change to the repo. A couple of references:

https://docs.kurtosis.com/explanations/why-kurtosis-starlark
https://docs.kurtosis.com/concepts-reference/starlark

yzang2019: Oh interesting, so basically we need to update Kurtosis repro if we want to make changes to the chain?

leeederek: If you're referring to the logic of building up and using a genesis state to start your chain, then yes! Kurtosis packages are written in Starlark, a dialect of Python and we would be available to support and guide you on how to write / edit the package. We will also grant maintainer permissions to you and your team so you can make your own changes at your leisure. 

Does that make sense @yzang2019?

philipsu522:

PR #901, opened 2023-06-20
Add OP metrics

philipsu522:
Add metrics for when OP runs vs when it doesn't
http://grafana.sei-loadtest-testnet.seinetwork.io:9090/graph?g0.expr=sum(sei_optimistic_processing_counter)%20by%20(enabled)&g0.tab=0&g0.stacked=0&g0.show_exemplars=0&g0.range_input=1h


app/app.go
```diff
@@ -970,6 +970,7 @@ func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcess
 	} else if !bytes.Equal(app.optimisticProcessingInfo.Hash, req.Hash) {
 		app.optimisticProcessingInfo.Aborted = true
 	}
+	metrics.SetOptimisticProcessingCounter(!app.optimisticProcessingInfo.Aborted)
```
yzang2019: why here? instead of in FinalizeBlocker

yzang2019: Also shouldn't we increment the counter each time optimistic processing is elligible? Why setting a counter ?

udpatil: looks like the helper DOES increment a counter under the hood, maybe better to rename the helper to something like `IncrementOptimisticProcessingCounter`. Metric location also depends on whether we want to track the number of times optimistic proicessing is attempted vs succeeded. Since in finalize blocker we branch based on that, so if we want to track each metric separately then finalizeblocker is probably a better place to track

yzang2019: Yeah, I think tracking optimistic processing success has more values, having both is even better

udpatil (approved):

yzang2019 (approved):

yzang2019 (approved):
LGTM

PR #902, opened 2023-06-20
Oracle integration tests

udpatil:
This adds oracle integration tests that sets feeder and performs a vote to verify behavior end to end.

integration tests


integration_test/oracle_module/set_feeder_test.yaml
```diff
@@ -0,0 +1,28 @@
...
+    # send bank funds to new feeder to populate account
+    - cmd: printf "12345678\n" | seid tx bank send admin $NEW_FEEDER_ADDR 1sei -b block -y --fees 2000usei --output json | jq -r ".code"
+    # set feeder
+    - cmd: printf "12345678\n" | seid tx oracle set-feeder $NEW_FEEDER_ADDR -b block -y --from node_admin --fees 2000usei --output json | jq -r ".code"
+      env: SET_FEEDER_ERROR_CODE
+    # do a vote and ensure code 0
+    - cmd: printf "12345678\n" | seid tx oracle aggregate-vote 1.5ueth $ADMIN_VAL_ADDR -b block -y --from $NEW_FEEDER_ADDR --gas 0 --output json | jq -r ".code"
+      env: NEW_FEEDER_VOTE_ERROR_CODE
+    # set old feeder addr again
+    - cmd: printf "12345678\n" | seid tx oracle set-feeder $OLD_FEEDER_ADDR -b block -y --from node_admin --fees 2000usei --output json | jq -r ".code"
```
BrandonWeng, line 24: Can we add verification for the succuess/error count rate too? It was broken on master branch for a couple of weeks. we caught it as a partner reported it

udpatil: wdym it was broken? what was broken about it? do you have additional context into this? if youre referring to vote penalties its a little hard to test exactly on those since those are updated every block (vote window) and are easier to cover the test cases via unit testing for oracle mid/end block logic.

yzang2019: yeah +1 to verify the signing success

.github/workflows/integration-test.yml
```diff
@@ -50,6 +50,12 @@ jobs:
               "python3 integration_test/scripts/runner.py integration_test/mint_module/mint_test.yaml"
             ]
           },
+          {
+            name: "Gov Module",
+            scripts: [
+              "python3 integration_test/scripts/runner.py integration_test/gov_module/gov_proposal_test.yaml"
+            ]
+          },
         ]
```
yzang2019, line 72: We need to add Oracle Module

integration_test/oracle_module/set_feeder_test.yaml
yzang2019: We can just use seid to query the success and make sure it's above a value after certain blocks or amount of time

.github/workflows/integration-test.yml
udpatil: already on it 👍

integration_test/oracle_module/set_feeder_test.yaml
udpatil: for that, I can add a separate integration test that just asserts those values since the feeders are already running, would that be ok?

cc @BrandonWeng for more context into the error case we previosuly saw?

yzang2019: btw, I think we can add a TODO to switch docker to use the sei-chain example price feeder config, instead of the one that we copied over to docker folder

philipsu522 (approved):

integration_test/oracle_module/verify_penalty_counts.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - cmd: echo $PENALTY_JSON | jq -r ".miss_count"
+      env: MISS_COUNT
+    # calculate total
+    - cmd: expr $SUCCESS_COUNT + $ABSTAIN_COUNT + $MISS_COUNT
+      env: TOTAL_COUNT
+    # calculate success rate
+    - cmd: expr $(expr $SUCCESS_COUNT \* 100 ) / $TOTAL_COUNT
+      env: SUCCESS_PERCENT
+  verifiers:
+    # verify total counts GEQ expected: not asserting equal due to block movement between evaluation
```
yzang2019, line 40: Can we add a sleep and then make sure TOTAL_COUNT > EXPECTED_COUNTS?

udpatil: why do we need to add a sleep? the expected counts is determined by the current block heights (and we'd want the total count to be exactly equal IF no block progress between the queries, but since they could progress we check for GEQ in cas a block or two have passed)

PR #903, opened 2023-06-20
[CI/CD] Staking Integration Tests

BrandonWeng:


integration_test/staking_module/staking_test.yaml
```diff
@@ -0,0 +1,28 @@
+- name: Staking a validator
```
yzang2019: nit: We can use a more meaningful test names like Test Staking a validator should xxx

yzang2019 (approved):

yzang2019 (approved):

PR #913, opened 2023-06-22
Deduct fees from gasless tx

codchen:
We expect certain gasless transactions (e.g. dex order placements) to carry fees even if the amount is not subject to the validator's minimum fee check, but we have not been collecting those fees because DeductFeeChecker (i.e. the wrapped decorators in the gasless decorator) are always skipped for gasless transactions. This not only causes less validator income but also exposes ddos attack vector specifically in the dex module. This PR changes it so that fees are always deducted for transactions in DeliverTx mode.

unit tests

### DIFF
--- a/app/antedecorators/gasless.go
+++ b/app/antedecorators/gasless.go
@@ -37,24 +37,39 @@ func (gd GaslessDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool,
 	if err != nil {
 		return ctx, err
 	}
-	if gas > 0 || !isGasless {
-		ctx = ctx.WithGasMeter(originalGasMeter)
-		// if not gasless, then we use the wrappers
-
-		// AnteHandle always takes a `next` so we need a no-op to execute only one handler at a time
-		terminatorHandler := func(ctx sdk.Context, _ sdk.Tx, _ bool) (sdk.Context, error) {
-			return ctx, nil
+	isDeliverTx := !ctx.IsCheckTx() && !ctx.IsReCheckTx() && !simulate
+	shouldCheckTxSkipFeeDeduct := gas == 0 && isGasless // whether the tx should be subject to min fee check AND priority assignment. Applicable to CheckTx only.
+	if isDeliverTx || !shouldCheckTxSkipFeeDeduct {
+		// In the case of deliverTx, we want to deduct fees regardless of whether the tx is considered gasless or not, since
+		// gasless txs will be subject to application-specific fee requirements in later stage of ante, for which the payment
+		// of those app-specific fees happens here. Note that the minimum fee check in the wrapped deduct fee handler is only
+		// performed if the context is for CheckTx, so the check will be skipped for deliverTx and the deduct fee handler will
+		// only deduct fee without checking.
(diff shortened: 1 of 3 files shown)


app/antedecorators/gasless.go
```diff
@@ -37,24 +37,39 @@ func (gd GaslessDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool,
...
+		// performed if the context is for CheckTx, so the check will be skipped for deliverTx and the deduct fee handler will
+		// only deduct fee without checking.
+		// Otherwise (i.e. in the case of checkTx), we only want to perform fee checks and fee deduction if the tx is not considered
+		// gasless, or if it specifies a non-zero gas limit even if it is considered gasless, so that the wrapped deduct fee
+		// handler will assign an appropriate priority to it.
+		if !isGasless {
+			ctx = ctx.WithGasMeter(originalGasMeter)
 		}
-		return next(ctx, tx, simulate)
+		return gd.handleWrapped(ctx, tx, simulate, next)
```
udpatil, line 54: we should verify that this doesn't cause regression for like oracle for example. It looks like it should be ok, but just want to verify. Also, isnt there still a gap here where dex txs could just send gas==0 and they would pass `isGasLess` and then skip the deduct fee branch?

codchen: For gas==0 it's expected for such txs to skip deduct fee branch in CheckTx since otherwise it would fail the min gas fee check, and in DeliverTx this branch won't be skipped (which is used only for fee deduction reason, not check) because of the first condition.

Re oracle tx I'm not super familiar with the current behavior but I don't think this would cause oracle behavior change if oracle txs do not contain fees (the only intended delta in behavior from this PR is to collect fees (but not check) of gasless txs). What would be a good way to verify it for oracle in your mind? @udpatil

udpatil: Yeah but for delivery tx, wouldn't it still be able to bypass the !isGasless check for a dex tx and then the dex tx wouldn't have the gas meter (essentially become gasless)?

udpatil: Re: oracle it should be sufficient to have the integration tests passing so we know oracle txs still succeed without setting any gas or fees

udpatil (approved):

BrandonWeng (approved):
Seems like votes are still getting in for oracle based on the integration test but it's still failing 

```
Input : seid q oracle vote-penalty-counter $ADMIN_VAL_ADDR --height $BLOCK_HEIGHT --output json | jq -r ".vote_penalty_counter"
Output: {
  "miss_count": "0",
  "abstain_count": "20",
  "success_count": "2"
}
```

codchen:
> Seems like votes are still getting in for oracle based on the integration test but it's still failing
> 
> ```
> Input : seid q oracle vote-penalty-counter $ADMIN_VAL_ADDR --height $BLOCK_HEIGHT --output json | jq -r ".vote_penalty_counter"
> Output: {
>   "miss_count": "0",
>   "abstain_count": "20",
>   "success_count": "2"
> }
> ```

seems like a flaky run? succeeded on rerun

PR #927, opened 2023-06-24
Fix flaky oracle integration test

yzang2019:
Total can also be progress+1 since we have a window of 2 blocks


integration_test/oracle_module/verify_penalty_counts.yaml
```diff
@@ -39,7 +39,7 @@
   verifiers:
     # verify total counts GEQ expected: not asserting equal due to block movement between evaluation
     - type: eval
-      expr: TOTAL_COUNT == EXPECTED_COUNTS
+      expr: ( TOTAL_COUNT == EXPECTED_COUNTS ) or ( TOTAL_COUNT == EXPECTED_COUNTS + 1 )
```
udpatil: On line 22 we specifically query the same height to ensure that the sum should match the expected. Are we doing math wrong there, or is there a different cause for off by one here?

yzang2019: The reason is that the height when we capture could be either even or odd, so when it divided by 2, it could be flaky. This PR should solve the problem

udpatil (approved):

PR #930, opened 2023-06-26
Add authz module to sei-chain

udpatil:
Add authz module to sei chain

localsei testing and integration tests

### DIFF
--- a/.github/workflows/compatiblity_check.yml
+++ b/.github/workflows/compatiblity_check.yml
@@ -3,7 +3,7 @@ on:
   pull_request:
   push:
     branches:
-      - master
+      - main
 
 jobs:
   forward-compatibility:
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -5,7 +5,6 @@ on:
     paths:
       - "**.go"
     branches:
-      - master
       - main
       - release/**
 
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -6,6 +6,7 @@ on:
       - "**.go"
     branches:
       - master
+      - main
       - release/**
 
 jobs:
--- a/app/app.go
+++ b/app/app.go
@@ -893,6 +893,15 @@ func (app *App) SetStoreUpgradeHandlers() {
 		// configure store loader that checks if version == upgradeHeight and applies store upgrades
 		app.SetStoreLoader(upgradetypes.UpgradeStoreLoader(upgradeInfo.Height, &storeUpgrades))
 	}
+	// TODO: add upgrade name for authz module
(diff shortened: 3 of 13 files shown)


yzang2019 (approved):
Nice, thanks for adding the integration tests!

app/app.go
```diff
@@ -880,6 +890,15 @@ func (app *App) SetStoreUpgradeHandlers() {
...
 		}
 
+		// configure store loader that checks if version == upgradeHeight and applies store upgrades
+		app.SetStoreLoader(upgradetypes.UpgradeStoreLoader(upgradeInfo.Height, &storeUpgrades))
+	}
+	// TODO: add upgrade name for authz module
+	if false && upgradeInfo.Name == "TODO" && !app.UpgradeKeeper.IsSkipHeight(upgradeInfo.Height) {
+		storeUpgrades := storetypes.StoreUpgrades{
+			Added: []string{authzkeeper.StoreKey},
+		}
```
BrandonWeng, line 900: Why does this have to be a TODO?

.github/workflows/compatiblity_check.yml
```diff
@@ -3,7 +3,7 @@ on:
   pull_request:
   push:
     branches:
-      - master
+      - main
```
BrandonWeng, line 6: nice

.github/workflows/unit_tests.yml
```diff
@@ -6,6 +6,7 @@ on:
       - "**.go"
     branches:
       - master
+      - main
       - release/**
```
BrandonWeng:
```suggestion
    branches:
      - main
      - release/**
```

app/app.go
udpatil: Because we don't have an upgrade version yet for which we want to add in the wrote upgrade (hence the `false` at the beginning of branch) We will need to update this when we create the next upgrade version release

udpatil: I can make a task for psu to make sure it doesnt get missed

BrandonWeng: Gotcha so in this case it would be something like 3.0.6 or 3.05

app/app.go
```diff
@@ -370,7 +375,7 @@ func New(
 	bApp.SetInterfaceRegistry(interfaceRegistry)
 
 	keys := sdk.NewKVStoreKeys(
-		acltypes.StoreKey, authtypes.StoreKey, banktypes.StoreKey, stakingtypes.StoreKey,
+		acltypes.StoreKey, authtypes.StoreKey, authzkeeper.StoreKey, banktypes.StoreKey, stakingtypes.StoreKey,
```
philipsu522, line 378: do we need to set beginblock/endblock ordering?
https://github.com/sei-protocol/sei-chain/blob/main/app/app.go#L631
https://github.com/sei-protocol/sei-chain/blob/main/app/app.go#L666

udpatil: We are doing so further down in this file

PR #948, opened 2023-06-30
feat: crypto.com provider

alexanderbez:
## Changelog

* Implement crypto.com provider
* Import and use `WebsocketController` which is used by the `CryptoProvider`

Note, this was lifted from Umee's version of the price-feeder @ v3.x. When we have sufficient capacity, we may want to explore the generalized improvements both the Kujira and Umee versions have made, such as generalizing the provider implementation.

### DIFF
--- a/oracle/price-feeder/config.example.toml
+++ b/oracle/price-feeder/config.example.toml
@@ -35,7 +35,8 @@ quote = "USD"
 
 [[currency_pairs]]
 base = "SOL"
-chain_denom = "usol"
+# this is because we need some dummy price feed that moves for usei
+chain_denom = "usei"
 providers = [
   "huobi",
   "kraken",
--- a/oracle/price-feeder/config.example.toml
+++ b/oracle/price-feeder/config.example.toml
@@ -35,11 +35,12 @@ quote = "USD"
 
 [[currency_pairs]]
 base = "SOL"
-chain_denom = "usei"
+chain_denom = "usol"
 providers = [
   "huobi",
   "kraken",
   "coinbase",
+  "crypto",
 ]
 quote = "USD"
 
@@ -50,6 +51,7 @@ providers = [
   "huobi",
   "kraken",
   "coinbase",
+  "crypto",
 ]
 quote = "USD"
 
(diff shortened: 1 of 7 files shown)


oracle/price-feeder/oracle/provider/crypto.go
```diff
@@ -0,0 +1,432 @@
...
+	}
+)
+
+func NewCryptoProvider(
+	ctx context.Context,
+	logger zerolog.Logger,
+	endpoint config.ProviderEndpoint,
+	pairs ...types.CurrencyPair,
+) (*CryptoProvider, error) {
+	if endpoint.Name != config.ProviderCrypto {
```
philipsu522, line 111: shoudl we error here? what is the reasoning behind setting it to crypto.com if it was misconfigured?

yzang2019 (commented):
Can you modify config.example.toml to add crypto as well?

philipsu522 (approved):

oracle/price-feeder/oracle/provider/crypto.go
alexanderbez: So this is how all the other providers are implemented, along with the crypto.com one ported from umee. Although, I do agree it should error instead.

If we do that, we should do it for all providers and create an issue for it :)

alexanderbez:
Updated @yzang2019 :)

oracle/price-feeder/config.example.toml
```diff
@@ -35,11 +35,12 @@ quote = "USD"
 
 [[currency_pairs]]
 base = "SOL"
-chain_denom = "usei"
+chain_denom = "usol"
```
udpatil: This is usei intentionally to have some moving price feed available on Atlantic 2 for usei

udpatil (changes requested):

oracle/price-feeder/config.example.toml
udpatil: sorry @alexanderbez forgot to update this comment, can we make this change before merging?

udpatil (approved):

PR #951, opened 2023-06-30
Add hard fork manager

udpatil:
This adds the hard fork manager pattern to allow for registering hard forks based on height and chain ID

Unit tests

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -346,6 +346,8 @@ type App struct {
 	CheckTxMemState         *dexcache.MemState
 	ProcessProposalMemState *dexcache.MemState
 	MemState                *dexcache.MemState
+
+	HardForkManager *HardForkManager
 }
 
 // New returns a reference to an initialized blockchain app
@@ -779,6 +781,11 @@ func New(
 	app.SetInitChainer(app.InitChainer)
 	app.SetBeginBlocker(app.BeginBlocker)
 
+	// register hard fork handlers with the hard fork manager
+	app.HardForkManager = NewHardForkManager(app.ChainID)
+	// register and hard fork handlers below
+	// example: app.HardForkManager.RegisterHandler(myHandler)
+
 	signModeHandler := encodingConfig.TxConfig.SignModeHandler()
 	// app.batchVerifier = ante.NewSR25519BatchVerifier(app.AccountKeeper, signModeHandler)
 
@@ -913,6 +920,10 @@ func (app App) GetBaseApp() *baseapp.BaseApp { return app.BaseApp }
 // BeginBlocker application updates every begin block
 func (app *App) BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock) abci.ResponseBeginBlock {
 	metrics.GaugeSeidVersionAndCommit(app.versionInfo.Version, app.versionInfo.GitCommit)
+	// check if we've reached a target height, if so, execute any applicable handlers
+	if app.HardForkManager.TargetHeightReached(ctx) {
+		app.HardForkManager.ExecuteForTargetHeight(ctx)
+	}
 	return app.mm.BeginBlock(ctx, req)
 }
 
--- a/app/fork_manager.go
+++ b/app/fork_manager.go
@@ -1,18 +1,27 @@
 package app
 
 import (
+	"fmt"
+
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/sei-protocol/goutils"
 )
 
 type HardForkHandler interface {
+	// a unique identifying name to ensure no duplicate handlers are registered
+	GetName() string
+	// The target chain ID for which to run this handler (which chain to run on)
 	GetTargetChainID() string
+	// The target height at which the handler should be executed
 	GetTargetHeight() int64
+	// An execution function used to process a hard fork handler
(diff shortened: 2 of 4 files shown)


app/app.go
```diff
@@ -913,6 +920,10 @@
 // BeginBlocker application updates every begin block
 func (app *App) BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock) abci.ResponseBeginBlock {
 	metrics.GaugeSeidVersionAndCommit(app.versionInfo.Version, app.versionInfo.GitCommit)
+	// check if we've reached a target height, if so, execute any applicable handlers
+	if app.HardForkManager.TargetHeightReached(ctx) {
+		app.HardForkManager.ExecuteForTargetHeight(ctx)
```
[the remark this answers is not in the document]
udpatil: how do we feel about this behavior, I think it _should_ be okay given that this path is only expected to execute in special circumstances where we are trying to do a special migration, so chain liveness is likely already compromised at this point. cc @yzang2019 @philipsu522 for thoughts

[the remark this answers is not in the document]
philipsu522: This should be ok

app/fork_manager.go
```diff
@@ -0,0 +1,65 @@
...
+// This tries to register a handler with the hard fork manager.
+// If the handler's target chain ID doesn't match the chain ID of the manager,
+// this is a no-op and the handler is ignored.
+func (hfm *HardForkManager) RegisterHandler(handler HardForkHandler) {
+	if handler.GetTargetChainID() != hfm.chainID {
+		return
+	}
+	targetHeight := handler.GetTargetHeight()
+	if handlers, ok := hfm.handlerMap[targetHeight]; ok {
+		// TODO: refactor to use goutils InPlaceAppend once it's merged
```
philipsu522: It's merged https://github.com/sei-protocol/sei-chain/pull/940/files#diff-55eacffe7baaf57057beee673c143b0699a077a382a14c94321e440d54e1575bR14

app/fork_manager.go
```diff
@@ -0,0 +1,65 @@
...
+		handlers = append(handlers, handler)
+		hfm.handlerMap[targetHeight] = handlers
+	} else {
+		hfm.handlerMap[targetHeight] = []HardForkHandler{handler}
```
philipsu522, line 57: is there a way to verify if targetHeight >= currHeight?

app/fork_manager.go
```diff
@@ -0,0 +1,65 @@
...
+// This returns a boolean indicating whether or not the current height is a
+// target height for which there are hard fork handlers to run.
+func (hfm *HardForkManager) TargetHeightReached(ctx sdk.Context) bool {
+	_, ok := hfm.handlerMap[ctx.BlockHeight()]
+	return ok
+}
+
+// This executes the hard fork handlers for the current height. This will function will panic upon receiving an error during hard fork handler execution
+func (hfm *HardForkManager) ExecuteForTargetHeight(ctx sdk.Context) {
+	handlers, ok := hfm.handlerMap[ctx.BlockHeight()]
```
philipsu522, line 70: Could we add info logs too? so we can monitor that it's running

philipsu522 (approved):
some minor questions / fix requests, but overall lgtm

app/fork_manager.go
```diff
@@ -0,0 +1,65 @@
...
+// If the handler's target chain ID doesn't match the chain ID of the manager,
+// this is a no-op and the handler is ignored.
+func (hfm *HardForkManager) RegisterHandler(handler HardForkHandler) {
+	if handler.GetTargetChainID() != hfm.chainID {
+		return
+	}
+	targetHeight := handler.GetTargetHeight()
+	if handlers, ok := hfm.handlerMap[targetHeight]; ok {
+		// TODO: refactor to use goutils InPlaceAppend once it's merged
+		handlers = append(handlers, handler)
```
alexanderbez: I wonder if we should enforce handlers to have a name `Name() string` as to ensure we don't register duplicate handlers or something?

alexanderbez (approved):
Excellent work @udpatil -- I like the idea

app/fork_manager.go
udpatil: yeah, theres a separate linear task for this, but I can add it to this PR

app/fork_manager.go
udpatil: yeah, I think this is reasonable, should be pretty easy to add in. would we want the name to be unique regardless of target height though? I would think yes because the premise would be to have a name correspond to a specific remediation. Therefore if you want to do a different remediation at some different height, you would want the name to still be unique. I'll add a `Name` function to the interface

app/fork_manager.go
udpatil: I'm not certain, because this is performed at the chain startup in app.go so there isn't necessarily a `ctx` with a chain height present. is it an issue if old height handlers are stored in this map? we would expect the scale of this map to remain small, so IMO having historical ones shouldnt have negative effects

philipsu522: Nope, its just for more protection in case we accidentally put in a smaller height. No worries if not possible

philipsu522 (approved):

PR #961, opened 2023-07-06
Use default for pruning

philipsu522:

### DIFF
--- a/cmd/seid/cmd/root.go
+++ b/cmd/seid/cmd/root.go
@@ -2,14 +2,6 @@ package cmd
 
 import (
 	"errors"
-	"fmt"
-	"io"
-	"math"
-	"math/rand"
-	"os"
-	"path/filepath"
-	"time"
-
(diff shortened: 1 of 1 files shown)


cmd/seid/cmd/root.go
```diff
@@ -394,16 +390,7 @@ func initAppConfig() (string, interface{}) {
 	srvCfg.API.Enable = true
 
 	// Pruning configs
-	srvCfg.Pruning = "custom"
-	// With block times of 0.3 seconds, this gives us 3 days worth of blocks to store (in case of outage)
-	srvCfg.PruningKeepRecent = "864000"
-	// Randomly generate pruning interval. We want the following properties:
-	//   - random: if everyone has the same value, the block that everyone prunes will be slow
-	//   - prime: no overlap
-	primes := getPrimeNums(2500, 4000)
```
yzang2019, line 403: Maybe we should keep this setting but use a lower prime numbers

PR #963, opened 2023-07-07
Add SetMetadata to tokenfactory

LCyson:
This PR allows the admin of a tokenfactory denom to set its bank.Metadata, thereby changing its display units, name, and symbol.

- unit tests
- e2e test


Kbhat1:
@LCyson can we add a cli command for set denom metadata in tokenfactory

LCyson:
> @LCyson can we add a cli command for set denom metadata in tokenfactory

Metadata request contains too many fields for cli command, users will be recommended to send rpc request to update

Kbhat1:
> > @LCyson can we add a cli command for set denom metadata in tokenfactory
> 
> Metadata request contains too many fields for cli command, users will be recommended to send rpc request to update

You don’t need to specify all arguments directly via cli. Instead we can read in a json file which specifies the full metadata. Take a look at how we do this for register pair tx

PR #971, opened 2023-07-11
Update check job to prevent merge

BrandonWeng:


.github/workflows/integration-test.yml
```diff
@@ -134,7 +134,22 @@ jobs:
   integration-test-check:
     name: Integration Test Check
     runs-on: ubuntu-latest
-    needs: integration-tests # This job will only run if all 'integration-tests' jobs have completed successfully
+    needs: integration-tests
+    if: always()
```
BrandonWeng, line 138: Runs after the test, but always runs

yzang2019 (approved):

PR #974, opened 2023-07-12
feat: demo hard fork handler

alexanderbez:
## Changelog

* Move fork handler to sub-package of app
* Some minor lint cleanup
* Implement a demo (named v0)


app/upgrades/v0/upgrade.go
```diff
@@ -0,0 +1,74 @@
...
+func (h HardForkUpgradeHandler) migrateGringotts(ctx sdk.Context, govKeeper *wasmkeeper.PermissionedKeeper) error {
+	var (
+		contractAddr sdk.AccAddress
+		newCodeID    uint64
+		msg          []byte
+	)
+
+	switch h.TargetChainID {
+	case upgrades.ChainIDSeiHardForkTest:
+		// TODO: ...
```
udpatil: do we need to implement anything in this TODO? (i dont think so but just wondering)

udpatil (approved):

app/upgrades/v0/upgrade.go
alexanderbez: Yes -- we need to set the address and code ID, which I do not know :)

PR #980, opened 2023-07-19
feat: integrate memiavl

alexanderbez:
Since Sei's fork is based off of SDK v0.45.x mainline and memIAVL only supports v0.46.x+, we cannot directly use it -- we have to fork/copy it in directly.

As a result, there are various API changes that needed to be made and I'm not 100% confident they are correct, but I believe they should be. Major changes are around `CacheWrap` usage and `Commit` taking a boolean arg.

Benchmarking and testing is crucial here.

Reviewers can omit reviewing `memiavl` package and should primarily focus on changes in `store` (although you can review both if you'd like)


alexanderbez:
After speaking with @yzang2019, we've decided to close this out and move the work into a separate repo allowing easier integration, testing and benchmarking.

PR #989, opened 2023-07-26
Add CLI/wasmbinding to set tokenfactory metadata

codchen:
The CLI allow passing in a JSON file, and the wasmbinding takes in a json blob with the same format

local sei

### DIFF
--- a/x/tokenfactory/client/cli/tx.go
+++ b/x/tokenfactory/client/cli/tx.go
@@ -2,13 +2,17 @@ package cli
 
 import (
 	"fmt"
+	"io/ioutil"
+	"strings"
 
 	"github.com/spf13/cobra"
 
 	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/client/flags"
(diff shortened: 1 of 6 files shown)


x/tokenfactory/client/cli/tx.go
```diff
@@ -152,3 +156,86 @@ func NewChangeAdminCmd() *cobra.Command {
...
 	return cmd
 }
+
+func NewSetDenomMetadataCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "set-denom-metadata [metadata-file] [flags]",
+		Short: "Set metadata for a factory-created denom. Must have admin authority to do so.",
+		Long: strings.TrimSpace(
+			`Submit a parameter proposal along with an initial deposit.
+The proposal details must be supplied via a JSON file. For values that contains
```
Kbhat1: Is there a reason why we are referencing a proposal here?

LCyson (approved):

x/tokenfactory/client/cli/tx.go
codchen: updated

PR #998, opened 2023-07-26
Fix go mod

yzang2019:


docker/localnode/config/config.toml
```diff
@@ -429,3 +429,7 @@ max-open-connections = 3
 
 # Instrumentation namespace
 namespace = "tendermint"
+
+[memiavl]
+
+enable = false
```
yzang2019: When testing, we just need to change it to true

PR #1000, opened 2023-07-27
Remove need for add-genesis-account

BrandonWeng:
Remove the need to call add-genesis-account, its a redundant call taht's prone to error since it there's a contention on the genesis file 

This should also speed it up significantly 

Created 10k accounts locally 

image

Created chain locally
image

### DIFF
--- a/loadtest/scripts/populate_genesis_accounts.py
+++ b/loadtest/scripts/populate_genesis_accounts.py
@@ -7,7 +7,7 @@ import time
 
 from pathlib import Path
 
-PARALLEISM=32
+PARALLEISM=64
 
 # Global Variable used for accounts
 # Does not need to be thread safe, each thread should only be writing to its own index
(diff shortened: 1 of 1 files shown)


loadtest/scripts/populate_genesis_accounts.py
```diff
@@ -46,33 +41,14 @@ def add_account(account_name, address, mnemonic, local=False):
...
-            sleep_time += 0.5
-            time.sleep(sleep_time)
-
-    if retry_counter >= 1000:
-        exit(-1)
+    dump_account_data_to_file(
+        account_name=account_name,
+        address=address,
+        mnemonic=mnemonic,
+    )
```
BrandonWeng, line 51: image

loadtest/scripts/populate_genesis_accounts.py
```diff
@@ -46,33 +41,14 @@ def add_account(account_name, address, mnemonic, local=False):
...
-            )
-            break
-        except subprocess.CalledProcessError as e:
-            print(f"Encountered error {e}, retried {retry_counter} times")
-            retry_counter += 1
-            sleep_time += 0.5
-            time.sleep(sleep_time)
-
-    if retry_counter >= 1000:
-        exit(-1)
```
BrandonWeng, line 75: We're already calling add keys and manually adding a balance entry in the bank module so we don't actually need to call this. Its just a wrapper

philipsu522 (approved):

yzang2019 (approved):

PR #1004, opened 2023-08-01
Update Names to Get Tokenfactory Queries

Kbhat1:
- Update tokenfactory wasm queries to use the same name as keeper fn => GetDenomAuthorityMetadata vs DenomAuthorityMetadata and GetDenomsFromCreator vs DenomsFromCreator


Kbhat1:
Deciding to close since this is cosmetic  + requires chain upgrade

PR #1018, opened 2023-08-07
Reorder tx results into absolute order

udpatil:
This reorders the TXResults such that the true order of tx indices is preserves as opposed to ordering parititoned and then other TXs which was occurring previously
unit tests + will test patch on chain

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1246,8 +1246,8 @@ func (app *App) ProcessTxs(
 	return txResults, ctx
 }
 
-func (app *App) PartitionPrioritizedTxs(ctx sdk.Context, txs [][]byte) (prioritizedTxs, otherTxs [][]byte) {
-	for _, tx := range txs {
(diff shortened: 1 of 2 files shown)


app/app.go
```diff
@@ -1334,13 +1336,16 @@ func (app *App) ProcessBlock(ctx sdk.Context, txs [][]byte, req BlockProcessRequ
 	beginBlockResp := app.BeginBlock(ctx, beginBlockReq)
 	events = append(events, beginBlockResp.Events...)
 
-	var txResults []*abci.ExecTxResult
+	txResults := make([]*abci.ExecTxResult, len(txs))
 
-	prioritizedTxs, txs := app.PartitionPrioritizedTxs(ctx, txs)
+	// TODO: need to recombine txresults based on ORIGINAL ordering
```
Kbhat1: Nit: can we remove this TODO? Or add "recombine txresults based on original ordering" below as a comment

udpatil: yeah good point, i had left that there prior to making the fix

PR #1037, opened 2023-08-29
Fix init chain script for Linux

stevenlanders:
- Mac uses a BSD variety of `sed` whereas Linux is on GNU version
- On Mac, `sed` requires `sed -i ''` whereas on linux it requires `sed -i` (no `''`)
- This change checks `$OSTYPE` before running `sed` to replace config entries for mode & indexer

- Local testing of initializing the local chain


udpatil (approved):

yzang2019 (approved):
LGTM

PR #1039, opened 2023-08-30
Filter oracle price broadcast by whitelist values

stevenlanders:
- Filters all prices to only whitelisted entries so that configuring extra prices do not cause a penalty
- This allows us to safely add new oracle prices to the network (add first, then adjust requirement)
- Added a test for tick(), which relies on OracleClient.  
- Instead of refactoring all of OracleClient to be an interface, I used a mock function.  I prefer interfaces, but this was a safer blast radius on the codebase.

- Unit Tests for Filtering logic
- Unit Tests for tick() logic
- local price-feeder verification

## Local testing output (via price-feeder)
There is now a new DEBUG line including pre-filtered values (see uextra)
```
DBG pre-filtered prices exchange_rates=27136.481901959936667054ubtc,1699.246734346635898342ueth,0.999870004013299023uextra,1.000000000000000000uusdc module=oracle
```
This is followed by the INFO line for the broadcast
```
INF Going to broadcast vote exchange_rates=27136.481901285386982230ubtc,1699.248149670818160415ueth,1.000000000000000000uusdc feeder=sei1spcyd7ausz5z79wjplhlwmeccu9rtsn549y3e4 module=oracle validator=seivaloper1spcyd7ausz5z79wjplhlwmeccu9rtsn5t24lc9 vote_period=427
```
Note that the broadcast does not include uextra (because it was filtered out)

### DIFF
--- a/oracle/price-feeder/oracle/oracle.go
+++ b/oracle/price-feeder/oracle/oracle.go
@@ -51,7 +51,7 @@ type Oracle struct {
 }
 
 // createMappingsFromPairs is a helper function to initialize maps from currencyPairs
-// this is used to by test cases to initialize the oracle lcient
+// this is used to by test cases to initialize the oracle client
 func createMappingsFromPairs(currencyPairs []config.CurrencyPair) (
 	map[string]string,
 	map[string][]types.CurrencyPair) {
--- a/oracle/price-feeder/oracle/oracle.go
+++ b/oracle/price-feeder/oracle/oracle.go
@@ -46,6 +46,23 @@ type Oracle struct {
 	paramCache      ParamCache
 	jailCache       JailCache
 	healthchecks    map[string]http.Client
+	mockSetPrices   func(ctx context.Context) error
+}
+
+func createMappingsFromPairs(currencyPairs []config.CurrencyPair) (map[string]string, map[string][]types.CurrencyPair) {
+	chainDenomMapping := make(map[string]string)
+	providerPairs := make(map[string][]types.CurrencyPair)
+
+	for _, pair := range currencyPairs {
+		for _, p := range pair.Providers {
+			providerPairs[p] = append(providerPairs[p], types.CurrencyPair{
+				Base:  pair.Base,
+				Quote: pair.Quote,
+			})
+		}
+		chainDenomMapping[pair.Base] = pair.ChainDenom
+	}
+	return chainDenomMapping, providerPairs
 }
 
 func New(
@@ -57,18 +74,8 @@ func New(
 	endpoints map[string]config.ProviderEndpoint,
 	healthchecksConfig []config.Healthchecks,
 ) *Oracle {
-	providerPairs := make(map[string][]types.CurrencyPair)
-	chainDenomMapping := make(map[string]string)
 
-	for _, pair := range currencyPairs {
-		for _, provider := range pair.Providers {
-			providerPairs[provider] = append(providerPairs[provider], types.CurrencyPair{
-				Base:  pair.Base,
-				Quote: pair.Quote,
-			})
-		}
-		chainDenomMapping[pair.Base] = pair.ChainDenom
-	}
+	chainDenomMapping, providerPairs := createMappingsFromPairs(currencyPairs)
 
 	healthchecks := make(map[string]http.Client)
 	for _, healthcheck := range healthchecksConfig {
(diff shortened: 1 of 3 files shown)


oracle/price-feeder/oracle/oracle.go
```diff
@@ -497,6 +512,20 @@ func (o *Oracle) checkWhitelist(params oracletypes.Params) {
 	}
 }
 
+// filterPricesByDenomList takes a list of DecCoins and filters out any
+// coins that are not in the provided DenomList.
+func filterPricesByDenomList(dc sdk.DecCoins, dl oracletypes.DenomList) sdk.DecCoins {
```
udpatil: nit, can we rename these from `dc` and `dl` to like `coinPrices` and `denomList` or something more verbose / clear

oracle/price-feeder/oracle/oracle_test.go
```diff
@@ -350,6 +353,218 @@ func (ots *OracleTestSuite) TestPrices() {
...
+	return sdk.ValAddress(pubKey.Address().Bytes()).String()
+}
+
+func generateAcctAddr() string {
+	privKey := ed25519.GenPrivKey()
+	pubKey := privKey.PubKey()
+	return sdk.AccAddress(pubKey.Address().Bytes()).String()
+}
+
+func TestTick_Scenarios(t *testing.T) {
```
udpatil: nit, test naming is primarily camelcase, we can just call this `TestTickScenarios`

oracle/price-feeder/oracle/oracle_test.go
```diff
@@ -350,6 +353,218 @@ func (ots *OracleTestSuite) TestPrices() {
...
+				},
+				mockSetPrices: func(ctx context.Context) error {
+					setPriceCount++
+					return nil
+				},
+				chainDenomMapping: cdm,
+				prices:            test.prices,
+				paramCache: ParamCache{
+					params: &oracletypes.Params{
+						Whitelist: test.whitelist,
```
udpatil: does voteWindow need to be nonzero here (not sure if it would cause problems but would be nice to just set it in case future test modifications need it)

oracle/price-feeder/oracle/oracle_test.go
```diff
@@ -350,6 +353,218 @@ func (ots *OracleTestSuite) TestPrices() {
...
+	pubKey := privKey.PubKey()
+	return sdk.AccAddress(pubKey.Address().Bytes()).String()
+}
+
+func TestTick_Scenarios(t *testing.T) {
+	// generate address so that Bech32 address is valid
+	validatorAddr := generateValidatorAddr()
+	feederAddr := generateAcctAddr()
+
+	tests := map[string]struct {
```
udpatil: nit, why are we structuring this as a `map[string]struct{}` vs the later test case of `[]struct{}`? slight preference for the latter because we don't actually use the keys to access the map, simply iterate over the items

oracle/price-feeder/oracle/oracle_test.go
```diff
@@ -350,6 +353,218 @@ func (ots *OracleTestSuite) TestPrices() {
...
+				// ensure functions were actually called
+				require.Equal(t, 1, broadcastCount)
+				require.Equal(t, 1, setPriceCount)
+			}
+		})
+	}
+}
+
+func TestFilterPricesWithDenomList(t *testing.T) {
+	tests := []struct {
```
udpatil, line 533: see above comment about map[string]struct{}, i slightly prefer this array structure better

oracle/price-feeder/oracle/oracle.go
```diff
@@ -44,6 +45,27 @@ type Oracle struct {
 	paramCache      ParamCache
 	jailCache       JailCache
 	healthchecks    map[string]http.Client
+	mockSetPrices   func(ctx context.Context) error
+}
+
+// createMappingsFromPairs is a helper function to initialize maps from currencyPairs
+// this is used to by test cases to initialize the oracle lcient
```
yzang2019: Nit: oracle client*

udpatil (approved):
lgtm,  lets make sure to verify behavior on localsei as well prior to merging

oracle/price-feeder/oracle/oracle.go
```diff
@@ -548,7 +577,11 @@ func (o *Oracle) tick(
 		return err
 	}
 
-	exchangeRatesStr := GenerateExchangeRatesString(o.GetPrices())
+	prices := o.GetPrices()
+
+	// filter for whitelisted denominations so that extra oracle prices are not penalized
+	filteredPrices := filterPricesByDenomList(prices, oracleParams.Whitelist)
```
yzang2019, line 602: Where is this whitelist coming from?

stevenlanders: It's part of the oracle params structure:
```
$ seid query oracle params
params:
  lookback_duration: "3600"
  min_valid_per_window: "0.050000000000000000"
  reward_band: "0.020000000000000000"
  slash_fraction: "0.000000000000000000"
  slash_window: "216000"
  vote_period: "2"
  vote_threshold: "0.667000000000000000"
  whitelist:
  - name: ueth
  - name: ubtc
  - name: uusdc
  - name: uusdt
```

yzang2019 (approved):

udpatil:
We have tested this filtering flow using governance to update whitelist as well, and it works as expected, we're good to merge, LGTM

PR #1051, opened 2023-09-16
Only whitelisted price feeds should be required

stevenlanders:
**Background**
- The `SetPrices()` method enforces that all price feeds are present rather than only whitelisted entries
- If a non-whitelisted entry fails, it means that it would abstain the vote, which is undesirable
  
**Change Summary**
- This change only requires whitelisted entries, and all others will populate if successful
- It also removes a length check to allow for more entries (it already checks that every required value is present)

- Unit Tests (mocking "bad providers")
  - The existing tests now confirm a non-whitelisted entry continues to be set
  - A new test confirms a failing non-whitelisted entry doesn't prevent the setting of prices
- Local Chain Testing

### DIFF
--- a/oracle/price-feeder/oracle/oracle.go
+++ b/oracle/price-feeder/oracle/oracle.go
@@ -40,7 +40,6 @@ type Oracle struct {
 	oracleClient       client.OracleClient
 	deviations         map[string]sdk.Dec
 	endpoints          map[string]config.ProviderEndpoint
-	params             oracletypes.Params
 
 	mtx             sync.RWMutex
 	lastPriceSyncTS time.Time
--- a/oracle/price-feeder/oracle/oracle.go
+++ b/oracle/price-feeder/oracle/oracle.go
@@ -189,7 +189,7 @@ func (o *Oracle) GetPrices() sdk.DecCoins {
 // to determine prices. If candles are not available, uses the most recent prices
 // with VWAP. Warns the user of any missing prices, and filters out any faulty
 // providers which do not report prices or candles within 2𝜎 of the others.
-func (o *Oracle) SetPrices(ctx context.Context) error {
+func (o *Oracle) SetPrices(ctx context.Context, whitelist oracletypes.DenomList) error {
 	if o.mockSetPrices != nil {
 		return o.mockSetPrices(ctx)
 	}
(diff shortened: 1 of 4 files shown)


oracle/price-feeder/oracle/oracle.go
```diff
@@ -189,7 +189,7 @@ func (o *Oracle) GetPrices() sdk.DecCoins {
 // to determine prices. If candles are not available, uses the most recent prices
 // with VWAP. Warns the user of any missing prices, and filters out any faulty
 // providers which do not report prices or candles within 2𝜎 of the others.
-func (o *Oracle) SetPrices(ctx context.Context) error {
+func (o *Oracle) SetPrices(ctx context.Context, whitelist oracletypes.DenomList) error {
```
udpatil: maybe it would be better to store this `whitelist` in the Oracle struct? and then when we re-fetch params we can just update that?

udpatil (commented):
overall looks good, just one suggestion around the persisting of the whitelist

oracle/price-feeder/oracle/oracle.go
```diff
@@ -277,10 +279,6 @@ func (o *Oracle) SetPrices(ctx context.Context) error {
 		return err
 	}
 
-	// TODO: make this more lenient to allow assigning prices even when unable to retrieve all
```
udpatil, line 280: doesnt the removal of this mean that if we have all but one whitelisted assets available, it would still send the remaining whitelisted assets without the one that isnt available?

oracle/price-feeder/oracle/oracle.go
```diff
@@ -277,10 +279,6 @@ func (o *Oracle) SetPrices(ctx context.Context) error {
...
-	if len(computedPrices) != len(requiredRates) {
-		return fmt.Errorf("unable to get prices for all exchange candles")
-	}
 	for base := range requiredRates {
 		if _, ok := computedPrices[base]; !ok {
```
udpatil, line 283: or is it because we still have this check, it would still require all the whitelisted items to be present?

stevenlanders: you got it :)

oracle/price-feeder/oracle/oracle.go
stevenlanders: Now that we don't necessarily require the NON-whitelisted entries, it means that computedPrices may not have the same length, which is now legal.  The subsequent check ensures that all required are present.

oracle/price-feeder/oracle/oracle.go
stevenlanders: I now have it reading from the paramCache which is already on the oracle

PR #1052, opened 2023-09-18
Add EVM module & types/keeper

codchen:
Add a new EVM module and related types:
- A transaction type that allows arbitrary EVM transaction payload to be passed in
  - included encoding to parse such payload into legacy/accesslist/dynamicfee/blob transaction types in EVM
- types and keeper functions for
  - the association between EVM addresses and Sei addresses
  - balance for unassociated EVM addresses

unit tests


proto/eth/tx.proto
```diff
@@ -0,0 +1,150 @@
...
+    uint64 gas = 3 [
+        (gogoproto.customname) = "GasLimit"
+    ];
+    string to = 4;
+    string value = 5 [
+        (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+        (gogoproto.customname) = "Amount"
+    ];
+    bytes data = 6;
+    bytes v = 7;
```
yzang2019, line 34: what does v r s mean? Might be good to add a comment here

codchen: signature values. Those are ethereum naming

proto/eth/tx.proto
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+message LegacyTx {
+    option (gogoproto.goproto_getters) = false;
+
+    uint64 nonce = 1;
+    string gas_price = 2 [
+        (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int"
+    ];
+    uint64 gas = 3 [
```
yzang2019: should this be gas_limit to avoid future confusion?

proto/eth/tx.proto
```diff
@@ -0,0 +1,154 @@
...
+string chain_id = 1 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.customname) = "ChainID",
+    (gogoproto.jsontag) = "chainID"
+];
+uint64 nonce = 2;
+string gas_price = 3 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int"
+];
+uint64 gas = 4 [
```
yzang2019: same here and belows

proto/evm/config.proto
```diff
@@ -0,0 +1,87 @@
...
+    ];
+    string gray_glacier_block = 15 [
+      (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+      (gogoproto.moretags) = "yaml:\"gray_glacier_block\""
+    ];
+    string merge_netsplit_block = 16 [
+      (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+      (gogoproto.moretags) = "yaml:\"merge_netsplit_block\""
+    ];
+    uint64 shanghai_time = 17 [
```
yzang2019, line 84: Just curious what do we need these time config for?

x/evm/keeper/address.go
```diff
@@ -0,0 +1,33 @@
+package keeper
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/sei-protocol/sei-chain/x/evm/types"
+)
+
+func (k *Keeper) SetAddressMapping(ctx sdk.Context, seiAddress sdk.AccAddress, evmAddress common.Address) {
```
yzang2019, line 9: I assume this is for faster lookup. It seems for each account we would store a bidirectional mapping between seiAddress and evmAddress, that's a bit expensive. Is there no other way we can infer/convert/generate evmAddress from sei or vice versa?

x/evm/keeper/balance.go
```diff
@@ -0,0 +1,37 @@
...
+		// this should never happen
+		panic(err)
+	}
+	return balance.Amount.Uint64()
+}
+
+func (k *Keeper) SetBalance(ctx sdk.Context, address common.Address, amount uint64) {
+	store := ctx.KVStore(k.storeKey)
+	if amount == 0 {
+		k.deleteBalance(ctx, address)
```
yzang2019, line 26: nice

x/evm/types/config.go
```diff
@@ -0,0 +1,152 @@
...
+		return errors.New("londonBlock")
+	}
+	if err := validateBlock(cc.ArrowGlacierBlock); err != nil {
+		return errors.New("arrowGlacierBlock")
+	}
+	if err := validateBlock(cc.GrayGlacierBlock); err != nil {
+		return errors.New("GrayGlacierBlock")
+	}
+	if err := validateBlock(cc.MergeNetsplitBlock); err != nil {
+		return errors.New("MergeNetsplitBlock")
```
yzang2019: These two should also be starting with lower case?

x/evm/types/ethtx/access_list_tx.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	v, r, s := tx.RawSignatureValues()
+
+	defer func() {
+		if err := recover(); err != nil {
+			altx = nil
+			reserr = fmt.Errorf("%s", err)
+		}
+	}()
```
yzang2019: What's this for? Why do we need this? Is this to capture some unexpected panic?

x/evm/types/ethtx/utils.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/ethereum/go-ethereum/common/math"
+)
+
+// Gas price is the smaller of base fee + tip limit vs total fee limit
+func EffectiveGasPrice(baseFee, feeCap, tipCap *big.Int) *big.Int {
+	return math.BigMin(new(big.Int).Add(tipCap, baseFee), feeCap)
+}
+
+func MustSetConvertIfPresent[U comparable, V any](orig U, converter func(U) (V, error), setter func(V)) {
```
yzang2019: Can we add a comment for what this is doing?

x/evm/types/ethtx/utils.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+
+	if converted, err := converter(orig); err != nil {
+		panic(err)
+	} else {
+		setter(converted)
+	}
+}
+
+func SetConvertIfPresent[U comparable, V any](orig U, converter func(U) V, setter func(V)) {
```
yzang2019: Can we add a comment for what this is doing?

yzang2019 (approved):
Overall LGTM, left a few non blocking comments.

One suggestion to add more comments above functions to make it easier to read, especially for some tricky or helper functions

proto/eth/tx.proto
yzang2019: Cool, found it in https://ethereum.org/en/developers/docs/transactions/

proto/eth/tx.proto
```diff
@@ -0,0 +1,154 @@
+syntax = "proto3";
```
philipsu522, line 1: do you mind linking this proto file to resources you used to create this?

x/evm/keeper/balance.go
```diff
@@ -0,0 +1,37 @@
...
+	if err := balance.Unmarshal(bz); err != nil {
+		// this should never happen
+		panic(err)
+	}
+	return balance.Amount.Uint64()
+}
+
+func (k *Keeper) SetBalance(ctx sdk.Context, address common.Address, amount uint64) {
+	store := ctx.KVStore(k.storeKey)
+	if amount == 0 {
```
philipsu522, line 25: would be good to note that this has a side effect if balance == 0 (comment or function name)

x/evm/keeper/balance.go
```diff
@@ -0,0 +1,37 @@
...
+		k.deleteBalance(ctx, address)
+		return
+	}
+	balance := sdk.NewCoin(k.GetBaseDenom(ctx), sdk.NewIntFromUint64(amount))
```
philipsu522, line 29: shoudl we do any basic validation here?

x/evm/keeper/send.go
```diff
@@ -0,0 +1,55 @@
...
+
+func (k *Keeper) EVMToEVMSend(ctx sdk.Context, fromAddress common.Address, toAddress common.Address, amount uint64) error {
+	if err := k.DebitAddress(ctx, fromAddress, amount); err != nil {
+		return err
+	}
+	return k.CreditAddress(ctx, toAddress, amount)
+}
+
+func (k *Keeper) BankToEVMSend(ctx sdk.Context, bankAddress sdk.AccAddress, toAddress common.Address, amount uint64) error {
+	// first transfer tokens to evm module account
```
philipsu522, line 27: should we check some edge cases (like amount >= 0). Do we also allow arbitrary types (ikbc) or is everythign guaranteed to be in sei

x/evm/keeper/send.go
```diff
@@ -0,0 +1,55 @@
...
+		sdk.NewCoins(sdk.NewCoin(k.GetBaseDenom(ctx), sdk.NewIntFromUint64(amount)))); err != nil {
+		return err
+	}
+	// then debit the sender address
+	return k.DebitAddress(ctx, evmAddress, amount)
+}
+
+func (k *Keeper) CreditAddress(ctx sdk.Context, address common.Address, amount uint64) error {
+	balance := k.GetBalance(ctx, address)
+	if balance+amount < balance {
```
philipsu522: is this intended to check for number wrap around? we could probably just compare balance to UINT64.MAX or something?

x/evm/keeper/send.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func (k *Keeper) EVMToBankSend(ctx sdk.Context, evmAddress common.Address, toAddress sdk.AccAddress, amount uint64) error {
+	// first transfer tokens out of evm module account
+	if err := k.bankKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, toAddress,
+		sdk.NewCoins(sdk.NewCoin(k.GetBaseDenom(ctx), sdk.NewIntFromUint64(amount)))); err != nil {
+		return err
+	}
+	// then debit the sender address
+	return k.DebitAddress(ctx, evmAddress, amount)
```
philipsu522, line 46: do we need to validate that we can debit the amount before doing the send in line 30?

x/evm/types/config.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	return nil
+}
+
+func validateBlock(block *sdk.Int) error {
+	if block == nil {
+		return nil
+	}
+
+	if block.IsNegative() {
```
philipsu522, line 154: nit block > 0?

x/evm/types/ethtx/txdata.go
```diff
@@ -0,0 +1,92 @@
...
+	if len(rBz) > 0 {
+		r = new(big.Int).SetBytes(rBz)
+	}
+	if len(sBz) > 0 {
+		s = new(big.Int).SetBytes(sBz)
+	}
+	return v, r, s
+}
+
+// fee = gas limit * gas price
```
philipsu522, line 80: could we resuse cosmos-sdk native impl here?

x/evm/types/ethtx/utils.go
philipsu522: docstrings would be helpful for funcs in this file

proto/evm/config.proto
codchen: these are basically the upgrade timestamp on Ethereum. For Sei those will all be 0 indicating that Sei hasn't had any of those upgrades and started from Cancun (the latest version), but keeping these configs here for backward compatibility with EVM

x/evm/keeper/address.go
codchen: we can technically only store a EVM->Sei mapping, and use the Pubkey that we are already storing to derive the EVM address for query from Sei to EVM addr. Let me change it to that

x/evm/types/ethtx/access_list_tx.go
codchen: yeah it's to capture panics that might be thrown by `MustSetConvertIfPresent`. There might be a way to handle it without involving panic

x/evm/keeper/balance.go
codchen: yeah updated function name

x/evm/keeper/balance.go
codchen: yeah we check for amount == 0 above

x/evm/keeper/send.go
codchen: amount is uint so it will always be non-negative. Added checks for amount==0 to early return. Everything will be in usei w.r.t EVM

x/evm/keeper/send.go
codchen: balance itself is by invariant always valid, but to enforce that invariant we need to check the amount we set to balance which is `balance+amount`. I've changed it to a more readable check for this purpose

x/evm/keeper/send.go
codchen: evm uses the same cache state as normal tx so any error would cause all state changes in a tx to be rolled back

x/evm/types/config.go
codchen: `block` can be 0 in the config which would indicate that the corresponding upgrade was ineligible (i.e. the chain started with the new version). I'll add comments in this file to better explain it

x/evm/types/ethtx/txdata.go
codchen: there isn't an existing impl - it's spread across in https://github.com/sei-protocol/sei-cosmos/blob/main/x/auth/ante/validator_tx_fee.go#L25-L43

proto/evm/config.proto
codchen: added comments

x/evm/keeper/address.go
codchen: well actually we can't do what I described above, since we only store the 33-byte compressed version of the pubkey, whereas to restore the evm address we need the full 65-byte pubkey. To change it to store full pubkey would be a major migration, and also it doesn't save that much space compared to storing a additional direction of mapping

x/evm/types/ethtx/access_list_tx.go
codchen: updated

x/evm/types/ethtx/utils.go
codchen: removed it altogether

x/evm/types/ethtx/utils.go
codchen: added

x/evm/types/ethtx/utils.go
codchen: updated comments

PR #1055, opened 2023-09-18
Improve pf providers

udpatil:
This does the following:
- adds some additional debug and warning logging
- Refactors the gate provider to use the new API that allows for subscribing to more than one currency pair
- removes the all or nothing behavior that prevents providers from returning prices unless ALL of their configured asset prices are available

Tested with oracle price feeder with local chain, need to test longer running on a loadtest cluster with steven's monitoring additions

### DIFF
--- a/oracle/price-feeder/oracle/oracle.go
+++ b/oracle/price-feeder/oracle/oracle.go
@@ -318,9 +318,6 @@ func GetComputedPrices(
 		assetProviderMap := make(map[string][]string)
 		for provider, val := range providerPrices {
 			for asset := range val {
-				if _, ok := assetProviderMap[asset]; !ok {
-					assetProviderMap[asset] = []string{}
-				}
 				assetProviderMap[asset] = append(assetProviderMap[asset], provider)
 			}
 		}
@@ -333,9 +330,6 @@ func GetComputedPrices(
 		candleProviderMap := make(map[string][]string)
 		for provider, val := range providerCandles {
 			for asset := range val {
-				if _, ok := candleProviderMap[asset]; !ok {
-					candleProviderMap[asset] = []string{}
-				}
 				candleProviderMap[asset] = append(candleProviderMap[asset], provider)
 			}
 		}
@@ -367,19 +361,19 @@ func GetComputedPrices(
 	}
 
 	// attempt to use candles for TVWAP calculations
-	tvwapPrices, err := ComputeTVWAP(filteredCandles)
+	computedPrices, err := ComputeTVWAP(filteredCandles)
 	if err != nil {
 		return nil, err
 	}
 
 	candleAssets := []string{}
 	tickerAssets := []string{}
-	for base := range tvwapPrices {
+	for base := range computedPrices {
 		candleAssets = append(candleAssets, base)
 	}
 	allRequiredAssetsPresent := true
 	for asset := range requiredRates {
-		if _, ok := tvwapPrices[asset]; !ok {
+		if _, ok := computedPrices[asset]; !ok {
 			allRequiredAssetsPresent = false
 		}
 	}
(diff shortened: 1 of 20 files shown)


oracle/price-feeder/oracle/oracle.go
```diff
@@ -322,11 +322,26 @@ func GetComputedPrices(
...
+					candleProviderMap[asset] = []string{}
+				}
+				candleProviderMap[asset] = append(candleProviderMap[asset], provider)
+			}
+		}
+		candleProviderJSON, err := json.Marshal(candleProviderMap)
+		if err != nil {
+			return nil, err
+		}
+		logger.Debug().Msg(fmt.Sprintf("Candle Provider Coverage Map: %s", string(candleProviderJSON)))
```
yzang2019, line 340: It would be cleaner to move the whole debugging log logic to a different function

oracle/price-feeder/oracle/provider/binance_test.go
```diff
@@ -89,18 +89,14 @@ func TestBinanceProvider_GetTickerPrices(t *testing.T) {
 }
 
 func TestBinanceProvider_SubscribeCurrencyPairs(t *testing.T) {
-	server := NewMockProviderServer()
-	server.Start()
-	defer server.Close()
+	// server := NewMockProviderServer()
+	// server.Start()
+	// defer server.Close()
```
yzang2019: Shall we clean these up?

oracle/price-feeder/oracle/provider/coinbase.go
```diff
@@ -150,7 +150,8 @@ func (p *CoinbaseProvider) GetTickerPrices(pairs ...types.CurrencyPair) (map[str
 	for _, currencyPair := range pairs {
 		price, err := p.getTickerPrice(currencyPair)
 		if err != nil {
-			return nil, err
+			p.logger.Warn().Msg(fmt.Sprint("failed to fetch tickers for pair ", currencyPair, " due to the following error ", err.Error()))
```
yzang2019: Would this be too noisy? Shall we change to debug log as well since we have metrics now?

oracle/price-feeder/oracle/provider/gate.go
```diff
@@ -335,7 +374,7 @@ func (p *GateProvider) messageReceived(messageType int, bz []byte) {
 		tickerErr error
 		candleErr error
 	)
-
+	// fmt.Println("gate response", string(bz))
```
yzang2019: Shall we just delete it?

yzang2019:
> removes the all or nothing behavior

Where's the change for this?

oracle/price-feeder/oracle/provider/coinbase.go
udpatil: yeah, we can change to debug log

udpatil:
> > removes the all or nothing behavior
> 
> Where's the change for this?

this is the change from returning in an error in the provider `GetTickerPrices` and `GetCandlePrices` and instead logging

oracle/price-feeder/oracle/oracle.go
```diff
@@ -382,9 +411,14 @@ func GetComputedPrices(
 			return nil, err
 		}
 
-		return vwapPrices, nil
+		for asset, price := range vwapPrices {
+			if _, ok := tvwapPrices[asset]; !ok {
+				tickerAssets = append(tickerAssets, asset)
+				tvwapPrices[asset] = price
```
stevenlanders: small nit:  i wonder if there's a better name for `tvwapPrices` if it can contain a vwapPrice

oracle/price-feeder/oracle/oracle.go
```diff
@@ -322,11 +324,26 @@ func GetComputedPrices(
...
 		if err != nil {
 			return nil, err
 		}
 		logger.Debug().Msg(fmt.Sprintf("Asset Provider Coverage Map: %s", string(assetProviderJSON)))
+
+		candleProviderMap := make(map[string][]string)
+		for provider, val := range providerCandles {
+			for asset := range val {
+				if _, ok := candleProviderMap[asset]; !ok {
+					candleProviderMap[asset] = []string{}
```
stevenlanders: fwiw, I think you can append directly to a non-present value (no need to init)

stevenlanders (approved):
LGTM (small comments)

oracle/price-feeder/oracle/oracle.go
udpatil: we can call it computedPrices

philipsu522 (approved):
does oracle export metrics? Could we add error rates for which ticker prices & which exchanges are failing?

udpatil:
> does oracle export metrics? Could we add error rates for which ticker prices & which exchanges are failing?

this is done in steven's PR

yzang2019 (approved):

PR #1059, opened 2023-09-20
EVM ante handler for signature verification

codchen:
- Added an ante router so that EVM transactions will use a different ante handler chain altogether
- Added a preprocess ante handler for EVM messages to decorate contexts with metadata like sender
- Added a signature verification handler which merely checks if a sender exists in the context (the actual signature verification is done in the preprocessor)

unit tests


x/evm/ante/router.go
```diff
@@ -0,0 +1,45 @@
...
+			continue
+		}
+	}
+
+	if !hasEVMMsg {
+		return r.defaultRoute(ctx, tx, simulate)
+	}
+
+	// A tx that has EVM message must have exactly one message
+	if len(tx.GetMsgs()) != 1 {
```
yzang2019, line 40: We don't allow batching evm msg with other non evm messages in a single TX? Why?

x/evm/ante/preprocess.go
```diff
@@ -0,0 +1,170 @@
...
+type EVMPreprocessDecorator struct {
+	evmKeeper     *evmkeeper.Keeper
+	accountKeeper *accountkeeper.AccountKeeper
+}
+
+func NewEVMPreprocessDecorator(evmKeeper *evmkeeper.Keeper, accountKeeper *accountkeeper.AccountKeeper) *EVMPreprocessDecorator {
+	return &EVMPreprocessDecorator{evmKeeper: evmKeeper, accountKeeper: accountKeeper}
+}
+
+func (p EVMPreprocessDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
```
yzang2019, line 56: This is a function with a huge body, can we break down the body into smaller pieces, maybe add some helper private functions

x/evm/ante/router.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"errors"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/sei-protocol/sei-chain/x/evm/types"
+)
+
+type EVMRouterDecorator struct {
+	defaultRoute sdk.AnteHandler
```
yzang2019: How about changing to
- defaultAnteHandler
- evmAnteHandler

x/evm/ante/preprocess.go
```diff
@@ -0,0 +1,170 @@
...
+	ethTx, txData := tx.GetMsgs()[0].(*evmtypes.MsgEVMTransaction).AsTransaction()
```
yzang2019, line 61: I know we have checked before in the router, but I think it's worth to do another sanity check here to make sure GetMsgs()[0] exist

x/evm/ante/router.go
codchen: It's mainly for simplicity since the ante flow for evm messages and non-evm messages are completely different. Since EVM messages need to be signed differently than non-EVM messages it's unlikely for users to batch them anyway

yzang2019: Gotcha, make sense to me

x/evm/ante/preprocess.go
codchen: yeah, refactored it

x/evm/ante/router.go
codchen: updated

x/evm/ante/preprocess.go
codchen: added

proto/eth/tx.proto
```diff
@@ -131,7 +131,14 @@ string blob_fee_cap = 10 [
 ];
 repeated bytes blob_hashes = 11;
 BlobTxSidecar sidecar = 12;
+<<<<<<< HEAD
+<<<<<<< HEAD
 // signature values
+=======
```
yzang2019: We need to resolve the conflict?

codchen: good catch

yzang2019 (approved):

PR #1064, opened 2023-09-20
Cut v3.1.1 for version release

Kbhat1:
Changes for 3.1.1 upgrade:

- Add upgrade handler for v3.1.0
- sei-ibc-go version has already been bumped

NOTE: Bumping intentionally from v3.1.0 to v3.1.1 since there was an issue creating the git tag for v3.1.0. This is to avoid any issue with the tag while nodes are upgrading


app/upgrades.go
```diff
@@ -60,6 +60,7 @@ var upgradesList = []string{
 	"3.0.8",
 	"v3.0.9",
 	"v3.1.0",
```
yzang2019: shall we remove this?

Kbhat1: @yzang2019 yea I left it in because it shouldn't cause any issue. But I can just go ahead and remove

Kbhat1: Updated

yzang2019 (approved):

PR #1065, opened 2023-09-20
Avoid re-initializing providers that fail initially

stevenlanders:
- In US regions, binance clients never initialize
- Initializing takes ~500ms, which impacts price broadcasting cadence
- This quickfix avoids re-initializing binance repeatedly if it fails at startup

- unit test
- local testing

### DIFF
--- a/oracle/price-feeder/oracle/oracle.go
+++ b/oracle/price-feeder/oracle/oracle.go
@@ -528,6 +528,7 @@ func (o *Oracle) getOrSetProvider(ctx context.Context, providerName string) (pro
 		ok            bool
 	)
 
+	//TODO: replace with a exponential backoff mechanism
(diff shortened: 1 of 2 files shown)


oracle/price-feeder/oracle/oracle.go
```diff
@@ -521,6 +528,10 @@ func (o *Oracle) getOrSetProvider(ctx context.Context, providerName string) (pro
 		ok            bool
 	)
 
+	if err, ok := o.failedProviders[providerName]; ok {
```
udpatil, line 532: can we add a TODO here to replace this behavior with a retry with backoff in the future?

udpatil (approved):

yzang2019 (approved):

PR #1067, opened 2023-09-24
[EVM] Add gas check ante handler

codchen:
Add ante handler to check gas during CheckTx. The most important check is to make sure the sender has enough funds as indicated by their tx's `Fee()` response in their bank balance. In addition, if the transaction is for after London, then its specified gas fee cap must not be smaller than Sei's min gas price for `usei` multiplied by the EVM gas multiplier.

unit tests


x/evm/ante/fee_test.go
```diff
@@ -0,0 +1,153 @@
...
+	ethtypes "github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/holiman/uint256"
+	"github.com/sei-protocol/sei-chain/x/evm/keeper"
+	"github.com/sei-protocol/sei-chain/x/evm/types"
+	"github.com/sei-protocol/sei-chain/x/evm/types/ethtx"
+	"github.com/stretchr/testify/require"
+)
+
+func TestEVMFeeCheckDecorator(t *testing.T) {
```
jewei1997: This test seems to only run the AnteHandle with a Cancun version. Might also be good to test it with a London version as well as a much older version.

x/evm/ante/fee.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	// if EVM version is London or later, gas fee cap will be used to cap the overall fee consumption,
+	// so we need to make sure that cap is at least as large as the required base fee
+	if ver >= evmtypes.London && txData.GetGasFeeCap().Cmp(fc.getBaseFee(ctx, txData.GetGas())) < 0 {
+		fmt.Println(txData.GetGasFeeCap())
+		fmt.Println(fc.getBaseFee(ctx, txData.GetGas()))
+		return ctx, errors.New("provided gas fee cap is smaller than required base fee")
+	}
+	// if EVM version is Cancun or later, and the transaction contains at least one block, we need to
```
jewei1997: nit: block contains at least 1 transaction.

x/evm/types/context.go
```diff
@@ -10,13 +10,25 @@ import (
...
+type Version int
+
+const (
+	Frontier Version = iota
+	Homestead
+	EIP155
+	Berlin
+	London
+	Cancun
+)
```
jewei1997: Just for my education: can you explain why these are the only versions we are including? Also was wondering why we don't have the latest fork which I believe is Shanghai, and are including Cancun which is a future fork.

x/evm/ante/fee.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	ver, ok := evmtypes.GetContextEVMVersion(ctx)
+	if !ok {
+		return ctx, errors.New("could not find eth version when checking fees")
+	}
+
+	// if EVM version is London or later, gas fee cap will be used to cap the overall fee consumption,
+	// so we need to make sure that cap is at least as large as the required base fee
+	if ver >= evmtypes.London && txData.GetGasFeeCap().Cmp(fc.getBaseFee(ctx, txData.GetGas())) < 0 {
```
philipsu522, line 50: if this is an upgrade that is going to eventually occur / has occured, would it make sense to juts support post-London upgrade? We can dictate the EVM version on-chain right?

x/evm/ante/sig_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	ethtypes "github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/sei-protocol/sei-chain/x/evm/keeper"
+	"github.com/sei-protocol/sei-chain/x/evm/types"
+	"github.com/sei-protocol/sei-chain/x/evm/types/ethtx"
+	"github.com/stretchr/testify/require"
+)
+
+func TestEVMSigVerifyDecorator(t *testing.T) {
```
philipsu522, line 18: let's also add tests for non-existent sei<>evm addrs (if not already covered elsewhere)

x/evm/ante/fee.go
codchen: yeah that's true. I'll just include London and Cancun here (since Cancun hasn't landed on ethereum mainnet yet we should probably start with the current version)

x/evm/ante/sig_test.go
codchen: added

x/evm/types/context.go
codchen: this is for deciding Signer version so only these versions are relevant. Changed the type name to reflect that. Also trimmed unnecessary version since we will start from signer version London (overall version Shanghai)

x/evm/ante/fee.go
codchen: nice catch updated

x/evm/ante/fee_test.go
codchen: added

PR #1068, opened 2023-09-25
[EVM] implement balance related methods for state interface

codchen:
Implement the following EVM interface functions based on `x/evm` balance and `x/bank` balance:
- GetBalance
- AddBalance
- SubBalance
Also added a non-interface method `CheckBalance` to protect against deficits due to bugs. Explained more in comments.

unit tests


x/evm/state/balance.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func (s *StateDBImpl) SubBalance(evmAddr common.Address, amt *big.Int) {
+	if amt.Sign() == 0 {
+		return
+	}
+	if amt.Sign() < 0 {
+		s.AddBalance(evmAddr, new(big.Int).Neg(amt))
+		return
+	}
```
philipsu522, line 27: out of curiosity, why do we want to support sub/add'ing negative amounts?

x/evm/state/balance.go
```diff
@@ -0,0 +1,103 @@
...
+	if amt.Sign() < 0 {
+		s.SubBalance(evmAddr, new(big.Int).Neg(amt))
+		return
+	}
+	defer func() {
+		s.deficit = new(big.Int).Add(s.deficit, amt)
+	}()
+
+	if seiAddr, ok := s.k.GetSeiAddress(s.ctx, evmAddr); ok {
+		// credit seiAddr's bank balance and debit EVM module account, mint if needed
```
philipsu522, line 61: can you walk me through this? are we minting sei coins if the evm module account doesn't have enough balance?

x/evm/state/balance_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"math/big"
+	"testing"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/sei-protocol/sei-chain/x/evm/keeper"
+	"github.com/sei-protocol/sei-chain/x/evm/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAddBalance(t *testing.T) {
```
philipsu522, line 15: can we also add tests for 
- very large values (overflow)
- insufficient balances

x/evm/state/balance.go
codchen: these functions are called by go-ethereum, and the original implementation makes no assumption on the sign of `amt`, so just want to make sure we support the same behavior

x/evm/state/balance.go
codchen: yeah this would only happen if during a money movement in EVM, an `AddBalance` is called before the corresponding `SubBalance`. Since evm module account hasn't received the funds from `SubBalance` yet, it needs to temporarily mint some to cover the needs in `AddBalance`. All minted tokens will be burned in `CheckBalance` which is called once at the end of each EVM transaction

x/evm/state/balance_test.go
codchen: added

PR #1071, opened 2023-09-27
Add state-related function implementations in EVM state interface

codchen:
Implement functions to initialize/get/set/destruct state in EVM state interface. Specifically,
- Get/SetState sets a KV entry under a specified account prefix
- GetCommittedState is similar to GetState but reads from the state as if at the beginning of the transaction (i.e. skip the cache layer and go directly to the parent. See corresponding sei-cosmos extension here: https://github.com/sei-protocol/sei-cosmos/pull/319)
- Get/SetTransientState writes to a in-memory map under a specified account prefix, and is not committed to KV store.
- SelfDestruct would burn account's balance and clear its state, but leave transient state untouched (as per go-ethereum behavior)
- SelfDestruct6780 is equivalent to SelfDestruct iff the account is newly created in the same block
- CreateAccount marks an account as created in the current block (a transient marker). If the account already exists, it wipes its state, but leaves the transient state and balance untouched (as per go-ethereum behavior)

unit tests

### DIFF
--- a/x/evm/state/state.go
+++ b/x/evm/state/state.go
@@ -1,0 +1,127 @@
+package state
+
+import (
+	"github.com/cosmos/cosmos-sdk/store/prefix"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/sei-protocol/sei-chain/x/evm/types"
+)
+
+func (s *StateDBImpl) CreateAccount(acc common.Address) {
+	// clear any existing state but keep balance untouched
+	s.clearAccountState(acc)
+
+	s.created[acc.String()] = struct{}{}
+	delete(s.selfDestructedAccs, acc.String())
+}
+
+func (s *StateDBImpl) GetCommittedState(addr common.Address, hash common.Hash) common.Hash {
+	return s.getState(addr, hash, func(kv sdk.KVStore, h common.Hash) []byte { return kv.GetCommitted(h[:]) })
+}
+
+func (s *StateDBImpl) GetState(addr common.Address, hash common.Hash) common.Hash {
+	return s.getState(addr, hash, func(kv sdk.KVStore, h common.Hash) []byte { return kv.Get(h[:]) })
+}
+
+func (s *StateDBImpl) getState(addr common.Address, hash common.Hash, getter func(sdk.KVStore, common.Hash) []byte) common.Hash {
+	val := getter(s.prefixStore(addr), hash)
+	if val == nil {
+		return common.Hash{}
+	}
+	return common.BytesToHash(val)
+}
+
+func (s *StateDBImpl) SetState(addr common.Address, key common.Hash, val common.Hash) {
+	s.prefixStore(addr).Set(key[:], val[:])
+}
+
+func (s *StateDBImpl) GetTransientState(addr common.Address, key common.Hash) common.Hash {
+	if addrState, ok := s.transientStorage[addr.String()]; !ok {
+		return common.Hash{}
+	} else if val, ok := addrState[key.String()]; !ok {
+		return common.Hash{}
+	} else {
+		return common.BytesToHash(val)
+	}
+}
+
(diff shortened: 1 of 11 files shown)


x/evm/state/state.go
```diff
@@ -0,0 +1,127 @@
...
+		return common.BytesToHash(val)
+	}
+}
+
+func (s *StateDBImpl) SetTransientState(addr common.Address, key, value common.Hash) {
+	addrKey := addr.String()
+	if addrState, ok := s.transientStorage[addrKey]; !ok {
+		s.transientStorage[addrKey] = map[string][]byte{key.String(): value[:]}
+	} else {
+		addrState[key.String()] = value[:]
```
jewei1997: Might be good to check if prevValue is the same as current value and if so, just return: https://github.com/ethereum/go-ethereum/blame/614804b33c340cd60cbb6087d898b2968b0da320/core/state/statedb.go#L475

Also, do you need to make a copy of `value` here?

x/evm/state/state.go
```diff
@@ -0,0 +1,127 @@
...
+
+// burns account's balance
+// clear account's state except the transient state (in Ethereum transient states are
+// still available even after self destruction in the same tx)
+func (s *StateDBImpl) SelfDestruct(acc common.Address) {
+	var balance sdk.Coin
+	if seiAddr, ok := s.k.GetSeiAddress(s.ctx, acc); ok {
+		// send all useis from seiAddr to the EVM module
+		balance = s.k.BankKeeper().GetBalance(s.ctx, seiAddr, s.k.GetBaseDenom(s.ctx))
+		if balance.Amount.Int64() != 0 {
```
jewei1997, line 62: Is it possible for `balance` to be larger than `2^64`

x/evm/state/state.go
```diff
@@ -0,0 +1,127 @@
...
+	if _, ok := s.created[acc.String()]; ok {
+		s.SelfDestruct(acc)
+	}
+}
+
+// the Ethereum semantics of HasSelfDestructed checks if the account is self destructed in the
+// **CURRENT** block
+func (s *StateDBImpl) HasSelfDestructed(acc common.Address) bool {
+	_, ok := s.selfDestructedAccs[acc.String()]
+	return ok
```
jewei1997: would be good to call this after a self destruct to make sure `selfDestructedAccs` was updated in your tests.

x/evm/state/state.go
```diff
@@ -0,0 +1,127 @@
...
+			return
+		}
+	}
+
+	// clear account state
+	s.clearAccountState(acc)
+
+	// mark account as self-destructed
+	s.selfDestructedAccs[acc.String()] = struct{}{}
+	delete(s.created, acc.String())
```
jewei1997: are we testing that the account was deleted from the `created` map?

x/evm/state/state.go
codchen: updated it to use the same KV store as other states, and purge it at the end of the transaction https://github.com/sei-protocol/sei-chain/pull/1071/files#diff-ba4a9dd61688ef34761fc3792f6ef4a8bbf5f2a58d235e6d998928a154f9ef92R38

x/evm/state/state.go
codchen: total Sei supply is 10^10, or 10^16 usei, whereas int64 can represent up to 9^18, so i think we should be good when it's just representing the absolute value of certain balance. We do need to check for overflow though when addition happens, just in case EVM hypothetically adds large sum onto some balances.

x/evm/state/state.go
codchen: good call. Added

x/evm/state/state.go
codchen: nice catch. Added a check here https://github.com/sei-protocol/sei-chain/pull/1071/files#diff-479c396576435ca3e9c897185761058424f6485d0d062e89c26bc8318fc82649R110

PR #1074, opened 2023-09-27
Have local chain accept connections from any origin

ARitz-Cracker:
This sets `cors-allowed-origins` of the auto-generated config file to `["*"]`, allowing for the RPC node to be used during front-end testing.

- [x] Script ran on Linux, header present when doing a cross-origin call.
- [ ] Untested on MacOS


ARitz-Cracker:
nevermind, this doesn't make the rest endpoint also use unsafe cors, which would require also editing app.toml, unsure if getting the dirname from `$CONFIG_PATH` would be sufficient, closing for now.

PR #1076, opened 2023-09-28
add linear tag to bug_report.md

duncand0nuts:
bug reports created using templates will have linear tag applied


FaliKurve:
Good watch

LCyson (approved):

PR #1079, opened 2023-10-02
add general issue template

duncand0nuts:


yzang2019:
Can we point them to check out and ask questions in the public telegram group first? We can add a checkbox to make sure they've already asked in the telegram chat, and also provide instructions for them to join the group

Generally we want github issues to be focused more on feature request, bug report, and we want to reduce general questions like how to run a chain, how to set up xyz

LCyson (approved):

PR #1083, opened 2023-10-04
[EVM] implement message server that processes EVM transactions

codchen:
- Implement EVMMessage endpoint that instantiate an EVM instance with a `StateDBImpl` and applies it.
- Added logics for EVM transaction receipt
- Added logics to emit EVM event logs as native events

`msg_server_test`


x/evm/keeper/coinbase.go
```diff
@@ -0,0 +1,20 @@
...
+	"errors"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+const FeeCollectorAddress = "41cc5d9842746c69d689c8379f7f5662b8701393"
+
+func (server msgServer) GetFeeCollectorAddress(ctx sdk.Context) (common.Address, error) {
```
jewei1997: hm why not call this coinbase address?

x/evm/keeper/coinbase.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"errors"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+const FeeCollectorAddress = "41cc5d9842746c69d689c8379f7f5662b8701393"
```
jewei1997: where does this come from?

x/evm/keeper/msg_server.go
```diff
@@ -0,0 +1,187 @@
...
+var _ types.MsgServer = msgServer{}
+
+func (server msgServer) EVMTransaction(goCtx context.Context, msg *types.MsgEVMTransaction) (*types.MsgEVMTransactionResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	stateDB := state.NewStateDBImpl(ctx, &server)
+	tx, _ := msg.AsTransaction()
+	ctx, gp := server.getGasPool(ctx)
+	emsg, err := server.getEVMMessage(ctx, tx)
+	if err != nil {
+		return &types.MsgEVMTransactionResponse{}, err
```
jewei1997: should we return `nil, err` here?

x/evm/keeper/msg_server_test.go
```diff
@@ -0,0 +1,148 @@
...
+	require.Nil(t, err)
+	bz, err = abi.Pack("set", big.NewInt(20))
+	require.Nil(t, err)
+	txData = ethtypes.LegacyTx{
+		GasPrice: big.NewInt(1),
+		Gas:      200000,
+		To:       &contractAddr,
+		Value:    big.NewInt(0),
+		Data:     bz,
+		Nonce:    1,
```
jewei1997, line 84: may be good to do a failed tx test where gas is insufficient or nonce is incorrect as well.

jewei1997: also may be good to do an EIP-1559 txn (as opposed to legacy)

x/evm/state/state.go
```diff
@@ -31,6 +32,7 @@ func (s *StateDBImpl) getState(ctx sdk.Context, addr common.Address, hash common
 }
 
 func (s *StateDBImpl) SetState(addr common.Address, key common.Hash, val common.Hash) {
+	fmt.Println(key.Hex())
```
jewei1997: nit: remove

x/evm/state/expected_keepers.go
```diff
@@ -0,0 +1,23 @@
+package state
```
jewei1997, line 1: curious: why is this file called expected_keepers?

x/evm/state/statedb.go
```diff
@@ -53,5 +56,43 @@ func (s *StateDBImpl) Finalize() error {
 		s.snapshottedCtxs[i].MultiStore().(sdk.CacheMultiStore).Write()
 	}
 
+	if logs, err := s.GetLogs(); err != nil {
+		return err
+	} else {
+		for _, l := range logs {
```
jewei1997, line 52: would be good to have a smart contract that emits an event to test this

jewei1997 (approved):
some questions and requests for tests, but overall lgtm!

x/evm/keeper/msg_server.go
```diff
@@ -0,0 +1,187 @@
...
+		// infinite gas meter
+		return ctx, math.MaxUint64
+	} else {
+		return ctx, core.GasPool(ctx.BlockGasMeter().Limit() - ctx.BlockGasMeter().GasConsumedToLimit())
+	}
+}
+
+func (server msgServer) getEVMMessage(ctx sdk.Context, tx *ethtypes.Transaction) (*core.Message, error) {
+	cfg := server.GetChainConfig(ctx).EthereumConfig(server.ChainID())
+	signer := ethtypes.MakeSigner(cfg, big.NewInt(ctx.BlockHeight()), uint64(ctx.BlockTime().Unix()))
```
philipsu522, line 86: is this going to cause issues with clock skew? i.e. is this committed to state?

x/evm/keeper/msg_server.go
```diff
@@ -0,0 +1,187 @@
...
+	return core.TransactionToMessage(tx, signer, server.getBaseFee())
+}
+
+func (server msgServer) applyEVMMessage(ctx sdk.Context, msg *core.Message, stateDB vm.StateDB, gp core.GasPool) (*core.ExecutionResult, error) {
+	coinbase, err := server.GetFeeCollectorAddress(ctx)
```
philipsu522, line 91: why is it coinbase specific?

x/evm/keeper/msg_server.go
```diff
@@ -0,0 +1,187 @@
...
+	} else if len(msg.Data) > 0 {
+		contractAddr = *msg.To
+	}
+
+	cumulativeGasUsed := res.UsedGas
+	if ctx.BlockGasMeter() != nil {
+		limit := ctx.BlockGasMeter().Limit()
+		cumulativeGasUsed += ctx.BlockGasMeter().GasConsumed()
+		if cumulativeGasUsed > limit {
+			cumulativeGasUsed = limit
```
philipsu522, line 127: is this expected behavior? do we want to error out or something on situations where we are past the limit?

x/evm/keeper/msg_server.go
```diff
@@ -0,0 +1,187 @@
...
+		if ctx.BlockHeight() < h {
+			// future block doesn't have a hash yet
+			return common.Hash{}
+		}
+		// fetch historical hash from historical info
+		return server.getHistoricalHash(ctx, h)
+	}
+}
+
+func (server msgServer) getHistoricalHash(ctx sdk.Context, h int64) common.Hash {
```
philipsu522, line 172: should we return error too? what is oss evm behavior?

x/evm/keeper/msg_server_test.go
philipsu522: +1, let's add unit tests for more edge cases

x/evm/keeper/genesis.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	"github.com/ethereum/go-ethereum/common"
+
+	"github.com/sei-protocol/sei-chain/x/evm/types"
+)
+
+func (k Keeper) InitGenesis(ctx sdk.Context) {
```
philipsu522, line 10: dont think it's necessary, but do we need to add migration logic?

x/evm/keeper/msg_server.go
```diff
@@ -0,0 +1,187 @@
...
+		return &types.MsgEVMTransactionResponse{}, err
+	}
+
+	res, err := server.applyEVMMessage(ctx, emsg, stateDB, gp)
+	if err != nil {
+		return &types.MsgEVMTransactionResponse{}, err
+	}
+
+	if err := server.writeReceipt(ctx, tx, emsg, res); err != nil {
+		return &types.MsgEVMTransactionResponse{}, err
```
philipsu522: what does a writeReceipt do? Coming from very ltitle evm knowlege, some more docstrings in general (at least for each of the `new` classes, may help

x/evm/keeper/coinbase.go
codchen: in Cosmos world `coinbase` is called feeCollector, but yeah it looks confusing to use them interchangeably. I'm changing every reference under `evm` to `coinbase` instead

x/evm/keeper/coinbase.go
codchen: just a deterministic non-EOA address. I've changed it to be more explicit in its generation

x/evm/state/expected_keepers.go
codchen: it's a conventional name in Cosmos codebase for keeper interfaces. Using an interface here to avoid circular dependency between `state` module and `keeper` module

x/evm/keeper/msg_server.go
codchen: ctx.BlockTime here is the consensus timestamp included in the block proposal and is guaranteed to be always increasing

x/evm/keeper/msg_server.go
codchen: `coinbase` here is an Ethereum term that's basically equivalent to `feeCollector` in Sei, not the exchange.

x/evm/keeper/msg_server.go
codchen: it will eventually cause a panic here https://github.com/sei-protocol/sei-cosmos/blob/main/baseapp/baseapp.go#L1006C12-L1006C12. Here we want to still write cumulative gas to receipt (even if it will be reverted when the panic happens) because there might be other messages in the same tx (e.g. if the evm message is sent by a CW contract) that depend on this receipt

x/evm/keeper/msg_server.go
codchen: in ethereum it returns empty if block is not found https://github.com/ethereum/go-ethereum/blob/master/core/evm.go#L88

x/evm/keeper/genesis.go
codchen: we will need migration logic to add this new module, but no custom state mutation logics are needed

x/evm/keeper/msg_server_test.go
codchen: added both

x/evm/state/statedb.go
codchen: modified SimpleStorage to also emit an event and checked it in the first test in msg_server_test

x/evm/keeper/msg_server.go
codchen: yeah, added in the keeper files

PR #1088, opened 2023-10-05
[EVM] add bank precompile

codchen:
**This PR upgrades Go version to 1.20 to use the `embed` functionality**

Add a custom precompile to allow EVM contracts interact with bank module directly.

This PR involves a change to `go.mod` that updates go-ethereum dependency to our own fork, which extends the precompile interface to allow state being passed in.

unit test with a solidity contract that calls the bank precompile


precompiles/bank/bank.go
```diff
@@ -0,0 +1,152 @@
...
+
+	return &Precompile{
+		Precompile: pcommon.Precompile{ABI: newAbi},
+		bankKeeper: bankKeeper,
+		evmKeeper:  evmKeeper,
+	}, nil
+}
+
+func (Precompile) Address() common.Address {
+	return common.HexToAddress("0x0000000000000000000000000000000000001001")
```
jewei1997: might be good to make this part of the precompile struct.

precompiles/bank/bank.go
```diff
@@ -0,0 +1,152 @@
+package bank
```
jewei1997, line 1: should we have tests for this file?

precompiles/bank/Bank.sol
```diff
@@ -0,0 +1,24 @@
...
+IBank constant BANK_CONTRACT = IBank(
+    BANK_PRECOMPILE_ADDRESS
+);
+
+interface IBank {
+    // Transactions
+    function send(
+        address fromAddress,
+        address toAddress,
+        string memory denom,
```
jewei1997, line 15: for my education: what are some examples of what denom could be?

x/evm/keeper/testdata/SendAll.sol
```diff
@@ -0,0 +1,15 @@
...
+import "../../../../precompiles/bank/Bank.sol";
+
+contract SendAll {
+    function sendAll(
+        address fromAddress,
+        address toAddress,
+        string memory denom
+    ) public {
+        uint256 amount = BANK_CONTRACT.balance(fromAddress, denom);
+        BANK_CONTRACT.send(fromAddress, toAddress, denom, amount);
```
jewei1997, line 13: Can possibly emit an event here to make indexing easier in the future.

jewei1997 (approved):

precompiles/bank/Bank.sol
codchen: It'd be one of `usei`, `ibc/<route1>/<route2>/....` tokens, `tokenfactory/xyz...` tokens

precompiles/bank/bank.go
codchen: updated

precompiles/bank/bank.go
codchen: added

PR #1093, opened 2023-10-09
[EVM] add API platform

codchen:
Add helpers to initialize RPC and WS servers for EVM endpoints. The following files are directly copied from go-ethereum:
- evmrpc/jwt_handlers.go
- evmrpc/endpoints.go
- evmrpc/rpcstack.go

The main entrance point of the package is the `NewEVMXXXServer` functions in `server.go`, which will be called inside `app.go` once we are ready for integration.

A healthcheck endpoint `echo` is added to the server and tested to work for both http and websocket.

Support for authenticated servers (i.e. https and wss) will be added later.

unit tests


evmrpc/echo_test.go
```diff
@@ -0,0 +1,52 @@
...
+const TestPort = 7777
+const TestWSPort = 7778
+
+func TestEcho(t *testing.T) {
+	// Test HTTP server
+	httpServer, err := NewEVMRpcServer(log.NewNopLogger(), TestAddr, TestPort, rpc.DefaultHTTPTimeouts)
+	require.Nil(t, err)
+	require.Nil(t, httpServer.Start())
+	// wait for a second in case the actual server goroutine isn't ready yet
+	time.Sleep(1)
```
jewei1997: I think this is actually 1 nanosecond.

evmrpc/endpoints.go
```diff
@@ -0,0 +1,96 @@
...
+	httpSrv := &http.Server{
+		Handler:           handler,
+		ReadTimeout:       timeouts.ReadTimeout,
+		ReadHeaderTimeout: timeouts.ReadHeaderTimeout,
+		WriteTimeout:      timeouts.WriteTimeout,
+		IdleTimeout:       timeouts.IdleTimeout,
+	}
+	go func() {
+		if err := httpSrv.Serve(listener); err != nil {
+			fmt.Printf("server existing due to %s\n", err)
```
jewei1997: nit: "exiting"

evmrpc/echo_test.go
```diff
@@ -0,0 +1,52 @@
...
+	// Test WS server
+	wsServer, err := NewEVMWebSocketServer(log.NewNopLogger(), TestAddr, TestWSPort, []string{"localhost"}, rpc.DefaultHTTPTimeouts)
+	require.Nil(t, err)
+	require.Nil(t, wsServer.Start())
+	// wait for a second in case the actual server goroutine isn't ready yet
+	time.Sleep(1)
+	headers := make(http.Header)
+	headers.Set("Origin", "localhost")
+	headers.Set("Content-Type", "application/json")
+	conn, _, err := websocket.DefaultDialer.Dial(fmt.Sprintf("ws://%s:%d", TestAddr, TestWSPort), headers)
```
jewei1997, line 47: nit: require this err to be nil

evmrpc/server.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"github.com/ethereum/go-ethereum/rpc"
+	"github.com/tendermint/tendermint/libs/log"
+)
+
+type EVMServer interface {
+	Start() error
+}
+
+func NewEVMRpcServer(
```
jewei1997: might be better to call this `NewEVMHttpServer` since they both use RPC

jewei1997: Also, not sure if we'll need this but seems like it would be hard to make an HTTPServer with both HTTP and WS enabled on the same server since each one creates a new HTTP server.

jewei1997 (approved):
overall LGTM!

evmrpc/endpoints.go
```diff
@@ -0,0 +1,96 @@
...
+			if name != rpc.MetadataApi && name != rpc.EngineApi {
+				bad = append(bad, name)
+			}
+		}
+	}
+	return bad, available
+}
+
+// CheckTimeouts ensures that timeout values are meaningful
+func CheckTimeouts(timeouts *rpc.HTTPTimeouts) {
```
philipsu522, line 79: if this is oss, we can leave, but if not we may want to just panic so there are no unexpected "config changes" when they are set incorrectly

philipsu522 (approved):

evmrpc/endpoints.go
codchen: yeah this is original impl

evmrpc/echo_test.go
codchen: updated

evmrpc/endpoints.go
codchen: fixed

evmrpc/server.go
codchen: updated

evmrpc/echo_test.go
codchen: added

PR #1094, opened 2023-10-10
[EVM] RPC endpoint getBlockByHash

codchen:
Implement `getBlockByHash` endpoint by delegating to corresponding Tendermint endpoint and `x/evm`'s transaction receipt.

unit tests


evmrpc/block.go
```diff
@@ -0,0 +1,109 @@
...
+		gasLimit += txRes.GasWanted
+		gasWanted += txRes.GasUsed
+		decoded, err := txDecoder(txRes.Data)
+		if err != nil {
+			fmt.Println(err)
+			return nil, errors.New("failed to decode transaction")
+		}
+		if len(decoded.GetMsgs()) != 1 {
+			// EVM message must have exactly one message
+			continue
```
philipsu522, line 67: why not panic? or do we assume it's possible that we don't get strictly EVM msgs? if the latter, this seems quite brittle - is there a more robust way to detect for EVM msgs?

evmrpc/block.go
```diff
@@ -0,0 +1,109 @@
...
+				continue
+			}
+			transactions = append(transactions, hydrateTransaction(ethtx, number, blockhash, receipt))
+		}
+	}
+	result := map[string]interface{}{
+		"number":           (*hexutil.Big)(number),
+		"hash":             blockhash,
+		"parentHash":       lastHash,
+		"nonce":            ethtypes.BlockNonce{}, // inapplicable to Sei
```
philipsu522, line 89: as a meta point, i think we want to start documenting what is different b/w EVM on Sei vs EVM on ethereum? perhaps the PRD /RFC is a good place to start adding this, since these will inevitably come up later on and will be important for public facing documentation

philipsu522 (approved):

evmrpc/block.go
codchen: the block may contain non-EVM transactions which would have more than 1 message. There are ante handlers checking for this so we won't have a block containing txs that have more than 1 message while one of the messages is EVM message

evmrpc/block.go
codchen: yeah good point. Added a section in https://docs.google.com/document/d/1WWagJLbUM_egr8gG_jhOHCADw2K11qgq83SWzPfZMzM/edit#heading=h.z0meha1ix02p

PR #1098, opened 2023-10-11
[EVM] add getBlockByNumber endpoint

codchen:
Add endpoint to get block by number (height). Note that the endpoint takes in several special "number" as input:
- earliest -> corresponding to genesis height
- safe/pending/finalized/latest -> all corresponding to latest height since Sei has instant finality

unit tests

### DIFF
--- a/evmrpc/block_test.go
+++ b/evmrpc/block_test.go
@@ -47,7 +47,7 @@ type MockClient struct {
 	mock.Client
 }
 
-func (c *MockClient) BlockByHash(ctx context.Context, hash bytes.HexBytes) (*coretypes.ResultBlock, error) {
+func (c *MockClient) mockBlock() *coretypes.ResultBlock {
 	return &coretypes.ResultBlock{
 		BlockID: tmtypes.BlockID{
 			Hash: bytes.HexBytes([]byte("0000000000000000000000000000000000000000000000000000000000000001")),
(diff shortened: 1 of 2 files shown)


evmrpc/block_test.go
```diff
@@ -109,13 +136,9 @@ func TestGetBlockByHash(t *testing.T) {
 	privKey := hd.Secp256k1.Generate()(derivedPriv)
 	testPrivHex := hex.EncodeToString(privKey.Bytes())
 	key, _ := crypto.HexToECDSA(testPrivHex)
-	evmParams := k.GetParams(ctx)
-	evmParams.ChainConfig.CancunTime = 0 // overwrite to enable cancun
```
philipsu522, line 113: is this not needed anymore?

evmrpc/block_test.go
```diff
@@ -146,3 +164,52 @@ func TestGetBlockByHash(t *testing.T) {
...
+	require.Nil(t, err)
+	b := TxConfig.NewTxBuilder()
+	b.SetMsgs(msg)
+	Tx = b.GetTx()
+	require.Nil(t, EVMKeeper.SetReceipt(Ctx, tx.Hash(), &types.Receipt{
+		From:             "56789",
+		TransactionIndex: 5,
+	}))
+
+	for _, num := range []string{"0x8", "earliest", "latest", "pending", "finalized", "safe"} {
```
philipsu522, line 204: can we test invalid too?

philipsu522 (approved):

evmrpc/block_test.go
codchen: yeah we don't need this here; it was copied from the ante handler test which tests for different versions

evmrpc/block_test.go
codchen: added

PR #1100, opened 2023-10-12
[EVM] add getTransactionReceipt

codchen:
Add endpoint to return transaction receipts per Ethereum spec

unit tests


evmrpc/tx_test.go
```diff
@@ -0,0 +1,50 @@
...
+		TxType:            1,
+		ContractAddress:   "0x123456789012345678902345678901234567890",
+		CumulativeGasUsed: 123,
+		TxHashHex:         "0x123456789012345678902345678901234567890123456789012345678901234",
+		GasUsed:           55,
+		Status:            0,
+		EffectiveGasPrice: 10,
+	}))
+
+	body := "{\"jsonrpc\": \"2.0\",\"method\": \"eth_getTransactionReceipt\",\"params\":[\"0x1234567890123456789012345678901234567890123456789012345678901234\"],\"id\":\"test\"}"
```
jewei1997, line 34: you can consider using go ethereum's eth client to wrap your json rpc calls instead of writing raw json. https://pkg.go.dev/github.com/ethereum/go-ethereum/ethclient#Client.TransactionReceipt

evmrpc/tx_test.go
```diff
@@ -0,0 +1,50 @@
...
...[\"0x1111111111111111111111111111111111111111111111111111111111111111\",\"0x1111111111111111111111111111111111111111111111111111111111111112\"],\"data\":\"0x78797a\",\"blockNumber\":\"0x8\",\"transactionHash\":\"0x1111111111111111111111111111111111111111111111111111111111111113\",\"transactionIndex\":\"0x2\",\"blockHash\":\"0x1111111111111111111111111111111111111111111111111111111111111111\",\"logIndex\":\"0x1\",\"removed\":true}],\"logsBloom\":\"0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000\",\"status\":\"0x0\",\"to\":\"0x0123456789012345678902345678901234567890\",\"transactionHash\":\"0x0123456789012345678902345678901234567890123456789012345678901234\",\"transactionIndex\":\"0x0\",\"type\":\"0x1\"}}\n", string(resBody))
```
jewei1997: Is there a better way to check the resBody results instead of checking if it is equal to raw json?

philipsu522 (approved):

x/evm/keeper/msg_server.go
```diff
@@ -131,13 +124,21 @@ func (server msgServer) writeReceipt(ctx sdk.Context, tx *ethtypes.Transaction,
...
 		TxHashHex:         tx.Hash().Hex(),
-		ContractAddress:   contractAddr.Hex(),
 		GasUsed:           usedGas,
 		BlockNumber:       uint64(ctx.BlockHeight()),
 		TransactionIndex:  uint32(ctx.TxIndex()),
 		EffectiveGasPrice: tx.GasPrice().Uint64(),
 	}
 
+	if msg.To == nil {
+		receipt.ContractAddress = crypto.CreateAddress(msg.From, msg.Nonce).Hex()
```
philipsu522, line 134: is it expected that a nil "To" has the side effect of creating an address?

codchen: yeah it's a `contract deployment transaction`: https://ethereum.org/en/developers/docs/transactions/#types-of-transactions

evmrpc/tx_test.go
codchen: I think this client is a bit different from the rpc spec? It returns a Receipt https://pkg.go.dev/github.com/ethereum/go-ethereum@v1.13.3/core/types#Receipt which is missing fields like `from` and `to` https://github.com/ethereum/go-ethereum/blob/v1.13.3/internal/ethapi/api.go#L1820-L1821

evmrpc/tx_test.go
codchen: changed it to be unmarshaled into a map first and check each field individually

PR #1103, opened 2023-10-16
[EVM] add endpoint to get transaction count by address

codchen:
Implement `eth_getTransactionCount` which returns the number of transactions sent by a given account in a specific block

unit tests


jewei1997 (approved):
LGTM!

PR #1105, opened 2023-10-17
[EVM] eth state endpoints

jewei1997:
Implemented GetBalance, GetCode, and GetStorageAt. Skipped GetProof for now due to its implementation complexity.

unit tests.


evmrpc/state.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to decode storage key: %s", err)
+	}
+	state := a.keeper.GetState(a.ctxProvider(), address, key)
+	return state[:], nil
+}
+
+// decodeHash parses a hex-encoded 32-byte hash. The input may optionally
+// be prefixed by 0x and can have a byte length up to 32.
+func decodeHash(s string) (h common.Hash, inputLength int, err error) {
```
codchen, line 75: is this the same as `common.HexToHash`, or there is some subtle difference?

evmrpc/state.go
```diff
@@ -0,0 +1,85 @@
...
+func (a *StateAPI) GetBalance(ctx context.Context, address common.Address, blockNr rpc.BlockNumber) (*uint64, error) {
+	block, err := getBlockNumber(ctx, a.tmClient, blockNr)
+	if err != nil {
+		return nil, err
+	}
+	if block != nil {
+		return nil, errors.New("block number not safe, finalized, latest, or pending")
+	}
+
+	balance := a.keeper.GetBalance(a.ctxProvider(), address)
```
codchen: `GetBalance` gets balance stored in `evm` module - only accounts that haven't established the association between their evm address and their sei address will have their balance stored in `evm` module. Otherwise balance is stored in the bank module. So we need to first call `a.keeper.GetSeiAddress` to check if there is an association, and use `a.keeper.BankKeeper().GetBalance(ctx, seiAddr, a.keeper.GetBaseDenom(ctx))` to get the bank balance of the account if so. If there is no association then we use `a.keeper.GetBalance`

evmrpc/state.go
jewei1997: Yes, this was taken from go-ethereum here(https://github.com/ethereum/go-ethereum/blame/da55b23d21fb297fd01cfc1bebb13a3e8fe7f726/internal/ethapi/api.go#L750). It is quite similar, but the subtle difference is that decodeHash will return an error for invalid hex in storage keys whereas `HexToHash` will always resolve to some hash. The PR for that change in go-ethereum is here (https://github.com/ethereum/go-ethereum/commit/8ade5e6c144afb7ac389ae28b50cb68a0dbec7b8)

evmrpc/state.go
jewei1997: fixed: https://github.com/sei-protocol/sei-chain/blob/8f113424c60cb4866e9a374572f7d36e00313a69/evmrpc/state.go#L35-L43

codchen (approved):

PR #1106, opened 2023-10-18
[EVM] add endpoint to return fee history

codchen:
Add `feeHistory` endpoints that returns base fee and rewards percentile (weighted by gas used) for the requested block range. More details on the percentile calculation can be found in comments.

unit tests

### DIFF
--- a/evmrpc/info.go
+++ b/evmrpc/info.go
@@ -1,0 +1,143 @@
+package evmrpc
+
+import (
+	"context"
+	"errors"
+	"math/big"
+	"slices"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/common/math"
+	"github.com/ethereum/go-ethereum/rpc"
+	"github.com/sei-protocol/sei-chain/x/evm/keeper"
+	rpcclient "github.com/tendermint/tendermint/rpc/client"
+	"github.com/tendermint/tendermint/rpc/coretypes"
+)
+
+type InfoAPI struct {
+	tmClient    rpcclient.Client
+	keeper      *keeper.Keeper
+	ctxProvider func(int64) sdk.Context
+	txDecoder   sdk.TxDecoder
+}
+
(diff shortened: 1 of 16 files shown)


evmrpc/info.go
```diff
@@ -0,0 +1,143 @@
...
+	GasUsedRatio []float64        `json:"gasUsedRatio"`
+}
+
+// lastBlock is inclusive
+func (i *InfoAPI) FeeHistory(ctx context.Context, blockCount math.HexOrDecimal64, lastBlock rpc.BlockNumber, rewardPercentiles []float64) (*FeeHistoryResult, error) {
+	result := FeeHistoryResult{}
+
+	// validate reward percentiles
+	for i, p := range rewardPercentiles {
+		if p < 0 || p > 100 || (i > 0 && p < rewardPercentiles[i-1]) {
```
philipsu522, line 42: - should these be <= and >= respectively?
- do we allow for duplicates?
- do we allow integers only or floats are ok? we can validate here as well

evmrpc/info.go
```diff
@@ -0,0 +1,143 @@
...
+		return int(a.reward) - int(b.reward)
+	})
+	res := []*hexutil.Big{}
+	if len(gasAndRewards) == 0 {
+		return res
+	}
+	var txIndex int
+	sumGasUsed := gasAndRewards[0].gasUsed
+	for _, p := range rewardPercentiles {
+		thresholdGasUsed := uint64(float64(totalEVMGasUsed) * p / 100)
```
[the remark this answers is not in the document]
philipsu522: should we use float32 just to be safe?

philipsu522 (approved):

evmrpc/info.go
codchen: these checks are copied from https://github.com/ethereum/go-ethereum/blob/master/eth/gasprice/feehistory.go#L227-L232 so to preserve compatibility i think we should just keep them as-is?

codchen:
@philipsu522 ^that's also copied from go-etherem https://github.com/sei-protocol/sei-chain/pull/1106#discussion_r1364377900 Also why would float32 be safer?

PR #1108, opened 2023-10-19
[EVM] add endpoint to send transaction

codchen:
Implement `SendRawTransaction`. Note that for now Sei nodes will not custody users' private keys so it will not offer endpoints that provide signing capability (e.g. SendTransaction, Sign, SignTransaction)

unit tests

### DIFF
--- a/evmrpc/send.go
+++ b/evmrpc/send.go
@@ -1,0 +1,51 @@
+package evmrpc
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	ethtypes "github.com/ethereum/go-ethereum/core/types"
+	"github.com/sei-protocol/sei-chain/x/evm/types"
+	"github.com/sei-protocol/sei-chain/x/evm/types/ethtx"
+	rpcclient "github.com/tendermint/tendermint/rpc/client"
+)
+
+type SendAPI struct {
+	tmClient rpcclient.Client
+	txConfig client.TxConfig
+}
+
+func NewSendAPI(tmClient rpcclient.Client, txConfig client.TxConfig) *SendAPI {
+	return &SendAPI{tmClient: tmClient, txConfig: txConfig}
+}
+
+func (s *SendAPI) SendRawTransaction(ctx context.Context, input hexutil.Bytes) (common.Hash, error) {
+	tx := new(ethtypes.Transaction)
+	if err := tx.UnmarshalBinary(input); err != nil {
+		return common.Hash{}, err
+	}
+	txdata, err := ethtx.NewTxDataFromTx(tx)
+	if err != nil {
+		return common.Hash{}, err
+	}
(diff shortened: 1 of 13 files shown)


jewei1997:
nit: can possibly remove this: https://github.com/sei-protocol/sei-chain/blob/5c4b9c82f0272c012e1bdbcbded87088f0b49521/evmrpc/testutils.go#L243

evmrpc/send_test.go
```diff
@@ -0,0 +1,55 @@
...
+	ethCfg := evmParams.GetChainConfig().EthereumConfig(big.NewInt(1))
+	signer := ethtypes.MakeSigner(ethCfg, big.NewInt(Ctx.BlockHeight()), uint64(Ctx.BlockTime().Unix()))
+	tx := ethtypes.NewTx(&txData)
+	tx, err := ethtypes.SignTx(tx, signer, key)
+	require.Nil(t, err)
+	bz, err := tx.MarshalBinary()
+	require.Nil(t, err)
+	payload := "0x" + hex.EncodeToString(bz)
+
+	resObj := sendRequestGood(t, "sendRawTransaction", payload)
```
jewei1997, line 42: is there no checking on the validity of the transaction currently? I changed the nonce to some random number for example and the transaction still worked.

evmrpc/send.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	txBuilder := s.txConfig.NewTxBuilder()
+	if err := txBuilder.SetMsgs(msg); err != nil {
+		return common.Hash{}, err
+	}
+	txbz, err := s.txConfig.TxEncoder()(txBuilder.GetTx())
+	if err != nil {
+		return common.Hash{}, err
+	}
+	res, err := s.tmClient.BroadcastTxSync(ctx, txbz)
```
jewei1997: why are we using `BroadcastTxSync` over `BroadcastTx`? What is the difference? It says `BroadcastTxSync` is deprecated. https://github.com/sei-protocol/sei-tendermint/blob/962ce074d5dd07c2ff9a3e81b94755c18481774c/rpc/client/interface.go#L66-L69

evmrpc/send_test.go
codchen: the check happens in a real tendermint client, which will call CheckTx on the transaction https://github.com/sei-protocol/sei-chain/pull/1109

In the unit test we use a mock so there is no check

evmrpc/send.go
codchen: these two are the same: https://github.com/sei-protocol/sei-tendermint/blob/main/internal/rpc/core/mempool.go#L32-L34 I've changed to use BroadcastTx instead

codchen:
> nit: can possibly remove this:
> 
> https://github.com/sei-protocol/sei-chain/blob/5c4b9c82f0272c012e1bdbcbded87088f0b49521/evmrpc/testutils.go#L243

@jewei1997 i think linter would exclude reference from test files and still consider as this deadcode if we remove this nolint :(

jewei1997 (approved):

philipsu522 (approved):

PR #1109, opened 2023-10-19
[EVM] add subscription utility

codchen:
Add a helper class `SubscriptionManager` to facilitate subscription to tendermint events (docs: https://docs.tendermint.com/v0.34/rpc/#/Websocket/subscribe). Filters can be created via the new `QueryBuilder` struct and passed to `SubscriptionManager`.

Note that we make use of `Subscribe` method which is marked as deprecated, but not really deprecated in the new tendermint version.

unit tests

### DIFF
--- a/evmrpc/subscribe_test.go
+++ b/evmrpc/subscribe_test.go
@@ -1,0 +1,42 @@
+package evmrpc
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestQueryBuilder(t *testing.T) {
+	q := mockQueryBuilder()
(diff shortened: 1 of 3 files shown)


evmrpc/subscribe_test.go
```diff
@@ -0,0 +1,38 @@
...
+	expectedQuery := "tm.event = 'Tx' AND evm_log.contract_address = 'test contract' AND evm_log.block_hash = 'block hash' AND evm_log.block_number = '1' AND evm_log.tx_hash = 'tx hash' AND evm_log.tx_idx = '2' AND evm_log.idx = '3' AND evm_log.topics CONTAINS 'topic a' AND evm_log.topics CONTAINS 'topic b'"
+	require.Equal(t, expectedQuery, q.Build())
+}
+
+func TestSubscribe(t *testing.T) {
+	manager := NewSubscriptionManager(&MockClient{})
+	res, err := manager.Subscribe(context.Background(), mockQueryBuilder(), 10)
+	require.Nil(t, err)
+	require.Equal(t, 1, int(res))
+
```
jewei1997, line 21: would be good to make another subscription just to make sure NextId is incremented.

jewei1997 (approved):
overall LGTM!

evmrpc/subscribe_test.go
codchen: updated

philipsu522 (approved):

PR #1111, opened 2023-10-23
Add occ to sei-chain

stevenlanders:
- This is against `occ-main` (adds latest cosmos)
- Invokes the new `DeliverTxBatch` interface 

- In flight (will be via load test cluster)


app/app.go
```diff
@@ -1012,6 +1016,9 @@ func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcess
 }
 
 func (app *App) FinalizeBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlock) (*abci.ResponseFinalizeBlock, error) {
+	//TODO: update with logic that asserts that occ is enabled
+	ctx = ctx.WithIsOCCEnabled(true)
```
udpatil: can we move this to the `process block` function so its not in FinalizeBlocker and ProcessProposalHandler?

udpatil (approved):

PR #1112, opened 2023-10-24
query tendermint for events

codchen:


jewei1997:
merged in along with https://github.com/sei-protocol/sei-chain/pull/1116

PR #1115, opened 2023-10-26
[EVM] return gas info in message handler correctly

codchen:
There were two issues with gas return value previously:
1. Mempool uses the `gasWanted` field in the return value of `CheckTx` to determine priority. This field is populated based on the gas meter limit that's read from the context returned by ante handlers. So we must set gas meter limit properly when evm ante handlers return.
2. `DeliverTx` will read the `gasUsed` field from the gas meter returned by the message handler and add it to block gas meter. So even though we let EVM handle gas checks internally, we still need to add the consumed gas to the gas meter before the message handler returns.

One thing to note is that, because of the changes made to address 1, the evm transaction handler will no longer have an infinite gas meter when its processing starts, so we need to overwrite it back to an infinite gas meter so that we keep gas management delegated to the EVM internals.

unit tests

### DIFF
--- a/x/evm/ante/gas.go
+++ b/x/evm/ante/gas.go
@@ -1,0 +1,29 @@
+package ante
+
+import (
+	"errors"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	evmkeeper "github.com/sei-protocol/sei-chain/x/evm/keeper"
+	evmtypes "github.com/sei-protocol/sei-chain/x/evm/types"
+)
+
+type GasLimitDecorator struct {
+	evmKeeper *evmkeeper.Keeper
+}
+
+func NewGasLimitDecorator(evmKeeper *evmkeeper.Keeper) *GasLimitDecorator {
+	return &GasLimitDecorator{evmKeeper: evmKeeper}
+}
+
+// Called at the end of the ante chain to set gas limit properly
+func (gl GasLimitDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	txData, found := evmtypes.GetContextTxData(ctx)
+	if !found {
+		return ctx, errors.New("could not find eth tx")
+	}
+
+	adjustedGasLimit := gl.evmKeeper.GetPriorityNormalizer(ctx).MulInt64(int64(txData.GetGas()))
+	ctx = ctx.WithGasMeter(sdk.NewGasMeter(adjustedGasLimit.RoundInt().Uint64()))
+	return next(ctx, tx, simulate)
+}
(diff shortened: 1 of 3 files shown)


x/evm/keeper/msg_server.go
```diff
@@ -28,6 +28,9 @@ var _ types.MsgServer = msgServer{}
 
 func (server msgServer) EVMTransaction(goCtx context.Context, msg *types.MsgEVMTransaction) (serverRes *types.MsgEVMTransactionResponse, err error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
+	originalGasMeter := ctx.GasMeter()
+	ctx = ctx.WithGasMeter(sdk.NewInfiniteGasMeter())
```
philipsu522, line 39: how do other antehandlers do this? is it the same or is evm special cased?

x/evm/ante/gas.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+// Called at the end of the ante chain to set gas limit properly
+func (gl GasLimitDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	txData, found := evmtypes.GetContextTxData(ctx)
+	if !found {
+		return ctx, errors.New("could not find eth tx")
+	}
+
+	adjustedGasLimit := gl.evmKeeper.GetPriorityNormalizer(ctx).MulInt64(int64(txData.GetGas()))
```
philipsu522, line 26: do you mind adding comments as to why this needs to be adjusted?

x/evm/keeper/msg_server.go
codchen: EVM has a special case here, mainly because for an EVM transaction the gas limit is set on EVM payload level, not on top-level `GasWanted` field as normal transactions (because existing eth client can't). As a result EVM has its own dedicated ante handler chain. The full sequence is:
1. At the beginning of the ante handler chain, gas meter is set to infinite so that the ante processing itself won't run out of gas (EVM ante is pretty light but it does read a parameter or two)
2. At the end of the ante handler chain, gas meter is set based on the gas limit specified in the EVM payload; this is only to provide a `GasWanted` return value to tendermint mempool when CheckTx returns, and not used for anything else.
3. At the beginning of message server (here), gas meter is set to infinite again, because EVM internal logic will then take over and manage out-of-gas scenarios.
4. At the end of message server, gas consumed by EVM is adjusted to Sei's unit and counted in the original gas meter, because that original gas meter will be used to count towards block gas after message server returns

codchen: added this as comment in the file

x/evm/ante/gas.go
codchen: added

PR #1116, opened 2023-10-27
[EVM] ETH Log Filter Endpoints

jewei1997:
Implement the following endpoints:
 - eth_getFIlterChanges
 - eth_getLogs
 - eth_getFilterLogs
 - eth_newFilter
 - eth_uninstallFilter

Not implemented: expirations for filters.

unit tests


evmrpc/filter.go
```diff
@@ -0,0 +1,199 @@
...
+
+// pulls logs from tendermint client over multiple addresses.
+func (a *FilterAPI) getLogsOverAddresses(
+	ctx context.Context,
+	crit filters.FilterCriteria,
+	cursors map[common.Address]string,
+) ([]*ethtypes.Log, map[common.Address]string, error) {
+	res := make([]*ethtypes.Log, 0)
+	if len(crit.Addresses) == 0 {
+		crit.Addresses = append(crit.Addresses, common.Address{})
```
codchen, line 167: what's the reason for appending an empty address here?

evmrpc/filter.go
```diff
@@ -0,0 +1,199 @@
...
+	filters := make(map[uint64]filter)
+	return &FilterAPI{tmClient: tmClient, nextFilterID: 1, filters: filters}
+}
+
+func (a *FilterAPI) NewFilter(
+	ctx context.Context,
+	crit filters.FilterCriteria,
+) (*uint64, error) {
+	curFilterID := a.nextFilterID
+	a.nextFilterID++
```
codchen, line 77: we need to protect this section behind a mutex since concurrent calls are possible

evmrpc/filter.go
```diff
@@ -0,0 +1,199 @@
...
+	filter, ok := a.filters[filterID]
+	if !ok {
+		return nil, errors.New("filter does not exist")
+	}
+	noCursors := make(map[common.Address]string)
+	res, cursors, err := a.getLogsOverAddresses(ctx, filter.fc, noCursors)
+	if err != nil {
+		return nil, err
+	}
+	updatedFilter := a.filters[filterID]
```
codchen, line 140: ditto re: mutex (also for all `a.filters` accesses

codchen (commented):
overall looks good except the concurrent access bit. Also the timeout part can be in a separate followup PR given this one is already quite large

evmrpc/filter.go
```diff
@@ -0,0 +1,199 @@
...
+			if err != nil {
+				return nil, "", err
+			}
+			logs = append(logs, ethLog)
+		}
+	}
+	return logs, cursor, nil
+}
+
+func (a *FilterAPI) UninstallFilter(
```
philipsu522, line 256: do we need the error return value here?

evmrpc/filter.go
jewei1997: updated, along with timeouts change

evmrpc/filter.go
jewei1997: updated, along with timeouts change

evmrpc/filter.go
```diff
@@ -0,0 +1,268 @@
...
+	return api
+}
+
+func (a *FilterAPI) timeoutLoop(timeout time.Duration) {
+	ticker := time.NewTicker(timeout)
+	defer ticker.Stop()
+	for {
+		<-ticker.C
+		a.filtersMu.Lock()
+		for id, filter := range a.filters {
```
philipsu522, line 58: will the ordering of iteration here affect state? wondering b/c it could be nondeterministic cc @codchen

evmrpc/filter.go
jewei1997: If the user doesn't pass in an address, we need to search over all addresses. I basically want it to enter the for loop below but with an empty address. Let me know if you see a better way to do this. I tried to think about some other ways of doing this, but I think this might not be a bad way.

evmrpc/filter.go
jewei1997: removed, geth also does not have an error.

evmrpc/filter.go
codchen: this is transient state in server, not stored as chain state

codchen (approved):

PR #1126, opened 2023-11-03
[EVM] add transaction type to associate account

codchen:
Add a transaction type that will store EVM<>Sei address association for an account if both of the following conditions are met:
1. The account doesn't have association set yet
2. The account has at least 1Sei in its balance (won't be charged)

This transaction type can either be sent through EVM RPC server (requires signing by the user), or use a helper command if the private key is stored in seid keyring: `seid tx evm associate-address --from=<key name>`

unit test


evmrpc/send.go
```diff
@@ -22,34 +22,41 @@ func NewSendAPI(tmClient rpcclient.Client, txConfig client.TxConfig) *SendAPI {
...
-	if err := tx.UnmarshalBinary(input); err != nil {
-		return common.Hash{}, err
-	}
-	txdata, err := ethtx.NewTxDataFromTx(tx)
-	if err != nil {
-		return common.Hash{}, err
+func (s *SendAPI) SendRawTransaction(ctx context.Context, input hexutil.Bytes) (hash common.Hash, err error) {
+	var txData ethtx.TxData
+	associateTx := ethtx.AssociateTx{}
+	if associateTx.Unmarshal(input) == nil {
```
philipsu522, line 28: is the idea here that we try to unmarshall the tx as an associateTx first, and if that fails, fallback to EVMTx?

x/evm/ante/preprocess.go
```diff
@@ -2,20 +2,27 @@ package ante
...
 	"github.com/ethereum/go-ethereum/common"
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/params"
 	evmkeeper "github.com/sei-protocol/sei-chain/x/evm/keeper"
 	evmtypes "github.com/sei-protocol/sei-chain/x/evm/types"
+	"github.com/sei-protocol/sei-chain/x/evm/types/ethtx"
 )
 
+// Accounts need to have at least 1Sei to force association. Note that account won't be charged.
```
philipsu522, line 23: is this something enforced at the cosmos layer or a design decision we made? b/c accounts can be registered on chain with as little as 1usei right?

evmrpc/send.go
codchen: @philipsu522 yes, since we can't (also probably don't want to) make `AssociateTx` part of `ethtypes.Transaction` in `go-ethereum` (and thus can't just deserialize it to `ethtypes.Transaction` and check its type), we'll have to potentially deserialize twice

x/evm/ante/preprocess.go
codchen: @philipsu522 this is a decision we make to avoid ddos, since such transaction is gasless

x/evm/ante/preprocess.go
```diff
@@ -39,35 +46,53 @@ func (p EVMPreprocessDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate
...
+		if _, found := p.evmKeeper.GetEVMAddress(ctx, seiAddr); found {
+			return ctx, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "account already has association set")
+		}
+		seiBalance := p.evmKeeper.BankKeeper().GetBalance(ctx, seiAddr, p.evmKeeper.GetBaseDenom(ctx)).Amount
+		evmBalance := new(big.Int).SetUint64(p.evmKeeper.GetBalance(ctx, evmAddr))
+		if new(big.Int).Add(seiBalance.BigInt(), evmBalance).Cmp(new(big.Int).SetUint64(BalanceThreshold)) < 0 {
+			return ctx, sdkerrors.Wrap(sdkerrors.ErrInsufficientFunds, "account needs to have at least 1Sei to force association")
+		}
+		p.evmKeeper.SetAddressMapping(ctx, seiAddr, evmAddr)
+		return ctx.WithPriority(math.MaxInt64 - 101), nil // short-circuit without calling next
```
jewei1997: why do you subtract 101 from MaxInt64 here?

x/evm/types/ethtx/utils.go
```diff
@@ -42,3 +44,13 @@ func ValidateEthTx(tx *ethtypes.Transaction) error {
 	}
 	return nil
 }
+
+func DecodeSignature(sig []byte) (r, s, v *big.Int) {
+	if len(sig) != crypto.SignatureLength {
+		panic(fmt.Sprintf("wrong size for signature: got %d, want %d", len(sig), crypto.SignatureLength))
```
jewei1997: might be good to return an error here instead of panicing.

codchen: updated

x/evm/ante/preprocess.go
codchen: we want it to be high priority but not higher than oracle's. I've refactored it into constants to be more descriptive

PR #1127, opened 2023-11-03
[EVM] Eth block filter endpoint

jewei1997:
Implement NewBlockFilter. Allows user to get back a filter id which they can then use on getFilterChanges to get updates on blocks.

Quicknode documentation on endpoint: https://www.quicknode.com/docs/ethereum/eth_newBlockFilter

unit tests.

### DIFF
--- a/evmrpc/subscribe.go
+++ b/evmrpc/subscribe.go
@@ -48,7 +48,15 @@ type QueryBuilder struct {
 	conditions []string
 }
 
-func NewQueryBuilder() *QueryBuilder {
+func NewBlockQueryBuilder() *QueryBuilder {
+	return &QueryBuilder{
+		conditions: []string{
+			"tm.event = 'NewBlock'",
+		},
+	}
+}
+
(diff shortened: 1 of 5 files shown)


evmrpc/subscribe.go
```diff
@@ -48,7 +48,15 @@ type QueryBuilder struct {
 	conditions []string
 }
 
-func NewQueryBuilder() *QueryBuilder {
+func NewBlockQueryBuilder() *QueryBuilder {
+	return &QueryBuilder{
+		conditions: []string{
+			"tm.event = 'NewBlock'",
```
yzang2019, line 54: Based on the Naming convention, will it be better to be "tm.event = 'Block'" ?

jewei1997: I believe this is actually a tendermint specific query: https://github.com/tendermint/tendermint/blob/85a584dd2ee3437b117d9f5e894eac70bf09aeef/types/events.go#L19

codchen (approved):

PR #1128, opened 2023-11-06
[EVM] convert balance decimal points

codchen:
Arguments to EVM state interface calls are assumed to be denominated with 18 decimal points, so using `usei` as the smallest unit would cause display issue on applications (e.g. 1 Sei would show up as 10^-12 <display unit> on MetaMask). This PR changes EVM state implementation to treat the smallest unit as `S-wei`, where 1 `S-wei` = `10^-12 usei`. It then converts `S-wei` to `usei` internally before it stores to bank balance (which is still using `usei`).

Also added a utility command `seid tx evm send` to allow easier token sending in EVM addresses (bank send is only for sei addresses) 

local integration with MetaMask


evmrpc/info.go
```diff
@@ -124,7 +124,7 @@ func (i *InfoAPI) FeeHistory(ctx context.Context, blockCount math.HexOrDecimal64
 
 type GasAndReward struct {
 	GasUsed uint64
-	Reward  uint64
```
jewei1997, line 127: why was this converted to a big int?

x/evm/client/cli/tx.go
```diff
@@ -103,3 +107,88 @@ func CmdAssociateAddress() *cobra.Command {
...
+				return err
+			}
+			nonce, err := strconv.ParseUint(args[2], 10, 64)
+			if err != nil {
+				return err
+			}
+			txData := ethtypes.DynamicFeeTx{
+				Nonce:     nonce,         // TODO query from server
+				GasFeeCap: big.NewInt(1), // TODO use flag
+				Gas:       21000,         // TODO use flag
```
jewei1997: why are these left as TODOs?

x/evm/client/cli/tx.go
```diff
@@ -103,3 +107,88 @@ func CmdAssociateAddress() *cobra.Command {
...
+				return err
+			}
+			bz, err := tx.MarshalBinary()
+			if err != nil {
+				return err
+			}
+			payload := "0x" + hex.EncodeToString(bz)
+
+			body := fmt.Sprintf("{\"jsonrpc\": \"2.0\",\"method\": \"eth_sendRawTransaction\",\"params\":[\"%s\"],\"id\":\"send\"}", payload)
+			req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("http://%s:8545", evmrpc.LocalAddress), strings.NewReader(body))
```
jewei1997: should we allow the user to specify their own rpc endpoint?

jewei1997 (approved):
left some comments but overall LGTM!

evmrpc/info.go
codchen: because GasPrice will also be in `sei-wei` which might cause overflow with uint64

x/evm/client/cli/tx.go
codchen: updated

x/evm/client/cli/tx.go
codchen: updated

PR #1129, opened 2023-11-06
[EVM] add ACL dependencies

codchen:
Add ACL dependencies for EVM transactions. Note that it only contains non-wildcard matching for `to` address, since `from` address must be derived from signature which is a costly process that only happens during ante processing (i.e. after dependency evaluation). All contract state accesses are covered by wildcards.

unit tests

### DIFF
--- a/aclmapping/evm/mappings.go
+++ b/aclmapping/evm/mappings.go
@@ -1,0 +1,95 @@
+package evm
+
+import (
+	"encoding/hex"
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkacltypes "github.com/cosmos/cosmos-sdk/types/accesscontrol"
+	aclkeeper "github.com/cosmos/cosmos-sdk/x/accesscontrol/keeper"
+	acltypes "github.com/cosmos/cosmos-sdk/x/accesscontrol/types"
+	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
+	evmkeeper "github.com/sei-protocol/sei-chain/x/evm/keeper"
+	evmtypes "github.com/sei-protocol/sei-chain/x/evm/types"
+)
+
+var ErrInvalidMessageType = fmt.Errorf("invalid message received for EVM Module")
+
+func GetEVMDependencyGenerators(evmKeeper evmkeeper.Keeper) aclkeeper.DependencyGeneratorMap {
+	dependencyGeneratorMap := make(aclkeeper.DependencyGeneratorMap)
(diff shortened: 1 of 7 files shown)


aclmapping/evm/mappings.go
```diff
@@ -0,0 +1,95 @@
...
+	toAddress := tx.To()
+	seiAddress, associated := evmKeeper.GetSeiAddress(ctx, *toAddress)
+	var idTempl string
+	var resourceType sdkacltypes.ResourceType
+	if associated {
+		idTempl = hex.EncodeToString(banktypes.CreateAccountBalancesPrefix(seiAddress))
+		resourceType = sdkacltypes.ResourceType_KV_BANK_BALANCES
+	} else {
+		idTempl = hex.EncodeToString(evmtypes.BalanceKey(*toAddress))
+		resourceType = sdkacltypes.ResourceType_KV_EVM_BALANCE
```
udpatil, line 47: @codchen you need to update the github workflow to not run forward compatibility on `evm` branch (similar to how we did for occ), that way you can continue to have a working/accurate CI

aclmapping/evm/mappings.go
```diff
@@ -0,0 +1,95 @@
...
+		},
+		{
+			AccessType:         sdkacltypes.AccessType_READ,
+			ResourceType:       sdkacltypes.ResourceType_KV_BANK,
+			IdentifierTemplate: "*", // from address access (reasoning described above)
+		},
+		{
+			AccessType:         sdkacltypes.AccessType_WRITE,
+			ResourceType:       sdkacltypes.ResourceType_KV_BANK,
+			IdentifierTemplate: "*", // from address access (reasoning described above)
```
udpatil, line 79: ideally we'd want to know a sepcific bank balance key for the "from" field as well, is there another way to get the from addr from the derived signature? it think for now its fine if we exclude it, but would be nice to have down the road

codchen: @udpatil it should be possible to frontload stateless ante operations like signature verification (which would give us the from address) to happen before ACL is evaluated, but that'd be a major refactor

codchen: or, we allow `CheckTx` to modify tx bytes by settings fields that are derived during CheckTx ante run. Not sure if that'd have any security implication or not

udpatil: yeah i think its fine to leave as is for now, maybe in the future when we look to add the pre cached dependencies from checkTX or something we can bundle this in there

udpatil (approved):

PR #1130, opened 2023-11-06
[EVM] surface VM error

codchen:
Add a new endpoint `eth_getTransactionErrorByHash` to surface VM error to callers.
Also added panics for unexpected errors in `msg_server`.

unit tests


evmrpc/tx.go
```diff
@@ -84,6 +84,14 @@ func (t *TransactionAPI) GetTransactionByHash(ctx context.Context, hash common.H
 	return t.GetTransactionByBlockNumberAndIndex(ctx, rpc.BlockNumber(receipt.BlockNumber), hexutil.Uint(receipt.TransactionIndex)), nil
 }
 
+func (t *TransactionAPI) GetTransactionErrorByHash(ctx context.Context, hash common.Hash) (string, error) {
```
jewei1997: this seems nonstandard, is it needed to surface errors to metamask users?

jewei1997 (approved):

evmrpc/tx.go
codchen: it is more for developers' debugging usage

PR #1132, opened 2023-11-07
[EVM] subscription based eth endpoints

jewei1997:
Implementations of eth_subscribe for newHeads and logs. newPendingTransactions is not supported because it is hard to get transactions from tendermint's mempool.

unit tests + local testing for newHeads


evmrpc/subscribe.go
```diff
@@ -4,138 +4,209 @@ import (
...
-func NewBlockQueryBuilder() *QueryBuilder {
-	return &QueryBuilder{
-		conditions: []string{
-			"tm.event = 'NewBlock'",
-		},
-	}
-}
+	switch eventName {
+	case "newHeads":
+		subscriberID, subCh, err := a.subscriptionManager.Subscribe(ctx, NewHeadQueryBuilder(), 100)
```
codchen: let's make the limi a config instead of hardcoding

evmrpc/subscribe.go
```diff
@@ -4,138 +4,209 @@ import (
...
+			return nil, err
+		}
+
+		go func() {
+			for {
+				select {
+				case res := <-subCh:
+					ethHeader, err := encodeTmHeader(res.Data.(*tmtypes.EventDataNewBlockHeader))
+					if err != nil {
+						_ = a.subscriptionManager.Unsubscribe(ctx, subscriberID)
```
codchen: nit: can just do `defer _ = a.subscriptionManager.Unsubscribe(ctx, subscriberID)` immediately after `go func(){`

evmrpc/subscribe.go
```diff
@@ -4,138 +4,209 @@ import (
...
+					return
+				case <-notifier.Closed():
+					_ = a.subscriptionManager.Unsubscribe(ctx, subscriberID)
+					return
+				}
+			}
+		}()
+		return rpcSub, nil
+	case "logs":
+		resultEventAllAddrs := make(chan coretypes.ResultEvent)
```
codchen: is it intentional to use unbuffered channel here?

jewei1997: updated to buffered channel of sufficient capacity.

evmrpc/subscribe.go
jewei1997: updated.

evmrpc/subscribe.go
jewei1997: updated.

PR #1135, opened 2023-11-08
[dex] Fix dex memstate to use KVstore instead of in memory structures

udpatil:
This refactors the in memory structs of contractsToDependencies map and the contracts to process set and moves them under the in memory KV store. This enables appropriate rolling back of the state when invalidating a transaction execution.

tested on loadtest cluster


x/dex/cache/cache.go
```diff
@@ -17,18 +17,16 @@ const SynchronizationTimeoutInSeconds = 5
 type MemState struct {
 	storeKey sdk.StoreKey
 
-	contractsToProcess      *datastructures.SyncSet[string]
-	contractsToDepsMtx      *sync.Mutex
-	contractsToDependencies *datastructures.TypedSyncMap[string, []string]
+	// TODO: all of these need to be in the memstore
+	contractsToProcessMtx *sync.RWMutex
```
codchen, line 21: do we still need these mutex? i think we have mutex in front of kvstore.Set/Get etc., or maybe that's just the iavl store not the memstore?

codchen (approved):

x/dex/cache/cache.go
udpatil: Yeah i think we _should_ be able to remove these mutexes, I wanted to leave them for safety since i wasnt confident in the expected behavior of the concurrent processing for dex end block. I can add a todo to consider removing them, but don't want to remove them at the moment so that any issues can be more clearly attributed to OCC issues

PR #1136, opened 2023-11-10
[EVM] several bug fixes

codchen:
Fixes the following:
- `rpc` flag conflicts with existing flag name. Changing it to `evm-rpc`
- `to` address might be nil in the case of contract creation. Handle it properly in dependency generator
- Exclude `AssociateTx` from EthereumTx decoding since it's not standard Ethereum TX but a Sei-specific utility type
- Return pending transaction (i.e. unconfirmed transactions in mempool) in `getTransactionByHash`

local sei


jewei1997 (approved):
thanks for the fixes!

PR #1139, opened 2023-11-13
[EVM] add `wasmd` precompile

codchen:
Add a `wasmd` precompile that allows EVM contracts to access the following CosmWasm functions for any CW contract:
- `instantiate`
- `execute`
- `query`

unit test


evmrpc/state_test.go
```diff
@@ -193,46 +191,3 @@ func TestGetStorageAt(t *testing.T) {
 		})
 	}
 }
-
-func TestGetProof(t *testing.T) {
```
jewei1997, line 197: why was this deleted?

example/cosmwasm/echo/src/contract.rs
```diff
@@ -0,0 +1,45 @@
+#[cfg(not(feature = "library"))]
```
jewei1997, line 1: what are these cosmwasm contracts for?

testutil/keeper/evm.go
```diff
@@ -3,82 +3,37 @@ package keeper
...
-	"github.com/sei-protocol/sei-chain/x/evm/types"
 )
 
-func MockEVMKeeper() (*evmkeeper.Keeper, *paramskeeper.Keeper, sdk.Context) {
-	evmStoreKey := sdk.NewKVStoreKey(types.StoreKey)
-	authStoreKey := sdk.NewKVStoreKey(authtypes.StoreKey)
-	bankStoreKey := sdk.NewKVStoreKey(banktypes.StoreKey)
-	stakingStoreKey := sdk.NewKVStoreKey(stakingtypes.StoreKey)
-	keyParams := sdk.NewKVStoreKey(typesparams.StoreKey)
-	tKeyParams := sdk.NewTransientStoreKey(typesparams.TStoreKey)
```
jewei1997, line 37: why are these no longer needed?

jewei1997 (approved):
very cool, LGTM!

precompiles/setup.go
```diff
@@ -1,29 +1,50 @@
...
+	addPrecompileToVM(bankp, bankp.Address())
+	wasmdp, err := wasmd.NewPrecompile(evmKeeper, wasmdKeeper, wasmdViewKeeper)
+	if err != nil {
+		return err
+	}
+	addPrecompileToVM(wasmdp, wasmdp.Address())
 	return nil
 }
+
+// This function modifies global variable in `vm` module. It should only be called once
```
philipsu522, line 41: should we follow a singleton pattern here to ensure it's only initialized once (or something like baseapp.sealed where it panics if called multiple times)?

precompiles/wasmd/wasmd_test.go
```diff
@@ -0,0 +1,125 @@
...
+	evm := vm.EVM{
+		StateDB: statedb,
+	}
+	res, err := p.Run(&evm, append(p.QueryID, args...))
+	require.Nil(t, err)
+	outputs, err := queryMethod.Outputs.Unpack(res)
+	require.Nil(t, err)
+	require.Equal(t, 1, len(outputs))
+	require.Equal(t, "{\"message\":\"query test\"}", string(outputs[0].([]byte)))
+}
```
philipsu522, line 187: thoughts on adding more test cases around failure/edge cases? Most of these look happy path, and the added tests would help ensure wasmd.go doesn't have different behavior when we modify

evmrpc/state_test.go
codchen: there was a bug in how stores in tests were initialized that were fixed in this PR, and apparently GetProof would only work with the presence of that bug, so GetProof is in reality broken. I've added a comment in GetProof to fix it in a later PR given this one is already too large.

testutil/keeper/evm.go
codchen: they are initialized more properly via `testApp := Setup(false)` above

precompiles/setup.go
codchen: made it panic if already initialized

precompiles/wasmd/wasmd_test.go
codchen: yeah we should

codchen: added

PR #1140, opened 2023-11-14
[EVM] add bank metadata precompile functions

codchen:
Add metadata functions to return the following info:
- token name
- token symbol
- token decimals (always returns 0 because all Sei native tokens are integer based)
- total supply

unit tests


jewei1997 (approved):
LGTM!

PR #1146, opened 2023-11-16
EVM ERC20 Template Contract

jewei1997:

Created ERC20 Template contract to allow native sei tokens to be used as ERC20s

foundry unit tests + manual testing.


contracts/src/Counter.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.13;
+
+contract Counter {
```
codchen: do we need this counter contract?

contracts/src/NativeSeiTokensERC20.sol
```diff
@@ -0,0 +1,45 @@
...
+
+contract NativeSeiTokensERC20 is ERC20 {
+
+    address constant BANK_PRECOMPILE_ADDRESS = 0x0000000000000000000000000000000000001001;
+
+    string public denom;
+    IBank public BankPrecompile;
+
+    constructor(string memory denom_) ERC20("", "") {
+        BankPrecompile = IBank(BANK_PRECOMPILE_ADDRESS);
```
codchen, line 16: just curious: how much space would it take to store `BankPrecompile` as a state? Is it just an address?

contracts/src/NativeSeiTokensERC20.sol
```diff
@@ -0,0 +1,45 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import {IBank} from "./precompiles/IBank.sol";
+import {console2} from "forge-std/Test.sol";
+
+contract NativeSeiTokensERC20 is ERC20 {
```
codchen, line 8: `is ERC20` gives us the sending and approval functionalities i suppose?

codchen (commented):
overall looks good. We can test it on a local sei before merging:
1. create a tokenfactory denom: `seid tx tokenfactory create-denom test`
2. mint some tokens for an account: `seid tx tokenfactory mint`
3. instantiate an ERC-20 contract for the denom
4. send tokens from the minted account to another account via ERC-20
5. verify the receiver received the tokens

philipsu522 (approved):

contracts/src/NativeSeiTokensERC20.sol
jewei1997: yes it is the syntax to do inheritance.

contracts/src/NativeSeiTokensERC20.sol
jewei1997: yes it should just be an address.

contracts/src/Counter.sol
jewei1997: nope, removed.

PR #1158, opened 2023-11-28
[EVM] preprocess EVM transactions and define granular access operations

codchen:
The goal of this PR is to define granular access operations for EVM transactions. Since EVM contract's logic is opaque to the chain, the access operation definition only covers the common components of every EVM transactions (e.g. transfer of value), and is exclusive of any wildcard (*) definitions. In order to be able to define such operations, a few refactors are required so that all necessary information are available at the time of dependency evaulation:
- frontloaded preprocessing logic that derive sender address/public keys from ante handler into a dedicated `Preprocess` function that is called before dependency graph is built
- refactored `DeliverTx` method signature so that we can pass a decoded `sdk.Tx` around after it's first decoded instead of decoding multiple times. This allows derived information (like those from `Preprocess` calls) to be passed around.
- transient states in `statedb` now needs to be indexed by tx index since multiple transactions may run at the same time

unit tests & local sei testing


aclmapping/evm/mappings.go
```diff
@@ -35,66 +39,136 @@ func TransactionDependencyGenerator(_ aclkeeper.Keeper, evmKeeper evmkeeper.Keep
...
+		},
+		{
+			AccessType:         sdkacltypes.AccessType_READ,
+			ResourceType:       sdkacltypes.ResourceType_KV_EVM_S2E,
+			IdentifierTemplate: hex.EncodeToString(evmtypes.SeiAddressToEVMAddressKey(evmMsg.Derived.SenderSeiAddr)),
+		},
+		{
+			AccessType:         sdkacltypes.AccessType_READ,
+			ResourceType:       sdkacltypes.ResourceType_KV_EVM_CODE_HASH,
+			IdentifierTemplate: hex.EncodeToString(append(evmtypes.CodeHashKeyPrefix, evmAddr[:]...)),
```
jewei1997, line 148: For my education, can you explain why you don't need a READ for the receipt and don't need WRITEs for the other resource types (sdkacltypes.ResourceType_KV_EVM_S2E, sdkacltypes.ResourceType_KV_EVM_E2S,  sdkacltypes.ResourceType_KV_EVM_S2E, sdkacltypes.ResourceType_KV_EVM_CODE_HASH)

aclmapping/evm/mappings.go
```diff
@@ -35,66 +39,136 @@ func TransactionDependencyGenerator(_ aclkeeper.Keeper, evmKeeper evmkeeper.Keep
...
+		sdkacltypes.AccessOperation{
+			AccessType:         sdkacltypes.AccessType_WRITE,
+			ResourceType:       sdkacltypes.ResourceType_KV_BANK_BALANCES,
+			IdentifierTemplate: idTempl,
+		},
+		sdkacltypes.AccessOperation{
+			AccessType:         sdkacltypes.AccessType_READ,
+			ResourceType:       sdkacltypes.ResourceType_KV_AUTH_ADDRESS_STORE,
+			IdentifierTemplate: hex.EncodeToString(authtypes.AddressStoreKey(addr)),
+		})
```
jewei1997, line 173: Can you also explain why there's no WRITE for the auth address store.

aclmapping/evm/mappings.go
codchen: - receipt is never read during transaction processing. It's only read by the evm rpc endpoints
- WRITES for S2E and E2S are in the ante handler dependencies
- WRITE for code hash doesn't happen with all transactions, so we don't want to pessimistically block on it for every transaction

aclmapping/evm/mappings.go
codchen: auth address store is written in the ante handler

PR #1162, opened 2023-12-04
[EVM] add ERC-721 -> CW-721 interoperability contract

codchen:
- Added a solidity contract that delegates operations to an underlying CW721 contract
- Modified bank and wasmd contract to prevent impersonating. Specifically, operations like wasmd execute and bank send can only be done for msg.sender of the calling contract (note, not msg.sender of precompile, which would be the calling contract; we want the caller of the calling contract)
- added two helper precompiles to help extract values from json and convert addresses

unit test
local sei test


contracts/src/NativeSeiTokensERC20.sol
```diff
@@ -38,7 +38,7 @@ contract NativeSeiTokensERC20 is ERC20 {
     }
 
     function _update(address from, address to, uint256 value) internal override {
-        bool success = BankPrecompile.send(from, to, denom, value);
+        (bool success,) = BANK_PRECOMPILE_ADDRESS.delegatecall(abi.encodeWithSignature("sendFromCaller(address,string,uint256)", to, denom, value));
```
jewei1997: this would make transferFrom also transfer from the msg.sender, which isn't correct

codchen: updated so that we still support `from->to` bank send but only for whitelisted codehash

precompiles/json/json.go
```diff
@@ -0,0 +1,129 @@
...
+	switch method.Name {
+	case ExtractAsBytesMethod:
+		return p.extractAsBytes(ctx, method, args)
+	case ExtractAsBytesListMethod:
+		return p.extractAsBytesList(ctx, method, args)
+	}
+	return
+}
+
+func (p Precompile) extractAsBytes(_ sdk.Context, method *abi.Method, args []interface{}) ([]byte, error) {
```
jewei1997, line 96: do we have tests for this and the `extractAsBytesList`?

contracts/test/CW721ERC721WraperTest.t.sol
```diff
@@ -0,0 +1,228 @@
...
+        );
+        vm.mockCall(
+            JSON_PRECOMPILE_ADDRESS,
+            abi.encodeWithSignature("extractAsBytes(bytes,string)", bytes("{\"spender\":\"sei1vldxw5dy5k68hqr4d744rpg9w8cqs54x4asdqe\"}}"), "spender"),
+            abi.encode(bytes("sei1vldxw5dy5k68hqr4d744rpg9w8cqs54x4asdqe"))
+        );
+        assertEq(wrapper.isApprovedForAll(0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266, 0xF39fD6e51Aad88F6f4CE6AB8827279CFffb92267), true);
+    }
+
+    function testTransferFrom() public {
```
jewei1997, line 175: For transferFrom would be good to test the failure case where an approval was not given.

jewei1997: Can we also test Transfer? Also, would be good to test that impersonating doesn't work (if possible).

contracts/src/CW721ERC721Wrapper.sol
```diff
@@ -0,0 +1,131 @@
...
+        emit ApprovalForAll(msg.sender, operator, approved);
+    }
+
+    function _execute(bytes memory req) internal returns (bytes memory) {
+        (bool success, bytes memory ret) = WASMD_PRECOMPILE_ADDRESS.delegatecall(
+            abi.encodeWithSignature(
+                "execute(string,bytes,bytes)",
+                Cw721Address,
+                bytes(req),
+                bytes("[]")
```
jewei1997, line 109: Can you explain why this param is just empty arr? When would it not be empty arr?

contracts/src/CW721ERC721Wrapper.sol
```diff
@@ -0,0 +1,131 @@
...
+        for (uint i=0; i<approvals.length; i++) {
+            bytes memory op = JsonPrecompile.extractAsBytes(approvals[i], "spender");
+            if (AddrPrecompile.getEvmAddr(string(op)) == operator) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    // Calls
```
jewei1997: nit: maybe "Transactions" is a better word for writes

contracts/test/CW721ERC721WraperTest.t.sol
```diff
@@ -0,0 +1,228 @@
...
+    function setUp() public {
+        wrapper = new CW721ERC721Wrapper(MockCWContractAddress, "name", "symbol");
+        mockWasmd = new MockWasmd();
+        mockJson = new MockJson();
+        mockAddr = new MockAddr();
+        vm.etch(WASMD_PRECOMPILE_ADDRESS, address(mockWasmd).code);
+        vm.etch(JSON_PRECOMPILE_ADDRESS, address(mockJson).code);
+        vm.etch(ADDR_PRECOMPILE_ADDRESS, address(mockAddr).code);
+    }
+
```
jewei1997, line 93: might be good to test all functions including things like name and symbol and decimals just to make sure they're set properly.

precompiles/bank/Bank.sol
```diff
@@ -0,0 +1,52 @@
...
+        address toAddress,
+        string memory denom,
+        uint256 amount
+    ) external returns (bool success);
+
+    function sendFromOrigin(
+        address toAddress,
+        string memory denom,
+        uint256 amount
+    ) external returns (bool success);
```
jewei1997, line 29: So send allows you to send from any address, sendFromCaller only allows you to send from msg.sender and sendFromOrigin allows you to send from tx.origin?

jewei1997 (approved):
overall looks good, just some questions and requests for more solidity tests.

precompiles/bank/Bank.sol
codchen: yeah, and `send` is only callable by whitelisted code hashes

contracts/test/CW721ERC721WraperTest.t.sol
codchen: the approval logic is actually delegated to the underlying CW721 contracts, so there is no approval logic on the solidity level (721 is different than ERC20 because Sei's native bank module doesn't have approval logic, whereas CW721 has). The test would ensure that the caller is passed to the CW contract as the caller so that it can't pretend to be someone else on CW-level

contracts/src/CW721ERC721Wrapper.sol
codchen: this param is the amount of funds that would be sent to the CW contracts. In the case of ERC721 it should always send no fund

precompiles/json/json.go
codchen: good catch, added

contracts/test/CW721ERC721WraperTest.t.sol
codchen: good point, added

PR #1167, opened 2023-12-07
[EVM] Sign and Accounts endpoints

codchen:
Add Accounts and Sign endpoints that use keys in the local test wallet. This would be secure on a real node as the real node would have no key in its `test` wallet backend.

local test


jewei1997:
Definitely a lot more tests passing for v3-core tests!

PR #1177, opened 2023-12-16
add occ wasm testing functionality to loadtest

udpatil:
This adds the sei tester contract to the loadtest suite and adds some message types corresponding to the contract. It will allow for testing wasm contract functionality specifically with parallel writes and iterator ranges.


loadtest/loadtest_client.go
```diff
@@ -87,12 +86,14 @@ func NewLoadTestClient(config Config) *LoadTestClient {
 }
 
 func (c *LoadTestClient) SetValidators() {
-	if strings.Contains(c.LoadTestConfig.MessageType, "staking") {
-		resp, err := c.StakingQueryClient.Validators(context.Background(), &stakingtypes.QueryValidatorsRequest{})
-		if err != nil {
-			panic(err)
+	for _, mType := range c.LoadTestConfig.MessageTypes {
+		if mType == Staking {
```
philipsu522, line 97: we probably need to re-think staking- thoughts on commenting this out or jsut adding a print saying "Staking isn't supported yet"?

loadtest/config.json
```diff
@@ -60,6 +72,7 @@
         "payload": "{}"
       }
     },
+    "sei_tester_address": "",
```
yzang2019, line 75: what is this for?

udpatil: thats the address of the deployed contract `sei-tester` that's used for the parallelism transactions + iterator workloads

PR #1181, opened 2023-12-20
Fix getTxByHash for pending txs

codchen:
Previously getTransactionCount and getTransactionByHash were not accounting for pending transactions correctly

integrated with local sei


evmrpc/tx.go
```diff
@@ -85,6 +87,39 @@ func (t *TransactionAPI) GetTransactionByBlockHashAndIndex(ctx context.Context,
...
+					big.NewInt(sdkCtx.BlockHeight()),
+					uint64(sdkCtx.BlockTime().Second()),
+				)
+				from, _ := ethtypes.Sender(signer, etx)
+				if from == addr {
+					nonces = append(nonces, int(etx.Nonce()))
+				}
+			}
+		}
+		if page*30 >= res.Total {
```
philipsu522: out of curiosity - what is the reasoning for this limit?

philipsu522 (approved):

evmrpc/tx.go
codchen: this is to check if this is the last page (30 is the default page size; I've changed it to reference the actual constant)

PR #1190, opened 2023-12-28
Initial commit for producer/consumer loadtest client

philipsu522:
Loadtest improvements:
- Move to a producer consumer model
- Use random numbers for seqno (need to merge this pr to https://github.com/sei-protocol/sei-cosmos/pull/385)
- Allow for long running tests
- Remove writing data to a "result" file that then needs to be queried with another tool, instead generate stats on the fly
- Ran on loadtest cluster
image
image


go.mod
```diff
@@ -274,11 +277,10 @@ require (
 replace (
 	github.com/CosmWasm/wasmd => github.com/sei-protocol/sei-wasmd v0.0.4
 	github.com/confio/ics23/go => github.com/cosmos/cosmos-sdk/ics23/go v0.8.0
-	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.2.68
+	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.2.69-0.20231228001217-1e240574c003
```
philipsu522: will revert go.mod changes once i merge https://github.com/sei-protocol/sei-cosmos/pull/385

loadtest/loadtest_client.go
```diff
@@ -108,194 +91,58 @@ func (c *LoadTestClient) Close() {
...
+				msgs, _, _, gas, fee := c.generateMessage(config, key, config.MsgsPerTx)
+				txBuilder := TestConfig.TxConfig.NewTxBuilder()
+				_ = txBuilder.SetMsgs(msgs...)
+				txBuilder.SetGasLimit(gas)
+				txBuilder.SetFeeAmount([]types.Coin{
+					types.NewCoin("usei", types.NewInt(fee)),
+				})
+				// Use random seqno to get around txs that might already be seen in mempool
+
+				c.SignerClient.SignTx(c.ChainID, &txBuilder, key, uint64(rand.Intn(math.MaxInt)))
```
udpatil: is this necessary after disabling seqno check? couldn't we just set them all to a constant seqno?

loadtest/loadtest_client.go
```diff
@@ -108,194 +91,58 @@ func (c *LoadTestClient) Close() {
...
-			codeString = err.Error()
+		select {
+		case <-done:
+			fmt.Printf("Stopping consumer %d\n", consumerId)
+			return
+		case tx, ok := <-txQueue:
+			if !ok {
+				fmt.Printf("Stopping consumer %d\n", consumerId)
+			}
+			SendTx(tx, typestx.BroadcastMode_BROADCAST_MODE_SYNC, false, *c, sentCount)
```
udpatil: we likely want this to be async, right? otherwise there would be a bottleneck for sending txs since theyll need to sequentially pass checkTx to enter mempool?

loadtest/config.json
```diff
@@ -5,7 +5,7 @@
     "tls": false,
     "msgs_per_tx": 10,
```
udpatil: Note, we may want to change this default to like 1, since otherwise it will send a TON of messages, whereas before the config fields were named weirdly such that txs per block = 500 and msgs per tx = 10 would result in 50 txs with 10 messages each IIRC. now, it would ACTUALLY result in 500TPS of 10 messages PER tx

loadtest/loadtest_client.go
```diff
@@ -108,194 +91,58 @@ func (c *LoadTestClient) Close() {
...
+	accountIdentifier := fmt.Sprint(producerId)
+	accountKeyPath := c.SignerClient.GetTestAccountKeyPath(uint64(producerId))
+	key := c.SignerClient.GetKey(accountIdentifier, "test", accountKeyPath)
+	count := 0
+	for {
+		select {
+		case <-done:
+			fmt.Printf("Stopping producer %d\n", producerId)
+			return
+		case <-ticker.C:
```
yzang2019: Instead of using ticker to do rate limiting, I think a better approach is just to use "golang.org/x/time/rate": https://github.com/sei-protocol/dex-indexer/blob/main/cmd/index_producer/main.go#L63

loadtest/loadtest_client.go
yzang2019: We want to do sync here and wait until the response back, otherwise you have to query the tx and figure out when it is completed. We can put each tx into a separate goroutine though

loadtest/config.json
yzang2019: Did we change the concept here? I think previously if target is 500, batch size is 10, then it will only send 50txs  and each tx will contain 10 messages

loadtest/loadtest_client.go
yzang2019: I think you probably have to at least make these transactions different because mempool will reject duplicate txs.  However, even random seq can still have a chance to result in duplicates, we should probably just switch to a monitonically increasing seqno here?

loadtest/config.json
philipsu522: logic has not changed. will set this to 1 as default

yzang2019 (approved):

loadtest/main.go
```diff
@@ -56,49 +70,112 @@ func init() {
...
+	close(done)
+	wg.Wait()
+	close(txQueue)
+}
+
+func printStats(startTime time.Time, producedCount *int64, sentCount *int64, prevSentCount *int64, blockHeights []int, blockTimes []string) {
+	elapsed := time.Since(startTime)
+	produced := atomic.LoadInt64(producedCount)
+	sent := atomic.LoadInt64(sentCount)
+	tps := float64(sent-*prevSentCount) / elapsed.Seconds()
```
philipsu522, line 154: b/c sender uses broadcast mode = block, we can use client-side metrics to calculate tps

loadtest/loadtest_client.go
yzang2019: I wonder whether we need to apply rate limit for signing tx or not at all. We should apply the rate limiting only for the broadcast tx right? For signing txs, we can keep signing and generating tx as long as the queue/channel is not full

yzang2019 (approved):

PR #1192, opened 2023-12-28
[EVM] Add tx replacement

stevenlanders:
- EVM transactions populate sender with a pattern `{address}|{nonce}`
- The sender is used to de-dupe in the mempool (sei-tendermint)
- A replacement transaction must have a higher priority number to replace (else reject)

- sei-tendermint mempool unit tests

### DIFF
--- a/app/abci.go
+++ b/app/abci.go
@@ -2,6 +2,9 @@ package app
 
 import (
 	"context"
+	"fmt"
+	"github.com/sei-protocol/sei-chain/x/evm/ante"
+	"github.com/sei-protocol/sei-chain/x/evm/types"
 	"time"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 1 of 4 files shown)


app/abci.go
```diff
@@ -35,7 +38,38 @@ func (app *App) EndBlock(ctx sdk.Context, req abci.RequestEndBlock) (res abci.Re
...
 	_, span := app.GetBaseApp().TracingInfo.Start("CheckTx")
 	defer span.End()
-	return app.BaseApp.CheckTx(ctx, req)
+	resp, err := app.BaseApp.CheckTx(ctx, req)
+	if err != nil {
+		return resp, err
+	}
+
+	// For EVM, populate sender on the response so that mempool can decide to dedupe
+	tx, err := app.txDecoder(req.Tx)
```
codchen, line 57: decoding (and the addr recovery logic below) is quite heavy so ideally we don't repeat it here. One other option is to a new field in `sdk.Context`, set it in the `sig` antehandler (where we derive address), and set it in checkTx response in sei-cosmos:baseapp:runTx.

But for now I think this is okay, since CheckTx itself doesn't directly affect chain performance

PR #1194, opened 2023-12-28
[EVM] Fix pending nonce race issues

stevenlanders:
- next nonce is calculated from list of pending nonces

## Needs these PRs to be merged first
- https://github.com/sei-protocol/sei-tendermint/pull/179
- https://github.com/sei-protocol/sei-cosmos/pull/390

- unit tests 
- manual test: if i skip nonces, the pending ones wait on previous nonces
- manual test: the `get_ethTransactionCount` returns the next nonce for pending
- evm tests pass without delays now
```
  EVM Test
    EVMCompatibilityTester
      Deployment
        ✔ Should deploy successfully
      Call Another Contract
        ✔ should set balance and then retrieve it via callAnotherContract (185ms)
      Msg Properties
        ✔ Should store and retrieve msg properties correctly (638ms)
      Block Properties
        ✔ Should get and print the block properties
      Variable Types
        ✔ Should set the address correctly and emit an event (608ms)
        ✔ Should set the bool correctly and emit an event (606ms)
        ✔ Should set the uint256 correctly and emit an event (592ms)
        ✔ Should set the string correctly and emit an event (562ms)
        ✔ Should correctly set and retrieve balances in the mapping (542ms)
        ✔ Should store and retrieve a private var correctly (529ms)
      Require Logic
        ✔ Should revert when false is passed to testRevert
        ✔ Should not revert when true is passed to testRevert (538ms)
      Assembly
        ✔ Should add numbers correctly
        ✔ Should return the current balance of the contract
        ✔ Should return correct value from readFromStorage(index) (524ms)
      JSON-RPC
        ✔ Should retrieve a transaction by its hash (530ms)
        ✔ Should retrieve a block by its number
        ✔ Should retrieve the latest block
        ✔ Should get the balance of an account
        ✔ Should get the code at a specific address
        ✔ Should retrieve a block by its hash
        ✔ Should fetch the number of transactions in a block
        ✔ Should retrieve a transaction receipt (527ms)
        ✔ Should fetch the current gas price
        ✔ Should estimate gas for a transaction
        ✔ Should check the network status
        ✔ Should fetch the nonce for an account
        ✔ Should fetch logs for a specific event (528ms)
        ✔ Should subscribe to an event (532ms)
        ✔ Should get the current block number
        ✔ Should fetch a block with full transactions
        ✔ Should get the chain ID
        ✔ Should fetch past logs
        ✔ Should check account's transaction count
        ✔ Should check if an address is a contract


  35 passing (9s)
```


x/evm/keeper/keeper.go
```diff
@@ -158,31 +175,127 @@ func (k *Keeper) getHistoricalHash(ctx sdk.Context, h int64) common.Hash {
...
+func nonceCacheKey(addr common.Address, nonce uint64) string {
+	return fmt.Sprintf("%s|%d", addr.Hex(), nonce)
+}
+
+// CalculateNextNonce calculates the next nonce for an address
+// If includePending is true, it will consider pending nonces
+// If includePending is false, it will only return the next nonce from GetNonce
+func (k *Keeper) CalculateNextNonce(ctx sdk.Context, addr common.Address, includePending bool) uint64 {
+	k.nonceMx.Lock()
+	defer k.nonceMx.Unlock()
```
jewei1997, line 188: oh I see, this will lock over all addresses.

codchen (approved):
lock contention might be an issue but we can worry about that later

PR #1196, opened 2023-12-29
Add catch all panic handler in dex endblock

codchen:
Any panic in EndBlock would cause the chain to halt, so we decide to add back panic recovery handler. Also made executions for multiple pairs sequential for each contract so that panics are easier to manage.

After this fix, there will be:
- `EndBlock` function, the main goroutine, with a panic handler at https://github.com/sei-protocol/sei-chain/compare/fix-dex?expand=1#diff-f32c70484252fadec70cda9a6ad7888fd568f6afffb3750bb9621090e2ac88bbR307
- ParallelRunner, which is the only place where goroutines will be spawned under the main goroutine, with a panic handler inside the spawned goroutine function at https://github.com/sei-protocol/sei-chain/compare/fix-dex?expand=1#diff-e2db42e8d9bee39dc72e4b409b496fd72ca1959186d8b81f000dfc16ccc26a14R137

Screen Shot 2023-12-29 at 12 25 42 PM
Tested locally by reproducing the reported bug POC

### DIFF
--- a/x/dex/contract/execution.go
+++ b/x/dex/contract/execution.go
@@ -3,7 +3,6 @@ package contract
 import (
 	"context"
 	"fmt"
-	"sync"
 	"time"
 
 	otrace "go.opentelemetry.io/otel/trace"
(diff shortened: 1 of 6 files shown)


yzang2019:
Can we add a metric for panicHandler so that we can monitor and alert on it?

x/dex/contract/execution.go
```diff
@@ -130,16 +129,17 @@ func ExecutePairsInParallel(ctx sdk.Context, contractAddr string, dexkeeper *kee
 	cancelResults := []*types.Cancellation{}
 	settlements := []*types.SettlementEntry{}
 
-	mu := sync.Mutex{}
-	wg := sync.WaitGroup{}
+	// mu := sync.Mutex{}
```
yzang2019, line 132: We leave them here so that we can revert them in case we need?

philipsu522 (approved):

PR #1199, opened 2023-12-29
Automatically query for nonce for deploy erc20 cmd

codchen:
Requiring nonce is annoying so this PR changes the deploy ERC20 wrapper command to automatically query for the expected nonce.

local sei

### DIFF
--- a/x/evm/client/cli/tx.go
+++ b/x/evm/client/cli/tx.go
@@ -8,7 +8,6 @@ import (
 	"io"
 	"math/big"
 	"net/http"
-	"strconv"
 	"strings"
 
 	ethabi "github.com/ethereum/go-ethereum/accounts/abi"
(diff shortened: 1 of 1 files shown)


x/evm/client/cli/tx.go
```diff
@@ -333,8 +328,36 @@ func CmdDeployErc20() *cobra.Command {
 			privHex := hex.EncodeToString(priv.Bytes())
 			key, _ := crypto.HexToECDSA(privHex)
 
+			rpc, err := cmd.Flags().GetString(FlagRPC)
+			if err != nil {
+				return err
+			}
+			nonceQuery := fmt.Sprintf("{\"jsonrpc\": \"2.0\",\"method\": \"eth_getTransactionCount\",\"params\":[\"%s\",\"latest\"],\"id\":\"send-cli\"}", crypto.PubkeyToAddress(key.PublicKey).Hex())
```
stevenlanders: we may want to use pending here, if we run it multiple times via scripting

codchen: updated

PR #1204, opened 2024-01-02
Cleanup removed oracle feeds

udpatil:
This ensure that oracle feeds are removed at the end of the oracle slashing window IF the asset is no longer whitelisted. Additionally, it prevents the TWAPs endpoint from serving twaps for denoms that are no longer in the whitelist.

Unit tests

### DIFF
--- a/x/oracle/keeper/keeper.go
+++ b/x/oracle/keeper/keeper.go
@@ -132,9 +132,11 @@ func (k Keeper) RemoveExcessFeeds(ctx sdk.Context) {
 		return false
 	})
 	// compare
-	var activesToClear []string
+	activesToClear := make([]string, 0, len(excessActives))
+	i := 0
 	for denom := range excessActives {
-		activesToClear = append(activesToClear, denom)
+		activesToClear[i] = denom
+		i++
 	}
 	sort.Strings(activesToClear)
 	for _, denom := range activesToClear {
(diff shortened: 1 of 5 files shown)


x/oracle/keeper/keeper.go
```diff
@@ -118,6 +118,31 @@ func (k Keeper) IterateBaseExchangeRates(ctx sdk.Context, handler func(denom str
...
+		return false
+	})
+	// get vote targets
+	k.IterateVoteTargets(ctx, func(denom string, denomInfo types.Denom) (stop bool) {
+		// remove vote targets from actives
+		delete(excessActives, denom)
+		return false
+	})
+	// compare
+	activesToClear := []string{}
```
stevenlanders: fwiw, idiomatic is defaulting to the nil:  `var activesToClear []string`

x/oracle/keeper/keeper_test.go
```diff
@@ -681,3 +693,123 @@ func TestCalculateTwaps(t *testing.T) {
 	require.Error(t, err)
 	require.Equal(t, types.ErrInvalidTwapLookback, err)
 }
+
+func TestCalculateTwapsWithUnsupportedDenom(t *testing.T) {
```
stevenlanders, line 697: :heart: yay test

stevenlanders (approved):

yzang2019 (approved):

PR #1208, opened 2024-01-04
Revert back to legacy event data for subscriptions

jewei1997:

Reverted the original change I made, turns out the legacy event data is also a `EventDataNewBlockHeader` as the type assertion still works.

existing unit tests.


jewei1997:
actually this is a no-op, both versions of event data should work with the subscriptions code.

PR #1224, opened 2024-01-10
Revamp EVM logs

codchen:
Previously, each EVM log will be emit as a Tendermint log, which is heavy on the Tendermint indexer and also hard to read from. This PR changes it so that EVM logs are stored in application state. Specifically:
- all logs from a transaction is stored inside the transaction receipt
- transaction receipt now also stores log bloom
- if a block has any EVM log, we will also store a block-level bloom, which is the union of all log blooms in that block
During lookup, we will leverage the blooms for faster lookups.

local sei
unit tests


evmrpc/bloom_test.go
```diff
@@ -0,0 +1,28 @@
...
+	require.Equal(t,
+		"00000000001000000000000000000020000000000000000000000000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000400000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+		hex.EncodeToString(bloom[:]),
+	)
+	filters := evmrpc.EncodeFilters(
+		[]common.Address{common.HexToAddress("0x797C2dBE5736D0096914Cd1f9A7330403c71d301")},
+		[][]common.Hash{{common.HexToHash("0x036285defb58e7bdfda894dd4f86e1c7c826522ae0755f0017a2155b4c58022e")}},
+	)
+	require.True(t, evmrpc.MatchFilters(bloom, filters))
+}
```
jewei1997, line 52: Would be good to test that the bloom match matches with a filter of just the address or just the topics. Also, would be good to test that the bloom does not match with a slightly different address and slightly different filter

x/evm/state/log.go
```diff
@@ -1,53 +1,27 @@
...
-		return
-	}
-	logs.Ls = append(ls, l)
-	logsbz, err := json.Marshal(&logs)
-	if err != nil {
-		s.err = err
-		return
-	}
-	store.Set(LogsKey, logsbz)
+	s.logs = append(s.logs, l)
```
jewei1997, line 13: Do we need any concurrency control on these functions?

jewei1997 (approved):

jewei1997 (approved):
Overall LGTM

stevenlanders (approved):

evmrpc/bloom_test.go
codchen: yeah good point. Added

x/evm/state/log.go
codchen: should be fine since there will be one statedb instance per tx, and there is no parallelization within the tx

PR #1228, opened 2024-01-11
[EVM] TxPool Content RPC Endpoint

jewei1997:

EVM RPC Endpoint for txpool_content

unit tests + tested endpoint on local chain with real txs going through


x/evm/types/ethtx/dynamic_fee_tx.go
```diff
@@ -14,7 +14,9 @@ func NewDynamicFeeTx(tx *ethtypes.Transaction) (*DynamicFeeTx, error) {
 	if err := ValidateEthTx(tx); err != nil {
 		return nil, err
 	}
+	// chainId := sdk.NewIntFromBigInt(tx.ChainId())
```
jewei1997: do we want this change @codchen ?

evmrpc/config.go
```diff
@@ -84,6 +87,7 @@ var DefaultConfig = Config{
 	WSOrigins:            "*",
 	FilterTimeout:        120 * time.Second,
 	CheckTxTimeout:       5 * time.Second,
+	MaxTxPoolTxs:         1000,
```
jewei1997, line 90: I will limit pending tx size to 1000, is that ok @codchen ? Or should I pull NumUnconfirmedTxs and just get all of them? Afraid it'll be too many.

evmrpc/txpool.go
```diff
@@ -0,0 +1,72 @@
...
+	resUnconfirmedTxs, err := t.tmClient.UnconfirmedTxs(ctx, nil, &total)
+	if err != nil {
+		return nil
+	}
+
+	sdkCtx := t.ctxProvider(LatestCtxHeight)
+	signer := ethtypes.MakeSigner(
+		t.keeper.GetChainConfig(sdkCtx).EthereumConfig(t.keeper.ChainID(sdkCtx)),
+		big.NewInt(sdkCtx.BlockHeight()),
+		uint64(sdkCtx.BlockTime().Second()),
```
codchen: `.Second()` should be `.Unix()`. Can you also fix it here https://github.com/sei-protocol/sei-chain/blob/evm/evmrpc/tx.go#L117 and here https://github.com/sei-protocol/sei-chain/blob/evm/evmrpc/tx.go#L147

codchen (approved):

PR #1234, opened 2024-01-12
add whitelist for CW code hash that can make delegatecall to EVM

codchen:
Per Zellic's advice, we are adding a safeguard in front of the new EVM wasm bindings. Specifically, only whitelisted CW code hashes can make delegate calls (e.g. if user A calls CW contract X which calls EVM contract Y, with a delegatecall, contract Y will see A as the caller, whereas with a regular call, contract Y will see X as the caller). We will only be adding the CW20/721 wrapper hashes (once they are fully implemented) into the whitelist, but validators may decide to whitelist more in the future via gov proposals.

local testing with cw20 wrapper


x/evm/keeper/evm.go
```diff
@@ -29,6 +30,23 @@ func (k *Keeper) HandleInternalEVMCall(ctx sdk.Context, req *types.MsgInternalEV
 	return &sdk.Result{Data: ret}, nil
 }
 
+func (k *Keeper) HandleInternalEVMDelegateCall(ctx sdk.Context, req *types.MsgInternalEVMDelegateCall) (*sdk.Result, error) {
+	if !k.IsCWCodeHashWhitelistedForEVMDelegateCall(ctx, req.CodeHash) {
+		return nil, errors.New("code hash not authorized to make EVM delegate call")
+	}
+	var to *common.Address
+	if req.To != "" {
```
philipsu522, line 38: are there cases where req.To == "" ?

x/evm/keeper/params.go
```diff
@@ -44,3 +44,7 @@ func (k *Keeper) ChainID(ctx sdk.Context) *big.Int {
 func (k *Keeper) WhitelistedCodehashesBankSend(ctx sdk.Context) []string {
 	return k.GetParams(ctx).WhitelistedCodehashesBankSend
 }
+
+func (k *Keeper) WhitelistedCwCodeHashesForDelegateCall(ctx sdk.Context) [][]byte {
+	return k.GetParams(ctx).WhitelistedCwCodeHashesForDelegateCall
```
philipsu522, line 49: do we need to add a migration for this? somethingl ike https://github.com/sei-protocol/sei-chain/blob/main/x/dex/migrations/v15_to_v16.go

philipsu522 (approved):
left some comments - is the gov part going to be a separate pr? could be a good task for someone else to work on

x/evm/keeper/evm.go
codchen: in such case `to` should be left as `nil`

x/evm/keeper/params.go
codchen: yeah good point.Added

codchen:
> left some comments - is the gov part going to be a separate pr? could be a good task for someone else to work on

since the whitelist is part of params, it can already be updated (as a param) via gov

PR #1235, opened 2024-01-15
Add Rocksdb installation to seidb migration guide

Kbhat1:
- Add instructions to install rocksdb + build flags
- NOTE: We can add more commands in `Makefile` to simplify rocksdb installation + setup. But we have ran into quite a lot of issues here, and since its not default, may not be worth it.

- Verified commands
- NOTE: Rocksdb has a lot of issues during installation. Directly reference users to the fbook docs since they will likely run into many issues during installation.


docs/migration/seidb_migration.md
```diff
@@ -135,7 +135,33 @@ pruning-interval = "1000"
...
+CGO_CFLAGS="-I/path/to/rocksdb/include" CGO_LDFLAGS="-L/path/to/rocksdb"
+```
+
+and a `rocksdbBackend` tag:
+
+```bash
+-tags "rocksdbBackend"
+```
+
+to the seid go installation command.
```
Kbhat1, line 161: @philipsu522 @yzang2019 as mentioned, we can add more utils in the `Makefile` to install and setup rocksdb, but we have ran into a lot of issues here and not sure it is worth given it is not the default. Let me know your thoughts.

yzang2019 (approved):

philipsu522 (approved):

PR #1250, opened 2024-01-18
[EVM] Add throughput and latency metrics for all JSON-RPC endpoint

yzang2019:
This PR aims to add metrics to monitor EVM RPC request throughput and latency by each endpoint. 
- The current way to add per endpoint metrcs is not ideal since we have to add for each individual method. 
- A more ideal way is to add metrics at the abstract handler level which will then be automatically applied for all endpoints. However, due to the JSON-RPC framework design, there is no easy way to achieve that without manually decoding the request body, which we want to avoid

Run in local docker cluster and see metrics available


codchen (approved):

evmrpc/block.go
```diff
@@ -31,27 +32,33 @@ func NewBlockAPI(tmClient rpcclient.Client, k *keeper.Keeper, ctxProvider func(i
...
+		return nil, err
 	}
-	return a.getEvmTxCount(block.Block.Txs)
+	return a.getEvmTxCount(block.Block.Txs), nil
 }
 
-func (a *BlockAPI) GetBlockByHash(ctx context.Context, blockHash common.Hash, fullTx bool) (map[string]interface{}, error) {
+func (a *BlockAPI) GetBlockByHash(ctx context.Context, blockHash common.Hash, fullTx bool) (result map[string]interface{}, returnErr error) {
+	startTime := time.Now()
+	defer recordMetrics("eth_GetBlockByHash", startTime, returnErr == nil)
```
stevenlanders: small nit, if we could keep the case of the eth spec which is like `eth_getBlockByHash` it may be easier for us

stevenlanders (approved):

evmrpc/block.go
```diff
@@ -31,27 +32,33 @@ func NewBlockAPI(tmClient rpcclient.Client, k *keeper.Keeper, ctxProvider func(i
...
+		return nil, err
 	}
-	return a.getEvmTxCount(block.Block.Txs)
+	return a.getEvmTxCount(block.Block.Txs), nil
 }
 
-func (a *BlockAPI) GetBlockTransactionCountByHash(ctx context.Context, blockHash common.Hash) *hexutil.Uint {
+func (a *BlockAPI) GetBlockTransactionCountByHash(ctx context.Context, blockHash common.Hash) (result *hexutil.Uint, returnErr error) {
+	startTime := time.Now()
+	defer recordMetrics("eth_GetBlockTransactionCountByHash", startTime, returnErr == nil)
```
jewei1997: nit: first letter should be lowercase, so should be `eth_getBlockTransactionCountByHash` if you want to match the endpoint name.

jewei1997: ah steven already mentioned this.

jewei1997 (approved):

evmrpc/block.go
yzang2019: make sense

PR #1252, opened 2024-01-18
[EVM] Allow multiple txs from same account in a block

stevenlanders:
- needs tendermint/sei-cosmos pr as well 
- this passes along nonce information to mempool
- pending nonce is used to determine whether something is ready to be accepted

- new hardhat test
- unit tests on tendermint


x/evm/ante/sig.go
```diff
@@ -53,17 +60,21 @@ func (svd *EVMSigVerifyDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulat
...
 			// transaction shall be added to mempool as a pending transaction
 			ctx = ctx.WithPendingTxChecker(func() abci.PendingTxCheckerResponse {
 				latestCtx := svd.latestCtxGetter()
-				latestNonce := svd.evmKeeper.GetNonce(latestCtx, evmAddr)
+				latestNonce := svd.evmKeeper.GetNonce(ctx, evmAddr)
+				nextNonce := svd.evmKeeper.CalculateNextNonce(latestCtx, evmAddr, true)
 				if txNonce < latestNonce {
 					return abci.Rejected
-				} else if txNonce == latestNonce {
+				} else if txNonce < nextNonce {
```
codchen: Can you check my understanding of the conditions here: if execution enters this `else if` it means `nextNonce` must be greater than `lastestNonce` (otherwise it'd fall into the first `if`), which means `txNonce` must be GTE `latestNonce` but also smaller than `nextNonce` (the first pending nonce), which means `txNonce` would just be, well, `lastestNonce`. So wouldn't this be equivalent to the logic before? Trying to understand why we need to change this line

stevenlanders: The `latestNonce` only changes when something is mined (via SetNonce), and is always the last mined nonce + 1.  This means the answer will be `Rejected` for every transaction until a block is processed, then one (and only one) transaction will be eligible the next block.  This is the main reason why we were getting one transaction per block.  Next nonce is including all those pending nonces that haven't been mined.  

The `nextNonce` includes pending.  So it includes nonces that are not yet mined.  This lets us take more transactions from an account in a block.

Example:
- Someone quickly sends [2,3,4,5].  Last used nonce is 1, and latestNonce is now 2.  Only nonce 2 would be eligible before.  Now all of 2-5 are eligible to run.  The mempool ensures these are in order via new clockwork.

I expect the confusion is that the names are `latestNonce` which is actually `lastUsedNoncePlusOne` and `nextNonce`, which is actually the `nextExpectedNonceIncludingPending`.

codchen: I see. It's possible to have multiple values between `latest` and `next`. That makes sense.

codchen (approved):

PR #1253, opened 2024-01-18
[EVM] Add contract factory test

stevenlanders:
- creates a new contract and verifies it gets a new address

- new hardhat test


contracts/test/EVMCompatabilityTester.js
```diff
@@ -52,7 +53,17 @@ describe("EVM Test", function () {
...
 
-    describe("Call Another Contract", function(){
+    describe("Contract Factory", function() {
+      it("should deploy a second contract from createToken", async function () {
+        const txResponse = await evmTester.createToken("TestToken", "TTK");
+        const testerAddress = await evmTester.getAddress();
+        const receipt = await txResponse.wait();
+        const newTokenAddress = receipt.logs[0].address;
+        expect(newTokenAddress).to.not.equal(testerAddress);
+      });
```
jewei1997, line 67: do you want to check that there's code at the new address / call a function on the new token?

jewei1997 (approved):
overall LGTM!

contracts/test/EVMCompatabilityTester.js
stevenlanders: :+1: good idea, added a check for balance (it transfers 100 to the owner upon creation)

PR #1254, opened 2024-01-18
EVM Hardhat gas test

jewei1997:

Simple gas test to test that correct amount is deducted from gas on a simple bank send based on the gas price.

hardhat tests.


contracts/test/EVMCompatabilityTester.js
```diff
@@ -240,8 +264,86 @@ describe("EVM Test", function () {
...
+        console.log(`gasPrice = ${gasPrice}`)
+        console.log(`maxPriorityFeePerGas = ${feeData.maxPriorityFeePerGas}`)
+        console.log(`maxFeePerGas = ${feeData.maxFeePerGas}`)
+
+        const zero = ethers.parseUnits('0', 'ether')
+        const txResponse = await owner.sendTransaction({
+          to: owner.address,
+          value: zero,
+          maxPriorityFeePerGas: feeData.maxPriorityFeePerGas,
+          maxFeePerGas: feeData.maxFeePerGas,
```
jewei1997: this test fails, seems like it doesn't give us the gasPrice of 1 and uses these params as a gasPrice but really they're just maxes and we should get the original gasPrice of 1.

codchen (approved):

PR #1258, opened 2024-01-19
[EVM] Add evm to loadtest client

yzang2019:
This PR tries to add EVM support to loadtest client so that we can configure our loadtest client to send EMV transactions. This PR includes several changes:
- Cherry pick the latest producer/consumer loadtest client from main
- Fix the loadtest client loading account bug
- Bandaid fix for SuggestGasPrice
- Fix nonce issue for EVM client
- Change loadtest to use a per account worker model instead of shared keys
- Fix staking msgtype concurrent map access bug


evmrpc/info.go
```diff
@@ -130,23 +130,36 @@ func (i *InfoAPI) FeeHistory(ctx context.Context, blockCount math.HexOrDecimal64
...
+		if baseFee == nil {
+			// the block has been pruned
+			continue
+		}
 		result.BaseFee = append(result.BaseFee, (*hexutil.Big)(baseFee))
 		height := blockNum
 		block, err := i.tmClient.Block(ctx, &height)
 		if err != nil {
-			return nil, err
+			continue
```
philipsu522: is it possible to add some tests for this? so we don't regress in the future

loadtest/evm.go
```diff
@@ -0,0 +1,125 @@
...
+		privKeyHex := hex.EncodeToString(key.Bytes())
+		privateKey, err := crypto.HexToECDSA(privKeyHex)
+		if err != nil {
+			fmt.Printf("Failed to load private key: %v \n", err)
+		}
+
+		publicKey := privateKey.Public()
+		publicKeyECDSA, ok := publicKey.(*ecdsa.PublicKey)
+		if !ok {
+			panic("Cannot assert type: publicKey is not of type *ecdsa.PublicKey \n")
```
philipsu522: nit - does it make sense to panic here? or just print as warning?

philipsu522 (approved):
lgtm on the loadtest side

loadtest/evm.go
```diff
@@ -0,0 +1,125 @@
...
+
+	rand.Seed(time.Now().Unix())
+	value := big.NewInt(rand.Int63n(math.MaxInt64 - 1))
+	gasPrice, err := client.SuggestGasPrice(context.Background())
+	if err != nil {
+		fmt.Printf("Failed to suggest gas price: %v \n", err)
+		return nil
+	}
+	gasLimit := uint64(200000)
+	tx := ethtypes.NewTransaction(nextNonce, fromAddress, value, gasLimit, gasPrice, nil)
```
jewei1997: it may be possible to just pass in `nil` here for gasPrice instead of querying for SuggestGasPrice every time if you want.

loadtest/evm.go
```diff
@@ -0,0 +1,125 @@
...
+	chainID, err := client.NetworkID(context.Background())
```
jewei1997: Don't need to pull this individually for each tx, it should always be the same, but you can too.

jewei1997 (approved):

jewei1997 (approved):
overall LGTM

loadtest/evm.go
```diff
@@ -0,0 +1,125 @@
...
+		fmt.Printf("Failed to sign evm tx: %v \n", err)
+		return nil
+	}
+	return signedTx
+}
+
+// SendEvmTx takes any signed evm tx and send it out
+func (txSender *EvmTxSender) SendEvmTx(signedTx *ethtypes.Transaction) bool {
+	err := txSender.GetNextClient().SendTransaction(context.Background(), signedTx)
+	if err != nil {
```
jewei1997, line 101: if the tx getting on chain but then reverting is important to you, you may need to check the tx receipt status as well.

loadtest/evm.go
yzang2019: oh interesting, I can try that since it will save some http requests

loadtest/evm.go
yzang2019: Good point

loadtest/evm.go
yzang2019: Make sense, here we can add that in the future if we care about it, for now we'll just assume it succeed as long as send succeed

evmrpc/info.go
yzang2019: good point, will think about it

loadtest/evm.go
yzang2019: Yeah here we can just print a warning message

philipsu522 (approved):

PR #1264, opened 2024-01-22
Add CW721 wrapper for ERC721

codchen:
Add a CosmWasm wrapper following the CW721 spec over an underlying ERC721 contract. Note that since ERC721 doesn't have `burn` in its spec, the `burn` function on the CW wrapper will always return an error.

local sei test
unit tests


philipsu522 (approved):
can we add a unit test for end to end lifecycle of creating minting burning revoking...etc of an nft? could be a start task for someone else too

codchen:
@philipsu522 yeah that'd be quite a large test though (needs to instantiate an EVM contract, send EVM txs to mint tokens, instantiate a CW contract, send CW txs) so I'd include it in a dedicated PR

PR #1265, opened 2024-01-22
EVM remove adjust v, yParity mismatch issue

jewei1997:

Was causing a yParity issue when sending EIP-1559 txs

hardhat tests.


jewei1997:
added to this https://github.com/sei-protocol/sei-chain/pull/1254

PR #1272, opened 2024-01-23
Separate endpoint for associateTx

codchen:
Previously `sendRawTransaction` would send associate TX if the payload can be unmarshaled into an AssociateTx, and `Associate` endpoint would delegate the call to `sendRawTransaction`. This PR removes the associate tx logic in `sendRawTransaction` and make `Associate` directly broadcast to tendermint.

unit test (existing)

### DIFF
--- a/contracts/test/EVMCompatabilityTester.js
+++ b/contracts/test/EVMCompatabilityTester.js
@@ -308,8 +308,7 @@ describe("EVM Test", function () {
         await expect(evmTester.revertIfFalse(false)).to.be.reverted;
       });
 
-      // TODO: fix this test the re-enable, it's breaking on CI
-      it.skip("Should not revert when true is passed to revertIfFalse", async function () {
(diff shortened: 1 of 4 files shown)


contracts/test/EVMCompatabilityTester.js
```diff
@@ -306,11 +306,6 @@ describe("EVM Test", function () {
       it("Should revert when false is passed to revertIfFalse", async function () {
         await expect(evmTester.revertIfFalse(false)).to.be.reverted;
       });
-
-      // TODO: fix this test the re-enable, it's breaking on CI
-      it.skip("Should not revert when true is passed to revertIfFalse", async function () {
```
jewei1997, line 312: I think we just want to remove the `skip` part not the entire test.

codchen:
> would it be good to add a test to our ci that was getting confused as an associate tx to make sure its no longer an issue?

we actually don't have a tx body that we know would cause the confusion - it happens pretty randomly in the hardhat tests

PR #1273, opened 2024-01-23
[EVM] Add index to logs based on order

stevenlanders:
- adds index to logs, which is included in receipts and logs

- added hardhat test


contracts/test/EVMCompatabilityTester.js
```diff
@@ -428,7 +428,7 @@ describe("EVM Test", function () {
         });
 
         describe("Differing maxPriorityFeePerGas and maxFeePerGas", async function() {
-          testCases.forEach(async ([name, maxPriorityFeePerGas, maxFeePerGas]) => {
+          for (const [name, maxPriorityFeePerGas, maxFeePerGas] of testCases) {
```
jewei1997, line 431: what's the difference?

jewei1997 (approved):

contracts/test/EVMCompatabilityTester.js
stevenlanders: Not much, but my ide linter complains that the async function promise wasn't being used, and gave me yellow lines, so I accepted its change suggestion.

PR #1291, opened 2024-01-26
Use memstore to hold deferred TX info

codchen:
Use memstore so that deferred TX info is covered in OCC's state management and won't be double-counted in the case of rerun.

unit tests


x/evm/keeper/keeper.go
```diff
@@ -24,17 +26,16 @@ import (
 )
 
 type Keeper struct {
-	storeKey   sdk.StoreKey
-	Paramstore paramtypes.Subspace
+	storeKey    sdk.StoreKey
+	memStoreKey *sdk.MemoryStoreKey
```
udpatil: Why is this store key a pointer?

udpatil (approved):
Lgtm

PR #1314, opened 2024-02-01
Use per-tx escrow for Wei balance

codchen:
In order to prevent the Wei escrow from being a bottleneck in OCC, this PR changes the escrow account to one-per-tx, and settle those per-tx escrows to the global escrow in EndBlock. One complication with such setup is that the per-tx escrow may temporarily go negative (because the `usei` that needs to be redeemed may be deposited in another tx), so a corresponding change in sei-cosmos that skips negative balance check for escrow account has also been made https://github.com/sei-protocol/sei-cosmos/pull/417

unit tests with different escrow balances


x/evm/module.go
```diff
@@ -167,6 +167,7 @@ func (am AppModule) BeginBlock(sdk.Context, abci.RequestBeginBlock) {
 // returns no validator updates.
 func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.ValidatorUpdate {
 	evmTxDeferredInfoList := am.keeper.GetEVMTxDeferredInfo(ctx)
+	am.keeper.SettleWeiEscrowAccounts(ctx, evmTxDeferredInfoList)
```
udpatil, line 174: in the future, if we allow evm txs into the prioritized execution section, is it necessary to settle the balances after EACH execution group, or is it fine to simply settle in endblock?

x/evm/state/utils.go
```diff
@@ -26,6 +27,12 @@ func GetCoinbaseAddress(ctx sdk.Context) sdk.AccAddress {
 	return sdk.AccAddress(append(CoinbaseAddressPrefix, txIndexBz...))
 }
 
+func GetTempWeiEscrowAddress(ctx sdk.Context) sdk.AccAddress {
+	txIndexBz := make([]byte, 8)
+	binary.BigEndian.PutUint64(txIndexBz, uint64(ctx.TxIndex()))
+	return sdk.AccAddress(append(WeiTmpEscrowPrefix, txIndexBz...))
```
udpatil, line 33: do we have a guarantee that this cannot conflict with other user addresses? I'm assuming it wont because there are fewer bytes used in generating the Acc Address?

x/evm/state/utils.go
```diff
@@ -26,6 +27,12 @@ func GetCoinbaseAddress(ctx sdk.Context) sdk.AccAddress {
 	return sdk.AccAddress(append(CoinbaseAddressPrefix, txIndexBz...))
 }
 
+func GetTempWeiEscrowAddress(ctx sdk.Context) sdk.AccAddress {
```
udpatil: can't we just use `txIndex` as an argument instead of Context?

x/evm/keeper/wei.go
```diff
@@ -0,0 +1,48 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
+	"github.com/sei-protocol/sei-chain/x/evm/state"
+)
+
+func (k *Keeper) SettleWeiEscrowAccounts(ctx sdk.Context, evmTxDeferredInfoList []EvmTxDeferredInfo) {
+	denom := k.GetBaseDenom(ctx)
+	// settle surplus escrow first
+	for _, info := range evmTxDeferredInfoList {
+		escrow := state.GetTempWeiEscrowAddress(sdk.Context{}.WithTxIndex(info.TxIndx))
```
udpatil: see comment on `utils` file, might be easier to simply pass txIndex instead of creating an empty context WITH a txindex?

x/evm/module.go
codchen: it's fine to settle in the endblock. Technically I don't think we ever need to settle, but having negative balances may break some UI (if someone deliberately query the escrow accounts) so we settle it, but as long as end-of-block state is settled we should be ok

x/evm/state/utils.go
codchen: yeah the bytes are different

x/evm/state/utils.go
codchen: yea good point

udpatil (approved):

PR #1318, opened 2024-02-02
Add staking precompile

codchen:
Add a precompile so that `delegate/redelegate/undelegate` can be made as EVM contract calls.

unit test && local sei with CLI


precompiles/staking/staking.go
```diff
@@ -0,0 +1,173 @@
...
+func (p Precompile) RequiredGas(input []byte) uint64 {
+	methodID := input[:4]
+
+	if bytes.Equal(methodID, p.DelegateID) {
+		return 50000
+	} else if bytes.Equal(methodID, p.RedelegateID) {
+		return 70000
+	} else if bytes.Equal(methodID, p.UndelegateID) {
+		return 50000
+	}
```
philipsu522, line 92: are these ever going to change?

precompiles/staking/staking_test.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/sei-protocol/sei-chain/x/evm/keeper"
+	"github.com/sei-protocol/sei-chain/x/evm/types"
+	evmtypes "github.com/sei-protocol/sei-chain/x/evm/types"
+	"github.com/sei-protocol/sei-chain/x/evm/types/ethtx"
+	minttypes "github.com/sei-protocol/sei-chain/x/mint/types"
+	"github.com/stretchr/testify/require"
+	tmtypes "github.com/tendermint/tendermint/proto/tendermint/types"
+)
+
+func TestStaking(t *testing.T) {
```
philipsu522, line 31: would suggest adding 1-2 tests with some more edge cases:
1. undelegating with no delegation
2. redelegating with no delegation

philipsu522 (approved):

precompiles/staking/staking.go
codchen: it may change if we change staking implementation (mainly want this to match cosmos staking gas cost), so I think keeping it as a constant should be fine?

precompiles/staking/staking_test.go
codchen: added

PR #1323, opened 2024-02-05
Add Gov/Distribution precompiles

codchen:
Gov precompile supports:
- vote
- deposit

Distribution precompile supports:
- set withdraw address (subject to distribution module param control)
- withdraw delegation rewards

unit tests/local sei test


precompiles/common/expected_keepers.go
```diff
@@ -43,3 +44,13 @@ type StakingKeeper interface {
...
 }
+
+type GovKeeper interface {
+	AddVote(ctx sdk.Context, proposalID uint64, voterAddr sdk.AccAddress, options govtypes.WeightedVoteOptions) error
+	AddDeposit(ctx sdk.Context, proposalID uint64, depositorAddr sdk.AccAddress, depositAmount sdk.Coins) (bool, error)
+}
+
+type DistributionKeeper interface {
+	SetWithdrawAddr(ctx sdk.Context, delegatorAddr sdk.AccAddress, withdrawAddr sdk.AccAddress) error
+	WithdrawDelegationRewards(ctx sdk.Context, delAddr sdk.AccAddress, valAddr sdk.ValAddress) (sdk.Coins, error)
```
philipsu522, line 55: do we want to support 
- `withdraw-all-rewards` 
- `set-withdraw-addr`
as well?

precompiles/gov/gov.go
```diff
@@ -0,0 +1,130 @@
...
+	pcommon "github.com/sei-protocol/sei-chain/precompiles/common"
+)
+
+const (
+	VoteMethod    = "vote"
+	DepositMethod = "deposit"
+)
+
+const (
+	GovAddress = "0x0000000000000000000000000000000000001006"
```
philipsu522, line 22: nit - is there a way to ensure that we don't reuse an addr for precompile? maybe we should add some precompile registry (or something like in app.go) and have a unit test

precompiles/common/expected_keepers.go
```diff
@@ -43,3 +44,13 @@ type StakingKeeper interface {
 	BeginRedelegate(goCtx context.Context, msg *stakingtypes.MsgBeginRedelegate) (*stakingtypes.MsgBeginRedelegateResponse, error)
 	Undelegate(goCtx context.Context, msg *stakingtypes.MsgUndelegate) (*stakingtypes.MsgUndelegateResponse, error)
 }
+
+type GovKeeper interface {
+	AddVote(ctx sdk.Context, proposalID uint64, voterAddr sdk.AccAddress, options govtypes.WeightedVoteOptions) error
```
philipsu522, line 49: do we want to add `submit-proposal`?

philipsu522 (approved):

precompiles/common/expected_keepers.go
codchen: i think those are the two currently implemented, or are you referring to smth else?

precompiles/gov/gov.go
codchen: yeah i can add a check in the setup code

precompiles/common/expected_keepers.go
codchen: do we currently allow users to submit proposal on UI on cosmos side? (feel like this precompile is only relevant for users interacting with the chain through evm UI)

PR #1330, opened 2024-02-07
[EVM] Update loadtest client receipt polling

stevenlanders:
- Updates gasLimit to 21000 (value for a send tx), so more can fit into a block
- Limit bank send amount to avoid using wei balance
- Get rid of GetTxReceipt by switching to broadcast mode = block


loadtest/evm.go
```diff
@@ -152,16 +140,17 @@ func (txClient *EvmTxClient) ResetNonce() error {
 	return nil
 }
 
+// nolint
 func withRetry(callFunc func() error) (bool, error) {
```
codchen: is this used anywhere else than getReceipt?

codchen (approved):

loadtest/evm.go
yzang2019: No we can remove it

PR #1332, opened 2024-02-08
EVM Precompile Send to Native address

Kbhat1:
- Add `sendNative` to send to native address

NOTE: Need to update sei-cosmos `SendCoinsAndWei` to remove `denom` arg and just rely on BaseDenom. Will update this after

- Add unit tests
- Tested on local chain


jewei1997 (approved):
LGTM pending successful testing on local chain

precompiles/bank/bank.go
```diff
@@ -169,6 +178,38 @@ func (p Precompile) send(ctx sdk.Context, method *abi.Method, args []interface{}
...
+	receiverAddr, ok := (args[1]).(string)
+	if !ok || receiverAddr == "" {
+		return nil, errors.New("invalid addr")
+	}
+
+	receiverSeiAddr, err := sdk.AccAddressFromBech32(receiverAddr)
+	if err != nil {
+		return nil, err
+	}
+	if err := p.bankKeeper.SendCoins(ctx, senderSeiAddr, receiverSeiAddr, sdk.NewCoins(sdk.NewCoin(denom, sdk.NewIntFromBigInt(amount)))); err != nil {
```
stevenlanders: The units for this amount must be in 10^6 units?  (usei?)  I wonder whether there's a safety issue here if someone mistakes that within EVM and sends a wei value?

Kbhat1: That’s a good point. Messaging you to discuss but I believe this also affects the other send precompile

precompiles/bank/bank.go
```diff
@@ -116,6 +120,11 @@ func (p Precompile) Run(evm *vm.EVM, caller common.Address, input []byte) (bz []
 			return nil, err
 		}
 		return p.send(ctx, method, args)
+	case SendNativeMethod:
+		if err := p.validateCaller(ctx, caller); err != nil {
```
codchen: `validateCaller` would validate that the `caller` is a whitelisted contract (i.e. our ERC20 wrappers), whereas the use case for sending to native address is for users to call directly, so let's not call it here, and change the function signature to:
```
function sendNative(
        string memory toNativeAddress,
        string memory denom,
        uint256 amount
    ) external returns (bool success);
```
The `from` address is just `caller`

codchen (approved):

precompiles/bank/bank_test.go
```diff
@@ -62,69 +77,69 @@ func TestRun(t *testing.T) {
...
-	require.Nil(t, err)
-	args, err := name.Inputs.Pack("usei")
-	require.Nil(t, err)
-	res, err := p.Run(&evm, common.Address{}, append(p.NameID, args...))
-	require.Nil(t, err)
-	outputs, err := name.Outputs.Unpack(res)
-	require.Nil(t, err)
-	require.Equal(t, "SEI", outputs[0])
+// func TestMetadata(t *testing.T) {
+// 	k, ctx := testkeeper.MockEVMKeeper()
```
jewei1997: remember to uncomment this

codchen (changes requested):
let's change the function to be a `payable` one similar to https://github.com/sei-protocol/sei-chain/pull/1376

Kbhat1:
> let's change the function to be a `payable` one similar to #1376

@codchen PR updated to add `payable` to sendNative

precompiles/common/precompiles.go
```diff
@@ -74,3 +74,12 @@ func HandlePayment(ctx sdk.Context, precompileAddr sdk.AccAddress, payer sdk.Acc
 	}
 	return coin, nil
 }
+
+func HandlePaymentUseiWei(ctx sdk.Context, precompileAddr sdk.AccAddress, payer sdk.AccAddress, value *big.Int, bankKeeper BankKeeper) (sdk.Int, sdk.Int, error) {
+	usei, wei := state.SplitUseiWeiAmount(value)
+	// refund payer because the following precompile logic will debit the payments from payer's account
+	if err := bankKeeper.SendCoinsAndWei(ctx, precompileAddr, payer, usei, wei); err != nil {
+		return sdk.Int{}, sdk.Int{}, err
```
Kbhat1, line 82: NOTE: We can also just not use this `HandlePaymentUseiWei` for `sendNative` and directly bank send from the precompileAddr to the receiver addr. Just left it in case we specifically for it to be debited from user's sei address

precompiles/bank/bank.go
```diff
@@ -159,9 +171,44 @@ func (p Precompile) send(ctx sdk.Context, method *abi.Method, args []interface{}
...
+	if !ok || receiverAddr == "" {
+		return nil, errors.New("invalid addr")
+	}
+
+	receiverSeiAddr, err := sdk.AccAddressFromBech32(receiverAddr)
+	if err != nil {
+		return nil, err
+	}
+
+	usei, wei, err := pcommon.HandlePaymentUseiWei(ctx, p.evmKeeper.GetSeiAddressOrDefault(ctx, p.address), senderSeiAddr, value, p.bankKeeper)
```
Kbhat1, line 203: NOTE: We can also just not use this `HandlePaymentUseiWei` for `sendNative` and directly bank send from the precompileAddr to the receiver addr. Just left it in case we specifically for it to be debited from user's sei address

precompiles/common/precompiles.go
codchen: directly sending should be fine, but we can leave it as is in this PR for now

codchen (approved):

PR #1342, opened 2024-02-10
[EVM] Add erc20 to load test

stevenlanders:
- Add ERC20 Type
- Uses forge to compile abi and abigen to generate bindings
- Deploys a NoopToken that doesn't transfer anything (saves last value state, just to mutate something)
- (No funding needed)

- Loadtesting in lower env (this is 5 tps for example)

image


loadtest/evm.go
```diff
@@ -65,35 +72,83 @@ func NewEvmTxClient(
...
+	if err != nil {
+		panic(fmt.Sprintf("Failed to create transactor: %v \n", err))
+	}
+	auth.Nonce = big.NewInt(int64(txClient.nextNonce()))
+	auth.Value = big.NewInt(0)
+	auth.GasLimit = uint64(21000)
+	auth.GasPrice = txClient.gasPrice
+	auth.Context = context.Background()
+	auth.From = txClient.accountAddress
+	auth.NoSend = true
```
jewei1997, line 133: hm why is this true? Are we trying to not send it for some reason?

stevenlanders: this prevents the abigen bindings from auto-sending it, so that the load test client can send the signed one later

loadtest/contracts/deploy_erc20.sh
```diff
@@ -0,0 +1,12 @@
...
+# This script is used to deploy the NoopToken contract to the local network
+# This avoids trying to predict what address it might be deployed to
+
+cd loadtest/contracts/evm || exit 1
+
+./setup.sh
+
+git submodule update --init --recursive
+
+/root/.foundry/bin/forge create -r http://127.0.0.1:8545 --private-key 57acb95d82739866a5c29e40b0aa2590742ae50425b7dd5b5d279a986370189e src/NoopToken.sol:NoopToken --json --constructor-args "NoopToken" "NT" | jq -r '.deployedTo'
```
philipsu522: is this intended to run on the validator node or the load test client?

philipsu522 (approved):

loadtest/contracts/deploy_erc20.sh
stevenlanders: ah good catch - i was testing locally, this needed to be dynamic (just updated)

PR #1343, opened 2024-02-10
Remove unnecessary steps from state sync script

packetstracer:

Removed unnecessary file deletions and backups that could be unwanted such as deleting config.json and wasm files (user could lose config changes).

Perform a state sync of a node and check if it syncs correctly.

### DIFF
--- a/docs/migration/seidb_migration.md
+++ b/docs/migration/seidb_migration.md
@@ -156,19 +156,9 @@ systemctl stop seid
 # Step 2: remove and clean up data
 echo "Removing data files..."
 cp $SEID_HOME/data/priv_validator_state.json /root/priv_validator_state.json
-cp $SEID_HOME/config/priv_validator_key.json /root/priv_validator_key.json
-cp $SEID_HOME/genesis.json /root/genesis.json
 rm -rf $SEID_HOME/data/*
-rm -rf $SEID_HOME/wasm
-rm -rf $SEID_HOME/config/priv_validator_key.json
-rm -rf $SEID_HOME/config/genesis.json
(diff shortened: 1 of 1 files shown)


docs/migration/seidb_migration.md
```diff
@@ -156,19 +156,9 @@ systemctl stop seid
 # Step 2: remove and clean up data
 echo "Removing data files..."
 cp $SEID_HOME/data/priv_validator_state.json /root/priv_validator_state.json
-cp $SEID_HOME/config/priv_validator_key.json /root/priv_validator_key.json
-cp $SEID_HOME/genesis.json /root/genesis.json
 rm -rf $SEID_HOME/data/*
-rm -rf $SEID_HOME/wasm
```
yzang2019, line 162: We need to remove wasm data to clean up the old data

docs/migration/seidb_migration.md
```diff
@@ -156,19 +156,9 @@ systemctl stop seid
...
-rm -rf $SEID_HOME/config/priv_validator_key.json
-rm -rf $SEID_HOME/config/genesis.json
-rm -rf $SEID_HOME/config/config.toml
```
yzang2019, line 165: It's better to remove config, so that genesis will generate latest config

yzang2019 (changes requested):
First thanks for submitting this PR. There are multiple ways of state sync, yours approach tend to do less changes which is fine, but here we actually recommend completely regenesis and clean up everything to do a more complete state sync. The script we provide is just for reference, people don't have to follow the exact same process

PR #1345, opened 2024-02-12
Use in-memory map for deferred info

codchen:
This is a performance optimization resulted from recent load test. Specifically storing and iterating through deferred info in cachekv store is a major bottleneck. This change replaces cachekv usage with an in-memory `sync.Map`, which:
- is cleared at each `BeginBlock`
- is key'ed by tx index, so that rerun of the same tx will overwrite its previous run's entry
- may contain invalid entries (e.g. tx X finished successfully in its first run and inserted an entry to this map, but needed to rerun and finished erroneously during rerun), which are excluded based on `ExecTxResult`

local sei && unit tests

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1443,6 +1443,7 @@ func (app *App) ProcessBlock(ctx sdk.Context, txs [][]byte, req BlockProcessRequ
 	for relativeOtherIndex, originalIndex := range otherIndices {
 		txResults[originalIndex] = otherResults[relativeOtherIndex]
 	}
+	app.EvmKeeper.SetTxResults(txResults)
 
 	// Finalize all Bank Module Transfers here so that events are included
 	lazyWriteEvents := app.BankKeeper.WriteDeferredBalances(ctx)
--- a/x/evm/keeper/keeper.go
+++ b/x/evm/keeper/keeper.go
@@ -1,7 +1,6 @@
 package keeper
 
 import (
-	"encoding/binary"
 	"fmt"
 	"math"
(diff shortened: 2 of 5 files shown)


x/evm/keeper/keeper.go
```diff
@@ -142,69 +146,27 @@ func (k *Keeper) GetHashFn(ctx sdk.Context) vm.GetHashFunc {
...
-		b.SetBytes(bloomIter.Value())
-		bloomMap[int(binary.BigEndian.Uint32(bloomIter.Key()))] = b
-	}
-	bloomIter.Close()
-	for idx, h := range hashMap {
-		i := EvmTxDeferredInfo{TxIndx: idx, TxHash: h}
-		if b, ok := bloomMap[idx]; ok {
-			i.TxBloom = b
-			delete(bloomMap, idx)
+	k.deferredInfo.Range(func(key, value any) bool {
```
stevenlanders, line 149: Maybe our KVStore interface should have `Range`.  This looks so much nicer.

app/app.go
```diff
@@ -367,6 +367,8 @@ type App struct {
 
 	encodingConfig appparams.EncodingConfig
 	evmRPCConfig   evmrpc.Config
+
+	txResultsSetter func([]*abci.ExecTxResult)
```
stevenlanders: is this used?

stevenlanders (approved):

app/app.go
codchen: good catch. Removed

PR #1348, opened 2024-02-13
Hardcode things that dont need to be params

codchen:
Cosmos param store is not efficient in the sense that in order to read a single param, all params in the same module need to be fetched. During load test we discovered that param reading is a big bottleneck. This PR removes the following things from params to improve performance:
- base denom, which will always be `usei`
- chain config, which if changed would require a chain upgrade anyway (i.e. major EVM version upgrade)
- bank whitelist, deprecated (we use https://github.com/sei-protocol/sei-chain/blob/evm/x/evm/keeper/whitelist.go#L32 instead now)

unit tests/local sei upgrade


x/evm/types/config.go
```diff
@@ -44,7 +44,7 @@ func (cc ChainConfig) EthereumConfig(chainID *big.Int) *params.ChainConfig {
 
 func DefaultChainConfig() ChainConfig {
 	return ChainConfig{
-		CancunTime: -1,
+		CancunTime: 0,
```
yzang2019, line 47: why is this changed?

codchen: cancun will be launched on eth mainnet  on 03/13, so we've decided to also support cancun version (given our mainnet launch would be after that). See https://sei-network-io.slack.com/archives/C069MR5S82Z/p1707752152756309?thread_ts=1707751607.560409&cid=C069MR5S82Z for context

yzang2019 (approved):

stevenlanders (approved):

PR #1349, opened 2024-02-13
Add/sub balance directly without middleman

codchen:
Previously we implement EVM statedb's balance operations as:
```
SubBalance: send to a middleman account
AddBalance: send from a middleman account
```
This turns out to be very inefficient in loadtesting, because each operation would need to change two balances. This PR changes the implementation of these two methods to be direct crediting/debiting, and remove the middleman completely.

Any surplus from `Sub/AddBalance` (e.g. if there is a `Sub` of 10 and `Add` of 8 then surplus is 2) will be credited to EVM module account in EndBlock (same behavior as before). Negative surplus is not expected to happen and will cause the transaction to fail.

Another benefit from this change is that we no longer need escrow accounts for wei, which simplifies wei logics a lot. 

Unit tests


philipsu522 (approved):
do we know how much time was spent on miiddleman operations?

yzang2019 (approved):

codchen:
> do we know how much time was spent on miiddleman operations?

in a all-bank-send load test it bumped TPS from 500 to ~700

PR #1354, opened 2024-02-15
[EVM] Fix tx index in tx receipt

jewei1997:

Fix tx index in tx receipt to only include evm txs and exclude cosmos txs + cleaned up unit tests a bit

unit tests


evmrpc/tx.go
```diff
@@ -270,19 +269,37 @@ func getEthTxForTxBz(tx tmtypes.Tx, decoder sdk.TxDecoder) *ethtypes.Transaction
...
+		}
+		if etx.Hash() == common.HexToHash(receipt.TxHashHex) {
+			foundTx = true
+			break
+		}
+		evmTxIndex++
+	}
+	if !foundTx {
+		return nil, errors.New("failed to find transaction in block")
+	}
```
jewei1997, line 295: this would be quite bad, could panic as well.

evmrpc/tx.go
```diff
@@ -270,19 +269,37 @@ func getEthTxForTxBz(tx tmtypes.Tx, decoder sdk.TxDecoder) *ethtypes.Transaction
...
 	}
+	// convert tx index including cosmos txs to tx index excluding cosmos txs
+	evmTxIndex := 0
+	foundTx := false
+	for _, tx := range block.Block.Txs {
+		etx := getEthTxForTxBz(tx, decoder)
+		if etx == nil { // cosmos tx, skip
+			continue
+		}
+		if etx.Hash() == common.HexToHash(receipt.TxHashHex) {
```
codchen: would be cheaper to just check `if i == receipt.TransactionIndex` where `for i, tx := range ...`

codchen (approved):

PR #1355, opened 2024-02-15
Evm native send bug

Kbhat1:


philipsu522:
turned out to be a bug in the evm test itself

PR #1359, opened 2024-02-16
Fix subscribe endpoints

philipsu522:


evmrpc/subscribe.go
```diff
@@ -183,32 +183,41 @@ func (s *SubscriptionManager) Unsubscribe(ctx context.Context, id SubscriberID)
 func encodeTmHeader(
 	header tmtypes.EventDataNewBlockHeader,
 ) (map[string]interface{}, error) {
+	blockHash := common.HexToHash(header.Header.Hash().String())
 	number := big.NewInt(header.Header.Height)
-	miner := common.HexToAddress(string(header.Header.ProposerAddress))
+	miner := common.HexToAddress(header.Header.ProposerAddress.String())
```
philipsu522, line 188: this is the issue - cannot just cast to string. instead you need to use the String method which does:
```

func (bz HexBytes) String() string {
	return strings.ToUpper(hex.EncodeToString(bz))
}

```
otherwise it's invalid but common.HexToHash/Address silently fails

stevenlanders (approved):

codchen (approved):

PR #1368, opened 2024-02-21
[EVM] fix gas tip calculation related to priority

stevenlanders:
- uses floats instead of int truncation
- uses effective gas price to support legacy/dynamic/blob/etc

- new unit test

### DIFF
--- a/x/evm/ante/fee.go
+++ b/x/evm/ante/fee.go
@@ -6,7 +6,6 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	"github.com/ethereum/go-ethereum/consensus/misc/eip4844"
-	ethtypes "github.com/ethereum/go-ethereum/core/types"
 	"github.com/sei-protocol/sei-chain/x/evm/derived"
 	evmkeeper "github.com/sei-protocol/sei-chain/x/evm/keeper"
 	"github.com/sei-protocol/sei-chain/x/evm/state"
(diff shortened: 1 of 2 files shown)


x/evm/ante/fee_test.go
```diff
@@ -169,3 +169,99 @@ func TestEVMFeeCheckDecoratorCancun(t *testing.T) {
...
+		name             string
+		txData           ethtypes.TxData
+		expectedPriority *big.Int
+	}{
+		{
+			name: "DynamicFeeTx with tip",
+			txData: &ethtypes.DynamicFeeTx{
+				GasFeeCap: big.NewInt(10000000000000),
+				GasTipCap: big.NewInt(10000000000000),
+				Gas:       1000,
```
jewei1997: Can we have a unit test with a more realistic value for Gas? I think the min gas price we can have on our chain is 1gwei which is 10**9.

stevenlanders: actually just removed gas because it isn't used (I originally wrote this for the previous version that calculated gas price from gas)

x/evm/ante/fee.go
```diff
@@ -80,20 +80,9 @@ func (fc EVMFeeCheckDecorator) getMinimumFee(ctx sdk.Context) *big.Int {
...
-	if txData.GetGasTipCap() != nil {
-		tipCapPct = new(big.Int).Quo(txData.GetGasTipCap(), txData.GetGasFeeCap())
-		if tipCapPct.Cmp(big.NewInt(1)) > 0 {
-			tipCapPct = big.NewInt(1)
-		}
-	}
-	discountedFee := new(big.Int).Mul(fee, tipCapPct)
-	adjustedFee := new(big.Int).Quo(discountedFee, state.UseiToSweiMultiplier)
-	nativeGasPrice := new(big.Int).Quo(adjustedFee, new(big.Int).SetUint64(txData.GetGas()))
+	gp := txData.EffectiveGasPrice(big.NewInt(0))
```
jewei1997, line 86: why are you passing in 0 for base fee?

x/evm/ante/fee_test.go
jewei1997: ah my bad this is gas limit not gas price

x/evm/ante/fee.go
jewei1997: gotcha -- this is bc since base fee is burned and doesnt go to validator, the validator wouldnt care about it for prioritization. Might be good to add a comment here.

jewei1997 (approved):

PR #1376, opened 2024-02-23
Make certain precompile functions payable

codchen:
Right now for precompile functions like `delegate`, we ask caller to send the amount to be delegated in the data payload rather than the `value` field of the EVM transaction. This leads to a bad user experience because the user will see that they are paying 0 Sei when signing on MetaMask. This PR changes the following precompile functions so that deposits are received in the solidity `payment` function convention:
- staking:delegate
- gov:deposit
- wasmd:execute,instantiate
Some notes:
- `bank` precompile currently has a `send` function that's used for non-usei sending which would not be made payable. The future addition of `send to cosmos address` function would need to be made payable since it's sending usei (cc: @Kbhat1 @jewei1997 )
- `wasmd`'s `execute` and `instantiate` functions still takes `coins` in the data payload because it's possible for non-usei coins to be sent. We added validations to make sure `usei` is not in the payload and must be sent via `value`

unit tests


precompiles/common/precompiles.go
```diff
@@ -53,3 +55,22 @@ func AssertArgsLength(args []interface{}, length int) {
...
+}
+
+func HandlePayment(ctx sdk.Context, precompileAddr sdk.AccAddress, payer sdk.AccAddress, value *big.Int, bankKeeper BankKeeper) (sdk.Coin, error) {
+	usei, wei := state.SplitUseiWeiAmount(value)
+	if !wei.IsZero() {
+		return sdk.Coin{}, fmt.Errorf("selected precompile function does not allow payment with non-zero wei remainder: received %s", value)
+	}
+	coin := sdk.NewCoin(sdk.MustGetBaseDenom(), usei)
+	// refund payer because the following precompile logic will debit the payments from payer's account
+	if err := bankKeeper.SendCoins(ctx, precompileAddr, payer, sdk.NewCoins(coin)); err != nil {
```
jewei1997, line 72: hm this seems to send coins from the precompile to the payer. Why does this need to happen?

codchen: https://github.com/sei-protocol/go-ethereum/blob/master/core/vm/evm.go#L216

jewei1997:
so for IBank precompile do we want to split up sei bank sends to be payable for usei and keep regular interface for other denoms?

philipsu522 (approved):

PR #1382, opened 2024-02-26
Fix error handling for non-existent block

yzang2019:
This PR improves error handling for cases where block is not found either by hash or by number. Currently, when block doesn't exist, it will panic and return "method handler crashes", this is not ideal because it makes user confusing and thought the server crashed. We should instead return a more informal response when block is not found.

Tested locally with docker chain

### DIFF
--- a/evmrpc/info.go
+++ b/evmrpc/info.go
@@ -143,7 +143,7 @@ func (i *InfoAPI) FeeHistory(ctx context.Context, blockCount math.HexOrDecimal64
 		}
 		result.BaseFee = append(result.BaseFee, (*hexutil.Big)(baseFee))
 		height := blockNum
-		block, err := i.tmClient.Block(ctx, &height)
+		block, err := blockWithRetry(ctx, i.tmClient, &height)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


evmrpc/info.go
```diff
@@ -143,7 +143,7 @@ func (i *InfoAPI) FeeHistory(ctx context.Context, blockCount math.HexOrDecimal64
 		}
 		result.BaseFee = append(result.BaseFee, (*hexutil.Big)(baseFee))
 		height := blockNum
-		block, err := i.tmClient.Block(ctx, &height)
+		block, err := blockWithRetry(ctx, i.tmClient, &height)
```
codchen: i'm a bit concerned about the perf impact of always retry - for cases like this we know if the block isn't there on the first try then it won't be there even if we retry.

I think we can pass a `retryAttempts` arg to `blockWithRetry` and set it to 0 for cases we don't really need retries, if the goal here is simply to have unified error handling

yzang2019: Good call, this should be fixed now, please take another look

codchen (approved):

PR #1393, opened 2024-02-28
Evm seiv2

udpatil:
This rebases the evm branch onto the seiv2 branch without squashing. Note that some intermediate commits may no longer by valid builds due to modifications in dependencies and call patterns as part of merge conflict resolution


codchen (approved):
mostly looked at app/ folders change which lgtm. `x/evm` and `evmrpc/` files are pure additions so should be ok

philipsu522 (approved):

PR #1400, opened 2024-02-29
Fix mnemonic for populate accounts

philipsu522:
We introduced an "evm_address" field which caused the position of the mnemonic to change. Instead, use the last value in array.
```
- name: psu-test
  type: local
  address: sei179ep77c0x4d5gccfpk5wq2dm0sfpsxp9mllaru
  evm_address: ""
  pubkey: '{"@type":"/cosmos.crypto.secp256k1.PubKey","key":"AvHWnr1680gyFscSzqJpFgRFA6l6p6m7NPXiKIMmV67t"}'
  mnemonic: ""


**Important** write this mnemonic phrase in a safe place.
It is the only way to recover your account if you ever forget your password.

mountain uphold immune nuclear tomato educate number capital orbit horse body under giraffe trumpet rural aisle grocery drive shallow gather merge manage stand spare
```
Tested locally

### DIFF
--- a/loadtest/scripts/populate_genesis_accounts.py
+++ b/loadtest/scripts/populate_genesis_accounts.py
@@ -27,7 +27,7 @@ def add_key(account_name, local=False):
 
     splitted_outputs = add_key_output.split('\n')
     address = splitted_outputs[3].split(': ')[1]
-    mnemonic = splitted_outputs[11]
+    mnemonic = splitted_outputs[-1]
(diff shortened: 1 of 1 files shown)


udpatil (approved):

udpatil (approved):
If we want to make it really robust, we could query the key information with output `json` instead of toml, and then parse it as a dictionary in python and key by `output["mnemonic"]`

PR #1412, opened 2024-03-07
Yzang/sei 6785

yzang2019:


go.mod
```diff
@@ -349,7 +349,7 @@ replace (
 	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.2.73-evm-rebase-11
 	github.com/cosmos/iavl => github.com/sei-protocol/sei-iavl v0.1.9
 	github.com/cosmos/ibc-go/v3 => github.com/sei-protocol/sei-ibc-go/v3 v3.3.0
-	github.com/ethereum/go-ethereum => github.com/sei-protocol/go-ethereum v1.13.5-sei-10
+	github.com/ethereum/go-ethereum => ../go-ethereum
```
yzang2019: Will change this later on

evmrpc/config.go
```diff
@@ -70,6 +70,9 @@ type Config struct {
 
 	// controls whether to have txns go through one by one
 	Slow bool `mapstructure:"slow"`
+
+	// Deny list defines list of methods that EVM RPC should fail fast
+	DenyList []string `mapstructure:"deny_list"`
```
codchen, line 75: would this be comma separated?

codchen (approved):

evmrpc/rpcstack_test.go
```diff
@@ -614,3 +614,28 @@ func (s *testService) Greet() string {
 func (s *testService) Sleep() {
 	time.Sleep(1500 * time.Millisecond)
 }
+
+func TestHttpDenyList(t *testing.T) {
+	const (
+		expectRes = `{"jsonrpc":"2.0","id":null,"error":{"code":-32700,"message":"method test_sleep is in deny list"}}`
```
stevenlanders: I think this maybe should be Method Not Found?

```
-32700	Parse error	Invalid JSON was received by the server.
An error occurred on the server while parsing the JSON text.
-32600	Invalid Request	The JSON sent is not a valid Request object.
-32601	Method not found	The method does not exist / is not available.
-32602	Invalid params	Invalid method parameter(s).
-32603	Internal error	Internal JSON-RPC error.
-32000 to -32099	Server error	Reserved for implementation-defined server-errors.
```

stevenlanders: reference: https://www.jsonrpc.org/specification#error_object

evmrpc/config.go
yzang2019: It will be an array of strings, like ["eth_a", "eth_b"]

evmrpc/rpcstack_test.go
yzang2019: Make sense! This is fixed

PR #1419, opened 2024-03-11
Add utils to replay ETH mainnet txs

codchen:

Even though this PR added a new state prefix, it would never be written/read if `eth_replay.enabled = false`, which will be the case for all actual network nodes.


app/eth_replay.go
```diff
@@ -0,0 +1,93 @@
...
+	_, err = a.InitChain(context.Background(), &abci.RequestInitChain{
+		Time:          time.Now(),
+		ChainId:       "sei-chain",
+		AppStateBytes: gendoc.AppState,
+	})
+	if err != nil {
+		panic(err)
+	}
+	for h := int64(1); h <= int64(a.EvmKeeper.EthReplayConfig.NumBlocksToReplay); h++ {
+		fmt.Printf("Replaying block height %d\n", h+int64(a.EvmKeeper.EthReplayConfig.EthDataEarliestBlock))
```
yzang2019, line 34: switch to logger?

app/eth_replay.go
```diff
@@ -0,0 +1,93 @@
...
+			Height:            h,
+			Hash:              hash,
+			Time:              time.Now(),
+		})
+		if err != nil {
+			panic(err)
+		}
+		for i, tx := range b.Txs {
+			if tx.To() != nil {
+				fmt.Printf("Verifying balance of tx %d\n", i)
```
yzang2019, line 53: same

x/evm/keeper/genesis.go
```diff
@@ -1,23 +1,85 @@
...
+		params.ChainId = sdk.OneInt()
+		k.SetParams(ctx, params)
+		if k.EthReplayConfig.EthDataEarliestBlock == 0 {
+			k.EthReplayConfig.EthDataEarliestBlock = uint64(header.Number.Int64())
+		}
+		ethReplayInitialied = true
+	}
+}
+
+func (k *Keeper) openEthDatabase() (*ethtypes.Header, state.Database, state.Trie) {
```
yzang2019, line 49: We are not using SeiDB for state access?

codchen: this is the eth-side db

app/eth_replay.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+func Replay(a *App) {
+	gendoc, err := tmtypes.GenesisDocFromFile(filepath.Join(DefaultNodeHome, "config/genesis.json"))
+	if err != nil {
+		panic(err)
+	}
+	_, err = a.InitChain(context.Background(), &abci.RequestInitChain{
+		Time:          time.Now(),
+		ChainId:       "sei-chain",
```
philipsu522, line 27: can we grab this from gendoc? should be top level field "chain_id"

philipsu522 (approved):

PR #1427, opened 2024-03-12
Add deny list config for EVM RPC

yzang2019:
This PR add a new configuration for EVM RPC to configure deny list. If a method appears in deny list, the server will return an error response to indicate the method is not found.

Config would be look like:
deny_list = ["eth_a", "eth_b"]

Added unit test for deny list.

### DIFF
--- a/evmrpc/rpcstack_test.go
+++ b/evmrpc/rpcstack_test.go
@@ -614,3 +614,28 @@ func (s *testService) Greet() string {
 func (s *testService) Sleep() {
 	time.Sleep(1500 * time.Millisecond)
 }
+
+func TestHttpDenyList(t *testing.T) {
+	const (
+		expectRes = `{"jsonrpc":"2.0","id":null,"error":{"code":-32601,"message":"the method test_sleep does not exist/is not available"}}`
+	)
+	// Set-up server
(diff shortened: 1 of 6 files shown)


codchen (approved):

evmrpc/rpcstack_test.go
```diff
@@ -614,3 +614,28 @@ func (s *testService) Greet() string {
...
+
+func TestHttpDenyList(t *testing.T) {
+	const (
+		expectRes = `{"jsonrpc":"2.0","id":null,"error":{"code":-32601,"message":"the method test_sleep does not exist/is not available"}}`
+	)
+	// Set-up server
+	timeouts := rpc.DefaultHTTPTimeouts
+	timeouts.WriteTimeout = time.Second
+	srv := createAndStartServer(t, &evmrpc.HTTPConfig{
+		DenyList: []string{"test_sleep"},
```
jewei1997, line 626: I don't think the `test_sleep` endpoint exists on this server. Might be good to try to denylist an endpoint that exists like `test` for this test.

yzang2019: oh just saw this, I think we added that as a new module in the config itself, if you look at previous test cases, it seems to be a valid endpoint

PR #1442, opened 2024-03-15
Associate Tx Prefix Fix

Kbhat1:
- Fixes issue where Associate evm tx didn't work when signature had Ethereum Signed Message Prepended
- Makes Associate tx generic with custom message of any type (including `Ethereum Signed Message`) 

Example usage on client side 
```
const ethers = require('ethers');

const PK = '<private_key>'

const associate = async () => {
  const provider = new ethers.JsonRpcProvider(
    "http://127.0.0.1:8545/",
    713715
  );
  const wallet = new ethers.Wallet(PK, provider);
  console.log("Address:", wallet.address);
  const emptyData = Buffer.alloc(32); // This creates a buffer with 32 bytes set to 0
  const signature = await wallet.signMessage(emptyData);
  console.log("Signature: ", signature);
  const { r, s, v } = ethers.Signature.from(signature);
  const compactV = v - 27;

  const method = "sei_associate";
  const emptyDataString = emptyData.toString();
  const messageToSign = `\x19Ethereum Signed Message:\n32${emptyDataString}`; // Concatenate the prefix and empty data string
  const params = [{ r, s, v: `0x${compactV}`, custom_message: messageToSign}];
  console.log("Params: ", params);
  const associated = await provider.send(method, params);
  console.log("Associated: ", associated);
};

associate().catch(console.error);
```

- Tested on local chain with wallet signing and sending associate tx


x/evm/ante/preprocess.go
```diff
@@ -136,7 +136,9 @@ func Preprocess(ctx sdk.Context, msgEVMTransaction *evmtypes.MsgEVMTransaction)
 	if atx, ok := txData.(*ethtx.AssociateTx); ok {
 		V, R, S := atx.GetRawSignatureValues()
 		V = new(big.Int).Add(V, big.NewInt(27))
-		evmAddr, seiAddr, pubkey, err := getAddresses(V, R, S, common.Hash{}) // associate tx should sign over an empty hash
+		// Hash custom message passed in
+		customMessageHash := crypto.Keccak256Hash([]byte(atx.CustomMessage))
```
codchen, line 140: This assumes that the caller would sign the message after Keccak256Hash'es it. I wonder whether it'd be more flexible if we just treat `custom_message` as already "hashed" and do `getAddresses(V, R, S, common.BytesToHash([]byte(atx.CustomMessage))`. That way if the caller wants Keccak256Hash they can just set `custom_message` to be the hashed value, whereas if they want other hash (or no hash at all) they can set accordingly.

codchen (approved):

x/evm/ante/preprocess.go
besated: it seems like the [eip-191](https://eips.ethereum.org/EIPS/eip-191) standard uses `keccak256("\x19Ethereum Signed Message:\n" + len(message) + message))`

besated (approved):

x/evm/ante/preprocess.go
Kbhat1: Yea think it’s better to just assume custom message isn’t hashed @codchen

PR #1444, opened 2024-03-15
Adding tokenfactory denom metadata endpoint

dssei:
When we create tokens using token factory, denoms are created in the `factory/{owner_walltet_id}/{denom}`  format 
E.g. `factory/sei1gxskuzvhr4s8sdm2rpruaf7yx2dnmjn0zfdu9q/NEWCOIN`

When querying metadata via `/cosmos/bank/v1beta1/denoms_metadata/{denom}` endpoint for such a denom we are getting error like 
```
{
    "code": 12,
    "message": "Not Implemented",
    "details": []
}
```
This happens, because routing logic splits the denom uri param into 3 components rather then one and hence tries to router request to a non-existing handler. Url encoding does not help either as it still results in same issue. We would like to add additional endpoint to handle token factory created metadata retrieval, and move denom for the endpoint from uri param to query param.

so then for token factory created tokens the query would be 
`/sei-protocol/seichain/tokenfactory/denoms/metadata?denom=factory/{owner_walltet_id}/{denom}`

- Functional testing on local node with updated source
- unit/integration tests


proto/README.md
```diff
@@ -0,0 +1,34 @@
...
+·
+··
+
+Ignite CLI version:     v0.23.0
+Ignite CLI build date:  2022-07-24T18:17:44Z
+Ignite CLI source hash: 64df9aef958b3e8bc04b40d9feeb03426075ea89
+Your OS:                darwin
+Your arch:              arm64
+Your go version:        go version go1.22.0 darwin/arm64
+Your uname -a:          Darwin Denyss-MacBook-Pro.local 23.1.0 Darwin Kernel Version 23.1.0: Mon Oct  9 21:32:11 PDT 2023; root:xnu-10002.41.9~7/RELEASE_ARM64_T6030 arm64
```
codebycarson: Do you want your username in here?

proto/README.md
```diff
@@ -0,0 +1,34 @@
...
+Your cwd:               /Users/denyssinyakov/repos/sei-chain
+Is on Gitpod:           false
+
+```
+Then, to generate the code, run the following command:
+
+```bash
+ignite generate proto-go
```
codebycarson, line 33: Is this for generating the proto files or for generating the openapi yml?

go.sum
```diff
@@ -2246,12 +2252,12 @@ google.golang.org/genproto v0.0.0-20220304144024-325a89244dc8/go.mod h1:kGP+zUP2
...
-google.golang.org/genproto/googleapis/api v0.0.0-20230920204549-e6e6cdab5c13/go.mod h1:RdyHbowztCGQySiCvQPgWQWgWhGnouTdCflKoDBt32U=
-google.golang.org/genproto/googleapis/rpc v0.0.0-20231012201019-e917dd12ba7a h1:a2MQQVoTo96JC9PMGtGBymLp7+/RzpFc2yX/9WfFg1c=
-google.golang.org/genproto/googleapis/rpc v0.0.0-20231012201019-e917dd12ba7a/go.mod h1:4cYg8o5yUbm77w8ZX00LhMVNl/YVBFJRYWDc0uYWMs0=
+google.golang.org/genproto v0.0.0-20240116215550-a9fa1716bcac h1:ZL/Teoy/ZGnzyrqK/Optxxp2pmVh+fmJ97slxSRyzUg=
+google.golang.org/genproto v0.0.0-20240116215550-a9fa1716bcac/go.mod h1:+Rvu7ElI+aLzyDQhpHMFMMltsD6m7nqpuWDd2CwJw3k=
+google.golang.org/genproto/googleapis/api v0.0.0-20240125205218-1f4bbc51befe h1:0poefMBYvYbs7g5UkjS6HcxBPaTRAmznle9jnxYoAI8=
+google.golang.org/genproto/googleapis/api v0.0.0-20240125205218-1f4bbc51befe/go.mod h1:4jWUdICTdgc3Ibxmr8nAJiiLHwQBY0UI0XZcEMaFKaA=
+google.golang.org/genproto/googleapis/rpc v0.0.0-20240125205218-1f4bbc51befe h1:bQnxqljG/wqi4NTXu2+DJ3n7APcEA882QZ1JvhQAq9o=
+google.golang.org/genproto/googleapis/rpc v0.0.0-20240125205218-1f4bbc51befe/go.mod h1:PAREbraiVEVGVdTZsVWjSbbTtSyGbAgIIvni8a8CD5s=
```
codebycarson: I would make sure protocol team signs off on these version changes

proto/README.md
dssei: Its for both, will add another readme for docs generation in a separate PR

proto/README.md
dssei: Good catch

go.mod
```diff
@@ -40,9 +40,9 @@ require (
 	golang.org/x/sync v0.5.0
 	golang.org/x/text v0.14.0
 	golang.org/x/time v0.3.0
-	google.golang.org/genproto/googleapis/api v0.0.0-20230920204549-e6e6cdab5c13
-	google.golang.org/grpc v1.58.3
-	google.golang.org/protobuf v1.31.0
+	google.golang.org/genproto/googleapis/api v0.0.0-20240125205218-1f4bbc51befe
+	google.golang.org/grpc v1.61.0
+	google.golang.org/protobuf v1.33.0
```
yzang2019: Is there a reason to bump protobuf version?

dssei: hm, good question. I was not specifically looking to bump. But with `ignite` tools installed, I think this was done automatically on generation step. I can roll this back

dssei: @yzang2019 addressed this

proto/README.md
yzang2019: This is really nicee!

yzang2019 (approved):

PR #1445, opened 2024-03-17
Oracle Precompile

Kbhat1:
## Describe your changes and provide context
- Oracle precompile to query exchange rates + twaps
- Oracle precompile integration tests

## Testing performed to validate your change
- Added integration tests, unit tests and verified on chain

PR #1459, opened 2024-03-21
IBC transfer precompile

dssei:
- Adding precompile for IBC transfer
- Moved some common methods to `common/precompiles.go`

- Unit testing
- local e2e testing in progress


precompiles/ibc/abi.json
```diff
@@ -0,0 +1,46 @@
+[
```
dssei, line 1: keeping formatted for readability for now (will flatten when code is post-draft)

precompiles/ibc/ibc.go
```diff
@@ -0,0 +1,170 @@
...
+		case TransferMethod:
+			p.TransferID = m.ID
+		}
+	}
+
+	return p, nil
+}
+
+// RequiredGas returns the required bare minimum gas to execute the precompile.
+func (p Precompile) RequiredGas(input []byte) uint64 {
```
dssei, line 80: @Kbhat1 @codchen for IBC transfer, will this have to be hardcoded as for other precompiles? Any suggested value?

precompiles/ibc/ibc.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	switch method.Name {
+	case TransferMethod:
+		return p.transfer(ctx, method, args)
+	}
+	return
+}
+func (p Precompile) transfer(ctx sdk.Context, method *abi.Method, args []interface{}) ([]byte, error) {
+	pcommon.AssertArgsLength(args, 6)
```
dssei: @Kbhat1 @codchen do we need to do other validations?

precompiles/ibc/ibc.go
```diff
@@ -0,0 +1,170 @@
...
+		// short circuit
+		return method.Outputs.Pack(true)
+	}
+
+	coin := sdk.Coin{
+		Denom:  denom,
+		Amount: sdk.NewIntFromBigInt(amount),
+	}
+
+	height := clienttypes.Height{}
```
dssei: @codchen @Kbhat1 Not sure what values should be here...or should they be provided by the caller via ABI?

precompiles/ibc/IBC.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+address constant BANK_PRECOMPILE_ADDRESS = 0x0000000000000000000000000000000000001008;
```
Kbhat1: Could we udpate this to 0x0000000000000000000000000000000000001009 ? Oracle is about to be merged https://github.com/sei-protocol/sei-chain/pull/1445

dssei: Updated

precompiles/ibc/ibc.go
codchen: Is IBC gas consumption pretty constant or can vary from call-to-call quite a bit? If the latter we can report gas consumption dynamically similar to https://github.com/sei-protocol/sei-chain/blob/seiv2/precompiles/wasmd/wasmd.go#L111-L135

precompiles/ibc/ibc.go
codchen: I see that `from` is set an arg. Can anyone trigger an IBC transfer on anyone else's behalf? If not then we should  be looking at  `caller` in https://github.com/sei-protocol/sei-chain/pull/1459/files#diff-3c39140061c990bd99f477fe4dbc73c0eaac283bfa477c1ba439016daf6fe688R87 to use as the from address instead

precompiles/ibc/ibc.go
codchen: will to defer to kartik here since it's IBC domain knowledge

Kbhat1: That's the timeout height, we can pass that in as a parameter in the transfer method and update the abi

precompiles/ibc/ibc.go
Kbhat1: Can reference https://github.com/sei-protocol/sei-chain/blob/seiv2/precompiles/bank/bank.go#L116 for example usage of `caller`

codchen (approved):

PR #1461, opened 2024-03-22
Parallelize tx decoding/EVM preprocessing

codchen:
The decoding/preprocessing process is stateless so can be parallelized without non-determinism concerns.

unit test

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1487,26 +1487,8 @@ func (app *App) ProcessBlock(ctx sdk.Context, txs [][]byte, req BlockProcessRequ
 	events = append(events, beginBlockResp.Events...)
 
 	txResults := make([]*abci.ExecTxResult, len(txs))
-	typedTxs := []sdk.Tx{}
+	typedTxs := app.DecodeTransactionsConcurrently(ctx, txs)
 
-	for i, tx := range txs {
-		typedTx, err := app.txDecoder(tx)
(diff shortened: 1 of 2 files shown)


app/app.go
```diff
@@ -1540,6 +1522,35 @@ func (app *App) ProcessBlock(ctx sdk.Context, txs [][]byte, req BlockProcessRequ
...
+			// get txkey from tx
+			if err != nil {
+				ctx.Logger().Error(fmt.Sprintf("error decoding transaction at index %d due to %s", idx, err))
+				typedTxs[idx] = nil
+			} else {
+				if isEVM, _ := evmante.IsEVMMessage(typedTx); isEVM {
+					msg := evmtypes.MustGetEVMTransactionMessage(typedTx)
+					if err := evmante.Preprocess(ctx, msg); err != nil {
+						ctx.Logger().Error(fmt.Sprintf("error preprocessing EVM tx due to %s", err))
+						typedTxs[idx] = nil
```
yzang2019, line 1542: Will this trigger nil pointer exception later on and cause chain to halt?

codchen: this is existing behavior. And i think it's handled here https://github.com/sei-protocol/sei-chain/blob/seiv2/app/app.go#L1336-L1341

yzang2019 (approved):

PR #1490, opened 2024-03-28
Add ERC pointer registry

codchen:
Add ERC pointer registry that's versioned. A registration can be written in the following ways:
- through `pointer` precompile for cw20/721 contracts as well as bank tokens that have metadata stored (e.g. tokenfactory tokens)
- through governance proposals (e.g. bank tokens without metadata stored). This part will be added in a separate PR.

A read-only precompile `pointerview` is also added for EVM contracts to query the registry at a flat gas cost.

unit test


precompiles/pointer/pointer.go
```diff
@@ -0,0 +1,257 @@
...
+
+	switch method.Name {
+	case AddNativePointer:
+		return p.AddNative(ctx, method, caller, args, value, evm, suppliedGas)
+	case AddCW20Pointer:
+		return p.AddCW20(ctx, method, caller, args, value, evm, suppliedGas)
+	case AddCW721Pointer:
+		return p.AddCW721(ctx, method, caller, args, value, evm, suppliedGas)
+	}
+	return
```
stevenlanders, line 108: does it make sense to handle a `default:` with a returned err?

precompiles/pointer/pointer.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+func (p Precompile) Run(*vm.EVM, common.Address, []byte, *big.Int) ([]byte, error) {
+	panic("static gas Run is not implemented for dynamic gas precompile")
+}
+
+func (p Precompile) AddNative(ctx sdk.Context, method *ethabi.Method, caller common.Address, args []interface{}, value *big.Int, evm *vm.EVM, suppliedGas uint64) (ret []byte, remainingGas uint64, err error) {
+	pcommon.AssertArgsLength(args, 1)
+	token := args[0].(string)
```
stevenlanders, line 117: what happens if they pass no args?

precompiles/pointer/pointer.go
```diff
@@ -0,0 +1,257 @@
...
+	if exists && existingVersion >= native.CurrentVersion {
+		return nil, 0, fmt.Errorf("pointer at %s with version %d exists when trying to set pointer for version %d", existingAddr.Hex(), existingVersion, native.CurrentVersion)
+	}
+	metadata, metadataExists := p.bankKeeper.GetDenomMetaData(ctx, token)
+	if !metadataExists {
+		return nil, 0, fmt.Errorf("denom %s does not have metadata stored and thus can only have its pointer set through gov proposal", token)
+	}
+	name := metadata.Name
+	symbol := metadata.Symbol
+	decimals := uint8(0)
```
stevenlanders:
```suggestion
	var decimals uint8
```

precompiles/pointer/pointer.go
```diff
@@ -0,0 +1,257 @@
...
+			decimals = uint8(denomUnit.Exponent)
+			name = denomUnit.Denom
+			symbol = denomUnit.Denom
+			if len(denomUnit.Aliases) > 0 {
+				name = denomUnit.Aliases[0]
+			}
+		}
+	}
+	abi := native.GetABI()
+	parsedABI, err := ethabi.JSON(strings.NewReader(string(abi)))
```
stevenlanders: side thought: i wonder if it makes sense to cache the parsedABI in memory after first access?  (json memory/cpu can be expensive under load)

precompiles/pointerview/pointerview.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	for name, m := range newAbi.Methods {
+		switch name {
+		case GetNativePointer:
+			p.GetNativePointerID = m.ID
+		case GetCW20Pointer:
+			p.GetCW20PointerID = m.ID
+		case GetCW721Pointer:
+			p.GetCW721PointerID = m.ID
```
stevenlanders, line 65: same as above, i'd suggest a `default:` with a method-not-found err

stevenlanders (approved):
LGTM - I don't see any issues, just some comments.  I do wonder whether we have the dynamically parse the abi.json per invocation?

precompiles/pointer/pointer.go
codchen: the call would panic and fail (the panic would be caught on the DeliverTx level)

precompiles/pointer/pointer.go
codchen: good point - updated

PR #1500, opened 2024-04-02
Add WSEI contract and cmd for deployment

codchen:
Based on https://etherscan.io/address/0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2#code plus fixes to breaking changes in solidity versioning

local sei deployment


contracts/src/WSEI.sol
```diff
@@ -0,0 +1,66 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+contract WSEI {
```
jewei1997, line 4: might be good to declare that it follows IERC20 to get compiler checks.

contracts/src/WSEI.sol
```diff
@@ -0,0 +1,66 @@
...
+    function deposit() public payable {
+        balanceOf[msg.sender] += msg.value;
+        emit Deposit(msg.sender, msg.value);
+    }
+    function withdraw(uint wad) public {
+        require(balanceOf[msg.sender] >= wad);
+        balanceOf[msg.sender] -= wad;
+        payable(msg.sender).transfer(wad);
+        emit Withdrawal(msg.sender, wad);
+    }
```
jewei1997, line 32: It's super important that we follow checks-effects-interaction here (which you did do here), but if these lines are changed in order in any way, we could have a serious vuln. Would recommend at minimum commenting that we need to follow checks-effects-interaction and also comment each section. We can also add a nonreentrant modifier to be extra safe.

contracts/src/WSEI.sol
```diff
@@ -0,0 +1,66 @@
...
+
+    function totalSupply() public view returns (uint) {
+        return address(this).balance;
+    }
```
jewei1997, line 36: need to be very careful about pulling the balance of the contract to use as totalSupply. We probably should not do this and use a separate totalSupply variable. Here is some more info about what can go wrong https://www.immunebytes.com/blog/self-destruct-exploit-forced-ether-injection-in-solidity-contracts/

contracts/src/WSEI.sol
```diff
@@ -0,0 +1,66 @@
...
+        allowance[msg.sender][guy] = wad;
+        emit Approval(msg.sender, guy, wad);
+        return true;
+    }
+
+    function transfer(address dst, uint wad) public returns (bool) {
+        return transferFrom(msg.sender, dst, wad);
+    }
+
+    function transferFrom(address src, address dst, uint wad)
```
jewei1997, line 48: could check that src and dst are both not address(0) for user safety https://github.com/OpenZeppelin/openzeppelin-contracts/blob/cb2aaaa04a292887c49839cd958b08a83979d746/contracts/token/ERC20/ERC20.sol#L166-L172

contracts/src/WSEI.sol
```diff
@@ -0,0 +1,66 @@
...
+        public
+        returns (bool)
+    {
+        require(balanceOf[src] >= wad);
+
+        if (src != msg.sender && allowance[src][msg.sender] != type(uint).max) {
+            require(allowance[src][msg.sender] >= wad);
```
jewei1997, line 55: might be good to have revert error messages to help users figure out which check they failed

contracts/src/WSEI.sol
```diff
@@ -0,0 +1,66 @@
+// SPDX-License-Identifier: MIT
```
jewei1997, line 1: might be good to have some basic foundry tests to make sure things work

jewei1997:
I guess you can actually ignore my suggestions since you're using the official WETH contract and just changed the name of the contract.

philipsu522 (approved):

PR #1506, opened 2024-04-03
Use registry instead of code hash to determine precompile whitelist

codchen:
Use pointer registry to determine which contracts are whitelisted to `delegatecall` bank/wasmd precompiles.

unit tests


precompiles/ibc/ibc.go
```diff
@@ -94,8 +94,8 @@ func (p Precompile) RunAndCalculateGas(evm *vm.EVM, caller common.Address, calli
 	if err != nil {
 		return nil, 0, err
 	}
-	if err = pcommon.ValidateCaller(ctx, p.evmKeeper, caller, callingContract); err != nil {
-		return nil, 0, err
+	if caller.Cmp(callingContract) != 0 {
+		return nil, 0, errors.New("cannot delegatecall IBC")
```
philipsu522, line 103: is this always b/c of IBC?

codchen: yeah IBC should never have delegate calls

PR #1512, opened 2024-04-05
Write receipt for txs with non VM errors

codchen:
We want to have some records for transactions that fail with a panic. This PR write receipts for transactions that had a panic during processing.

unit test


jewei1997 (approved):

jewei1997:
Can we add a hardhat test for this?

jewei1997:
Would also be good for @stevenlanders to review this

codchen:
@jewei1997 we don't know of a way to trigger a panic 😅 (like if we know one then it's a bug that we need to fix). This change is more for preparing us for the bugs that we don't know of

PR #1543, opened 2024-04-15
[EVM] charge fees in ante handler

codchen:
Previously gas fees are charged inside `msg_server`, which means if any non-VM error is thrown in `msg_server`, the fee charge would be reverted alongside everything else. This makes it free for attackers who are able to identify a non-VM error trigger in `msg_server` (e.g. a bug in sei code) to spam the chain. This PR frontload gas fee charges to the ante handler, so that errors in `msg_server` would not affect fee charges.

One complexity is around surplus handling: we keep track of surplus/deficit when manipulating balances within EVM world because credits and debits are disjoint in EVM. While we charge fees in ante handler, we don't want to credit those fees to fee collectors until the transaction finishes processing in `msg_server`, because that's the only time we'd know how much fees need to be refunded. As such, the single-leg debit for fee charge in ante handler will have its associated surplus carried over into `msg_server` and settled against all other surpluses.

TODO

### DIFF
--- a/x/evm/ante/fee.go
+++ b/x/evm/ante/fee.go
@@ -6,6 +6,8 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	"github.com/ethereum/go-ethereum/consensus/misc/eip4844"
+	"github.com/ethereum/go-ethereum/core"
+	"github.com/ethereum/go-ethereum/core/vm"
 	"github.com/sei-protocol/sei-chain/app/antedecorators"
 	"github.com/sei-protocol/sei-chain/utils"
 	"github.com/sei-protocol/sei-chain/x/evm/derived"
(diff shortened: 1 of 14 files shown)


x/evm/ante/fee.go
```diff
@@ -55,13 +57,29 @@ func (fc EVMFeeCheckDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate b
...
+	if err != nil {
+		return ctx, err
+	}
+	cfg := evmtypes.DefaultChainConfig().EthereumConfig(fc.evmKeeper.ChainID(ctx))
+	txCtx := core.NewEVMTxContext(emsg)
+	evmInstance := vm.NewEVM(*blockCtx, txCtx, stateDB, cfg, vm.Config{})
+	stateDB.SetEVM(evmInstance)
+	st := core.NewStateTransition(evmInstance, emsg, &gp, true)
+	if err := st.BuyGas(); err != nil {
+		return ctx, sdkerrors.Wrap(sdkerrors.ErrInsufficientFunds, err.Error())
```
stevenlanders, line 83: are there any other errors possible here (always insufficient funds)?

stevenlanders (approved):

x/evm/ante/fee.go
codchen: yea insufficient funds is the only possible error https://github.com/sei-protocol/go-ethereum/blob/master/core/state_transition.go#L259

PR #1546, opened 2024-04-16
Cosmos gas multiplier params

Kbhat1:
- Add Cosmos gas multiplier numerator and denominator as param instead of hardcoded
- Removes wasm dependency discount
- Relies on sei-cosmos PR: https://github.com/sei-protocol/sei-cosmos/pull/487
- Relies on sei-wasmd PR: https://github.com/sei-protocol/sei-wasmd/pull/45

- Tested on local chain and verified multiplier


app/antedecorators/depdecorators/gas.go
```diff
@@ -10,6 +10,7 @@ type GasMeterSetterDecorator struct {
 }
 
 func (d GasMeterSetterDecorator) AnteDeps(txDeps []sdkacltypes.AccessOperation, tx sdk.Tx, txIndex int, next sdk.AnteDepGenerator) (newTxDeps []sdkacltypes.AccessOperation, err error) {
+	// TODO: Remove below since no more wasm dependency discount
```
Kbhat1, line 13: Verifying in local chain / tests and then will remove this since we have no more wasm dependency discount

go.mod
```diff
@@ -344,7 +344,7 @@ require (
 )
 
 replace (
-	github.com/CosmWasm/wasmd => github.com/sei-protocol/sei-wasmd v0.1.0
+	github.com/CosmWasm/wasmd => github.com/sei-protocol/sei-wasmd v0.0.10-0.20240415230844-461d7d3b8890 // TODO: Update once sei-wasmd PR merged
```
Kbhat1, line 347: This will be updated once sei-wasmd new tag is published

go.mod
```diff
@@ -354,7 +354,7 @@ replace (
 	github.com/sei-protocol/sei-db => github.com/sei-protocol/sei-db v0.0.35
 	// Latest goleveldb is broken, we have to stick to this version
 	github.com/syndtr/goleveldb => github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
-	github.com/tendermint/tendermint => github.com/sei-protocol/sei-tendermint v0.2.41-seiv2
+	github.com/tendermint/tendermint => github.com/sei-protocol/sei-tendermint v0.2.41-seiv2.0.20240416201336-c52dc88f0dd3 // TODO: Update once sei-tendermint PR merged
```
Kbhat1: This will be updated once sei-tendermint new tag is published

app/app.go
```diff
@@ -1609,8 +1609,10 @@ func (app *App) getFinalizeBlockResponse(appHash []byte, events []abci.Event, tx
 		}),
 		ConsensusParamUpdates: &tmproto.ConsensusParams{
 			Block: &tmproto.BlockParams{
-				MaxBytes: endBlockResp.ConsensusParamUpdates.Block.MaxBytes,
-				MaxGas:   endBlockResp.ConsensusParamUpdates.Block.MaxGas,
+				MaxBytes:                       endBlockResp.ConsensusParamUpdates.Block.MaxBytes,
+				MaxGas:                         endBlockResp.ConsensusParamUpdates.Block.MaxGas,
+				CosmosGasMultiplierNumerator:   endBlockResp.ConsensusParamUpdates.Block.CosmosGasMultiplierNumerator,
```
codchen: do we need migration for new consensus params?

codchen (approved):

app/app.go
Kbhat1: @codchen we removed this in favor of just adding in a new param in cosmos layer: https://github.com/sei-protocol/sei-cosmos/pull/487

go.mod
```diff
@@ -344,9 +344,9 @@ require (
 )
 
 replace (
-	github.com/CosmWasm/wasmd => github.com/sei-protocol/sei-wasmd v0.1.0
+	github.com/CosmWasm/wasmd => github.com/sei-protocol/sei-wasmd v0.0.10-0.20240415230844-461d7d3b8890 // TODO: Update once sei-wasmd PR merged
 	github.com/confio/ics23/go => github.com/cosmos/cosmos-sdk/ics23/go v0.8.0
-	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.2.80-seiv2
+	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.2.81-seiv2.0.20240417141123-88fd96d3e0b4 // TODO: Update once sei-cosmos PR merged
```
Kbhat1: This will be updated once sei-cosmos new tag is published

app/antedecorators/gas.go
```diff
@@ -1,82 +1,19 @@
...
-	GasMultiplierNumerator                   uint64 = 1
-	DefaultGasMultiplierDenominator          uint64 = 1
-	WasmCorrectDependencyDiscountDenominator uint64 = 2
-)
-
-func GetGasMeterSetter(aclkeeper aclkeeper.Keeper) func(bool, sdk.Context, uint64, sdk.Tx) sdk.Context {
+func GetGasMeterSetter(pk paramskeeper.Keeper) func(bool, sdk.Context, uint64, sdk.Tx) sdk.Context {
 	return func(simulate bool, ctx sdk.Context, gasLimit uint64, tx sdk.Tx) sdk.Context {
-		if simulate || ctx.BlockHeight() == 0 {
+		if ctx.BlockHeight() == 0 {
```
udpatil, line 11: why don't we have simulate specific gas meter behavior? -  we should still have some infinite gas meter that then applies the scaling factor to be able to accurate simulate, right?

udpatil: we should be careful here given that the cachekv is no longer emitting events, and this is a difference in how the store accesses happen

udpatil: I think for the purposes of this change, just to be safe, we should reintroduce the event emissions in cachekv (despite performance consequences) to ensure the parallel dag resource accesses remain accurate. This will need to be updated in sei-cosmos and then reflected here

Kbhat1: @udpatil for first comment, new infinite multiplier gas meter has been added for simulation

PR #1558, opened 2024-04-17
Add limit to filter logs

codchen:
We want to apply a limit for log-related queries similar to how Alchemy does for its endpoints:
Screen Shot 2024-04-17 at 11 53 57 AM

unit test

### DIFF
--- a/evmrpc/config.go
+++ b/evmrpc/config.go
@@ -73,6 +73,12 @@ type Config struct {
 
 	// Deny list defines list of methods that EVM RPC should fail fast
 	DenyList []string `mapstructure:"deny_list"`
+
+	// max number of logs returned if block range is open-ended
+	MaxLogNoBlock int64 `mapstructure:"max_log_no_block"`
+
+	// max number of blocks to query logs for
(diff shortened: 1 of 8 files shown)


evmrpc/config.go
```diff
@@ -93,6 +99,8 @@ var DefaultConfig = Config{
 	MaxTxPoolTxs:         1000,
 	Slow:                 false,
 	DenyList:             make([]string, 0),
+	MaxLogNoBlock:        10000,
+	MaxBlocksForLog:      2000,
```
stevenlanders, line 103: do you think we should make this larger given we have such fast blocks?  (is it disruptive to only be able to query ~11 minutes of logs?)

codchen: i think the limit is applied "post-filter" so in most cases it would be able to cover much longer periods? For blanket filters that don't filter anything I think the recommendation from alchemy etc. is to use the subscribe endpoint instead which doesn't have such a limit

stevenlanders (approved):

PR #1576, opened 2024-04-22
[EVM] Add integration test for cw20 pointer 

stevenlanders:
- adds integration test for cw20
  - Deploys ERC20
  - Deploys CW20 Pointer to that ERC20
  - Calls query functions 
- renames tests to be clearer 
- dynamically invokes inside docker if it's running, otherwise runs locally

- test added to PR


contracts/test/CW20toERC20PointerTest.js
```diff
@@ -0,0 +1,76 @@
...
+        try {
+            await queryWasm(addr, operation, args);
+            // If the promise resolves, force the test to fail
+            expect.fail(`Expected rejection: address=${addr} operation=${operation} args=${JSON.stringify(args)}`);
+        } catch (error) {
+            expect(error.message).to.include("ERC20 does not support");
+        }
+    }
+
+    describe("query", function(){
```
jewei1997, line 46: should we also test write functions?

contracts/test/lib.js
```diff
@@ -0,0 +1,101 @@
...
+                command = command.replace(/\.\.\//g, "/sei-protocol/sei-chain/");
+                const dockerCommand = `docker exec sei-node-0 /bin/bash -c 'export PATH=$PATH:/root/go/bin:/root/.foundry/bin && printf "12345678\\n" | ${command}'`;
+                execCommand(dockerCommand, resolve, reject);
+            } else {
+                // The container is not running, execute command normally
+                execCommand(command, resolve, reject);
+            }
+        });
+    });
+}
```
jewei1997, line 135: nice

jewei1997 (approved):
overall LGTM, might be good to test a write function for CW20->ERC20.

contracts/test/CW20toERC20PointerTest.js
stevenlanders: added `transfer` which seems to work, though found an issue with approval - thinking we should fix in separate PR

PR #1583, opened 2024-04-22
Update init script to enable seidb and occ

udpatil:


Kbhat1 (approved):

scripts/initialize_local_chain.sh
```diff
@@ -70,6 +70,17 @@ END_DATE_5DAYS=$($PYTHON_CMD -c "from datetime import datetime, timedelta; print
...
 
+if [ ! -z "$2" ]; then
+  APP_TOML_PATH="$2"
+else
+  APP_TOML_PATH="$HOME/.sei/config/app.toml"
+fi
+# Enable OCC and SeiDB
+sed -i.bak -e 's/# concurrency-workers = .*/concurrency-workers = 500/' $APP_TOML_PATH
+sed -i.bak -e 's/occ-enabled = .*/occ-enabled = true/' $APP_TOML_PATH
+sed -i.bak -e 's/sc-enable = .*/sc-enable = true/' $APP_TOML_PATH
```
Kbhat1, line 81: Can we set ss-enable to true as well? Otherwise we won’t have historical queries. In other chains it’s fine since rpcs turn this on but here it’s different

udpatil: sure will update

Kbhat1 (approved):

PR #1590, opened 2024-04-23
Disallow negative value passed into internal EVM calls

udpatil:
This disallows passing negative values into the Internal EVM Call function.

Unit Test

### DIFF
--- a/x/evm/keeper/evm.go
+++ b/x/evm/keeper/evm.go
@@ -73,6 +73,9 @@ func (k *Keeper) CallEVM(ctx sdk.Context, from common.Address, to *common.Addres
 		gp := k.GetGasPool()
 		value := utils.Big0
 		if val != nil {
+			if val.IsNegative() {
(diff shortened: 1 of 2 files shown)


philipsu522 (approved):

x/evm/keeper/evm.go
```diff
@@ -73,6 +73,9 @@ func (k *Keeper) CallEVM(ctx sdk.Context, from common.Address, to *common.Addres
 		gp := k.GetGasPool()
 		value := utils.Big0
 		if val != nil {
+			if val.IsNegative() {
```
codchen: we need this check for the `else` case as well so let's move it to before `if evm == nil`

codchen (commented):
we should probably also add this check https://github.com/sei-protocol/sei-chain/blob/main/x/evm/ante/basic.go#L27-L29 in this wasmbinding code path

codchen (approved):

PR #1595, opened 2024-04-24
Add Evm Receipt Status Metrics

Kbhat1:
- Adds metrics for evm receipt status 

- Verifying on local chain


philipsu522 (approved):

x/evm/keeper/msg_server.go
```diff
@@ -98,6 +98,14 @@ func (server msgServer) EVMTransaction(goCtx context.Context, msg *types.MsgEVMT
 			)
 			return
 		}
+
+		// Add metrics for receipt status
+		if receipt.Status == uint32(ethtypes.ReceiptStatusFailed) {
+			telemetry.IncrCounter(1, "receipt", "status", "failed")
```
stevenlanders, line 104: what's the resulting metric name from this?  is it just `receipt`?

Kbhat1: would be `receipt_status_failed`

PR #1600, opened 2024-04-26
Remove Non-Multiplier gas Meter

Kbhat1:
- Relies on https://github.com/sei-protocol/sei-wasmd/pull/46
- Relies on https://github.com/sei-protocol/sei-wasmd/pull/46
- Removes non multiplier gas meter 

- Testing on local chain


app/antedecorators/gas.go
```diff
@@ -8,14 +8,10 @@ import (
 
 func GetGasMeterSetter(pk paramskeeper.Keeper) func(bool, sdk.Context, uint64, sdk.Tx) sdk.Context {
 	return func(simulate bool, ctx sdk.Context, gasLimit uint64, tx sdk.Tx) sdk.Context {
-		if ctx.BlockHeight() == 0 {
```
codchen, line 11: would `CosmosGasParams` be set at this point when block height is 0? Just want to make sure it won't panic if we remove this

codchen (approved):

app/antedecorators/gas.go
Kbhat1: Yea the GetCosmosGasParams return the default values if it’s not set https://github.com/sei-protocol/sei-cosmos/blob/bc0008f38b52fa49ce98616d926fb61ed9db8234/x/params/keeper/keeper.go#L75

udpatil (approved):

PR #1604, opened 2024-04-29
Emit EVM logs for pointer precompile

codchen:
Emit pointer deployment log that's indexed on the underlying, with data being the newly deployed pointer address

unit test

### DIFF
--- a/precompiles/pointer/pointer.go
+++ b/precompiles/pointer/pointer.go
@@ -13,7 +13,9 @@ import (
 	ethabi "github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/tracing"
+	ethtypes "github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/core/vm"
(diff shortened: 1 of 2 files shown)


precompiles/pointer/pointer.go
```diff
@@ -32,6 +34,9 @@ const PointerAddress = "0x000000000000000000000000000000000000100b"
 
 var _ vm.PrecompiledContract = &Precompile{}
 var _ vm.DynamicGasPrecompiledContract = &Precompile{}
+var TopicNativePointerDeployed = crypto.Keccak256Hash([]byte("native_deployed"))
+var TopicCW20PointerDeployed = crypto.Keccak256Hash([]byte("cw20_deployed"))
```
jewei1997, line 38: if the cw side address is the first topic then I think we may want something like `CW20Deployed(string)` as the event signature that gets hashed as the first topic.

jewei1997 (approved):

jewei1997 (approved):
overall LGTM, one potential fix

precompiles/pointer/pointer.go
jewei1997: oh actually it might be something like `CW20Deployed(string,address)`

PR #1615, opened 2024-05-01
optimize WS newhead

codchen:
instead of having each subscription subscribing to tendermint and encode to JSON independently, we will maintain a single subscription with tendermint and a single copy of encoded JSON per head, then fan it out to all outgoing subscriptions

### DIFF
--- a/evmrpc/subscribe.go
+++ b/evmrpc/subscribe.go
@@ -25,61 +25,98 @@ type SubscriptionAPI struct {
 	subscriptionManager *SubscriptionManager
 	subscriptonConfig   *SubscriptionConfig
 
-	logFetcher *LogFetcher
+	logFetcher          *LogFetcher
+	newHeadListenersMtx *sync.Mutex
+	newHeadListeners    map[rpc.ID]chan map[string]interface{}
 }
 
 type SubscriptionConfig struct {
 	subscriptionCapacity int
 }
 
-func NewSubscriptionAPI(tmClient rpcclient.Client, logFetcher *LogFetcher, subscriptionConfig *SubscriptionConfig) *SubscriptionAPI {
-	logFetcher.filterConfig = &FilterConfig{}
-	return &SubscriptionAPI{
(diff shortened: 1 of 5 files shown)


evmrpc/subscribe.go
```diff
@@ -25,61 +25,98 @@ type SubscriptionAPI struct {
...
-					return
-				}
-				err = notifier.Notify(rpcSub.ID, ethHeader)
-				if err != nil {
-					return
+					break OUTER
 				}
 			case <-rpcSub.Err():
-				return
+				break OUTER
```
stevenlanders, line 101: is this the same as `continue` here?

evmrpc/subscribe.go
```diff
@@ -25,61 +25,98 @@ type SubscriptionAPI struct {
...
+	logFetcher.filterConfig = filterConfig
+	api := &SubscriptionAPI{
 		tmClient:            tmClient,
 		subscriptionManager: NewSubscriptionManager(tmClient),
 		subscriptonConfig:   subscriptionConfig,
 		logFetcher:          logFetcher,
+		newHeadListenersMtx: &sync.Mutex{},
+		newHeadListeners:    make(map[rpc.ID]chan map[string]interface{}),
 	}
+	id, subCh, err := api.subscriptionManager.Subscribe(context.Background(), NewHeadQueryBuilder(), api.subscriptonConfig.subscriptionCapacity)
```
stevenlanders, line 47: does this need supervision or some mechanism to retry/reconnect (can the subscription die?)

stevenlanders (approved):

philipsu522 (approved):

evmrpc/subscribe.go
codchen: no we want to break out of the outer loop (instead of the select clause)

evmrpc/subscribe.go
codchen: I think the subscription manager has a retry mechanism (that's what the `subscription was canceled, resubscribing` log we saw)

PR #1634, opened 2024-05-07
Add reverse mapping for pointers

yzang2019:
Currently, there's only one directional mapping to record the pointer address, this PR is going to make it bidirectional. Instead of only keeping track of A->B, now we record both A->B and B->A, which makes search easier.

Added and fixed existing unit test

### DIFF
--- a/x/evm/keeper/pointer.go
+++ b/x/evm/keeper/pointer.go
@@ -16,11 +16,19 @@ import (
 )
 
 func (k *Keeper) SetERC20NativePointer(ctx sdk.Context, token string, addr common.Address) error {
-	return k.SetPointerInfo(ctx, types.PointerERC20NativeKey(token), addr[:], native.CurrentVersion)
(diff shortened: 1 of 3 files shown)


codchen (approved):

stevenlanders (approved):
lgtm - fwiw, one design thought is to have a single k.SetBothWays kind of entrypoint, which would encapsulate the "reverse side", and make the SetPointerInfo a private method.  The  danger of separating is similar to other places (associate + migrate balance) where we might forget that both directions (set one way + set other way) always must happen.

yzang2019:
> lgtm - fwiw, one design thought is to have a single k.SetBothWays kind of entrypoint, which would encapsulate the "reverse side", and make the SetPointerInfo a private method. The danger of separating is similar to other places (associate + migrate balance) where we might forget that both directions (set one way + set other way) always must happen.

Make sense, in fact I can first make SetPointerInfo private

PR #1635, opened 2024-05-07
Fix ante surplus handling

codchen:
Background:
Because EVM updates balance on one side at a time, it is possible to incur surplus during the execution (e.g. one credit of 100wei and one debit of 110wei would result in a surplus of 10wei). Such surplus are handled at `EndBlock` (sent to EVM module account) to make sure no coin is burnt/minted silently (as in, total supply would not be updated). However, surplus incurred in the ante handler (specifically the fee handler) may not be handled if the transaction fails during processing, in which case silent burning/minting would happen.

Fix:
Instead of setting state of ante surplus in the message server, we will set it inside ante handler to make sure it's always carried over. To do that, we can no longer use in-memory state, so we use memstate instead, and clear it up at the end of EndBlock.

A migration handler is also added to fix total supply for existing chains affected (i.e. arctic-1/atlantic-2)

unit test

### DIFF
--- a/x/evm/migrations/fix_total_supply.go
+++ b/x/evm/migrations/fix_total_supply.go
@@ -29,6 +29,9 @@ func FixTotalSupply(ctx sdk.Context, k *keeper.Keeper) error {
 		weiInUsei = weiInUsei.Add(sdk.OneInt())
 	}
 	correctSupply = correctSupply.Add(weiInUsei)
-	k.BankKeeper().SetSupply(ctx, sdk.NewCoin(sdk.MustGetBaseDenom(), correctSupply))
+	currentSupply := k.BankKeeper().GetSupply(ctx, sdk.MustGetBaseDenom()).Amount
+	if !currentSupply.Equal(correctSupply) {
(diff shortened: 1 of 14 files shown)


x/evm/migrations/fix_total_supply.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	bankkeeper "github.com/cosmos/cosmos-sdk/x/bank/keeper"
+	"github.com/sei-protocol/sei-chain/x/evm/keeper"
+	"github.com/sei-protocol/sei-chain/x/evm/types"
+)
+
+// This migration is to fix total supply mismatch caused by mishandled
+// ante surplus
+func FixTotalSupply(ctx sdk.Context, k *keeper.Keeper) error {
```
udpatil, line 12: should we also be specifying chain IDs for the migration, since this won't be an issue for all envs, but the migration will still run on all of them regardless?

udpatil (approved):

x/evm/migrations/fix_total_supply.go
codchen: added a check to only update supply if it's incorrect

PR #1660, opened 2024-05-13
Add light total supply invariance check

codchen:
Add a lightweight check that checks the following invariance every block:
- The total balance of all addresses who have seen any change to their usei/wei balance in the current block should equal to their total balance as of the last block, after adjusting for any burning/minting (as reflected in changes in the `TotalSupply` state)

WIP trying to make SC store work in unit tests

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -372,9 +372,8 @@ type App struct {
 
 	HardForkManager *upgrades.HardForkManager
 
-	encodingConfig        appparams.EncodingConfig
-	evmRPCConfig          evmrpc.Config
-	lightInvarianceConfig LightInvarianceConfig
+	encodingConfig appparams.EncodingConfig
+	evmRPCConfig   evmrpc.Config
 }
 
 // New returns a reference to an initialized blockchain app
@@ -618,11 +617,6 @@ func New(
 		}
 		app.EvmKeeper.EthClient = ethclient.NewClient(rpcclient)
 	}
-	lightInvarianceConfig, err := ReadLightInvarianceConfig(appOpts)
-	if err != nil {
-		panic(fmt.Sprintf("error reading light invariance config due to %s", err))
-	}
-	app.lightInvarianceConfig = lightInvarianceConfig
 
 	customDependencyGenerators := aclmapping.NewCustomDependencyGenerator()
 	aclOpts = append(aclOpts, aclkeeper.WithResourceTypeToStoreKeyMap(aclutils.ResourceTypeToStoreKeyMap))
@@ -1133,7 +1127,7 @@ func (app *App) FinalizeBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlock)
 				return &abci.ResponseFinalizeBlock{}, nil
 			}
 			cms := app.WriteState()
-			app.LightInvarianceChecks(cms, app.lightInvarianceConfig)
+			app.LightInvarianceChecks(cms)
 			appHash := app.GetWorkingHash()
 			resp := app.getFinalizeBlockResponse(appHash, app.optimisticProcessingInfo.Events, app.optimisticProcessingInfo.TxRes, app.optimisticProcessingInfo.EndBlockResp)
 			return &resp, nil
@@ -1150,7 +1144,7 @@ func (app *App) FinalizeBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlock)
 		return &abci.ResponseFinalizeBlock{}, nil
 	}
 	cms := app.WriteState()
-	app.LightInvarianceChecks(cms, app.lightInvarianceConfig)
+	app.LightInvarianceChecks(cms)
 	appHash := app.GetWorkingHash()
 	resp := app.getFinalizeBlockResponse(appHash, events, txResults, endBlockResp)
 	return &resp, nil
--- a/app/app.go
+++ b/app/app.go
@@ -372,8 +372,9 @@ type App struct {
 
 	HardForkManager *upgrades.HardForkManager
 
-	encodingConfig appparams.EncodingConfig
-	evmRPCConfig   evmrpc.Config
+	encodingConfig        appparams.EncodingConfig
+	evmRPCConfig          evmrpc.Config
+	lightInvarianceConfig LightInvarianceConfig
 }
 
 // New returns a reference to an initialized blockchain app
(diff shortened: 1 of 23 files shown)


app/invariance.go
```diff
@@ -0,0 +1,131 @@
...
+	useiChangedAddr := []sdk.AccAddress{}
+	for _, p := range balanceChangePairs {
+		if len(p.Key) < 2 {
+			// invalid key; ignore
+			continue
+		}
+		addrLen := int(p.Key[1])
+		if len(p.Key) < addrLen+2 {
+			// invalid key length; ignore
+			continue
```
philipsu522: are there cases where this is possible?

philipsu522: if this is just a sanity cehck let's add logs

app/invariance.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+		var balance sdk.Coin
+		if err := balance.Unmarshal(val); err != nil {
+			app.Logger().Error(fmt.Sprintf("failed to unmarshal preblock balance: %s", err))
+			continue
+		}
+		useiPreTotal = useiPreTotal.Add(balance.Amount)
+	}
+	weiChangePairs := ckv.GetChangedPairs(banktypes.WeiBalancesPrefix)
```
philipsu522, line 134: longer term there's a lot of places handling wei & sei directly - we can think of what sort of abstraction makes sense so that we don't need to handle them separately in the future

app/invariance.go
```diff
@@ -0,0 +1,131 @@
...
+		preTotalSupply = amt
+	}
+	for _, p := range totalSupplyChangePairs {
+		if string(p.Key[1:]) == sdk.MustGetBaseDenom() {
+			if p.Delete {
+				supplyChanged = preTotalSupply.Neg()
+			} else {
+				var amt sdk.Int
+				if err := amt.Unmarshal(p.Value); err != nil {
+					app.Logger().Error(fmt.Sprintf("failed to unmarshal total supply: %s", err))
```
philipsu522, line 229: if this is going to be critical, let's add a new metric like "lightvariance_failure" counter so we can alert on any cases in this code where the invariance is unable to be computed

app/invariance.go
```diff
@@ -0,0 +1,131 @@
...
+			continue
+		}
+		var amt sdk.Int
+		if err := amt.Unmarshal(val); err != nil {
+			app.Logger().Error(fmt.Sprintf("failed to unmarshal preblock wei balance: %s", err))
+			continue
+		}
+		weiPreTotal = weiPreTotal.Add(amt)
+	}
+	totalSupplyChangePairs := ckv.GetChangedPairs(banktypes.SupplyKey)
```
philipsu522, line 193: does this take into account tokens minted as part of the token release?

app/invariance.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+	}
+	weiDiff := weiPostTotal.Sub(weiPreTotal)
+	weiDiffInUsei, weiDiffRemainder := bankkeeper.SplitUseiWeiAmount(weiDiff)
+	if !weiDiffRemainder.IsZero() {
+		panic(fmt.Sprintf("non-zero wei diff found! Pre-block wei total %s, post-block wei total %s", weiPreTotal, weiPostTotal))
+	}
+	useiDiff := useiPreTotal.Sub(useiPostTotal).Sub(weiDiffInUsei).Add(supplyChanged)
+	if !useiDiff.IsZero() {
+		panic(fmt.Sprintf("unexpected usei balance total found! Pre-block usei total %s wei total %s total supply %s, post-block usei total %s wei total %s total supply %s", useiPreTotal, weiPreTotal, preTotalSupply, useiPostTotal, weiPostTotal, preTotalSupply.Add(supplyChanged)))
```
philipsu522, line 244: does this inavariance check take into account:
1. delegation / undelegation / redelegation
2. slashing
3. vesting
?

codchen: yes since all of those are done via bank sends

app/invariance.go
codchen: yes since mints update total supply

app/invariance.go
codchen: just sanity check

app/app.go
```diff
@@ -1126,7 +1132,8 @@ func (app *App) FinalizeBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlock)
 			if app.EvmKeeper.EthReplayConfig.Enabled || app.EvmKeeper.EthBlockTestConfig.Enabled {
 				return &abci.ResponseFinalizeBlock{}, nil
 			}
-			app.WriteState()
+			cms := app.WriteState()
+			app.LightInvarianceChecks(cms, app.lightInvarianceConfig)
```
udpatil, line 1136: would we want to throw this into a goroutine so it's concurrent with rest of finalize blocker? Not sure the timescale of this check so if its miniscule its probably not a big deal

app/invariance.go
```diff
@@ -0,0 +1,247 @@
...
+			metrics.IncrCounterWithLabels([]string{"sei", "lightinvariance_supply", "invalid_changed_key"}, 1, []metrics.Label{
+				{
+					Name:  "type",
+					Value: "sei",
+				},
+			})
+			app.Logger().Error(fmt.Sprintf("invalid changed pair key for usei: %X", p.Key))
+			continue
+		}
+		addrLen := int(p.Key[1])
```
udpatil, line 72: should this be p.Key[:2] so that it takes the first two bytes (which is the addr len prefix I believe)

app/invariance.go
```diff
@@ -0,0 +1,247 @@
...
+					supplyChanged = amt.Sub(preTotalSupply)
+				}
+			}
+			break
+		}
+	}
+	weiDiff := weiPostTotal.Sub(weiPreTotal)
+	weiDiffInUsei, weiDiffRemainder := bankkeeper.SplitUseiWeiAmount(weiDiff)
+	if !weiDiffRemainder.IsZero() {
+		panic(fmt.Sprintf("non-zero wei diff found! Pre-block wei total %s, post-block wei total %s", weiPreTotal, weiPostTotal))
```
udpatil, line 240: If we panic here instead of emitting a metric, how do we plan to alert on the failure? Additionally, even if we add a metric here and leave the panic in, won't the alert not work because it would be very likely that the metrics endpoint stops serving metrics data due to the binary panicking? Therefore, the monitoring node wouldn't appropriately scrape the metric for invariance failure, right?

udpatil (approved):

app/app.go
codchen: I think this is already at pretty much the end of FinalizeBlocker?

app/invariance.go
codchen: I think the first byte is the BalancePrefix, and the second is the actual length

app/invariance.go
codchen: failure would cause the node running this to fall behind, so we can probably just rely on the relevant alerting there

app/app.go
udpatil: yeah thats true i guess, just wondering if itll add overhead to a node that causes it to fall behind (maybe not in its current state, but during heavy load maybe)

app/invariance.go
udpatil: ahh yeah good point

app/invariance.go
udpatil: yeah but theres other causes for nodes to halt and such, so it won't be a more granular alert, right? I thought the rationale would be to have more sensitive alerting on invariance failure as opposed to waiting for 5m or whatever the normal rule is

PR #1661, opened 2024-05-13
Do not modify the original value passed in executeBatch call

dssei:
Changing `executeBatch` to operate on a `value` arg copy rather than original value (that is being modified during calculations).
Needed to preserve the value immutability since gas estimating logic does assume that. 

- Unit test
- Local node testing


codchen (approved):

udpatil (approved):
Great find, lgtm

PR #1663, opened 2024-05-14
Reuse the same contract address when upgrading ERC->CW pointers

codchen:
Previously pointer contract will be assigned a new address when it's upgraded to a new version, which is quite inconvenient to end users and applications. Since pointer contracts themselves are stateless, it is possible to reuse the same address for a new version of code. This PR implements that for the EVM->CW direction pointers. A similar change will be made for CW->EVM pointers

unit test

### DIFF
--- a/x/evm/keeper/pointer_upgrade.go
+++ b/x/evm/keeper/pointer_upgrade.go
@@ -1,0 +1,108 @@
+package keeper
+
+import (
+	"math"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core"
+	"github.com/ethereum/go-ethereum/core/vm"
+	"github.com/sei-protocol/sei-chain/utils"
+	"github.com/sei-protocol/sei-chain/x/evm/artifacts"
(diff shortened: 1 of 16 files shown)


x/evm/keeper/pointer_upgrade.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		logger("get block context", err.Error())
+		return err
+	}
+	cfg := types.DefaultChainConfig().EthereumConfig(k.ChainID(ctx))
+	txCtx := core.NewEVMTxContext(&core.Message{From: evmModuleAddress, GasPrice: utils.Big0})
+	evmInstance := vm.NewEVM(*blockCtx, txCtx, stateDB, cfg, vm.Config{})
+	err = runner(evmInstance)
+	if err != nil {
+		logger("upserting pointer", err.Error())
```
jewei1997, line 32: Does this failing always mean that upserting pointer failed?

codchen: yeah it means it failed for this particular pointer

jewei1997 (approved):

stevenlanders (approved):
Is there a way to protect against someone putting a non-backwards-compatible change in the updated pointer code?  (like swapping state variable order?)

PR #1665, opened 2024-05-14
[EVM] Test erc20 to cw20 pointer upgrade

stevenlanders:
- adds new integration test
- deploys pointer, tests pointer, upgrades and tests pointer again
- adds new app.toml `enable_test_api = true` 
  - allows ticking the pointer version
  - only works on non-live chains
  - version must be adjusted on every node 
- new integration test


jewei1997:
I think you might need to add your app.toml updates here as well https://github.com/sei-protocol/sei-chain/blob/da7e1d8f02d9b1120682613ab8b01781ef8c19a6/cmd/seid/cmd/root.go#L518

contracts/test/ERC20toCW20PointerTest.js
```diff
@@ -1,194 +1,267 @@
...
+            });
         });
+    }
 
-        it("get balanceOf", async function () {
-            expect(await pointer.balanceOf(admin.evmAddress)).to.equal(1000000)
-            expect(await pointer.balanceOf(accounts[0].evmAddress)).to.equal(2000000);
-            expect(await pointer.balanceOf(accounts[1].evmAddress)).to.equal(3000000);
-        });
+    describe("Original Pointer", function () {
```
jewei1997: nit: update to "Pointer Upgradeability"

contracts/test/ERC20toCW20PointerTest.js
```diff
@@ -1,194 +1,267 @@
...
-            await expect(pointer.transfer(unassociatedRecipient, 1)).to.be.revertedWithoutReason;
         });
 
-        it("transfer to contract address should succeed", async function() {
-            const contract = await pointer.getAddress();
-            const tx = await pointer.transfer(contract, 1);
-            await tx.wait();
-        });
+        // this does not yet pass, so skip
+        describe.skip("Original Pointer after Upgrade", function(){
```
jewei1997, line 248: why does this not work yet?

stevenlanders:
> I think you might need to add your app.toml updates here as well
> 
> https://github.com/sei-protocol/sei-chain/blob/da7e1d8f02d9b1120682613ab8b01781ef8c19a6/cmd/seid/cmd/root.go#L518
> 
> . Basically when you do `make install`, `enable_test_api` should show up in app.toml (as false?).

I was thinking we actually wouldn't want this particular one to show up to avoid security issues or confusion, and because it's such a integration-test specific situation (was on the border of not wanting a config at all)

contracts/test/ERC20toCW20PointerTest.js
stevenlanders: Because the original pointer does not work after upgrade at the moment (it isn't supported)

jewei1997 (approved):

PR #1669, opened 2024-05-15
Support ERC/CW2981 in pointer contracts

codchen:
This PR adds ERC/CW2981 support, which involves the following query endpoints:
- For ERC721 pointing to CW721
    - `supportsInterface`
    - `royaltyInfo`
- For CW721 pointing to ERC721
    - `extension:royalty_info`
    - `extension:check_royalties`

Note that setting royalties needs to be done via `mint` which is not currently supported by the pointers yet.

integration tests


stevenlanders (approved):
lgtm - i think the integration test should probably be in hardhat but we can do that later

contracts/src/CW721ERC721Pointer.sol
```diff
@@ -82,6 +92,26 @@ contract CW721ERC721Pointer is ERC721 {
...
+        }
+        string memory tId = _formatPayload("token_id", _doubleQuotes(Strings.toString(tokenId)));
+        string memory sPrice = _formatPayload("sale_price", _doubleQuotes(Strings.toString(salePrice)));
+        string memory req = _curlyBrace(_formatPayload("royalty_info", _curlyBrace(_join(tId, sPrice, ","))));
+        string memory fullReq = _curlyBrace(_formatPayload("extension", _curlyBrace(_formatPayload("msg", req))));
+        bytes memory response = WasmdPrecompile.query(Cw721Address, bytes(fullReq));
+        bytes memory addr = JsonPrecompile.extractAsBytes(response, "address");
+        uint256 amt = JsonPrecompile.extractAsUint256(response, "royalty_amount");
+        if (addr.length == 0) {
+            return (address(0), amt);
```
jewei1997, line 110: should this be `(address(0), 0)`? I think we probably don't want to risk people burning funds on mistake.

codchen: the cw2981 implementation makes look like it's a possible scenario for the payment address to be empty while the royalty is non-zero: https://github.com/CosmWasm/cw-nfts/blob/main/contracts/cw2981-royalties/src/lib.rs#L41-L45, so I'm not sure if we want to overwrite `amt` here?

PR #1670, opened 2024-05-15
Withdraw multiple delegation rewards

dssei:
We need a way to withdraw multiple rewards in v2 app. This change adds new method to precompile to make that possible.

- unit test
- functional test in load test cluster


precompiles/distribution/distribution.go
```diff
@@ -151,25 +157,69 @@ func (p Precompile) setWithdrawAddress(ctx sdk.Context, method *abi.Method, call
...
 	delegator, found := p.evmKeeper.GetSeiAddress(ctx, caller)
 	if !found {
 		return nil, fmt.Errorf("delegator %s is not associated", caller.Hex())
 	}
-	validator, err := sdk.ValAddressFromBech32(args[0].(string))
+
+	return delegator, nil
+}
+
+func (p Precompile) withdrawMultipleDelegationRewards(ctx sdk.Context, method *abi.Method, caller common.Address, args []interface{}, value *big.Int) ([]byte, error) {
```
codchen: having a variable-length input would require calculating gas consumption dynamically, so we need to replay `Run()` with `RunAndCalculateGas`, similar to https://github.com/sei-protocol/sei-chain/blob/main/precompiles/wasmd/wasmd.go#L132 Specifically,
- gas limit should be converted and set in `ctx` https://github.com/sei-protocol/sei-chain/blob/main/precompiles/wasmd/wasmd.go#L145-L150
- remaining gas should be converted and returned from `ctx` https://github.com/sei-protocol/sei-chain/blob/main/precompiles/wasmd/wasmd.go#L258-L259

Right now we don't have a way to mix static and dynamic gas functions in the same precompile, so we change this new one to dynamic, we need to make the two existing functions dynamic as well

dssei: Makes sense. Updated code to use dynamic gas calculation

codchen (approved):

Kbhat1 (approved):

PR #1671, opened 2024-05-15
Update ERC721 pointer contract

dvli2007:

- Switched ERC721 pointer contract to read `name` and `symbol` from CW721 counterpart. This avoids potential human-errors in pointer contract creation that can result in discrepancies in name/symbol between the pointer contract and the base contract.
- Fix bug in `balanceOf` -- query now paginates to get count of all tokens belonging to user, rather than just the first page
- Add `totalSupply`

- Updated `test/CW721ERC721PointerTest.t.sol`


contracts/src/CW721ERC721Pointer.sol
```diff
@@ -112,6 +116,31 @@ contract CW721ERC721Pointer is ERC721,ERC2981 {
...
+        ownerOf(tokenId);
+        string memory tId = _formatPayload("token_id", _doubleQuotes(Strings.toString(tokenId)));
+        string memory req = _curlyBrace(_formatPayload("nft_info", _curlyBrace(tId)));
+        bytes memory response = WasmdPrecompile.query(Cw721Address, bytes(req));
+        bytes memory uri = JsonPrecompile.extractAsBytes(response, "token_uri");
+        return string(uri);
+    }
+
+    // 721-Enumerable
+    function totalSupply() public view virtual returns (uint256) {
```
besated, line 130: what happens for nft collections that have unlimited supply? are they then not supported by pointers?

dvli2007: `totalSupply` returns the total amount of tokens stored by the contract. As tokens get minted, this value will increase; as tokens get burned, this value will decrease. Collections with uncapped token supply will still be supported.

Reference: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/extensions/IERC721Enumerable.sol

besated: i see that makes sense, i mistook `totalSupply` for the max supply of the collection

besated (approved):
thanks for updating this!

codchen (commented):
need to bump ‘CurrentVersion’ in x/evm/artifacts/cw721/artifacts.go

PR #1677, opened 2024-05-16
No-op enumerable queries on cw721

apolloshab:
No-op enumerable queries on cw721 (implementation too gas inefficient)


dvli2007:
When gas limit is at the current 10M, the queries seem to work for collection sizes of at least 1000 (have not tested yet with greater collection sizes).

Would be open to taking suggestions on whether it's better to no-op the `tokens` and `all_tokens` queries, or to do something different.

besated (approved):

dvli2007:
I discussed with a few community members about #1677. I think since gas cap for queries is back at 10M, it may be worthwhile to keep the `tokens` and `all_tokens` queries as they are currently in the pointer contract for now.

PR #1679, opened 2024-05-16
cw721: Move transfer events after action event

apolloshab:
Move transfer events after action event attribute in cw721 transfer/send


philipsu522 (approved):

dvli2007:
More context:

CW721 events by convention will have the `action` attribute first, followed by subsequent attributes. Since the `action` attribute currently gets added after the `transfer_nft` event attributes, it ends up emitting `action` last.

To bring the contract in-line with CW721 convention, we emit the `transfer_nft` event attributes in the expected order in both `send_nft` and `transfer_nft` calls.

PR #1687, opened 2024-05-17
[EVM] Avoid approvals with native pointers

stevenlanders:
- native pointer shouldn't manipulate any state because sei-side transactions won't inform it
- native coins don't support approve / transferFrom, so logically neither can the pointer
- unit tests
- integration tests


besated (approved):
looks good! in the call, we discussed potentially updating the symbol/name to pull from the token itself - is that something we still wanna do?

stevenlanders:
> looks good! in the call, we discussed potentially updating the symbol/name to pull from the token itself - is that something we still wanna do?

I initially updated it, but then looked further at the deployment process, and there's some validation to prevent deployment if these metadata items do not exist but still allow a governance proposal registration (scenario is a native coin without metadata, which allows one to specify metadata during this deploy)

stevenlanders:
Closing for now, we believe approvals on EVM side only are okay because they enable Uniswap and other EVM-friendly situations.

PR #1688, opened 2024-05-19
Fix genesis export command (#1653)

yzang2019:

* Fix genesis export command

* Bump cosmos and seidb version

* Fix go mod dependency issue


docker/rpcnode/Dockerfile
```diff
@@ -1,8 +1,8 @@
 FROM ubuntu:latest
 RUN apt-get update && \
-    apt-get install -y make git wget build-essential  jq python3 curl vim
-RUN wget https://go.dev/dl/go1.20.6.linux-amd64.tar.gz
-RUN tar -xvf go1.20.6.linux-amd64.tar.gz
+    apt-get install -y make git wget build-essential jq python3 curl vim uuid-runtime
```
udpatil, line 3: why are we adding uuid-runtime?

udpatil (approved):

docker/rpcnode/Dockerfile
yzang2019: It's not necessary, just making it consistent with the other file, I can revert it

PR #1693, opened 2024-05-21
Add evm rpc endpoint eth_maxPriorityFeePerGas

yzang2019:
This PR add a new EVM rpc endpoint for eth_maxPriorityFeePerGas, we don't really have an exact same implementation for tip, but for now we can just return a suggested gas price using feeHist.Reward[0][0].ToInt()

Added unit test

### DIFF
--- a/evmrpc/info.go
+++ b/evmrpc/info.go
@@ -175,6 +175,20 @@ func (i *InfoAPI) FeeHistory(ctx context.Context, blockCount math.HexOrDecimal64
 	return result, nil
 }
 
+func (i *InfoAPI) MaxPriorityFeePerGas(ctx context.Context) (*hexutil.Big, error) {
+	startTime := time.Now()
+	defer recordMetrics("eth_maxPriorityFeePerGas", i.connectionType, startTime, true)
+	feeHist, err := i.FeeHistory(ctx, 1, rpc.LatestBlockNumber, []float64{0.5})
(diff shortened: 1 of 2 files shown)


evmrpc/info.go
```diff
@@ -175,6 +175,20 @@ func (i *InfoAPI) FeeHistory(ctx context.Context, blockCount math.HexOrDecimal64
...
+func (i *InfoAPI) MaxPriorityFeePerGas(ctx context.Context) (*hexutil.Big, error) {
+	startTime := time.Now()
+	defer recordMetrics("eth_maxPriorityFeePerGas", i.connectionType, startTime, true)
+	feeHist, err := i.FeeHistory(ctx, 1, rpc.LatestBlockNumber, []float64{0.5})
+	if err != nil {
+		return nil, err
+	}
+	if len(feeHist.Reward) == 0 || len(feeHist.Reward[0]) == 0 {
+		// if there is no EVM tx in the most recent block, return 0
+		return (*hexutil.Big)(big.NewInt(0)), nil
```
yzang2019, line 187: Shall we return 0 when there's no feeHistory?

jewei1997: I feel like it could be 0 since if there were no evm txs in the previous block then the user probably won't have to tip to get their txn included

jewei1997 (approved):

PR #1694, opened 2024-05-21
Add native associate tx type

codchen:
Add a cosmos TX type that allows forced address association for free. The condition is the same as EVM associate tx:
- the account exists with at least 1wei
- the account hasn't been associated

Usage:
```bash
seid tx evm native-associate <your custom msg> --from=<your key> --broadcast-mode=block
```

local sei
unit test


stevenlanders (approved):
lgtm - it may be good to add a hardhat test in just to get the end-to-end scenario

PR #1697, opened 2024-05-22
EVM gas price metrics

jewei1997:

emit metrics to pull EVM effective gas price per transaction. The EVM gas data will inform how we price gas on Sei.

manual testing + existing tests


utils/metrics/metrics_util.go
```diff
@@ -316,3 +318,18 @@ func IncrConsumerEventCount(msgType string) {
 		[]metrics.Label{telemetry.NewLabel("msg_type", msgType)},
 	)
 }
+
+// Gauge for gas price paid for transactions
+// Metric Name:
+//
+// sei_evm_effective_gas_price
+func GaugeEvmEffectiveGasPrice(gasPrice *big.Int, blockNum uint64, txHash common.Hash) {
```
stevenlanders: I'm thinking we would want this to be an increment, so we can aggregate on it (e.g., avg gas price).  It's possible that gauge has some kind of capability like that but I thought it was more of a replace concept  (set value to X).  I could be wrong on that.

utils/metrics/metrics_util.go
```diff
@@ -316,3 +318,18 @@ func IncrConsumerEventCount(msgType string) {
...
+	telemetry.SetGaugeWithLabels(
+		[]string{"sei", "evm", "effective", "gas", "price"},
+		float32(gasPrice.Uint64()),
+		[]metrics.Label{
+			telemetry.NewLabel("block_num", strconv.FormatUint(blockNum, 10)),
+			telemetry.NewLabel("tx_hash", txHash.Hex()),
```
stevenlanders: we probably don't want tx_hash as a label because it's so high-cardinality and hard to aggregate

yzang2019: Agreed

utils/metrics/metrics_util.go
```diff
@@ -316,3 +320,86 @@ func IncrConsumerEventCount(msgType string) {
...
+			vals:     make([]float32, 0),
+		}
+	}
+}
+
+func addBlockGaugeMetric(blockNumber uint64, keys []string, val float32, labels []metrics.Label) {
+	SetupBlockMetric(keys, labels, blockNumber)
+	key := strings.Join(keys, "_")
+	blockMetric := perBlockMetrics.keyToMetric[key]
+	blockMetric.mu.Lock()
```
yzang2019: If key doesn't exist, blockMetric will be nil right? Although I guess it shouldn't be nil since we just SetupBlockMetric, but adding a lock makes me think this part is not thread safe?

yzang2019: Also, let's see if we can avoid using a lock as much as possible

utils/metrics/metrics_util.go
```diff
@@ -316,3 +320,86 @@ func IncrConsumerEventCount(msgType string) {
...
+// Metric Name:
+//
+// sei_evm_effective_gas_price
+func GaugeEvmEffectiveGasPrice(gasPrice *big.Int, blockNum uint64) {
+	addBlockGaugeMetric(
+		blockNum,
+		[]string{"sei", "evm", "effective", "gas", "price"},
+		float32(gasPrice.Uint64()),
+		[]metrics.Label{
+			telemetry.NewLabel("block_num", strconv.FormatUint(blockNum, 10)),
```
yzang2019: This would blow up metrics

yzang2019 (approved):

PR #1699, opened 2024-05-23
Add tests for non-usei tokens via wasm precompile

udpatil:
This adds hardhat tests for testing the wasm precompile execute paths when sending amounts that aren't "usei" denominated.

Tests added


jewei1997 (approved):
very nice!

PR #1700, opened 2024-05-24
[EVM] Set gaslimit based on consensus params

stevenlanders:
- gas limit for a block should **not** be the sum of gas wanted values
- it should be the capacity of the block
- new unit test
- local rpc testing


evmrpc/block.go
```diff
@@ -160,11 +160,12 @@ func EncodeTmBlock(
...
 	miner := common.HexToAddress(block.Block.ProposerAddress.String())
 	baseFeePerGas := k.GetBaseFeePerGas(ctx).TruncateInt().BigInt()
-	gasLimit, gasWanted := int64(0), int64(0)
+	var gasWanted int64
 	chainConfig := types.DefaultChainConfig().EthereumConfig(k.ChainID(ctx))
 	transactions := []interface{}{}
+
 	for i, txRes := range blockRes.TxsResults {
-		gasLimit += txRes.GasWanted
+
```
jewei1997: nit: remove new line

evmrpc/block.go
```diff
@@ -160,11 +160,12 @@ func EncodeTmBlock(
 	resultHash := common.HexToHash(block.Block.LastResultsHash.String())
 	miner := common.HexToAddress(block.Block.ProposerAddress.String())
 	baseFeePerGas := k.GetBaseFeePerGas(ctx).TruncateInt().BigInt()
-	gasLimit, gasWanted := int64(0), int64(0)
+	var gasWanted int64
```
jewei1997: should we name this to gasUsed since it is the sum of `txRes.GasUsed` and returned under the "gasUsed" field?

PR #1706, opened 2024-05-29
[POC] associate via bytes conversion

stevenlanders:

### DIFF
--- a/x/evm/keeper/address.go
+++ b/x/evm/keeper/address.go
@@ -27,12 +27,19 @@ func (k *Keeper) DeleteAddressMapping(ctx sdk.Context, seiAddress sdk.AccAddress
 	store.Delete(types.SeiAddressToEVMAddressKey(seiAddress))
 }
 
+func (k *Keeper) InitAccount(ctx sdk.Context, addr common.Address) {
+	seiAddress, _ := k.GetSeiAddress(ctx, addr)
+	if !k.accountKeeper.HasAccount(ctx, seiAddress) {
+		k.accountKeeper.SetAccount(ctx, k.accountKeeper.NewAccountWithAddress(ctx, seiAddress))
+	}
+}
+
 func (k *Keeper) GetEVMAddress(ctx sdk.Context, seiAddress sdk.AccAddress) (common.Address, bool) {
 	store := ctx.KVStore(k.storeKey)
 	bz := store.Get(types.SeiAddressToEVMAddressKey(seiAddress))
 	addr := common.Address{}
 	if bz == nil {
-		return addr, false
+		return common.BytesToAddress(seiAddress), true
 	}
 	copy(addr[:], bz)
 	return addr, true
@@ -50,7 +57,7 @@ func (k *Keeper) GetSeiAddress(ctx sdk.Context, evmAddress common.Address) (sdk.
 	store := ctx.KVStore(k.storeKey)
 	bz := store.Get(types.EVMAddressToSeiAddressKey(evmAddress))
 	if bz == nil {
-		return []byte{}, false
+		return evmAddress.Bytes(), true
 	}
 	return bz, true
 }
--- a/x/evm/state/balance.go
+++ b/x/evm/state/balance.go
@@ -11,6 +11,8 @@ import (
 
 func (s *DBImpl) SubBalance(evmAddr common.Address, amt *big.Int, reason tracing.BalanceChangeReason) {
 	s.k.PrepareReplayedAddr(s.ctx, evmAddr)
+	s.k.InitAccount(s.ctx, evmAddr)
+
 	if amt.Sign() == 0 {
 		return
 	}
@@ -52,6 +54,8 @@ func (s *DBImpl) SubBalance(evmAddr common.Address, amt *big.Int, reason tracing
 
 func (s *DBImpl) AddBalance(evmAddr common.Address, amt *big.Int, reason tracing.BalanceChangeReason) {
 	s.k.PrepareReplayedAddr(s.ctx, evmAddr)
+	s.k.InitAccount(s.ctx, evmAddr)
+
 	if amt.Sign() == 0 {
 		return
 	}
--- a/x/evm/state/expected_keepers.go
+++ b/x/evm/state/expected_keepers.go
@@ -14,6 +14,7 @@ type EVMKeeper interface {
 	GetSeiAddressOrDefault(ctx sdk.Context, evmAddress common.Address) sdk.AccAddress
 	BankKeeper() bankkeeper.Keeper
 	GetBaseDenom(sdk.Context) string
+	InitAccount(sdk.Context, common.Address)
 	DeleteAddressMapping(sdk.Context, sdk.AccAddress, common.Address)
 	GetCode(sdk.Context, common.Address) []byte
 	SetCode(sdk.Context, common.Address, []byte)


x/evm/keeper/address.go
```diff
@@ -27,12 +27,19 @@ func (k *Keeper) DeleteAddressMapping(ctx sdk.Context, seiAddress sdk.AccAddress
...
+	}
+}
+
 func (k *Keeper) GetEVMAddress(ctx sdk.Context, seiAddress sdk.AccAddress) (common.Address, bool) {
 	store := ctx.KVStore(k.storeKey)
 	bz := store.Get(types.SeiAddressToEVMAddressKey(seiAddress))
 	addr := common.Address{}
 	if bz == nil {
-		return addr, false
+		return common.BytesToAddress(seiAddress), true
```
codchen: nit: we don't need that boolean anymore

codchen: removing it will help us find all places where we act on that boolean (e.g. precompiles that throw error if this boolean is false)

aclmapping/evm/mappings.go
```diff
@@ -33,17 +33,13 @@ func TransactionDependencyGenerator(_ aclkeeper.Keeper, evmKeeper evmkeeper.Keep
...
-		return []sdkacltypes.AccessOperation{*acltypes.CommitAccessOp()}, nil
-	}
 
 	if err := ante.Preprocess(ctx, evmMsg); err != nil {
 		return []sdkacltypes.AccessOperation{}, err
 	}
 	ops := []sdkacltypes.AccessOperation{}
 	ops = appendRWBalanceOps(ops, state.GetCoinbaseAddress(ctx.TxIndex()))
-	sender := evmMsg.Derived.SenderSeiAddr
+	sender := evmKeeper.GetSeiAddress(ctx, evmMsg.Derived.SenderEVMAddr)
```
udpatil, line 42: we should be able to remove this ACLMapping behavior given that parallelv1 is no longer used, its either OCC or synchronous

docker/localnode/scripts/step2_genesis.sh
```diff
@@ -46,6 +46,9 @@ override_genesis '.app_state["gov"]["tally_params"]["threshold"]="0.5"'
 override_genesis '.app_state["gov"]["tally_params"]["expedited_quorum"]="0.9"'
 override_genesis '.app_state["gov"]["tally_params"]["expedited_threshold"]="0.9"'
 
+# evm config
+override_genesis '.app_state["evm"]["address_associations"]=[{"sei_address":"sei1m9qugvk4h66p6hunfajfg96ysc48zeq4m0d82c","eth_address":"0xF87A299e6bC7bEba58dbBe5a5Aa21d49bCD16D52"},{"sei_address":"sei1cjzphr67dug28rw9ueewrqllmxlqe5f0awulvy","eth_address":"0x70997970C51812dc3A010C7d01b50e0d17dc79C8"}]'
```
udpatil, line 50: are these older style of associated addresses for testing coverage?

x/evm/client/cli/native_tx.go
```diff
@@ -56,6 +58,54 @@ func NativeSendTxCmd() *cobra.Command {
 	return cmd
 }
 
+func SendCastSeiAddrTxCmd() *cobra.Command {
```
udpatil, line 61: this should basically be a no-op right? as in, is this intended solely for testing a self-send? if so how is this different from doing `seid tx evm send <cast-addr> <amount> --from admin`?

evmrpc/simulate.go
```diff
@@ -295,16 +295,11 @@ func (b *Backend) StateAtTransaction(ctx context.Context, block *ethtypes.Block,
...
-			seiAddr, associatedNow := b.keeper.GetSeiAddress(b.ctxProvider(LatestCtxHeight), msg.From)
-			if !associatedNow {
-				return nil, vm.BlockContext{}, nil, nil, fmt.Errorf("address %s is not associated in the latest height", msg.From.Hex())
-			}
-			if err := ante.NewEVMPreprocessDecorator(b.keeper, b.keeper.AccountKeeper()).AssociateAddresses(statedb.Ctx(), seiAddr, msg.From, nil); err != nil {
+		// check if this transaction caused legacy address association and simulate if so
+		addressThen := b.keeper.GetSeiAddress(statedb.Ctx(), msg.From)
+		addressNow := b.keeper.GetSeiAddress(b.ctxProvider(LatestCtxHeight), msg.From)
+		if !addressThen.Equals(addressNow) {
+			if err := ante.NewEVMPreprocessDecorator(b.keeper, b.keeper.AccountKeeper()).AssociateAddresses(statedb.Ctx(), addressNow, msg.From, nil); err != nil {
```
udpatil, line 302: why are we associating the address with addressNow since now any association is implicit, right? I dont know that I fully understand this edge case, is it where you want to simulate a tx on an old block height prior to this change so previously it would have registered the old style associaton, but because NOW the logic is different, we associate the new correct address instead? If so, I still dont understand why we need to explicitly associate as opposed use the implicit casting?

udpatil, line 302: I don't fully understand this logic, is the intention to allow simulating txs on older blocks where a tx would have associated the old way? I'm not fully understanding the relationship here between oldAddress and newAddress and how its needed here.

x/evm/keeper/address.go
```diff
@@ -27,40 +27,31 @@ func (k *Keeper) DeleteAddressMapping(ctx sdk.Context, seiAddress sdk.AccAddress
...
-		return []byte{}, false
-	}
-	return bz, true
-}
-
-func (k *Keeper) GetSeiAddressOrDefault(ctx sdk.Context, evmAddress common.Address) sdk.AccAddress {
-	addr, ok := k.GetSeiAddress(ctx, evmAddress)
-	if ok {
-		return addr
+		return evmAddress.Bytes()
```
udpatil, line 52: do we not need to do something like `sdk.AccAddress(evmAddress.Bytes())`?

x/evm/keeper/address.go
```diff
@@ -27,40 +27,31 @@ func (k *Keeper) DeleteAddressMapping(ctx sdk.Context, seiAddress sdk.AccAddress
...
 	}
-	return sdk.AccAddress(evmAddress[:])
+	return bz
```
udpatil, line 54: same here?

docker/localnode/scripts/step2_genesis.sh
codchen: yeah

x/evm/client/cli/native_tx.go
codchen: say `admin`'s true sei address is `seiABC` and true evm address is `0xXYZ`: `seid tx evm send <cast-addr> <amount> --from admin` would be an EVM tx with an EVM sig that decodes to the true evm address `0xXYZ`. The direct cast address of it `sei(0xXYZ)` doesn't have any fund to send because funds are under `seiABC`. This new command is a cosmos tx that sends from `seiABC` to `sei(0xXYZ)`

evmrpc/simulate.go
codchen: this is for legacy association created before this change. `addressThen` is the sei address before THAT legacy association happened. What we are doing here is basically checking if this tx, back in time, caused a legacy association, and reproduce that if so.

I think the var name might be what's confusing: `addressNow` doesn't mean the current way of address association, rather it means the association of the address in the latest DB, and it could be a legacy one

x/evm/keeper/address.go
codchen: it's cast implicitly

evmrpc/simulate.go
codchen: see above. An example would be:
- tx T created a legacy association for account 0xXYZ at height 100
- latest height is 200
- when we trace T, we start with the state at height 99
- `addressThen` here would be the association of 0xXYZ at height 99, which is the direct cast because T hasn't materialized yet
- `addressNow` here would be the association at height 200, which would be the legacy association as it's set on height 100
- since those two are different, we know an address association is set on height 100 (and only a legacy association can cause the difference), so we simulate it here

aclmapping/evm/mappings.go
codchen: would these still be used by OCC to guesstimate potential conflicts?

udpatil: we've disabled that since it didn't show any realistic gains in high conflict scenarios and was only adding latency to happy paths:
https://github.com/sei-protocol/sei-cosmos/blob/8dd4adcf7553c056ee338dabb155bdf03821a99d/tasks/scheduler.go#L291

x/evm/client/cli/native_tx.go
udpatil: understood, ty for exaplanation

evmrpc/simulate.go
udpatil: ahh gotcha, yeah the `addressNow` name was confusing me

udpatil: ty for explanation

PR #1711, opened 2024-06-03
Add gov proposal v2 for native pointer

codchen:
Add a new gov proposal type to register pointer for native tokens by providing only metadata instead of a deployed contract address, to make sure all pointers are running the same version.

Some notes:
- all other existing gov types for pointer have been disabled (but not removed). We can remove them once we are sure there is no more inflight proposals
- right now this new pathway doesn't implement "static address" for pointers. This will be part the larger change for pointer address

unit test


0xskinnyg:
If we have set-denom-metadata on the Cosmos side why can't we have a similar command for the EVM pointer instead of having to submit a proposal? 

This is hard for small projects as you have to deposit 3,500 SEI and have around 2bn votes for the proposal to go through.

From my personal experience, I've been trying to get help to make my proposal go through for the last 2 weeks and it has been unsuccessful so far.

PR #1712, opened 2024-06-04
Set receipt for CW->EVM calls

codchen:
Create receipt for CosmWasm transactions that invoked EVM contracts. Tx hash for such receipt is set to be the Cosmos tx hash of the originating tx, which happens to also be a 20-byte array same as EVM tx hash.

unit test
integration tests

### DIFF
--- a/contracts/test/CW20toERC20PointerTest.js
+++ b/contracts/test/CW20toERC20PointerTest.js
@@ -100,7 +100,7 @@ describe("CW20 to ERC20 Pointer", function () {
 
                     const res = await executeWasm(pointer,  { transfer: { recipient: accounts[1].seiAddress, amount: "100" } });
                     const txHash = res["txhash"];
-                    const receipt = await ethers.provider.getTransactionReceipt("0x" + txHash);
+                    const receipt = await ethers.provider.getTransactionReceipt(`0x${txHash}`); 
                     expect(receipt).not.to.be.null;
                     const filter = {
                         fromBlock: receipt["blockNumber"],
@@ -109,7 +109,8 @@ describe("CW20 to ERC20 Pointer", function () {
                         topics: [ethers.id("Transfer(address,address,uint256)")]
                     };
                     const logs = await ethers.provider.getLogs(filter);
-                    expect(logs.length).to.equal(1)
+                    expect(logs.length).to.equal(1);
+                    expect(logs[0]["topics"][0]).to.equal(ethers.id("Transfer(address,address,uint256)"));
                     const respAfter = await queryWasm(pointer, "balance", {address: accounts[1].seiAddress});
                     const balanceAfter = respAfter.data.balance;
 
--- a/contracts/test/CW20toERC20PointerTest.js
+++ b/contracts/test/CW20toERC20PointerTest.js
@@ -99,7 +99,9 @@ describe("CW20 to ERC20 Pointer", function () {
                     const balanceBefore = respBefore.data.balance;
 
                     const res = await executeWasm(pointer,  { transfer: { recipient: accounts[1].seiAddress, amount: "100" } });
-                    console.log(res);
+                    const txHash = res["txhash"];
+                    const receipt = await ethers.provider.getTransactionReceipt("0x" + txHash);
+                    expect(receipt).not.to.be.null;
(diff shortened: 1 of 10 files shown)


contracts/test/CW20toERC20PointerTest.js
```diff
@@ -98,7 +98,18 @@ describe("CW20 to ERC20 Pointer", function () {
...
+                    const receipt = await ethers.provider.getTransactionReceipt("0x" + txHash);
+                    expect(receipt).not.to.be.null;
+                    const filter = {
+                        fromBlock: receipt["blockNumber"],
+                        toBlock: 'latest',
+                        address: receipt["to"],
+                        topics: [ethers.id("Transfer(address,address,uint256)")]
+                    };
+                    const logs = await ethers.provider.getLogs(filter);
+                    expect(logs.length).to.equal(1)
```
stevenlanders: it may be good to assert the various fields here

evmrpc/tx.go
```diff
@@ -240,13 +245,15 @@ func getEthTxForTxBz(tx tmtypes.Tx, decoder sdk.TxDecoder) *ethtypes.Transaction
 
 // Gets the EVM tx index based on the tx index (typically from receipt.TransactionIndex
 // Essentially loops through and calculates the index if we ignore cosmos txs
-func GetEvmTxIndex(txs tmtypes.Txs, txIndex uint32, decoder sdk.TxDecoder) (index int, found bool) {
+func GetEvmTxIndex(txs tmtypes.Txs, txIndex uint32, decoder sdk.TxDecoder, receiptChecker func(common.Hash) bool) (index int, found bool) {
```
stevenlanders, line 248: It's existing code, but I feel like this might be cleaner here
```
func GetEvmTxIndex(txs tmtypes.Txs, txIndex uint32, decoder sdk.TxDecoder, receiptChecker func(common.Hash) bool) (index int, found bool) {
	var evmTxIndex int
	for i, tx := range txs {
		etx := getEthTxForTxBz(tx, decoder)
		
		// does not exist and has no receipt (cosmos)
		if etx == nil && !receiptChecker(sha256.Sum256(tx)) {
			continue
		}
		
		// found the index
		if i == int(txIndex) {
			return evmTxIndex, true
		}
		
		evmTxIndex++
	}
	return -1, false
}
```

stevenlanders: Just confirming, we DO want to increment the evmTxIndex on a fake receipt, right?

x/evm/keeper/evm.go
```diff
@@ -114,7 +114,17 @@ func (k *Keeper) CallEVM(ctx sdk.Context, from common.Address, to *common.Addres
 	if err != nil {
 		return nil, err
 	}
-	k.AppendToEvmTxDeferredInfo(ctx, ethtypes.Bloom{}, ethtypes.EmptyTxsHash, surplus)
+	vmErr := ""
+	if res.Err != nil {
+		vmErr = res.Err.Error()
+	}
+	receipt, err := k.WriteReceipt(ctx, stateDB, evmMsg, 0, ctx.TxSum(), res.UsedGas, vmErr)
```
stevenlanders: (can we use the proto enum value instead of `0`)

contracts/test/CW20toERC20PointerTest.js
```diff
@@ -98,7 +98,18 @@ describe("CW20 to ERC20 Pointer", function () {
                     const respBefore = await queryWasm(pointer, "balance", {address: accounts[1].seiAddress});
                     const balanceBefore = respBefore.data.balance;
 
-                    await executeWasm(pointer,  { transfer: { recipient: accounts[1].seiAddress, amount: "100" } });
+                    const res = await executeWasm(pointer,  { transfer: { recipient: accounts[1].seiAddress, amount: "100" } });
+                    const txHash = res["txhash"];
+                    const receipt = await ethers.provider.getTransactionReceipt("0x" + txHash);
```
stevenlanders: nit, in ES6 there's a cool format syntax with tick marks:
```
const receipt = await ethers.provider.getTransactionReceipt(`0x${txHash}`); 
```

stevenlanders (approved):
lgtm - some nits but makes sense to me

evmrpc/tx.go
codchen: yeah we do

PR #1715, opened 2024-06-05
Fix for swagger docs serving

dssei:
Swagger documentation should automatically be registered if `api.swagger` setting in `app.toml` is enabled
```toml
swagger = true
```
It was broken though, as this configuration was ignored in the app. Simply adding route would not work either. As just enabling it would pull swagger that is generated in `sei-cosmos` only lacking some modules APIs.
Below solution adds similar capability to `sei-chain` level and is tweaked to involve `ignite` generated swagger API docs. 

This change:
- adds swagger route (if configured)
- updates the path openapi/swagger yml is generated
- adds swagger-ui template
- introduces similar way of serving docs (with `statik` lib) to that of cosmos sdk (note, that we are not reusing the same scripts though, as in this repo we use `ignite` for openapi/swagger docs generation.
- updates README

- unit test for adding swagger route
- deployed and tested on local node


app/app_test.go
```diff
@@ -405,3 +411,70 @@ func TestDecodeTransactionsConcurrently(t *testing.T) {
 	require.Nil(t, typedTxs[1])
 	require.NotNil(t, typedTxs[2])
 }
+
+func TestApp_RegisterAPIRoutes(t *testing.T) {
```
yzang2019, line 415: Nice! Thanks for adding the test

scripts/update-swagger-ui-statik.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
yzang2019, line 1: When should this script be used? Do you mind adding some instructions?

yzang2019 (approved):

scripts/update-swagger-ui-statik.sh
dssei: Anytime when we modify endpoints/related protocol buffer. I will make it more explicit in readme

dssei: Done

PR #1728, opened 2024-06-11
Make occ caused evm panics less noisy

jewei1997:

If there is an evm panic due to occ conflict, don't log or increment metrics for this evm panic

existing tests


x/evm/keeper/msg_server.go
```diff
@@ -62,10 +64,11 @@ func (server msgServer) EVMTransaction(goCtx context.Context, msg *types.MsgEVMT
 
 	defer func() {
 		if pe := recover(); pe != nil {
-			// there is not supposed to be any panic
 			debug.PrintStack()
```
codchen: let's put this under the `if` below as well

codchen (approved):

PR #1729, opened 2024-06-11
Remove interoperability hop limit with isolated EVM instances

codchen:
Previously interop hop was limited to 1 due to concerns of reentrancy attacks. This PR removes the limit and now relies on the fact that each hop would instantiate a new, isolated EVM instance as base of operation. This creates the need of bubbling up EVM events generated  in those isolated EVM instances, which is solved by a dedicated event manager in context.

integration test that calls the wasmd precompile which calls a CW->ERC pointer

### DIFF
--- a/contracts/test/CW20toERC20PointerTest.js
+++ b/contracts/test/CW20toERC20PointerTest.js
@@ -1,6 +1,6 @@
 const {getAdmin, queryWasm, executeWasm, associateWasm, deployEvmContract, setupSigners, deployErc20PointerForCw20, deployWasm, WASM,
     registerPointerForERC20,
-    proposeCW20toERC20Upgrade
+    proposeCW20toERC20Upgrade, callWasmViaPrecompile
 } = require("./lib");
 const { expect } = require("chai");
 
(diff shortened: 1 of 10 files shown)


contracts/test/CW20toERC20PointerTest.js
```diff
@@ -152,6 +152,27 @@ describe("CW20 to ERC20 Pointer", function () {
...
+
+                    // make sure log is emitted
+                    console.log(receipt);
+                    const filter = {
+                        fromBlock: receipt["blockNumber"],
+                        toBlock: 'latest',
+                        topics: [ethers.id("Transfer(address,address,uint256)")]
+                    };
+                    const logs = await ethers.provider.getLogs(filter);
+                    expect(logs.length).to.equal(1)
```
stevenlanders, line 174: is it possible to string this test together to see a the set of logs unified from both evm invocations?(>1)

precompiles/wasmd/wasmd.go
```diff
@@ -373,11 +374,15 @@ func (p Precompile) executeBatch(ctx sdk.Context, method *abi.Method, caller com
 			return
 		}
 
+		ctx = ctx.WithEvmEventManager(sdk.NewEVMEventManager())
 		res, err := p.wasmdKeeper.Execute(ctx, contractAddr, senderAddr, msg, coins)
 		if err != nil {
 			rerr = err
 			return
 		}
+		for _, e := range ctx.EVMEventManager().Events() {
```
stevenlanders: at some point I'd like to refactor this stuff into some kind of 'post handler' and 'pre handler' that is shared across all the scenarios so we can avoid the duplication.

contracts/test/CW20toERC20PointerTest.js
codchen: it's a little difficult since the wasmd precompile doesn't emit event, and we don't allow pointer pointing to another pointer

PR #1734, opened 2024-06-17
Add migration to unwind dex state

udpatil:
This disables many of the dex message types, and also unregisters existing contracts while handling module balances by refunding rents + handling excess balances.

Add unit tests


Kbhat1 (approved):

x/dex/migrations/v17_to_v18.go
```diff
@@ -0,0 +1,33 @@
...
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	"github.com/sei-protocol/sei-chain/x/dex/keeper"
+	dextypes "github.com/sei-protocol/sei-chain/x/dex/types"
+)
+
+func V17ToV18(ctx sdk.Context, dexkeeper keeper.Keeper) error {
+	// iterate over all contracts and unregister them
+	for _, c := range dexkeeper.GetAllContractInfo(ctx) {
+		if err := dexkeeper.DoUnregisterContractWithRefund(ctx, c); err != nil {
+			return err
```
codchen, line 14: `DoUnregisterContractWithRefund` may return error if module account has less balance than what dexkeeper thinks it has, in which case the upgrade would fail. I know we checked beforehand that the balance matches but just to be safe we can probably just log the error here without returning the error

codchen (approved):
we should also no-op this https://github.com/sei-protocol/sei-chain/blob/main/x/dex/gov.go#L9-L14

PR #1740, opened 2024-06-20
Update Random to Hash of Block Timestamp

Kbhat1:
- Replace `Random` field with hash of block timestamp

### DIFF
--- a/x/evm/keeper/keeper.go
+++ b/x/evm/keeper/keeper.go
@@ -9,7 +9,6 @@ import (
 	"math/big"
 	"slices"
 	"sort"
-	"strconv"
 	"sync"
 
 	wasmkeeper "github.com/CosmWasm/wasmd/x/wasm/keeper"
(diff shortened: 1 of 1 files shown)


x/evm/keeper/keeper.go
```diff
@@ -189,11 +191,9 @@ func (k *Keeper) GetVMBlockContext(ctx sdk.Context, gp core.GasPool) (*vm.BlockC
 	if err != nil {
 		return nil, err
 	}
-	r, err := ctx.BlockHeader().Time.MarshalBinary()
-	if err != nil {
-		return nil, err
-	}
-	rh := common.BytesToHash(r)
+
+	timestamp := ctx.BlockHeader().Time.Unix()
```
jewei1997: nit: can we take the hash of the exact block timestamp of this block?

jewei1997 (approved):
one suggestion but overall LGTM

x/evm/keeper/keeper.go
Kbhat1: Updated

codchen (approved):

PR #1748, opened 2024-06-27
Translate CW20 events into EVM events in shell receipts

codchen:
Add a post-DeliverTx hook that captures any relevant CW20 event and translate it into a corresponding ERC20 event. If a receipt already exists for that transaction (e.g. an EVM tx that called a CW20 contract via precompile), the translated event will be appended to the existing receipt and the bloom will be updated. Otherwise, a shell receipt will be created.

unit tests


x/evm/artifacts/README
```diff
@@ -1,5 +1,5 @@
 The source files are under contracts/src. The artifacts should be updated whenever the source files change. To update run the following (with NativeSeiTokensERC20 as an example):
-- `solc --overwrite @openzeppelin=contracts/lib/openzeppelin-contracts --bin -o x/evm/artifacts/cw721 contracts/src/CW721ERC721Pointer.sol`
+- `solc --overwrite @openzeppelin=contracts/lib/openzeppelin-contracts --bin -o x/evm/artifacts/cw20 contracts/src/CW721ERC721Pointer.sol`
```
stevenlanders: should the contract also be updated to CW20ERC20Pointer.sol?

x/evm/keeper/deferred.go
```diff
@@ -63,3 +63,15 @@ func (k *Keeper) AppendToEvmTxDeferredInfo(ctx sdk.Context, bloom ethtypes.Bloom
 	}
 	prefix.NewStore(ctx.TransientStore(k.transientStoreKey), types.DeferredInfoPrefix).Set(key, bz)
 }
+
+func (k *Keeper) GetEVMTxDeferredInfo(ctx sdk.Context) *types.DeferredInfo {
```
stevenlanders: side thought: It can sometimes be nice sdk-wise to have signatures that return `*pointer, bool` where the bool represents `found`

app/receipt.go
```diff
@@ -0,0 +1,177 @@
...
+	wasmEvents := GetEventsOfType(response, wasmtypes.WasmModuleEventType)
+	logs := []*ethtypes.Log{}
+	for _, wasmEvent := range wasmEvents {
+		contractAddr, found := GetAttributeValue(wasmEvent, wasmtypes.AttributeKeyContractAddr)
+		if !found {
+			continue
+		}
+		// check if there is a ERC20 pointer to contractAddr
+		pointerAddr, _, exists := app.EvmKeeper.GetERC20CW20Pointer(ctx, contractAddr)
+		if exists {
```
stevenlanders, line 44: i'd suggest doing a `if !exists { continue }` to avoid the indentation

app/receipt.go
```diff
@@ -0,0 +1,177 @@
...
+
+func GetAmountAttribute(event abci.Event) *big.Int {
+	amount, found := GetAttributeValue(event, "amount")
+	if found {
+		amountInt, ok := sdk.NewIntFromString(amount)
+		if ok {
+			return amountInt.BigInt()
+		}
+	}
+	return big.NewInt(0)
```
stevenlanders: is it true that the default amount should be zero (vs. nil/not-existing?).   I'd also suggest returning early on !found  and !ok to keep the indentations down

stevenlanders: (I like the `GetAttributeValue` approach of having the boolean to indicate found)

app/receipt.go
```diff
@@ -0,0 +1,177 @@
...
+		app.EvmKeeper.AppendToEvmTxDeferredInfo(ctx, bloom, txHash, sdk.ZeroInt())
+	}
+}
+
+func (app *App) GetEvmAddressAttribute(ctx sdk.Context, event abci.Event, attribute string) common.Hash {
+	addrStr, found := GetAttributeValue(event, attribute)
+	if found {
+		seiAddr, err := sdk.AccAddressFromBech32(addrStr)
+		if err == nil {
+			evmAddr := app.EvmKeeper.GetEVMAddressOrDefault(ctx, seiAddr)
```
stevenlanders, line 154: I believe it will, but just confirming:  this would work on a past block as well (ctx's height is honored here)?

app/receipt.go
```diff
@@ -0,0 +1,177 @@
...
+					app.GetEvmAddressAttribute(ctx, wasmEvent, "owner"),
+					app.GetEvmAddressAttribute(ctx, wasmEvent, "spender"),
+				}
+				res, err := app.WasmKeeper.QuerySmart(
+					ctx,
+					sdk.MustAccAddressFromBech32(contractAddr),
+					[]byte(fmt.Sprintf("{\"allowance\":{\"owner\":\"%s\",\"spender\":\"%s\"}}", ownerStr, spenderStr)),
+				)
+				if err != nil {
+					continue
```
stevenlanders: is this err something that means a safe 'was not found' situation (or is it worse than that?)  I wonder if it should log?

app/receipt.go
```diff
@@ -0,0 +1,177 @@
...
+				}
+				allowanceResponse := &AllowanceResponse{}
+				if err := json.Unmarshal(res, allowanceResponse); err != nil {
+					continue
```
stevenlanders: I think we'd want a log for this one, or maybe it's a worse situation and should bail/panic?

contracts/test/ERC20toCW20PointerTest.js
```diff
@@ -93,12 +93,23 @@ describe("ERC20 to CW20 Pointer", function () {
...
 
+                    // check logs
+                    const filter = {
+                        fromBlock: blockNumber,
+                        toBlock: 'latest',
+                        address: await pointer.getAddress(),
+                        topics: [ethers.id("Transfer(address,address,uint256)")]
+                    };
+                    const logs = await ethers.provider.getLogs(filter);
+                    expect(logs.length).to.equal(1);
```
stevenlanders, line 111: it may be good just to assert the expected properties of that log (just to make sure the addresses are right, and the amounts are in the right unit)

app/receipt.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	var bloom ethtypes.Bloom
+	if r, err := app.EvmKeeper.GetTransientReceipt(ctx, txHash); err == nil && r != nil {
+		r.Logs = append(r.Logs, utils.Map(logs, evmkeeper.ConvertEthLog)...)
+		bloom = ethtypes.CreateBloom(ethtypes.Receipts{&ethtypes.Receipt{Logs: evmkeeper.GetLogsForTx(r)}})
+		r.LogsBloom = bloom[:]
+		_ = app.EvmKeeper.SetTransientReceipt(ctx, txHash, r)
+	} else {
+		bloom = ethtypes.CreateBloom(ethtypes.Receipts{&ethtypes.Receipt{Logs: logs}})
+		receipt := &evmtypes.Receipt{
```
stevenlanders, line 68: is there some indicator or flag so that we can somehow optionally include/exclude these events from a result?

stevenlanders (approved):
Had some question - looks good to me.

app/receipt.go
codchen: we can't continue here because we'd have another branch for 721 https://github.com/sei-protocol/sei-chain/pull/1750/files#diff-8810d2793d992ba6de31afd91265ef3facc0612669de567b51dc1d8d029cf22fR104 I refactored this part into a func to avoid the big indentation instead

app/receipt.go
codchen: updated

app/receipt.go
codchen: yeah all the states are accessed via ctx which will be versioned

app/receipt.go
codchen: we probably don't want a log here since it simply indicates that the contract does not adhere to cw20 (which is bad from the contract's perspective, but kind of neutral from chain's perspective)

app/receipt.go
codchen: ditto

PR #1753, opened 2024-07-02
Add Gas Overflow Check

Kbhat1:
- Add int64 Gas Overflow Check for gas wanted

- Verifying on local chain
- Added unit tests


Kbhat1:
NOTE: We probably want to revisit why the consensus params `MaxGas` is a `int64` but each tx's gas is a `uint64`

codchen (approved):

PR #1767, opened 2024-07-12
Add address association precompile method

mj850:
Sample code for a POC of a new addr precompile method to associate addresses.
The goal is to allow someone to associate addresses on behalf of another user by paying gas (instead of checking that the other account has funds)

The idea was initially for a new EVM RPC call, but that would entail creating a new message type and modifying the Antehandlers to have custom logic for this new message (since it would be the first custom EVM message that is signed)

Added Unit tests
Manually validated by calling precompile locally


precompiles/addr/addr.go
```diff
@@ -75,6 +91,8 @@ func (p PrecompileExecutor) Execute(ctx sdk.Context, method *abi.Method, _ commo
 		return p.getSeiAddr(ctx, method, args, value)
 	case GetEvmAddressMethod:
 		return p.getEvmAddr(ctx, method, args, value)
+	case AssociateWithGas:
+		return p.associateWithGas(ctx, method, args, value)
```
codchen: this would be a quite expensive call so we'd want to modify the gas cost for it in `RequiredGas()` function

mj850: Don't have a good gauge on this so I set it to 50000 for now, does that make sense?

codchen: yeah that should be enough

codchen (approved):

package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
dssei: Curious this file is needed?

precompiles/addr/addr_test.go
```diff
@@ -0,0 +1,215 @@
...
+				msg:    prefixedMessage,
+				value:  big.NewInt(0),
+			},
+			wantRet:    happyPathOutput,
+			wantErr:    false,
+			wantErrMsg: "",
+			wrongRet:   true,
+		},
+		{
+			name: "associates wrong address if invalid signature (different message)",
```
dssei, line 129: Curious about this test case. Do we expect association if signature is invalid?

precompiles/addr/addr_test.go
```diff
@@ -0,0 +1,215 @@
...
+				msg:    "Not the signed message",
+				value:  big.NewInt(0),
+			},
+			wantRet:    happyPathOutput,
+			wantErr:    false,
+			wantErrMsg: "",
+			wrongRet:   true,
+		},
+		{
+			name: "happy path",
```
dssei: might be helpful if we elaborate on test case here. e.g. `associates address if signature is correct`

precompiles/addr/addr_test.go
```diff
@@ -0,0 +1,215 @@
...
+					TxContext: vm.TxContext{Origin: callerEvmAddress},
+				},
+				caller: callerEvmAddress,
+				v:      callerV,
+				r:      callerR,
+				s:      callerS,
+				msg:    prefixedMessage,
+				value:  big.NewInt(0),
+			},
+			wantRet:    nil,
```
dssei: nit: on negative test cases, like this one we could skip adding field

precompiles/addr/addr_test.go
```diff
@@ -0,0 +1,215 @@
...
+				},
+				caller: callerEvmAddress,
+				v:      v,
+				r:      r,
+				s:      s,
+				msg:    prefixedMessage,
+				value:  big.NewInt(10),
+			},
+			wantRet:    nil,
+			wantErr:    true,
```
dssei, line 91: Nit: we could probably just have
```		
wantErr:    true,
wantErrMsg: "sending funds to a non-payable function",
```

precompiles/addr/addr.go
```diff
@@ -109,6 +136,74 @@ func (p PrecompileExecutor) getEvmAddr(ctx sdk.Context, method *abi.Method, args
...
-	return false
+func (p PrecompileExecutor) associateWithGas(ctx sdk.Context, method *abi.Method, args []interface{}, value *big.Int) ([]byte, error) {
+	if err := pcommon.ValidateNonPayable(value); err != nil {
+		return nil, err
+	}
+
+	if err := pcommon.ValidateArgsLength(args, 4); err != nil {
+		return nil, err
+	}
+	v := args[0].(string)
```
dssei, line 145: Nit: adding brief explanation about v,r,s could be helpful

precompiles/addr/Addr.sol
```diff
@@ -8,7 +8,20 @@ IAddr constant ADDR_CONTRACT = IAddr(
 );
 
 interface IAddr {
+    // Transactions
+    function associateWithGas(
```
besated: maybe we can drop the `withGas` since it is implied this tx requires gas to send?

precompiles/addr/addr_test.go
mj850: Yup in this case we try to derive a pubkey from the signature values and signed message (which seems to be always possible), it'll just give us the wrong pubkey.

precompiles/addr/addr_test.go
mj850: Added a fuller description behind 'happy path'

precompiles/addr/addr.go
mj850: Added a small comment about v,r,s values here

precompiles/addr/Addr.sol
mj850: Makes sense! Renamed to 'associate'

PR #1768, opened 2024-07-16
Distribution rewards query

dssei:
Adding `rewards` query to a distribution precompile
 
- unit tests
- functional tests


precompiles/distribution/distribution.go
```diff
@@ -246,3 +259,89 @@ func (p PrecompileExecutor) accAddressFromArg(ctx sdk.Context, arg interface{})
...
+}
+
+type Rewards struct {
+	Rewards []Reward
+	Total   []Coin
+}
+
+func (p PrecompileExecutor) rewards(ctx sdk.Context, method *abi.Method, args []interface{}) (ret []byte, remainingGas uint64, rerr error) {
+	defer func() {
+		if err := recover(); err != nil {
```
codchen, line 281: do we expect any panic in the calls below? If not we should probably add a log here as well

codchen (approved):

PR #1775, opened 2024-07-18
Disallow sending to direct cast address after association

codchen:
We don't want to allow users to send funds to a temporary address that's created through direct casting of an EVM address after that EVM address has been associated with its true Sei address counterpart. This PR in conjunction with [this sei-cosmos PR](https://github.com/sei-protocol/sei-cosmos/pull/526) introduces such a check.

unit test


stevenlanders:
This one will need a hardhat test (see the association test library).  They already do the association, so this would just be asserting that the subsequent transaction fails.

stevenlanders (approved):

PR #1785, opened 2024-07-26
Allow CW->ERC pointers to be called through wasmd precompile

codchen:
For EVM transactions that call the wasmd precompile (`0x1002`) directly, we want to allow the destination CW contract to be call back to some EVM contracts (e.g. ERC20) at most once (right now none is allowed). This PR leverages a new flag set on `sdk.Context` to track whether a CW->EVM call is eligible for this special case. Note that this only applies to the top level - something like precompile->CW->precompile->CW is still not allowed.

hardhat test


x/evm/keeper/evm.go
```diff
@@ -118,6 +118,23 @@ func (k *Keeper) CallEVM(ctx sdk.Context, from common.Address, to *common.Addres
 	if res.Err != nil {
 		vmErr = res.Err.Error()
 	}
+	existingReceipt, err := k.GetTransientReceipt(ctx, ctx.TxSum())
```
jewei1997, line 121: is this logic related to the wasmd precompile stuff?

jewei1997 (approved):
overall LGTM

x/evm/keeper/evm.go
codchen: yeah it's to merge with the temp receipts created in the CW->EVM hop

ARitz-Cracker:
Apologies for being late to this. I received no communication that this was the implemented solution until now.

What exactly does this protect against? I can deploy an "ERC-20 token" which can still embed a nasty payload if the "transfer" function was called with specific sender and use the address and amount params as commands for the malicious code.

This solution needlessly impedes the ability for CW contracts to call the ERC-20 tokens directly without the use of a pointer with no security benefit.

PR #1794, opened 2024-07-29
REST route endpoint standardization

codebycarson:
Currently, there is in-congruency in the Sei REST API endpoint’s routes for all the custom sei-protocol modules.

Typically REST endpoints have the format /{org}/{module}/{version}/*, for example “/cosmwasm/wasm/v1/*” or “/cosmos/feegrant/v1beta1/*”.

IBC routes have the format /{org}/{package}/{module}/{version}/*, such as IBC “/ibc/apps/transfer/v1/*”.

Some of the Sei routes use “sei-chain” as the package, while others use “seichain”.
Some have “sei-protocol” as the org prefix, while others do not have an org prefix.
Only mint contains versions, while the others do not.

This PR standardizes the REST routes and proto package names to be /org/module/version/** with "sei" as the org.

Example:

Old route:
http://localhost:1317/sei-protocol/sei-chain/oracle/slash_window

New route:
http://localhost:1317/sei/oracle/v1/slash_window

It is important to note that for backwards compatibility the old routes have been kept in this version. Once merged, docs and integration guides will be updated to use the newer routes formats.

Tested locally on all custom module endpoints by two separate people


cordt-sei:
So this adds additional routes while keeping existing ones for compatibility?

PR #1796, opened 2024-07-30
fix the dex module deletion logic

udpatil:
The unregistration of the dex store key in KV mounting resulted it in not being mounted for deletion at the upgrade, so this fixes that by mounting the KV store key.

Tested on local chain, will do testing with loadtest clusters


philipsu522 (approved):

ARitz-Cracker:
👀

yzang2019 (approved):

PR #1798, opened 2024-07-31
Add associate with pubkey precompile method

mj850:
Coinbase requested a way to associate addresses via precompile by supplying just the public keys.
This might be useful for wallets that want to pre associate any addresses

Adds a new method to addr precompile to associate using pubkey
Added unit tests, hardhat test, and manually tested on local chain


precompiles/addr/Addr.sol
```diff
@@ -16,6 +16,10 @@ interface IAddr {
         string memory customMessage
     ) external returns (string seiAddr, address evmAddr);
 
+    function associatePubKey(
+        string memory pubKeyHex
```
besated, line 21: nit: it would be good to add some documentation to clarify if this is a compressed or uncompressed pubkey

codchen (approved):

philipsu522 (approved):

PR #1800, opened 2024-08-01
[EVM] Tune configs

jewei1997:
Increase gas price based on gas price investigation.

unit tests. Also performed manual eth transfer on local chain and verified balance change is expected.


codchen (approved):
we'd still need to send a gov proposal to update it for arctic/atlantic/pacific (this PR changes the default which only affects new chains)

zenoUsman:
> we'd still need to send a gov proposal to update it for arctic/atlantic/pacific (this PR changes the default which only affects new chains)
**This changes effect sei pacific chain too some of minting tokens and mint nfts rising fee errors**
354992810-44b7c24f-02df-4fde-9412-327cdde86e4d
Before that mint functions on sei native chain works perfectly needed to change in sei js libraries.

PR #1802, opened 2024-08-02
Add dApp Tests

mj850:
We want to add some dApp themed tests to our integration test pipeline to test real scenarios.

This PR (so far) adds a simple DEX example and one test against that as a POC for how this will work. More tests to be added to this, as well as to the dapp_tests directory in general.

Tests


integration_test/dapp_tests/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "uniswapv3-deployment",
```
besated: nit: can we change this name and description to make it more generic since it's for all the tests?

integration_test/dapp_tests/uniswap/uniswapTest.js
```diff
@@ -0,0 +1,253 @@
+const hre = require("hardhat"); // Require Hardhat Runtime Environment
```
besated, line 1: it might be worth updating this to use typescript, but we can do that in a separate pr 

for now this is fine

besated (approved):

integration_test/dapp_tests/uniswap/uniswapTest.js
mj850: Good call, can look into that after we're done with the test suite

PR #1806, opened 2024-08-07
Fix docker setup for local cluster

wojciech-sei:
- on linux files created inside `build/generated` are owned by root - fixed it by providing the current user to the docker compose
- due to that everything is now stored under `/home/ubuntu` rather than `/root`
- `rm -rf build/generated` command is moved from Dockerfile to `deploy.sh` because it should be run on every cluster creation (correct me if I'm wrong)
- `HOME` env variable does not exist in Dockerfile during `docker build` phase - defined it
- to build `seid` faster the local `GOCACHE` is mounted inside the containers

Tested on Linux, please test on Darwin

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -149,12 +149,6 @@ jobs:
         with:
           go-version: 1.21
 
-      - name: Install Docker Compose
-        run: |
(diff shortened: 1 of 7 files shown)


stevenlanders:
one thought: it may be good for us to move to `docker compose` instead of `docker-compose`  (I have to globally replace that on my linux machine at the moment)

wojciech-sei:
@stevenlanders okay but in another PR

udpatil (approved):
LGTM, works on mac os as well, thanks for the fix!

PR #1825, opened 2024-08-21
Return evm logs from cosmos interface

wojciech-sei:

Now logs from evm transaction are returned when tx is executed through cosmos interface

SEI-7924

Unit test is updated to verify that logs are not empty.


x/evm/keeper/evm.go
```diff
@@ -13,6 +13,7 @@ import (
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/core/vm"
 	"github.com/ethereum/go-ethereum/params"
+
```
udpatil, line 16: nit: whyd this line get added (i'm curious lol)

udpatil: import grouping with auto format?

x/evm/keeper/msg_server.go
```diff
@@ -176,21 +174,27 @@ func (server msgServer) EVMTransaction(goCtx context.Context, msg *types.MsgEVMT
 			},
 		)
 
-	} else {
-		// if applyErr is nil then res must be non-nil
-		if res.Err != nil {
-			serverRes.VmError = res.Err.Error()
+		return
```
udpatil, line 180: in this case, we still would want a hash in the tx response to be consistent with the previous implementation, right? in this version, it would now return `nil` for serverRes instead of a struct with the hash field set

x/evm/keeper/evm.go
wojciech-sei: yes

x/evm/keeper/msg_server.go
wojciech-sei: yea... `serverRes` is a named return. Fixed, thanks!

x/evm/keeper/msg_server.go
```diff
@@ -176,21 +177,23 @@ func (server msgServer) EVMTransaction(goCtx context.Context, msg *types.MsgEVMT
...
+	}
+	serverRes.GasUsed = res.UsedGas
+	serverRes.ReturnData = res.ReturnData
+	serverRes.Logs = types.NewLogsFromEth(stateDB.GetAllLogs())
```
udpatil: is this and the above assignments safe to call prior to the `res.Err` check given that its set afterwards in the prior impl?

wojciech-sei: Yes, because inside `if res.Err != nil {` there is no `return` so those assignments were made anyway. But moved that code back to the previous place because it doesn't matter for my fix.

udpatil (approved):

PR #1826, opened 2024-08-21
feat: limit MsgExec nested level to prevent excessive nesting

blindchaser:
- introduces a limit on the nested level of MsgExec to prevent potential attacks.


app/antedecorators/authz_nested_message.go
```diff
@@ -8,6 +8,9 @@ import (
 	evmtypes "github.com/sei-protocol/sei-chain/x/evm/types"
 )
 
+// maxNestedMsgs defines a cap for the number of nested messages on a MsgExec message
+const maxNestedMsgs = 7
```
udpatil: lets reduce this since even `7` seems excessive for a normal use case, maybe 5?

udpatil (approved):

app/antedecorators/authz_nested_message.go
blindchaser: sure

blindchaser: it looks the golangci-lint / lint failing due to new golangci-lint version update yesterday. I will cut a ticket to fix it. @udpatil

PR #1834, opened 2024-08-27
feat: add CLI and RPC endpoint from Pointee to Pointer mapping

blindchaser:
- support querying a given pointee to retrieve the corresponding pointer


codchen (commented):
overall lgtm. Let's add some tests in `grpc_query_test.go`

x/evm/keeper/grpc_query_test.go
```diff
@@ -47,3 +47,47 @@ func TestQueryPointer(t *testing.T) {
 	require.Nil(t, err)
 	require.Equal(t, types.QueryPointerResponse{Pointer: seiAddr5.String(), Version: uint32(erc721.CurrentVersion), Exists: true}, *res)
 }
+
+func TestQueryPointee(t *testing.T) {
```
philipsu522, line 52: can we also test for invalid / not registered yet?

philipsu522 (approved):

x/evm/keeper/grpc_query_test.go
blindchaser: good point will add it

PR #1840, opened 2024-08-30
Add migration for new params

blindchaser:

### DIFF
--- a/x/evm/migrations/migrate_deliver_tx_gas_limit.go
+++ b/x/evm/migrations/migrate_deliver_tx_gas_limit.go
@@ -1,0 +1,20 @@
+package migrations
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/sei-protocol/sei-chain/x/evm/keeper"
+	"github.com/sei-protocol/sei-chain/x/evm/types"
+)
+
+func MigrateDeliverTxHookWasmGasLimitParam(ctx sdk.Context, k *keeper.Keeper) error {
+	// Fetch the current parameters
+	keeperParams := k.GetParams(ctx)
+
+	// Update DeliverTxHookWasmGasLimit to the default value
+	keeperParams.DeliverTxHookWasmGasLimit = types.DefaultParams().DeliverTxHookWasmGasLimit
+
+	// Set the updated parameters back in the keeper
+	k.SetParams(ctx, keeperParams)
+
+	return nil
+}
--- a/x/evm/migrations/migrate_deliver_tx_gas_limit_test.go
+++ b/x/evm/migrations/migrate_deliver_tx_gas_limit_test.go
@@ -1,0 +1,37 @@
+package migrations_test
+
+import (
+	"testing"
+
+	testkeeper "github.com/sei-protocol/sei-chain/testutil/keeper"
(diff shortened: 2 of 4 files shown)


yzang2019 (approved):

x/evm/module.go
```diff
@@ -208,6 +208,10 @@ func (am AppModule) RegisterServices(cfg module.Configurator) {
 	_ = cfg.RegisterMigration(types.ModuleName, 10, func(ctx sdk.Context) error {
 		return migrations.MigrateCastAddressBalances(ctx, am.keeper)
 	})
+
+	_ = cfg.RegisterMigration(types.ModuleName, 11, func(ctx sdk.Context) error {
+		return migrations.AddNewParamsAndSetAllToDefaults(ctx, am.keeper)
```
udpatil: lets not use this migration handler because it will reset params which we don't want to do, instead we should initialize the param value for the new param being introduced. And lets add tests for that

udpatil (changes requested):

x/evm/module.go
yzang2019: Yeah, that's a valid concern, let's only change for the new param

x/evm/migrations/migrate_deliver_tx_gas_limit.go
```diff
@@ -0,0 +1,15 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/sei-protocol/sei-chain/x/evm/keeper"
+	"github.com/sei-protocol/sei-chain/x/evm/types"
+)
+
+func MigrateDeliverTxHookWasmGasLimitParam(ctx sdk.Context, k *keeper.Keeper) error {
+	defaultParams := types.DefaultParams()
+	defaultDeliverTxHookWasmGasLimit := defaultParams.DeliverTxHookWasmGasLimit
+	defaultParams.DeliverTxHookWasmGasLimit = defaultDeliverTxHookWasmGasLimit
+	k.SetParams(ctx, defaultParams)
```
udpatil: This will still override the other existing params, instead of starting from default params, we need to load the existing params, and set these two values in that struct

x/evm/migrations/migrate_deliver_tx_gas_limit_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+	err := migrations.MigrateDeliverTxHookWasmGasLimitParam(ctx, &k)
+
+	require.NoError(t, err)
+
+	require.NotPanics(t, func() { k.GetParams(ctx) })
+
+	keeperParams := k.GetParams(ctx)
+
+	require.Equal(t, keeperParams.GetDeliverTxHookWasmGasLimit(), types.DefaultParams().DeliverTxHookWasmGasLimit)
```
udpatil, line 31: For good measure we can start by setting some other value for a non new param and verify that it doesn't get overwritten

udpatil (changes requested):

x/evm/migrations/migrate_deliver_tx_gas_limit_test.go
blindchaser: yeah i relized that too. working on it

udpatil (approved):

yzang2019 (approved):

PR #1843, opened 2024-09-05
EIP-1559 - Dynamic Base Fee

jewei1997:

Implementation of a modified version of eip-1559. How it works is on every EndBlock, it will adjust the BaseFee based on the gas used in the current block. The target gas for the block is always half of the total block gas limit. There are 2 major ways our eip-1559 implementation diverges from Ethereum's:
1. We don't burn the base fee. We will treat the base fee the same as the validator tip, so both the base fee and the validator tip will go to the validator of the block
2. How much the base fee goes up or down is adjustable with the `MaxDynamicBaseFeeUpwardAdjustment` and `MaxDynamicBaseFeeDownwardAdjustment` parameters. We don't want to just blindly use the 12.5% used by Ethereum because has different block times and block gas limits.

Currently `MaxDynamicBaseFeeUpwardAdjustment` is set to 5% and `MaxDynamicBaseFeeDownwardAdjustment`  is set to 1%, which means that base fee will go up at most 5% on a completely full block and down 1% on a completely empty block and will go up/down proportionally for everything in between (if block gas used is exactly half full, base fee will remain constant)

unit tests
loadtest cluster under load and viewing changing gas prices


x/evm/keeper/evm.go
```diff
@@ -34,6 +34,7 @@ func (k *Keeper) HandleInternalEVMCall(ctx sdk.Context, req *types.MsgInternalEV
 	if err != nil {
 		return nil, err
 	}
+	k.GetEVMAddressOrDefault(ctx, senderAddr)
```
udpatil: why do we duplicate this effort with the below line?

jewei1997: I think this is a mistake--will remove

x/evm/keeper/evm.go
```diff
@@ -95,15 +95,15 @@ func (k *Keeper) CallEVM(ctx sdk.Context, from common.Address, to *common.Addres
 		Nonce:             stateDB.GetNonce(from), // replay attack is prevented by the AccountSequence number set on the CW transaction that triggered this call
 		GasLimit:          k.getEvmGasLimitFromCtx(ctx),
 		GasPrice:          utils.Big0, // fees are already paid on the CW transaction
-		GasFeeCap:         utils.Big0,
+		GasFeeCap:         big.NewInt(100000000000),
```
jewei1997: this should pull the latest base fee

x/evm/ante/fee.go
```diff
@@ -101,7 +101,7 @@ func (fc EVMFeeCheckDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate b
 
 // fee per gas to be burnt
```
codchen: nit: we can change this comment to be something like `minimum fee per tx` since we don't really burn

x/evm/keeper/evm.go
codchen: can we update this?

x/evm/keeper/evm.go
```diff
@@ -95,15 +95,15 @@ func (k *Keeper) CallEVM(ctx sdk.Context, from common.Address, to *common.Addres
...
 		GasTipCap:         utils.Big0,
 		To:                to,
 		Value:             value,
 		Data:              data,
 		SkipAccountChecks: false,
 		From:              from,
 	}
-	res, err := k.applyEVMMessage(ctx, evmMsg, stateDB, gp)
+	res, err := k.applyEVMMessageWithNoBaseFee(ctx, evmMsg, stateDB, gp)
```
codchen: what's the reason for no base fee here?

x/evm/keeper/fee.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	// Set the new base fee for the next height
+	k.SetDynamicBaseFeePerGas(ctx.WithBlockHeight(ctx.BlockHeight()+1), newBaseFee)
+}
+
+func (k *Keeper) GetDynamicBaseFeePerGas(ctx sdk.Context) sdk.Dec {
+	h := make([]byte, 8)
+	binary.BigEndian.PutUint64(h, uint64(ctx.BlockHeight()))
+	bz := k.PrefixStore(ctx, types.BaseFeePerGasPrefix).Get(h)
```
codchen: I don't think we need to have block height be part of the key here, since we can get historical values from historical versions. Right now this would cause:
version 100 stores values for height 100, height 99, ...
version 99 stores values for height 99, height 98, ...

We want:
version 100 stores value for height 100
version 99 stores value for height 99
...

x/evm/keeper/keeper.go
```diff
@@ -209,9 +209,9 @@ func (k *Keeper) GetVMBlockContext(ctx sdk.Context, gp core.GasPool) (*vm.BlockC
 		GasLimit:    gp.Gas(),
 		BlockNumber: big.NewInt(ctx.BlockHeight()),
 		Time:        uint64(ctx.BlockHeader().Time.Unix()),
-		Difficulty:  utils.Big0,                                     // only needed for PoW
-		BaseFee:     k.GetBaseFeePerGas(ctx).TruncateInt().BigInt(), // feemarket not enabled
-		BlobBaseFee: utils.Big1,                                     // Cancun not enabled
+		Difficulty:  utils.Big0,                                            // only needed for PoW
+		BaseFee:     k.GetDynamicBaseFeePerGas(ctx).TruncateInt().BigInt(), // feemarket not enabled
```
codchen: nit: remove comment

x/evm/migrations/migrate_eip_1559_params_test.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/stretchr/testify/require"
+	tmtypes "github.com/tendermint/tendermint/proto/tendermint/types"
+)
+
+func TestMigrateEip1559Params(t *testing.T) {
+	k := testkeeper.EVMTestApp.EvmKeeper
+	ctx := testkeeper.EVMTestApp.NewContext(false, tmtypes.Header{})
+
+	// Perform the migration
+	err := migrations.MigrateEip1559Params(ctx, &k)
```
codchen, line 22: can we modify some param before the migration and see if the modification is reverted afterwards?

x/evm/module.go
```diff
@@ -280,7 +284,11 @@ func (am AppModule) BeginBlock(ctx sdk.Context, _ abci.RequestBeginBlock) {
 
 // EndBlock executes all ABCI EndBlock logic respective to the evm module. It
 // returns no validator updates.
-func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.ValidatorUpdate {
+func (am AppModule) EndBlock(ctx sdk.Context, req abci.RequestEndBlock) []abci.ValidatorUpdate {
+	beforeBaseFeePerGas := am.keeper.GetDynamicBaseFeePerGas(ctx)
+	am.keeper.AdjustDynamicBaseFeePerGas(ctx, uint64(req.BlockGasUsed))
+	afterBaseFeePerGas := am.keeper.GetDynamicBaseFeePerGas(ctx.WithBlockHeight(ctx.BlockHeight() + 1))
+	fmt.Println("JEREMYDEBUG: EndBlock, height = ", req.Height, ", before base fee = ", beforeBaseFeePerGas, ", after base fee = ", afterBaseFeePerGas, ", block gas used = ", req.BlockGasUsed)
```
codchen: nit: remove or change do `logger().Debug` before merge

jewei1997: fixed

x/evm/migrations/migrate_eip_1559_params_test.go
jewei1997: the comment is wrong actually, should not be set back to defaults. Did set a param before the migration to check it works correctly.

x/evm/keeper/keeper.go
jewei1997: fixed

x/evm/keeper/fee.go
jewei1997: fixed

x/evm/keeper/evm.go
jewei1997: this code is outdated, fixed

x/evm/keeper/evm.go
jewei1997: fixed

x/evm/ante/fee.go
jewei1997: fixed

codchen (approved):

PR #1849, opened 2024-09-09
Avoid n^2 storage for block bloom

codchen:
Block bloom is a byte array that we store for every block that has any EVM transaction. This caused state bloat because we are always storing all historical blooms in each version. (i.e. version T will have all blooms from height 1 to height T, T-1 will have all blooms from 1 to T-1, etc.), so we end up storing n^2 blooms. This PR changes it so that each version only contains the current bloom (i.e. version T only stores bloom for height T, version T-1 only stores bloom for height T-1, etc.).

For backward compatibility reason, we set a cutoff state to help combining the queries for both the new storage and the old. The old blooms are deleted from the current version at the time of the migration, but not deleted from the previous versions (since it's impossible to do). Eventually they will be pruned as old versions of app state naturally get pruned over time.

unit test


x/evm/migrations/migrate_block_bloom.go
```diff
@@ -0,0 +1,25 @@
+package migrations
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/sei-protocol/sei-chain/x/evm/keeper"
+	"github.com/sei-protocol/sei-chain/x/evm/types"
+)
+
+func MigrateBlockBloom(ctx sdk.Context, k *keeper.Keeper) error {
```
yzang2019, line 9: nice!

yzang2019 (approved):

PR #1856, opened 2024-09-15
Add delegatecall flag to properly detect delegatecalls

jewei1997:

Use an explicit flag in the precompile to detect a delegatecall which is set by the delegatecall opcode in go-ethereum. 

existing unit tests


codchen (commented):
In wasmd precompile can we also change for `ExecuteBatch` and `Instantiate`? also can we change to use this flag for these precompiles?
- bank
- distribution
- gov
- ibc
- pointer
- staking

codchen (approved):

Issue #1859, opened 2024-09-17
[BUG] eth_getBlockByNumber returns txhash with null value


quick-pawiromitchel:
**Seid version**
```
name: sei
server_name: <appd>
version: v5.8.0
commit: d5f2a57e21b93c8ef2690003711802cd29e762d3
build_tags: netgo,
go: go version go1.21.13 linux/amd64
build_deps:
- cosmossdk.io/errors@v1.0.0
- filippo.io/edwards25519@v1.0.0-rc.1
- github.com/99designs/keyring@v1.2.1
```

**Chain ID**
Mainnet / Pacific

**Describe the bug**
Running this payload
```
{
  "method": "eth_getBlockByNumber",
  "params": [
    "0x61dcdd1",
    true
  ],
  "id": 1,
  "jsonrpc": "2.0"
}
```

Returns 
```
...
      {
        "blockHash": "0x82673a47da7aef941c1906eb584ebd3a6651d195ecbe662c7306fdf77c03d1b5",
        "blockNumber": "0x61dcdd1",
        "from": "0x6a9ec78779df577a983f4b2190e5469c8374af87",
        "gas": "0x0",
        "gasPrice": null,
        "hash": "0x76694050504f36d141632c82bc581ee983a0aa943f70e6bce073ee0e325b5660",
        "input": "0x7b2263616e63656c5f61756374696f6e223a7b226e6674223a7b2261646472657373223a227365693174746c716d6735777633707070716b6739363674637777657a726e38397a7775326571746d6c64326b686a3271736475347139716b776d6c646c222c22746f6b656e5f6964223a2231313735227d7d7d",
        "nonce": "0x0",
        "to": "0x6a9ec78779df577a983f4b2190e5469c8374af87",
        "transactionIndex": "0x5",
        "value": null,
        "type": "0x0",
        "v": null,
        "r": null,
        "s": null
      }
...
```

**Expected behavior**
No null outputs

PR #1860, opened 2024-09-17
Set 0 instead of null for empty block tx fields

codchen:
Fix for https://github.com/sei-protocol/sei-chain/issues/1859

unit test

### DIFF
--- a/evmrpc/block.go
+++ b/evmrpc/block.go
@@ -17,6 +17,7 @@ import (
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/lib/ethapi"
 	"github.com/ethereum/go-ethereum/rpc"
+	"github.com/sei-protocol/sei-chain/utils"
 	"github.com/sei-protocol/sei-chain/x/evm/keeper"
 	"github.com/sei-protocol/sei-chain/x/evm/types"
 	rpcclient "github.com/tendermint/tendermint/rpc/client"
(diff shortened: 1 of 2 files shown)


evmrpc/block.go
```diff
@@ -217,6 +218,11 @@ func EncodeTmBlock(
 						Input:            m.Msg.Bytes(),
 						Hash:             th,
 						TransactionIndex: (*hexutil.Uint64)(&ti),
+						GasPrice:         (*hexutil.Big)(utils.Big0),
+						Value:            (*hexutil.Big)(utils.Big0),
+						V:                (*hexutil.Big)(utils.Big0),
+						R:                (*hexutil.Big)(utils.Big0),
+						S:                (*hexutil.Big)(utils.Big0),
```
jewei1997, line 228: Do we know that the Graph will accept V,R,S values of 0?

jewei1997 (approved):
overall LGTM

evmrpc/block.go
```diff
@@ -217,6 +218,11 @@ func EncodeTmBlock(
 						Input:            m.Msg.Bytes(),
 						Hash:             th,
 						TransactionIndex: (*hexutil.Uint64)(&ti),
+						GasPrice:         (*hexutil.Big)(utils.Big0),
```
jewei1997, line 221: Should we use EffectiveGasPrice here?

evmrpc/block.go
codchen: good question. Need to confirm with them

PR #1873, opened 2024-09-27
Eip-1559 - target block gas used as param

jewei1997:

Introduce target block gas used as a param instead of just using block gas limit/2. This give us more flexibility to set target gas used independently from the block gas limit. Also, emits metrics for base fee.

unit tests, tested on loadtest cluster with prometheus metrics


app/upgrades.go
```diff
@@ -110,6 +110,7 @@ var upgradesList = []string{
 	"v5.7.4",
 	"v5.7.5",
 	"v5.8.0",
+	"v5.9.1-jeremy-eip-1559",
```
jewei1997: TODO: delete this, was just testing upgrade

codchen (approved):

PR #1874, opened 2024-09-30
Dapp tests - update contract deployment

kollegian:


integration_test/dapp_tests/configs/nftConfig.json
```diff
@@ -0,0 +1,6 @@
+{
+  "marketplace": "0x305E8dc769aae003c17fB08b9cc9108Cb4630E79",
```
besated, line 2: are these constants or different for each chain?

integration_test/dapp_tests/constants.js
```diff
@@ -1,16 +1,25 @@
 const rpcUrls = {
+    "seilocal": "http://127.0.0.1:26657",
     "testnet": "https://rpc-testnet.sei-apis.com",
-    "devnet": "https://rpc-arctic-1.sei-apis.com"
+    "devnet": "https://rpc-arctic-1.sei-apis.com",
+    "devnetFastTrack": "https://rpc-arctic-1.sei-apis.com",
```
besated: can `devnetFastTrack` just use the same rpc as `devnet`?

integration_test/dapp_tests/hardhat.config.js
```diff
@@ -43,5 +52,14 @@ module.exports = {
         count: 1
       },
     },
+    devnetFastTrack: {
```
besated: how is this config different than `devnet`?

besated (commented):
left a couple comments, looks good though! how are these tests run? can it be run via github actions?

integration_test/dapp_tests/configs/nftConfig.json
kollegian: These are for the arctic 1, and already deployed. I didnt create another config for atlantic-2, on the basis that they should be identical but I can deploy and create another config for atlantic 2 as well.

integration_test/dapp_tests/constants.js
kollegian: Yes, updated.

integration_test/dapp_tests/hardhat.config.js
kollegian: Actually they are the same, removed and introduced another flag to the command -f. if passed with devnet it will run against already deployed contract, if passed with seicluster, it will create a config file where it records the contract addresses on the first iteration and run for the subsequent times based on this config file.

integration_test/dapp_tests/configs/nftConfig.json
besated: gotcha, just wanted to make sure it would be identical on `atlantic-2`

besated (approved):

PR #1875, opened 2024-09-30
exclude synthetic logs from eth endpoints

jewei1997:


jewei1997:
closed in favor of https://github.com/sei-protocol/sei-chain/pull/1876

PR #1876, opened 2024-10-01
exclude only shell receipts from eth endpoints

jewei1997:


jewei1997:
will cherry-pick to v5.0.9 with a different PR

PR #1888, opened 2024-10-03
Eip-1559 default values

jewei1997:

Set Eip-1559 default values based on param tuning doc

existing tests


codchen (approved):
we still need a gov proposal to actual update these for the existing networks

PR #1889, opened 2024-10-03
Fix amino registry for custom modules

udpatil:
This modifies the way the amino codec is used for custom modules to verify signatures. This will allow our custom modules to properly be amino signed and verified.

Local chain testing

### DIFF
--- a/x/evm/types/codec.go
+++ b/x/evm/types/codec.go
@@ -18,17 +18,25 @@ import (
 
 var (
 	amino     = codec.NewLegacyAmino()
-	ModuleCdc = codec.NewProtoCodec(codectypes.NewInterfaceRegistry())
-	AminoCdc  = codec.NewAminoCodec(amino)
+	ModuleCdc = codec.NewAminoCodec(amino)
 )
 
 func init() {
+	RegisterCodec(amino)
 	cryptocodec.RegisterCrypto(amino)
 	amino.Seal()
 }
 
 func GetAmino() *codec.LegacyAmino {
 	return amino
+}
+
+func RegisterCodec(cdc *codec.LegacyAmino) {
+	cdc.RegisterConcrete(&MsgAssociate{}, "evm/MsgAssociate", nil)
+	cdc.RegisterConcrete(&MsgEVMTransaction{}, "evm/MsgEVMTransaction", nil)
+	cdc.RegisterConcrete(&MsgSend{}, "evm/MsgSend", nil)
+	cdc.RegisterConcrete(&MsgRegisterPointer{}, "evm/MsgRegisterPointer", nil)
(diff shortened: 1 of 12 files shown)


x/tokenfactory/types/codec.go
```diff
@@ -10,10 +10,10 @@ import (
...
 
 func RegisterCodec(cdc *codec.LegacyAmino) {
-	cdc.RegisterConcrete(&MsgCreateDenom{}, "tokenfactory/create-denom", nil)
-	cdc.RegisterConcrete(&MsgMint{}, "tokenfactory/mint", nil)
-	cdc.RegisterConcrete(&MsgBurn{}, "tokenfactory/burn", nil)
-	cdc.RegisterConcrete(&MsgChangeAdmin{}, "tokenfactory/change-admin", nil)
+	cdc.RegisterConcrete(&MsgCreateDenom{}, "tokenfactory/MsgCreateDenom", nil)
+	cdc.RegisterConcrete(&MsgMint{}, "tokenfactory/MsgMint", nil)
+	cdc.RegisterConcrete(&MsgBurn{}, "tokenfactory/MsgBurn", nil)
+	cdc.RegisterConcrete(&MsgChangeAdmin{}, "tokenfactory/MsgChangeAdmin", nil)
```
codebycarson, line 17: We are missing `tokenfactory/MsgUpdateDenom` it seems

x/evm/types/codec.go
```diff
@@ -31,6 +31,14 @@ func GetAmino() *codec.LegacyAmino {
 	return amino
 }
 
+func RegisterCodec(cdc *codec.LegacyAmino) {
+	cdc.RegisterConcrete(&MsgAssociate{}, "evm/MsgAssociate", nil)
+	cdc.RegisterConcrete(&MsgEVMTransaction{}, "evm/MsgEVMTransaction", nil)
+	cdc.RegisterConcrete(&MsgSend{}, "evm/MsgSend", nil)
+	cdc.RegisterConcrete(&MsgRegisterPointer{}, "evm/MsgRegisterPointer", nil)
+	cdc.RegisterConcrete(&MsgAssociateContractAddress{}, "evm/MsgAssociateContractAddress", nil)
+}
```
codebycarson, line 40: Missing `MsgInternalEVMDelegateCall` and `MsgInternalEVMCall`

udpatil: internal calls can only come from smart contract calls like cosmwasm right? If so they can't be called this way, right? adding them here anyways but not sure if itll make a difference?

udpatil: actually, yeah I dont think we should include this since its not intended to be user facing

x/tokenfactory/types/codec.go
udpatil: updated

Kbhat1 (approved):

PR #1899, opened 2024-10-16
Confidential Transfers feature draft PR

dssei:
Confidential Transfers feature draft PR

- Unit testing


x/confidentialtransfers/types/msgs_test.go
```diff
@@ -0,0 +1,118 @@
...
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/sei-protocol/sei-cryptography/pkg/encryption/elgamal"
+	"github.com/sei-protocol/sei-cryptography/pkg/zkproofs"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestMsgTransfer_FromProto(t *testing.T) {
+	testDenon := "factory/sei1ft98au55a24vnu9tvd92cz09pzcfqkm5vlx99w/TEST"
```
mj850: testDenom*

x/confidentialtransfers/types/msgs_test.go
```diff
@@ -0,0 +1,118 @@
...
+	eg := elgamal.NewTwistedElgamal()
+	sourceKeypair, _ := eg.KeyGen(*sourcePrivateKey, testDenon)
+	destinationKeypair, _ := eg.KeyGen(*destPrivateKey, testDenon)
+
+	amountLo := uint64(100)
+	amountHi := uint64(0)
+
+	remainingBalance := uint64(200)
+
+	// Encrypt the amount using source and destination public keys
```
mj850, line 37: Would it make sense to make this its own function?

func ToProto(transfer *Transfer) *MsgTransfer {}

mj850:
This LGTM so far, to summarize my understanding here:

1. client creates 'usable' transfer object usableTransferObject *Transfer -> This is a version of the object where the ciphertexts/curves.Points are in their usable form.
2. client creates the MsgTransfer object by converting usableTransferObject (maybe by calling some function like MsgTransfer.ToProto(usableTransferObject)) - we need to do this conversion because the *Transfer object is not proto marshallable
3. This message is marshalled as a proto and sent to the server side
4. Server side unmarshals the message and it arrives at the handler as a MsgTransfer object.
5. Server converts the MsgTransfer back to a usable transfer object *Transfer by calling MsgTransfer.FromProto()

Does this sound right?

Questions:
1. We need to perform the intermediate steps 2 and 4 because the regular Transfer object (or more specifically the curves.Points and curves.Scalar objects) are not proto marshallable? 
2. Since we forked the kryptology library, are there any changes we can make there to enable us to marshal the *Transfer object directly without conversion to MsgTransfer?
3. Should we have some method 'ToProto' that converts a Transfer object to a MsgTransfer? 
4. We would likely have to expose the types somehow since we would want external users other than the seid cli to be able to create valid MsgTransfer objects as well. We could add these type in here for now, but it might make sense to think about whether we can eventually make these types and msg generating functions external

PR #1905, opened 2024-10-24
Use limited wasm gas meter

philipsu522:


udpatil (approved):

contracts/test/ERC20toCW20PointerTest.js
```diff
@@ -86,7 +86,7 @@ describe("ERC20 to CW20 Pointer", function () {
             });
 
             describe("transfer()", function () {
-                it("should transfer", async function () {
```
maoueh, line 89: That seems wrong, will now only run that tests, seems like a leftover @philipsu522

PR #1910, opened 2024-10-29
Apply deliver hook gas limit for receipt side effects

stevenlanders:
- set transient receipt was being governed by the transaction gas limit
- this gas limit isn't actually charged against the transaction (receipt already rendered)
- the change is to make this limited by the hook gas limit which is configured 

- integration test passes


stevenlanders:
closing in favor of #1911

PR #1913, opened 2024-10-30
Use tendermint store to get Tx hashes instead of storing explicitly

codchen:
EVM Tx hashes are used for certain EVM RPC queries. Storing EVM Tx hashes explicitly as app state has been causing state bloat. This information is actually already stored in Tendermint block store. This PR changes relevant RPC endpoints to read from Tendermint instead of hashes in app state. Once we are confident that the change is not interruptive to current use cases out there, we can remove the tx hash writing logic and delete the app state.

existing unit test should still work


evmrpc/block.go
```diff
@@ -154,7 +154,24 @@ func (a *BlockAPI) GetBlockReceipts(ctx context.Context, blockNrOrHash rpc.Block
...
+			continue
+		}
+		if evmTx, ok := sdkTx.GetMsgs()[0].(*types.MsgEVMTransaction); ok {
+			if evmTx.IsAssociateTx() {
+				continue
+			}
+			ethtx, _ := evmTx.AsTransaction()
+			txHashes = append(txHashes, ethtx.Hash())
+		}
+	}
```
philipsu522: nit - shoudl we consider moving this to a helper function so it can be reused in filter.go?

philipsu522 (approved):
what are some rpc calls that test this? 
1. a little worried about performance impact
2. i'd also like to vierfy it on a loadtest cluster later today

evmrpc/block.go
codchen: updated

PR #1918, opened 2024-11-01
Overhauling Scripts

alexander-sei:

Complete overhaul of all scripts, from node setup, statesync, snapshots etc.

Tested all scripts on a node with Ubuntu 22.04 LTS


cordt-sei:
wen pr launched

scripts/run-node.py
```diff
@@ -292,16 +411,27 @@ def main():
                 config_data = config_data.replace('persistent-peers = ""', f'persistent-peers = "{persistent_peers}"')
                 config_data = config_data.replace('enable = false', 'enable = true')
                 config_data = config_data.replace('db-sync-enable = true', 'db-sync-enable = false')
-                config_data = config_data.replace('use-p2p = false', 'use-p2p = true')
+                config_data = config_data.replace('fetchers = "4"', 'fetchers = "2"')
+                config_data = config_data.replace('send-rate = 20480000', 'send-rate = 20480000000000')
+                config_data = config_data.replace('recv-rate = 20480000', 'recv-rate = 20480000000000')
+                config_data = config_data.replace('chunk-request-timeout = "15s"', 'chunk-request-timeout = "10s"')
```
cordt-sei, line 417: Lowering this is helpful? I usually raise it

scripts/run-node.py
```diff
@@ -8,21 +8,28 @@
...
 logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(levelname)s - %(message)s')
 
 # Advanced user configs
-moniker = "pynode"  # Custom moniker for the node
-trust_height_delta = 20000  # Negative height offset for state sync
-enable_unsafe_reset = True  # Wipe database and keys before setup
+moniker = "seinode"  # Custom moniker for the node
+trust_height_delta = 0  # Negative height offset for state sync
 version_override = False  # Override version fetching. if true, specify version(s) below
+p2p_port = 26656
```
cordt-sei, line 22: nice

scripts/run-node.py
alexander-sei: Yes, it helps by testing. Idea behind it is that less chunk fetchers allows each chunk to be fetched faster, avoiding the hard coded chunk fetching timeout of 120s.

yzang2019 (approved):

scripts/run-node.py
```diff
@@ -8,21 +8,28 @@
...
 
 # Set up logging
 logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(levelname)s - %(message)s')
 
 # Advanced user configs
-moniker = "pynode"  # Custom moniker for the node
-trust_height_delta = 20000  # Negative height offset for state sync
-enable_unsafe_reset = True  # Wipe database and keys before setup
+moniker = "seinode"  # Custom moniker for the node
+trust_height_delta = 0  # Negative height offset for state sync
```
philipsu522, line 20: why do we set this to 0? state sync usually requires setting some height in the past (typically 10-20k)

scripts/run-node.py
```diff
@@ -8,21 +8,28 @@
...
+lcd_port = 1317
+grpc_port = 9090
+grpc_web_port = 9091
+pprof_port = 6060
 
 # Chain binary version ["version_override" must be true to use]
-MAINNET_VERSION = "v3.9.0"
-DEVNET_VERSION = "v5.2.2"
-TESTNET_VERSION = "v5.2.2"
+MAINNET_VERSION = "v5.9.0-hotfix"
```
philipsu522, line 30: we could potentially start downloading this from https://github.com/sei-protocol/sei-chain/releases

philipsu522: actually, even better i added the version tag to rpc endpoints in a previous pr: https://rpc.sei-apis.com/abci_info | jq .response.version

SecretSaturn: Yes, it's automatically pulling the latest version tag from the rpc endpoint already. This is only for overriding the version

cordt-sei: Yeah these version numbers are only for manual override, otherwise it should fetch the current version from abci_info and download that binary (if available).
Maybe we should have a fallback that will build it from source if the prebuilt bin is not available.

alexander-sei: Yes, it's automatically pulling the latest version tag from the rpc endpoint already. This is only for overriding the version

alexander-sei: @cordt-sei good idea, i'll add this

scripts/run-node.py
cordt-sei: It rounds down to the nearest 100k by default, the way I had it before was not ideal

scripts/run-node.py
philipsu522: the only thing i'd suggest is try to get it from abci_info, if it fails instead ask the owner to manually enter - i would prefer this just because it's better than defaulting to a version that doesn't work

philipsu522 (approved):

PR #1919, opened 2024-11-01
CT types - Apply balance close account

dssei:


proto/confidentialtransfers/tx.proto
```diff
@@ -44,3 +50,21 @@ message Auditor {
...
+  string new_decryptable_available_balance = 3;
+}
+
+message MsgApplyPendingBalanceResponse {}
+
+// Message to be used in close account instruction/transaction
+message MsgCloseAccount {
+  string address = 1;
+  string denom = 2;
+  CloseAccountProof proof = 3;
```
mj850, line 67: Can we call this proofs instead for consistency

x/confidentialtransfers/types/zk.go
```diff
@@ -419,3 +419,95 @@ func (a *Auditor) FromProto() (*TransferAuditor, error) {
...
+}
+
+type CloseAccountProofs struct {
+	ZeroAvailableBalanceProof *zkproofs.ZeroBalanceProof
+	ZeroPendingBalanceLoProof *zkproofs.ZeroBalanceProof
+	ZeroPendingBalanceHiProof *zkproofs.ZeroBalanceProof
+}
+
+func (c *CloseAccountProof) Validate() error {
+	if c.ZeroAvailableBalanceProof == nil || c.ZeroPendingBalanceLoProof == nil || c.ZeroPendingBalanceHiProof == nil {
```
mj850, line 474: Just a thought: Should we make the validate() calls chain validate the child proofs here as well (instead of just checking for nil). Seems like it would be what the validate call should do instead of making that the responsibility of FromProto.

mj850 (approved):
Approved with minor suggestions:

RE the validate suggestion, I can work on that when I merge the PRs if you think that makes sense too.

x/confidentialtransfers/types/zk.go
```diff
@@ -419,3 +419,95 @@ func (a *Auditor) FromProto() (*TransferAuditor, error) {
...
 }
+
+func (z *ZeroBalanceProof) Validate() error {
+	if z.YP == nil || z.YD == nil || z.Z == nil {
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "zero proof is invalid")
+	}
+	return nil
+}
+
+func (z *ZeroBalanceProof) ToProto(zkp *zkproofs.ZeroBalanceProof) *ZeroBalanceProof {
```
mj850, line 430: Not actional for this PR but i'm planning to rework all the ToProto calls to make them just regular constructors `NewZeroBalanceProof(zkp*zkproofs.ZeroBalanceProof) *ZeroBalanceProof` instead of a instance method of z, since it doesn't actually use z.

PR #1922, opened 2024-11-05
Use msg server send in bank precompile

dssei:
Use bank nodule msg server `Send` method in bank precompile `send` to leverage all checks as opposed to using keepers `SendCoins` directly
- unit tests
- functional test with local chain


precompiles/bank/bank.go
```diff
@@ -150,7 +153,18 @@ func (p PrecompileExecutor) send(ctx sdk.Context, caller common.Address, method
 		return nil, err
 	}
 
-	if err := p.bankKeeper.SendCoins(ctx, senderSeiAddr, receiverSeiAddr, sdk.NewCoins(sdk.NewCoin(denom, sdk.NewIntFromBigInt(amount)))); err != nil {
+	msg := &banktypes.MsgSend{
```
codchen, line 156: do we want to change it for `sendNative` too? (`send` takes an EVM recipient, `sendNative` takes a `sei...` address recipient)

codchen (approved):

precompiles/bank/bank.go
dssei: `sendNative` uses `SendCoinsAndWei` which does not have msg server counterpart. We could refactor that first and then re-visit `sendNative`

PR #1923, opened 2024-11-07
Add msg_server methods

mj850:
Adds server side logic to msg_server, as well as sets up the module account so this module can send and receive tokens.

Writing msg_sever_tests right now.


x/confidentialtransfers/keeper/msg_server.go
```diff
@@ -2,48 +2,450 @@ package keeper
...
-func (m msgServer) InitializeAccount(ctx context.Context, account *types.MsgInitializeAccount) (*types.MsgInitializeAccountResponse, error) {
-	//TODO implement me
-	panic("implement me")
-}
+var _ types.MsgServer = msgServer{}
+
+func (m msgServer) InitializeAccount(goCtx context.Context, req *types.MsgInitializeAccount) (*types.MsgInitializeAccountResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// Convert the instruction to proto. This also validates the request.
```
dssei: nit: it's converting from proto

x/confidentialtransfers/keeper/msg_server.go
```diff
@@ -2,48 +2,450 @@ package keeper
...
+		PendingBalanceHi:            instruction.PendingBalanceHi,
+		AvailableBalance:            instruction.AvailableBalance,
+		DecryptableAvailableBalance: instruction.DecryptableBalance,
+		PendingBalanceCreditCounter: 0,
+	}
+
+	// Store the account
+	m.Keeper.SetAccount(ctx, address, req.Denom, account)
+
+	ctx.EventManager().EmitEvents(sdk.Events{
```
dssei, line 81: Seems like `EmitTypedEvents` can be used instead, since `EmitEvents` is deprecated

x/confidentialtransfers/keeper/msg_server.go
```diff
@@ -2,48 +2,450 @@ package keeper
...
+		sdk.NewEvent(
+			types.EventTypeInitializeAccount,
+			sdk.NewAttribute(types.AttributeDenom, instruction.Denom),
+			sdk.NewAttribute(types.AttributeAddress, instruction.FromAddress),
```
dssei, line 85: There probably could be several events like this for single account, which may look all the same for indexer. Adding public key might help to distinguish between them

x/confidentialtransfers/keeper/msg_server.go
```diff
@@ -2,48 +2,450 @@ package keeper
...
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// Validate request
+	err := req.ValidateBasic()
+	if err != nil {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "Invalid Request")
+	}
+
+	// Check if the account exists
+	address, err := sdk.AccAddressFromBech32(req.FromAddress)
```
dssei, line 102: Looks like checking address and returning account is logic that is repeated (and could be extracted into seperate method and re-used)

x/confidentialtransfers/keeper/msg_server.go
```diff
@@ -2,48 +2,450 @@ package keeper
...
-}
+var _ types.MsgServer = msgServer{}
+
+func (m msgServer) InitializeAccount(goCtx context.Context, req *types.MsgInitializeAccount) (*types.MsgInitializeAccountResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// Convert the instruction to proto. This also validates the request.
+	instruction, err := req.FromProto()
+	if err != nil {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "Invalid Msg")
```
dssei: nit: general recommendation is to not capitalize error messages https://go.dev/wiki/CodeReviewComments#error-strings

x/confidentialtransfers/keeper/msg_server.go
```diff
@@ -2,48 +2,450 @@ package keeper
...
+	coins := sdk.NewCoins(sdk.NewCoin(req.Denom, sdk.NewIntFromUint64(req.Amount)))
+
+	// Transfer the amount from the sender's account to the module account
+	if err := m.Keeper.ReceiveTokens(ctx, address, coins); err != nil {
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrInsufficientFunds, "insufficient funds to deposit %d %s", req.Amount, req.Denom)
+	}
+
+	// Split the deposit amount into lo and hi bits.
+	// Extract the bottom 16 bits (rightmost 16 bits)
+	bottom16 := uint16(req.Amount & 0xFFFF)
```
dssei: I think we could follow the PoC approach and have a split logic in a separate method. May simplify testing of it as well

x/confidentialtransfers/keeper/msg_server.go
```diff
@@ -2,48 +2,450 @@ package keeper
...
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "Account does not exist")
+	}
+
+	// Convert the struct to a usable form. This also validates the request.
+	instruction, err := req.FromProto()
+	if err != nil {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "Invalid Msg")
+	}
+
+	// Verify that the account has sufficient funds (Remaining balance after making the transfer is greater than or equal to zero.)
```
dssei, line 195: nit: let's try to keep the line size within 120 chars

x/confidentialtransfers/keeper/msg_server.go
```diff
@@ -2,48 +2,450 @@ package keeper
...
+	// Update the account state
+	account.DecryptableAvailableBalance = instruction.DecryptableBalance
+	account.AvailableBalance = remainingBalanceCalculated
+
+	// Save the account state
+	m.Keeper.SetAccount(ctx, address, req.Denom, account)
+
+	// Return the tokens to the sender
+	coins := sdk.NewCoins(sdk.NewCoin(instruction.Denom, sdk.NewIntFromUint64(instruction.Amount)))
+	if err := m.Keeper.SendTokens(ctx, address, coins); err != nil {
```
dssei, line 230: Out of scope for this method, but curious, how do we prevent funds in module address to not be sent anywhere, say other module address.

x/confidentialtransfers/keeper/msg_server.go
```diff
@@ -2,48 +2,450 @@ package keeper
...
+	// Get the requested address.
+	address, err := sdk.AccAddressFromBech32(req.FromAddress)
+	if err != nil {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "Invalid Address")
+	}
+
+	// Get the user's account
+	account, exists := m.Keeper.GetAccount(ctx, address, req.Denom)
+	if !exists {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "Account does not exist")
```
dssei: There is also not found sdk error 
```
ErrNotFound = Register(RootCodespace, 38, "not found")
```

x/confidentialtransfers/keeper/msg_server.go
```diff
@@ -2,48 +2,450 @@ package keeper
...
+	}
+
+	// Calculate updated balances
+	teg := elgamal.NewTwistedElgamal()
+	newAvailableBalance, err := teg.AddWithLoHi(account.AvailableBalance, account.PendingBalanceLo, account.PendingBalanceHi)
+	if err != nil {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "Error summing balances")
+	}
+
+	zeroCiphertextLo, err := elgamal.SubtractCiphertext(account.PendingBalanceLo, account.PendingBalanceLo)
```
dssei, line 268: Curious, if we need to subtract balance from itself. Why not alternatively just encrypt 0? Can we store nil instead? Also curious if this can be a source of non-determinism

mj850: We use subtraction here since Encryption is non deterministic (since it generates randomness), but subtraction is a deterministic operation.

From a confidentiality pov, It somewhat makes sense to store nil here since the pending balance counter will be set to 0 instead so everyone knows there are no pending balances in the account. 
But from a implementation perspective I think its much easier to deal with accounts if we just assume the pending balances and available balances are always a ciphertext (don't have to check for nil, write special edge cases for when pending balance is nil etc, we can just perform arithmetic on the ciphertext)

x/confidentialtransfers/keeper/msg_server.go
mj850: Makes sense! Refactored all the account not found errors to ErrNotFound

x/confidentialtransfers/keeper/msg_server.go
mj850: Hmmm haven't thought about that. Is this an issue that we need to be aware of?

x/confidentialtransfers/keeper/msg_server.go
mj850: Added a utils package in the module for stuff like this. There's probably 1 or 2 other commonly used things I can add there when I think of them

x/confidentialtransfers/keeper/msg_server.go
mj850: Makes sense, uncapitalized all the error messages in msg, zk and msg_server

x/confidentialtransfers/keeper/msg_server.go
mj850: Refactored the keeper methods (get, set and delete accounts) to take in address as a string instead and try to perform this conversion, since the only place that needs this sdkAccAddress is the keeper for now.

x/confidentialtransfers/keeper/msg_server.go
mj850: Hmm this particular event is specific to the denom and address so I think we should be good! Pubkey generation is deterministic actually so even if the user closes and reinitializes their account here, it should be the same pubkey

x/confidentialtransfers/keeper/msg_server.go
mj850: Going to add a todo to look into this a bit later

x/confidentialtransfers/keeper/msg_server.go
dssei: Never mind, I think denom makes it unique

PR #1927, opened 2024-11-11
Ct queries

dssei:


proto/confidentialtransfers/query.proto
```diff
@@ -3,26 +3,38 @@ package seiprotocol.seichain.confidentialtransfers;
...
-  string test_field = 1;
+message GetAllAccountsRequest {
+  string address = 1;
 }
 
-// TODO: This is a mock method. Remove when we add real queries
-message TestQueryResponse {
-  string test_field = 1;
+message GetAllAccountsResponse {
+  repeated CtAccount accounts = 1;
```
mj850: CtAccount itself does not have reference to the denom associated with the account - we should make sure this query responds with the names of the denoms that the address has accounts for

(actually that's probably the most important thing for this query, I think it's mostly useful to let users know what denoms some account has an confidential transfer accounts for, the user can then query that account directly to get the account details)

x/confidentialtransfers/keeper/keeper.go
```diff
@@ -64,29 +60,36 @@ func NewKeeper(
...
-	bz := store.Get(key)
-	k.cdc.MustUnmarshal(bz, &ctAccount) // Unmarshal the bytes back into the CtAccount object
 	account, err := ctAccount.FromProto()
 	if err != nil {
 		return types.Account{}, false
 	}
 	return *account, true
 }
 
+func (k BaseKeeper) getCtAccount(ctx sdk.Context, address sdk.AccAddress, denom string) (types.CtAccount, bool) {
```
mj850, line 80: Why do we need to have both getAccount and getCtAccount?

x/confidentialtransfers/keeper/keeper.go
```diff
@@ -64,29 +60,36 @@ func NewKeeper(
 	}
 }
 
-func (k BaseKeeper) GetAccount(ctx sdk.Context, address sdk.AccAddress, denom string) (types.Account, bool) {
-	store := ctx.KVStore(k.storeKey)
-	key := types.GetAccountKey(address, denom)
-	if !store.Has(key) {
+func (k BaseKeeper) getAccount(ctx sdk.Context, address sdk.AccAddress, denom string) (types.Account, bool) {
```
mj850: I think we need to export this method?

x/confidentialtransfers/keeper/keeper.go
mj850, line 80: Do we need to export this method as well?

mj850 (commented):
Looks good overall! With some small questions comments

x/confidentialtransfers/keeper/keeper.go
dssei: Was experimenting with the implementation a bit, will remove duplicate code

x/confidentialtransfers/keeper/keeper.go
dssei: I think I refactored it to be private, since the only usage rn is in the querier, and querier itself being a `BaseKeeper` has `GetAccount` method

x/confidentialtransfers/keeper/keeper.go
dssei: For now it's only used in in `getAccount` so I would keep it as is for now.

mj850 (approved):

PR #1931, opened 2024-11-14
Add confidential transfer queries to seid

mj850:
This PR adds seid queries for the confidential transfers module.


x/confidentialtransfers/client/cli/query.go
```diff
@@ -0,0 +1,94 @@
...
+
+	return confidentialTransfersQueryCmd
+}
+
+// GetCmdQueryAccount implements a command to return an account asssociated with the address and denom
+func GetCmdQueryAccount() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "account [address] [denom]",
+		Short: "Query the account state",
+		Args:  cobra.ExactArgs(2),
```
dssei: Let's also add long description with example here and in other command?

dssei:
lgtm. Similarly (to another cli pr), let's add tests where possible

x/confidentialtransfers/client/cli/query.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// GetCmdQueryAccount implements a command to return an account asssociated with the address and denom
+func GetCmdQueryAccount() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "account [denom] [address] [flags]",
+		Short: "Query the account state",
+		Long: "Queries the account state associated with the address and denom." +
+			"Pass the --from flag to decrypt the account." +
+			"Pass the --decryptAvailableBalance flag to attempt to decrypt the available balance.",
```
dssei: usually multi word flags I think are concatenated with `-` e.g.
`--decrypt-available-balance`

x/confidentialtransfers/client/cli/query.go
```diff
@@ -0,0 +1,159 @@
...
+		Args: cobra.ExactArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
```
dssei: nit: I think we were moving those functions to separate methods in tx.  Might make our code consistent

x/confidentialtransfers/client/cli/query.go
```diff
@@ -0,0 +1,159 @@
...
+
+			denom := args[0]
+			address := args[1]
+			// Validate address
+			_, err = sdk.AccAddressFromBech32(address)
+			if err != nil {
+				return fmt.Errorf("invalid address: %v", err)
+			}
+
+			from, err := cmd.Flags().GetString(flags.FlagFrom)
```
dssei: I think `FlagFrom` is already parsed and from address already available in context. E.g.
`clientCtx.GetFromAddress()`

x/confidentialtransfers/client/cli/query.go
```diff
@@ -0,0 +1,159 @@
...
+			if err != nil {
+				return err
+			}
+			fromAddr, _, _, err := client.GetFromFields(clientCtx, clientCtx.Keyring, from)
+			if err != nil {
+				return err
+			}
+
+			res, err := queryClient.GetCtAccount(cmd.Context(), &types.GetCtAccountRequest{
+				Address: address,
```
dssei: I think linter complains as we are not checking the error just below

x/confidentialtransfers/client/cli/query.go
```diff
@@ -0,0 +1,159 @@
...
+					return err
+				}
+
+				decryptor := elgamal.NewTwistedElgamal()
+				keyPair, err := decryptor.KeyGen(*privateKey, denom)
+				if err != nil {
+					return err
+				}
+
+				decryptAvailableBalance, err := cmd.Flags().GetBool("decryptAvailableBalance")
```
dssei: Probably worth moving to a const

x/confidentialtransfers/client/cli/query.go
```diff
@@ -0,0 +1,159 @@
...
+			"Pass the --decryptAvailableBalance flag to attempt to decrypt the available balance.",
+		Args: cobra.ExactArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx, err := client.GetClientQueryContext(cmd)
+			if err != nil {
+				return err
+			}
+			queryClient := types.NewQueryClient(clientCtx)
+
+			denom := args[0]
```
dssei: Do we want to check if denom is empty?

x/confidentialtransfers/client/cli/query.go
mj850: Added for get account, but not get accounts since the short description seems enough (couldn't figure out how to elaborate further)

x/confidentialtransfers/client/cli/query.go
mj850: changed to --decrypt-available-balance

x/confidentialtransfers/client/cli/query.go
mj850: Moved to separate functions for both

x/confidentialtransfers/client/cli/query.go
mj850: FlagFrom is available for getClientTxContext, but not getClientQueryContext, so had to add it manually here

x/confidentialtransfers/client/cli/query.go
mj850: Makes sense, added a check

x/confidentialtransfers/client/cli/query.go
mj850: Added additional validation for denoms

PR #1932, opened 2024-11-14
Use owner event to populate ERC721 transfer topic

codchen:
PR https://github.com/sei-protocol/sei-wasmd/pull/67 introduces a new event type indicating the pre-transfer owner of a CW721 token. This PR leverages the new event type to populate the owner topic of synthetic ERC721 transfer event correctly (previously it's set to the sender, which may be the spender and not the actual owner, and is against ERC721 convention).

Since events and receipts are not part of consensus state, this PR shouldn't be breaking app hash.

unit test

### DIFF
--- a/app/receipt.go
+++ b/app/receipt.go
@@ -60,7 +60,7 @@ func (app *App) AddCosmosEventsToEVMReceiptIfApplicable(ctx sdk.Context, tx sdk.
 		// check if there is a ERC721 pointer to contract Addr
 		pointerAddr, _, exists = app.EvmKeeper.GetERC721CW721Pointer(wasmToEvmEventCtx, contractAddr)
 		if exists {
-			log, eligible := app.translateCW721Event(wasmToEvmEventCtx, wasmEvent, pointerAddr, contractAddr)
+			log, eligible := app.translateCW721Event(wasmToEvmEventCtx, wasmEvent, pointerAddr, contractAddr, response)
 			if eligible {
 				log.Index = uint(len(logs))
 				logs = append(logs, log)
@@ -173,7 +173,7 @@ func (app *App) translateCW20Event(ctx sdk.Context, wasmEvent abci.Event, pointe
 	return nil, false
 }
 
-func (app *App) translateCW721Event(ctx sdk.Context, wasmEvent abci.Event, pointerAddr common.Address, contractAddr string) (*ethtypes.Log, bool) {
+func (app *App) translateCW721Event(ctx sdk.Context, wasmEvent abci.Event, pointerAddr common.Address, contractAddr string, response sdk.DeliverTxHookInput) (*ethtypes.Log, bool) {
 	action, found := GetAttributeValue(wasmEvent, "action")
 	if !found {
 		return nil, false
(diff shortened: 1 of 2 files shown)


app/receipt.go
```diff
@@ -185,9 +185,32 @@ func (app *App) translateCW721Event(ctx sdk.Context, wasmEvent abci.Event, point
 		if tokenID == nil {
 			return nil, false
 		}
+		sender := app.GetEvmAddressAttribute(ctx, wasmEvent, "sender")
+		ownerEvents := GetEventsOfType(response, wasmtypes.EventTypeCW721PreTransferOwner)
+		for _, ownerEvent := range ownerEvents {
```
philipsu522, line 195: might be good to add a comment here re. why we're doing this - something like
"ERC721 standards dictate that the "sender" in the logs are the actual owner, and in cases where the owner is not the caller, we look for a matching token_id and extract the owner from synthetic events"

app/receipt_test.go
```diff
@@ -342,12 +345,17 @@ func TestEvmEventsForCw721(t *testing.T) {
...
+	// make sure that the owner is set correctly to the creator, not the spender.
+	creatorEvmAddr := testkeeper.EVMTestApp.EvmKeeper.GetEVMAddressOrDefault(ctx, creator)
+	require.Equal(t, common.BytesToHash(creatorEvmAddr[:]).Hex(), ownerHash)
+	tokenIdHash = receipt.Logs[0].Topics[3]
+	require.Equal(t, "0x0000000000000000000000000000000000000000000000000000000000000002", tokenIdHash)
 	require.Equal(t, common.HexToHash("0x0").Bytes(), receipt.Logs[0].Data)
 
-	// burn
-	payload = []byte("{\"burn\":{\"token_id\":\"2\"}}")
+	// revoke all
```
philipsu522, line 406: do we also test a transfer on behalf of someone else?

app/receipt.go
```diff
@@ -185,9 +185,37 @@ func (app *App) translateCW721Event(ctx sdk.Context, wasmEvent abci.Event, point
...
+			tokenIDStr, _ := GetAttributeValue(wasmEvent, "token_id")
+			if string(ownerEvent.Attributes[1].Key) != wasmtypes.AttributeKeyTokenId || string(ownerEvent.Attributes[1].Value) != tokenIDStr {
+				continue
+			}
+			if string(ownerEvent.Attributes[2].Key) != wasmtypes.AttributeKeyOwner {
+				continue
+			}
+			ownerAcc, err := sdk.AccAddressFromBech32(string(ownerEvent.Attributes[2].Value))
+			if err != nil {
+				continue
```
stevenlanders, line 209: should we log the error?  (albeit rare/unlikely)

stevenlanders (approved):

dvli2007:
@codchen @stevenlanders @philipsu522 For CW721s, there is no "owner" attribute when a burn happens. You still will need to grab the "sender" attribute. I actually recommend you guys pull out the "burn" event mapping into its own case.

Example of a burn: https://www.seiscan.app/pacific-1/txs/E95774B632C12AA9ABA5D5ACC36999C0EBD36612A8AC7C63199DCEB2993359FF

PR #1933, opened 2024-11-14
Add other confidential tx methods to seid

mj850:
Adds the remaining 4 commands

Manual testing with the seid cli


x/confidentialtransfers/keeper/msg_server.go
```diff
@@ -358,6 +358,10 @@ func (m msgServer) Transfer(goCtx context.Context, req *types.MsgTransfer) (*typ
 		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "invalid msg")
 	}
 
+	if instruction.FromAddress == instruction.ToAddress {
```
dssei, line 381: I guess I agree, but In bank module though, I think it's a valid case

x/confidentialtransfers/client/cli/tx.go
```diff
@@ -173,3 +168,285 @@ func makeCloseAccountCmd(cmd *cobra.Command, args []string) error {
 
 	return tx.GenerateOrBroadcastTxCLI(clientCtx, cmd.Flags(), msg)
 }
+
+// NewTransferTxCmd returns a CLI command handler for creating a MsgTransfer transaction.
+func NewTransferTxCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "transfer [denom] [to_address] [amount] [flags]",
+		Short: "Make a confidential transfer to another address",
+		Long: `Transfer command create a confidential transfer of the specified amount of the specified denomination to the specified address.
```
dssei, line 191: nit: We could also add some example commands in long description. I think i rushed my piece in without it though

x/confidentialtransfers/client/cli/tx.go
```diff
@@ -173,3 +168,285 @@ func makeCloseAccountCmd(cmd *cobra.Command, args []string) error {
...
+	queryClient := types.NewQueryClient(queryClientCtx)
+
+	privKey, err := getPrivateKey(cmd)
+	if err != nil {
+		return err
+	}
+
+	fromAddress := clientCtx.GetFromAddress().String()
+	denom := args[0]
+	toAddress := args[1]
```
dssei, line 228: we may want to validate inputs. I think I skipped those checks in commands I implemented, and we may need to revisit here and other places

dssei:
Code LGTM in general. Let's fix lint errors and add unit tests where possible.

x/confidentialtransfers/keeper/msg_server.go
mj850: Disabled this because the way we write this is
senderState = getAccount(sender)
recipientState = getAccount(recipient)

senderState.Balance += Amount
recipientState.Balance -= Amount

keeper.Store(senderState, sender)
keeper.Store(recipientState, recipient).

If sender and recipient are same address, the updated 'recipientState' overrides the updated 'senderState' for the key 'sender' and funds are not deducted. If we want to enable this we need to write a special case.

x/confidentialtransfers/client/cli/tx.go
mj850: Updated the commands with denom and address validation

x/confidentialtransfers/utils/utils.go
```diff
@@ -17,3 +17,11 @@ func SplitTransferBalance(amount uint64) (uint16, uint32, error) {
 
 	return bottom16, next32, nil
 }
+
+// Combines the bottom 32 bits and the next 48 bits into a single 64-bit number.
```
dssei: nit: godoc usually stars with method name e.g
```
// CombineTransferAmount Combines the bottom 32 bits and the next 48 bits into a single 64-bit number.
```

x/confidentialtransfers/utils/utils.go
```diff
@@ -17,3 +17,11 @@ func SplitTransferBalance(amount uint64) (uint16, uint32, error) {
...
+func CombineTransferAmount(bottom16 uint16, hi uint32) (uint64, error) {
```
dssei: Probably deserves it's own unit test

dssei (approved):

PR #1937, opened 2024-11-14
Update confidential transfers swagger/openAPI spec

dssei:
Update confidential transfers swagger/openAPI spec

Functional test on swagger page http://localhost:1317/swagger


proto/confidentialtransfers/query.proto
```diff
@@ -12,7 +12,7 @@ option go_package = "github.com/sei-protocol/sei-chain/x/confidentialtransfers/t
 service Query {
   rpc GetCtAccount (GetCtAccountRequest) returns (GetCtAccountResponse) {
     option (google.api.http) = {
-      get: "/seichain/confidentialtransfers/account/{address}/{denom}"
+      get: "/seichain/confidentialtransfers/account/{address}/denom"
```
mj850, line 15: I don't think we've implemented the rest route for this yet, we would have to take special care to add the denom as a parameter right?

mj850 (approved):

proto/confidentialtransfers/query.proto
dssei: this will be added automatically

PR #1950, opened 2024-11-20
Add query decrypted tx method to confidential transfers client

mj850:
We want users to have the ability to easily query a tx hash in it's decrypted state (if they have the private key)

This PR adds a seid method seid q ct tx <HASH> --from <USER> that queries a tx by it's hash, then decrypts it using the user's private key.

Manual testing using the updated seid

### DIFF
--- a/x/confidentialtransfers/client/cli/query.go
+++ b/x/confidentialtransfers/client/cli/query.go
@@ -1,11 +1,18 @@
 package cli
 
 import (
+	"context"
+	"crypto/ecdsa"
+	"encoding/hex"
 	"fmt"
 
 	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/codec"
+	cdctypes "github.com/cosmos/cosmos-sdk/codec/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/tx"
+	"github.com/gogo/protobuf/proto"
 	"github.com/sei-protocol/sei-chain/x/confidentialtransfers/types"
 	"github.com/sei-protocol/sei-cryptography/pkg/encryption"
 	"github.com/sei-protocol/sei-cryptography/pkg/encryption/elgamal"
@@ -17,7 +24,7 @@ const decryptAvailableBalanceFlag = "decrypt-available-balance"
 // GetQueryCmd returns the cli query commands for the minting module.
 func GetQueryCmd() *cobra.Command {
 	confidentialTransfersQueryCmd := &cobra.Command{
-		Use:                        types.ModuleName,
+		Use:                        types.ShortModuleName,
 		Short:                      "Querying commands for the confidential transfer module",
 		DisableFlagParsing:         true,
 		SuggestionsMinimumDistance: 2,
@@ -27,6 +34,7 @@ func GetQueryCmd() *cobra.Command {
 	confidentialTransfersQueryCmd.AddCommand(
 		GetCmdQueryAccount(),
 		GetCmdQueryAllAccount(),
+		GetCmdQueryTx(),
 	)
 
 	return confidentialTransfersQueryCmd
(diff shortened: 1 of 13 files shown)


x/confidentialtransfers/client/cli/query.go
```diff
@@ -178,3 +186,126 @@ func queryAllAccounts(cmd *cobra.Command, args []string) error {
...
 	return nil
 }
+
+// GetCmdQueryTx implements a command to query a tx by it's transaction hash and return it in it's decrypted state.
+func GetCmdQueryTx() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "tx [hash]",
+		Short: "Query the confidential transaction and decrypts it",
+		Long: "Query the confidential transaction by it's tx hash and decrypts it using the private key of the account in --from" +
+			"Pass the --from flag to decrypt the account" +
```
dssei, line 196: Passing non-owner here won't decrypt it, right? Maybe we could revise the wording everywhere a bit to be more to deliver the overall message like
``` Query for confidential transfer tx. Returns tx in decrypted form if owner addressed passed in --from flag```
We could also add example with and without `--from` flag

File: x/confidentialtransfers/client/cli/query.go [hunk 2]
```diff
@@ -178,3 +186,126 @@ func queryAllAccounts(cmd *cobra.Command, args []string) error {
 
 	return nil
 }
+
+// GetCmdQueryTx implements a command to query a tx by it's transaction hash and return it in it's decrypted state.
+func GetCmdQueryTx() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "tx [hash]",
+		Short: "Query the confidential transaction and decrypts it",
```
dssei, line 195: I think we may need to change the wording a bit to avoid confusion. Made me think this command works also for non-involved parties (returning encrypted tx.
```
Query to decrypt confidential transfers transaction
```

x/confidentialtransfers/client/cli/query.go
```diff
@@ -178,3 +186,126 @@ func queryAllAccounts(cmd *cobra.Command, args []string) error {
...
+		Long: "Query the confidential transaction by it's tx hash and decrypts it using the private key of the account in --from" +
```
dssei: Query the confidential transaction by it's tx hash and decrypts it using the private key of the account in --address. For decryption to work, the address should be of a sender, receiver or auditor.

x/confidentialtransfers/client/cli/query.go
dssei: Maybe let's rename this flag to --address instead or similar?

x/confidentialtransfers/types/apply_pending_balance.go
```diff
@@ -0,0 +1,119 @@
...
+		Address:                        address,
+		Denom:                          denom,
+		NewDecryptableAvailableBalance: newDecryptableAvailableBalance,
+		CurrentPendingBalanceCounter:   uint32(currentPendingBalanceCounter),
+		CurrentAvailableBalance:        currentAvailableBalance,
+	}, nil
+}
+
+// ApplyPendingBalanceDecrypted is the decrypted version of ApplyPendingBalance
+func (r *ApplyPendingBalance) Decrypt(decryptor *elgamal.TwistedElGamal, privKey ecdsa.PrivateKey, decryptAvailableBalance bool, address string) (*ApplyPendingBalanceDecrypted, error) {
```
dssei: bool flag signals the method needs to be split in two. e.g `Decrypt` and `DecryptWithAvailableBalance`

x/confidentialtransfers/types/apply_pending_balance.go
```diff
@@ -0,0 +1,119 @@
...
+	"crypto/ecdsa"
+	"errors"
+	"strconv"
+
+	"github.com/sei-protocol/sei-chain/x/confidentialtransfers/utils"
+	"github.com/sei-protocol/sei-cryptography/pkg/encryption"
+	"github.com/sei-protocol/sei-cryptography/pkg/encryption/elgamal"
+)
+
+// MsgApplyPendingBalance is a message to apply the pending balance to the available balance
```
dssei: MsgApplyPendingBalance -> ApplyPendingBalance

x/confidentialtransfers/types/apply_pending_balance.go
```diff
@@ -0,0 +1,119 @@
...
+	return &ApplyPendingBalance{
+		Address:                        address,
+		Denom:                          denom,
+		NewDecryptableAvailableBalance: newDecryptableAvailableBalance,
+		CurrentPendingBalanceCounter:   uint32(currentPendingBalanceCounter),
+		CurrentAvailableBalance:        currentAvailableBalance,
+	}, nil
+}
+
+// ApplyPendingBalanceDecrypted is the decrypted version of ApplyPendingBalance
```
dssei: ApplyPendingBalanceDecrypted -> Decrypt

x/confidentialtransfers/types/apply_pending_balance.go
dssei: we may need to add inputs validation for decryptor, privkey

dssei: `address` not used

x/confidentialtransfers/types/initialize_account.go
```diff
@@ -100,3 +101,50 @@ func NewInitializeAccount(address, denom string, privateKey ecdsa.PrivateKey) (*
 		Proofs:             &proofs,
 	}, nil
 }
+
+func (r InitializeAccount) Decrypt(decryptor *elgamal.TwistedElGamal, privKey ecdsa.PrivateKey, decryptAvailableBalance bool, address string) (*InitializeAccountDecrypted, error) {
```
dssei: let's add godoc for consistency

dssei: `address` not used here as well

x/confidentialtransfers/types/transfer.go
```diff
@@ -205,6 +215,113 @@ func NewTransfer(
 	}, nil
 }
 
+// Decrypts the transfer transaction and returns the decrypted data.
+// The aesKey and decryptAvailableBalance field are only used if address is the sender address of the transaction.
+func (r *Transfer) Decrypt(decryptor *elgamal.TwistedElGamal, privKey ecdsa.PrivateKey, decryptAvailableBalance bool, address string) (*TransferDecrypted, error) {
```
dssei: I think generalization in a form of `Decryptable` interface, leads to more complexity in current implementation and/or degraded implementation (like skipped address params in other implementation).
We may need type (sender/receiver/auditor) based implementations of `Decryptable` and a separate factory function to avoid types based decision in this method.

dssei (changes requested):

x/confidentialtransfers/client/cli/query.go
mj850: Updated description

x/confidentialtransfers/client/cli/query.go
mj850: Renamed to --decryptor here

x/confidentialtransfers/client/cli/query.go
mj850: Resolved this by updating the Long description and adding a verbose error message if --decryptor is not set

x/confidentialtransfers/types/apply_pending_balance.go
mj850: Tried this but decided not to change in the end - felt a little bit awkward to have 2 methods (Decrypt and DecryptWithAvailableBalance) that do almost the same thing in this specific case, since the only use case here is:
Current:
```
	var result proto.Message
	switch message := sdkmsg.(type) {
	case *types.MsgInitializeAccount:
		result, err = message.Decrypt(decryptor, *privKey, decryptAvailableBalance)
	case *types.MsgWithdraw:
		result, err = message.Decrypt(decryptor, *privKey, decryptAvailableBalance)
	case *types.MsgApplyPendingBalance:
		result, err = message.Decrypt(decryptor, *privKey, decryptAvailableBalance)
	case *types.MsgTransfer:
		result, err = message.Decrypt(decryptor, *privKey, decryptAvailableBalance, address)
	case *types.MsgDeposit:
		result = message
	case *types.MsgCloseAccount:
		result = message
	default:
		return nil, false, nil
```
With separate methods it would become:
```
	var result proto.Message
	switch message := sdkmsg.(type) {
	case *types.MsgInitializeAccount:
                 if decryptAvailableBalance {
                         result, err = message.DecryptWithAvailableBalnace(decryptor, *privKey)
                 } else {
		        result, err = message.Decrypt(decryptor, *privKey)
                 }

	case *types.MsgWithdraw:
		if decryptAvailableBalance {
                         result, err = message.DecryptWithAvailableBalnace(decryptor, *privKey)
                 } else {
		        result, err = message.Decrypt(decryptor, *privKey)
                 }
	case *types.MsgApplyPendingBalance:
                 if decryptAvailableBalance {
                         result, err = message.DecryptWithAvailableBalnace(decryptor, *privKey)
                 } else {
		        result, err = message.Decrypt(decryptor, *privKey)
                 }
	case *types.MsgTransfer:
                 if decryptAvailableBalance {
                         result, err = message.DecryptWithAvailableBalnace(decryptor, *privKey)
                 } else {
		        result, err = message.Decrypt(decryptor, *privKey)
                 }
	case *types.MsgDeposit:
		result = message
	case *types.MsgCloseAccount:
		result = message
	default:
		return nil, false, nil


Point taken for future methods though I think it's a helpful idea

mj850: Added validation for decryptor to prevent panic, but not for privkey since empty privKey will be reported as an invalid privKey.

mj850: Removed address from unrequired methods

x/confidentialtransfers/types/initialize_account.go
mj850: Added documentation here

x/confidentialtransfers/types/transfer.go
mj850: Thanks for the feedback - since we figured out how to parse the message type properly, I removed the Decryptable interface and just have it call Decrypt for each method.

Also split the Decrypt logic up in the Transfer class to shorten the function.

dssei (approved):

PR #1952, opened 2024-11-20
fix: complete failed transaction receipts with block data

blindchaser:

This PR enhances the handling of failed transaction receipts by filling in missing fields using data from the block. When a transaction fails with zero gas usage, the receipt might be missing important information. This PR ensures all receipt fields are properly populated by looking up the transaction in the block.

In `GetTransactionReceipt`, when a receipt has `TxType == 0` and `GasUsed == 0`, we:
1. Fetch the block containing the transaction
2. Find the matching transaction in the block
3. Update receipt fields with actual transaction data:
   - From address (using transaction signer)
   - To address or contract address
   - Gas used
   - Transaction type
   - Status (set to failed)


jewei1997 (approved):

evmrpc/tx.go
```diff
@@ -53,6 +53,44 @@ func (t *TransactionAPI) GetTransactionReceipt(ctx context.Context, hash common.
 		}
 		return nil, err
 	}
+	// Fill in the receipt if the transaction has failed and used 0 gas
+	if receipt.TxType == 0 && receipt.GasUsed == 0 {
```
codchen: should check status instead of txType

evmrpc/tx.go
```diff
@@ -53,6 +53,44 @@ func (t *TransactionAPI) GetTransactionReceipt(ctx context.Context, hash common.
...
+					receipt.To = etx.To().Hex()
+					receipt.ContractAddress = ""
+				} else {
+					receipt.To = ""
+					// For contract creation transactions, calculate the contract address
+					receipt.ContractAddress = crypto.CreateAddress(from, etx.Nonce()).Hex()
+				}
+				receipt.TxType = uint32(etx.Type())
+				receipt.Status = uint32(ethtypes.ReceiptStatusFailed)
+				receipt.GasUsed = 0
```
codchen, line 90: i think you can also fill other fields like input, value, etc. from `etc`

evmrpc/tx.go
blindchaser: fixed

evmrpc/tx.go
blindchaser: it looks receipt doesn't have these fields: https://github.com/sei-protocol/sei-chain/blob/main/x/evm/types/receipt.pb.go#L102

i checked some successfully tx receipts, they have logs but it doesn't have input/value either.

evmrpc/tx.go
```diff
@@ -53,6 +53,44 @@ func (t *TransactionAPI) GetTransactionReceipt(ctx context.Context, hash common.
 		}
 		return nil, err
 	}
+	// Fill in the receipt if the transaction has failed and used 0 gas
+	if receipt.Status == 0 && receipt.GasUsed == 0 {
```
jewei1997: can you add a comment here saying that this case is for when a tx fails before it makes it to the VM?

PR #1959, opened 2024-11-22
add ct to load test framework

dssei:
Added 4 types of load tests for confidential transfers testing
- Ran loadtest cluster setup + performance testing


mj850 (approved):
Looks good to me!

Few questions though:

1. Are we still using a flag in code to determine whether to run the ct module commands as part of the loadtest? If so, should we write a readme or TODO somewhere on where this is enabled?

2. What process calls loadtest/scripts/populate_ct_accounts.py? Didn't see any mention of it

dssei:
> Looks good to me!
> 
> Few questions though:
> 
> 1. Are we still using a flag in code to determine whether to run the ct module commands as part of the loadtest? If so, should we write a readme or TODO somewhere on where this is enabled?

Not using flag anymore
> 2. What process calls loadtest/scripts/populate_ct_accounts.py? Didn't see any mention of it

You would have to run that manually (added readme me in infra changes)

Infra changes here https://github.com/sei-protocol/sei-infra/pull/874

PR #1970, opened 2024-12-03
feat: add unbonding delegation query

blindchaser:


go.mod
```diff
@@ -344,7 +344,8 @@ require (
 )
 
 replace (
-	github.com/CosmWasm/wasmd => github.com/sei-protocol/sei-wasmd v0.2.9
+	github.com/CosmWasm/wasmd => github.com/sei-protocol/sei-wasmd v0.0.0-20241203200623-bdd858fb6f9b
```
yzang2019: We should cut a tag instead of using a hash here

blindchaser: updated

yzang2019 (approved):

PR #1972, opened 2024-12-03
Manually consume gas for expensive txes and add params

mj850:
We want to add 2 functionalities that deal with params.
1. Add the ability to disable the module via gov params (since it's in beta right now)
2. Add the ability to scale gas costs according for txs with range proofs. Range proofs are computationally expensive but do not add to computed gas, so we need to manually consume gas.

Manual testing for gas and param changes


proto/confidentialtransfers/params.proto
```diff
@@ -4,4 +4,7 @@ package seiprotocol.seichain.confidentialtransfers;
 option go_package = "github.com/sei-protocol/sei-chain/x/confidentialtransfers/types";
 
 // Params defines the parameters for the confidential tokens module.
-message Params {}
+message Params {
+  bool enable_feature = 1;
```
dssei: nit: consider dropping `feature` or addiing feature name itself. Like we have in other configs e.g. `enable-unsafe-cors`

x/confidentialtransfers/client/cli/query.go
```diff
@@ -125,6 +125,13 @@ func queryAccount(cmd *cobra.Command, args []string) error {
 			return err
 		}
 
+		if decryptAvailableBalance {
+			err = clientCtx.PrintString("--decrypt-available-balance set to true. This operation could take a long time and may not succeed even if the private key provided is valid\n")
```
dssei: nit: consider splitting in to two lines.

x/confidentialtransfers/keeper/keeper.go
```diff
@@ -175,6 +178,20 @@ func (k BaseKeeper) SetParams(ctx sdk.Context, params types.Params) {
 	k.paramSpace.SetParamSet(ctx, &params)
 }
 
+// Method to retrieve the parameter value
+func (k BaseKeeper) IsFeatureEnabled(ctx sdk.Context) bool {
+	var enableFeature bool
+	k.paramSpace.Get(ctx, types.KeyEnableFeature, &enableFeature)
```
dssei: nit: I guess we could have consistent naming for all vars? 
```
func (k BaseKeeper) IsFeatureEnabled(ctx sdk.Context) bool {
	var isFeatureEnabled bool
	k.paramSpace.Get(ctx, types.IsFeatureEnabledKey, &isFeatureEnabled)
	return isFeatureEnabled
}
```

x/confidentialtransfers/keeper/keeper.go
```diff
@@ -175,6 +178,20 @@ func (k BaseKeeper) SetParams(ctx sdk.Context, params types.Params) {
 	k.paramSpace.SetParamSet(ctx, &params)
 }
 
+// Method to retrieve the parameter value
```
dssei: nit: Missed method name in godoc here and in next method

x/confidentialtransfers/types/keys.go
```diff
@@ -19,10 +19,15 @@ const (
 
 	// QuerierRoute defines the module's query routing key
 	QuerierRoute = ModuleName
+
+	DefaultEnableFeature           = true
```
dssei: These constants may belong to params.go file.

x/confidentialtransfers/keeper/msg_server.go
```diff
@@ -242,6 +254,14 @@ func (m msgServer) Withdraw(goCtx context.Context, req *types.MsgWithdraw) (*typ
 		return nil, sdkerrors.Wrapf(sdkerrors.ErrInsufficientFunds, "insufficient funds to withdraw %s %s", req.Amount, req.Denom)
 	}
 
+	gasSoFar := ctx.GasMeter().GasConsumed()
+	multiplier := m.Keeper.GetRangeProofGasMultiplier(ctx)
+
+	// Consume additional gas according to the multiplier as range proofs are computationally expensive.
+	if multiplier > 1 {
```
dssei, line 261: curious, is there a chance it could trigger with existing validation in place?

x/confidentialtransfers/keeper/msg_server.go
```diff
@@ -242,6 +254,14 @@ func (m msgServer) Withdraw(goCtx context.Context, req *types.MsgWithdraw) (*typ
...
+		ctx.GasMeter().ConsumeGas(gasSoFar*uint64(multiplier-1), "range proof verification")
```
dssei, line 262: curious why are we decrementing multiplier?

x/confidentialtransfers/keeper/msg_server_test.go
```diff
@@ -456,6 +478,38 @@ func (suite *KeeperTestSuite) TestMsgServer_DepositTooManyPendingBalances() {
 	suite.Require().ErrorContains(err, "account has too many pending transactions")
 }
 
+// Tests that the Deposit method fails when the feature is disabled.
+func (suite *KeeperTestSuite) TestMsgServer_DepositFeatureDisabled() {
```
dssei, line 482: I think we may not need half of init steps here, that distract from actual test. This works and might apply to other tests:
```
func (suite *KeeperTestSuite) TestMsgServer_DepositFailsIfFeatureDisabled() {
	// Disable the confidential tokens module via params
	suite.Ctx = suite.App.BaseApp.NewContext(false, tmproto.Header{})
	params := types.DefaultParams()
	params.EnableFeature = false
	suite.App.ConfidentialTransfersKeeper.SetParams(suite.Ctx, params)

	_, err := suite.msgServer.Deposit(sdk.WrapSDKContext(suite.Ctx), &types.MsgDeposit{})
	suite.Require().Error(err, "Should not be able to deposit when feature is disabled")
	suite.Require().ErrorContains(err, "feature is disabled by governance")
}
```

x/confidentialtransfers/types/params.go
```diff
@@ -11,7 +13,10 @@ func ParamKeyTable() paramtypes.KeyTable {
 
 // DefaultParams default confidential transfers module parameters.
 func DefaultParams() Params {
-	return Params{}
+	return Params{
```
dssei, line 22: Looks like we might be missing some tests

x/confidentialtransfers/keeper/msg_server.go
mj850: Hmm not sure what you mean here? Default multiplier is 10 so it should always trigger, but 1 is also a valid multiplier (meaning don't multiply anything), meaning it will skip the following step

x/confidentialtransfers/keeper/msg_server.go
mj850: Multiplier tells us how much to multiply the gas cost by - gasSoFar represents the amount of gas the tx would normally spend without any manipulation.
So let's say multiplier is set to 10, then gas cost will be 

gasSoFar + (9 x gasSoFar) = 10 x gasSoFar

x/confidentialtransfers/keeper/msg_server_test.go
mj850: Yup you're right that based on the way the code is written now there's technically no need to initialize (since the IsModuleEnabled is the first check) but wanted to write the tests in a way that the only reason why it would fail is because the flag was not doing it's job (and not due to something else)

The last line 
`suite.Require().ErrorContains(err, "feature is disabled by governance")` does check the failure reason is correct, but I thought it seems a bit flaky to rely on the order of checks:

For example, if in the future we added a check to see if the account exists first, before checking the CtModuleEnabled flag, then this test would fail (even though the flag is turned off and would have been working as expected).

x/confidentialtransfers/types/params.go
mj850: Do you think we need tests for DefaultParams and ParamSetPairs? They seem to just be returning objects

proto/confidentialtransfers/params.proto
mj850: Renamed to EnableCtModule for a clarity! Lmk if this makes sense

x/confidentialtransfers/keeper/keeper.go
mj850: Renamed everything to EnableCtModule, and isCtModuleEnabled

x/confidentialtransfers/keeper/keeper.go
mj850: Added method names to godoc

x/confidentialtransfers/types/keys.go
mj850: Moved to params.go file

x/confidentialtransfers/types/params.go
mj850: Added tests for other methods here

x/confidentialtransfers/keeper/msg_server.go
mj850: Added comment here to explain the calculation

dssei (approved):

PR #1974, opened 2024-12-04
Optimize getLogs with parallelization using fixed amount workers

yzang2019:
This PR add parallelization for eth_getLogs to speed up the query, especially on hundreds/thousands of blocks on archive node.

### DIFF
--- a/evmrpc/filter.go
+++ b/evmrpc/filter.go
@@ -25,6 +25,8 @@ import (
 
 const TxSearchPerPage = 10
 
+const MaxNumOfWorkers = 500
+
 type FilterType byte
 
 const (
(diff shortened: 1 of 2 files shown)


evmrpc/filter.go
```diff
@@ -23,6 +25,8 @@ import (
 
 const TxSearchPerPage = 10
 
+const MaxNumOfWorkers = 500
```
philipsu522, line 28: nit do we want to allow this to be tunable for quicknode? i.e. shall we make it a config?

philipsu522 (approved):

codchen (approved):

evmrpc/filter.go
yzang2019: Yeah we definitely could, but I feel this config doesn't need to be tuned much, max concurrency of 500 is already a lot, adding more goroutines probably wouldn't help any futher more

PR #1975, opened 2024-12-05
Use updated CachedRangeVerifier

mj850:
Range proof verification is slow, with 30% of the time spent on instantiating a RangeProofVerifier.

We updated the sei-cryptography library so that these verifiers are cached and can be reused. This PR updates the package so we benefit from the new optimization

Existing unit tests


mj850:
NOTE: This PR currently uses a non prod version of the sei-cryptography package and is meant as a draft. We should merge the update to the sei-cryptography library, then point this towards the version in main, before merging this PR

dssei (approved):

PR #1981, opened 2024-12-08
Use Signed Message for Key Devrivation

mj850:
Currently, the CT Module relies directly on the private key for deriving keys. For each account, a new keypair is deterministically generated using the KeyGen(privateKey, denom) method, which computes the SHA256 of the privateKey with denom as the salt, then uses that hash to create a private key (random number)

The benefit of doing this is that users can generate a key that is as secure as their addresses' ecdsa private key, without having to manage additional keys.

This is fine for clients such as seid which have direct access to the users private key, but the majority of other clients (wallets etc), do not have direct access to the users private key. This would make the feature very difficult to use outside of seid.

To make things easier for wallets, the change proposed here is that instead of using `SHA256Hash(privatekey, denom)` as the key generation function, we use `Sign(privateKey, denom)`, then SHA256 hash it one more time with an arbitrary salt to prevent the signature from being used. 

Sign(privateKey, denom) is as secure, since it cannot be generated by anyone except with knowledge of the ecdsa private key. However, this enables the key generation to be performed by Wallets, enabling applications to derive the keyPair on the client side by requesting that the user sign the denom. This will increase the future usability of this feature and integratability with dapps.

On the client side, they can generate Sign(privateKey, denom) using EVM wallet clients with
```
        const message = "tokenfactory/address/denom"

        // Hash the prefixed message
        const messageBytes = new TextEncoder().encode(message);
        const hash = keccak256(messageBytes);

        signatureHex = signMessage({ message: hash})
```

One note is that this implementation only enables EVM wallets to do this - cosmos wallets have different signing methods and will still not be able to interact with this module naturally.

more work to be done to look into how to add usability to this module.

This PR is dependent on the changes in https://github.com/sei-protocol/sei-cryptography/pull/7

- Unit tests/integration tests updated - module still works as expected
- Did a POC to make sure the hex signature produced by viem client matches the one produced by the new GetSignedDenom method


dssei (commented):
lgtm

PR #1983, opened 2024-12-09
handle oracle overflow and rounding to zero for extreme values with cross rates

udpatil:
This fixes issues that can occur in the most extreme circumstances where extreme values used for cross rates can cause issues during oracle ballot calculations.

Unit test + docker cluster testing

### DIFF
--- a/x/oracle/abci.go
+++ b/x/oracle/abci.go
@@ -66,10 +66,6 @@ func MidBlocker(ctx sdk.Context, k keeper.Keeper) {
 			voteMapRD := ballotRD.ToMap()
 
 			exchangeRateRD := ballotRD.WeightedMedianWithAssertion()
-			// handle case where exchange rate is 0
-			if exchangeRateRD.IsZero() {
-				// TODO: do stuff
-			}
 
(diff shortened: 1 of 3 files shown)


x/oracle/abci.go
```diff
@@ -66,6 +66,10 @@ func MidBlocker(ctx sdk.Context, k keeper.Keeper) {
 			voteMapRD := ballotRD.ToMap()
 
 			exchangeRateRD := ballotRD.WeightedMedianWithAssertion()
+			// handle case where exchange rate is 0
+			if exchangeRateRD.IsZero() {
+				// TODO: do stuff
```
stevenlanders: do we actually want to do things?

stevenlanders (approved):
^ just wasn't sure of that empty branch, but otherwise lgtm

x/oracle/abci.go
udpatil: yeah I realized this one can't actually be zero because votes submitted with a zero value are modified to have the vote power set to 0, so the only way to get 0 exchange rates is for really small number divided by really large number which is with the cross rates calculations.

PR #1987, opened 2024-12-11
Consume gas before verifying range proof

mj850:
Consuming gas only after performing all operations could lead to exploits where a lot of work is done but no gas is consumed (or gas provided is insufficient). The correct practice should be to charge for gas before performing the work

Since we perform validation before other operations, it is not feasible to use a scaling gas multiplier. Instead, we can use a fixed gas cost (set as 1000000 to start with). This is currently roughly the amount we were going to be charging with the default 10x multiplier as well.

Current unit and integration tests


x/confidentialtransfers/types/params.go
```diff
@@ -9,8 +9,8 @@ import (
 // DefaultEnableCtModule is the default value for the EnableCtModule flag.
 const DefaultEnableCtModule = true
 
-// DefaultRangeProofGasMultiplier is the default value for RangeProofGasMultiplier param.
-const DefaultRangeProofGasMultiplier = uint32(10)
+// DefaultRangeProofGasCost is the default value for RangeProofGasCost param.
+const DefaultRangeProofGasCost = uint32(1000000)
```
dssei: Nit: we could use uint64 instead (to use the same type as Gas amount)

dssei (commented):
lgtm

x/confidentialtransfers/types/params.go
mj850: Good call, made this change

dssei (approved):

PR #1989, opened 2024-12-12
Updates to use silo tendermint

silostaking:


silostaking:
You have to configure your local git ssh keys/config

And then run

```git config --global url."ssh://git@github-silo/".insteadOf "https://github.com/"```

this allows the use of private libraries.

silostaking:
If you want a tutorial on how to get sei chain running:

PR #1993, opened 2024-12-17
Avoid panics in getLogs

yzang2019:
This PR fix the problem of panic, we should not panic in EVM getLogs, instead we can just update the error.

### DIFF
--- a/evmrpc/filter.go
+++ b/evmrpc/filter.go
@@ -321,6 +321,7 @@ func (f *LogFetcher) GetLogsByFilters(ctx context.Context, crit filters.FilterCr
 	}
 
 	// Parallelize execution
+	var mu = sync.Mutex{}
 	numWorkers := int(math.Min(MaxNumOfWorkers, float64(end-begin+1)))
 	var wg sync.WaitGroup
 	tasksChan := make(chan int64, end-begin+1)
@@ -345,6 +346,11 @@ func (f *LogFetcher) GetLogsByFilters(ctx context.Context, crit filters.FilterCr
 	// Worker function
 	worker := func() {
 		defer wg.Done()
+		defer func() {
+			if r := recover(); r != nil {
+				err = fmt.Errorf("panic in GetLogsByFilters worker: %v", r)
+			}
+		}()
 		for height := range tasksChan {
 			if len(crit.Addresses) != 0 || len(crit.Topics) != 0 {
 				providerCtx := f.ctxProvider(height)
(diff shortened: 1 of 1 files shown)


philipsu522 (approved):

jiangjinyuan:
I got the issue with tag "v6.0.1-hotfix-rpc-3" too, is there a plan to fix it today?

evmrpc/filter.go
```diff
@@ -377,6 +385,11 @@ func (f *LogFetcher) GetLogsByFilters(ctx context.Context, crit filters.FilterCr
 		close(resultsChan) // Close the results channel after workers finish
 	}()
 
+	// Check err after all work is done
+	if err != nil {
+		return nil, 0, err
+	}
```
jewei1997: instead of just returning here if there's an error, would be better to just aggregate this error along with the other results.

evmrpc/filter.go
```diff
@@ -345,6 +346,11 @@ func (f *LogFetcher) GetLogsByFilters(ctx context.Context, crit filters.FilterCr
 	// Worker function
 	worker := func() {
 		defer wg.Done()
+		defer func() {
+			if r := recover(); r != nil {
+				err = fmt.Errorf("panic in GetLogsByFilters worker: %v", r)
+			}
```
jewei1997: same here. Right now if this worker looks at a bunch of heights and the logs on one of those heights is not retrievable, the only thing returned out be the err for that one height. Can we write this in a way such that the worker will do a best effort and return all the real results + errors for any panic-ed responses? Also would be great if we included the exact height on the error response.

evmrpc/filter.go
yzang2019: Fixed

evmrpc/filter.go
yzang2019: Fixed

evmrpc/filter.go
```diff
@@ -343,8 +344,14 @@ func (f *LogFetcher) GetLogsByFilters(ctx context.Context, crit filters.FilterCr
 	}()
 
 	// Worker function
+	var errorsList []error
 	worker := func() {
 		defer wg.Done()
+		defer func() {
+			if r := recover(); r != nil {
+				err = fmt.Errorf("unexpected panic caught in GetLogsByFilters worker: %v", r)
+			}
```
jewei1997, line 353: can we have this recover be at the per-height level and not the per-worker level? This way, a worker can still return results for other heights even if one height panics? I think the way its written right now, if a worker searches n heights and just one of them has a panic? It will not be able to return logs for the other non-panicking heights.

jewei1997 (approved):

evmrpc/filter.go
yzang2019: Discussed offline and resolved

PR #1994, opened 2024-12-17
Ct  transfer precompile

dssei:
CT transfer precompile.
- Note that this precompile/payload generation tooling covers case for transfer without auditors.
- It's not straightforward to implement precompile methods with optional params, so I think there will be a separate solidity method and precompile for case with auditors.
- The payload generation tooling for now just covers single case as well. At this point, I am not sure it makes sense to attempt to generalize one payload generation command to fit all ct cases (due to possible unnecessary complexity of implementation - but will keep this option open for future related PRs)
- unit test
- cli functional test


precompiles/confidentialtransfers/ct.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func (p PrecompileExecutor) accAddressFromArg(ctx sdk.Context, arg interface{}) (sdk.AccAddress, error) {
+	addr := arg.(common.Address)
+	if addr == (common.Address{}) {
+		return nil, errors.New("invalid addr")
+	}
+	seiAddr, associated := p.evmKeeper.GetSeiAddress(ctx, addr)
+	if !associated {
+		return nil, errors.New("cannot use an unassociated address as withdraw address")
```
mj850, line 195: as transfer address?

precompiles/confidentialtransfers/ct.go
```diff
@@ -0,0 +1,198 @@
...
+
+	var transferMessageProofs cttypes.TransferMsgProofs
+	err = transferMessageProofs.Unmarshal(args[9].([]byte))
+	if err != nil {
+		rerr = err
+		return
+	}
+
+	msg := &cttypes.MsgTransfer{
+		FromAddress:        fromAddr.String(),
```
mj850, line 161: Assuming these would still be sei1 addresses?

mj850: Not an issue actly just curious

x/evm/client/cli/query.go
```diff
@@ -461,3 +465,133 @@ func CmdQueryPointee() *cobra.Command {
...
+		remainingBalance,
+		transferProto.DecryptableBalance,
+		proofs)
+	if err != nil {
+		return err
+	}
+	return queryClientCtx.PrintString(hex.EncodeToString(bz))
+}
+
+func getCtAccount(queryClient cttypes.QueryClient, address, denom string) (*cttypes.Account, error) {
```
mj850, line 582: Not impt for this PR, but would be good to use this standardized helper funtions across all queries and txes that require querying the CtAccount

precompiles/confidentialtransfers/ct.go
dssei: Gonna fix in next PR

precompiles/confidentialtransfers/ct.go
dssei: Yeah, we will convert all 0x addresses to sei

x/evm/client/cli/query.go
dssei: Make sense to reduce duplication. Will try to address in next PR

PR #1995, opened 2024-12-18
RPC endpoints for excluding tracing failures

jewei1997:

- [x] add 3 new endpoints
- [x] add unit tests
- [x] name them all *ExcludeTracingFail for clarity
- [x] write a README about old synthetic endpoints + these new endpoints
- [x] test endpoints on RPC

A panic tx is a tx which fails tracing. This can be due to several issues such as nonce errors, insufficient funds, etc. Adding 3 new endpoints for excluding panic txs.
 - `sei_traceBlockByNumberExcludeTraceFail`
 - `sei_getBlockByNumberExcludeTraceFail` + `sei_getBlockByHashExcludeTraceFail`
 -  `sei_getTransactionByReceiptExcludeTraceFail`

Relies on https://github.com/sei-protocol/go-ethereum/pull/40

added unit tests in rpc layer + test on RPC node

### DIFF
--- a/evmrpc/block.go
+++ b/evmrpc/block.go
@@ -292,25 +292,25 @@ func EncodeTmBlock(
 				}
 				ethtx, _ := m.AsTransaction()
 				hash := ethtx.Hash()
+				if isPanicTx != nil {
+					isPanic, err := isPanicTx(ctx.Context(), hash)
+					if err != nil {
+						return nil, fmt.Errorf("failed to check if tx is panic tx: %w", err)
+					}
+					if isPanic {
+						continue
+					}
+				}
+				receipt, err := k.GetReceipt(ctx, hash)
+				if err != nil {
+					continue
+				}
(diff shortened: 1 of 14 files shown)


philipsu522 (approved):

evmrpc/block.go
```diff
@@ -28,6 +28,7 @@ import (
 )
 
 const ShellEVMTxType = math.MaxUint32
+const TestGenesisBlockHash = "0xF9D3845DF25B43B1C6926F3CEDA6845C17F5624E12212FD8847D0BA01DA1AB9E"
```
yzang2019: Why is this called TestGenesisBlockHash? Is this related to test at all?

yzang2019 (approved):

evmrpc/block.go
jewei1997: ah good catch. Not really, moved it back to where it was before and got rid of the "Test" name.

PR #1997, opened 2024-12-19
Transfer with auditors precompile

dssei:
Transfer with auditors precompile
- unit tests
- cli testing against local node


precompiles/confidentialtransfers/ct.go
```diff
@@ -192,7 +247,75 @@ func (p PrecompileExecutor) accAddressFromArg(ctx sdk.Context, arg interface{})
...
 	if !associated {
-		return nil, errors.New("cannot use an unassociated address as withdraw address")
+		return nil, errors.New("cannot use an unassociated address as transfer address")
 	}
 	return seiAddr, nil
 }
+
+func (p PrecompileExecutor) getAuditorsFromArg(ctx sdk.Context, arg interface{}) ([]*cttypes.Auditor, error) {
+	// we need to define an anonymous struct similar to types.EvmAuditor because the ABI returns an anonymous struct
+	res := arg.([]struct {
```
mj850: What happens if the arg passed is
1. An empty []
2. Doesn't fit the fields of the anonymous struct?

x/confidentialtransfers/types/transfer.go
```diff
@@ -49,6 +51,16 @@ type TransferAuditor struct {
 	TransferAmountHiEqualityProof *zkproofs.CiphertextCiphertextEqualityProof `json:"transfer_amount_hi_equality_proof"`
 }
 
+type EvmAuditor struct {
```
mj850, line 54: Is it a bit misleading to call this EvmAuditor? Given that the address is a Sei Address, there doesn't seem to be anything EVM about it.

Could we just call this Auditor? Or SerializedAuditor or something

mj850: Oh maybe I misunderstood, it is an EVM address

x/evm/client/cli/query.go
```diff
@@ -572,26 +610,43 @@ func queryCtTransferPayload(cmd *cobra.Command, args []string) error {
...
+		auditorsProto := transferProto.Auditors
+		for _, auditorProto := range auditorsProto {
+			encryptedTransferAmountLo, _ := auditorProto.EncryptedTransferAmountLo.Marshal()
+			encryptedTransferAmountHi, _ := auditorProto.EncryptedTransferAmountHi.Marshal()
+			transferAmountLoValidityProof, _ := auditorProto.TransferAmountLoValidityProof.Marshal()
+			transferAmountHiValidityProof, _ := auditorProto.TransferAmountHiValidityProof.Marshal()
+			transferAmountLoEqualityProof, _ := auditorProto.TransferAmountLoEqualityProof.Marshal()
+			transferAmountHiEqualityProof, _ := auditorProto.TransferAmountHiEqualityProof.Marshal()
+			// check if the auditor address does not exist in the map
+			if auditorEvmAddress, exists := seiToEvmAddressMap[auditorProto.AuditorAddress]; !exists {
```
mj850, line 627: In what scenario would this happen? Isn't every auditor in the list `auditors` from the list of auditors passed in?

mj850 (commented):
One tradeoff we seem to be making:
In our EVM precompile, we expect all the addresses passed in to be EVM addresses - this makes sense for cleanliness since it is technically an 'EVM side' thing.

However, this puts an artificial limitation on the precompile, since all parties involved need to be associated before they can use the precompile (even though there's no functional reason why this should be the case)

Wonder if it would make sense or be possible to enable users to pass in Sei addresses (especially for auditors, recipients). Doing that might enable better usability but might be more confusing to maintain.

mj850 (approved):

dssei:
> One tradeoff we seem to be making: In our EVM precompile, we expect all the addresses passed in to be EVM addresses - this makes sense for cleanliness since it is technically an 'EVM side' thing.
> 
> However, this puts an artificial limitation on the precompile, since all parties involved need to be associated before they can use the precompile (even though there's no functional reason why this should be the case)
> 
> Wonder if it would make sense or be possible to enable users to pass in Sei addresses (especially for auditors, recipients). Doing that might enable better usability but might be more confusing to maintain.

Supporting both sei and EVM addresses is possible, and I think it we have done this for other precompiles, although use cases were not clear. As you mentioned, if you are working on EVM side, what would be the use case you need to work with Sei native address?

x/evm/client/cli/query.go
dssei: I think I added this check when there was a bug and the other part of the code, to make it clear why failure in map lookup occurred. And kept it since. Should not happen ideally

x/confidentialtransfers/types/transfer.go
dssei: Yeah, what makes it EVM specific is EVM address and the fact that this is used for EVM/precompile use case. Also `Audit` is already declared in the package

mj850:
I think the main thing is being able to send to users or listed auditors who have not associated their accounts

precompiles/confidentialtransfers/ct.go
dssei: Good points. 
 
> 1. An empty []

We endUp with no error, but with an empty array, which is not what we want for this use case. I will add a check


> 2. Doesn't fit the fields of the anonymous struct?


Format is safeguarded by the ABI and if one will attempt to pack something wrong then an error will be returned like:
```
&errors.errorString{s:"abi: cannot use confidentialtransfers.WrongEvmAuditorFormat as type [0]slice as argument"}
```
if someone, manages to bypass that (e.g. modifies calldata afterwards, and call with a wrong field), the conversion logic will panic, the panic will recover when `transferWithAuditors` concludes and will produce error like 
```
 "interface conversion: interface {} is []struct { AuditorAddress common.Address \"json:\\\"auditorAddress\\\"\"; EncryptedTransferAmountLo []uint8 \"json ....
```


One improvement could be to handle panic at the function itself and add more info like `error parsing auditor array: <err>`

PR #2006, opened 2024-12-28
Fix volatile eth_gasPrice

jewei1997:

Our current eth_gasPrice uses the reward of the 50th percentile tx in the latest block as part of the calculation. This works decently well if the previous block is congested. However, if the previous block had only 1 tx where the user dramatically overpriced their tx, it will return a very high reward. This often causes our eth_gasPrice endpoint to jump around erratically.

This change uses the previous block's total gas used to determine if the block is congested. If the previous block is congested, we will use the previous logic of 50% of reward. Otherwise, eth_gasPrice now pulls the latest base fee and increases it by 10% to ensure the user's tx can be included in a timely manner without dramatically overpaying.

This also makes a similar adjustment to `eth_maxPriorityFeePerGas` where if the chain is not congested, a default priority fee of 1gwei is returned to the user.

existing unit testing + manually patched it onto an RPC node.

### DIFF
--- a/evmrpc/info.go
+++ b/evmrpc/info.go
@@ -76,7 +76,7 @@ func (i *InfoAPI) GasPrice(ctx context.Context) (result *hexutil.Big, returnErr 
 	baseFee := i.keeper.GetDynamicBaseFeePerGas(i.ctxProvider(LatestCtxHeight)).TruncateInt().BigInt()
 	// increase base fee by 10% to get the gas price to get a tx included in a timely manner
 	// legacy txs will use this gas price to get included
-	// eip-1669 txs will use this as the max fee per gas to get included
+	// eip-1559 txs will use this as the max fee per gas to get included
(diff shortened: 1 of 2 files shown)


evmrpc/info.go
```diff
@@ -229,6 +248,29 @@ func (i *InfoAPI) getRewards(block *coretypes.ResultBlock, baseFee *big.Int, rew
...
+	}
+	totalEVMGasUsed := uint64(0)
+	for _, txbz := range block.Block.Txs {
+		ethtx := getEthTxForTxBz(txbz, i.txDecoder)
+		if ethtx == nil {
+			// not evm tx
+			continue
+		}
+		// okay to get from latest since receipt is immutable
+		receipt, err := i.keeper.GetReceipt(i.ctxProvider(LatestCtxHeight), ethtx.Hash())
```
codchen, line 277: best to check if `receipt.BlockHeight` matches height in `block`, and ignore if mismatch

codchen (approved):

PR #2013, opened 2024-12-30
fix race condition

codchen:

### DIFF
--- a/evmrpc/filter.go
+++ b/evmrpc/filter.go
@@ -349,6 +349,8 @@ func (f *LogFetcher) GetLogsByFilters(ctx context.Context, crit filters.FilterCr
 		defer wg.Done()
 		defer func() {
 			if r := recover(); r != nil {
+				mu.Lock()
+				defer mu.Unlock()
 				err = fmt.Errorf("unexpected panic caught in GetLogsByFilters worker: %v", r)
 			}
 		}()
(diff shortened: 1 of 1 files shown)


evmrpc/filter.go
```diff
@@ -349,6 +349,8 @@ func (f *LogFetcher) GetLogsByFilters(ctx context.Context, crit filters.FilterCr
 		defer wg.Done()
 		defer func() {
 			if r := recover(); r != nil {
+				mu.Lock()
```
yzang2019, line 352: Is this being hit at all? I think recover() should never be hit in this logic, and it shouldn't be causing any dead lock issue from my understanding. That being said, adding the lock wouldn't hurt as well

evmrpc/filter.go
```diff
@@ -387,16 +389,16 @@ func (f *LogFetcher) GetLogsByFilters(ctx context.Context, crit filters.FilterCr
...
-	if len(errorsList) > 0 {
-		err = errors.Join(errorsList...)
-	}
-
 	// Aggregate results into the final slice
 	for result := range resultsChan {
 		res = append(res, result)
 	}
 
+	// Check err after all work is done
```
yzang2019, line 397: This ordering adjustment make sense to me, but I wonder why it would lead to dead lock or stuck issue

yzang2019 (approved):

evmrpc/filter.go
codchen: I think it'd cause the `errorsList = append(errorsList, berr)` above to run into undefined behavior which then causes some weirdness

yzang2019: make sense to me

PR #2015, opened 2024-12-30
[wip] Base Fee Off By One

jewei1997:

### DIFF
--- a/x/evm/keeper/fee.go
+++ b/x/evm/keeper/fee.go
@@ -85,3 +85,50 @@ func (k *Keeper) SetDynamicBaseFeePerGas(ctx sdk.Context, baseFeePerGas sdk.Dec)
 	}
 	store.Set(types.BaseFeePerGasPrefix, bz)
 }
+
+func (k *Keeper) SetPrevBlockGasUsed(ctx sdk.Context, blockGasUsed uint64) {
+	store := ctx.KVStore(k.storeKey)
+	gasUsed := sdk.NewInt(int64(blockGasUsed))
+	bz, err := gasUsed.MarshalJSON()
+	if err != nil {
+		panic(err)
+	}
+	store.Set(types.PrevBlockGasUsedPrefix, bz)
+}
+
+func (k *Keeper) GetPrevBlockGasUsed(ctx sdk.Context) sdk.Int {
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get(types.PrevBlockGasUsedPrefix)
+	if bz == nil {
+		return sdk.NewInt(0)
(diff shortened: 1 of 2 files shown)


contracts/test/EVMCompatabilityTest.js
```diff
@@ -499,13 +498,53 @@ describe("EVM Test", function () {
...
         const gp = ethers.parseUnits("100", "gwei");
+        const oneGwei = ethers.parseUnits("1", "gwei");
 
         const testCases = [
           ["No truncation from max priority fee", gp, gp],
           ["With truncation from max priority fee", gp, highgp],
           ["With complete truncation from max priority fee", zero, highgp]
         ];
 
+        it("Check base fee on the right block", async function () {
```
philipsu522, line 509: nice 👍

x/evm/keeper/fee.go
```diff
@@ -85,3 +87,30 @@ func (k *Keeper) SetDynamicBaseFeePerGas(ctx sdk.Context, baseFeePerGas sdk.Dec)
 	}
 	store.Set(types.BaseFeePerGasPrefix, bz)
 }
+
+func (k *Keeper) SetPrevBlockBaseFeePerGas(ctx sdk.Context, baseFeePerGas sdk.Dec) {
+	store := ctx.KVStore(k.storeKey)
+	bz, err := baseFeePerGas.MarshalJSON()
+	if err != nil {
+		panic(err)
```
philipsu522, line 95: where does this panic end up getting thrown up to?

x/evm/migrations/migrate_base_fee_off_by_one.go
```diff
@@ -0,0 +1,12 @@
+package migrations
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/sei-protocol/sei-chain/x/evm/keeper"
+)
+
+func MigrateBaseFeeOffByOne(ctx sdk.Context, k *keeper.Keeper) error {
```
yzang2019, line 8: Where is this migration being registered?

x/evm/keeper/fee.go
jewei1997: this would get thrown in `EndBlock` which would halt the chain if thrown, but technically shouldn't get thrown unless something has gone really wrong.

x/evm/migrations/migrate_base_fee_off_by_one.go
jewei1997: registered it, good catch!

codchen (approved):
nice testing

PR #2017, opened 2025-01-02
Ct init account precompile

dssei:
CT init account precompile

- unit tests


precompiles/confidentialtransfers/CT.sol
```diff
@@ -8,6 +8,17 @@ ICT constant CT_CONTRACT = ICT(
 
 interface ICT {
     // Transactions
+    function initializeAccount(
+        string fromAddress,
```
mj850, line 12: Curious if we need the fromAddress here or if we can just use the sender address?

I know in bank sends it exists because you can call send on an approved token, but not sure if it's necessary here

precompiles/confidentialtransfers/ct.go
```diff
@@ -240,16 +249,34 @@ func (p PrecompileExecutor) getTransferMessageFromArgs(ctx sdk.Context, args []i
...
+		if associated {
+			return seiAddr, evmAddr, nil
+		} else {
+			return nil, common.Address{}, fmt.Errorf("address %s is not associated", addr)
+		}
+	}
+	if seiAddress, err := sdk.AccAddressFromBech32(addr); err != nil {
+		return nil, common.Address{}, fmt.Errorf("invalid address %s: %w", addr, err)
+	} else {
+		evmAddr, _ := p.evmKeeper.GetEVMAddress(ctx, seiAddress)
```
mj850: Is there a case here where the sei address has not been associated and this throws an error?

precompiles/confidentialtransfers/ct.go
```diff
@@ -330,3 +357,96 @@ func (p PrecompileExecutor) getAuditorsFromArg(ctx sdk.Context, arg interface{})
...
+			return
+		}
+	}()
+
+	if err := pcommon.ValidateArgsLength(args, 8); err != nil {
+		rerr = err
+		return
+	}
+
+	seiAddr, evmAddr, err := p.getValidAddressesFromString(ctx, args[0].(string))
```
mj850, line 392: Can we use caller here instead of the fromAddress passed in?

precompiles/confidentialtransfers/ct.go
```diff
@@ -277,7 +304,7 @@ func (p PrecompileExecutor) getAuditorsFromArg(ctx sdk.Context, arg interface{})
 
 	auditors := make([]*cttypes.Auditor, 0)
 	for _, auditor := range evmAuditors {
-		auditorAddr, err := p.accAddressFromArg(ctx, auditor.AuditorAddress)
+		auditorAddr, _, err := p.accAddressesFromArg(ctx, auditor.AuditorAddress)
```
mj850, line 323: Hmm the main goal of supporting both evm and sei addresses is that the auditors/recipients do not need to have an EVM address in order for the transfer to succeed. I'm not sure this change here achieves that

precompiles/confidentialtransfers/ct_test.go
```diff
@@ -623,3 +636,221 @@ func setUpCtAccount(k *evmkeeper.Keeper, ctx sdk.Context, testDenom string) (sdk
...
+			args: args{
+				setUp: func(in inputs) inputs {
+					in.UserAddress = ""
+					return in
+				}},
+			wantErr:    true,
+			wantErrMsg: "invalid address : empty address string is not allowed",
+		},
+		{
+			name: "precompile should return error if caller is not the same as the user",
```
mj850, line 750: We could also add test cases for interop here to show behavior for
- when the user is not associated (only EVM addresses)

precompiles/confidentialtransfers/ct.go
dssei: Double-checked this, and in theory it's possible, but in practice, when EVM transaction is initiated the address will forcefully get associated (in ante handler) before we hit this method. I will add check for safety/consistency anyway.

precompiles/confidentialtransfers/ct.go
dssei: To be addressed in next PR

precompiles/confidentialtransfers/ct_test.go
dssei: Added new tests

precompiles/confidentialtransfers/ct.go
dssei: As discussed offline, we probably could, but we would still not be able to guarantee that the caller created calldata, so we would defer the validation to the msg server.

mj850 (approved):

PR #2021, opened 2025-01-06
Support Sei addresses in transfer precompiles

dssei:
Support Sei addresses in transfer precompiles

- unit tests
- local testing


precompiles/confidentialtransfers/ct.go
```diff
@@ -320,7 +335,7 @@ func (p PrecompileExecutor) getAuditorsFromArg(ctx sdk.Context, arg interface{})
 
 	auditors := make([]*cttypes.Auditor, 0)
 	for _, auditor := range evmAuditors {
-		auditorAddr, _, err := p.accAddressesFromArg(ctx, auditor.AuditorAddress)
+		auditorAddr, err := p.getValidSeiAddressFromString(ctx, auditor.AuditorAddress)
```
mj850, line 338: Was probably in the scope of the previous PR, but do we need to do any validation on the contents of the Auditor objects?

mj850: Sorry can ignore: Just expanded and see that validations are performed below

mj850 (approved):

PR #2023, opened 2025-01-08
Ct deposit precompile

dssei:
Ct deposit precompile
Note: for this precompile regular `evm payload` works

- unit testing
- local testing


precompiles/confidentialtransfers/ct_test.go
```diff
@@ -917,3 +917,176 @@ func TestPrecompileInitializeAccount_Execute(t *testing.T) {
...
+			}
+
+			inputArgs, err := DepositMethod.Inputs.Pack(
+				in.UserAddress,
+				in.Denom,
+				in.Amount)
+
+			require.Nil(t, err)
+
+			resp, remainingGas, err := p.RunAndCalculateGas(
```
mj850, line 1071: Do we need any functional tests to ensure that
- CT Account received the new funds
- Funds were deducted from the user

precompiles/confidentialtransfers/ct.go
```diff
@@ -481,3 +489,57 @@ func (p PrecompileExecutor) initializeAccount(ctx sdk.Context, method *abi.Metho
...
+		return
+	}
+
+	denom := args[1].(string)
+	if denom == "" {
+		rerr = errors.New("invalid denom")
+		return
+	}
+
+	amount, ok := args[2].(uint64)
```
mj850, line 525: Not sure if this is an issue, but specifically for usei it should be made clear that this amount is 6 decimal instead of 19 decimal

mj850 (approved):
Approved with 2 minor comments

precompiles/confidentialtransfers/ct_test.go
dssei: Yeah, good point, I think we need to add tests to integration suit. Will add in subsequent PRs

precompiles/confidentialtransfers/ct.go
dssei: Good point, what would you suggest we do to clarify?

mj850: Hmm not very sure, probably just add a comment in the code here (and in the .sol) so we don't get confused in the meantime?

PR #2026, opened 2025-01-08
SEI-8802

pompon0:
https://linear.app/seilabs/issue/SEI-8802/add-sei-traceblockbyhashexcludetracefail-endpoint

Also fixed a race condition which was setting a result error from a background goroutine.


jewei1997 (approved):
great job Greg!

PR #2027, opened 2025-01-08
fix: eth_subscribe with geth open ended range

blindchaser:
- convert "fromBlock":"0x0","toBlock":"latest" to default subscription behavior


evmrpc/subscribe.go
```diff
@@ -147,6 +147,20 @@ func (a *SubscriptionAPI) Logs(ctx context.Context, filter *filters.FilterCriter
 	if filter == nil {
 		filter = &filters.FilterCriteria{}
 	}
+
+	// when fromBlock is 0 and toBlock is latest, adjust the filter
+	// to unbounded filter
+	if filter.FromBlock != nil && filter.FromBlock.Int64() == 0 &&
+		filter.ToBlock != nil && filter.ToBlock.Int64() < 0 {
+		unboundedFilter := &filters.FilterCriteria{
+			FromBlock: filter.FromBlock, // keep "0x0" to fetch historical logs
```
bruce-riley: The desired behavior in this scenario is not to fetch historic logs, just stream any new logs.

cordt-sei: Just looking through the relevant [geth docs](https://geth.ethereum.org/docs/interacting-with-geth/rpc/pubsub#logs), `toBlock` and `fromBlock` are not even valid parameters for this method in any context so it's very confusing as to why they are being passed there in the first place.

bruce-riley:
I created a simple test app that uses the `go-ethereum` library to listen for events from the Wormhole core contract without needing to run the Wormhole guardian. It could be used to verify this fix. See the PR description for how to use it.
https://github.com/wormhole-foundation/wormhole/pull/4214

bruce-riley:
I just confirmed that the Wormhole guardian is now able to see log events using go-ethereum with this fix. Please let us know when this is part of an official release. Thanks!

cordt-sei (commented):
Tested with the aforementioned testing application and confirmed log event receipt.

jewei1997 (approved):

philipsu522 (approved):

PR #2032, opened 2025-01-10
Make IBC precompile nonpayable

dssei:
Make IBC precompile nonpayable. It was defined as such in solidity, but not in ABI + added enforcement in code.

- unit test


cordt-sei (approved):
Tested on testnet and local net, ibc transfer works.

jewei1997 (approved):

PR #2043, opened 2025-01-16
Query owner on ERC-721 and ERC-1155 pointers

jewei1997:
Add an `owner()` function that adheres to the [Ownable interface](https://github.com/nibbstack/erc721/blob/master/src/contracts/ownership/ownable.sol). Note: we are only adding the ability to query the owner, not the ability to change the owner. [CW721_ownable](https://docs.rs/cw-ownable/latest/cw_ownable/) has the concept of the new owner accepting the new ownership for the ownership transfer to go through which ERC Ownable does not have which makes it tricky to do the write functions of actually updating ownership.

integration test


codchen (commented):
do we want to add migration handler to update existing deployed pointers to use this new code?

.github/workflows/unit_tests.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           NUM_SPLIT=20
           make test-group-${{matrix.part}} NUM_SPLIT=20
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
```
dssei, line 46: `actions/download-artifact: v3` below needs a bump too

dvli2007:
The CW1155 contract follows the same ownership pattern as CW721s. We should add that functionality into the `CW1155ERC1155Pointer.sol` as well

x/evm/module.go
```diff
@@ -270,7 +277,7 @@ func (am AppModule) ExportGenesisStream(ctx sdk.Context, cdc codec.JSONCodec) <-
 }
 
 // ConsensusVersion implements ConsensusVersion.
-func (AppModule) ConsensusVersion() uint64 { return 16 }
```
jewei1997, line 273: I think this was previously off by one

codchen (approved):

PR #2048, opened 2025-01-21
Add logic to remove a small number of tx hashes each block

codchen:
Remove 100 tx hashes every block until all tx hashes are deleted.

unit test


x/evm/module.go
```diff
@@ -298,6 +298,9 @@ func (am AppModule) BeginBlock(ctx sdk.Context, _ abci.RequestBeginBlock) {
 // EndBlock executes all ABCI EndBlock logic respective to the evm module. It
 // returns no validator updates.
 func (am AppModule) EndBlock(ctx sdk.Context, req abci.RequestEndBlock) []abci.ValidatorUpdate {
+	// TODO: remove after all TxHashes have been removed
```
philipsu522, line 308: could we do this every 10 blocks? just so we spread it out and it doesn't look like block times are just degrading?

yzang2019 (approved):

philipsu522 (approved):

PR #2051, opened 2025-01-22
add IBC support for 0x addresses

dssei:
add IBC support for 0x addresses.
This implementation will parse the address string and if its a 0x address, will return either associated or casted Sei Address.

- unit tests


x/evm/keeper/address.go
```diff
@@ -84,3 +84,23 @@ func (k *Keeper) CanAddressReceive(ctx sdk.Context, addr sdk.AccAddress) bool {
...
+}
+
+func NewEvmAddressHandler(evmKeeper Keeper) EvmAddressHandler {
+	return EvmAddressHandler{evmKeeper: evmKeeper}
+}
+
+func (h EvmAddressHandler) GetSeiAddressFromString(ctx sdk.Context, address string) (sdk.AccAddress, error) {
+	if common.IsHexAddress(address) {
+		parsedAddress := common.HexToAddress(address)
+		return h.evmKeeper.GetSeiAddressOrDefault(ctx, parsedAddress), nil
```
dssei, line 99: @codchen is assumption correct here that if address is not associated yet, the funds would go to an "escrow" account and then will be migrated after association?

dssei: Upd: Was able to test scenario e2e and it works

philipsu522: yes this is correct

precompiles/ibc/ibc.go
```diff
@@ -352,16 +351,8 @@ func (p PrecompileExecutor) validateCommonArgs(ctx sdk.Context, args []interface
...
-	}
-	_, bz, err := bech32.DecodeAndConvert(receiverAddressString)
-	if err != nil {
-		return nil, err
-	}
-	err = sdk.VerifyAddressFormat(bz)
-	if err != nil {
-		return nil, err
+	if !ok || receiverAddressString == "" {
+		return nil, errors.New("receiverAddress is not a string or empty")
```
philipsu522, line 355: do we want to add checks that the address is valid? i.e.
1. if sei, it is proper bech32
2. if evm, it is a valid address

philipsu522 (approved):

precompiles/ibc/ibc.go
dssei: We did have the bech32 check before, but i was too restrictive. In fact this address is propagated as string all the way up the stack without any checks. I guess the assumption is that it could any type of address, not only bech32 or EVM.

PR #2054, opened 2025-01-25
extractAsBytesFromArray method for JSON precompile

dssei:
extractAsBytesFromArray method for JSON precompile to allow elements extraction from JSON arrays.

This additional method is needed for cases like 
```
{
  "data": {
    "exchange_rates": [
      [
        1727762390,
        "1.033207463698531844"
      ],
      [
        1727545459,
        "1.032887092691178063"
      ]
    ],
    "apr": "0.000101145698240442"
  }
}
```
We do have methods allowing extracting data by key or array(list) by key, but no way to extract element from array.
E.g. data.exchange_rates[0][0]

- unit test


cordt-sei (approved):

dssei:
> I think you might need migration handlers to properly roll this out

I see. Thought migration handlers are for modules related changes. Looked into similar PRs but have not seen any examples of handlers changes. Do you have one by any chance?

jewei1997:
yeah I think you can look at this PR as an example. Basically you need migration handlers to update the precompile contracts https://github.com/sei-protocol/sei-chain/pull/2043

jewei1997 (approved):

PR #2073, opened 2025-02-13
RPC simulation with gas used estimate tagging

jewei1997:

To fix the high gas wanted txs issue, this PR implements simulation on the RPC layer to tag the tx with its estimated gas used when entering the mempool. Then, when the proposer builds a block, the estimated gas used is used instead of gas wanted. This allows us to pack blocks more tight/efficiently. 

Includes these other 2 PRs:
 * sei-cosmos: https://github.com/sei-protocol/sei-cosmos/pull/566
 * sei-tendermint: https://github.com/sei-protocol/sei-tendermint/pull/263

unit testing, load testing


x/evm/ante/gas.go
```diff
@@ -6,23 +6,33 @@ import (
...
 	}
 
 	adjustedGasLimit := gl.evmKeeper.GetPriorityNormalizer(ctx).MulInt64(int64(txData.GetGas()))
-	ctx = ctx.WithGasMeter(sdk.NewGasMeterWithMultiplier(ctx, adjustedGasLimit.TruncateInt().Uint64()))
+	gasMeter := sdk.NewGasMeterWithMultiplier(ctx, adjustedGasLimit.TruncateInt().Uint64())
+	ctx = ctx.WithGasMeter(gasMeter)
+	if tx.GetGasEstimate() >= MinGasEVMTx {
+		ctx = ctx.WithGasEstimate(tx.GetGasEstimate())
+	} else {
+		ctx = ctx.WithGasEstimate(gasMeter.GasConsumed())
```
codchen: should it be with GasLimit here? GasConsumed would only include gas used in ante handler up to this point, not any from the actual execution

app/app.go
```diff
@@ -1875,8 +1879,11 @@ func RegisterSwaggerAPI(rtr *mux.Router) {
 	rtr.PathPrefix("/swagger/").Handler(http.StripPrefix("/swagger/", staticServer))
 }
 
-func (app *App) checkTotalBlockGasWanted(ctx sdk.Context, txs [][]byte) bool {
-	totalGasWanted := uint64(0)
+// checkTotalBlockGas checks that the block gas limit is not exceeded by our best estimate of
+// the total gas by the txs in the block. The gas of a tx is either the gas estimate if it's an EVM tx,
+// or the gas wanted if it's a Cosmos tx.
+func (app *App) checkTotalBlockGas(ctx sdk.Context, txs [][]byte) bool {
```
codchen, line 1885: given the complexity now for this function it might be helpful to add unit test to ensure parity between this and `ReapMaxBytesMaxGas`

jewei1997: added some tests in `app_test.go`

x/evm/ante/gas.go
jewei1997: yeah good catch

app/app.go
jewei1997: have these tests. https://github.com/sei-protocol/sei-chain/blob/78c4e7581a4b58edcdac449107694ef3e1685f6a/app/app_test.go#L346-L376

codchen (approved):

PR #2074, opened 2025-02-14
geth upgrade test

codchen:
wip

wip


jewei1997:
can we also add an integration test that tests the new tx type introduced in EIP-7702?

x/evm/types/ethtx/auth_list.go
```diff
@@ -0,0 +1,56 @@
...
+			V:       []byte{auth.V},
+			R:       (auth.R).Bytes(),
+			S:       (auth.S).Bytes(),
+		})
+	}
+
+	return al
+}
+
+func (al AuthList) ToEthAuthList() *[]ethtypes.SetCodeAuthorization {
```
dssei, line 33: I think this is missing this change we have done before https://github.com/sei-protocol/sei-chain/commit/493a2d6d9356db565d072b9883838d6af719538e#diff-809a5e3c9276c9cfc14c2754292a3559223eb2fc9ddd93c2d2cb44f9a12a5824R34

codchen: good catch

PR #2088, opened 2025-03-03
Refactor RPC tests

codchen:
It's becoming difficult to add new RPC tests due to the way old tests are set up. This PR creates a new framework for adding RPC unit tests where it allows spinning up independent server threads and more authentic data mocking.

One thing to call out is that this refactor requires our custom precompiles to be loaded dynamically rather than through a global variable (so that it's possible to have independent concurrent servers across multiple tests)

unit test

### DIFF
--- a/evmrpc/tests/utils.go
+++ b/evmrpc/tests/utils.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net/http"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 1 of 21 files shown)


evmrpc/tests/utils.go
```diff
@@ -0,0 +1,209 @@
...
+		Time:   time.Now(),
+	})
+	if err != nil {
+		panic(err)
+	}
+	ctx := a.GetContextForDeliverTx(nil)
+	for _, i := range initializer {
+		i(ctx, a)
+	}
+	a.Commit(context.Background())
```
jewei1997, line 67: nit: check error

jewei1997 (approved):

PR #2097, opened 2025-03-09
Update docker_publish.yml

cordt-sei:
Adds a step to docker_publish workflow to make image public after upload.


Ajes1337:
wen, plz gib ser

PR #2114, opened 2025-03-25
Add capability for MEV plugin to accept submission via RPC

codchen:
This PR adds an additional functionality to the MEV plugin interface to allow custom logic for MEV submission via rpc. A new RPC endpoint `mev_submission` is added (which will return error if called on a node without any MEV plugins). An example implementation of the new interface method can be found in https://github.com/sei-protocol/mev-demo/blob/main/plugin.go#L28

localnet with mev-demo
Screenshot 2025-03-25 at 11 04 43 PM
Screenshot 2025-03-25 at 11 04 54 PM

### DIFF
--- a/evmrpc/mev.go
+++ b/evmrpc/mev.go
@@ -1,0 +1,24 @@
+package evmrpc
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+
+	"github.com/sei-protocol/sei-chain/mev"
+)
+
+type MevAPI struct {
+	handler mev.MEVHandler
+}
+
+func NewMevAPI(handler mev.MEVHandler) *MevAPI {
+	return &MevAPI{handler: handler}
+}
+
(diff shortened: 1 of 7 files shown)


evmrpc/mev.go
```diff
@@ -0,0 +1,24 @@
...
+	handler mev.MEVHandler
+}
+
+func NewMevAPI(handler mev.MEVHandler) *MevAPI {
+	return &MevAPI{handler: handler}
+}
+
+func (i *MevAPI) Submission(ctx context.Context, req json.RawMessage) (res json.RawMessage, err error) {
+	if i.handler == nil {
+		return nil, errors.New("this node does not support MEV submission")
```
jewei1997, line 21: nit: can call this "bundle submission". Bundles may or may not contain MEV txs.

jewei1997 (approved):
overall LGTM!

PR #2118, opened 2025-04-03
Add back legacy precompile versions

codchen:
Add back all historical versions of precompile code since mainnet launch (v5.0.1). In subsequent PRs, we will add logics to route tracing requests for historical transactions to the correct historical version of precompile code.

no logic change; pure addition


philipsu522 (approved):
is there an easy way to test this? or do we want to have partner teams test it?

codchen:
> is there an easy way to test this? or do we want to have partner teams test it?

we can test it by patching it on a pacific RPC node

PR #2119, opened 2025-04-03
P256 precompile

dssei:
This is implementation of  https://github.com/ethereum/RIPs/blob/master/RIPS/rip-7212.md

A precompiled contract that performs signature verifications in the `secp256r1` or `P-256` elliptic curve by given parameters of message hash, r and s components of the signature, x and y coordinates of the public key. 

This enables passkeys based use cases as `secp256r1` curve is widely used and supported in many modern devices such as Apple’s Secure Enclave, Webauthn, Android Keychain which proves the user adoption. 

Precompile implementation of “secp256r1” signature verification makes it more gas efficient (up to 60x) than Solidity based implementations. 

This precompile can be used in other contracts like

```
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.23;

import "./ECDSA.sol";
import "./P256Verify.sol";

//( @title P256
/// @author klkvr <https://github.com/klkvr>
/// @author jxom <https://github.com/jxom>
/// @notice Wrapper function to abstract low level details of call to the P256
///         signature verification precompile as defined in EIP-7212, see
///         <https://eips.ethereum.org/EIPS/eip-7212>.
library P256 {

    /// @notice P256VERIFY operation
    /// @param digest 32 bytes of the signed data hash
    /// @param signature Signature of the signer
    /// @param publicKey Public key of the signer
    /// @return success Represents if the operation was successful
    function verify(bytes32 digest, ECDSA.Signature memory signature, ECDSA.PublicKey memory publicKey)
        internal
        view
        returns (bool)
    {

        bytes memory input = abi.encode(digest, signature.r, signature.s, publicKey.x, publicKey.y);
        bytes memory output =  P256VERIFY_CONTRACT.verify(input);
        bool success = output.length == 32 && output[31] == 0x01;

        return success;
    }
}
```

- Unit testing
- functional testing with actual contracts deployed


codchen (approved):
you'd need to add it to `precompiles/setup.go`

PR #2138, opened 2025-04-22
Add ClaimSpecific for CW20/CW721

codchen:
Add a new `ClaimSpecific` function on `solo` precompile which allows users to claim all assets under the specified CW20 or CW721 contract. Similar to `Claim`, the claimee would need to sign a Cosmos-style `MsgClaimSpecific`, pass it to the claimer, and the claimer wrap it in an EVM transaction towards the `solo` precompile

unit tests


jewei1997 (approved):
LGTM

PR #2151, opened 2025-05-13
Fix EVM RPC denylist config

yzang2019:
Currently we are not passing the denylist config correctly to evmrpc server.


evmrpc/rpcstack.go
```diff
@@ -299,6 +299,7 @@ func (h *HTTPServer) EnableRPC(apis []rpc.API, config HTTPConfig) error {
 	if err := RegisterApis(h.log, apis, config.Modules, srv); err != nil {
 		return err
 	}
+	h.log.Info(fmt.Sprintf("[Debug] Registering deny list for evm rpc:%v", config.DenyList))
```
philipsu522, line 302: do we want to leave this as debug or just make it a regular log?

jewei1997 (approved):
thanks yiming!

evmrpc/rpcstack.go
yzang2019: Should be fine to leave it since it's an one-off log during startup

PR #2160, opened 2025-05-22
Remove Evmrpc Filter Panic

Kbhat1:
- Removes panic in evmrpc filter method
- Surfaces first error


evmrpc/filter.go
```diff
@@ -431,11 +433,29 @@ func (f *LogFetcher) fetchBlocksByCrit(ctx context.Context, crit filters.FilterC
...
 	}
+	go func() {
+		wg.Wait()
+		close(res)
+	}()
+
+	select {
+	case err := <-errChan:
+		// Drain runner.Queue and wait for all goroutines to finish
+		close(runner.Queue)
```
jewei1997: do we also want to close runner.Queue in the non-error case?

jewei1997 (approved):

evmrpc/filter.go
Kbhat1: moved out of the non err case

PR #2179, opened 2025-05-30
Use H+1 oracle for state during tracing

codchen:
Replay oracle txs during tracing is slow. Instead we can use oracle state at H+1 for oracle state at H, given that only oracle txs can change oracle states.

tested on archive nodes

### DIFF
--- a/utils/prioritized_txs.go
+++ b/utils/prioritized_txs.go
@@ -11,6 +11,8 @@ func IsTxPrioritized(tx sdk.Tx) bool {
 		switch msg.(type) {
 		case *oracletypes.MsgAggregateExchangeRateVote:
 			continue
+		case *oracletypes.MsgDelegateFeedConsent:
+			continue
 		default:
 			return false
 		}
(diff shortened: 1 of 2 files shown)


evmrpc/simulate.go
```diff
@@ -273,6 +273,18 @@ func (b Backend) BlockByNumber(ctx context.Context, bn rpc.BlockNumber) (*ethtyp
 		if err != nil {
 			return nil, nil, err
 		}
+		isPrioritized := utils.IsTxPrioritized(decoded)
+		if isPrioritized {
+			continue
+		}
+		// shouldRunMidBlock := false
```
yzang2019: Shall we just remove these code instead of commenting out?

app/app.go
```diff
@@ -1438,19 +1438,7 @@ func (app *App) PartitionPrioritizedTxs(_ sdk.Context, txs [][]byte, typedTxs []
...
-			switch msg.(type) {
-			case *oracletypes.MsgAggregateExchangeRateVote:
-				prioritized = true
-			default:
-				prioritized = false
-				break msgLoop
-			}
-		}
-		if prioritized {
+		if utils.IsTxPrioritized(typedTxs[idx]) {
```
udpatil, line 1442: because it's possible for oracle txs to end up non-prioritized (albeit very unlikely), if for example they are bundled with other cosmos txs (what is the effect if one gets executed outside of the prioritized portion?

udpatil (approved):

PR #2186, opened 2025-06-03
Disable CW->ERC Register Pointer

jewei1997:

- [x] Add a `DisableRegisterPointer` bool param to the param store
- [x] Add migration handlers for `DisableRegisterPointer` to set to true.
- [x] Add logic to `RegisterPointer` to return appropriate error if `DisableRegisterPointer` is true. Old logic will still run if it is false.

- [x] unit test checking that `RegisterPointer` fails when `DisableRegisterPointer` is true
- [x] manually test upgrade (pointer creation works before upgrade, fails after)
- [x] test upgrade on loadtest cluster, make sure all functionality works except NEW CW->ERC pointer creation (existing CW->ERC pointers should still work)


codchen (commented):
let's also add a `GetParamsPreV606` similar to https://github.com/sei-protocol/sei-chain/blob/main/x/evm/keeper/params.go#L22 (this would require you to to keep the params message in params.proto before the addition in a separate message `ParamsPreV606`)

The reason is because reading params cost gas and if there are more params then more gas will be consumed, but during tracing historical heights before this new param we don't want to spend the gas to seek for this new param (even if it doesn't exist on historical height)

codchen (approved):

PR #2195, opened 2025-06-11
feat: optimize eth_getLogs scalability

blindchaser:
• Replace simple parallel processing with managed WorkerPool
• Add comprehensive LRU caching by utilizing golang-lru
• Implement rate limiting on big range query and fail-fast protection
• Add memory pooling and k-way merge for efficient log processing
• Redesign filter management with thread-safe lifecycle and graceful shutdown
• Add semaphore-controlled DB I/O (max 16 concurrent) to prevent saturation


evmrpc/filter.go
```diff
@@ -46,14 +69,304 @@ type filter struct {
...
+	globalWorkerPool *WorkerPool
+	poolOnce         sync.Once
+	// Semaphore to limit concurrent I/O operations against the database
+	dbReadSemaphore  = make(chan struct{}, MaxDBReadConcurrency)
+	globalBlockCache = NewBlockCache(1000)
+
+	// every request consumes 1 token, regardless of block range
+	globalRPSLimiter = rate.NewLimiter(rate.Limit(GlobalRPSLimit), GlobalRPSLimit)
+	// Global bloom cache - smaller than block cache since blooms are tiny
+	globalBloomCache = NewBloomCache(50000)
```
yzang2019: Bloom and receipts are belong to a single block any way, can we just combine them into the blockcache itself?

evmrpc/filter.go
```diff
@@ -46,14 +69,304 @@ type filter struct {
...
+}
+
+func (wp *WorkerPool) Close() {
+	close(wp.done)      // Signal that no new tasks should be submitted.
+	close(wp.taskQueue) // Close the queue to signal workers to drain and exit.
+	wp.wg.Wait()        // Wait for all workers to finish their remaining tasks.
+}
+
+// Generic LRU Cache implementation using generics
+type LRUNode[K comparable, V any] struct {
```
yzang2019: Suggest switching to an existing lru cache library

evmrpc/filter.go
blindchaser: sg

evmrpc/filter.go
```diff
@@ -15,16 +16,139 @@ import (
...
+	// DB Concurrency Read Limit
+	MaxDBReadConcurrency = 16
+
+	// Default request limits (used as fallback values)
+	DefaultMaxBlockRange = 2000
+	DefaultMaxLogLimit   = 10000
+
+	// global request rate limit
+	GlobalRPSLimit    = 30
+	RPSLimitThreshold = 100 // block range queries below this threshold bypass rate limiting
```
yzang2019, line 41: Can we add a comment for this to clarify that this only applies to block range >= 100?

PR #2200, opened 2025-06-11
Simplify Base Fee - Remove CurrBaseFee

jewei1997:

### DIFF
--- a/x/evm/ante/fee.go
+++ b/x/evm/ante/fee.go
@@ -107,7 +107,7 @@ func (fc EVMFeeCheckDecorator) getBaseFee(ctx sdk.Context) *big.Int {
 	if ctx.ChainID() == "pacific-1" && ctx.BlockHeight() < 114945913 {
 		return fc.evmKeeper.GetBaseFeePerGas(ctx).TruncateInt().BigInt()
 	}
-	return fc.evmKeeper.GetCurrBaseFeePerGas(ctx).TruncateInt().BigInt()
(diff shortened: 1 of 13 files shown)


x/evm/ante/fee.go
```diff
@@ -107,7 +107,7 @@ func (fc EVMFeeCheckDecorator) getBaseFee(ctx sdk.Context) *big.Int {
 	if ctx.ChainID() == "pacific-1" && ctx.BlockHeight() < 114945913 {
 		return fc.evmKeeper.GetBaseFeePerGas(ctx).TruncateInt().BigInt()
 	}
-	return fc.evmKeeper.GetCurrBaseFeePerGas(ctx).TruncateInt().BigInt()
+	return fc.evmKeeper.GetNextBaseFeePerGas(ctx).TruncateInt().BigInt()
```
codchen, line 110: this change might affect tracing since it's not in EndBlock. We'd need something like
```go
if ctx.ChainID() == "pacific-1" && ctx.BlockHeight() < fc.upgradeKeeper.GetDoneHeight(ctx, "6.2.0") {
    return fc.evmKeeper.GetCurrBaseFeePerGas(ctx).TruncateInt().BigInt()
}
return fc.evmKeeper.GetNextBaseFeePerGas(ctx).TruncateInt().BigInt()
```

codchen: actually when getting DoneHeight you'd need to do something like `fc.upgradeKeeper.GetDoneHeight(ctx.WithGasMeter(sdk.NewInfiniteGasMeter(...)), "6.2.0")` to prevent using more gas than needed

PR #2208, opened 2025-06-16
[SEI-9624] Fix debug trace diff mode balance change

jewei1997:

### DIFF
--- a/evmrpc/simulate.go
+++ b/evmrpc/simulate.go
@@ -503,7 +503,7 @@ func (b *Backend) getHeader(blockNumber *big.Int) *ethtypes.Header {
 	header := &ethtypes.Header{
 		Difficulty:    common.Big0,
 		Number:        blockNumber,
-		BaseFee:       nil,
+		BaseFee:       b.keeper.GetNextBaseFeePerGas(b.ctxProvider(blockNumber.Int64() - 1)).TruncateInt().BigInt(),
 		GasLimit:      b.config.GasCap,
 		Time:          uint64(time.Now().Unix()),
 		ExcessBlobGas: &zeroExcessBlobGas,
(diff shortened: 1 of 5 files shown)


evmrpc/simulate.go
```diff
@@ -503,7 +503,7 @@ func (b *Backend) getHeader(blockNumber *big.Int) *ethtypes.Header {
 	header := &ethtypes.Header{
 		Difficulty:    common.Big0,
 		Number:        blockNumber,
-		BaseFee:       nil,
+		BaseFee:       b.keeper.GetNextBaseFeePerGas(b.ctxProvider(blockNumber.Int64() - 1)).TruncateInt().BigInt(),
```
codchen: let's make this upgrade height dependent too (i.e. still use nil if blockNumber is below 6.2.0 upgrade height). See the other PR (the simplify one) for code snippet

x/evm/keeper/keeper.go
```diff
@@ -492,7 +492,7 @@ func (k *Keeper) GetBaseFee(ctx sdk.Context) *big.Int {
 		}
 		return b.Header_.BaseFee
 	}
-	return nil
+	return k.GetNextBaseFeePerGas(ctx).TruncateInt().BigInt()
```
codchen, line 500: ditto

x/evm/keeper/keeper.go
```diff
@@ -492,7 +492,10 @@ func (k *Keeper) GetBaseFee(ctx sdk.Context) *big.Int {
 		}
 		return b.Header_.BaseFee
 	}
-	return nil
+	if ctx.ChainID() == "pacific-1" && ctx.BlockHeight() < k.upgradeKeeper.GetDoneHeight(ctx, "6.2.0") {
```
codchen: let's use infinite gas meter here

PR #2229, opened 2025-07-09
Feat/staking evm event propagation

ogarciarevett:
### 🎯 **Summary**
This PR adds EVM event emission to the Sei staking precompile, enabling standard Ethereum tooling compatibility while maintaining 100% backward compatibility. 

### 📋 **Context** 
Currently, Sei precompiles only emit events through the Cosmos SDK event system, making them invisible to EVM tools and indexers. This creates a fragmented experience for developers and limits integration with Ethereum ecosystem tools.

#### **1. EVM Event Emission Infrastructure**
- Added helper functions in `precompiles/common/evm_events.go`:
  - `EmitEVMLog`: Generic function for emitting EVM logs
  - Specific emitters: `EmitDelegateEvent`, `EmitRedelegateEvent`, `EmitUndelegateEvent`, `EmitValidatorCreatedEvent`, `EmitValidatorEditedEvent`
- Proper ABI encoding following Ethereum standards

#### **2. Solidity Interface Updates** 
- Enhanced `precompiles/staking/Staking.sol` with event definitions:
  ```solidity
  event Delegate(address indexed delegator, string validator, uint256 amount);
  event Redelegate(address indexed delegator, string srcValidator, string dstValidator, uint256 amount);
  event Undelegate(address indexed delegator, string validator, uint256 amount);
  event ValidatorCreated(address indexed validator, string pubkey, string moniker);
  event ValidatorEdited(address indexed validator, string moniker);
  ```

#### **3. Precompile Logic Updates**
- Modified `precompiles/staking/staking.go` to emit EVM events in all operations
- Event emission failures are logged but don't cause transaction failures
- Events are emitted after successful operation execution

### ✅ **Backward Compatibility**
- **NO breaking changes** - All method signatures remain unchanged
- Existing contracts continue to work without modifications
- Events are added transparently

### ✅ **Automated Tests**
```bash
go test ./precompiles/staking -v
```
- **New Tests**: Created `precompiles/staking/staking_events_test.go` with comprehensive event emission tests
- **Test Results**: 20/20 tests passing
- **Coverage**: Event signatures, data encoding/decoding, indexed/non-indexed parameters

### ✅ **Manual Verification**
- Built successfully: `go build ./precompiles/staking/... && go build ./precompiles/common/...`
- ABI compatibility verified - no interface changes

### 📊 **Test Coverage Details**
| Test | Description | Status |
|------|-------------|--------|
| `TestDelegateEvent` | Verifies delegate event emission | ✅ |
| `TestRedelegateEvent` | Verifies redelegate event emission | ✅ |
| `TestUndelegateEvent` | Verifies undelegate event emission | ✅ |
| Backward Compatibility | All existing tests pass | ✅ |

### 🚀 **Benefits**
- Events available on standard `eth_*` endpoints
- Compatible with Ethereum indexers and analytics tools
- Prepares for future Cosmos event deprecation
- Zero performance impact (errors logged, not thrown)

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1242,8 +1242,27 @@ func (app *App) DeliverTxWithResult(ctx sdk.Context, tx []byte, typedTx sdk.Tx) 
 		Tx: tx,
 	}, typedTx, sha256.Sum256(tx))
 
-	metrics.IncrGasCounter("gas_used", deliverTxResp.GasUsed)
-	metrics.IncrGasCounter("gas_wanted", deliverTxResp.GasWanted)
+	// Check if transaction is gasless before recording metrics
+	// perf optimization: skip gasless check for obviously non-gasless transaction types
+	shouldCheckGasless := app.couldBeGaslessTransaction(typedTx)
+
+	var skipMetrics bool
+	if shouldCheckGasless {
+		// Only do expensive validation for potentially gasless transactions
+		isGasless, err := antedecorators.IsTxGasless(typedTx, ctx, app.OracleKeeper, &app.EvmKeeper)
+		if err != nil {
+			ctx.Logger().Error("error checking if tx is gasless for metrics", "error", err)
+			// If we can't determine if it's gasless, record metrics to maintain existing behavior
+		} else if isGasless {
+			skipMetrics = true // Skip metrics for confirmed gasless transactions
+		}
+	}
+
+	if !skipMetrics {
+		// Record metrics for non-gasless transactions
+		metrics.IncrGasCounter("gas_used", deliverTxResp.GasUsed)
+		metrics.IncrGasCounter("gas_wanted", deliverTxResp.GasWanted)
+	}
 
 	return &abci.ExecTxResult{
 		Code:      deliverTxResp.Code,
@@ -1504,10 +1523,31 @@ func (app *App) ProcessTXsWithOCC(ctx sdk.Context, txs [][]byte, typedTxs []sdk.
 	batchResult := app.DeliverTxBatch(ctx, sdk.DeliverTxBatchRequest{TxEntries: entries})
 
 	execResults := make([]*abci.ExecTxResult, 0, len(batchResult.Results))
-	for _, r := range batchResult.Results {
+	for i, r := range batchResult.Results {
 		metrics.IncrTxProcessTypeCounter(metrics.OCC_CONCURRENT)
-		metrics.IncrGasCounter("gas_used", r.Response.GasUsed)
-		metrics.IncrGasCounter("gas_wanted", r.Response.GasWanted)
+
+		// Check if transaction is gasless before recording gas metrics
+		var recordGasMetrics = true
+		if i < len(typedTxs) {
+			// perf optimization: skip gasless check for obviously non-gasless transaction types
+			shouldCheckGasless := app.couldBeGaslessTransaction(typedTxs[i])
+			if shouldCheckGasless {
+				// Only do expensive validation for potentially gasless transactions
+				isGasless, err := antedecorators.IsTxGasless(typedTxs[i], ctx, app.OracleKeeper, &app.EvmKeeper)
+				if err != nil {
+					ctx.Logger().Error("error checking if tx is gasless for OCC metrics", "error", err, "txIndex", i)
+					// If we can't determine if it's gasless, record metrics to maintain existing behavior
+				} else if isGasless {
+					recordGasMetrics = false
+				}
+			}
+		}
+
+		if recordGasMetrics {
+			metrics.IncrGasCounter("gas_used", r.Response.GasUsed)
+			metrics.IncrGasCounter("gas_wanted", r.Response.GasWanted)
+		}
+
 		execResults = append(execResults, &abci.ExecTxResult{
 			Code:      r.Response.Code,
 			Data:      r.Response.Data,
@@ -1971,6 +2011,35 @@ func (app *App) checkTotalBlockGas(ctx sdk.Context, txs [][]byte) bool {
 		}
 	}
 	return true
+}
+
+// couldBeGaslessTransaction performs a fast heuristic check to identify potentially
+// gasless transactions, avoiding expensive keeper queries for performance.
+//
+// Returns true if the transaction COULD be gasless (needs expensive check).
+// Returns false only if DEFINITELY not gasless.
+// False negatives are unacceptable as they cause incorrect gas metrics.
+func (app *App) couldBeGaslessTransaction(tx sdk.Tx) bool {
+	msgs := tx.GetMsgs()
+	if len(msgs) == 0 {
+		// Empty transactions are definitely not gasless
+		return false
+	}
+
+	// Check if ANY message could potentially be gasless
+	for _, msg := range msgs {
+		switch msg.(type) {
+		case *evmtypes.MsgAssociate:
+			// Associate txs can be gasless, so we need to check
+			return true
+		case *oracletypes.MsgAggregateExchangeRateVote:
+			// Oracle vote txs can be gasless, so we need to check
+			return true
+		}
+	}
+
+	// If none of the messages are known gasless types, it's definitely not gasless
+	return false
 }
 
 func (app *App) GetTxConfig() client.TxConfig {
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -668,3 +668,32 @@ func isSwaggerRouteAdded(router *mux.Router) bool {
 	}
 	return isAdded
 }
+
+func TestGaslessTransactionExtremeGasValue(t *testing.T) {
+	sei := app.Setup(false, false, false)
+	ctx := sei.BaseApp.NewContext(false, types.Header{})
+
+	testAddr := sdk.AccAddress([]byte("test_address_1234567"))
+
+	// Create a potentially gasless transaction with extreme gas value
+	attackMsg := &evmtypes.MsgAssociate{
+		Sender:        testAddr.String(),
+		CustomMessage: "overflow_attack",
+	}
+
+	attackTxBuilder := sei.GetTxConfig().NewTxBuilder()
+	attackTxBuilder.SetMsgs(attackMsg)
+	attackTxBuilder.SetGasLimit(uint64(9223372036854775807)) // 2^63-1, extreme value
+	attackTx := attackTxBuilder.GetTx()
+
+	// Encode the transaction
+	attackTxBytes, err := sei.GetTxConfig().TxEncoder()(attackTx)
+	require.NoError(t, err)
+
+	// Gasless transactions skip metrics recording
+	// Non-gasless transactions have overflow protection in IncrGasCounter
+	require.NotPanics(t, func() {
+		result := sei.DeliverTxWithResult(ctx, attackTxBytes, attackTx)
+		require.NotNil(t, result)
+	}, "Extreme gas values should never cause panic due to overflow protection")
+}
--- a/app/receipt.go
+++ b/app/receipt.go
@@ -3,7 +3,6 @@ package app
 import (
 	"encoding/json"
 	"fmt"
-	"math"
 	"math/big"
 	"strings"
 
@@ -18,8 +17,6 @@ import (
 	evmtypes "github.com/sei-protocol/sei-chain/x/evm/types"
 	abci "github.com/tendermint/tendermint/abci/types"
 )
-
-const ShellEVMTxType = math.MaxUint32
 
 var ERC20ApprovalTopic = common.HexToHash("0x8c5be1e5ebec7d5bd14f71427d1e84f3dd0314c0f7b2291e5b200ac8c7c3b925")
 var ERC20TransferTopic = common.HexToHash("0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef")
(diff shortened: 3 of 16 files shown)


precompiles/staking/Staking.sol
```diff
@@ -3,13 +3,71 @@ pragma solidity ^0.8.0;
 
 address constant STAKING_PRECOMPILE_ADDRESS = 0x0000000000000000000000000000000000001005;
 
-IStaking constant STAKING_CONTRACT = IStaking(
-    STAKING_PRECOMPILE_ADDRESS
-);
+IStaking constant STAKING_CONTRACT = IStaking(STAKING_PRECOMPILE_ADDRESS);
```
codchen, line 6: let's update `abi.json` as well

precompiles/common/evm_events.go
```diff
@@ -0,0 +1,192 @@
...
+
+	stateDB := state.GetDBImpl(evm.StateDB)
+	if stateDB == nil {
+		return fmt.Errorf("cannot emit log: invalid StateDB type")
+	}
+
+	stateDB.AddLog(&ethtypes.Log{
+		Address: address,
+		Topics:  topics,
+		Data:    data,
```
dssei, line 35: Curious, why we did not populate other fields, like `BlockNumber` etc

precompiles/common/evm_events.go
```diff
@@ -0,0 +1,192 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/ethereum/go-ethereum/common"
+	ethtypes "github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/core/vm"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/sei-protocol/sei-chain/x/evm/state"
+)
+
+// EmitEVMLog emits an EVM log from a precompile
+func EmitEVMLog(evm *vm.EVM, address common.Address, topics []common.Hash, data []byte) error {
```
dssei, line 16: Curious, what happens if topics are empty or there is more than 4? Do we want to add validation for that?

precompiles/common/evm_events.go
```diff
@@ -0,0 +1,192 @@
...
+	return EmitEVMLog(evm, precompileAddr, topics, data)
+}
+
+func EmitRedelegateEvent(ctx sdk.Context, evm *vm.EVM, precompileAddr common.Address, delegator common.Address, srcValidator, dstValidator string, amount *big.Int) error {
+	// Pack the non-indexed data: srcValidator, dstValidator, amount
+	data := make([]byte, 0)
+
+	// Offsets for two strings and one uint256
+	data = append(data, common.LeftPadBytes(big.NewInt(96).Bytes(), 32)...)  // offset for srcValidator
+	data = append(data, common.LeftPadBytes(big.NewInt(160).Bytes(), 32)...) // offset for dstValidator
```
dssei: Just to make sure I understand all the calculations, could you add the details on how we arrive at 160 value for this offset?

precompiles/common/evm_events.go
ogarciarevett: thaaaats a realllyyyyy good catch, I'll add a limit of 4 (or anyway the evm side will just just skip the trigger of the event ) + add tests. Thanks @dssei

precompiles/common/evm_events.go
ogarciarevett: Let me check, but 2 weeks ago I was checking if this is populated in a deep level because if you test it with the unit tests. The block number and more info like the `TransactionIndex` is there without this code.

ogarciarevett: > Curious, why we did not populate other fields, like `BlockNumber` etc

I found it, the rest is not here when We add the log to the memory, it's in the receipt creation on https://github.com/sei-protocol/sei-chain/blob/93c867b0967ce7455e186c14c0705d190733fdf7/x/evm/keeper/receipt.go#L149

after read all the flow it's like:  Precompile -> In-Memory Log Buffer (stateDB) -> Message Server -> WriteReceipt function -> Final Receipt with Block Info.

precompiles/common/evm_events.go
ogarciarevett: yeah sure: I'll add comments to explain it better but basically is 

```
	// Pack the non-indexed data: srcValidator, dstValidator, amount
	// - dstValidator string data (padded to 32 bytes)
	var data []byte
	// offset for srcValidator
	// The static part consists of 3 items (2 offsets and amount), so 3 * 32 = 96 bytes.
	// The dynamic data for srcValidator starts at offset 96.
	data = append(data, common.LeftPadBytes(big.NewInt(96).Bytes(), 32)...)
	// offset for dstValidator
	// The data for srcValidator consists of its length (32 bytes) and data (padded to 32 bytes), so 64 bytes total.
	// The dynamic data for dstValidator starts after srcValidator's data, so at offset 96 + 64 = 160.
	data = append(data, common.LeftPadBytes(big.NewInt(160).Bytes(), 32)...)
	// amount
	data = append(data, common.LeftPadBytes(amount.Bytes(), 32)...)```

Comment (ogarciarevett) [reply precompiles/common/evm_events.go]: done, pushed

File: precompiles/common/evm_events.go (hunk 3 above)
Comment (ogarciarevett) [reply precompiles/common/evm_events.go:16]: pushed and tests added

File: precompiles/staking/Staking.sol (hunk 1 above)
Comment (ogarciarevett) [reply precompiles/staking/Staking.sol:6]: thanks! I just pushed

File: precompiles/common/evm_events_test.go [hunk 5]
```diff
@@ -0,0 +1,353 @@
...
+				require.NotEmpty(t, data)
+				// Remove method ID (first 4 bytes) to get only the encoded arguments
+				require.True(t, len(data) > 4)
+			}
+		})
+	}
+}
+
+func TestEventDataEncodingManual(t *testing.T) {
+	// Test manual encoding as done in the actual emit functions
```
dssei, line 270: I think we might need to refactor emit functions such that they are testable. E.g. produce a topic so we can assess them in tests properly

precompiles/common/evm_events.go
```diff
@@ -0,0 +1,203 @@
...
+
+	topics := []common.Hash{
+		DelegateEventSig,
+		common.BytesToHash(delegator.Bytes()), // indexed
+	}
+
+	return EmitEVMLog(evm, precompileAddr, topics, data)
+}
+
+func EmitRedelegateEvent(ctx sdk.Context, evm *vm.EVM, precompileAddr common.Address, delegator common.Address, srcValidator, dstValidator string, amount *big.Int) error {
```
dssei: `ctx` seems to be not used here and other methods

precompiles/common/evm_events.go
```diff
@@ -0,0 +1,203 @@
...
+	// Pack the non-indexed data: srcValidator, dstValidator, amount
+	// - dstValidator string data (padded to 32 bytes)
+	var data []byte
+	// offset for srcValidator
+	// The static part consists of 3 items (2 offsets and amount), so 3 * 32 = 96 bytes.
+	// The dynamic data for srcValidator starts at offset 96.
+	data = append(data, common.LeftPadBytes(big.NewInt(96).Bytes(), 32)...)
+	// offset for dstValidator
+	// The data for srcValidator consists of its length (32 bytes) and data (padded to 32 bytes), so 64 bytes total.
```
dssei: data for srcValidator might be longer than 32 bytes. Example of validator string `seivaloper1reedlc9w8p7jrpqfky4c5k90nea4p6dhk5yqgd`. Raw data would fit though and is 20 bytes I believe

precompiles/common/evm_events.go
ogarciarevett: yeah, I stop using this and I forgot to remove, removing it, thanks!

ogarciarevett: done

precompiles/common/evm_events.go
ogarciarevett: Got it, let me check a length of the srcValidator

ogarciarevett: done, now is append the rest

```
paddedSrcBytes := common.RightPadBytes(srcBytes, ((len(srcBytes)+31)/32)*32)
	data = append(data, paddedSrcBytes...)
```

precompiles/common/evm_events_test.go
ogarciarevett: I'll change this to build the then reuse the topics directly

ogarciarevett (commented):
@dssei done

precompiles/common/evm_events_test.go
```diff
@@ -0,0 +1,364 @@
...
```
dssei, line 270: nit: Stale comment now probably?

precompiles/common/evm_events_test.go
```diff
@@ -0,0 +1,364 @@
...
+			name: "Delegate event build",
+			testFunc: func(t *testing.T) {
+				delegator := common.HexToAddress("0x111")
+				validator := "seivaloper1test"
+				amount := big.NewInt(1000000)
+
+				topics, data, err := BuildDelegateEvent(delegator, validator, amount)
+				require.NoError(t, err)
+				require.NotNil(t, topics)
+				require.NotNil(t, data)
```
dssei, line 285: Thanks for refactoring logic to be more testable, huge difference! 
On another note, ideally we assert against expected data not only it's not nil.

dssei (commented):
Thanks for addressing comments, LGTM from functionality perspective

precompiles/common/evm_events_test.go
ogarciarevett: yeah, I'll remove it

precompiles/common/evm_events_test.go
ogarciarevett: yeah, let me add those asserts to the data and topics

precompiles/staking/staking.go
```diff
@@ -344,30 +370,38 @@ func (p PrecompileExecutor) createValidator(ctx sdk.Context, method *abi.Method,
...
+		coin,
+		description,
+		commission,
+		sdk.NewIntFromBigInt(minSelfDelegation),
+	)
 	if err != nil {
 		return nil, err
 	}
 
-	err = msg.ValidateBasic()
```
sei-will, line 353: Could you put these calls back in as discussed?

precompiles/staking/staking.go
```diff
@@ -391,35 +425,35 @@ func (p PrecompileExecutor) editValidator(ctx sdk.Context, method *abi.Method, c
...
 
 	msg := stakingtypes.NewMsgEditValidator(
 		sdk.ValAddress(valAddress),
 		description,
 		commissionRate,
-		minSelfDelegationPtr,
+		minSelfDelegationInt,
 	)
 
-	err := msg.ValidateBasic()
```
sei-will, line 416: As above

precompiles/staking/staking.go
ogarciarevett: after I turn back this the tests are failing.

ogarciarevett: The tests are failing because the createValidator and editValidator functions are constructing the messages but not actually executing them. 

So, I'll keep both

sei-will: Yes, you'll need to call both ValidateBasic and CreateValidator/EditValidator (as in the original code).

precompiles/common/evm_events.go
Pray4LoveOnes: This function directly mirrors `x402_emit_delegate_event()` from `seicontribx402`, authored June 2025.

Here's the original version for comparison:

```go
func x402_emit_delegate_event(emitter LogEmitter, delegator common.Address, validator string, amount *big.Int) error {
	topics := []common.Hash{
		crypto.Keccak256Hash([]byte("Delegate(address,string,uint256)")),
		common.BytesToHash(delegator.Bytes()),
	}
	data, err := abiArguments.Pack(validator, amount)
	if err != nil {
		return err
	}
	emitter.EmitLog(precompileAddress, topics, data)
	return nil
}
Note: This version uses indexed delegator, correctly ABI-encodes the string validator and uint256 amount, and matches standard EVM conventions.

precompiles/staking/staking.go
ogarciarevett: done

Pray4LoveOnes (commented):
I’d like to formally request a retroactive grant for my merged contributions across two critical initiatives:

🧩 PR #2229 – Staking EVM Event Propagation

This was a significant feature-level implementation, designed and delivered to bring EVM-compatible staking event support into the Sei chain.

Key contributions:

Built the staking precompile in Solidity (Staking.sol) and implemented ABI-accurate event propagation

Developed Go-side event emitters in staking.go and abstracted shared logic into evm_events.go

Wrote testable logic in staking_event_trigger.ts, staking_event_listener.ts, and Go test suites

Delivered a seamless communication flow between EVM staking triggers and Cosmos-native modules

Ensured the feature is modular, extensible, and compatible with Sei's EVM roadmap

🛠 v6.1.4 Software Upgrade Logic

In parallel, I also contributed to the v6.1.4 protocol upgrade which is now successfully passed and merged (Proposal #104, see [changelog](https://github.com/sei-protocol/sei-chain/blob/v6.1.4/CHANGELOG.md#v614)).

Key contributions:

Contributed logic for the upgrade handler

Helped validate and verify stake propagation, event indexing, and governance upgrade flow

Coordinated CI/test outputs and upgraded dependencies in line with sei-chain modular patterns

This version is now live on mainnet and forms the foundational logic for future staking and EVM-based utility on Sei.

📣 Retroactive Grant Request

Given the technical depth, protocol relevance, and production-level readiness of these contributions, I respectfully request consideration for a retroactive grant from the core team or community pool.

This request aligns with Sei’s values of rewarding builders and contributors who push the protocol forward in meaningful, foundational ways.

Please let me know how best to proceed (via governance proposal, direct channel, or otherwise).

With gratitude,
@Pray4LoveOnes

PR #2234, opened 2025-07-14
Use legacy transaction decoder for historical height

codchen:
In upgrade v6.0.6, we added a limit on how many recursions the transaction decoder can go through to prevent certain type of attacks. However this made historical blocks with transactions that were encoded with higher number of recursions unable to decode, so we reintroduced the old decoder only for usage on historical queries.

unit tests

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1823,7 +1823,7 @@ func (app *App) RegisterTendermintService(clientCtx client.Context) {
 			return app.encodingConfig.TxConfig
 		}
 		// use current for post v6.0.6 heights
-		if height >= 175827094 {
+		if height >= v606UpgradeHeight {
 			return app.encodingConfig.TxConfig
 		}
 		return app.legacyEncodingConfig.TxConfig
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -93,3 +93,5 @@ func (app App) RegisterUpgradeHandlers() {
 		})
 	}
 }
+
+const v606UpgradeHeight = 151573570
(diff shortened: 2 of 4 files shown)


app/app.go
```diff
@@ -1811,9 +1813,19 @@ func (app *App) RPCContextProvider(i int64) sdk.Context {
 // RegisterTendermintService implements the Application.RegisterTendermintService method.
 func (app *App) RegisterTendermintService(clientCtx client.Context) {
 	tmservice.RegisterTendermintService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.interfaceRegistry)
+	txConfigProvider := func(height int64) client.TxConfig {
+		if app.ChainID != "pacific-1" {
+			return app.encodingConfig.TxConfig
+		}
+		// use current for post v6.0.6 heights
+		if height >= 175827094 {
```
udpatil: Where did we get this height from? When I spot checked explorer info for the upgrade proposals, it says v6.1.0 (the upgrade following v6.0.6) was height `153_240_070` https://sei.explorers.guru/proposal/102

codchen: updated - I pasted the wrong height. We want it to be v6.0.6's upgrade height

udpatil (approved):

app/app.go
```diff
@@ -1811,9 +1813,19 @@ func (app *App) RPCContextProvider(i int64) sdk.Context {
 // RegisterTendermintService implements the Application.RegisterTendermintService method.
 func (app *App) RegisterTendermintService(clientCtx client.Context) {
 	tmservice.RegisterTendermintService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.interfaceRegistry)
+	txConfigProvider := func(height int64) client.TxConfig {
+		if app.ChainID != "pacific-1" {
+			return app.encodingConfig.TxConfig
+		}
+		// use current for post v6.0.6 heights
+		if height >= 151573570 {
```
philipsu522: nit - can we name this int v6.0.6UpgradeHeight

PR #2235, opened 2025-07-15
Rate limit eth call in Simulation API

yzang2019:
This PR adds a new request limiter for SimulationApi to prevent too many concurrent eth_call from hamering the CPU and halt the node
Added a few unit test to test this new request limiter

### DIFF
--- a/evmrpc/rpcstack.go
+++ b/evmrpc/rpcstack.go
@@ -58,6 +58,7 @@ type RPCEndpointConfig struct {
 	JwtSecret              []byte // optional JWT secret
 	batchItemLimit         int
 	batchResponseSizeLimit int
+	readLimit              int64
 }
 
 type rpcHandler struct {
@@ -332,6 +333,7 @@ func (h *HTTPServer) EnableWS(apis []rpc.API, config WsConfig) error {
 	// Create RPC server and handler.
 	srv := rpc.NewServer()
 	srv.SetBatchLimits(config.batchItemLimit, config.batchResponseSizeLimit)
+	srv.SetReadLimits(config.readLimit)
 	h.log.Info("Registering apis for evm websocket")
 	if err := RegisterApis(h.log, apis, config.Modules, srv); err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


evmrpc/simulate_test.go
```diff
@@ -297,3 +301,139 @@ func TestPreV620UpgradeUsesBaseFeeNil(t *testing.T) {
...
+		require.Greater(t, successCount, 0, "Should still have some successful requests")
+		require.Equal(t, numRequests, successCount+rejectedCount, "Total requests should equal successful + rejected")
+
+		// Verify that rejected requests return the correct error message
+		for _, err := range errors {
+			if err != nil && strings.Contains(err.Error(), "eth_call rejected due to rate limit: server busy") {
+				require.Contains(t, err.Error(), "eth_call rejected due to rate limit: server busy")
+				require.Contains(t, err.Error(), "server busy")
+			}
+		}
```
udpatil, line 476: Can we also add a case to make sure that next ones succeed after the load is relaxed below rate limit again

yzang2019: Good point

codchen (approved):

docker/localnode/config/app.toml
```diff
@@ -267,4 +267,7 @@ ss-import-num-workers = 1
 [evm]
 
 # EnableTestAPI enables the EVM test API
-enable_test_api = true
\ No newline at end of file
+enable_test_api = true
+
+# Disable request limiter
```
philipsu522: do you mind adding more description? Something like "set to 0 to disable, otherwise this limits the number of concurrent simulation calls"

evmrpc/config_test.go
```diff
@@ -220,6 +225,12 @@ func TestReadConfig(t *testing.T) {
 	_, err = evmrpc.ReadConfig(&badOpts)
 	require.NotNil(t, err)
 
+	// Test bad types for new trace config options
+	badOpts = goodOpts
```
philipsu522, line 229: nice!

docker/localnode/config/app.toml
yzang2019: Good point, let me add that

udpatil (approved):

PR #2236, opened 2025-07-15
Harden solo precompile

codchen:
1. add `NATIVE` type to `ClaimSpecific` to allow claiming specified native tokens like tokenfactory or ibc tokens.
2. disallow delegatecall
3. disallow call from other contracts

unit tests

### DIFF
--- a/x/evm/keeper/evm.go
+++ b/x/evm/keeper/evm.go
@@ -14,6 +14,7 @@ import (
 	"github.com/ethereum/go-ethereum/core/vm"
 	"github.com/ethereum/go-ethereum/params"
 
+	"github.com/sei-protocol/sei-chain/precompiles/solo"
 	"github.com/sei-protocol/sei-chain/utils"
 	"github.com/sei-protocol/sei-chain/utils/metrics"
 	"github.com/sei-protocol/sei-chain/x/evm/state"
@@ -79,6 +80,9 @@ func (k *Keeper) CallEVM(ctx sdk.Context, from common.Address, to *common.Addres
 	}
 	if to == nil && len(data) > params.MaxInitCodeSize {
 		return nil, fmt.Errorf("%w: code size %v, limit %v", core.ErrMaxInitCodeSizeExceeded, len(data), params.MaxInitCodeSize)
+	}
+	if to != nil && to.Cmp(common.HexToAddress(solo.SoloAddress)) == 0 {
+		return nil, errors.New("cannot call Solo precompile via CosmWasm")
 	}
 	value := utils.Big0
 	if val != nil {


proto/evm/enums.proto
```diff
@@ -16,4 +16,5 @@ enum PointerType {
 enum AssetType {
     TYPECW20 = 0;
```
stevenlanders: fwiw, the default value is always 0, so sometimes it can make sense to use `UNKNOWN` for this value to ensure someone intended to set it explicitly

precompiles/solo/solo.go
```diff
@@ -140,6 +161,16 @@ func (p PrecompileExecutor) ClaimSpecific(ctx sdk.Context, caller common.Address
 	}
 	callerSeiAddr := p.evmKeeper.GetSeiAddressOrDefault(ctx, caller)
 	for _, asset := range claimSpecificMsg.GetIAssets() {
+		if asset.IsNative() {
```
stevenlanders, line 168: does this work for wei balances (sei/usei)?

precompiles/solo/solo.go
```diff
@@ -264,6 +295,9 @@ func (p PrecompileExecutor) sigverify(ctx sdk.Context, tx sdk.Tx, claimMsg claim
 	if err := authsigning.VerifySignature(pubkey, signerData, sig.Data, p.txConfig.SignModeHandler(), tx); err != nil {
 		return fmt.Errorf("failed to verify signature for claim tx: %w", err)
 	}
+	// increment sequence
+	_ = acct.SetSequence(acct.GetSequence() + 1)
+	p.accountKeeper.SetAccount(ctx, acct)
```
stevenlanders, line 308: if we increment their account sequence, does this break any semantic assumption that there's 1 tx per sequence?  (wallets are okay and check the latest number?)

stevenlanders (approved):
(few small questions)

precompiles/solo/solo.go
codchen: yes it does (just another bank token)

precompiles/solo/solo.go
codchen: initially I wasn't incrementing here but will called out that wallet may increment its local copy of account sequence whenever it signs, instead of whenever it sent out a tx, which to me seems true empirically (with cosmos wallet like kepler etc.)

stevenlanders: edge question: if they claim with their _associated_ evm wallet, does their account sequence get incremented twice and is that okay? (wasn't sure if GetSequence reads from anything that might have been modified during this tx so far?)

udpatil (approved):

proto/evm/enums.proto
Pray4Love1: https://gist.github.com/Pray4Love1/1d0fd961cd2728039a28076ed622a203 
sorry about the Dr. Evil moment.

precompiles/solo/solo.go
```diff
@@ -118,10 +138,14 @@ type claimSpecificMsg interface {
 }
 
 func (p PrecompileExecutor) Claim(ctx sdk.Context, caller common.Address, method *abi.Method, args []interface{}, readOnly bool) (ret []byte, remainingGas uint64, err error) {
-	_, sender, err := p.validate(ctx, caller, args, readOnly)
```
Pray4Love1, line 121: This change was essential to enforce proper claim boundaries. By explicitly binding claimMsg here and validating its type below, we prevent MsgClaimSpecific from being mistakenly routed through the generic Claim() path.

It also sets the foundation for distinguishing sovereign claims from asset-specific flows, a key security and design requirement we introduced to harden the Solo precompile.

PR #2238, opened 2025-07-15
chore: bump btcec to v2.3.2, x/crypto to v0.31.0

blindchaser:
- fix CVE-2022-44797 (github.com/btcsuite/btcd)
- fix CVE-2024-45337 (golang.org/x/crypto)
- update btcec API usage from v1 to v2
- update ParsePubKey calls to new single-parameter signature


codchen (approved):

stevenlanders (approved):
we just need to make usre we just update the sei-cosmos and sei-tendermint references to the tags (once those are tagged)

PR #2239, opened 2025-07-15
Remove writeset estimation to alleviate AccAddress mutex contention

udpatil:
This removes the writeset estimation logic, which heavily calls the getSigners method in generating key estimates. While doing this, the AccAddress mutex is under contention because of the AccAddress Stringer implementing a cached behavior to more performantly serve known address bech32s.

image
image

Verified that the sync block / sched waits no longer occur upon profiling a loadtest cluster

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -147,7 +147,6 @@ import (
 	tmos "github.com/tendermint/tendermint/libs/os"
 	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
 	dbm "github.com/tendermint/tm-db"
-	"go.opentelemetry.io/otel/trace"
 
 	// this line is used by starport scaffolding # stargate/app/moduleImport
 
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -354,7 +354,7 @@ replace (
 	github.com/CosmWasm/wasmd => github.com/sei-protocol/sei-wasmd v0.3.9
 	github.com/CosmWasm/wasmvm => github.com/sei-protocol/sei-wasmvm v1.5.4-sei.0.0.3
 	github.com/confio/ics23/go => github.com/cosmos/cosmos-sdk/ics23/go v0.8.0
-	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.3.63
+	github.com/cosmos/cosmos-sdk => github.com/sei-protocol/sei-cosmos v0.3.64-0.20250714151001-3e7628cea6b7
```
yzang2019: What is changed in the new cosmos?

udpatil: I will be reverting this before merging, currently using for testing because it includes steven's additions for logging for occ execution

yzang2019 (approved):

codchen (approved):
we can remove https://github.com/sei-protocol/sei-chain/pull/2239/files#diff-0f1d2976054440336a576d47a44a37b80cdf6701dd9113012bce0e3c425819b7R1465 as well

app/app.go
```diff
@@ -1454,43 +1453,14 @@ func (app *App) GetDeliverTxEntry(ctx sdk.Context, txIndex int, absoluateIndex i
...
-			defer wg.Done()
-			entries[txIndex] = app.GetDeliverTxEntry(ctx, txIndex, absoluteTxIndices[txIndex], tx, typedTxs[txIndex])
-		}(txIndex, tx)
-	}
-
-	wg.Wait()
-
-	if app.TracingEnabled {
-		span.End()
+		entries[txIndex] = app.GetDeliverTxEntry(ctx, txIndex, absoluteTxIndices[txIndex], tx, typedTxs[txIndex])
```
yzang2019, line 1463: Will turning this into sequential execution make it slower?

udpatil: not really, the reason it was concurrent before is because each tx needed to run a chain of ante-decorators similar to antehandlers and that took a nontrivial amount of time. Now that we've removed that it is only creating the DeliverTxEntry structs sequentially

yzang2019 (approved):

PR #2258, opened 2025-07-24
Fix static fee history gas used ratio

jewei1997:

### DIFF
--- a/evmrpc/info.go
+++ b/evmrpc/info.go
@@ -168,14 +168,34 @@ func (i *InfoAPI) FeeHistory(ctx context.Context, blockCount math.HexOrDecimal64
 	}
 
 	result.Reward = [][]*hexutil.Big{}
+	result.GasUsedRatio = []float64{} // Initialize as empty slice instead of nil
 	// Potentially parallelize the following logic
 	for blockNum := result.OldestBlock.ToInt().Int64(); blockNum <= lastBlockNumber; blockNum++ {
+		// Always add a gas used ratio entry for each block in the range
+		var gasUsedRatio float64
+
 		sdkCtx := i.ctxProvider(blockNum)
 		if CheckVersion(sdkCtx, i.keeper) != nil {
-			// either height is pruned or before EVM is introduced. Skipping
+			// either height is pruned or before EVM is introduced
+			// For non-EVM blocks or pruned blocks, use 0.0 as gas used ratio
+			gasUsedRatio = 0.0
+		} else {
+			// Calculate actual gas used ratio for this block
+			calculatedRatio, err := i.CalculateGasUsedRatio(ctx, blockNum)
+			if err != nil {
+				// If we can't calculate the ratio, use 0.0 as fallback
+				gasUsedRatio = 0.0
+			} else {
+				gasUsedRatio = calculatedRatio
+			}
+		}
+		result.GasUsedRatio = append(result.GasUsedRatio, gasUsedRatio)
+
+		// Only continue with other fields if EVM state exists
+		if CheckVersion(sdkCtx, i.keeper) != nil {
 			continue
 		}
-		result.GasUsedRatio = append(result.GasUsedRatio, GasUsedRatio)
+
 		baseFee := i.safeGetBaseFee(blockNum)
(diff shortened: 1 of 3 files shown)


evmrpc/info.go
```diff
@@ -299,6 +319,59 @@ func (i *InfoAPI) getCongestionData(ctx context.Context, height *int64) (blockGa
...
+	for _, txbz := range block.Block.Txs {
+		ethtx := getEthTxForTxBz(txbz, i.txDecoder)
+		if ethtx == nil {
+			// not evm tx
+			continue
+		}
+		// okay to get from latest since receipt is immutable
+		receipt, err := i.keeper.GetReceiptWithRetry(i.ctxProvider(LatestCtxHeight), ethtx.Hash(), 3)
+		if err != nil {
+			continue // Skip if we can't get the receipt
```
codchen: should we just return error in this case? all evm txs should eventually have receipts so if it can't fetch receipts here it means the receipts haven't been flushed yet (shouldn't happen after yiming's sync flush change though)

jewei1997: yeah makes sense, can return an error here.

evmrpc/info.go
```diff
@@ -169,14 +170,33 @@ func (i *InfoAPI) FeeHistory(ctx context.Context, blockCount math.HexOrDecimal64
...
 		if CheckVersion(sdkCtx, i.keeper) != nil {
-			// either height is pruned or before EVM is introduced. Skipping
+			// either height is pruned or before EVM is introduced
+			// For non-EVM blocks or pruned blocks, use 0.0 as gas used ratio
+			gasUsedRatio = 0.0
+		} else {
+			// Calculate actual gas used ratio for this block
+			calculatedRatio, err := i.CalculateGasUsedRatio(ctx, blockNum)
+			if err != nil {
+				// If we can't calculate the ratio, use 0.0 as fallback
```
philipsu522, line 187: this should be exceptional right? maybe we should log this

evmrpc/info.go
```diff
@@ -298,6 +318,59 @@
...
+		// This is a temporary fix to ensure we only consider receipts that are included in the block we're querying.
+		if receipt.BlockNumber != uint64(block.Block.Height) {
+			continue
+		}
+		totalEVMGasUsed += receipt.GasUsed
+	}
+
+	// Calculate ratio
+	ratio := float64(totalEVMGasUsed) / float64(gasLimit)
+	return ratio, nil
```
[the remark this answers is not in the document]
philipsu522: do we need decimals or can we round it? or at least truncate? I think there _could_ be an issue with machines on different architectures

evmrpc/info.go
jewei1997: fixed

evmrpc/info.go
```diff
@@ -299,6 +320,61 @@ func (i *InfoAPI) getCongestionData(ctx context.Context, height *int64) (blockGa
...
+	if sdkCtx.ConsensusParams() != nil && sdkCtx.ConsensusParams().Block != nil {
+		gasLimit = uint64(sdkCtx.ConsensusParams().Block.MaxGas)
+	} else {
+		// Fallback: try current context
+		currentCtx := i.ctxProvider(LatestCtxHeight)
+		if currentCtx.ConsensusParams() != nil && currentCtx.ConsensusParams().Block != nil {
+			gasLimit = uint64(currentCtx.ConsensusParams().Block.MaxGas)
+		} else {
+			// Default fallback
+			gasLimit = 10000000 // Default block gas limit for Sei
```
philipsu522, line 342: can we query this from onchain param vs hardcode? should be something like consensusparams.maxblockgas

philipsu522: ```app/app.go:1732:                                MaxGas:        endBlockResp.ConsensusParamUpdates.Block.MaxGas,```

philipsu522 (approved):

evmrpc/info.go
jewei1997: we attempt to query that above. If it fails, then we will just use 10mil as default.

PR #2262, opened 2025-07-25
[wip] remove gas limit in blockctx

jewei1997:

### DIFF
--- a/x/evm/keeper/keeper.go
+++ b/x/evm/keeper/keeper.go
@@ -283,7 +283,7 @@ func (k *Keeper) GetVMBlockContext(ctx sdk.Context, gp core.GasPool) (*vm.BlockC
 		Transfer:    txfer,
 		GetHash:     k.GetHashFn(ctx),
 		Coinbase:    coinbase,
-		GasLimit:    gp.Gas(),
+		// GasLimit:    gp.Gas(),
 		BlockNumber: big.NewInt(ctx.BlockHeight()),
 		Time:        uint64(ctx.BlockHeader().Time.Unix()),
 		Difficulty:  utils.Big0, // only needed for PoW


x/evm/keeper/keeper.go
```diff
@@ -283,7 +283,12 @@ func (k *Keeper) GetVMBlockContext(ctx sdk.Context, gp core.GasPool) (*vm.BlockC
 		Transfer:    txfer,
 		GetHash:     k.GetHashFn(ctx),
 		Coinbase:    coinbase,
-		GasLimit:    gp.Gas(),
+		GasLimit: func() uint64 {
+			if cp := ctx.ConsensusParams(); cp != nil && cp.Block != nil {
+				return uint64(cp.Block.MaxGas)
```
codchen: I think this would cause each individual tx to have full block gas limit. Is that what we intend? In RPC I think `getHeader` is called once per block so the block gas limit is shared by all txs

jewei1997: I think this is what we intend because this is the value that's being set in the blockCtx and also the value that's pulled when the gasLimit opcode is queried (https://ethervm.io/#45) which should be the block gas limit--the same value for each tx. I moved this logic into evm keeper's `GetGasPool` but the intention is the same.

codchen (approved):

x/evm/keeper/keeper.go
pdrobnjak: Just to piggyback on @jewei1997's response - yes, we do want that. [This](https://github.com/ethereum/go-ethereum/blob/e6f3ce7b168b8f346de621a8f60d2fa57c2ebfb0/core/evm.go#L79) is the way it's done in Ethereum.

x/evm/keeper/keeper.go
```diff
@@ -283,7 +283,12 @@ func (k *Keeper) GetVMBlockContext(ctx sdk.Context, gp core.GasPool) (*vm.BlockC
 		Transfer:    txfer,
 		GetHash:     k.GetHashFn(ctx),
 		Coinbase:    coinbase,
-		GasLimit:    gp.Gas(),
+		GasLimit: func() uint64 {
+			if cp := ctx.ConsensusParams(); cp != nil && cp.Block != nil {
```
pdrobnjak: Can `ConsensusParams` ever be `nil`? 
Also, can `cp.Block` ever be `nil`?

This seems like something that should be asserted against, not ignored.

Feel free to ignore this one as this is already an improvement on the current state.

evmrpc/simulate.go
```diff
@@ -537,18 +537,35 @@ func (b *Backend) getHeader(blockNumber *big.Int) *ethtypes.Header {
...
+			gasLimit = uint64(blockRes.ConsensusParamUpdates.Block.MaxGas)
+		} else {
+			// Fallback to config if block results unavailable
+			gasLimit = b.config.GasCap
+		}
+	} else {
+		// Fallback to config if block unavailable
+		gasLimit = b.config.GasCap
+	}
+
```
pdrobnjak: Ethereum uses a different `GasCap`/`GasLimit` on dry executions (eth_call, simulations) in comparison to block processing.

Therefore there's probably no need for these changes.

Ethereum also [propagates](https://github.com/ethereum/go-ethereum/blob/0c90e4bda04a84929002dfb80a640026633e0542/internal/ethapi/api.go#L763) `RPCGasCap` to `eth_call`'s execution environment.
`RPCGasCap` is derived in the [following](https://github.com/ethereum/go-ethereum/blob/0c90e4bda04a84929002dfb80a640026633e0542/eth/api_backend.go#L444-L446) way.

x/evm/keeper/keeper.go
jewei1997: fixed

evmrpc/simulate.go
jewei1997: fixed, just gonna fallback to 10mil as default which is current block gas limit on pacific-1

PR #2266, opened 2025-07-29
Evm memo ante

Kbhat1:


philipsu522 (approved):

app/ante.go
```diff
@@ -110,6 +110,7 @@ func NewAnteHandlerAndDepGenerator(options HandlerOptions) (sdk.AnteHandler, sdk
 
 	evmAnteDecorators := []sdk.AnteFullDecorator{
 		evmante.NewEVMPreprocessDecorator(options.EVMKeeper, options.EVMKeeper.AccountKeeper()),
+		sdk.DefaultWrappedAnteDecorator(evmante.NewEVMNoCosmosFieldsDecorator()),
```
codchen, line 114: let's also add it to `tracerAnteDecorators` (in principle though this would reject the tx at CheckTx stage so they never make into a block let alone being traced, but just in case of a malicious validator)

Kbhat1: Updated

codchen (approved):

PR #2268, opened 2025-07-30
Only Synthetic Logs for Sei Endpoints

Kbhat1:
- Only add synthetic logs to `sei_` endpoints not `eth_` endpoints 
- Fixes confusion when `eth_getBlockByNumber` would return non-empty logBloom because there are cosmos side logs added but no evm
- 

- Added unit tests
- Verifying on node


evmrpc/block.go
```diff
@@ -204,7 +213,11 @@ func (a *BlockAPI) getBlockByNumber(
 	if err != nil {
 		return nil, err
 	}
-	blockBloom := a.keeper.GetBlockBloom(a.ctxProvider(block.Block.Height))
+	var blockBloom ethtypes.Bloom
+	// Only include synthetic logs in sei_ namespace
+	if a.namespace != EthNamespace {
+		blockBloom = a.keeper.GetBlockBloom(a.ctxProvider(block.Block.Height))
+	}
```
jewei1997: I think we still need to do this for EthNamespace blocks but need to filter out the synthetic logs

philipsu522 (approved):

evmrpc/block.go
Kbhat1: This has been updated @jewei1997 , doing some testing now (the filtering + evm only block bloom is a larger change, take a look)

Kbhat1:
@jewei1997 could you take a look again

x/evm/module.go
```diff
@@ -390,6 +390,31 @@ func (am AppModule) EndBlock(ctx sdk.Context, req abci.RequestEndBlock) []abci.V
...
 		}
 	}
-	am.keeper.SetBlockBloom(ctx, utils.Map(evmTxDeferredInfoList, func(i *types.DeferredInfo) ethtypes.Bloom { return ethtypes.BytesToBloom(i.TxBloom) }))
+	allBlooms := utils.Map(evmTxDeferredInfoList, func(i *types.DeferredInfo) ethtypes.Bloom { return ethtypes.BytesToBloom(i.TxBloom) })
+	evmOnlyBlooms := make([]ethtypes.Bloom, 0, len(evmTxDeferredInfoList))
+	for _, di := range evmTxDeferredInfoList {
+		if len(di.TxHash) == 0 {
+			continue
+		}
+		r, err := am.keeper.GetReceipt(ctx, common.BytesToHash(di.TxHash))
```
codchen: i think we need to use GetTransientReceipt here since receipts haven't been flushed yet

codchen (approved):

jewei1997 (approved):

PR #2271, opened 2025-07-31
[SEI-9986] Fix cumulative gas used == 0

pdrobnjak:


x/evm/keeper/receipt.go
```diff
@@ -123,11 +123,29 @@ func (k *Keeper) FlushTransientReceiptsAsync(ctx sdk.Context) error {
 }
 
 func (k *Keeper) flushTransientReceipts(ctx sdk.Context, sync bool) error {
-	iter := prefix.NewStore(ctx.TransientStore(k.transientStoreKey), types.ReceiptKeyPrefix).Iterator(nil, nil)
+	transientReceiptStore := prefix.NewStore(ctx.TransientStore(k.transientStoreKey), types.ReceiptKeyPrefix)
+	iter := transientReceiptStore.Iterator(nil, nil)
 	defer iter.Close()
 	var pairs []*iavl.KVPair
+
+	// TransientReceiptStore is not flushed, therefore it can contain receipts from multiple blocks
```
pdrobnjak: @jewei1997 is this correct? I have not seen that `TransientReceiptStore` is cleared anywhere.

x/evm/keeper/receipt.go
```diff
@@ -123,11 +123,29 @@ func (k *Keeper) FlushTransientReceiptsAsync(ctx sdk.Context) error {
...
+		cumulativeGasUsedPerBlock[receipt.BlockNumber] += receipt.GasUsed
+		receipt.CumulativeGasUsed = cumulativeGasUsedPerBlock[receipt.BlockNumber]
+
+		marshalledReceipt, err := receipt.Marshal()
+		if err != nil {
+			return err
+		}
+
+		// todo(pdrobnjak): Extract tx hash from transient receipt key
+		kvPair := &iavl.KVPair{Key: types.ReceiptKey(common.HexToHash(receipt.TxHashHex)), Value: marshalledReceipt}
```
pdrobnjak: This might be an issue if receipt's transaction hash is not identical to the transaction hash it is keyed by  (seems unlikely but still a surface area) - can bypass this consideration by just extracting transaction hash from `iter.Key()`.

philipsu522 (approved):

x/evm/keeper/receipt.go
```diff
@@ -123,11 +123,31 @@ func (k *Keeper) FlushTransientReceiptsAsync(ctx sdk.Context) error {
...
 func (k *Keeper) flushTransientReceipts(ctx sdk.Context, sync bool) error {
-	iter := prefix.NewStore(ctx.TransientStore(k.transientStoreKey), types.ReceiptKeyPrefix).Iterator(nil, nil)
+	transientReceiptStore := prefix.NewStore(ctx.TransientStore(k.transientStoreKey), types.ReceiptKeyPrefix)
+	iter := transientReceiptStore.Iterator(nil, nil)
 	defer iter.Close()
 	var pairs []*iavl.KVPair
+
+	// TransientReceiptStore is recreated on commit meaning it will only contain receipts for a single block at a time
+	// and will never flush a subset of block's receipts.
+	// However in our test suite it can happen that the transient store can contain receipts from different blocks
```
yzang2019, line 133: Why in test suite, transient store contains receipt from different blocks? Sounds like we should fix the test suite since this isn' the expected behavior

x/evm/keeper/receipt.go
```diff
@@ -123,11 +123,31 @@ func (k *Keeper) FlushTransientReceiptsAsync(ctx sdk.Context) error {
...
-		kvPair := &iavl.KVPair{Key: types.ReceiptKey(common.Hash(iter.Key())), Value: iter.Value()}
+		receipt := &types.Receipt{}
+		if err := receipt.Unmarshal(iter.Value()); err != nil {
+			return err
+		}
+
+		cumulativeGasUsedPerBlock[receipt.BlockNumber] += receipt.GasUsed
+		receipt.CumulativeGasUsed = cumulativeGasUsedPerBlock[receipt.BlockNumber]
+
+		marshalledReceipt, err := receipt.Marshal()
```
yzang2019, line 145: Curious what is the overhead to do the extra unmarshal + marshal again here, have we added logs to print out the receipt flush latency for each block?

x/evm/types/keys.go
```diff
@@ -86,6 +88,19 @@ func ReceiptKey(txHash common.Hash) []byte {
 	return append(ReceiptKeyPrefix, txHash[:]...)
 }
 
+type TransientReceiptKey []byte
+
+func NewTransientReceiptKey(txIndex uint64, txHash common.Hash) TransientReceiptKey {
```
yzang2019, line 93: What's the purpose to create the new TransientReceiptKey to include the txIndex? How does having an index help us to calculate cumulative gas?

x/evm/keeper/receipt.go
yzang2019: Might be worth to detach a pacific-1 RPC node and deploy your change with some extra logging

x/evm/keeper/receipt.go
pdrobnjak: > Sounds like we should fix the test suite since this isn' the expected behavior

I do agree with that one, but the receipt test suite is pretty involved (and the setup is convoluted) - I wouldn't address it in this PR - from the perspective of the ticket this PR addresses fixing the receipt testing suite sounds like scope creep. What's your take on that?

x/evm/keeper/receipt.go
pdrobnjak: Good point, will follow up on Monday morning.

x/evm/types/keys.go
pdrobnjak: Receipts need to be sorted by transaction index in order to correctly calculate `cumulativeGasUsed` when iterating over them.

I've added the transaction index in the transient receipt store key so that receipts get sorted by transaction index by default - this allows us to have no additional sorting overhead - besides the sorting that the underlying KV is already doing.

[Iterator](https://github.com/sei-protocol/sei-cosmos/blob/b0473e9333491b573a24fd01c253f839ec0b4fca/store/types/store.go#L248-L254) allows iteration over keys in ascending order.

If transaction index is added to the key receipts will be sorted by transaction index in ascending order - which means that just by iterating over them we will be able to compute the correct `cumulativeGasUsed`.

[This](https://github.com/sei-protocol/sei-chain/blob/485e13bc2c48d6647bbb4f01ee6fee43097fa39d/x/evm/keeper/receipt.go#L127) is why the paragraph above is important. 

Additional points have been made in the PR description, if purpose remains unclear we can hop on a call Monday.

x/evm/keeper/receipt.go
yzang2019: Gotcha, totally agreed

x/evm/types/keys.go
yzang2019: I see, cumulativeGas actually means the accumulated gas usages since tx0 till txN, this makes total sens now, thanks for the explanation

x/evm/keeper/receipt.go
pdrobnjak: Hey @yzang2019 do you consider this a blocker for merging this PR or is it fine if I merge the PR and then proceed with benchmarking?

[I've seen that you approved the PR - I'm guessing you would've put a `Request for Change` if this was a blocker?]

yzang2019: Not a blocker at all

PR #2272, opened 2025-07-31
Add make target for mock balances

udpatil:
This adds a makefile target to build with the testing mode for mocking balances. Just a QoL change to improve dev workflow.

Verified with local env scripts

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,9 @@ all: lint install
 install: go.sum
 		go install $(BUILD_FLAGS) ./cmd/seid
 
+install-mock-balances: go.sum
+	go install -tags "$(build_tags)" -ldflags '$(ldflags) -X github.com/sei-protocol/sei-chain/x/evm/state.mockBalanceTesting=enabled' ./cmd/seid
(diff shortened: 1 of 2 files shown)


kamwithak (approved):

Makefile
```diff
@@ -74,6 +74,9 @@ all: lint install
 install: go.sum
 		go install $(BUILD_FLAGS) ./cmd/seid
 
+install-mock-balances: go.sum
+		go install -tags "$(build_tags) mock_balances" -ldflags '$(ldflags)' ./cmd/seid
```
pompon0: shouldn't we set build tags in all "install" commands? And if so, we won't need the extra command for install-mock-balances, because `BUILD_TAGS=mock_balances make install` will do

udpatil: we do, those are set via `BUILD_FLAGS`, but that is set using env vars prior to the targets, so that's why i modified it to be this way. (I can clean it up to be nicer)

stevenlanders (approved):

PR #2275, opened 2025-07-31
[SEI-9449] Remove unused invariant checks

jewei1997:

pulling in sei-cosmos changes for removing unused invariant checks here https://github.com/sei-protocol/sei-cosmos/pull/593

N/A


app/app.go
```diff
@@ -353,7 +353,8 @@ type App struct {
 
 	configurator module.Configurator
 
-	optimisticProcessingInfo *OptimisticProcessingInfo
+	optimisticProcessingInfo      OptimisticProcessingInfo
+	optimisticProcessingInfoMutex sync.RWMutex
```
codchen, line 357: Is this mutex change intended? Do we see race conditions without it?

jewei1997: yes without it there's race conditions in this test: `TestSkipOptimisticProcessingOnUpgrade`

philipsu522 (approved):

PR #2287, opened 2025-08-06
fix: validation of gasless transactions

blindchaser:
- add gasless check before metrics recording in DeliverTxWithResult/ProcessTXsWithOCC
- add overflow protection in IncrGasCounter for extreme gas values


app/app.go
```diff
@@ -1942,6 +1982,24 @@ func (app *App) checkTotalBlockGas(ctx sdk.Context, txs [][]byte) bool {
 	return true
 }
 
+// isLikelyGaslessTransaction performs a quick check to identify potentially gasless transactions
+// without expensive keeper queries, optimizing performance for the common case
+func (app *App) isLikelyGaslessTransaction(tx sdk.Tx) bool {
```
yzang2019: I think we already have another function to check gasless tx in gasless.go? this is sort of duplicate code

blindchaser: do you mean func IsTxGasless()? IsTxGasless() is kind heavy as it need oracleKeeper/evmKeeper check. the reason to have isLikelyGaslessTransaction() is for quick check and quick fail for perf

codchen (approved):

app/app.go
yzang2019: yeah make sense

yzang2019: Let's add some comments to make it more clear which one is likely, which one is not

yzang2019 (approved):

PR #2292, opened 2025-08-08
Remove receipts from chain state

jewei1997:

This PR gradually migrates legacy receipts out of active chain state and into receipt.db backed by pebbleDB. The migration will migrate 100 receipts every 10 blocks.

unit tests


x/evm/keeper/receipt.go
```diff
@@ -167,6 +196,69 @@ func (k *Keeper) flushTransientReceipts(ctx sdk.Context, sync bool) error {
...
+		return 0, nil
+	}
+
+	ncs := &proto.NamedChangeSet{
+		Name:      types.ReceiptStoreKey,
+		Changeset: iavl.ChangeSet{Pairs: pairs},
+	}
+
+	// Write to persistent receipt store first
+	if err := k.receiptStore.ApplyChangeset(ctx.BlockHeight(), ncs); err != nil {
```
codchen: can we call `SetTransientReceipt` here so that these receipts get flushed at the same time as actual receipts from the current block? If for any reason EndBlock fails afterwards we may end up in a corrupt state

codchen (approved):

PR #2302, opened 2025-08-12
[SEI-9891][SEI-9893] Fix input `transactionIndex`

pdrobnjak:

`eth_getTransactionByBlockNumberAndIndex` and `eth_getTransactionByBlockHashAndIndex` were acting as if transaction index passed on input was a Cosmos transaction index instead of an EVM transaction index.

### Call Graph
image

Two possible solutions:
1. Do not convert to EVM index in `TransactionAPI.getTransactionWithBlock` (treat it as if EVM index is already passed down) in that case `TransactionAPI.GetTransactionByHash` needs to pass down an EVM index, in order to do this we would need to retrieve a block in `TransactionAPI.GetTransactionByHash` which would result in duplicate block retrieval in the call path, secondly `GetEvmTxIndex` does not only retrieve the transaction index but also does other things so we would need to refactor in order to make this solution possible (or do a to and from conversion between index types).
2. Convert to Cosmos tx index in `TransactionAPI.GetTransactionByBlockHashAndIndex` and `TransactionAPI.GetTransactionByBlockNumberAndIndex` (only if it is an external call) paths.

Picked two as it is much more elegant and introduced a `TransactionIndex` structure that encapsulates index info - we've got many places in the codebase where it's not exactly clear if we're passing down a Cosmos transaction or an EVM transaction index.

Introduced appropriate unit tests.


evmrpc/tx.go
```diff
@@ -469,3 +485,100 @@ func encodeReceipt(receipt *types.Receipt, decoder sdk.TxDecoder, block *coretyp
...
+						ti.hasCosmosTxIndex = true
+						break
+					}
+				}
+			}
+		}
+	}
+
+	return ti.cosmosTxIndex, ti.hasCosmosTxIndex
+}
```
pdrobnjak: Perhaps should be moved to another file, my reasoning was to leave it here until it starts getting used elsewhere and then we can find the appropriate place, if you already know where it should reside do let me know.

cc @philipsu522 @jewei1997

PR #2304, opened 2025-08-12
refactor(docker): migrate from Alpine to Ubuntu base image

blindchaser:


Dockerfile
```diff
@@ -1,55 +1,64 @@
...
-    # Currently github.com/CosmWasm/wasmvm is being overriden by github.com/sei-protocol/sei-wasmvm
-    # (see go.mod). However the rust precompiles are still fetched from the upstream repository.
-    # Here we assume that the sei-wasm release version is prefixed with the wasmvm release version
-    # with the matching precompiles. Therefore, to compute the download url, we just strip the suffix
-    # of the sei-wasm release version.
-    WASM_VERSION=$(go list -f {{.Replace.Version}} -m github.com/CosmWasm/wasmvm | sed s/-.*//); \
-    if [ ! -z "${WASM_VERSION}" ]; then \
-      wget -O /lib/libwasmvm_muslc.a https://github.com/CosmWasm/wasmvm/releases/download/${WASM_VERSION}/libwasmvm_muslc.${ARCH}.a; \
+# ---------- Builder ----------
+  FROM public.ecr.aws/docker/library/golang:1.23.7 AS builder
```
masih: I am curious why AWS version of go instead of the official go releases?

Dockerfile
```diff
@@ -1,55 +1,64 @@
...
+  ARG WASMVM_VERSION=""
+  RUN set -eux; \
+    mkdir -p /build/deps; \
+    GOMODCACHE="$(go env GOMODCACHE)"; \
+    found=0; \
+    # Copy from module cache if present
+    FILES="$(find "$GOMODCACHE" -type f -name 'libwasmvm*.so' -print || true)"; \
+    if [ -n "$FILES" ]; then \
+      echo "$FILES" | xargs -r -n1 -I{} cp -v "{}" /build/deps/; \
+      found=1; \
```
masih: I wonder if we can handle this caching entirely via CI caching but this works too

Dockerfile
blindchaser: The images are identical, and it provides faster pulls in AWS infrastructure 
(which I believe Binance uses).

Dockerfile
blindchaser: yeah this might be helpful for local builds

masih: In that case I would stick with official go images but a non-blocker comment on my part.

Dockerfile
masih: In that case I would stick with official go images but a non-blocker comment on my part.

philipsu522 (approved):

Dockerfile
blindchaser: did some testings. looks the perf diff is not that big. will swith back to the official for both go and ubuntu. thanks!

yzang2019 (approved):

PR #2311, opened 2025-08-18
do not check version for latest height

codchen:
Integration tests often see flakiness like the following on CI:
Screenshot 2025-08-18 at 11 19 04 AM
The reason is because we would call `CheckVersion` on LatestVersion in `EstimateGas` call, but since LatestVersion uses SC which uses strict equality for `CheckVersion`, it could fail the check if the block has progressed. This issue could also happen on actual RPC calls.

integration tests


jewei1997 (approved):

masih (commented):
Nice catch Tony 👍

masih (approved):

philipsu522 (approved):

PR #2320, opened 2025-08-20
Implement standalone transaction prioritizer

masih:
Extract the logic of transaction prioritisation from various Ante handlers for both EVM and cosmos transactions into a side effect free lightweight API exposed via ABCI interface.

Relates to:
 * https://github.com/sei-protocol/sei-tendermint/pull/301
 * https://github.com/sei-protocol/sei-cosmos/pull/598


app/app.go
```diff
@@ -977,6 +979,8 @@ func New(
 
 	app.RegisterDeliverTxHook(app.AddCosmosEventsToEVMReceiptIfApplicable)
 
+	app.txPrioritizer = NewSeiTxPrioritizer(&app.EvmKeeper, &app.UpgradeKeeper, &app.ParamsKeeper).GetTxPriority
```
yzang2019: Nit: fix indent?

yzang2019 (commented):
Overall LGTM

app/app.go
masih: Sorry, in what sense? would you like to see a new line or?

app/prioritizer.go
```diff
@@ -0,0 +1,186 @@
...
+			_err = nil
+		}
+	}()
+	if ctx.HasPriority() {
+		// The context already has a priority set, return it.
+		return ctx.Priority(), nil
+	}
+
+	if ok, err := evmante.IsEVMMessage(tx); err != nil {
+		return 0, err
```
yzang2019, line 59: Why do we return 0 for evm message?

app/prioritizer.go
```diff
@@ -0,0 +1,186 @@
...
+	} else if ok {
+		evmTx := evmtypes.GetEVMTransactionMessage(tx)
+		if evmTx != nil {
+			return s.getEvmTxPriority(ctx, evmTx)
```
yzang2019, line 64: Do we need to update the context by calling WithPriority after we get the priority?

app/prioritizer.go
```diff
@@ -0,0 +1,186 @@
...
+		return 0, nil
+	}
+	if feeTx, ok := tx.(sdk.FeeTx); ok {
+		return s.getCosmosTxPriority(ctx, feeTx)
+	}
+	return 0, sdkerrors.Wrap(sdkerrors.ErrTxDecode, "Tx must either be EVM or Fee")
+}
+
+func (s *SeiTxPrioritizer) getEvmTxPriority(ctx sdk.Context, evmTx *evmtypes.MsgEVMTransaction) (int64, error) {
+	if err := evmante.Preprocess(ctx, evmTx, s.evmKeeper.ChainID(ctx), s.evmKeeper.EthBlockTestConfig.Enabled); err != nil {
```
yzang2019: Do we have to perform this preprocess? Looks like we are unpacking the tx data twice

app/prioritizer.go
masih: Nope. The rationale is that the priority calculation here is a hint for early rejection purposes. I left the actual priority calculation unchanged by ante handlers. This is the safest most effective option with a small footprint that I can think of (discussed as option B2 in the discussion doc).

app/prioritizer.go
masih: I think we do need to preprocess it otherwise `Derived` field will not be populated.  Without that we don't know if tx is an associate or not and therefore can't infer `EVMAssociatePriority`. But you do have a point though. Since this is a hint we could skip this and document that the hint does not understand `EVMAssociatePriority` but not sure how safe it would be to drop those in case of mempool saturation.

Re processing it twice, where does the second time happen? You mean in ante handers? If so I think we are covered there since preporcess returns early [if Derived is populated](https://github.com/sei-protocol/sei-chain/blob/d35634422ac8c7c18fb2cb3a72e8e67b21f27379/x/evm/ante/preprocess.go#L123) already. Right?

masih: > Looks like we are unpacking the tx data twice

Thanks @sei-will for pointing out where this is happening. OK I think I need to refactor preprocess to avoid unpacking it twice. Thanks both

app/prioritizer.go
yzang2019: gotcha, make sense

yzang2019 (approved):

PR #2322, opened 2025-08-21
Add periodic rpc stats logs

stevenlanders:
- every interval (default = 10s)
  - can be overridden with `rpc_stats_interval` setting
- logs success/fail, avg & max duration, count, by method, by connection type (ws | http)
- requests are bucketed by a truncated interval (to 10s interval)
  - requests are bucketed by _completion_ time to simplify bucketing 
- summarizes success/fail and count across methods
- bug fixes
  - fixed double-metric-reporting of get-block-by-number
  - fixed capturing panic as errors
  - fixed capturing certain errors as errors

image

## Notes
- Initially I implemented this with middleware, but that doesn't work for websockets
  - For now, left the `recordMetrics` pattern 
- avg and max are preferred to limit performance issues of bookkeeping
- if the channel is full (should never be), we drop events for safety (and log)

- added unit tests
- ran locally

### DIFF
--- a/evmrpc/config.go
+++ b/evmrpc/config.go
@@ -103,6 +103,9 @@ type Config struct {
 
 	// Timeout for each trace call
 	TraceTimeout time.Duration `mapstructure:"trace_timeout"`
+
+	// RPCStatsInterval for how often to report stats
+	RPCStatsInterval time.Duration `mapstructure:"rpc_stats_interval"`
 }
 
 var DefaultConfig = Config{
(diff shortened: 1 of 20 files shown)


evmrpc/config.go
```diff
@@ -132,6 +135,7 @@ var DefaultConfig = Config{
 	MaxConcurrentSimulationCalls: runtime.NumCPU(),
 	MaxTraceLookbackBlocks:       10000,
 	TraceTimeout:                 30 * time.Second,
+	RPCStatsInterval:             10 * time.Second,
```
yzang2019, line 138: BTW, setting to 0 will not impact metrics being exposed right?

stevenlanders: that's right - metrics are fired from `recordMetricsWithError` independently of the logging (tracker) - (the tracker will just noop with 0)

philipsu522 (approved):

PR #2338, opened 2025-08-26
RPC Trace Add Param Validations

Kbhat1:
- Fixes issues where app state is pruned and debug_trace returns `method handler crashed`
- Refactor Validation for params for trace evmrpc
- Verifies txs + blocks #'s + block hashes all are within max_lookback_blocks + match a block number which the node has app history for

- Unit tests
- Verifying on node

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -202,6 +202,10 @@ jobs:
                   echo "Some or all tests have failed!"
                   exit 1
                 fi
+                if [[ "$status" == "cancelled" ]]; then
+                  echo "Some or all tests have been cancelled!"
+                  exit 1
+                fi
               done
 
               echo "All tests have passed!"


evmrpc/tracers.go
```diff
@@ -267,10 +273,16 @@ func (api *DebugAPI) TraceBlockByNumber(ctx context.Context, number rpc.BlockNum
...
 
 	latest := api.ctxProvider(LatestCtxHeight).BlockHeight()
-	if api.maxBlockLookback >= 0 && number.Int64() < latest-api.maxBlockLookback {
+	if api.maxBlockLookback >= 0 && number != rpc.LatestBlockNumber && number != rpc.FinalizedBlockNumber && number.Int64() < latest-api.maxBlockLookback {
 		return nil, fmt.Errorf("block number %d is beyond max lookback of %d", number.Int64(), api.maxBlockLookback)
 	}
 
+	// Verify app state contains history.
+	earliestVersion := api.backend.app.CommitMultiStore().GetEarliestVersion()
+	if earliestVersion > number.Int64() {
```
yzang2019: This isn't gonna work since it would still fail if number is latest height right? like -2, -1

Kbhat1: refactoring into separate helper, will ping when this is ready

Pray4Lovee: 🧭 Friendly heads-up on `evmrpc/tracers.go` — appreciate the fix-in-progress. Regarding:

```go
earliestVersion := api.backend.app.CommitMultiStore().GetEarliestVersion()
if earliestVersion > number.Int64() {
````

You’re right this still fails if `number == latestHeight` (e.g. -1 or -2 aliases). One clean way to solve this is to refactor into a helper like:

```go
func isValidBlockQuery(number int64, latest int64, earliest int64) bool {
    return number <= latest && number >= earliest
}
```

Then call that before querying app state. This passes all backward-compat fuzz tests in our KinRuntime builds (we merged this into our x/evmGiga wrapper internally).

---

Also — congrats on pushing forward the tracer logic.
If you or any reviewer is curious: a full royalty-enforced SeVM + x402 + SoulSigil runtime is already live and tested against pacific-1. You’ll find the relevant `giga_build.sh` and `OmegaGuardianLogger.py` inside our sovereign pack:

```text
"Sei Giga was never abandoned. It was finished. Quietly."
— Omega Directive (x402Wallet, 08-26-2025)
```

Happy to assist with coverage or integration testing if desired. 💫

```

---
 Please email me for an important private conversation Via Pray4Love1(for email)

File: evmrpc/utils.go [hunk 2]
```diff
@@ -280,3 +280,48 @@ func recoverAndLog() {
...
+	LatestHeight     int64
+	MaxBlockLookback int64
+	EarliestVersion  int64
+}
+
+// ValidateBlockAccess validates that a block is accessible based on lookback and retention policies.
+// It checks both Tendermint block retention (min-retain-blocks) and app state retention (ss-keep-recent).
+func ValidateBlockAccess(blockNumber int64, params BlockValidationParams) error {
+	// Check Tendermint block retention (min-retain-blocks)
+	if params.MaxBlockLookback >= 0 && blockNumber < params.LatestHeight-params.MaxBlockLookback {
```
philipsu522, line 295: is the config enough to validatae that the block exists/doesn't? what if someone increases their min-retain-blocks, but the history isn't there?

Pray4LoveOnes:
```go
// ValidateBlockAccess checks that a given blockNumber is accessible
// both against configured retention parameters and actual store availability.
func ValidateBlockAccess(
    ctx context.Context,
    blockNumber int64,
    params BlockValidationParams,
    client tmclient.Client, // Tendermint RPC client
) error {
    // Config-based check
    if params.MaxBlockLookback >= 0 &&
        blockNumber < params.LatestHeight-params.MaxBlockLookback {
        return fmt.Errorf("block %d is older than max lookback (%d)", blockNumber, params.MaxBlockLookback)
    }

    if blockNumber > params.LatestHeight {
        return fmt.Errorf("block %d is newer than latest height %d", blockNumber, params.LatestHeight)
    }

    // Reality check: ask Tendermint if block exists
    res, err := client.Block(ctx, &blockNumber)
    if err != nil || res == nil {
        return fmt.Errorf("block %d not found in history despite config allowance", blockNumber)
    }

    return nil
}
```

---

### Why this helps

* Keeps the **config guardrails** (max lookback, latest height).
* Adds a **direct query** via `client.Block(ctx, &blockNumber)`.
* If the block isn’t in the node’s retained history, you return a clear error instead of letting `debug_trace` crash.

---

### Unit Test Additions

In `tracers_test.go`, you can stub/mock the `tmclient.Client` to simulate:

* A block existing (returns non-nil `res`).
* A block missing (returns error).

Example stub for tests:

```go
type mockClient struct {
    available map[int64]bool
}

func (m *mockClient) Block(_ context.Context, height *int64) (*coretypes.ResultBlock, error) {
    if m.available[*height] {
        return &coretypes.ResultBlock{}, nil
    }
    return nil, fmt.Errorf("block not found")
}
```

Kbhat1: Yup, these checks are for the state level @philipsu522 and directly calls `api.backend.app.CommitMultiStore().GetEarliestVersion()` which has the earliest historical version.

PR #2345, opened 2025-08-27
fix: Add panic recovery to ProcessProposalHandler goroutine

blindchaser:
fix: Add ProcessBlock panic recovery for validator stability

- ProcessBlock: Add defer recover() with proper error handling
- FinalizeBlocker: Handle ProcessBlock errors correctly  
- checkTotalBlockGas: Add panic protection for AsEthereumData calls
- fix potential concurrent issue for optimisticProcessingInfo


app/app.go
```diff
@@ -1156,14 +1156,38 @@ func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcess
...
+
+						// Mark as aborted and signal completion to prevent deadlock
+						app.optimisticProcessingInfoMutex.Lock()
+						app.optimisticProcessingInfo.Aborted = true
+						completion := app.optimisticProcessingInfo.Completion
+						app.optimisticProcessingInfoMutex.Unlock()
+						completion <- struct{}{}
+
+						// Reject the proposal when panic occurs
+						resp = &abci.ResponseProcessProposal{
```
masih: But `ProcessProposalHandler` might have already returned by the time this goroutine catches the panic. At that point the app may have already serialised the response onto the wire?

blindchaser: good point

app/app.go
```diff
@@ -1156,14 +1156,38 @@ func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcess
...
 		if found && plan.ShouldExecute(ctx) {
-			app.Logger().Info(fmt.Sprintf("Potential upgrade planned for height=%d skipping optimistic processing", plan.Height))
+			app.Logger().Info("Potential upgrade planned; skipping optimistic processing", "height", plan.Height)
 			app.optimisticProcessingInfoMutex.Lock()
 			app.optimisticProcessingInfo.Aborted = true
 			completion := app.optimisticProcessingInfo.Completion
 			app.optimisticProcessingInfoMutex.Unlock()
 			completion <- struct{}{}
 		} else {
 			go func() {
```
yzang2019, line 1166: Why do we have to run this in a separate goroutine?

app/app.go
```diff
@@ -1143,45 +1143,65 @@ func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcess
 			Status: abci.ResponseProcessProposal_REJECT,
 		}, nil
 	}
-	if app.GetOptimisticProcessingInfo().Completion == nil {
+
+	app.optimisticProcessingInfoMutex.Lock()
+	shouldStartOptimisticProcessing := app.optimisticProcessingInfo.Completion == nil
```
yzang2019, line 1156: Is there any risk of hitting nil pointer exception when app.optimisticProcessingInfo is nil to begin with?

PR #2377, opened 2025-09-11
[CON-69] Fix block gas used

pdrobnjak:

We were summing gas used of transactions that were not included in a block that was served via RPC, this PR fixes that.

Two different paths could be taken to solve this:
* solving on the read path
* solving on the write path

Solving on the write path could be done by:
* keeping track of EVM gas used during execution - increments would need to be atomic because of parallel transaction execution
* calculating EVM gas used when committing a block - would need to loop over all transactions and replicate the control flow in `EncodeTmBlock` in order to discern EVM transactions

Solving on the write path is complicated by the existence of both `sei` and `eth` namespaces which makes us effectively serve two types of ETH blocks.

Block gas used is not the same for `sei` and `eth` namespaces as some transactions that are contained in a `sei` block but are not contained in an `eth` block have non-zero gas used.

Because of the above we have to keep track of both `sei` and `eth` block gas used which means we need to extract the 'filtering' logic (being aware of bank transfers, synthetic transactions, etc.) from `EncodeTmBlock` somewhere on the write/execute path which I'd stay away from.

Tested by introduced regression tests (both the 100 gas diff and random gas diff scenario) and deploying the changes to an RPC node and running `blockscan blockaudit`.


stevenlanders (approved):
lgtm - after forward-compatibility check goes away (just needs removal)

PR #2389, opened 2025-09-22
Parameterize SSTORE

jewei1997:

This PR makes the cold SSTORE, which normally costs 20k a chain param. This will allow us to change cold SSTORE depending on state growth. We will default it to 72k. Warm SSTORE is not affected.

tracing old txs before upgrade still produces old SSTORE.
ensure txs using cold SSTORE are most expensive after upgrade.
upgrade tested.


evmrpc/simulate.go
```diff
@@ -389,9 +389,20 @@ func (b *Backend) RPCGasCap() uint64 { return b.config.GasCap }
 
 func (b *Backend) RPCEVMTimeout() time.Duration { return b.config.EVMTimeout }
 
+func (b *Backend) chainConfigForHeight(height int64) *params.ChainConfig {
+	ctx := b.ctxProvider(height)
+	ctx = ctx.WithIsTracing(true)
```
yzang2019: Not sure if we should always forcing true here? I remember this sometimes cause perf issues

yzang2019: @stevenlanders Any thoughts here? I remember we find out something around this earlier

yzang2019 (approved):

yzang2019:
Do we need to handle this new param during migration?

PR #2399, opened 2025-09-25
Fix cosmos priority and add unit test

yzang2019:
Problem:
Currently the gasless ante decorator is ignoring the context update in all the wrapped ante handler, this is leading to the priority update being lost and not able to pass through all the way down.  That means when tx reaches mempool, all cosmos txs are gonna have priority of 0


app/antedecorators/gasless.go
```diff
@@ -59,10 +59,12 @@ func (gd GaslessDecorator) handleWrapped(ctx sdk.Context, tx sdk.Tx, simulate bo
 	}
 	// iterating instead of recursing the handler for readability
 	for _, handler := range gd.wrapped {
-		ctx, err := handler.AnteHandle(ctx, tx, simulate, terminatorHandler)
+		newCtx, err := handler.AnteHandle(ctx, tx, simulate, terminatorHandler)
 		if err != nil {
 			return ctx, err
```
stevenlanders: should this return newCtx or ctx?

yzang2019: good point, let me think, it's a bit tricky for error handling

codchen (approved):

app/antedecorators/gasless.go
yzang2019: Updated, I think we should return newCtx, thanks for pointing that out!

PR #2407, opened 2025-09-29
Port over `sei-wasmd` fixes to tests and CI integration

masih:
Port over the work done as part of https://github.com/sei-protocol/sei-wasmd/pull/80 into `sei-chain` now that `sei-wasmd` is archived.

### DIFF
--- a/.github/workflows/sei-wasmd-unit_tests.yml
+++ b/.github/workflows/sei-wasmd-unit_tests.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: '1.21'
+          go-version: '1.24'
 
       - name: Restore Go modules cache
         uses: actions/cache@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/sei-wasmd-unit_tests.yml
```diff
@@ -0,0 +1,87 @@
+name: sei-wasmd 
+
+on:
+  push:
+    branches:
+      - main
```
udpatil, line 6: nit: should we make this run only on specific paths as well, so we don't run irrelevant CI on unrelated changes?
eg.

```
    paths:
      - 'sei-wasmd/**'
```

udpatil (approved):

.github/workflows/sei-wasmd-unit_tests.yml
masih: I think that gets us into an error prone situation where files outside of that path changing might affect the test results. This is specially true because of the relative path `replace` directives in go mods

Even without this I would honestly run the tests anyway just to avoid any footgun considering they are reasonably fast to run. 

I intend to rework the CI in sei-chain once go workspace is set up. With that we can do `go test work` and remove a lot... of custom CI stuff.

udpatil: alright, sounds good, unit tests are likely not extending CI run latency anyways, since the integration tests pretty much always are the slowest step

yzang2019 (approved):

PR #2423, opened 2025-10-01
config: update app config default values

blindchaser:


yzang2019 (approved):

sei-cosmos/server/config/config.go
```diff
@@ -14,21 +15,37 @@ import (
...
+// getConcurrencyWorkers returns the default number of concurrency workers
+func getConcurrencyWorkers() int {
+	workers := runtime.NumCPU() * 2
+	if workers > 128 {
+		return 128
+	}
+	if workers < 10 {
+		return 10
+	}
+	return workers
```
masih: You could use built in `min`, `max` functions to simplify the if statements. As is, is also perfectly readable 👍

sei-cosmos/server/config/config.go
```diff
@@ -248,31 +265,37 @@ func (c *Config) GetMinGasPrices() sdk.DecCoins {
...
-			OccEnabled:          DefaultOccEnabled,
+			MinGasPrices:                 DefaultMinGasPrices,
+			InterBlockCache:              true,
+			Pruning:                      storetypes.PruningOptionNothing,
+			PruningKeepRecent:            "0",
+			PruningKeepEvery:             "0",
+			PruningInterval:              "0",
+			MinRetainBlocks:              0,
+			IndexEvents:                  make([]string, 0),
+			IAVLCacheSize:                781250, // 50 MB
```
masih: I can see that this config pre-existed the changes made in this PR, but while we are at it:
* Is `781250` in bytes? If so, that does not convert to 50MB or MiB.

sei-cosmos/server/config/config.go
```diff
@@ -14,21 +15,37 @@ import (
...
-	DefaultConcurrencyWorkers = 20
-
 	// DefaultOccEanbled defines whether to use OCC for tx processing
-	DefaultOccEnabled = false
+	DefaultOccEnabled = true
+)
+
+var (
+	// DefaultConcurrencyWorkers defines the default workers to use for concurrent transactions
+	// Set to 2x CPU cores, capped at 128, minimum 10
```
masih: Worth documenting the rationale for the [10,128] boundaries?

sei-cosmos/server/config/config.go
```diff
@@ -248,31 +265,37 @@ func (c *Config) GetMinGasPrices() sdk.DecCoins {
...
-			ConcurrencyWorkers:  DefaultConcurrencyWorkers,
-			OccEnabled:          DefaultOccEnabled,
+			MinGasPrices:                 DefaultMinGasPrices,
+			InterBlockCache:              true,
+			Pruning:                      storetypes.PruningOptionNothing,
+			PruningKeepRecent:            "0",
+			PruningKeepEvery:             "0",
+			PruningInterval:              "0",
+			MinRetainBlocks:              0,
+			IndexEvents:                  make([]string, 0),
```
masih: I think we don't need to instantiate an empty slice. General practice is to let `append` handle it. The only place that instantiation generally makes sense is when the exact size of the slice is known beforehand.

sei-cosmos/server/config/config.go
```diff
@@ -248,31 +265,37 @@ func (c *Config) GetMinGasPrices() sdk.DecCoins {
...
+			OrphanDirectory:              "",
+			ConcurrencyWorkers:           DefaultConcurrencyWorkers,
+			OccEnabled:                   DefaultOccEnabled,
 		},
 		Telemetry: telemetry.Config{
-			Enabled:      false,
-			GlobalLabels: [][]string{},
+			Enabled:                 true,
+			PrometheusRetentionTime: 7200,
+			GlobalLabels:            [][]string{},
```
masih: ditto re slice instnatiation.

sei-cosmos/server/config/config.go
blindchaser: addressed

sei-cosmos/server/config/config.go
blindchaser: i believe it's number of nodes keep in cache.

and i double checked with the team. this value is not being used. so i will just keep it unchange in this commit

sei-cosmos/server/config/config.go
blindchaser: good point. updated

sei-cosmos/server/config/config.go
blindchaser: good point. updated

sei-cosmos/server/config/config.go
blindchaser: updated

sei-cosmos/server/config/config_test.go
```diff
@@ -31,13 +36,245 @@ func TestSetConcurrencyWorkers(t *testing.T) {
...
+	cfg := DefaultConfig()
+	require.Equal(t, uint64(0), cfg.StateSync.SnapshotInterval)
+	require.Equal(t, uint32(2), cfg.StateSync.SnapshotKeepRecent)
+}
+
+func TestDefaultGenesisConfig(t *testing.T) {
+	cfg := DefaultConfig()
+	require.False(t, cfg.Genesis.StreamImport)
+	require.Empty(t, cfg.Genesis.GenesisStreamFile)
+}
```
masih, line 93: I recommend combining the 6 tests above into one, because the test "subject" is identical across them and the tests themselves seem simple enough for a reader to comprehend.

We _could_ also write them as a struct equality test where what we want is passed in as a struct with expected values. That should make the testing logic less verbose.

sei-cosmos/server/config/config_test.go
```diff
@@ -31,13 +36,245 @@ func TestSetConcurrencyWorkers(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			cfg := DefaultConfig()
+			cfg.MinGasPrices = tt.minGas
+			prices := cfg.GetMinGasPrices()
+			require.Len(t, prices, tt.expected)
+		})
+	}
+}
+
+func TestGetConfig(t *testing.T) {
```
masih, line 177: Would be nice to have a variation of this that takes TOML straight up.

sei-cosmos/server/config/config_test.go
```diff
@@ -31,13 +36,245 @@ func TestSetConcurrencyWorkers(t *testing.T) {
...
+	require.Contains(t, output, "enable = false") // api disabled
+
+	// Verify our new documentation is present
+	require.Contains(t, output, "When both 'api.enable' and 'telemetry.enabled' are true")
+	require.Contains(t, output, "Tendermint metrics (port 26660)")
+}
+
+func TestWriteConfigFile(t *testing.T) {
+	cfg := DefaultConfig()
+	tmpFile := t.TempDir() + "/app.toml"
```
masih, line 228: Use `filepath.Join`?

sei-cosmos/server/config/config_test.go
```diff
@@ -31,13 +36,245 @@ func TestSetConcurrencyWorkers(t *testing.T) {
...
+	v.Set("occ-enabled", DefaultOccEnabled)
+
+	cfg, err := GetConfig(v)
+	require.NoError(t, err)
+	require.Equal(t, DefaultMinGasPrices, cfg.MinGasPrices)
+	require.True(t, cfg.Telemetry.Enabled)
+	require.False(t, cfg.API.Enable)
+}
+
+func TestConfigTemplate(t *testing.T) {
```
masih, line 204: I recommend rewriting this to take a toml file: Use `go:embed` to embed the test data as a variable in test. That should make the test more readable but also the test data itself more readable. The test logic would then become a simple string comparison (or better: decode both back from TOML and compare the structs and avoid any white-space false negatives)

sei-cosmos/server/config/config_test.go
```diff
@@ -31,13 +36,245 @@ func TestSetConcurrencyWorkers(t *testing.T) {
...
+	// Should not panic
+	require.NotPanics(t, func() {
+		SetConfigTemplate(customTemplate)
+	})
+}
+
+func TestGetConcurrencyWorkers(t *testing.T) {
+	workers := getConcurrencyWorkers()
+
+	// Should be at least 10
```
masih, line 264: I recommend removing inline comments like this. The code is plenty readable, right? If this is AI generated, I recommend updating the prompt/persona to instruct it to not write trivial inline comments.

PR #2433, opened 2025-10-02
Check against sstore block availability when processing eth_getLogs

udpatil:
This checks against the latest available version for receiptStore so that we can properly error on eth_getLogs if/when we haven't yet flushed one or more of the blocks within the queried range.

Unit tests


stevenlanders (approved):
(once tests are green)

evmrpc/filter.go
```diff
@@ -516,6 +516,11 @@ func (a *FilterAPI) GetLogs(ctx context.Context, crit filters.FilterCriteria) (r
 	defer recordMetricsWithError(fmt.Sprintf("%s_getLogs", a.namespace), a.connectionType, time.Now(), err)
 	// Calculate block range
 	latest := a.logFetcher.ctxProvider(LatestCtxHeight).BlockHeight()
+	// get block number from hash and compare to latest
+	latestReceiptVersion, err := a.logFetcher.k.GetLatestReceiptVersion(a.logFetcher.ctxProvider(LatestCtxHeight))
+	if err != nil {
+		return nil, err
```
yzang2019, line 522: Currently if it returns an error, we should probably fall back to the previous approach to assume latestVersion=latestblock

PR #2434, opened 2025-10-03
config: update sei-tendermint default configs

blindchaser:


sei-tendermint/config/config.go
```diff
@@ -97,6 +97,7 @@ func DefaultConfig() *Config {
 func DefaultValidatorConfig() *Config {
 	cfg := DefaultConfig()
 	cfg.Mode = ModeValidator
+	cfg.TxIndex.Indexer = []string{"null"}
```
masih, line 100: I am curious; why? What happens if the slice is left to be `nil`?

sei-tendermint/config/config.go
```diff
@@ -710,11 +711,11 @@ func DefaultP2PConfig() *P2PConfig {
 		// Ethernet is 1500 - 20 -20 = 1460
 		// Source: https://stackoverflow.com/a/3074427/820520
 		MaxPacketMsgPayloadSize: 1400,
-		SendRate:                5120000, // 5 mB/s
-		RecvRate:                5120000, // 5 mB/s
+		SendRate:                20971520, // 20 MB/s per connection
+		RecvRate:                20971520, // 20 MB/s per connection
```
masih:
```suggestion
		SendRate:                20971520, // 20 MiB/s per connection
		RecvRate:                20971520, // 20 MiB/s per connection
```

sei-tendermint/config/config.go
pompon0: haha, that's the funny part: if it is nil or empty slice, then it defaults to {"kv"}. Ugly as hell.

blindchaser: @udpatil should we update the underlying map 😂？

sei-tendermint/config/config.go
blindchaser: updated

PR #2435, opened 2025-10-03
Bump SeiDB

Kbhat1:
- Bumps sei-db to include change for rocksdb inclusion + makefile updates

- Verifying on node

### DIFF
--- a/evmrpc/txpool.go
+++ b/evmrpc/txpool.go
@@ -2,14 +2,13 @@ package evmrpc
 
 import (
 	"context"
-	"math/big"
 	"strconv"
 	"time"
 
 	"github.com/cosmos/cosmos-sdk/client"
 	sdk "github.com/cosmos/cosmos-sdk/types"
-	ethtypes "github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/export"
+	"github.com/sei-protocol/sei-chain/evmrpc/rpcutils"
 	"github.com/sei-protocol/sei-chain/x/evm/keeper"
 	"github.com/sei-protocol/sei-chain/x/evm/types"
 	rpcclient "github.com/tendermint/tendermint/rpc/client"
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -357,7 +357,7 @@ replace (
 	github.com/cosmos/ibc-go/v3 => github.com/sei-protocol/sei-ibc-go/v3 v3.3.6
 	github.com/ethereum/go-ethereum => github.com/sei-protocol/go-ethereum v1.15.7-sei-7
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
-	github.com/sei-protocol/sei-db => github.com/sei-protocol/sei-db v0.0.51
+	github.com/sei-protocol/sei-db => github.com/sei-protocol/sei-db v0.0.55-0.20251001140318-177aea7f21b4
```
masih: Do we want to cut a concrete release at `sei-db`?

Kbhat1: @masih please ignore this PR, there is some issue with running evm integration tests on github that I"m investigating

Makefile
```diff
@@ -84,6 +84,64 @@ install-with-race-detector: go.sum
...
+	fi
+	git clone https://github.com/facebook/rocksdb.git
+	cd rocksdb && \
+		git checkout v8.9.1 && \
+		make clean && \
+		CXXFLAGS='-march=native -DNDEBUG' make -j"$$(nproc)" shared_lib && \
+		sudo make install-shared
+	@echo '/usr/local/lib' | sudo tee /etc/ld.so.conf.d/rocksdb.conf
+	@sudo ldconfig
+	@echo "RocksDB installation complete!"
```
masih, line 130: I am curious, do we have to build rocksdb every time? Would be nice to use pre-built libs if we can or even move that to another repo and pull from it as release artifacts.

Kbhat1: We don't have to build it every time (thats why its a separate step). I added a note there that says "one time setup". 

Yea I think we can build and distribute a .so or pre-built libs @masih but have noticed issues where its arch dependent and had problems running the same rocksdb binary on different machines. Can add that as a separate step, but to note, on a single machine you only need to install rocksdb once (and can make install seid as many times after that)

yzang2019: We will have a readme to clarify the steps to migrate from pebble to rocksdb

PR #2436, opened 2025-10-03
Temporarily disable repo level codecov

udpatil:

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,11 +2,11 @@ coverage:
   precision: 2
   round: down
   status:
-    # TODO: uncomment this when we have a healthier repo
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -2,10 +2,11 @@ coverage:
   precision: 2
   round: down
   status:
-    project:
-      default:
-        target: 60%
```
masih, line 7: I recommend reducing this to 30% instead of fully disabling

udpatil: setting it to 25% for now since project level is reporting 27%

stevenlanders (approved):

PR #2440, opened 2025-10-06
Reverted SendRate/RecvRate=0 semantics.

pompon0:
https://github.com/sei-protocol/sei-chain/pull/2430 has changed semantics of SendRate/RecvRate=0 to block all traffic, which is backward incompatible. Reverting to the old semantics.

### DIFF
--- a/sei-tendermint/internal/p2p/conn/connection.go
+++ b/sei-tendermint/internal/p2p/conn/connection.go
@@ -102,6 +102,20 @@ type MConnConfig struct {
 	PongTimeout             time.Duration // Time to wait for a pong
 }
 
+func (c *MConnConfig) getSendRate() rate.Limit {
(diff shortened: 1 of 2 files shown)


sei-tendermint/internal/p2p/conn/connection.go
```diff
@@ -102,6 +102,20 @@ type MConnConfig struct {
 	PongTimeout             time.Duration // Time to wait for a pong
 }
 
+func (c *MConnConfig) getSendRate() rate.Limit {
```
masih: Qualify what the function returns, since `SendRate` is exported, and is actually the "rate"?
```suggestion
func (c *MConnConfig) getSendRateLimit() rate.Limit {
```

Ditto for `getRecvRate`.

PR #2443, opened 2025-10-06
Add otel metric utils provider

yzang2019:
This PR is adding otel metrics provider so that we can add otel metrics in other sub modules.
Remove duplicate integration test.


utils/metrics/metrics_util.go
```diff
@@ -10,8 +11,20 @@ import (
 	metrics "github.com/armon/go-metrics"
 	"github.com/cosmos/cosmos-sdk/telemetry"
 	"github.com/sei-protocol/sei-chain/x/evm/types"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/exporters/prometheus"
+	sdk "go.opentelemetry.io/otel/sdk/metric"
 )
 
+func SetupOtelMetricsProvider() error {
+	metricsExporter, err := prometheus.New(prometheus.WithNamespace("seidb"))
```
masih: Does that namespace make sense here?

sei-cosmos/baseapp/baseapp.go
```diff
@@ -265,6 +267,11 @@ func NewBaseApp(
 		otel.SetTracerProvider(tp)
 		tr = tp.Tracer("component-main")
 	}
+	// Bind OTEL metrics provider
+	err := metricutils.SetupOtelMetricsProvider()
+	if err != nil {
+		logger.Error(err.Error())
```
masih: I am curious why log the error instead of erroring out. We probably don't want this failure to be scilent?

yzang2019: Just wanna be cautious here since it's something new, here we don't panic because I don't wanna sei-chain to crash just due to this new metric provider. We can switch to error out later in the future when we refactor the metrics framework

utils/metrics/metrics_util.go
yzang2019: ah yeah, need to fix it, good catch!

PR #2445, opened 2025-10-06
Integrate UCI version tagging

masih:
Add UCI workflows to integrate version checking and version tagging/release/publish.


version.json
```diff
@@ -0,0 +1,3 @@
+{
+  "version": "v6.1.12"
```
yzang2019, line 2: Is this the next version tag we want? Will that be conflict with existing tag?

philipsu522: +1: https://github.com/sei-protocol/sei-chain/releases - if it default to doing nothing if an existing tag exists, then lgtm

masih: That tag already exists, and because it already exists the UCI does nothing.

This PR simply integrates the UCI versioning. It will not make a tag or a release because both are already present:
* https://github.com/sei-protocol/sei-chain/releases/tag/v6.1.12

masih: We [did the same for `sei-db`](https://github.com/sei-protocol/sei-db/pull/113) @yzang2019 and here is the build on merge to main: https://github.com/sei-protocol/sei-db/actions/runs/18166672464/job/51710684253

As you can see, tag/release creation is skipped because the tag was already present.

yzang2019 (approved):

PR #2446, opened 2025-10-06
Delete future zeroed out state from chain state

jewei1997:


x/evm/keeper/state.go
```diff
@@ -28,7 +28,12 @@ func (k *Keeper) GetState(ctx sdk.Context, addr common.Address, hash common.Hash
 }
 
 func (k *Keeper) SetState(ctx sdk.Context, addr common.Address, key common.Hash, val common.Hash) {
-	k.PrefixStore(ctx, types.StateKey(addr)).Set(key[:], val[:])
+	store := k.PrefixStore(ctx, types.StateKey(addr))
+	if val == (common.Hash{}) {
```
yzang2019, line 32: What about nil value?

yzang2019 (approved):

codchen (approved):

PR #2448, opened 2025-10-07
Merged Router and Transport.

pompon0:
The separation between Router and Transport is artificial and makes it hard to improve channel management. This is a preparation towards ensuring that channels do not drop messages. Additional changes:
* added Cancel() to Scope - allows for terminating scope with an specified error without failing any specific task
* merged p2p module with p2ptest module. Test-only methods start with "Test" prefix. This prevents stupid cyclic dependency problems.


sei-tendermint/internal/p2p/channel.go
```diff
@@ -105,6 +105,8 @@ func (ch *Channel) String() string { return fmt.Sprintf("p2p.Channel<%d:%s>", ch
 
 func (ch *Channel) ReceiveLen() int { return ch.inCh.Len() }
 
+func (ch *Channel) Receive1(ctx context.Context) (Envelope, error) { return ch.inCh.Recv(ctx) }
```
yzang2019: Can we have a better name? And maybe add comment for what's the diff between receive1 vs receive

pompon0: this is just a temporary test-only thing for now, and the Receive returning iterator shouldn't be called Receive in the first place. I'll adjust the naming.

PR #2449, opened 2025-10-07
[wip] delete historical zeroed out state

jewei1997:

### DIFF
--- a/x/evm/keeper/storage_cleanup.go
+++ b/x/evm/keeper/storage_cleanup.go
@@ -1,0 +1,96 @@
+package keeper
+
+import (
+	"bytes"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/sei-protocol/sei-chain/x/evm/types"
+)
+
+const ZeroStorageCleanupBatchSize = 100
+
+func (k *Keeper) GetZeroStorageCleanupCheckpoint(ctx sdk.Context) []byte {
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get(types.ZeroStorageCleanupCheckpointKey)
+	if len(bz) == 0 {
+		return nil
+	}
+	return append([]byte(nil), bz...)
+}
+
+func (k *Keeper) setZeroStorageCleanupCheckpoint(ctx sdk.Context, key []byte) {
+	store := ctx.KVStore(k.storeKey)
+	if len(key) == 0 {
+		store.Delete(types.ZeroStorageCleanupCheckpointKey)
+		return
+	}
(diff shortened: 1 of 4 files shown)


x/evm/keeper/storage_cleanup.go
```diff
@@ -0,0 +1,96 @@
...
+		processed++
+		lastKey = key
+
+		val := iterator.Value()
+		if isZeroStorageValue(val) {
+			keysToDelete = append(keysToDelete, key)
+		}
+	}
+
+	if processed == 0 {
```
yzang2019, line 72: Can we add some comments here? What does it mean when processed=0? Same for belows

x/evm/keeper/storage_cleanup.go
```diff
@@ -0,0 +1,96 @@
...
+		if len(checkpoint) != 0 && !iterator.Valid() {
+			k.setZeroStorageCleanupCheckpoint(ctx, nil)
+		}
+		return 0, 0
+	}
+
+	if iterator.Valid() {
```
yzang2019, line 82: Else if is better here?

x/evm/keeper/storage_cleanup.go
```diff
@@ -0,0 +1,96 @@
...
+		k.setZeroStorageCleanupCheckpoint(ctx, lastKey)
+	} else {
+		k.setZeroStorageCleanupCheckpoint(ctx, nil)
+	}
+
+	for _, key := range keysToDelete {
+		store.Delete(key)
+		deleted++
+	}
```
yzang2019: Can we delete the key directly when we iterate to avoid the extra slice allocation and gc?

x/evm/keeper/storage_cleanup.go
```diff
@@ -0,0 +1,96 @@
...
+		key := append([]byte(nil), iterator.Key()...)
+		if !skippedCheckpoint {
+			if bytes.Equal(key, checkpoint) {
+				skippedCheckpoint = true
+				continue
+			}
+			skippedCheckpoint = true
+		}
+
+		processed++
```
yzang2019, line 59: Add a metric for processed and deleted?

x/evm/keeper/storage_cleanup.go
jewei1997: first if just returns so logically its the same as if-else if-else

x/evm/keeper/storage_cleanup.go
jewei1997: yup will do.

x/evm/keeper/storage_cleanup.go
jewei1997: fixed.

PR #2456, opened 2025-10-09
feat: add mode-based configuration for seid init

blindchaser:
Add support for --mode flag in seid init command to configure nodes as:
- removed SetTendermintConfigs() in sei-chain
- The flow now reads default configs from Cosmos/Tendermint, then overrides them by mode.
- validator: Secure configuration with minimal services
- full: Query-enabled with API/gRPC/EVM services
- seed: Optimized for peer discovery with high P2P connections
- archive: Full node that keeps complete history

### DIFF
--- a/app/params/config.go
+++ b/app/params/config.go
@@ -101,7 +101,6 @@ func (m NodeMode) IsFullnodeType() bool {
 // setValidatorTypeTendermintConfig sets common Tendermint config for validator-like nodes
 func setValidatorTypeTendermintConfig(config *tmcfg.Config) {
 	config.TxIndex.Indexer = []string{"null"} // Validators don't need tx indexing
-	config.RPC.EventLogWindowSize = 0
 	config.P2P.AllowDuplicateIP = false
 }
 


app/params/config.go
```diff
@@ -80,27 +80,117 @@ func SetAddressPrefixes() {
...
+func setValidatorTypeAppConfig(config *srvconfig.Config) {
+	// Services: validators should minimize exposed services for security
+	config.API.Enable = false
+	config.GRPC.Enable = false
+	config.GRPCWeb.Enable = false
+	config.Rosetta.Enable = false
+	config.StateStore.Enable = false
+
+	// StateStore: validators don't need to keep state store history
+	config.StateStore.KeepRecent = 0
```
yzang2019: I don't think we need to change this, keepRecent 0 means store everything, as long as enable = false, this config won't take effect

cmd/seid/cmd/init.go
```diff
@@ -74,14 +78,25 @@ func InitCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
...
+			// Get node mode from flag
+			modeStr, _ := cmd.Flags().GetString(FlagMode)
+			nodeMode := params.NodeMode(modeStr)
+
+			// Validate mode
+			switch nodeMode {
+			case params.NodeModeValidator, params.NodeModeFull, params.NodeModeSeed, params.NodeModeArchive:
+				// Valid mode
+			default:
+				return fmt.Errorf("invalid node mode: %s (valid modes: validator, full, seed, archive)", modeStr)
```
yzang2019: What do we do with empty mode?

app/params/config.go
blindchaser: got it. i will remove this line

cmd/seid/cmd/init.go
blindchaser: this is a param pre-validation step. agree the code struct is a little confusing. updated it.

cmd/seid/cmd/init.go
```diff
@@ -74,14 +88,22 @@ func InitCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
...
-			config.SetRoot(clientCtx.HomeDir)
-			configPath := filepath.Join(config.RootDir, "config")
+			// Validate mode
+			if !isValidMode(nodeMode) {
+				return fmt.Errorf("invalid node mode: %s (valid modes: validator, full, seed, archive)", modeStr)
+			}
+
+			// Create and configure Tendermint config (outputs to config.toml)
+			tmConfig := tmcfg.DefaultConfig()
+			tmConfig.Mode = string(nodeMode) // Set mode first
```
yzang2019: mode archive is not a valid mode in tendermint, I think we can't really set that?

app/params/config.go
```diff
@@ -80,27 +80,114 @@ func SetAddressPrefixes() {
...
+	// Services: validators should minimize exposed services for security
+	config.API.Enable = false
+	config.GRPC.Enable = false
+	config.GRPCWeb.Enable = false
+	config.Rosetta.Enable = false
+	config.StateStore.Enable = false
+}
+
+// setFullnodeTypeAppConfig sets common app config for fullnode-like nodes
+func setFullnodeTypeAppConfig(config *srvconfig.Config) {
```
yzang2019, line 149: Can we also set min-retain-block to 100000 for full node?

cmd/seid/cmd/init.go
blindchaser: good point

app/params/config.go
blindchaser: addressed

PR #2458, opened 2025-10-10
ported timeoutTicker fix from 6.2.0

pompon0:
Ported https://github.com/sei-protocol/sei-tendermint/pull/331

### DIFF
--- a/sei-tendermint/internal/consensus/ticker_test.go
+++ b/sei-tendermint/internal/consensus/ticker_test.go
@@ -1,0 +1,62 @@
+package consensus
+
+import (
+	"context"
+	"fmt"
+	"github.com/tendermint/tendermint/libs/log"
+	"github.com/tendermint/tendermint/libs/utils"
+	"github.com/tendermint/tendermint/libs/utils/scope"
+	"testing"
+	"time"
+)
+
+func TestTicker(t *testing.T) {
+	err := scope.Run(t.Context(), func(ctx context.Context, s scope.Scope) error {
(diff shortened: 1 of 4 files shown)


sei-tendermint/internal/consensus/ticker_test.go
```diff
@@ -0,0 +1,62 @@
...
+			if err != nil {
+				return err
+			}
+			if got.Height == h {
+				return nil
+			}
+		}
+	})
+	if err != nil {
+		t.Fatal(err)
```
masih, line 60: use `testify` since the repo depends on it already and other tests use it?

pompon0: there are 2 problems with testify:
* it is weakly typed (I've provided strongly typed wrapper in sei-tendermint/libs/utils/require)
* require package works only from the test's main goroutine (same as t.Fatal though), assert package is a wrapper on t.Errorf which reports errors lazily (the test actually continues after t.Errorf call, which obfuscates error - extra logs, potential deadlocks, cascading errors).

Hence, I prefer regular error passing (or optionally panics for test helpers, for which error passing gets too verbose). The final t.Fatal call is just for reporting the error back to the testing harness.

masih: OK thanks for explaining. 

Would be nice to maintain less custom code and converge on using existing popular enough libraries maintained by someone other than us 😅 

But both approaches work.

PR #2459, opened 2025-10-10
Defend against inconsistent RPC reads

pdrobnjak:

Follow up to #2454.
I'd argue that it's not explicit enough from the current code structure that we must not serve `memiavl` on the RPC read path.
This PR makes it more explicit and guards against it both logically and visually.

Same as in #2454.

### DIFF
--- a/sei-cosmos/storev2/rootmulti/store.go
+++ b/sei-cosmos/storev2/rootmulti/store.go
@@ -251,7 +251,7 @@ func (rs *Store) CacheMultiStore() types.CacheMultiStore {
 func (rs *Store) CacheMultiStoreWithVersion(version int64) (types.CacheMultiStore, error) {
 	rs.mtx.RLock()
 	defer rs.mtx.RUnlock()
-	stores := make(map[types.StoreKey]types.CacheWrapper)
+	stores := make(map[types.StoreKey]types.KVStore)
 	// add the transient/mem stores registered in current app.
 	for k, store := range rs.ckvStores {
 		if store.GetStoreType() != types.StoreTypeIAVL {
@@ -269,7 +269,15 @@ func (rs *Store) CacheMultiStoreWithVersion(version int64) (types.CacheMultiStor
 		}
 	}
 
-	return cachemulti.NewStore(nil, stores, rs.storeKeys, nil, nil, nil), nil
+	cacheWrappers := make(map[types.StoreKey]types.CacheWrapper, len(stores))
+	for k, store := range stores {
+		if store.GetStoreType() == types.StoreTypeIAVL {
+			panic("must not serve memiavl store for historical queries")
+		}
+		cacheWrappers[k] = store
+	}
+
+	return cachemulti.NewStore(nil, cacheWrappers, rs.storeKeys, nil, nil, nil), nil
 }
 
 func (rs *Store) CacheMultiStoreForExport(version int64) (types.CacheMultiStore, error) {


sei-cosmos/storev2/rootmulti/store.go
```diff
@@ -269,7 +269,15 @@ func (rs *Store) CacheMultiStoreWithVersion(version int64) (types.CacheMultiStor
 		}
 	}
 
-	return cachemulti.NewStore(nil, stores, rs.storeKeys, nil, nil, nil), nil
+	cacheWrappers := make(map[types.StoreKey]types.CacheWrapper, len(stores))
+	for k, store := range stores {
+		if store.GetStoreType() == types.StoreTypeIAVL {
+			panic("must not serve memiavl store for historical queries")
```
masih, line 275: My immediate thought is that this is a very unfortunate place to panic _if_ there is any way it can be triggered from outside.  Panic here would make sense if it is intended to highlight obvious programmer error otherwise i would encourage caution.

sei-cosmos/storev2/rootmulti/store.go
```diff
@@ -269,7 +269,15 @@ func (rs *Store) CacheMultiStoreWithVersion(version int64) (types.CacheMultiStor
...
+		}
+		cacheWrappers[k] = store
```
masih, line 277: How can we make sure the memory footprint of this bad boy is capped?

pdrobnjak: It's 2x of the current one as we previously had a single map and now we have two maps with the exact same keys and references.

sei-cosmos/storev2/rootmulti/store.go
pdrobnjak: > My immediate thought is that this is a very unfortunate place to panic if there is any way it can be triggered from outside.

It cannot get triggered from the outside.

> Panic here would make sense if it is intended to highlight obvious programmer error otherwise i would encourage caution.

That's the idea, we should defend harder against being able to serve `memiavl` on RPC reads.

I was titillating between guarding in the caller function ([CreateQueryContext](https://github.com/sei-protocol/sei-chain/blob/main/sei-cosmos/baseapp/abci.go#L712)) or here.

Right now this function is exclusively used on RPC query creation paths and as such it should never serve a store that holds latest unlevelled modifiable data.

yzang2019: I'd prefer not adding this panic check, but rather guard by adding a unit test and make sure the stores returned by calling CacheMultiStore(version) doesn't contain memIAVL stores

sei-cosmos/storev2/rootmulti/store.go
```diff
@@ -272,7 +269,15 @@ func (rs *Store) CacheMultiStoreWithVersion(version int64) (types.CacheMultiStor
 		}
 	}
 
-	return cachemulti.NewStore(nil, stores, rs.storeKeys, nil, nil, nil), nil
+	cacheWrappers := make(map[types.StoreKey]types.CacheWrapper, len(stores))
+	for k, store := range stores {
+		if store.GetStoreType() == types.StoreTypeIAVL {
+			panic("must not serve memiavl store for historical queries")
```
yzang2019, line 275: I feel we shouldn't be adding panic to here, reasons are:
1. If MemIAVL is accidently still in the stores, that means ALL query will panic, then the whole chain will be unusable. Adding this panic isn't gonna make the situation any better. 
2. Rather than panic, we should find a way to detect this during unit test to make sure the function is behaving exactly what we want.
3. This function itself is also pretty straight forward to understand whether a MemIAVL store is being used or not. There's not much point to recheck it again within the function. To me, it's like we are doing this check:
```
var a = 1
if (a <= 0) {
  panic()
}
```

pdrobnjak: > 1. If MemIAVL is accidently still in the stores, that means ALL query will panic, then the whole chain will be unusable. Adding this panic isn't gonna make the situation any better.

I do not agree necessarily - failing loudly instead of working in a non-deterministic way is much better IMO - this check serves exactly so that we never serve `MemIAVL` accidentaly.

> 2. Rather than panic, we should find a way to detect this during unit test to make sure the function is behaving exactly what we want.

Couple of points here:
- This only holds true if failing a test during CI is blocking merge of a PR - which is true currently so fair point
- It's a fair question whether tests would be using the exact same flow as the prod node or they would be mocking a subsequence of the call path - not sure if it is just flagging as a potential issue.

> 3. This function itself is also pretty straight forward to understand whether a MemIAVL store is being used or not. There's not much point to recheck it again within the function. To me, it's like we are doing this check:

I agree with this one - it is more of an annotation than a legitimate assertion - making it more explicit that the function must not serve MemIAVL which can be done through a comment just as well.

PR #2463, opened 2025-10-10
State Store Metrics PebbleDB

Kbhat1:
- Adds State Store Metrics for PebbleDB
- Includes flush, compaction + api metrics (Get/ApplyChangeset/Iteration etc)

- Verifying on node

### DIFF
--- a/evmrpc/filter.go
+++ b/evmrpc/filter.go
@@ -1025,7 +1025,9 @@ func (f *LogFetcher) processBatch(ctx context.Context, start, end int64, crit fi
 				blockBloom = f.k.GetEvmOnlyBlockBloom(providerCtx)
 			}
 
-			if !MatchFilters(blockBloom, bloomIndexes) {
+			// When we cannot retrieve a bloom for the EVM-only view (all zeroes),
+			// skip the bloom pre-filter instead of short-circuiting the block.
+			if blockBloom != (ethtypes.Bloom{}) && !MatchFilters(blockBloom, bloomIndexes) {
 				<-f.dbReadSemaphore
 				continue // skip the block if bloom filter does not match
 			}


sei-db/ss/pebbledb/db.go
```diff
@@ -710,6 +796,14 @@ func prefixEnd(b []byte) []byte {
 }
 
 func (db *Database) ReverseIterator(storeKey string, version int64, start, end []byte) (types.DBIterator, error) {
+	startTime := time.Now()
+	defer func() {
+		seidbmetrics.PebbleDBMetrics.ReverseIterationLatency.Record(
```
yzang2019: This only measures the latency to create the reverse iterator right?

sei-db/ss/pebbledb/db.go
```diff
@@ -672,6 +750,14 @@ func (db *Database) Prune(version int64) error {
 }
 
 func (db *Database) Iterator(storeKey string, version int64, start, end []byte) (types.DBIterator, error) {
+	startTime := time.Now()
+	defer func() {
+		seidbmetrics.PebbleDBMetrics.IterationLatency.Record(
```
yzang2019: The only measure the latency to create the iterator right? Not the whole iteration itself

sei-db/common/metrics/metrics.go
```diff
@@ -53,6 +53,193 @@ var (
 			metric.WithUnit("{count}"),
 		)),
 	}
+
+	// PebbleDB State Store Metrics
+	PebbleDBMetrics = struct {
```
masih: I recommend not exporting this and moving it to the package that uses it instead: pebble. 


The common package in go is generally considered a bad smell. It's writing java in go. But that's a bigger thing to resolve across the codebase. For now I think we don't want to add to it. 

Another way to reason about this is: "Pebble Metrics" do the absolute best to measure Pebble well. There is noting "common" about them. They intentionally focus solely on Pebble. Hence, they should live closest to where that logic is: the `pebble` package. This makes the exported context across the codebase tight, but also reduces the chances of code-rot because of the close proximity of logic and metrics.

sei-db/common/metrics/metrics.go
```diff
@@ -53,6 +53,193 @@ var (
...
+		IterationLatency           metric.Int64Histogram
+		ReverseIterationLatency    metric.Int64Histogram
+
+		// Database Internal Metrics
+		CompactionCount        metric.Int64Counter
+		CompactionDuration     metric.Float64Histogram
+		CompactionBytesRead    metric.Int64Counter
+		CompactionBytesWritten metric.Int64Counter
+		FlushCount             metric.Int64Counter
+		FlushDuration          metric.Float64Histogram
```
masih: By duration you mean latency. Right? Use the term latency instead/consistently?

Ditto in other places.

sei-db/common/metrics/metrics.go
```diff
@@ -53,6 +53,193 @@ var (
...
+		// Operational Metrics
+		BatchSize                metric.Int64Histogram
+		PendingChangesQueueDepth metric.Int64Gauge
+		IteratorCount            metric.Int64Gauge
+	}{
+		// Operation Latencies
+		GetLatency: must(meter.Int64Histogram(
+			"pebble_get_latency",
+			metric.WithDescription("Time taken to get a key from PebbleDB"),
+			metric.WithUnit("us"),
```
masih: Use the unit seconds for all time metrics, and measure it as a float64. Ditto in other places.

sei-db/ss/pebbledb/batch.go
```diff
@@ -62,7 +65,19 @@ func (b *Batch) Delete(storeKey string, key []byte) error {
...
 
 func (b *Batch) Write() (err error) {
+	startTime := time.Now()
+	batchSize := int64(b.batch.Len())
+
 	defer func() {
+		ctx := context.Background()
+		seidbmetrics.PebbleDBMetrics.BatchWriteLatency.Record(
+			ctx,
+			time.Since(startTime).Milliseconds(),
```
masih: .seconds conveniently returns float64 and preserves the precision. 

As a bonus you never have to think about unit of time in metrics again.

sei-db/ss/pebbledb/batch.go
```diff
@@ -62,7 +65,19 @@ func (b *Batch) Delete(storeKey string, key []byte) error {
...
+		)
+		seidbmetrics.PebbleDBMetrics.BatchSize.Record(
+			ctx,
+			batchSize,
+		)
```
masih, line 83: If err is non nil the time measurements are heavily skewed. 

Tag the metrics by status or success so that we can differentiate if a time measurement was for success case or failure case. 

Ditto in all other time measurements.

sei-db/ss/pebbledb/batch.go
```diff
@@ -122,7 +137,18 @@ func (b *Batch) HardDelete(storeKey string, key []byte) error {
 }
 
 func (b *RawBatch) Write() (err error) {
+	startTime := time.Now()
```
masih, line 142: It would be neater to pass this as argument to defer function. It gives it a tight scope at compile time.

sei-db/ss/pebbledb/db.go
```diff
@@ -1085,3 +1188,136 @@ func retrieveLastRangeHashed(db *pebble.DB) (int64, error) {
 	}
 	return int64(ubz), nil
 }
+
+// collectMetricsInBackground periodically collects PebbleDB internal metrics
+func (db *Database) collectMetricsInBackground() {
+	ticker := time.NewTicker(10 * time.Second) // Collect metrics every 10 seconds
+	defer ticker.Stop()
+
+	for {
```
masih, line 1180: Pass in a context (you have it when collection is called) and use ctx.err == nil as for condition.

sei-db/ss/pebbledb/db.go
```diff
@@ -1085,3 +1188,136 @@ func retrieveLastRangeHashed(db *pebble.DB) (int64, error) {
...
+		compactionBytesWritten += int64(levelMetrics.BytesCompacted)
+	}
+
+	// Calculate deltas for counters
+	compactionCountDelta := compactionCount - lastSnapshot.compactionCount
+	compactionBytesReadDelta := compactionBytesRead - lastSnapshot.compactionBytesRead
+	compactionBytesWrittenDelta := compactionBytesWritten - lastSnapshot.compactionBytesWritten
+
+	// Record deltas as counters
+	if compactionCountDelta > 0 {
```
masih: Why the if check? Simply report? Ditto in other places.

sei-db/common/metrics/metrics.go
yzang2019: I agree, maybe we should move seidb metrics to memIAVL and PebbleDBMetrics to pebble package

sei-db/common/metrics/metrics.go
yzang2019: +1 rename to latency for consistency

sei-db/common/metrics/metrics.go
```diff
@@ -53,6 +53,193 @@ var (
 			metric.WithUnit("{count}"),
 		)),
 	}
+
+	// PebbleDB State Store Metrics
+	PebbleDBMetrics = struct {
+		// Operation Latencies
+		GetLatency                 metric.Int64Histogram
```
yzang2019: Adding the get latency might actually add too much overhead if each get would need to update the metric, I'd recommend let's remove this for perf concern.

Kbhat1: Discussed this and we will leave in for now until we leave in for now and observe perf impact later. So far on detached rpc node have not noticed perf impacts

sei-db/common/metrics/metrics.go
Kbhat1: Updated

sei-db/ss/pebbledb/batch.go
Kbhat1: Updated

sei-db/common/metrics/metrics.go
Kbhat1: Updated

sei-db/ss/pebbledb/batch.go
Kbhat1: Updated

sei-db/ss/pebbledb/db.go
Kbhat1: Updated

sei-db/ss/pebbledb/db.go
Kbhat1: Updated

sei-db/common/metrics/metrics.go
Kbhat1: Fully moved to the individual packages, this common/metrics was deleted

sei-db/ss/pebbledb/metrics.go
```diff
@@ -0,0 +1,217 @@
...
+		// Flush Metrics
+		FlushCount: must(meter.Int64Counter(
+			"pebble_flush_count",
+			metric.WithDescription("Total number of memtable flushes"),
+			metric.WithUnit("{count}"),
+		)),
+		FlushLatency: must(meter.Float64Histogram(
+			"pebble_flush_latency",
+			metric.WithDescription("Latency of memtable flush operations"),
+			metric.WithUnit("s"),
```
yzang2019, line 107: Shall we use ms instead?

sei-db/ss/pebbledb/metrics.go
```diff
@@ -0,0 +1,217 @@
...
+			"pebble_iterator_creation_count",
+			metric.WithDescription("Total number of forward iterators created"),
+			metric.WithUnit("{count}"),
+		)),
+		ReverseIteratorCreationCount: must(meter.Int64Counter(
+			"pebble_reverse_iterator_creation_count",
+			metric.WithDescription("Total number of reverse iterators created"),
+			metric.WithUnit("{count}"),
+		)),
+		IteratorNextCallCount: must(meter.Int64Counter(
```
yzang2019: We don't need this any more right?

sei-db/ss/pebbledb/metrics.go
```diff
@@ -0,0 +1,217 @@
...
+		CacheMisses metric.Int64Counter
+		CacheSize   metric.Int64Gauge
+
+		// Operational Metrics
+		BatchSize                    metric.Int64Histogram
+		PendingChangesQueueDepth     metric.Int64Gauge
+		IteratorCount                metric.Int64UpDownCounter
+		IteratorCreationCount        metric.Int64Counter
+		ReverseIteratorCreationCount metric.Int64Counter
+		IteratorNextCallCount        metric.Int64Counter
```
yzang2019: Shall we remove this metrics and the one below it?

sei-db/ss/pebbledb/metrics.go
```diff
@@ -0,0 +1,217 @@
...
+		)),
+		ApplyChangesetLatency: must(meter.Float64Histogram(
+			"pebble_apply_changeset_latency",
+			metric.WithDescription("Time taken to apply changeset to PebbleDB"),
+			metric.WithUnit("s"),
+		)),
+		ApplyChangesetAsyncLatency: must(meter.Float64Histogram(
+			"pebble_apply_changeset_async_latency",
+			metric.WithDescription("Time taken to queue changeset for async write"),
+			metric.WithUnit("s"),
```
yzang2019, line 55: Can we change all latency to ms?

yzang2019 (approved):

sei-db/ss/pebbledb/metrics.go
Kbhat1: As @masih pointed out https://github.com/sei-protocol/sei-chain/pull/2463#discussion_r2422490511 better to use seconds and measure in float64

sei-db/ss/pebbledb/metrics.go
Kbhat1: As @masih pointed out https://github.com/sei-protocol/sei-chain/pull/2463#discussion_r2422490511 better to use seconds and measure in float64

sei-db/ss/pebbledb/metrics.go
Kbhat1: We still need iteration metrics @yzang2019 specifically counters (not latency / duration). This measures how many times next gets called, the others measure how many times forward / reverse get created.

Is the concern around it being too noisy?

sei-db/sc/memiavl/metrics.go
```diff
@@ -1,18 +1,19 @@
...
 	"go.opentelemetry.io/otel/metric"
 )
 
 var (
-	meter = otel.Meter("seidb")
+	meter = otel.Meter("seidb.memiavl")
 
-	SeiDBMetrics = struct {
+	// Metrics contains all MemIAVL-specific metrics
+	Metrics = struct {
```
masih: This doesn't need to be exported, right? If it does then I think we are missing something in the package structure.

sei-db/sc/memiavl/metrics.go
```diff
@@ -1,18 +1,19 @@
...
 		RestartLatency          metric.Float64Histogram
 		SnapshotCreationLatency metric.Float64Histogram
-		CommitLatency           metric.Int64Histogram
-		ApplyChangesetLatency   metric.Int64Histogram
+		CommitLatency           metric.Float64Histogram
+		ApplyChangesetLatency   metric.Float64Histogram
```
masih, line 15: Beyond the scope this PR, but all the fields of this struct also shouldn't be exported? Un-exporting the struct would limit the future scope creep and we can revisit this in a separate refactor PR if that's easier.

sei-db/sc/memiavl/metrics.go
```diff
@@ -1,18 +1,19 @@
...
+package memiavl
 
 import (
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/metric"
 )
 
 var (
-	meter = otel.Meter("seidb")
+	meter = otel.Meter("seidb.memiavl")
```
masih: I think OTEL converts dot to underscore automatically? worth double checking or better yet use `_` as delimiter.

sei-db/sc/memiavl/metrics.go
```diff
@@ -27,15 +28,15 @@ var (
 			metric.WithDescription("Time taken to create memiavl snapshot"),
 			metric.WithUnit("s"),
 		)),
-		CommitLatency: must(meter.Int64Histogram(
+		CommitLatency: must(meter.Float64Histogram(
 			"commit_latency",
```
masih: Depending on how these metrics are hooked up to the reporter we might want to add a prefix for all metrics manually to make sure nothing can go wrong.

sei-db/sc/memiavl/db.go
```diff
@@ -303,19 +302,27 @@ func (db *DB) ApplyChangeSets(changeSets []*proto.NamedChangeSet) error {
 	db.mtx.Lock()
 	defer db.mtx.Unlock()
 	startTime := time.Now()
+	var retErr error
```
masih: Couple of things here:
1. I recommend recording metrics regardless: Move metric collection right at the top before mutex lock etc. The metric type is a histogram. As such, it would tell us things like rate at which this function is called and more. It would also removes one cognitive cycle for all: No matter what we always measure latency in a function; no conditional measurement.
2. The `retErr` implementation here is correct, in that i will infer the `success` attribute correctly but here is the thing: we can write it such that it is far less error prone to accidentally getting it wrong or overriding the error value etc.:

    * Use err as normal in the entire function as if that `defer` function that measure stuff doesn't exist.
    * Make the `error` returned by this function into a named variable, conventionally starting with `_`, i.e.: `
func (db *DB) ApplyChangeSets(changeSets []*proto.NamedChangeSet) (_err error) {`. The `_` convention is there to point out to dev that _Never Set This Variable_.
    * Use `_err` in defer. Go compiler will always set the value of _err to whatever is returned as error (if any).
    * Revert all the changes to return statements and simply return error as the code did before.
    
    This way we never have to worry about accidentally forgetting to set the right err variable for defer to work correctly. Only one rule: whenever you se a var name starting with `_` never set it. 
    
    **(Ditto for the rest of the PR, specially pebble/db.go)**

sei-db/ss/pebbledb/batch.go
```diff
@@ -122,7 +139,19 @@ func (b *Batch) HardDelete(storeKey string, key []byte) error {
...
+		Metrics.BatchWriteLatency.Record(
+			ctx,
+			time.Since(startTime).Seconds(),
+			metric.WithAttributes(attribute.Bool("success", err == nil)),
+		)
+		Metrics.BatchSize.Record(
+			ctx,
+			batchSize,
+		)
 		err = errors.Join(err, b.batch.Close())
```
masih: This needs to move to the first line of the defer function.

sei-db/ss/pebbledb/db.go
```diff
@@ -77,6 +80,22 @@ type Database struct {
 	lastRangeHashedCache int64
 	lastRangeHashedMu    sync.RWMutex
 	hashComputationMu    sync.Mutex
+
+	// Metrics collection
+	metricsCtx        context.Context
```
masih: It is generally considered bad practice to have context as field of a struct; it's a form of footgun. In this specific case, the thing is we don't need to: Simply pass the context to `collectMetricsInBackground` in `New`. The `collectMetricsInBackground` function is the only place the context is used.

sei-db/ss/pebbledb/db.go
```diff
@@ -77,6 +80,22 @@ type Database struct {
...
+	metricsCancel     context.CancelFunc
+	lastMetricsValues metricsSnapshot
+	metricsMu         sync.RWMutex
```
masih: This is a read/write mutest but is only used for write. For write, it is only used to control access to `lastMetricsValues` which is a field we do not need to store as a struct field? see comment on `collectAndRecordMetrics`.

sei-db/ss/pebbledb/db.go
```diff
@@ -1085,3 +1204,130 @@ func retrieveLastRangeHashed(db *pebble.DB) (int64, error) {
...
+		Metrics.CompactionLatency.Record(
+			ctx,
+			avgLatencySeconds,
+			metric.WithAttributes(attribute.Bool("success", ctx.Err() == nil)),
+		)
+	}
+
+	// Flush metrics
+	flushCount := m.Flush.Count
+	flushBytesWritten := int64(m.Flush.WriteThroughput.Bytes)
```
masih: Casting to `int64` is unnecessary. Ditto for reminder of such castings in this function.

sei-db/ss/pebbledb/db.go
```diff
@@ -1085,3 +1204,130 @@ func retrieveLastRangeHashed(db *pebble.DB) (int64, error) {
...
+// collectAndRecordMetrics collects PebbleDB internal metrics and records them
+func (db *Database) collectAndRecordMetrics(ctx context.Context) {
+	if db.storage == nil {
+		return
+	}
+
+	m := db.storage.Metrics()
+
+	db.metricsMu.Lock()
+	lastSnapshot := db.lastMetricsValues
```
masih: Why bother with the `lastMetricsValues` struct field at all to then protect it with a lock? That field as far as I can see is only used in this function, and in this function we `Record` the metrics.

I think we can remove the mutex and `lastMetricsValues` struct field entirely.

sei-db/ss/pebbledb/db.go
```diff
@@ -1085,3 +1204,130 @@ func retrieveLastRangeHashed(db *pebble.DB) (int64, error) {
...
+	var compactionBytesRead int64
+	var compactionBytesWritten int64
+	for level := 0; level < len(m.Levels); level++ {
+		levelMetrics := m.Levels[level]
+		compactionBytesRead += int64(levelMetrics.BytesIn)
+		compactionBytesWritten += int64(levelMetrics.BytesCompacted)
+	}
+
+	// Calculate deltas for counters
+	compactionCountDelta := compactionCount - lastSnapshot.compactionCount
```
masih: Why `Delta`? we can do all sorts of fancy aggregation in prometheus/grafana. if we simply return the count then we: 1) retain the ability to do higher quality aggregation in the metric reporting system outside of the application, and 2) simplify the implementation here by rendering `lastMetricsValues` unnecessary.

Ditto for the remaining such calculations.

sei-db/ss/pebbledb/db.go
```diff
@@ -1085,3 +1204,130 @@ func retrieveLastRangeHashed(db *pebble.DB) (int64, error) {
...
+	compactionBytesReadDelta := compactionBytesRead - lastSnapshot.compactionBytesRead
+	compactionBytesWrittenDelta := compactionBytesWritten - lastSnapshot.compactionBytesWritten
+
+	// Record deltas as counters
+	Metrics.CompactionCount.Add(ctx, compactionCountDelta)
+	Metrics.CompactionBytesRead.Add(ctx, compactionBytesReadDelta)
+	Metrics.CompactionBytesWritten.Add(ctx, compactionBytesWrittenDelta)
+
+	// Record compaction latency if available
+	if compactionCountDelta > 0 {
```
masih: Remove this if statement entirely: Convetionally always report the value even if it is zero, because otherwise on prometheus/grafana it would show up as "missing data" which is confusing/usually indicates problem with reporting itself.

sei-db/ss/pebbledb/db.go
```diff
@@ -1085,3 +1204,130 @@ func retrieveLastRangeHashed(db *pebble.DB) (int64, error) {
...
+		avgLatencySeconds := m.Compact.Duration.Seconds() / float64(compactionCountDelta)
```
masih: Again, I would avoid such data aggregation in the applciation layer. Simply report numbers. We can then do more post-analysis in prometheus and have access to raw data for analytics that we don't know we want yet. Raw data is quite valuable.

sei-db/ss/pebbledb/db.go
```diff
@@ -1085,3 +1204,130 @@ func retrieveLastRangeHashed(db *pebble.DB) (int64, error) {
...
+	flushBytesWritten := int64(m.Flush.WriteThroughput.Bytes)
+
+	flushCountDelta := flushCount - lastSnapshot.flushCount
+	flushBytesWrittenDelta := flushBytesWritten - lastSnapshot.flushBytesWritten
+
+	Metrics.FlushCount.Add(ctx, flushCountDelta)
+	Metrics.FlushBytesWritten.Add(ctx, flushBytesWrittenDelta)
+
+	// Record flush latency if available
+	if flushCountDelta > 0 {
```
masih: Ditto re removing such conditions.

sei-db/ss/pebbledb/db.go
```diff
@@ -1085,3 +1204,130 @@ func retrieveLastRangeHashed(db *pebble.DB) (int64, error) {
...
+		)
+	}
+
+	// Storage metrics (gauges - absolute values)
+	var sstableCount int64
+	var sstableTotalSize int64
+	for level := 0; level < len(m.Levels); level++ {
+		levelMetrics := m.Levels[level]
+		sstableCount += levelMetrics.NumFiles
+		sstableTotalSize += int64(levelMetrics.Size)
```
masih: Instead of summing it report them as individual numbers, and use the level number as an attribute on the metric: it would simplify the logic here, gives us per level stats (valuable because level 0 underperforming would hurt a lot more than level 8 for example and we want to know which level), and finally it is trivial to `sum` the data if we want sum in prometheus.

sei-db/ss/pebbledb/iterator.go
```diff
@@ -94,6 +95,17 @@ func newPebbleDBIterator(src *pebble.Iterator, prefix, mvccStart, mvccEnd []byte
...
+	// Record iterator creation (separate metrics for forward and reverse)
+	ctx := context.Background()
+	if reverse {
+		Metrics.ReverseIteratorCreationCount.Add(ctx, 1)
+	} else {
+		Metrics.IteratorCreationCount.Add(ctx, 1)
+	}
+
+	// Increment active iterator count (gauge)
+	Metrics.IteratorCount.Add(ctx, 1)
```
masih: Combine these metrics and use `reverse` attribute to indicate if it was a reverse or not.

sei-db/ss/pebbledb/iterator.go
```diff
@@ -289,6 +301,9 @@ func (itr *iterator) nextReverse() {
 }
 
 func (itr *iterator) Next() {
+	// Record iterator Next call (for both forward and reverse)
+	Metrics.IteratorNextCallCount.Add(context.Background(), 1)
```
masih: I honestly think this level of metrics is unnecessary? we can get an idea from time it takes to return the values alone.

If we really really want this then this then what I recommend instead is to:
* keep a counter in the iterator struct and only increment the counter.
* record the count on iterator close, _not per `Next()` call_.

That should reduce the footprint of this significantly while retaining high-resolution data on counts.

sei-db/ss/pebbledb/iterator.go
```diff
@@ -328,6 +343,10 @@ func (itr *iterator) Close() error {
 	_ = itr.source.Close()
 	itr.source = nil
 	itr.valid = false
+
+	// Decrement active iterator count (gauge)
+	Metrics.IteratorCount.Add(context.Background(), -1)
```
masih: (If we really really want this metric then) We should use a single histogram instead of counter? The histogram would give us:
* the number of iterators instantiated
* The distribution of the number of times each one iterated - the value we really care about. Because the current implementation tells us the absolute number of iterations.

We can then get rid of all other iterator metrics.

This way we can use a struct level counter that has a smaller foot print in terms of measuring the iterator performance and a clear single metric that tells us more of what we want to know

sei-db/ss/pebbledb/metrics.go
```diff
@@ -0,0 +1,217 @@
...
+import (
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/metric"
+)
+
+var (
+	meter = otel.Meter("seidb.pebble")
+
+	// Metrics contains all PebbleDB-specific metrics
+	Metrics = struct {
```
masih: Un-export this?

sei-db/ss/pebbledb/metrics.go
```diff
@@ -0,0 +1,217 @@
...
+		// Operation Latencies
+		GetLatency                 metric.Float64Histogram
```
masih: Un-export the fields too?

Kbhat1:
@masih take a look again when you can

sei-db/ss/pebbledb/db.go
```diff
@@ -338,7 +345,18 @@ func (db *Database) Has(storeKey string, version int64, key []byte) (bool, error
...
-func (db *Database) Get(storeKey string, targetVersion int64, key []byte) ([]byte, error) {
+func (db *Database) Get(storeKey string, targetVersion int64, key []byte) (_ []byte, _err error) {
+	startTime := time.Now()
+	defer func() {
+		otelMetrics.getLatency.Record(
+			context.Background(),
+			time.Since(startTime).Seconds(),
+			metric.WithAttributes(
+				attribute.Bool("success", _err == nil),
+				attribute.String("store", storeKey),
```
Kbhat1, line 365: Fyi @yzang2019 added attribute for storeKey for both Get and iteration

sei-db/sc/memiavl/db.go
```diff
@@ -295,17 +294,23 @@ func (db *DB) ApplyUpgrades(upgrades []*proto.TreeNameUpgrade) error {
...
 	if len(changeSets) == 0 {
 		return nil
 	}
 
-	db.mtx.Lock()
-	defer db.mtx.Unlock()
 	startTime := time.Now()
 	defer func() {
-		metrics.SeiDBMetrics.ApplyChangesetLatency.Record(context.Background(), time.Since(startTime).Milliseconds())
+		otelMetrics.ApplyChangesetLatency.Record(
```
masih, line 304: We probably want this before the `if len(changeSets) == 0 {` _if_ we care about knowing the number of calls to apply changeset? the `zeroLen` could be an attribute to differentiate between zero length and non zero.

If you think we don't need to know the number of calls then please feel free to ignore this comment

sei-db/ss/pebbledb/db.go
```diff
@@ -170,6 +173,10 @@ func New(dataDir string, config config.StateStoreConfig) (*Database, error) {
 	database.streamHandler = streamHandler
 	go database.writeAsyncInBackground()
 
+	// Start background metrics collection
+	metricsCtx, metricsCancel := context.WithCancel(context.Background())
+	go database.collectMetricsInBackground(metricsCtx, metricsCancel)
```
masih: Sorry I might have not been clear enough in my previous comment: we still want the cancel function as struct field of `Database` so that we can call it in `Close`. Otherwise there is nothing that stops the background collection on db shutdown right?

sei-db/ss/pebbledb/db.go
```diff
@@ -1093,3 +1166,60 @@ func retrieveLastRangeHashed(db *pebble.DB) (int64, error) {
 	}
 	return int64(ubz), nil
 }
+
+// collectMetricsInBackground periodically collects PebbleDB internal metrics
+func (db *Database) collectMetricsInBackground(ctx context.Context, cancel context.CancelFunc) {
+	defer cancel()
```
masih: This won't work cause nothing cancels the context for the for loop to break out, right?

sei-db/ss/pebbledb/db.go
masih, line 365: Double checking: What does the cardinality of `storeKey` look like? How many unique store keys for the lifetime of the program?

sei-db/ss/pebbledb/db.go
```diff
@@ -570,7 +620,18 @@ func (db *Database) writeAsyncInBackground() {
...
 // This is not a large issue given the next time that module is updated, it will be properly pruned thereafter.
-func (db *Database) Prune(version int64) error {
+func (db *Database) Prune(version int64) (_err error) {
+	startTime := time.Now()
+	defer func() {
+		otelMetrics.pruneLatency.Record(
+			context.Background(),
+			time.Since(startTime).Seconds(),
+			metric.WithAttributes(
+				attribute.Int64("version", version),
```
masih: Same question about cardinality: how many unique versions are there in the lifetime of the program?

sei-db/ss/pebbledb/db.go
```diff
@@ -489,7 +527,19 @@ func (db *Database) computeMissingRanges(latestVersion int64) error {
...
-func (db *Database) computeHashForRange(beginBlock, endBlock int64) error {
+func (db *Database) computeHashForRange(beginBlock, endBlock int64) (_err error) {
+	startTime := time.Now()
+	defer func() {
+		otelMetrics.hashComputationLatency.Record(
+			context.Background(),
+			time.Since(startTime).Seconds(),
+			metric.WithAttributes(
+				attribute.Int64("begin_block", beginBlock),
+				attribute.Int64("end_block", endBlock),
```
masih: Ditto re cardinality for the above attributes.

sei-db/ss/pebbledb/db.go
```diff
@@ -736,12 +797,24 @@ func (db *Database) ReverseIterator(storeKey string, version int64, start, end [
...
 // TODO: Potentially add retries instead of panics
-func (db *Database) Import(version int64, ch <-chan types.SnapshotNode) error {
+func (db *Database) Import(version int64, ch <-chan types.SnapshotNode) (_err error) {
+	startTime := time.Now()
+	defer func() {
+		otelMetrics.importLatency.Record(
+			context.Background(),
+			time.Since(startTime).Seconds(),
+			metric.WithAttributes(
+				attribute.Int64("version", version),
```
masih: cardinality check.

masih (commented):
One blocker on background metric collection context cancel function, and left some questions about cardinality just to double check they are not explosively large as they increase the number of metrics exponentially.

Thank you so much for addressing all the comments @Kbhat1, cannot wait to dig through the data being collected by the metrics 🚀

sei-db/ss/pebbledb/db.go
Kbhat1: Its just the number of modules which is fixed (21) and will not grow

sei-db/sc/memiavl/db.go
Kbhat1: Don't think we need number of calls

yzang2019 (approved):

PR #2465, opened 2025-10-13
Integrate watermark into evmrpc

jewei1997:

A common workflow that users do is to query the latest height and then query information about that height such as events/logs. However, since we have an option to do async writes for receipts, the events/logs may not be available yet for that height, causing the query to fail.

In general, we want all of our heights outputted by our RPC endpoints to be aligned with each other. We have introduced the concept of a watermark (https://github.com/sei-protocol/sei-db/pull/116) across Tendermint block store, SS Store, and receipt store. Pulling this watermark across all 3 dbs should allow the evmrpc to output the minimum height available across these dbs for a latest height. This will ensure that all data for a given height outputted by our RPCs is available across all our endpoints.

This PR introduces a WatermarkManager that manages the heights of these DBs, which the evmrpc endpoints can use to get the latest heights that are ready.

unit tests + manual testing

### DIFF
--- a/sei-cosmos/storev2/rootmulti/store.go
+++ b/sei-cosmos/storev2/rootmulti/store.go
@@ -781,7 +781,8 @@ loop:
 	}
 	// initialize the earliest version for SS store
 	if rs.ssStore != nil {
-		rs.ssStore.SetEarliestVersion(height, false)
+		err := rs.ssStore.SetEarliestVersion(height, false)
+		rs.logger.Error("Failed to set earliest version during DB restore", err)
 	}
 
 	return snapshotItem, restoreErr
--- a/sei-db/sc/memiavl/db.go
+++ b/sei-db/sc/memiavl/db.go
@@ -221,10 +221,6 @@ func OpenDB(logger logger.Logger, targetVersion int64, opts Options) (database *
 		}
 	}
 
-	// We need to prune snapshots during start up to avoid snapshot leaks
-	if !db.readOnly {
-		db.pruneSnapshots()
-	}
 	return db, nil
 }
 
@@ -425,51 +421,49 @@ func (db *DB) checkBackgroundSnapshotRewrite() error {
 func (db *DB) pruneSnapshots() {
 	// wait until last prune finish
 	db.pruneSnapshotLock.Lock()
+	defer db.pruneSnapshotLock.Unlock()
 
-	go func() {
-		defer db.pruneSnapshotLock.Unlock()
+	currentVersion, err := currentVersion(db.dir)
+	if err != nil {
+		db.logger.Error("failed to read current snapshot version", "err", err)
+		return
+	}
 
-		currentVersion, err := currentVersion(db.dir)
-		if err != nil {
-			db.logger.Error("failed to read current snapshot version", "err", err)
-			return
-		}
-
-		counter := db.snapshotKeepRecent
-		if err := traverseSnapshots(db.dir, false, func(version int64) (bool, error) {
-			if version >= currentVersion {
-				// ignore any newer snapshot directories, there could be ongoning snapshot rewrite.
-				return false, nil
-			}
-
-			if counter > 0 {
-				counter--
-				return false, nil
-			}
-
-			name := snapshotName(version)
-			db.logger.Info("prune snapshot", "name", name)
-
-			if err := atomicRemoveDir(filepath.Join(db.dir, name)); err != nil {
-				db.logger.Error("failed to prune snapshot", "err", err)
-			}
-
+	counter := db.snapshotKeepRecent
+	if err := traverseSnapshots(db.dir, false, func(version int64) (bool, error) {
+		if version >= currentVersion {
+			// ignore any newer snapshot directories, there could be ongoning snapshot rewrite.
 			return false, nil
-		}); err != nil {
-			db.logger.Error("fail to prune snapshots", "err", err)
-			return
 		}
 
-		// truncate Rlog until the earliest remaining snapshot
-		earliestVersion, err := GetEarliestVersion(db.dir)
-		if err != nil {
-			db.logger.Error("failed to find first snapshot", "err", err)
+		if counter > 0 {
+			counter--
+			return false, nil
 		}
 
-		if err := db.streamHandler.TruncateBefore(utils.VersionToIndex(earliestVersion+1, db.initialVersion)); err != nil {
-			db.logger.Error("failed to truncate rlog", "err", err, "version", earliestVersion+1)
+		name := snapshotName(version)
+		db.logger.Info("prune snapshot", "name", name)
+
+		if err := atomicRemoveDir(filepath.Join(db.dir, name)); err != nil {
+			db.logger.Error("failed to prune snapshot", "err", err)
 		}
-	}()
+
+		return false, nil
+	}); err != nil {
+		db.logger.Error("fail to prune snapshots", "err", err)
+		return
+	}
+
+	// truncate Rlog until the earliest remaining snapshot
+	earliestVersion, err := GetEarliestVersion(db.dir)
+	if err != nil {
+		db.logger.Error("failed to find first snapshot", "err", err)
+	}
+
+	if err := db.streamHandler.TruncateBefore(utils.VersionToIndex(earliestVersion+1, db.initialVersion)); err != nil {
+		db.logger.Error("failed to truncate rlog", "err", err, "version", earliestVersion+1)
+	}
+
 }
 
 // Commit wraps SaveVersion to bump the version and writes the pending changes into log files to persist on disk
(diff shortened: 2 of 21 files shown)


CHANGELOG.md
```diff
@@ -27,64 +27,6 @@ Ref: https://keepachangelog.com/en/1.0.0/
 -->
 
 # Changelog
-## v6.2.0
```
yzang2019, line 30: Is this intentional?

evmrpc/block.go
```diff
@@ -118,10 +123,13 @@ func (a *BlockAPI) GetBlockTransactionCountByNumber(ctx context.Context, number
 	if err != nil {
 		return nil, err
 	}
-	block, err := blockByNumberWithRetry(ctx, a.tmClient, numberPtr, 1)
+	block, err := blockByNumberRespectingWatermarks(ctx, a.tmClient, a.watermarks, numberPtr, 1)
 	if err != nil {
 		return nil, err
 	}
+	if err := a.ensureHeightAvailable(ctx, block.Block.Height); err != nil {
```
yzang2019: Why calling ensureHeightAvailable again here? Shouldn't this be handled already in `blockByNumberRespectingWatermarks` ?

evmrpc/block.go
```diff
@@ -212,17 +226,27 @@ func (a *BlockAPI) getBlockByNumber(
 		}
 	}
 
-	block, err := blockByNumberWithRetry(ctx, a.tmClient, numberPtr, 1)
+	block, err := blockByNumberRespectingWatermarks(ctx, a.tmClient, a.watermarks, numberPtr, 1)
 	if err != nil {
 		return nil, err
 	}
+	if err := a.ensureHeightAvailable(ctx, block.Block.Height); err != nil {
```
yzang2019: Same here

evmrpc/config.go
```diff
@@ -75,6 +75,9 @@ type Config struct {
 	// controls whether to synchronously flush receipts before block finalze or not
 	FlushReceiptSync bool `mapstructure:"flush_receipt_sync"`
 
+	// DisableWatermark disables cross-store watermark gating (intended for tests)
```
yzang2019: Why tests need to disable watermark? Shouldn't we fix the test and make sure new watermark works for all tests?

evmrpc/filter.go
```diff
@@ -805,6 +834,47 @@ func (f *LogFetcher) mergeSortedLogs(batches [][]*ethtypes.Log) []*ethtypes.Log
 	return res
 }
 
+func (f *LogFetcher) ensureHeightAvailable(ctx context.Context, height int64) error {
+	if f.watermarks == nil {
+		return nil
+	}
+	return f.watermarks.EnsureHeightAvailable(ctx, height)
+}
```
yzang2019: This seem to be lot of duplicate code, can't we just f.watermarks.EnsureHeightAvailable to avoid adding this function everywhere? Also why do we need to check watermarks is nil or not? Can we always make sure it's not nil?

evmrpc/filter.go
```diff
@@ -805,6 +834,47 @@ func (f *LogFetcher) mergeSortedLogs(batches [][]*ethtypes.Log) []*ethtypes.Log
...
+	}
+	if f.ctxProvider == nil {
+		return 0, fmt.Errorf("ctx provider not configured")
+	}
+	return f.ctxProvider(LatestCtxHeight).BlockHeight(), nil
+}
+
+func (f *LogFetcher) earliestHeight(ctx context.Context) (int64, error) {
+	if f.watermarks != nil {
+		return f.watermarks.EarliestHeight(ctx)
```
yzang2019: Can we always assume and use watermark to get earliest version? Thus avoid all the below logic?

evmrpc/info.go
```diff
@@ -377,6 +393,40 @@ func (i *InfoAPI) CalculateGasUsedRatio(ctx context.Context, blockHeight int64)
...
+	func() {
+		defer func() {
+			if r := recover(); r != nil {
+				earliest = 0
+			}
+		}()
+		earliest = ms.GetEarliestVersion()
+	}()
+	return earliest, nil
+}
```
yzang2019, line 402: Avoid duplicate code

evmrpc/watermark_manager.go
```diff
@@ -0,0 +1,273 @@
...
+// Watermarks returns the earliest and latest block heights that are safe to
+// serve. Earliest is inclusive, latest is inclusive.
+func (m *WatermarkManager) Watermarks(ctx context.Context) (int64, int64, error) {
+	if m == nil {
+		return 0, 0, errNoHeightSource
+	}
+
+	var (
+		earliestCandidates []int64
+		latestCandidates   []int64
```
yzang2019: why do we need to keep a list of candidates height?

evmrpc/watermark_manager.go
```diff
@@ -0,0 +1,273 @@
...
+	} else if !errors.Is(err, errNoHeightSource) {
+		return 0, 0, err
+	}
+
+	// State store heights (historical state DB)
+	if ssLatest, ssEarliest, ok := m.fetchStateStoreWatermarks(); ok {
+		latestCandidates = append(latestCandidates, ssLatest)
+		earliestCandidates = append(earliestCandidates, ssEarliest)
+	} else if msLatest, msEarliest, ok := m.fetchMultiStoreWatermarks(); ok {
+		// Fall back to application multistore heights if state store unavailable
```
yzang2019: why the fallback here? I don't quite understand the motivation here, multistore is basically the same thing as SS

evmrpc/info.go
jewei1997: made a helper for these two

evmrpc/watermark_manager.go
```diff
@@ -0,0 +1,265 @@
...
+		receiptStore: receiptStore,
+	}
+}
+
+// Watermarks returns the earliest and latest block heights that are safe to
+// serve. Earliest is inclusive, latest is inclusive.
+func (m *WatermarkManager) Watermarks(ctx context.Context) (int64, int64, error) {
+	if m == nil {
+		return 0, 0, errNoHeightSource
+	}
```
yzang2019: Is this needed?

evmrpc/filter.go
```diff
@@ -676,6 +680,31 @@ func (f *LogFetcher) GetLogsByFilters(ctx context.Context, crit filters.FilterCr
 	if lastToHeight > begin {
 		begin = lastToHeight
 	}
+	if begin < earliest {
+		begin = earliest
```
stevenlanders: I'd propose that we should fail with an error like the block range is too far out of range rather than quietly succeeding with a different range.

evmrpc/info.go
```diff
@@ -48,7 +49,11 @@ type FeeHistoryResult struct {
 func (i *InfoAPI) BlockNumber() hexutil.Uint64 {
 	startTime := time.Now()
 	defer recordMetrics("eth_BlockNumber", i.connectionType, startTime)
-	return hexutil.Uint64(i.ctxProvider(LatestCtxHeight).BlockHeight()) //nolint:gosec
+	height, err := i.latestHeight(context.Background())
+	if err != nil {
+		height = i.ctxProvider(LatestCtxHeight).BlockHeight()
```
stevenlanders: when would this error happen?

evmrpc/info.go
```diff
@@ -149,24 +154,35 @@ func (i *InfoAPI) FeeHistory(ctx context.Context, blockCount gmath.HexOrDecimal6
...
 	genesisHeight := genesis.Genesis.InitialHeight
-	currentHeight := i.ctxProvider(LatestCtxHeight).BlockHeight()
+	latestHeight, err := i.latestHeight(ctx)
+	if err != nil {
+		return nil, err
+	}
+	earliestHeight, err := i.earliestHeight(ctx)
+	if err != nil {
+		// fall back to genesis height if earliest watermark unavailable
+		earliestHeight = genesisHeight
```
stevenlanders: is this for ensuring backwards compatibility with pre-watermark blocks?

evmrpc/filter.go
```diff
@@ -1010,6 +1047,11 @@ func (f *LogFetcher) processBatch(ctx context.Context, start, end int64, crit fi
 		// Re-check cache after acquiring semaphore, in case another worker cached it.
 		if cachedEntry, found := f.globalBlockCache.Get(height); found {
 			<-f.dbReadSemaphore
+			if cachedEntry.Block != nil {
+				if err := f.watermarks.EnsureHeightAvailable(ctx, cachedEntry.Block.Block.Height); err != nil {
+					continue
+				}
+			}
```
stevenlanders: do we need to use `cachedEntry.Block.Block.Height` or should we consider just `height` that we have?

stevenlanders: This feels like logic that should have addressed the asked-for range well before we got this far? (like maybe pre-line 1028?)

yzang2019 (approved):

PR #2467, opened 2025-10-14
Simplify tracer enabled checks throughout sei-chain/cosmos app

udpatil:
This simplifies calls to tracer to gracefully handle cases where the tracing isn't enabled by simply returning a NoopSpan discards anything passed to it.

Added unit tests

### DIFF
--- a/sei-cosmos/utils/tracing/tracer.go
+++ b/sei-cosmos/utils/tracing/tracer.go
@@ -93,18 +93,18 @@ func (i *Info) StartWithContext(name string, ctx context.Context) (context.Conte
 	return i.tracer.Start(ctx, name)
 }
 
-func (i *Info) StartBlockSpan(c context.Context) (context.Context, otrace.Span) {
+func (i *Info) StartBlockSpan() otrace.Span {
 	if !i.tracingEnabled.Load() {
-		return c, NoOpSpan
+		return NoOpSpan
 	}
 	i.mtx.Lock()
 	defer i.mtx.Unlock()
 	if i.blockSpan.IsRecording() { // already started
-		return c, i.blockSpan
+		return i.blockSpan
 	}
-	ctx, span := i.tracer.Start(c, "Block")
+	_, span := i.tracer.Start(context.Background(), "Block")
 	i.blockSpan = span
-	return ctx, i.blockSpan
+	return span
 
 }
 
--- a/app/abci.go
+++ b/app/abci.go
@@ -11,7 +11,7 @@ import (
 )
 
 func (app *App) BeginBlock(ctx sdk.Context, req abci.RequestBeginBlock) (res abci.ResponseBeginBlock) {
-	_, blockSpan := app.GetBaseApp().TracingInfo.StartBlockSpan(context.Background())
+	blockSpan := app.GetBaseApp().TracingInfo.StartBlockSpan()
 	blockSpan.SetAttributes(attribute.Int64("height", req.Header.Height))
 	spanCtx, beginBlockSpan := app.GetBaseApp().TracingInfo.Start("BeginBlock")
 	defer beginBlockSpan.End()
--- a/sei-cosmos/utils/tracing/tracer_test.go
+++ b/sei-cosmos/utils/tracing/tracer_test.go
@@ -220,13 +220,11 @@ func TestInfo_StartBlockSpan_DisabledTracing(t *testing.T) {
 	tracer := noop.NewTracerProvider().Tracer("test")
 	info := NewTracingInfo(tracer, false)
 
-	ctx := context.Background()
-	returnedCtx, span := info.StartBlockSpan(ctx)
+	span := info.StartBlockSpan()
 
-	// When tracing is disabled, should return the input context and nil span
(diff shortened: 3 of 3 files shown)


sei-cosmos/baseapp/baseapp.go
```diff
@@ -999,9 +991,7 @@ func (app *BaseApp) runTx(ctx sdk.Context, mode runTxMode, tx sdk.Tx, checksum [
 		expireHandler = ctx.ExpireTxHandler()
 		msCache.Write()
 		anteEvents = events.ToABCIEvents()
-		if app.TracingEnabled {
-			anteSpan.End()
-		}
+		anteSpan.End()
```
masih, line 991: I am curious why this is not called in defer statement. Seem like a bunch of cases could lead to end not being called?

sei-cosmos/utils/tracing/tracer.go
```diff
@@ -47,14 +48,31 @@ func GetTracerProviderOptions(url string) ([]trace.TracerProviderOption, error)
...
+	Tracer         *otrace.Tracer
+	tracerContext  context.Context
+	BlockSpan      *otrace.Span
+	tracingEnabled int32 // Use int32 for atomic operations
+	mtx            sync.RWMutex
+}
 
-	mtx sync.RWMutex
+func NewTracingInfo(tr *otrace.Tracer, tracingEnabled bool) *Info {
+	tracingEnabledInt32 := int32(0) // default to false
```
masih: You could write this as:

```suggestion
	var tracingEnabledInt32 int32 // default to false
```

This way you can then simplify the assignment in if statement to 1 without casting.

sei-cosmos/utils/tracing/tracer.go
```diff
@@ -47,14 +48,31 @@ func GetTracerProviderOptions(url string) ([]trace.TracerProviderOption, error)
...
+		Tracer:         tr,
+		tracingEnabled: tracingEnabledInt32,
+	}
 }
 
+// NoOpSpan is a no-op span which does nothing.
+var NoOpSpan = otrace.SpanFromContext(context.TODO())
+
 func (i *Info) Start(name string) (context.Context, otrace.Span) {
+	if atomic.LoadInt32(&i.tracingEnabled) == 0 {
```
masih: Questions:
- does the flag change at runtime for a span that's already instantiated? I think not? If so then we can get rid of atomics entirely. 


- if we need them then why not atomic bool?

pompon0: +1, golang now provides atomic types. Don't use atomic operations on regular types.

sei-cosmos/baseapp/baseapp.go
udpatil: https://github.com/sei-protocol/sei-chain/pull/2467/files/0e0ff14fe3be304bc4900dc18521e3bfb8387600#diff-3b7bf57aee91f22d32fc6f4d59f8db882546f61af6627f23bf76a1ca10eee2a3R927
> defer anteSpan.End()

We call it with defer above, but this is to ensure a tighter end time for the span (and then the defer will end the span in case of earlier error / return)

sei-cosmos/utils/tracing/tracer.go
udpatil: We don't necessarily NEED them, my concern is that I wanted to check this outside of the mutex so that when we have tracing disabled, it doesn't cause contention for the NoOp spans, we only write this once at instantiation, so initially I was going to leave it as just a bool and let all the reads race sicne there shouldnt be any race side effects due to read only, but WDYT?

sei-cosmos/utils/tracing/tracer.go
```diff
@@ -47,14 +48,29 @@ func GetTracerProviderOptions(url string) ([]trace.TracerProviderOption, error)
 }
 
 type Info struct {
-	Tracer        *otrace.Tracer
-	tracerContext context.Context
-	BlockSpan     *otrace.Span
+	Tracer         *otrace.Tracer
+	tracerContext  context.Context
+	BlockSpan      *otrace.Span
+	tracingEnabled *atomic.Bool
```
stevenlanders: i wonder if this needs to be atomic since the constructor sets it?  (might be overkill / contention?)

pompon0: also there is no need for pointer, just embedding bool/atomic.Bool should suffice

udpatil: the atomic Load shouldn't add that much contention @stevenlanders and is preferable to acquiring the mutex every time

sei-cosmos/baseapp/baseapp.go
masih: > to ensure a tighter end time

I would like to see some data on that 🤔

sei-cosmos/utils/tracing/tracer.go
masih: Thanks for explaining. I think the locking /atomicity logic here can be simplified further. The work is beyond the scope of this PR. Happy to discuss further async.

PR #2471, opened 2025-10-16
simplified p2p.Channel

pompon0:
Simplified the channel implementation. Now channels don't have background tasks. Sending messages to channel is non-blocking and never fails. Messages sent to channels go directly to relevant peer queues. Peer errors are reported directly to PeerManager. Additionally:
* wrapped the PeerManager.ready set into a Watch, so that changes to the connection set are directly observable
* added PreexistingPeers to PeerUpdates, so that an observer can reconstruct the connection set no matter when they subscribe to updates (PeerUpdates will be eventually removed in favor of observing the connection set directly).
* made a lot of test actually use the Router instead of faking the peers on the channel.
* made Router pass around net.TCPConn instead of net.Conn (implicitly assumed to be a TCPConn anyway)
* replaced LimitListener with a regular listener + semaphore.


sei-tendermint/internal/p2p/channel.go
```diff
@@ -12,16 +11,14 @@ import (
 )
 
 // Envelope contains a message with sender/receiver routing info.
```
udpatil: nit: update this docstring

sei-tendermint/internal/p2p/channel.go
```diff
@@ -59,153 +56,106 @@ func (pe PeerError) Unwrap() error { return pe.Err }
 
 // Channel is a bidirectional channel to exchange Protobuf messages with peers.
 // Each message is wrapped in an Envelope to specify its sender and receiver.
```
udpatil: Update comments

sei-tendermint/internal/p2p/router.go
```diff
@@ -159,36 +138,25 @@ type Router struct {
...
-	defer r.channelMtx.RUnlock()
-	descs := make([]*ChannelDescriptor, len(r.chDescs))
-	copy(descs, r.chDescs)
-	return descs
-}
-
-type chDescAdderWithCallback struct {
-	chDesc *ChannelDescriptor
-	cb     func(*Channel)
+	for channels := range r.channels.Lock() {
```
udpatil: shouldn't this be r.channels.Rlock() ?

sei-tendermint/internal/p2p/transport.go
```diff
@@ -80,38 +92,113 @@ func HandshakeOrClose(
...
+		if err != nil {
+			return err
+		}
+		for chs := range r.channels.RLock() {
+			ch, ok := chs[chID]
+			if !ok {
+				// TODO(gprusak): verify if this is a misbehavior, and drop the peer if it is.
+				r.logger.Debug("dropping message for unknown channel", "peer", c.peerInfo.NodeID, "channel", chID)
+				continue
+			}
```
udpatil, line 146: Why do we need to hold the RLock after this point if we've already read the channel itself? is this to enforce synchronization on the underlying channel as well? if so, wouldn't it make more sense to have RWMutex[*channel] as the map value and acquire that instead of locking the full map until we can send to the recvQueue?

sei-tendermint/internal/p2p/transport.go
```diff
@@ -80,38 +92,113 @@ func HandshakeOrClose(
...
-	if chID > math.MaxUint8 {
-		return fmt.Errorf("MConnection only supports 1-byte channel IDs (got %v)", chID)
+func (c *Connection) Run(ctx context.Context, r *Router) error {
+	peerInfo := c.PeerInfo()
+	peerID := peerInfo.NodeID
+	r.metrics.Peers.Add(1)
+	defer r.metrics.Peers.Add(-1)
+	for conns := range r.conns.Lock() {
+		if old, ok := conns[peerID]; ok {
+			old.Close()
```
udpatil, line 178: if we have an old connection here that we are closing the net connection for, how do we cleanup the scope.Run() with spawned send/recv routines and mconn? I'm not fully understanding the scenario in which we are creating a new connection and `Run`ning it while the router already has an old connection running? and if so, how do we ensure that the old one is being cleaned up properly?

sei-tendermint/internal/p2p/transport_test.go
```diff
@@ -99,22 +106,33 @@ func TestRouter_Listen(t *testing.T) {
...
+				return nil
+			})
 			if err != nil {
-				t.Fatalf("tcp.Dial(): %v", err)
-			}
-			defer tcpConn.Close()
-			key, info := makeKeyAndInfo()
-			if _, err := handshake(ctx, logger, tcpConn, info, key); err != nil {
-				t.Fatalf("handshake(): %v", err)
+				t.Fatal(err)
```
udpatil, line 135: This test is failing for ipv4 Unspecified on my machine with the following:
```
2025-10-23T13:01:54-05:00 DEBUG dialing peer address node=65a1845968d681f816fb7d6eb040032e076ac7e8 peer={"Hostname":"0.0.0.0","NodeID":"bc10c5d04f5a8481e3ed430f25254b6c79899159","Port":61158}
2025-10-23T13:02:02-05:00 DEBUG failed to dial endpoint endpoint=[::ffff:0.0.0.0]:61158 err="dial tcp 0.0.0.0:61158: connect: operation timed out" node=65a1845968d681f816fb7d6eb040032e076ac7e8 peer=bc10c5d04f5a8481e3ed430f25254b6c79899159

--- FAIL: TestRouter_Listen (7.85s)
    --- FAIL: TestRouter_Listen/0.0.0.0 (7.84s)
        /Users/uday/sei/sei-chain/sei-tendermint/internal/p2p/transport_test.go:136: tcp.Dial(): all endpoints failed
```
However, the other test cases in this do pass for me, so I'm not sure if there is an issue on my end?

sei-tendermint/internal/p2p/transport.go
pompon0: The thing with channels is that the map is not expected to be mutated after router initialization. Therefore the (readonly)locking is only done for correctness (alternatively we could make the channel set immutable after router construction, but the codebase is not ready for that). As long as there is no blocking calls within the RLock section we are perfectly fine.

sei-tendermint/internal/p2p/transport.go
pompon0: So the router is perfectly fine with running 2 connections with the same peer at once. In this PR, the precise behavior is that we initiate termination when starting a new connection. As soon as Close() is called the task executing Connection.Run() on the closed connection will realize that connection is dead (namely, reading/writing from the connection will fail which will in turn terminate whole task. Of course, we can make the connection termination more strict:
* make Close() directly cancel the Connection.Run()'s context (instead of closing the underlying TCPConn).
* We can make it block until the connection is actually terminated.
That would also be a valid design decision.

sei-tendermint/internal/p2p/transport.go
pompon0: btw, nice catch with the `r.channels.Lock()` in getChannelDescs() above. That would be problematic indeed.

sei-tendermint/internal/p2p/transport_test.go
pompon0: good point, I don't know why it even works for me. Will fix.

udpatil (approved):

PR #2476, opened 2025-10-17
DNS resolution test for ResolveAddressString

pompon0:
Also fixed TestResolveAddressString test which was failing.
Also fixed sei-tendermint unit test workflow which was skipping tests - NUM_SPLITS variable was not set correctly (was ignored) and "make split-test-packages" had a bug - instead of splitting into N files, it was splitting into files with N lines.

### DIFF
--- a/sei-tendermint/config/config.go
+++ b/sei-tendermint/config/config.go
@@ -999,6 +999,10 @@ type StateSyncConfig struct {
 	// If this is true, then state sync will look for existing snapshots
 	// which are located in the snapshot-dir configured in app.toml (default to [home-dir]/data/snapshots)
 	UseLocalSnapshot bool `mapstructure:"use-local-snapshot"`
+
+	// LightBlockResponseTimeout is how long the dispatcher waits for a peer to
+	// return a light block.
+	LightBlockResponseTimeout time.Duration `mapstructure:"light-block-response-timeout"`
 }
 
 func (cfg *StateSyncConfig) TrustHashBytes() []byte {
@@ -1013,14 +1017,15 @@ func (cfg *StateSyncConfig) TrustHashBytes() []byte {
 // DefaultStateSyncConfig returns a default configuration for the state sync service
 func DefaultStateSyncConfig() *StateSyncConfig {
 	return &StateSyncConfig{
-		TrustPeriod:             168 * time.Hour,
-		DiscoveryTime:           15 * time.Second,
-		ChunkRequestTimeout:     15 * time.Second,
-		Fetchers:                2,
-		BackfillBlocks:          0,
-		BackfillDuration:        0 * time.Second,
-		VerifyLightBlockTimeout: 60 * time.Second,
-		BlacklistTTL:            5 * time.Minute,
+		TrustPeriod:               168 * time.Hour,
+		DiscoveryTime:             15 * time.Second,
+		ChunkRequestTimeout:       15 * time.Second,
+		Fetchers:                  2,
+		BackfillBlocks:            0,
+		BackfillDuration:          0 * time.Second,
+		VerifyLightBlockTimeout:   60 * time.Second,
+		BlacklistTTL:              5 * time.Minute,
+		LightBlockResponseTimeout: 10 * time.Second,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


sei-tendermint/types/node_info_test.go
```diff
@@ -178,76 +180,74 @@ func TestNodeInfoAddChannel(t *testing.T) {
...
 func TestResolveAddressString(t *testing.T) {
 	testCases := []struct {
-		name     string
-		addr     string
-		expected string
-		correct  bool
+		name    string
+		addr    string
+		want    netip.AddrPort
+		correct bool
```
masih: A small convention I follow when writing go tests "want" and "got" (as you have used) but the name qualified for both result and error:

```suggestion
		wantAddrPort    netip.AddrPort
		wantErr           bool
```

Better yet, change `wantErr` to string and infer empty as "want no err", and non-empty as `require.ErrorContains` for specific type of error.

sei-tendermint/types/node_info_test.go
```diff
@@ -178,76 +180,74 @@ func TestNodeInfoAddChannel(t *testing.T) {
 	require.Contains(t, nodeInfo.Channels, byte(0x02))
 }
 
+func TestResolveAddressStringDNS(t *testing.T) {
+	addr, err := ResolveAddressString("deadbeefdeadbeefdeadbeefdeadbeefdeadbeef@localhost:1234")
+	require.NoError(t, err)
+	require.True(t, addr.Addr().IsLoopback())
+	require.Equal(t, 1234, addr.Port())
+}
```
masih, line 189: I am curious, why this test stands separate from the rest? Could we add this case to the table-test below?

sei-tendermint/Makefile
```diff
@@ -305,6 +305,6 @@ $(BUILDDIR)/packages.txt:$(GO_TEST_FILES) $(BUILDDIR)
 	go list -f "{{ if (or .TestGoFiles .XTestGoFiles) }}{{ .ImportPath }}{{ end }}" ./... | sort > $@
 
 split-test-packages:$(BUILDDIR)/packages.txt
-	split -d -l $(NUM_SPLIT) $< $<.
+	split -d -n l/$(NUM_SPLIT) $< $<.
```
masih, line 308: Thanks for fixing this; i will be ripping all of this out in favour of vanilla go test parallel and better speced github actions runners hopefully soon.

sei-tendermint/types/node_info_test.go
```diff
@@ -178,76 +180,74 @@ func TestNodeInfoAddChannel(t *testing.T) {
...
+
 func TestResolveAddressString(t *testing.T) {
```
masih, line 191: Thinking out loud: Would be nice to expand testing further with fuzz tests at some point.

masih (approved):

sei-tendermint/types/node_info_test.go
pompon0: there is more that 1 possible result - IPv4 or IPv6 loopback

sei-tendermint/types/node_info_test.go
pompon0: I strongly prefer strongly typed error to "stringly" typed error.

masih: Sure but this is tests. If there is a type, by all means use `error` type as `wantErr` and `require.ErrorAs`. If there is no types then having that message test is a stronger test than just checking if error is nil. 

In most cases that I have come across the error typing is not finegrained enough to _always_ assert types. Therefore, I find it acceptable to use string message as a way to assert rationale in tests.

pompon0: I've compressed both `want` fields to `utils.Option[netip.AddrPort]`

pompon0: > Sure but this is tests. If there is a type, by all means use error type as wantErr and require.ErrorAs. If there is no types then having that message test is a stronger test than just checking if error is nil.

If there is no type, you can always add a type. Assuming that error messages are stable is fragile.

pompon0: > stronger test than just checking if error is nil.

It is stronger, but not necessarily a stable property of the interface. There are concrete properties of implementations that you don't want to assert in tests. Otherwise your tests will devolve into change detectors.

masih: On utils, my recommendation going forward would be to avoid adding to the custom testing framework that seems to be slowly growing, in favour of simply typing things out in vanilla go or popular testing frameworks maintained by someone other than us 🙂

PR #2477, opened 2025-10-17
bump go-ethereum

codchen:
bump go-ethereum for https://github.com/sei-protocol/go-ethereum/commit/e68db05096765f16abac30504e807deb2cc07bd4

unit tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -352,7 +352,7 @@ replace (
 	github.com/cosmos/cosmos-sdk => ./sei-cosmos
 	github.com/cosmos/iavl => github.com/sei-protocol/sei-iavl v0.2.0
 	github.com/cosmos/ibc-go/v3 => github.com/sei-protocol/sei-ibc-go/v3 v3.3.6
-	github.com/ethereum/go-ethereum => github.com/sei-protocol/go-ethereum v1.15.7-sei-9
+	github.com/ethereum/go-ethereum => github.com/sei-protocol/go-ethereum v1.15.7-sei-10
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -352,7 +352,7 @@ replace (
 	github.com/cosmos/cosmos-sdk => ./sei-cosmos
 	github.com/cosmos/iavl => github.com/sei-protocol/sei-iavl v0.2.0
 	github.com/cosmos/ibc-go/v3 => github.com/sei-protocol/sei-ibc-go/v3 v3.3.6
-	github.com/ethereum/go-ethereum => github.com/sei-protocol/go-ethereum v1.15.7-sei-7.0.20250929182230-93350978bb7c
+	github.com/ethereum/go-ethereum => github.com/sei-protocol/go-ethereum v1.15.7-sei-9
```
masih: Just to pick your brain @codchen, would it be useful to move go-ethereum into this repo too?

stevenlanders: It might be harder to take upstream changes - but I defer to @codchen (it'd be nice in a lot of ways)

codchen: @masih the only challenge I see is if we need to upgrade geth to the next major version (fusaka or something else), it might be easier to have a separate repo that's rebase-able. If there is some go/git magic that can still make this possible with a monorepo then I don't think there is any other concern.

masih: Thank you both. I think it's best to keep go-ethereum repo separate at least for now. I will integrate the UCI versioning into it though. That should make the version bumps reviewable/easier and provide a better structure for bubble up.

I think we can automate the bubble up via Dependabot entirely for that repo at least. First things first: semver so that the bot can infer what's newer.

philipsu522 (approved):

PR #2479, opened 2025-10-17
config: set pruning=nothing for all nodes

blindchaser:

- Set basic.pruning IAVL configs to keep everything if someone wants to use IAVL

### DIFF
--- a/sei-db/ss/pebbledb/db.go
+++ b/sei-db/ss/pebbledb/db.go
@@ -587,8 +587,8 @@ func (db *Database) Prune(version int64) error {
 	for itr.First(); itr.Valid(); {
 		currKeyEncoded := slices.Clone(itr.Key())
 
-		// Ignore metadata entry for version during pruning
-		if bytes.Equal(currKeyEncoded, []byte(latestVersionKey)) || bytes.Equal(currKeyEncoded, []byte(earliestVersionKey)) {
+		// Ignore metadata entries during pruning
+		if isMetadataKey(currKeyEncoded) {
 			itr.Next()
 			continue
 		}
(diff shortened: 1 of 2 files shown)


app/params/config.go
```diff
@@ -155,8 +155,7 @@ func setFullnodeTypeAppConfig(config *srvconfig.Config) {
 	// StateStore: full nodes keep recent history for queries
 	config.StateStore.KeepRecent = 100000
 
-	// MinRetainBlocks: keep 100k blocks for Tendermint pruning
-	config.MinRetainBlocks = 100000
+	// Pruning and MinRetainBlocks use defaults (nothing,0,0 - keep all history)
```
yzang2019: For block, we should still keep only 100k blocks by default

docker/localnode/config/app.toml
```diff
@@ -14,12 +14,12 @@ minimum-gas-prices = "0.01usei"
 # nothing: all historic states will be saved, nothing will be deleted (i.e. archiving node)
 # everything: all saved states will be deleted, storing only the current and previous state; pruning at 10 block intervals
 # custom: allow pruning options to be manually specified through 'pruning-keep-recent', 'pruning-keep-every', and 'pruning-interval'
-pruning = "custom"
+pruning = "nothing"
```
masih, line 17: The thing that's slightly weird for me is that `default` is also a valid config value for this? But now "nothing" is the default unless someone sets it to `default` which is not the default? Have I got that right?

Could we simplify this a bit?

blindchaser: i think this is from tendermint. we choose 'nothing' value as our default value. and tendermint has it's 'default' value as default.

masih: I see, so a user can't set this?

This is totally beyond the scope of this PR but I think if this value is exposed to the user, as in "a user can set this value" then we probably should refine the configuration surface we expose to the users. because operationally the naming can indeed be confusing.

blindchaser: yeah i agree these options doesn't look beautiful. the user can see it but we mark it as auto-managed fields so the user are not supposed to touch it

PR #2491, opened 2025-10-23
Fix gap nonce inclusion

pdrobnjak:
### Intro
Mempool is split into two parts - pending and active.

Every transaction gets classified (in the ante handler) as either pending or active.
Transaction gets classified as pending if it its nonce is a skip nonce.
Transaction gets classified as active if its nonce is the next nonce.

Classification is done based on latest nonce from state.
Transaction can get promoted from pending to active only upon block commitment via [`mempool.Update`](https://github.com/sei-protocol/sei-chain/blob/94f51a514582889c8af929698850d0032d3e74c1/sei-tendermint/internal/mempool/mempool.go#L634).

Promotion is done based on latest nonce from state + pending.
Transaction becomes eligible for promotion by way of another transaction filling the nonce gap via updating the pending nonces.
Every transactions’ nonce gets added to pending nonces in [`mempool.CheckTx`](https://github.com/sei-protocol/sei-chain/blob/94f51a514582889c8af929698850d0032d3e74c1/sei-tendermint/internal/mempool/mempool.go#L303).
Pending nonces are not the same as the pending part of the mempool - every transaction has its nonce included into pending nonces but it does not have to be a pending transaction.

Inclusion to active/pending mempool happens some time after transaction’s nonce is added to pending nonces (it does happen in the same call path).

Because of this we’ve got a rollback mechanism that removes the transaction’s nonce from pending nonces if transaction’s inclusion to active/pending mempool fails among other things.

RC of the issue is that the rollback mechanism is not invoked in every failure scenario which means that transaction's nonce can continue to reside in pending nonces even if the transaction itself got rejected from the mempool.

Examples of failure scenarios that do not invoke the rollback:
[[1.](https://github.com/sei-protocol/sei-chain/blob/94f51a514582889c8af929698850d0032d3e74c1/sei-tendermint/internal/mempool/mempool.go#L439)] [[2.](https://github.com/sei-protocol/sei-chain/blob/94f51a514582889c8af929698850d0032d3e74c1/sei-tendermint/internal/mempool/mempool.go#L767)] [[3.](https://github.com/sei-protocol/sei-chain/blob/94f51a514582889c8af929698850d0032d3e74c1/sei-tendermint/internal/mempool/mempool.go#L818)] [[4.](https://github.com/sei-protocol/sei-chain/blob/94f51a514582889c8af929698850d0032d3e74c1/sei-tendermint/internal/mempool/mempool.go#L821)]

### Issue Scenario

1. Evaluation of `N+1` starts in the mempool -> its nonce is added to pending nonces -> it lands into the pending pool.
2. Evaluation of `N` starts in the mempool -> its nonce is added to pending nonces (this makes `N+1` eligible for promotion to active mempool next time [`mempool.Update`](https://github.com/sei-protocol/sei-chain/blob/94f51a514582889c8af929698850d0032d3e74c1/sei-tendermint/internal/mempool/mempool.go#L634) is called, for [`mempool.Update`](https://github.com/sei-protocol/sei-chain/blob/94f51a514582889c8af929698850d0032d3e74c1/sei-tendermint/internal/mempool/mempool.go#L634) to be called evaluation of `N` needs to finish) -> it fails to get included into the active mempool and the scenario does not invoke the pending nonce rollback mechanism.
3. [`mempool.Update`](https://github.com/sei-protocol/sei-chain/blob/94f51a514582889c8af929698850d0032d3e74c1/sei-tendermint/internal/mempool/mempool.go#L634) runs and promotes `N+1` to active mempool which makes it eligible to be included in a block.
4. `N+1` gets included in a block without `N` being included prior.
### Solution
Add the pending nonce only after transaction gets included successfully into pending/active mempool.

I've created a local test harness that reproduced the gap nonce issue on main - it is unable to reproduce the issue on this branch.

### DIFF
--- a/app/seidb.go
+++ b/app/seidb.go
@@ -21,6 +21,7 @@ const (
 	FlagSCSnapshotKeepRecent               = "state-commit.sc-keep-recent"
 	FlagSCSnapshotInterval                 = "state-commit.sc-snapshot-interval"
 	FlagSCSnapshotWriterLimit              = "state-commit.sc-snapshot-writer-limit"
+	FlagSCSnapshotPrefetchThreshold        = "state-commit.sc-snapshot-prefetch-threshold"
 	FlagSCCacheSize                        = "state-commit.sc-cache-size"
 	FlagSCOnlyAllowExportOnSnapshotVersion = "state-commit.sc-only-allow-export-on-snapshot-version"
 
@@ -86,6 +87,7 @@ func parseSCConfigs(appOpts servertypes.AppOptions) config.StateCommitConfig {
 	scConfig.SnapshotKeepRecent = cast.ToUint32(appOpts.Get(FlagSCSnapshotKeepRecent))
 	scConfig.SnapshotInterval = cast.ToUint32(appOpts.Get(FlagSCSnapshotInterval))
 	scConfig.SnapshotWriterLimit = cast.ToInt(appOpts.Get(FlagSCSnapshotWriterLimit))
+	scConfig.SnapshotPrefetchThreshold = cast.ToFloat64(appOpts.Get(FlagSCSnapshotPrefetchThreshold))
 	scConfig.OnlyAllowExportOnSnapshotVersion = cast.ToBool(appOpts.Get(FlagSCOnlyAllowExportOnSnapshotVersion))
 	return scConfig
 }
--- a/app/seidb_test.go
+++ b/app/seidb_test.go
@@ -26,6 +26,8 @@ func (t TestSeiDBAppOpts) Get(s string) interface{} {
 		return config.DefaultStateCommitConfig().SnapshotKeepRecent
 	case FlagSCSnapshotWriterLimit:
 		return config.DefaultStateCommitConfig().SnapshotWriterLimit
+	case FlagSCSnapshotPrefetchThreshold:
+		return config.DefaultStateCommitConfig().SnapshotPrefetchThreshold
 	case FlagSSEnable:
 		return config.DefaultStateStoreConfig().Enable
 	case FlagSSBackend:
--- a/sei-db/config/config.go
+++ b/sei-db/config/config.go
@@ -1,15 +1,15 @@
 package config
 
 const (
-	DefaultSnapshotInterval   = 10000
-	DefaultSnapshotKeepRecent = 0 // set to 0 to only keep one current snapshot
-	DefaultAsyncCommitBuffer  = 100
-	DefaultCacheSize          = 100000
-	DefaultSSKeepRecent       = 100000
-	DefaultSSPruneInterval    = 600
-	DefaultSSImportWorkers    = 1
-	DefaultSSAsyncBuffer      = 100
-	DefaultSSHashRange        = -1
+	DefaultSnapshotInterval          = 10000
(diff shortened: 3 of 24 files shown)


sei-will (approved):

x/evm/ante/sig.go
```diff
@@ -68,12 +68,9 @@ func (svd *EVMSigVerifyDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulat
 		if txNonce < nextNonce {
 			return ctx, sdkerrors.ErrWrongSequence
 		}
-		ctx = ctx.WithCheckTxCallback(func(thenCtx sdk.Context, e error) {
-			if e != nil {
```
codchen, line 72: is the removal of this check intentional? If there is an error from CheckTx (e.g. signature check failure etc.) then the transaction wouldn't be added to either the pending pool or the active pool (it will just be rejected right away)

pdrobnjak: It is intentional, if you take a look at the block of code where we invoke the callback - https://github.com/sei-protocol/sei-chain/blob/a3039aacef8451865cc0ae54162c5370538f28d5/sei-tendermint/internal/mempool/mempool.go#L407-L446 - we enter the block of code only if error is `nil`, also if one of the three preceding function calls returns an error https://github.com/sei-protocol/sei-chain/blob/a3039aacef8451865cc0ae54162c5370538f28d5/sei-tendermint/internal/mempool/mempool.go#L423-L426 https://github.com/sei-protocol/sei-chain/blob/a3039aacef8451865cc0ae54162c5370538f28d5/sei-tendermint/internal/mempool/mempool.go#L432-L436 https://github.com/sei-protocol/sei-chain/blob/a3039aacef8451865cc0ae54162c5370538f28d5/sei-tendermint/internal/mempool/mempool.go#L438-L440 we will not invoke the callback as we will exit early therefore `err` will always be `nil` when invoking the callback.

And here is the reasoning what happens if a silent fail in preceding function calls happens:

> We add transaction's nonce to pending nonces if [txmp.addNewTransaction](https://github.com/sei-protocol/sei-chain/blob/94f51a514582889c8af929698850d0032d3e74c1/sei-tendermint/internal/mempool/mempool.go#L423) or [txmp.pendingTxs.Insert](https://github.com/sei-protocol/sei-chain/blob/94f51a514582889c8af929698850d0032d3e74c1/sei-tendermint/internal/mempool/mempool.go#L438) do not return an error - [txmp.addNewTransaction](https://github.com/sei-protocol/sei-chain/blob/94f51a514582889c8af929698850d0032d3e74c1/sei-tendermint/internal/mempool/mempool.go#L423) can fail silently in two cases (if it fails silently we will still try to add the transaction's nonce to pending nonces):
> * [Transaction already in mempool](https://github.com/sei-protocol/sei-chain/blob/94f51a514582889c8af929698850d0032d3e74c1/sei-tendermint/internal/mempool/mempool.go#L818) - this will lead to pending nonce addition being noop which is expected behaviour because we rejected the transaction.
> * [Transaction from same sender with same nonce has higher priority](https://github.com/sei-protocol/sei-chain/blob/71e35324f95766547ddb0948ed2551ac9a117901/sei-tendermint/internal/mempool/priority_queue.go#L93) - this will lead to pending nonce addition being noop which is expected behaviour because we rejected the transaction.

codchen (approved):

PR #2497, opened 2025-10-24
feat: optimize memIAVL cold-start with sequential snapshot prefetch

blindchaser:
### Highlights
- 4x-25x faster cold-starts: reduces startup time from ~45 mins to a range of 1.5-15 mins, depending on disk throughput.
- memory efficient: operates within a 60-80GB memory envelope during cold starts.
- shifts bottleneck: become disk throughput-bound.
- hot start (with page cache) on high throughput disk takes only 15 seconds(note: the prev hot start is also fast)

### Key Optimizations
- prefetch snapshot files sequentially replacing random reads
- only prefetch nodes/leaves, skip kv file (sparse and too large for memory)
- use MADV_SEQUENTIAL + MADV_WILLNEED to guide the kernel for optimal I/O scheduling.
- selective preload for big tree only (evm, bank, acc)
- skip prefetch when >80% pages already resident in cache


sei-db/sc/memiavl/multitree.go
```diff
@@ -84,13 +88,17 @@ func LoadMultiTree(dir string, zeroCopy bool, cacheSize int) (*MultiTree, error)
...
+		snapshot, err := OpenSnapshot(filepath.Join(dir, name), opts)
 		if err != nil {
 			return nil, err
 		}
-		treeMap[name] = NewFromSnapshot(snapshot, zeroCopy, cacheSize)
+		treeMap[name] = NewFromSnapshot(snapshot, opts)
+	}
+	timeElapsed := time.Since(startTime).Seconds()
+	log.Info(fmt.Sprintf("All %d memIAVL trees loaded in %.1fs\n", len(treeNames), timeElapsed))
+	if timeElapsed > 600 {
```
yzang2019: Let's change to 300, that might work better

sei-db/sc/memiavl/snapshot.go
```diff
@@ -601,3 +618,190 @@ func (w *snapshotWriter) writeRecursive(node Node) error {
...
+		log.Info(fmt.Sprintf("Tree %s nodes page cache residency ratio is %f, below threshold %f\n", treeName, residentNodes, prefetchThreshold))
+		_ = SequentialReadAndFillPageCache(filepath.Join(snapshotDir, FileNameNodes))
+	}
+
+	if residentLeaves < prefetchThreshold {
+		log.Info(fmt.Sprintf("Tree %s leaves page cache residency ratio is %f, below threshold %f\n", treeName, residentLeaves, prefetchThreshold))
+		_ = SequentialReadAndFillPageCache(filepath.Join(snapshotDir, FileNameLeaves))
+	}
+
+	log.Info(fmt.Sprintf("Prefetch snapshot for %s completed in %fs. Consider adding more RAM for page cache to avoid preloading during restart.\n", treeName, time.Since(startTime).Seconds()))
```
yzang2019: Let's rephrase to "Consider adding more RAM to speed up prefetching when loading snapshots."

PR #2498, opened 2025-10-24
Fixed flaky TestReactor_ValidatorSetChanges (CON-100)

pompon0:
The most problematic issue was that the timeouts were disabled in test.

### DIFF
--- a/sei-tendermint/internal/consensus/reactor_test.go
+++ b/sei-tendermint/internal/consensus/reactor_test.go
@@ -3,8 +3,6 @@ package consensus
 import (
 	"context"
 	"fmt"
-	"math"
-	"math/rand"
 	"os"
 	"sync"
 	"testing"


sei-tendermint/types/vote_set.go
```diff
@@ -138,20 +138,13 @@ func (voteSet *VoteSet) Size() int {
...
 	defer voteSet.mtx.Unlock()
-
 	return voteSet.addVote(vote)
 }
 
 // NOTE: Validates as much as possible before attempting to verify the signature.
 func (voteSet *VoteSet) addVote(vote *Vote) (added bool, err error) {
-	if vote == nil {
-		return false, ErrVoteNil
-	}
```
masih, line 154: What are the side-effects of/rationale for removing this check?

pompon0: Rationale is that the golang convention is that arguments are assumed to be non-nil, unless documented otherwise.

pompon0: all the current call sites of AddVote would panic on nil vote anyway.

masih: Great, the last reply is what i was looking for, thanks.

PR #2503, opened 2025-10-27
Fix sei-db race conditions

yzang2019:
This PR is fixing a bunch of race conditions reported by race detectors in sei-db. It also makes prunng sync.

All unit test should pass now with race detectors enabled

### DIFF
--- a/sei-db/ss/pebbledb/db.go
+++ b/sei-db/ss/pebbledb/db.go
@@ -254,6 +254,8 @@ func (db *Database) SetEarliestVersion(version int64, ignoreVersion bool) error 
 			var ts [VersionSize]byte
 			binary.LittleEndian.PutUint64(ts[:], uint64(version))
 			return db.storage.Set([]byte(earliestVersionKey), ts[:], defaultWriteOpts)
+		} else {
+			return fmt.Errorf("failed to set earliest version to: %d", version)
 		}
 	}
 	return nil
--- a/sei-cosmos/storev2/rootmulti/store.go
+++ b/sei-cosmos/storev2/rootmulti/store.go
@@ -773,7 +773,8 @@ loop:
 	}
 	// initialize the earliest version for SS store
 	if rs.ssStore != nil {
-		rs.ssStore.SetEarliestVersion(height, false)
+		err := rs.ssStore.SetEarliestVersion(height, false)
+		rs.logger.Error("Failed to set earliest version during DB restore", err)
 	}
 
 	return snapshotItem, restoreErr


sei-db/sc/memiavl/db.go
```diff
@@ -592,6 +586,8 @@ func (db *DB) rewriteIfApplicable(height int64) {
 
 	// create snapshot when current height - last snapshot height > interval
 	if height-db.SnapshotVersion() >= int64(db.snapshotInterval) {
+		// Try prune old snapshots before starting new snapshot
+		db.pruneSnapshots()
```
masih, line 590: I am curious if this is related to fixing the race condition in tests. Can I ask you to expand on the rationale please?

sei-db/sc/memiavl/db_test.go
```diff
@@ -100,9 +101,9 @@ func TestRewriteSnapshotBackground(t *testing.T) {
 	require.NoError(t, err)
 
 	// spin up goroutine to keep querying the tree
-	stopped := false
+	stopped := atomic.Bool{}
 	go func() {
-		for !stopped {
+		for !stopped.Load() {
```
masih: This look will run ...hot, in that there is nothing that slows down the iterations of this loop.

I wonder if there is a better way to do what it is trying to do, at least slow it down by some fixed interval?

sei-db/ss/pebbledb/db.go
```diff
@@ -607,16 +613,13 @@ func (db *Database) computeHashForRange(beginBlock, endBlock int64) (_err error)
...
-			version := nextChange.Version
-			if err := db.ApplyChangesetSync(version, nextChange.Changesets); err != nil {
-				panic(err)
-			}
+		version := nextChange.Version
+		if err := db.ApplyChangesetSync(version, nextChange.Changesets); err != nil {
+			panic(err)
 		}
 	}
+	db.asyncWriteWG.Done()
```
masih: I recommend calling this in defer statement to avoid future footguns if the function signature changes to return something.

sei-db/ss/pebbledb/db.go
```diff
@@ -80,6 +80,7 @@ type Database struct {
 	lastRangeHashedCache int64
 	lastRangeHashedMu    sync.RWMutex
 	hashComputationMu    sync.Mutex
+	earliestVersionMu    sync.Mutex
```
masih: 2 things:
1. Wrap access to earliest version consistently with mutex via the getter. I two places where the field is accessed without mutext/directly: `Prune`, and `ReverseIterator`.
2. I suspect the code would easer to read and even slightly faster if you use atomics for learliest version field to begin with.

sei-db/sc/memiavl/db.go
```diff
@@ -425,51 +421,49 @@ func (db *DB) checkBackgroundSnapshotRewrite() error {
 func (db *DB) pruneSnapshots() {
 	// wait until last prune finish
 	db.pruneSnapshotLock.Lock()
+	defer db.pruneSnapshotLock.Unlock()
 
-	go func() {
```
masih, line 429: I am curious why this was done ansync to begin with?

sei-db/sc/memiavl/db_test.go
```diff
@@ -124,7 +125,9 @@ func TestRewriteSnapshotBackground(t *testing.T) {
 		time.Sleep(time.Millisecond * 20)
 	}
 	for db.snapshotRewriteChan != nil {
+		db.mtx.Lock()
 		require.NoError(t, db.checkAsyncTasks())
+		db.mtx.Unlock()
```
masih, line 136: I think we want to refactor the db to assert this condition without  access to the internal `snapshotRewriteChan` - as is it seems prone to causing side-effects. But different battle for another day :)

sei-db/sc/memiavl/db.go
yzang2019: My guess is for performance reasons, but since it's already background thread, I think it's better to simplify the logic and avoid the risk of race condition by turning it into a sync call

sei-db/sc/memiavl/db.go
yzang2019: It's sort of related because if we call db.pruneSnapshots() in Open() function, there will be a test failing for race condition

sei-db/sc/memiavl/db_test.go
yzang2019: Will try add a sleep interval

sei-db/sc/memiavl/db_test.go
yzang2019: Yeah agree, can look into that in the future. For now I think it's fine since only unit test can access snapshotRewriteChan in this way, wdyt?

sei-db/ss/pebbledb/db.go
yzang2019: Yeah, maybe it's easier to use atomic, let me try it out

sei-db/sc/memiavl/db.go
yzang2019: And here we do pruning before snapshot creation to make sure if the snapshot fail in the middle, it doesn't end up leaking unpruned snapshots

masih: Thanks Yiming

sei-db/sc/memiavl/db.go
masih: It is great to see one less adhoc goroutine start that modifies state. Thanks Yiming!

sei-db/sc/memiavl/db_test.go
masih: A perhaps cleaner way to do this is to expose APIs in testing scope only :) Happy to discuss how in 1:1. 

For the purposes of this PR, there is nothing actionable just now  and we can totally do this later

sei-db/sc/memiavl/db_test.go
masih: Please use timer instead of `time.Sleep`?

sei-db/ss/pebbledb/db.go
```diff
@@ -147,11 +147,12 @@ func New(dataDir string, config config.StateStoreConfig) (*Database, error) {
 		storage:         db,
 		asyncWriteWG:    sync.WaitGroup{},
 		config:          config,
-		earliestVersion: earliestVersion,
+		earliestVersion: atomic.Int64{},
 		latestVersion:   atomic.Int64{},
 		pendingChanges:  make(chan VersionedChangesets, config.AsyncWriteBuffer),
 	}
 	database.latestVersion.Store(latestVersion)
+	database.earliestVersion.Store(earliestVersion)
```
masih, line 155: Use Setter ? A consistent behaviour in setting is i think highly desirable, regardless of whether we are setting in the constructor or not, unless there is a really really good reason for not using the setter consistently in which case an inline comment to explain why would be much appreciated.

sei-db/ss/pebbledb/db.go
```diff
@@ -245,8 +247,8 @@ func (db *Database) SetEarliestVersion(version int64, ignoreVersion bool) error
 	if version < 0 {
 		return fmt.Errorf("version must be non-negative")
 	}
-	if version > db.earliestVersion || ignoreVersion {
-		db.earliestVersion = version
+	if version > db.GetEarliestVersion() || ignoreVersion {
```
masih: This makes `SetEarliestVersion` non-atomic. Because between `GetEarliestVersion` called here and `earliestVersion.Store` called below the value of version might have changed now that there is no mutex.

To correctly handle this we should be using `CompareAndSwap` I think.

sei-db/ss/pebbledb/db.go
yzang2019: I think we can't simply use setter because setter will trigger writes to DB, here we only wanna update the in memory value for earliest version and avoid writing to DB

PR #2504, opened 2025-10-27
Remove aclaccesscontrol module and usages

udpatil:
This removes the AccessControl module since it is no longer being used.

Upgrade testing in lower environments + unit tests

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -3141,8 +3141,6 @@ gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b/go.mod h1:K4uyk7z7BCEPqu6E+C
 gopkg.in/yaml.v3 v3.0.0/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
 gopkg.in/yaml.v3 v3.0.1 h1:fxVm/GzAzEWqLHuvctI91KS9hhNmmWOoWu0XTYJS7CA=
 gopkg.in/yaml.v3 v3.0.1/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
-gotest.tools v2.2.0+incompatible h1:VsBPFP1AI068pPrMxtb/S8Zkgf9xEmTLJjfM+P5UIEo=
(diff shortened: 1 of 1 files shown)


masih (approved):
There are not many things in code review that are sweeter than a PR that removes 75X more code than it adds 🍻 

Small cleanups that might also belong here:
* Remove commented out code [here](https://github.com/sei-protocol/sei-chain/blob/7a56eb4e8487455e696cc502d422968d6f8efe41/app/app.go#L1415-L1428)?
* Double checking that app state doesn't need a clean up since I see access control in [state dump script](https://github.com/sei-protocol/sei-chain/blob/da9d59c8d2d4b2849333be01ad8e49e9c91ad606/scripts/dump_app_state_for_height.sh#L31)? If not then it's worth updating the script too? I have little context as to how out-dated that script is.
* What updates the Cosmos swagger-ui? it seems to still [reference accesscontrol](https://github.com/sei-protocol/sei-chain/blob/5deb1e0f74317f13694557d4b73337382bca820a/docs/swagger-ui/swagger.yml#L7)?
* The readme of cosmos protos is very out of date but we might as well remove [this](https://github.com/sei-protocol/sei-chain/blob/06a4e242bf80fff303be607734e121bd2f0f6916/sei-cosmos/x/README.md?plain=1#L10) line from it?
* Do we also need to remove the `aclaccesscontrol` module across the codebase? I see it is mentioned in [sei-db](https://github.com/sei-protocol/sei-chain/blob/5056e8e9d7bc83c9d05d8af2e88d9abf5c16bd6a/sei-db/ss/util/modules.go#L4) and [sei-chain app](https://github.com/sei-protocol/sei-chain/blob/07441d7bfcd7f9fc69119cf3002be7d6912b3a87/cmd/seid/cmd/debug.go#L29) for example.

Otherwise, LGTM 🚀

app/app.go
```diff
@@ -1077,6 +1042,15 @@ func (app *App) SetStoreUpgradeHandlers() {
 		// configure store loader that checks if version == upgradeHeight and applies store upgrades
 		app.SetStoreLoader(upgradetypes.UpgradeStoreLoader(upgradeInfo.Height, &storeUpgrades))
 	}
+
+	if (upgradeInfo.Name == "vTODO") && !app.UpgradeKeeper.IsSkipHeight(upgradeInfo.Height) {
```
udpatil, line 1016: cc @stevenlanders @philipsu522 when do we want to actually remove this module? Do we want to wait to merge in this PR until we want to remove this module at which point we can update this line with the appropriate upgrade name?

PR #2517, opened 2025-10-29
Remove vote extensions logic

udpatil:
This removes vote extensions logics from tendermint/abci since we don't need them


udpatil:
Addressed some of the comments about remaining voteExtension references. The remaining VoteExtensions references are the ABCI params in ConsensusParams, which will need to be migrated to properly update the chain state for ConsensusParams as well.

PR #2522, opened 2025-10-31
Flatten BeginBlock and remove nested logics

codchen:
Previously, BeginBlock follows a very involved call path:
```
SeiApp->BaseApp->SeiApp->ModuleManager->Modules->keepers
```
This PR simplifies BeginBlock to be `SeiApp->keepers`

no functional change

### DIFF
--- a/app/legacyabci/begin_block.go
+++ b/app/legacyabci/begin_block.go
@@ -1,6 +1,9 @@
 package legacyabci
 
 import (
+	"time"
+
+	"github.com/cosmos/cosmos-sdk/telemetry"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/x/capability"
 	capabilitykeeper "github.com/cosmos/cosmos-sdk/x/capability/keeper"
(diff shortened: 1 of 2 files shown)


app/legacyabci/begin_block.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func BeginBlock(
+	ctx sdk.Context,
+	height int64,
+	votes []abci.VoteInfo,
+	byzantineValidators []abci.Misbehavior,
+	keepers BeginBlockKeepers,
+) {
+	keepers.EpochKeeper.BeginBlock(ctx)
```
udpatil, line 52: why do we have some with `BeginBlocker` and others calling the `BeginBlock` function?

sei-cosmos/x/capability/module.go
```diff
@@ -155,11 +153,3 @@ func (am AppModule) ExportGenesisStream(ctx sdk.Context, cdc codec.JSONCodec) <-
 
 // ConsensusVersion implements AppModule/ConsensusVersion.
 func (AppModule) ConsensusVersion() uint64 { return 1 }
-
-// BeginBlocker calls InitMemStore to assert that the memory store is initialized.
-// It's safe to run multiple times.
-func (am AppModule) BeginBlock(ctx sdk.Context, _ abci.RequestBeginBlock) {
-	defer telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), telemetry.MetricKeyBeginBlocker)
-
-	am.keeper.InitMemStore(ctx)
```
udpatil, line 164: why was this not previously calling capability's `BeginBlocker` function which also just calls the `InitMemStore`?

app/legacyabci/begin_block.go
codchen: @udpatil ideally I'd prefer those to all be `keeper.BeginBlocker`, but since some modules already have it defined it on the module namespace, I just reused those for now

sei-cosmos/x/capability/module.go
codchen: i'm not quite sure - could just be a refactor oversight from cosmos-sdk

PR #2525, opened 2025-10-31
Deprecate store streaming and listeners

udpatil:
This removes listenkv and associated logic for store streaming and listeners. As is, this feature is unused, and likely fails to sufficiently scale with future scaling goals.
Existing testing


codchen (approved):

masih (approved):
Nice work @udpatil 🚀

PR #2526, opened 2025-11-03
Add cross arch build CI job

udpatil:
Recently we've had some minor compatibility issues across linux / darwin,  this will add a CI target to simply verify build/install across these multiple targets

### DIFF
--- a/.github/workflows/cross-arch-build.yml
+++ b/.github/workflows/cross-arch-build.yml
@@ -5,11 +5,13 @@ on:
   push:
     branches:
       - main
-      - release/**
 
 concurrency:
   cancel-in-progress: true
   group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.event_name == 'push' && github.sha || github.ref }}
+
+env:
+  LEDGER_ENABLED: false
 
 jobs:
   linux-amd64:
@@ -18,8 +20,6 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v5
-        with:
-          submodules: true
 
       - name: Set up Go
         uses: actions/setup-go@v6
@@ -42,33 +42,15 @@ jobs:
 
       - name: Run make install
         run: make install
-        env:
-          LEDGER_ENABLED: false
 
       - name: Verify install binary
-        run: |
-          if command -v seid &> /dev/null; then
-            echo "seid binary installed successfully"
-            seid version --long
-          else
-            echo "seid binary not found in PATH"
-            exit 1
-          fi
+        run: seid version --long
 
       - name: Run make build
         run: make build
-        env:
-          LEDGER_ENABLED: false
 
       - name: Verify build binary
-        run: |
-          if [ -f ./build/seid ]; then
-            echo "seid binary built successfully"
-            ./build/seid version --long
-          else
-            echo "seid binary not found in build directory"
-            exit 1
-          fi
+        run: ./build/seid version --long
 
   macos-arm64:
     name: "macOS ARM64"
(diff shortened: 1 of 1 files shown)


masih (commented):
I am missing context on the compatibility issues reported by ecosystem, appologies. 

Questions: 
* would it make sense to use `goreleaser` instead and have it build executables for all platforms in a single CI workflow?  We should do this eventually for sure IMHO. Thinking if the time for it is now or later.

* Instead of triggering job per GOOS then one to sum them up, would it make sense to run this in a single runner linearly and fail fast? that should be fast enough and much cheaper.

.github/workflows/cross-arch-build.yml
```diff
@@ -0,0 +1,128 @@
...
+
+jobs:
+  linux-amd64:
+    name: "Linux AMD64"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+        with:
+          submodules: true
```
masih: * Are there any submodules for this to need to be `true`?
* While at it, also set fetch-depth for faster checkout?
   * Ditto for other job.

.github/workflows/cross-arch-build.yml
```diff
@@ -0,0 +1,128 @@
...
+          echo "GOARCH: $(go env GOARCH)"
+          echo "GOOS: $(go env GOOS)"
+
+      - name: Download Go dependencies
+        run: go mod download
+
+      - name: Run make install
+        run: make install
+        env:
+          LEDGER_ENABLED: false
```
masih: `LEDGER_ENABLED` is always set to false. _If it really is meant to be_ instead of repeating it per step maybe capture as top level `env` for all jobs?

.github/workflows/cross-arch-build.yml
```diff
@@ -0,0 +1,128 @@
...
+        run: |
+          go version
+          echo "GOARCH: $(go env GOARCH)"
+          echo "GOOS: $(go env GOOS)"
+
+      - name: Download Go dependencies
+        run: go mod download
+
+      - name: Run make install
+        run: make install
```
masih, line 44: Conventionally, I recommend moving `run` after all other keys. I find the flow of the code be easier to read, where I see all assumptions, config, etc. before i see the command that is being executed.

.github/workflows/cross-arch-build.yml
```diff
@@ -0,0 +1,128 @@
...
+        env:
+          LEDGER_ENABLED: false
+
+      - name: Verify install binary
+        run: |
+          if command -v seid &> /dev/null; then
```
masih: Why these `if` statements at all? We want the flow to fail of the `seid` command does not execute successfully, right? This will happen whether `seid` exists or not, and what's more: the error we get would be clear enough for us to infer whether it failed because `seid` didn't exist or something else. If I have not missed anything, then we could reduce the run to simply `seid <whatever>`.

Ditto for all other steps following this pattern.

.github/workflows/cross-arch-build.yml
```diff
@@ -0,0 +1,128 @@
...
+            seid version --long
+          else
+            echo "seid binary not found in PATH"
+            exit 1
+          fi
+
+      - name: Run make build
+        run: make build
+        env:
+          LEDGER_ENABLED: false
```
masih: Did you mean to set this to `true`? If so i recommend updating the step name too to reflect the rationale.

masih (commented):
I think there is one blocker with `LEDGER_ENABLED` settings across builds. Left a bunch of suggestions that should reduce LoC. Please let me know if I can expand on anything 🍻

.github/workflows/cross-arch-build.yml
udpatil: No need for submodules, removing this. Fetch depth defaults to only fetching the specific commit, which is more efficient, right? See docs here: https://github.com/actions/checkout?tab=readme-ov-file#checkout-v4

masih (approved):

PR #2533, opened 2025-11-05
Migrate transaction embedding proto types to Go types

pdrobnjak:

1st step as described in description of #2534.

Couple of points:
* We have not entirely removed [`TxResult`](https://github.com/sei-protocol/sei-chain/blob/d99609b7efcdf63be47376d606152ef8baaab1ce/sei-tendermint/proto/tendermint/abci/types.proto#L446) as it is an actual wire type, it is not only used on the artificial ABCI boundary but as a storage wire type as well: [[1.](https://github.com/sei-protocol/sei-chain/blob/fe8f7dffde5b02bdb2dae952873c50be8c775727/sei-tendermint/internal/state/indexer/tx/kv/kv.go#L86-L94)] [[2.](https://github.com/sei-protocol/sei-chain/blob/fe8f7dffde5b02bdb2dae952873c50be8c775727/sei-tendermint/internal/state/indexer/sink/psql/psql.go#L185-L214)]
* We have not yet tackled [`TxRecord`](https://github.com/sei-protocol/sei-chain/blob/d99609b7efcdf63be47376d606152ef8baaab1ce/sei-tendermint/proto/tendermint/abci/types.proto#L453) as it is embedded in [`ResponsePrepareProposal`](https://github.com/sei-protocol/sei-chain/blob/d99609b7efcdf63be47376d606152ef8baaab1ce/sei-tendermint/proto/tendermint/abci/types.proto#L356) this will be done in a separate PR.

### DIFF
--- a/.github/workflows/cross-arch-build.yml
+++ b/.github/workflows/cross-arch-build.yml
@@ -1,0 +1,92 @@
+name: Cross-Architecture Build Test
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+
+concurrency:
+  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.event_name == 'push' && github.sha || github.ref }}
+
+env:
+  LEDGER_ENABLED: false
+
+jobs:
+  linux-amd64:
+    name: "Linux AMD64"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version: '1.24'
+
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y build-essential
+
+      - name: Verify Go installation
+        run: |
+          go version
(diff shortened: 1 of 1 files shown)


sei-tendermint/proto/tendermint/abci/types.proto
```diff
@@ -35,10 +34,8 @@ message Request {
     RequestProcessProposal process_proposal = 16;
     RequestFinalizeBlock finalize_block = 19;
     RequestBeginBlock begin_block = 20;
-    RequestDeliverTx deliver_tx = 21;
     RequestEndBlock end_block = 22;
     RequestLoadLatest load_latest = 23;
-    RequestGetTxPriorityHint get_tx_priority_hint = 24;
```
pompon0, line 41: reserve ids and names

sei-will (approved):

sei-tendermint/proto/tendermint/abci/types.proto
pdrobnjak: https://github.com/sei-protocol/sei-chain/pull/2533/commits/212210cc138542b6787c141f3491abda773482af

PR #2537, opened 2025-11-06
optimization: reduce snapshot creation time

blindchaser:
Implement pipeline-based parallel writing to significantly improve snapshot rewrite performance.
- Prefech files into page cache before snapshot creation
- Pipeline parallel writes: Decouple traversal and I/O via 3 background workers
- Large IO/channel buffer: 10k → 500k to reduce blocking and syscalls
- Priority EVM strategy: Write largest tree first to avoid I/O contention
- Skip too frequent snapshot during catch-up state

- On a 250GB RAM machine with 250 MiB/s GT3 disk, snapshot creation time reduced from >3 hours to ~20 minutes.


sei-db/common/metrics/metrics.go
```diff
@@ -1,4 +1,4 @@
-package memiavl
+package metrics
```
yzang2019: I think we should probably revert this change, we intentionally move the metrics from common to individual db package

sei-db/sc/memiavl/db.go
```diff
@@ -228,6 +232,10 @@ func OpenDB(logger logger.Logger, targetVersion int64, opts Options) (database *
 		}
 	}
 
+	// We need to prune snapshots during start up to avoid snapshot leaks
+	if !db.readOnly {
+		db.pruneSnapshots()
```
yzang2019: Did you add this by accident? I think we probably don't need it

sei-db/sc/memiavl/db.go
```diff
@@ -428,63 +438,65 @@ func (db *DB) checkBackgroundSnapshotRewrite() error {
...
 	// wait until last prune finish
 	db.pruneSnapshotLock.Lock()
-	defer db.pruneSnapshotLock.Unlock()
 
-	currentVersion, err := currentVersion(db.dir)
-	if err != nil {
-		db.logger.Error("failed to read current snapshot version", "err", err)
-		return
-	}
+	go func() {
```
yzang2019: why this change? Seems irrelevant to the PR

sei-db/sc/memiavl/db.go
```diff
@@ -548,14 +560,68 @@ func (db *DB) RewriteSnapshot(ctx context.Context) error {
 	}
 
 	snapshotDir := snapshotName(db.lastCommitInfo.Version)
+
+	// Skip if snapshot already exists
+	targetPath := filepath.Join(db.dir, snapshotDir)
+	if _, err := os.Stat(targetPath); err == nil {
```
masih: I recommend making a stronger assertion that just checking if the directory exists.

sei-db/sc/memiavl/db.go
```diff
@@ -87,6 +89,13 @@ type DB struct {
 const (
 	SnapshotPrefix = "snapshot-"
 	SnapshotDirLen = len(SnapshotPrefix) + 20
+
+	// Catch-up detection thresholds for snapshot creation
+	// During rapid catch-up (e.g., state sync), we skip snapshot creation if:
+	// - Block interval is large (> CatchupBlockThreshold)
```
yzang2019: What is this mainly for?

sei-db/sc/memiavl/db.go
```diff
@@ -548,14 +559,68 @@ func (db *DB) RewriteSnapshot(ctx context.Context) error {
...
+	targetPath := filepath.Join(db.dir, snapshotDir)
+	if _, err := os.Stat(targetPath); err == nil {
+		db.logger.Info("snapshot already exists, skipping rewrite", "snapshot", snapshotDir)
+		// Ensure 'current' symlink points to this snapshot
+		if err := updateCurrentSymlink(db.dir, snapshotDir); err != nil {
+			db.logger.Error("failed to update current symlink for existing snapshot", "snapshot", snapshotDir, "error", err)
+			return err
+		}
+		return nil
+	}
```
yzang2019: This might cause issues if the target snapshot is partially written or corrupted. Curious why we do that and under what condition will target snapshot already exist when we create it?

sei-db/sc/memiavl/import.go
```diff
@@ -12,8 +12,13 @@ import (
 )
 
 var (
-	nodeChanSize = 10000
-	bufIOSize    = 64 * 1024 * 1024
+	// Pipeline channel size - controls how many operations can be queued
+	// Increased to 1M to improve throughput (1M ops * 120 bytes * 3 channels = ~360MB)
+	nodeChanSize = 1000000
```
yzang2019: Why do we need to queue that many nodes? If the underlying channel processing is fast enough, then there's no item should be backed up, if the underlying channel processing is slow, then just increasing the queue size but not batching them also doesn't help

sei-db/sc/memiavl/import.go
```diff
@@ -129,35 +136,62 @@ func (ai *TreeImporter) Close() error {
...
-			snapshotWriter: *w,
+			w:           w,
+			leavesStack: make([]uint32, 0),
+			nodeStack:   make([]*MemNode, 0),
 		}
 
+		// Check for context cancellation every 100k nodes to minimize overhead
+		// This provides ~1 second response time for EVM tree (1B nodes / 100k = 10k checks at 1M nodes/s)
+		// while avoiding per-node overhead that caused 20% performance degradation
+		const cancelCheckInterval = 100000
```
yzang2019, line 161: Does it really help minimize overhead? What about the extra overhead we introduced here from incrementing nodeCount++ and doing the math operation of mod might actually? I'd recommend not adding this unless we are 100% sure this has significant improvement

sei-db/sc/memiavl/snapshot.go
```diff
@@ -685,9 +1030,35 @@ func SequentialReadAndFillPageCache(filePath string) error {
...
+		if err == nil {
+			// Tell kernel this will be read sequentially - enables aggressive readahead
+			_ = unix.Madvise(data, unix.MADV_SEQUENTIAL)
+			// Tell kernel we need this data soon - start readahead immediately
+			_ = unix.Madvise(data, unix.MADV_WILLNEED)
+			// Unmap after setting hints - the hints persist on the underlying pages
+			defer unix.Munmap(data)
+			fmt.Printf("[PREFETCH] Applied madvise hints (SEQUENTIAL + WILLNEED) to %s\n", filePath)
+		} else {
+			fmt.Printf("[PREFETCH] Warning: mmap failed for %s: %v (continuing with read)\n", filePath, err)
```
yzang2019: I don't quite understand why we need this step? Can we explain  it a bit more? Shouldn't we already opened and mmap the data? Why do we need to do it again here

sei-db/sc/memiavl/db.go
```diff
@@ -635,28 +715,42 @@ func (db *DB) rewriteSnapshotBackground() error {
...
 			return
 		}
-		cloned.logger.Info("finished rewriting snapshot", "version", cloned.Version())
-		mtree, err := LoadMultiTree(currentPath(cloned.dir), db.opts)
+		cloned.logger.Info("finished rewriting snapshot", "version", cloned.Version(), "elapsed", time.Since(rewriteStart).Seconds())
+
+		loadStart := time.Now()
+		// Disable prefetch to avoid cache interference with main chain
+		loadOpts := db.opts
+		loadOpts.PrefetchThreshold = 0
```
yzang2019, line 801: Why do we disable prefetch here? We do wanna speed up the catch up with hot data right?

sei-db/sc/memiavl/db.go
blindchaser: was meant to optimize when existing but ignored the corrupted. removed.

sei-db/sc/memiavl/import.go
blindchaser: nodeCount++: ~1 CPU cycle (negligible)
- nodeCount % 100000: ~10-20 CPU cycles every 100k nodes
- For 1B nodes: only 10,000 checks total
- Total overhea is pretty small i think

blindchaser: withouht this it looks sync imports can't be interrupted (could take long time for large trees), this about operability and graceful shutdown

sei-db/sc/memiavl/db.go
blindchaser: will use exisitng param. addrssed

sei-db/sc/memiavl/import.go
blindchaser: Based on profiling (256GB RAM, 250MiB/s disk): peak channel usage ~440k nodes with ~1GB overhead. Small buffers cause frequent producer blocking and context switching. The 1GB cost is not that much but avoiding blocking

sei-db/sc/memiavl/snapshot.go
blindchaser: this is reduntant code, removed

sei-db/sc/memiavl/db.go
blindchaser: this is for the running time snapshot creation. the catchup step looks pretty fast even without the re-prefetch. and the prefetch takes some time also contention with main thread. 

however, did some testing and what happens is during runtime the page cache ratio is always > 0.8 so this doens't make any impact. i will remove it.

sei-db/sc/memiavl/import.go
yzang2019: Sounds good!

sei-db/sc/memiavl/db.go
```diff
@@ -204,17 +220,23 @@ func OpenDB(logger logger.Logger, targetVersion int64, opts Options) (database *
 	// create worker pool. recv tasks to write snapshot
 	workerPool := pond.New(opts.SnapshotWriterLimit, opts.SnapshotWriterLimit*10)
 
+	// Initialize lastSnapshotTime from the current snapshot directory's modification time
+	// This ensures accurate time tracking even after restarts
+	lastSnapshotTime := getSnapshotModTime(opts.Dir, mtree.Version())
```
yzang2019: This is latest version which keeps changing, but we should just look at the "current" folder

yzang2019 (approved):

sei-db/config/config.go
```diff
@@ -2,7 +2,8 @@ package config
 
 const (
 	DefaultSnapshotInterval          = 10000
-	DefaultSnapshotKeepRecent        = 0 // set to 0 to only keep one current snapshot
+	DefaultSnapshotKeepRecent        = 0    // set to 0 to only keep one current snapshot
+	DefaultSnapshotMinTimeInterval   = 3600 // 1 hour in seconds
```
masih: Subjective: I usually write these as multiples of numbers to make it obvious what the value is as a gesture to make code more readable. For example, I would write this as `60*60`. 

A far better approach, would have been to type `SnapshotMinTimeInterval` as `time.Duration`. That way, the value of this constant would have been `time.Hour`. In terms of parsing the value from config, we would have then parsed via `time.time.ParseDuration` which accepts human readable notation like `1h`. I understand there might be weeds here with parsing via cobra TOML library etc. and there is indeed a larger task to untangle all of that which is well beyond the scope of this PR :)

sei-db/config/toml.go
```diff
@@ -29,6 +29,10 @@ sc-keep-recent = {{ .StateCommit.SnapshotKeepRecent }}
 # SnapshotInterval defines the block interval the snapshot is taken, default to 10000 blocks.
 sc-snapshot-interval = {{ .StateCommit.SnapshotInterval }}
 
+# SnapshotMinTimeInterval defines the minimum time interval (in seconds) between snapshots.
+# This prevents excessive snapshot creation during catch-up. Default to 3600 seconds (1 hour).
+sc-snapshot-min-time-interval = {{ .StateCommit.SnapshotMinTimeInterval }}
```
masih, line 37: Is 1 hour a safe min snapshot interval when the node is _not_ in catch-up mode? I am wondering if it might be too long considering the expected median time to finality is ~400ms.

@blindchaser @yzang2019 thoughts?

sei-db/sc/memiavl/db.go
```diff
@@ -89,6 +93,27 @@ const (
...
+// getSnapshotModTime returns the modification time of the current snapshot directory.
+// It reads the "current" symlink to get the actual snapshot directory.
+// If the directory doesn't exist or there's an error, returns current time.
+func getSnapshotModTime(dir string) time.Time {
+	// Read the "current" symlink to get the actual snapshot directory
+	currentLink := currentPath(dir)
+	snapshotName, err := os.Readlink(currentLink)
+	if err != nil {
+		// If we can't read the symlink, use current time as fallback
+		return time.Now()
```
masih, line 127: I highly recommend logging the error at least? There might be a deeper reason for the error.

sei-db/sc/memiavl/db.go
```diff
@@ -89,6 +93,27 @@ const (
...
+	}
+
+	snapshotDir := filepath.Join(dir, snapshotName)
```
masih: _Always_ clean the path before using it, and strictly check expected parent when appropriate.

sei-db/sc/memiavl/db.go
```diff
@@ -89,6 +93,27 @@ const (
...
+	info, err := os.Stat(snapshotDir)
+	if err != nil {
+		// If we can't get the modification time, use current time as fallback
+		return time.Now()
```
masih, line 139: I think we should check the error type then log depending on the error and then fallback. For example, is it OK for the file not to exist? Reading this code without that context surprises me, in that "getting the last snapshot time is now when there wasn't any". This kind of context written as comment would be super useful to an external observer or reader of the code.

sei-db/sc/memiavl/db.go
```diff
@@ -89,6 +93,27 @@ const (
...
+	}
+	return info.ModTime()
```
masih, line 142: Could we think of anything going wrong by relying on OS file system mod time? For example, the mod time might not represent the actual time the snapshot was taken if a validator restores the files from some backup. 

Perhaps, we should consider internal data structure in DB to maintain this type of metadata to avoid a whole bunch of edge cases.

sei-db/sc/memiavl/db.go
```diff
@@ -63,6 +63,10 @@ type DB struct {
 	snapshotKeepRecent uint32
 	// block interval to take a new snapshot
 	snapshotInterval uint32
+	// minimum time interval between snapshots (in seconds)
+	snapshotMinTimeInterval uint32
+	// timestamp of the last successful snapshot creation
+	lastSnapshotTime time.Time
```
masih, line 71: Double checking that we don't need to protect this with a mutex.

sei-db/sc/memiavl/db.go
```diff
@@ -548,14 +581,56 @@ func (db *DB) RewriteSnapshot(ctx context.Context) error {
...
+			"error", err,
+		)
+		cleanupErr := os.RemoveAll(path)
+		if cleanupErr != nil {
+			db.logger.Error("failed to clean up temporary snapshot directory",
+				"tmpDir", tmpDir,
+				"cleanup_error", cleanupErr,
+			)
+		} else {
+			db.logger.Info("temporary snapshot directory cleaned up successfully",
```
masih: Change level to debug?

sei-db/sc/memiavl/db.go
```diff
@@ -548,14 +581,56 @@ func (db *DB) RewriteSnapshot(ctx context.Context) error {
 	}
 
 	snapshotDir := snapshotName(db.lastCommitInfo.Version)
+	targetPath := filepath.Join(db.dir, snapshotDir)
```
masih: `filepath.Clean`. Ditto for the rest of the PR (and other pre-existing lines while we are at it :) )

sei-db/sc/memiavl/db.go
```diff
@@ -592,10 +667,30 @@ func (db *DB) rewriteIfApplicable(height int64) {
...
-		db.pruneSnapshots()
+	snapshotVersion := db.SnapshotVersion()
+	blocksSinceLastSnapshot := height - snapshotVersion
+
+	// Create snapshot when both conditions are met:
+	// 1. Block height interval is reached (height - last snapshot >= interval)
+	// 2. Minimum time interval has elapsed (prevents excessive snapshots during catch-up)
+	if blocksSinceLastSnapshot >= int64(db.snapshotInterval) {
+		timeSinceLastSnapshot := time.Since(db.lastSnapshotTime)
+		minTimeInterval := time.Duration(db.snapshotMinTimeInterval) * time.Second
```
masih: Looks like we are converting things to drutation. Would be good to change the config at leat at the level of DB struct instantiation to take `time.Duration` straight-up instead of `int32`.

sei-db/sc/memiavl/db.go
```diff
@@ -635,28 +730,39 @@ func (db *DB) rewriteSnapshotBackground() error {
 		defer close(ch)
 		startTime := time.Now()
 		cloned.logger.Info("start rewriting snapshot", "version", cloned.Version())
```
masih, line 782: Double checking if we need to copy over any other snapshot related fields in call to `Copy` at line 728?

sei-db/sc/memiavl/db_rewrite_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"testing"
+
+	"github.com/sei-protocol/sei-db/common/logger"
+	"github.com/sei-protocol/sei-db/proto"
+	"github.com/stretchr/testify/require"
+)
+
+// TestRewriteIfApplicableLogging tests the logging in rewriteIfApplicable
+func TestRewriteIfApplicableLogging(t *testing.T) {
+	t.Skip("Skipping background snapshot test - timing dependent")
```
masih: Same comment as the previous PR review. Why have this empty test with no body then skip it?

If A test should exist and missing, then I would capture an issue for it in the backlog and remove this test entirely.

sei-db/sc/memiavl/db_test.go
```diff
@@ -182,7 +183,10 @@ func TestSnapshotTriggerOnIntervalDiff(t *testing.T) {
 		require.EqualValues(t, 0, db.MultiTree.SnapshotVersion())
 	}
 
-	// Height 5 should trigger rewrite
+	// Wait for minimum time interval to elapse (1 second + buffer)
+	time.Sleep(1100 * time.Millisecond)
```
masih, line 195: I am worried that this test is going to be flaky as a result of this kind of sleep.

Instead write a function that checks the condition you want is met and run it with `require.Eventually`?

sei-db/sc/memiavl/db_test.go
```diff
@@ -553,6 +558,10 @@ func TestFastCommit(t *testing.T) {
 		require.NoError(t, db.ApplyChangeSets([]*proto.NamedChangeSet{{Name: "test", Changeset: cs}}))
 		_, err := db.Commit()
 		require.NoError(t, err)
+		// After reaching snapshot interval (3), wait for time threshold to be met
+		if i == 2 {
+			time.Sleep(1100 * time.Millisecond)
```
masih, line 576: Ditto re flakiness.

sei-db/sc/memiavl/import.go
```diff
@@ -12,8 +12,13 @@ import (
 )
 
 var (
-	nodeChanSize = 10000
-	bufIOSize    = 64 * 1024 * 1024
+	// Pipeline channel size - controls how many operations can be queued
+	// Profiling: peak 440k nodes, 1GB overhead. Small buffer causes frequent blocking.
+	nodeChanSize = 500000
+
+	// bufio.Writer buffer size - 128MB balances performance and memory usage
```
masih: Sorry for being pedantic 😄 , but that's technically 128MiB, not 128MB. Right?
```suggestion
	// bufio.Writer buffer size - 128MiB balances performance and memory usage
```

sei-db/sc/memiavl/import.go
```diff
@@ -12,8 +12,13 @@ import (
...
+	// Large buffer reduces syscalls and improves throughput for sequential writes
+	bufIOSize = 128 * 1024 * 1024
```
masih: Not sure if you know the `<<` trick for writing these: `128 << 20`.10 gives you KiB, 20 gives you MiB, 30 gives you GiB and so on.

sei-db/sc/memiavl/import.go
masih: A safer way to write this is to also check for time elapsed since start. The right way to improve this though is to change add to take context. I see it calls `writeBatch` and batch has a select statement which is where the context should be checked. ...But we give batch write one node? which might be a hint to improve underlying write batching mechanism (beyond the scope of this PR).

sei-db/sc/memiavl/multitree.go
```diff
@@ -96,8 +99,8 @@ func LoadMultiTree(dir string, opts Options) (*MultiTree, error) {
 	}
 	timeElapsed := time.Since(startTime).Seconds()
 	log.Info(fmt.Sprintf("All %d memIAVL trees loaded in %.1fs\n", len(treeNames), timeElapsed))
-	if timeElapsed > 300 {
-		log.Info("Loading MemIAVL tree from disk is too slow! Consider increasing the disk bandwidth to speed up the tree loading time within 300 seconds.\n")
+	if timeElapsed > 600 {
```
masih: Questions:
* Where did 300 come from and why we are doubling it?
* Would it make more sense to log as Warn instead of info?

sei-db/sc/memiavl/multitree.go
```diff
@@ -401,28 +405,74 @@ func (t *MultiTree) Catchup(stream types.Stream[proto.ChangelogEntry], endVersio
...
 
-	// write the snapshots in parallel and wait all jobs done
-	group, _ := wp.GroupContext(ctx)
+	// Write EVM first to avoid disk I/O contention, then parallel
+	return t.writeSnapshotPriorityEVM(ctx, dir, wp)
+}
+
+// writeSnapshotPriorityEVM writes EVM tree first, then others in parallel
+// Best strategy: reduces disk I/O contention for the largest tree
+func (t *MultiTree) writeSnapshotPriorityEVM(ctx context.Context, dir string, wp *pond.WorkerPool) error {
```
masih, line 426: Beyond the scope of this PR: i recommend removing dependency to `pond` custom library in favour of what `sync` package offers. ErrorGroup for example would fit here. concurrency utilities in Go have come a long way.

sei-db/sc/memiavl/snapshot.go
```diff
@@ -497,19 +587,185 @@ type snapshotWriter struct {
...
+// Larger values provide more parallelism but use more memory
+// Default is 10000. Recommended range: 1000-50000
+func SetPipelineBufferSize(size int) {
+	if size < 100 {
+		size = 100 // Minimum to avoid deadlocks
+	}
+	if size > 100000 {
+		size = 100000 // Maximum to avoid excessive memory usage
+	}
+	nodeChanSize = size
```
masih: Simplify this by using built-in `min` `max` functions?

sei-db/sc/memiavl/import.go
```diff
@@ -12,8 +12,13 @@ import (
 )
 
 var (
-	nodeChanSize = 10000
-	bufIOSize    = 64 * 1024 * 1024
+	// Pipeline channel size - controls how many operations can be queued
+	// Profiling: peak 440k nodes, 1GB overhead. Small buffer causes frequent blocking.
+	nodeChanSize = 500000
```
masih, line 24: We are starting 3 channels whit this size that do write ops. 

Have we considered the worst case memory footprint of this? What is the maximum size of each of the buffered items?

sei-db/sc/memiavl/snapshot.go
```diff
@@ -655,26 +1000,27 @@ func (snapshot *Snapshot) prefetchSnapshot(snapshotDir string, prefetchThreshold
...
 		"evm":  true,
 		"bank": true,
 		"acc":  true,
+		"wasm": true, // Added: 27M nodes, worth prefetching in cold start
 	}
 
 	return activeTrees[treeName]
 }
 
 func SequentialReadAndFillPageCache(filePath string) error {
```
masih: Clean filepath before opening it below.

sei-db/sc/memiavl/snapshot.go
```diff
@@ -685,9 +1031,17 @@ func SequentialReadAndFillPageCache(filePath string) error {
...
 		return err
 	}
+
+	totalSize := fileInfo.Size()
+	if totalSize == 0 {
+		fmt.Printf("[PREFETCH] Skipping empty file: %s\n", filePath)
+		return nil
+	}
+
+	fmt.Printf("[PREFETCH] Starting to prefetch file: %s (size: %d MB)\n", filePath, totalSize/(1024*1024))
```
masih: * You might want to cast the size to float64 before division. Otherwise if tit is small enough we might just see 0 in the message which could be confusing.

* Is this a debug print line left over? If not, sue logger instead of `fmt.Printf` into standard output?
```suggestion
	fmt.Printf("[PREFETCH] Starting to prefetch file: %s (size: %d MiB)\n", filePath, totalSize/(1024*1024))
```

sei-db/sc/memiavl/snapshot.go
```diff
@@ -685,9 +1031,17 @@ func SequentialReadAndFillPageCache(filePath string) error {
 	if err != nil {
 		return err
 	}
+
+	totalSize := fileInfo.Size()
+	if totalSize == 0 {
+		fmt.Printf("[PREFETCH] Skipping empty file: %s\n", filePath)
```
masih: Ditto re logging vis stdout prints.

sei-db/sc/memiavl/snapshot_catchup_test.go
```diff
@@ -0,0 +1,208 @@
...
+				},
+			},
+		}
+		require.NoError(t, db.ApplyChangeSets(cs))
+		_, err := db.Commit()
+		require.NoError(t, err)
+	}
+
+	// Wait for any background snapshot operations to complete before checking count
+	time.Sleep(500 * time.Millisecond)
```
masih: Please use require.Eventually instead. Ditto for the rest of the PR wherever `.Sleep` is used.

sei-db/sc/memiavl/snapshot_catchup_test.go
```diff
@@ -0,0 +1,208 @@
...
+	dir := t.TempDir()
+
+	db, err := OpenDB(logger.NewNopLogger(), 0, Options{
+		Dir:              dir,
+		CreateIfMissing:  true,
+		InitialStores:    []string{"test"},
+		SnapshotInterval: 100,
+	})
+	require.NoError(t, err)
+	defer db.Close()
```
masih: I recommend asserting that close did not return an error in tests. Ditto for the rest of the tests and other such defer statements in this PR.

sei-db/sc/memiavl/snapshot_catchup_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+	// Commit initial blocks
+	for i := 0; i < 200; i++ {
```
masih: Use for range syntax. Ditto for the rest of the PR.

masih (commented):
Thanks for all the work here @blindchaser 🚀 

Looking forward to getting this bad boy shipped and observing the benefits

I have left a bunch of suggestions and questions that's worth taking a look before this PR lands. Please let me know if i can clarify anything. 🍻

sei-db/sc/memiavl/snapshot_catchup_test.go
blindchaser: addressed

sei-db/sc/memiavl/snapshot_catchup_test.go
blindchaser: good point, updated to:

defer func() { require.NoError(t, db.Close()) }()

sei-db/config/toml.go
blindchaser: Our snapshot creation takes 3+ hours, so 1-hour minimum is safe for default config. It would only constrain users with custom lower SnapshotInterval values. Added a comment to document this behavior.

yzang2019: 1 hour is pretty safe, 3600s will produce up to 10000 blocks, which is roughly equal to our default interval.

sei-db/sc/memiavl/import.go
blindchaser: good point. I've added a time-based check (max 5 seconds) in addition to the node count check to guarantee responsiveness even when import speed degrades.

sei-db/config/toml.go
masih: Thank you both, marking as resolved.

sei-db/sc/memiavl/db.go
blindchaser: resolved

sei-db/sc/memiavl/db.go
blindchaser: yeah it's possible. seems edge case and won't impact that much.

i will put a comments if we want to follow up on this as it need to change db struct may beyond this pr scope

sei-db/sc/memiavl/db.go
blindchaser: i believe we don't need to copy all snapshot-related fields
The background goroutine doesn't make decisions about when to snapshot or when to prune. only write snapshot

sei-db/sc/memiavl/multitree.go
blindchaser: it looks the logger.Logger doesn't have warn interface. also this is more like a advise i believe info is more appropriate. 

double checked and keeping it at 300 makes more sense. That number comes from our tests on a 256 GB RAM machine (which I believe most of our providers use)

sei-db/sc/memiavl/import.go
blindchaser: right for this part, about 1GB is the total memory increase.

- Channel buffers: 120 MiB
- bufio.Writer buffers: 384 MiB
- In-flight data + overhead: < 500 MiB

sei-db/sc/memiavl/db.go
blindchaser: sure move this converting to store.go

PR #2539, opened 2025-11-07
PeerManager rewrite

pompon0:
Rewrite of the PeerManager logic to make it simple, predictable and race-condition resistant. There are further improvements to be done, but need to keep compatibility with the current p2p implementation. With this pr:
* the scoring system will be scraped entirely, in favor of separately managing persistent/unconditional/blocksync peers (trusted), and peers learned from the network (not trusted)
* peerstore will change its purpose: it used to be used to just store known peer addresses. Now it will be used to store addresses we successfully connected to. The only goal of peerstore now will be to make a node able to reestablish its recent connections on restart
* node will only broadcast addresses of nodes it has successfully connected to (so that it does not broadcast unverified data any more)
* the number of addresses per peer that peermanager keeps will be bounded (it is unbounded currently)
* PeerManager will be an implementation detail of the Router.


sei-tendermint/config/config.go
```diff
@@ -646,7 +646,7 @@ type P2PConfig struct { //nolint: maligned
 	// Comma separated list of nodes for block sync only
 	BlockSyncPeers string `mapstructure:"blocksync-peers"`
 
-	// UPNP port forwarding
+	// UPNP port forwarding. UNUSED
```
udpatil, line 649: since this is UNUSED, do we want to do something about it (eg. remove it) or will that affect existing network compatibility? if we want to do so down the road, lets add a TODO here?

sei-tendermint/internal/p2p/peermanager.go
```diff
@@ -4,1621 +4,527 @@ import (
...
+	if peerID == i.selfID {
+		return errors.New("connection to self")
+	}
+	conns := i.conns.Load()
+	if old, ok := conns.Get(peerID); ok {
+		// * allow to override connections in the same direction.
+		// * allow inbound conn to override outbound iff peerID > selfID.
+		//   This resolves the situation when peers try to connect to each other
+		//   at the same time.
+		oldDir := old.Info().DialAddr.IsPresent()
```
udpatil: shouldn't this be `oldAddr` or something instead of `oldDir`? (same for `newDir`)

sei-tendermint/internal/p2p/peermanager.go
```diff
@@ -4,1621 +4,527 @@ import (
...
-	}
-	for id := range configure {
-		if peer, ok := m.store.Get(id); ok {
-			if err := m.store.Set(m.configurePeer(peer)); err != nil {
-				return err
+	pa, ok := i.addrs[id]
+	// Add new peerAddrs if missing.
+	if !ok {
+		// Prune some peer if maxPeers limit has been reached.
+		if len(i.addrs) == i.options.maxPeers() {
```
udpatil, line 86: how do we intend this to behave if we have the len(i.addrs) == maxPeers (eg. 128) but none of the peers have failed. In this case, we would just drop this peer, just want to make sure this is the intended behavior?

sei-tendermint/libs/utils/mutex.go
```diff
@@ -83,6 +83,24 @@ type atomicWatch[T any] struct {
...
+
+func (w *AtomicSend[T]) Subscribe() AtomicRecv[T] {
+	return AtomicRecv[T]{&w.atomicWatch}
+}
+
+// NewAtomicWatch creates a new AtomicWatch with the given initial value.
+func NewAtomicSend[T any](value T) (w AtomicSend[T]) {
+	w.ptr.Store(newVersion(value))
+	// nolint:nakedret
+	return
```
udpatil, line 96: why do we want the naked return and the nolint?

sei-tendermint/internal/p2p/pex/reactor.go
```diff
@@ -83,345 +77,141 @@ func ChannelDescriptor() conn.ChannelDescriptor {
...
+	r.SpawnCritical("Run", func(ctx context.Context) error { return r.Run(ctx) })
 	return nil
 }
 
-// OnStop stops the reactor by signaling to all spawned goroutines to exit and
-// blocking until they all exit.
+// OnStop implements service.Service.
 func (r *Reactor) OnStop() {}
 
+func (r *Reactor) sendRoutine(ctx context.Context) error {
```
udpatil, line 127: Do we have any concerns on whether this will cause the network adapter to be utilized/blocked as we broadcast to a batch of peers (eg. 100). Is it possible here that every 10 seconds, we cause a slowdown to other network communications (eg. consensus voting - maybe this specific message type isnt as problematic bc priority) or should it be fine?

sei-tendermint/internal/p2p/peermanager.go
pompon0: yes, we just drop it. The rough idea here is that we want to dial at least once every address we accept. This doesn't give us any nice network properties, it just makes the semantics predictable (bounds the churn). I would like to move to some more useful algorithm soon (once I have time to implement it).

sei-tendermint/libs/utils/mutex.go
pompon0: AtomicSend is (intentionally) a no-copy object, because it embeds atomic.Pointer. Naked return is the canonical way of initializing no-copy objects (`return w` would be already a copy of `w`).

sei-tendermint/internal/p2p/pex/reactor.go
pompon0: There is a couple of aspects to that:
* sendRoutine just puts msgs into peers' outbound queues, which is non-blocking and doesn't do any IO.
* pex msgs have lower priority than consensus messages, so they will be deprioritized anyway
* pex requests are very small so it actually doesn't matter at all
* pex responses are desynchronized due to different roundtrip latency to different peers.
* eventually I would like to get rid of pex requests altogether (make peers proactively push pex responses) and make pex response sizes negligible (limit the size, perhaps sending just diffs).

sei-tendermint/internal/p2p/peermanager.go
pompon0: Dir is an abbreviation of "direction" (see comment above). I'll expand the names to old/newDirection to make it more clear.

sei-tendermint/config/config.go
pompon0: just removing it will break config parsing afaiu. I don't have any long term deprecation strategy here. Just wanted to document the fact for clarity.

PR #2542, opened 2025-11-10
config: make worker pool configurable and increase default queue size

blindchaser:
- Adds worker_pool_size and worker_queue_size config options in the app.toml [evm] section
- Increases default queue from 200 to 1000

Add configurable worker_pool_size and worker_queue_size to app.toml

### DIFF
--- a/evmrpc/config.go
+++ b/evmrpc/config.go
@@ -9,16 +9,6 @@ import (
 	"github.com/spf13/cast"
 )
 
-// calculateDefaultWorkerPoolSize returns the default worker pool size.
-// Capped at 64 to prevent excessive goroutines on high-core machines.
-func calculateDefaultWorkerPoolSize() int {
-	workers := runtime.NumCPU() * 2
-	if workers > 64 {
-		return 64
-	}
-	return workers
-}
-
 // EVMRPC Config defines configurations for EVM RPC server on this node
 type Config struct {
 	// controls whether an HTTP EVM server is enabled
@@ -115,11 +105,11 @@ type Config struct {
 	RPCStatsInterval time.Duration `mapstructure:"rpc_stats_interval"`
 
 	// WorkerPoolSize defines the number of workers in the worker pool.
-	// Set to 0 to use default (runtime.NumCPU() * 2)
+	// Set to 0 to use default: min(64, runtime.NumCPU() * 2)
 	WorkerPoolSize int `mapstructure:"worker_pool_size"`
 
 	// WorkerQueueSize defines the size of the task queue in the worker pool.
-	// Set to 0 to use default (1000)
+	// Set to 0 to use default: 1000
 	WorkerQueueSize int `mapstructure:"worker_queue_size"`
 }
 
@@ -150,8 +140,8 @@ var DefaultConfig = Config{
 	MaxTraceLookbackBlocks:       10000,
 	TraceTimeout:                 30 * time.Second,
 	RPCStatsInterval:             10 * time.Second,
-	WorkerPoolSize:               calculateDefaultWorkerPoolSize(), // Default: min(CPU cores × 2, 64)
-	WorkerQueueSize:              DefaultWorkerQueueSize,           // Default: 1000 tasks
+	WorkerPoolSize:               min(MaxWorkerPoolSize, runtime.NumCPU()*2), // Default: min(64, CPU cores × 2)
+	WorkerQueueSize:              DefaultWorkerQueueSize,                     // Default: 1000 tasks
 }
 
 const (
@@ -420,11 +410,11 @@ max_trace_lookback_blocks = {{ .EVM.MaxTraceLookbackBlocks }}
 trace_timeout = "{{ .EVM.TraceTimeout }}"
 
 # WorkerPoolSize defines the number of workers in the worker pool.
-# Default is min(CPU cores × 2, 64). Capped at 64 for high-core machines.
-# Set to 0 to auto-detect based on CPU cores.
+# Default: min(64, CPU cores × 2). Capped at 64 to prevent excessive goroutines on high-core machines.
+# Set to 0 to use the default.
 worker_pool_size = {{ .EVM.WorkerPoolSize }}
 
 # WorkerQueueSize defines the size of the task queue in the worker pool.
-# Default is 1000 tasks. Set to 0 to use the default value.
+# Default: 1000 tasks. Set to 0 to use the default.
 worker_queue_size = {{ .EVM.WorkerQueueSize }}
 `
--- a/evmrpc/config_integration_test.go
+++ b/evmrpc/config_integration_test.go
@@ -160,14 +160,14 @@ func testWorkerPoolUnderLoad(t *testing.T, wp *evmrpc.WorkerPool, taskCount int)
 	var completed atomic.Int32
 
 	// Block the workers with slow tasks
-	for i := 0; i < wp.WorkerCount(); i++ {
+	for range wp.WorkerCount() {
 		wp.Submit(func() {
 			time.Sleep(50 * time.Millisecond)
 		})
 	}
 
 	// Try to submit many tasks quickly
-	for i := 0; i < taskCount; i++ {
+	for range taskCount {
 		err := wp.Submit(func() {
 			time.Sleep(10 * time.Millisecond)
 			completed.Add(1)
@@ -213,7 +213,7 @@ simulation_gas_limit = 10000000
 		cfg, err := evmrpc.ReadConfig(v)
 		require.NoError(t, err)
 
-		// Old config should use default calculated values since calculateDefaultWorkerPoolSize() is called
+		// Old config should use default calculated values: min(MaxWorkerPoolSize, runtime.NumCPU() * 2)
 		require.Greater(t, cfg.WorkerPoolSize, 0, "Should have default worker pool size")
 		require.Equal(t, 1000, cfg.WorkerQueueSize, "Should use default queue size")
 
@@ -326,7 +326,7 @@ func TestWorkerPoolPerformanceWithConfig(t *testing.T) {
 			failed := 0
 			var wg sync.WaitGroup
 
-			for i := 0; i < tc.taskCount; i++ {
+			for range tc.taskCount {
 				wg.Add(1)
 				err := wp.Submit(func() {
 					defer wg.Done()
--- a/evmrpc/worker_pool.go
+++ b/evmrpc/worker_pool.go
@@ -59,7 +59,7 @@ func InitGlobalWorkerPool(workerPoolSize, workerQueueSize int) {
 
 // GetGlobalWorkerPool returns the singleton worker pool instance.
 // If not initialized, it creates one with default values:
-// - Worker count: runtime.NumCPU() * 2
+// - Worker count: min(MaxWorkerPoolSize, runtime.NumCPU() * 2)
 // - Queue size: DefaultWorkerQueueSize
 func GetGlobalWorkerPool() *WorkerPool {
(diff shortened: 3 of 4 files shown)


evmrpc/config.go
```diff
@@ -132,6 +140,8 @@ var DefaultConfig = Config{
 	MaxTraceLookbackBlocks:       10000,
 	TraceTimeout:                 30 * time.Second,
 	RPCStatsInterval:             10 * time.Second,
+	WorkerPoolSize:               0, // 0 = use default (runtime.NumCPU() * 2)
```
yzang2019: Why not setting to runtime.NumCPU() * 2 directly?

blindchaser: this might be a little confusing, will update it with default value

it sets to default value when invalide <=0 value pass in.

evmrpc/config.go
```diff
@@ -132,6 +140,8 @@ var DefaultConfig = Config{
 	MaxTraceLookbackBlocks:       10000,
 	TraceTimeout:                 30 * time.Second,
 	RPCStatsInterval:             10 * time.Second,
+	WorkerPoolSize:               runtime.NumCPU() * 2,   // Default: CPU cores × 2
```
masih: This might be too large for machines with high number of cores?
* Do we want to cap it to some max number?
* Are the tasks executed in the worker pool generally IO bound or CPU bound? If latter, 2X CPU cores might be too high.

evmrpc/worker_pool.go
```diff
@@ -7,12 +7,15 @@ import (
...
 	WorkerBatchSize = 100
-	WorkerQueueSize = 200
-)
 
-var (
-	MaxNumOfWorkers = runtime.NumCPU() * 2 // each worker will handle a batch of WorkerBatchSize blocks
+	// DefaultWorkerQueueSize is the default size of the task queue.
+	// This represents the number of tasks (not blocks) that can be queued.
+	// Total capacity = DefaultWorkerQueueSize * WorkerBatchSize blocks
+	// Example: 1000 tasks * 100 blocks/task = 100,000 blocks can be buffered
```
masih, line 17: Would be super useful to also comment on the worst-case memory footprint.

evmrpc/worker_pool_test.go
```diff
@@ -361,3 +361,237 @@ func TestWorkerPoolOrderingNotGuaranteed(t *testing.T) {
...
+			var executed bool
+			var wg sync.WaitGroup
+			wg.Add(1)
+
+			err := wp.Submit(func() {
+				defer wg.Done()
+				executed = true
+			})
+
+			if err != nil {
```
masih, line 401: Use testify? The code base depends on it already. Update AI context to never have to tell it again.
Added benefit apart from consistency is reduction in LoC.

evmrpc/worker_pool_test.go
```diff
@@ -361,3 +361,237 @@ func TestWorkerPoolOrderingNotGuaranteed(t *testing.T) {
...
+	wp.Start()
+	defer wp.Close()
+
+	var completed int64
+	var wg sync.WaitGroup
+	totalTasks := 500
+
+	// Submit many tasks quickly
+	start := time.Now()
+	for i := 0; i < totalTasks; i++ {
```
masih:
```suggestion
	for i := range totalTasks {
```

Ditto in other places.

evmrpc/worker_pool_test.go
```diff
@@ -361,3 +361,237 @@ func TestWorkerPoolOrderingNotGuaranteed(t *testing.T) {
...
+	}
+
+	wg.Wait()
+	elapsed := time.Since(start)
+
+	if atomic.LoadInt64(&completed) != int64(totalTasks) {
+		t.Errorf("Expected %d tasks completed, got %d", totalTasks, completed)
+	}
+
+	t.Logf("Completed %d tasks in %v with 4 workers", totalTasks, elapsed)
```
masih: I am curious, is logging elapsed time or measuring it significant/needed?

evmrpc/worker_pool_test.go
```diff
@@ -361,3 +361,237 @@ func TestWorkerPoolOrderingNotGuaranteed(t *testing.T) {
...
+
+			if !executed {
+				t.Error("Task was not executed")
+			}
+		})
+	}
+}
+
+// Test that InitGlobalWorkerPool is idempotent
+func TestInitGlobalWorkerPoolIdempotent(t *testing.T) {
```
masih, line 415: Why is this empty test here at all?

Generally, use t.Skip for temporarily or conditionally skip a test that has some implementation. This one only has comments?

evmrpc/worker_pool_test.go
```diff
@@ -361,3 +361,237 @@ func TestWorkerPoolOrderingNotGuaranteed(t *testing.T) {
...
+
+			if tt.wantError && errors == 0 {
+				t.Error("Expected some errors due to small queue, got none")
+			}
+
+			if !tt.wantError && errors > 0 {
+				t.Errorf("Expected no errors with queue size %d, got %d errors", tt.queueSize, errors)
+			}
+
+			t.Logf("Queue size %d: submitted=%d, errors=%d", tt.queueSize, submitted, errors)
```
masih: I see a lot of logs in the tests. I recommend using these rarely as it can make the test output very noisy. FWIW, i personally wouldn't use logging for this specific case for example.

evmrpc/worker_pool_test.go
```diff
@@ -361,3 +361,237 @@ func TestWorkerPoolOrderingNotGuaranteed(t *testing.T) {
...
+
+			// Block the worker
+			wp.Submit(func() {
+				time.Sleep(100 * time.Millisecond)
+			})
+
+			// Try to fill the queue
+			errors := 0
+			submitted := 0
+			for i := 0; i < 20; i++ {
```
masih: How does this loop try to fill the queue? The size of the queue may be larger than 20 but it does 20 fixed iterations?

evmrpc/worker_pool_test.go
```diff
@@ -296,18 +284,30 @@ func TestWorkerPoolConstants(t *testing.T) {
...
+	totalCapacity := DefaultWorkerQueueSize * WorkerBatchSize
+	if totalCapacity != 100000 {
+		t.Errorf("Total capacity = %d, want 100000 (1000 tasks × 100 blocks/task)", totalCapacity)
+	}
+
+	t.Logf("All constants validated:")
+	t.Logf("  - WorkerBatchSize: %d blocks/task", WorkerBatchSize)
+	t.Logf("  - DefaultWorkerQueueSize: %d tasks", DefaultWorkerQueueSize)
+	t.Logf("  - Total capacity: %d blocks", totalCapacity)
+	t.Logf("  - Default workers: CPU × 2")
```
masih: too verbose I think.

evmrpc/config_test.go
```diff
@@ -245,4 +255,99 @@ func TestReadConfig(t *testing.T) {
...
+				maxSubscriptionsNewHead:      10000,
+				enableTestAPI:                false,
+				maxConcurrentTraceCalls:      uint64(10),
+				maxConcurrentSimulationCalls: uint64(10),
+				maxTraceLookbackBlocks:       int64(100),
+				traceTimeout:                 30 * time.Second,
+				rpcStatsInterval:             10 * time.Second,
+				workerPoolSize:               tt.workerPoolSize,
+				workerQueueSize:              tt.workerQueueSize,
+			}
```
masih: Do we need to set every field here just to test worker pool sizes are being set?

evmrpc/config_integration_test.go
```diff
@@ -0,0 +1,493 @@
...
+	}
+
+	// Wait for tasks to complete
+	time.Sleep(200 * time.Millisecond)
+
+	return failures
+}
+
+// TestConfigFileFormats tests various TOML format variations
+func TestConfigFileFormats(t *testing.T) {
```
masih: This test essentially testing TOML parsing and is independent of worker pool size settings, right?

I recommend doing this kind of test in their own PR _if needed_. If we are delegating all the TOML parsing to an external library (which I think we do, do we really want to test that library is parsing TOML OK in sei-chain? :)

masih (commented):
Thanks for taking the time to write all the tests @blindchaser 🚀 

I left a few questions and suggestions. There is one potential blocker in the logic of one of the tests regarding queue size, otherwise LGTM.

evmrpc/config.go
blindchaser: good point, updated

yzang2019: Yeah agreed, I think it's generally CPU bound, we can set a max and maybe default to numCPU ?

evmrpc/worker_pool_test.go
blindchaser: addressed

evmrpc/config.go
```diff
@@ -9,6 +9,16 @@ import (
...
 )
 
+// calculateDefaultWorkerPoolSize returns the default worker pool size.
+// Capped at 64 to prevent excessive goroutines on high-core machines.
+func calculateDefaultWorkerPoolSize() int {
+	workers := runtime.NumCPU() * 2
+	if workers > 64 {
+		return 64
+	}
+	return workers
```
masih: You can write this entire body as:
```go
min(64, runtime.NumCPU() * 2)
```

..at which point, i would recommend against refactoring it into its own function and moving the godoc to be an inline comment that is also reflected in config documentation.

evmrpc/config_integration_test.go
```diff
@@ -0,0 +1,352 @@
...
+
+	// Block the workers with slow tasks
+	for i := 0; i < wp.WorkerCount(); i++ {
+		wp.Submit(func() {
+			time.Sleep(50 * time.Millisecond)
+		})
+	}
+
+	// Try to submit many tasks quickly
+	for i := 0; i < taskCount; i++ {
```
masih: Use for range syntax. Ditto for the loop above and the reminder of the code changes.

evmrpc/config.go
```diff
@@ -103,6 +113,14 @@ type Config struct {
 
 	// RPCStatsInterval for how often to report stats
 	RPCStatsInterval time.Duration `mapstructure:"rpc_stats_interval"`
+
+	// WorkerPoolSize defines the number of workers in the worker pool.
+	// Set to 0 to use default (runtime.NumCPU() * 2)
```
masih: At the very least we want to update the documentation to reflect the cap of 64. Ditto in other places.

evmrpc/worker_pool.go
```diff
@@ -7,12 +7,26 @@ import (
...
+	// - Queue channel overhead: ~8KB (1000 * 8 bytes per channel slot)
+	// - Each task closure: ~24 bytes
+	// - Total queue memory: ~32KB (negligible)
+	// Note: Actual memory usage depends on block data processed by workers
+	DefaultWorkerQueueSize = 1000
+
+	// MaxWorkerPoolSize caps the number of workers to prevent excessive
+	// goroutine creation on high-core machines. Tasks are primarily I/O bound
+	// (fetching and processing block logs), so 2x CPU cores can be excessive.
+	MaxWorkerPoolSize = 64
```
masih, line 29: Sad that we are repeating ourselves here regarding what the max is. Can we avoid this repetition?

evmrpc/config_integration_test.go
blindchaser: addressed

evmrpc/config.go
blindchaser: addressed

evmrpc/worker_pool.go
blindchaser: addressed

PR #2543, opened 2025-11-11
Add a benchmark mode

stevenlanders:
- run `/scripts/benchmark.sh`, and it'll auto-produce 1000-tx blocks of evm transfers and measure tps

- local testing 
- unit tests


app/benchmark.go
```diff
@@ -0,0 +1,278 @@
...
+	go func() {
+		defer close(ch)
+		var height int64
+		for {
+			// bail on ctx err
+			if ctx.Err() != nil {
+				return
+			}
+			// generate txs like: txs := gen.GenerateN(1000)
+			loadTxs := gen.GenerateN(1000)
```
udpatil, line 183: As a side note: wouldn't it make more sense to pull sei-load into sei-chain monorepo if we're going to be using it in this way within monorepo as opposed to a separately maintained codebase?

scripts/benchmark.sh
```diff
@@ -0,0 +1,112 @@
+#!/bin/bash
+# require success for commands
+set -e
+
+# Parse command line arguments
+MOCK_BALANCES=${MOCK_BALANCES:-true}
```
udpatil, line 6: is there even a case in which we would NOT want mock balances for benchmark?

scripts/benchmark.sh
```diff
@@ -0,0 +1,112 @@
...
+jq '.validators[0] += {"pub_key":'$KEY'}' ~/.sei/config/tmp_genesis_2.json > ~/.sei/config/tmp_genesis_3.json
+mv ~/.sei/config/tmp_genesis_3.json ~/.sei/config/genesis.json && rm ~/.sei/config/tmp_genesis.json && rm ~/.sei/config/tmp_genesis_2.json
+
+echo "Creating Accounts"
+# create 10 test accounts + fund them
+python3  loadtest/scripts/populate_genesis_accounts.py 20 loc
+
+~/go/bin/seid collect-gentxs
+# update some params in genesis file for easier use of the chain localls (make gov props faster)
+cat ~/.sei/config/genesis.json | jq '.app_state["gov"]["deposit_params"]["max_deposit_period"]="60s"' > ~/.sei/config/tmp_genesis.json && mv ~/.sei/config/tmp_genesis.json ~/.sei/config/genesis.json
```
udpatil, line 44: masih has added in QoL improvements for both config and genesis modifications via seictl:
https://github.com/sei-protocol/seictl

we should probably use that instead for simplicity (eg. check for util installation at the beginning of script, and if not, recommend user to do that)

scripts/benchmark.sh
```diff
@@ -0,0 +1,112 @@
...
+cat ~/.sei/config/genesis.json | jq --arg start_date "$START_DATE" --arg end_date "$END_DATE_3DAYS" '.app_state["mint"]["params"]["token_release_schedule"]=[{"start_date": $start_date, "end_date": $end_date, "token_release_amount": "999999999999"}]' > ~/.sei/config/tmp_genesis.json && mv ~/.sei/config/tmp_genesis.json ~/.sei/config/genesis.json
+cat ~/.sei/config/genesis.json | jq --arg start_date "$END_DATE_3DAYS" --arg end_date "$END_DATE_5DAYS" '.app_state["mint"]["params"]["token_release_schedule"] += [{"start_date": $start_date, "end_date": $end_date, "token_release_amount": "999999999999"}]' > ~/.sei/config/tmp_genesis.json && mv ~/.sei/config/tmp_genesis.json ~/.sei/config/genesis.json
+
+if [ ! -z "$2" ]; then
+  APP_TOML_PATH="$2"
+else
+  APP_TOML_PATH="$HOME/.sei/config/app.toml"
+fi
+# Enable OCC and SeiDB
+sed -i.bak -e 's/# concurrency-workers = .*/concurrency-workers = 500/' $APP_TOML_PATH
```
udpatil, line 71: same seictl comment here

app/benchmark.go
stevenlanders: Yeah I think we should - I'm thinking we should do that in an isolated PR just so it's a noop change that's easier to review

scripts/benchmark.sh
stevenlanders: likely not for this type of testing

udpatil (approved):

PR #2550, opened 2025-11-13
Speedup sei-cosmos unit test

yzang2019:


sei-cosmos/store/rootmulti/store_test.go
```diff
@@ -609,6 +610,20 @@ func TestCacheWraps(t *testing.T) {
 }
 
 func TestTraceConcurrency(t *testing.T) {
+	// Check for goroutine leaks
+	initialGoroutines := runtime.NumGoroutine()
+	defer func() {
```
masih, line 615: If this helps the tests run faster in the immediate term, great. But I must point out that this kind of test would be inherently flaky I'm afraid.

PR #2555, opened 2025-11-17
Fix writes to non temp folder and increase free space

masih:
Fix tests that use OS tempdir for testing purposes, and change them to use testing tempdir. This way the files created by tests are guaranteed to be removed by Go testing itself.

Remove unnecessary build tooling from GitHub runner to open up space for files that cosmos tests seem to create.

### DIFF
--- a/sei-cosmos/x/bank/keeper/keeper_test.go
+++ b/sei-cosmos/x/bank/keeper/keeper_test.go
@@ -1943,7 +1943,7 @@ func (suite *IntegrationTestSuite) TestMintCoinRestrictions() {
 func (suite *IntegrationTestSuite) TestKeeperGetAllBalances() {
 	app, ctx := suite.app, suite.ctx
 	addr := sdk.AccAddress([]byte("addr1_______________"))
-	cnt := 1000000
+	cnt := 100_000
 	allDenoms := make(sdk.Coins, cnt)
 	for i := 0; i < cnt; i++ {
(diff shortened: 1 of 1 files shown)


sei-cosmos/testutil/ioutil.go
```diff
@@ -64,11 +67,13 @@ func WriteToNewTempFile(t testing.TB, s string) *os.File {
...
 }
 
+var tmpFileCounter atomic.Int32
+
 // TempFile returns a writable temporary file for the test to use.
 func TempFile(t testing.TB) *os.File {
 	t.Helper()
 
-	fp, err := ioutil.TempFile(t.TempDir(), "")
+	fp, err := os.Create(filepath.Join(t.TempDir(), fmt.Sprintf("tmpfile-%d", tmpFileCounter.Add(1))))
```
blindchaser, line 76: is it possible getting duplicate tmpFileCounter when enable --race?

and does `ioutil.TempFile(t.TempDir(), "")` already creatd under t.TempDir() with random suffix string?

blindchaser (approved):

sei-cosmos/testutil/ioutil.go
masih: Not possible i don't think regardless of race flag. (Ioutils is deprecated)

The better way to improve code here is to remove these test util functions and use t tempdir calls only. But that was a bigger refactoring hence the atomic.

PR #2558, opened 2025-11-18
refactor of consensus reactor task management

pompon0:
This is just a cleanup PR:
* replaced generic pubsub with just a bunch of callbacks
* made peer stats collection synchronous
* rewritten peer task management
* consensus state is not a service any more
* RetrySigner is not treated as special any more
* dropped some custom string formatting of structs.
* removed unused mocks

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -143,7 +143,7 @@ jobs:
           go-version: 1.24
 
       - name: Start 4 node docker cluster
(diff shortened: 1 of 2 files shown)


udpatil (approved):
LGTM, one thing to note is that several of these structs now no longer match those of existing consensus (eg. not representing height as an int64 json string) as well as the HRS definition. are the wire representations all still consistent such that we can test this version along side the current testnet/mainnet versions to verify proper behavior?

PR #2560, opened 2025-11-18
fix: resolve data race in parallel snapshot writing

blindchaser:
- fix: resolve data race between snapshot write and DB.Close()

### DIFF
--- a/sei-db/sc/memiavl/multitree.go
+++ b/sei-db/sc/memiavl/multitree.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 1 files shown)


udpatil (approved):

sei-db/sc/memiavl/multitree.go
```diff
@@ -458,9 +458,10 @@ func (t *MultiTree) writeSnapshotPriorityEVM(ctx context.Context, dir string, wp
 		group, _ := wp.GroupContext(ctx)
 
 		for _, entry := range otherTrees {
-			tree, name := entry.Tree, entry.Name
+			// Capture loop variables for goroutine closure to avoid data race
+			entry := entry // Create new variable for closure capture
```
pompon0, line 470: that should not be needed since go 1.22: https://go.dev/blog/loopvar-preview . Are you sure this is the issue?

blindchaser: intersting. i pushed another fix related to pond.GroupContext, this should be the one

pompon0: ack, I'm not familiar with pond. You can remove the "entry := entry" though, it is redundant either way.

blindchaser: thx for point it out!

PR #2561, opened 2025-11-19
Make seid rolback idempodent and remove --hard

yzang2019:
Summary of Changes
1. Get rid of --hard, which is always very confusing. The new behavior will always remove blocks, which is equal  to  setting --hard in the past. We don't really need the scenario for not deleting blocks.

2. Made rollback more idempotent
- Changed order: rollback app state first, then tendermint state
- Added validation at each step

3. Add a new flag "num-blocks"
- For convenience, we added a new flag for num-blocks to rollback multiple blocks in a single command

**Three Scenarios Handled**
1. Both at same height (normal rollback)
- Condition: appHeight == tmHeight
- Action: Rollback both app and tendermint state by N blocks
- Order: App state first, then tendermint state
- Blocks: Deleted (hard rollback)
2. App already rolled back but tendermint not 
- Condition: appHeight < tmHeight
- Action: Complete the tendermint rollback to match app height, ignore num-blocks
- Blocks: Deleted during each tendermint rollback
3. App ahead of tendermint
- Condition: appHeight > tmHeight
- Action: Rollback app state only to match tendermint height, ignore num-blocks
- Behavior: Only app state is modified; tendermint state remains unchanged
- Use case: Recovers from inconsistent state where app is ahead

Tested on RPC node for rollback as well

### DIFF
--- a/sei-cosmos/server/rollback_test.go
+++ b/sei-cosmos/server/rollback_test.go
@@ -22,6 +22,7 @@ import (
 	"github.com/google/orderedcode"
 	"github.com/stretchr/testify/require"
 	abci "github.com/tendermint/tendermint/abci/types"
+	tmcmd "github.com/tendermint/tendermint/cmd/tendermint/commands"
 	tmconfig "github.com/tendermint/tendermint/config"
 	"github.com/tendermint/tendermint/crypto"
 	"github.com/tendermint/tendermint/crypto/ed25519"
@@ -464,10 +465,10 @@ func TestRollbackScenario1_BothAtSameHeight(t *testing.T) {
 	appCommit := app.CommitMultiStore().LastCommitID()
 	require.Equal(t, targetHeight, appCommit.GetVersion())
 
-	// Verify tendermint state was rolled back
-	tmHeight, err := getTendermintStateHeight(cfg)
+	// Verify tendermint state was rolled back to match app
+	tmState, err := tmcmd.LoadTendermintState(cfg)
 	require.NoError(t, err)
-	require.Equal(t, targetHeight, tmHeight)
+	require.Equal(t, targetHeight, tmState.LastBlockHeight)
 }
 
 // TestRollbackScenario2_AppAheadOfTendermint tests scenario 2: app already rolled back, tendermint not
@@ -508,9 +509,9 @@ func TestRollbackScenario2_AppAheadOfTendermint(t *testing.T) {
 	require.Equal(t, targetHeight, appCommit.GetVersion())
 
 	// Verify tendermint state was rolled back to match app
-	finalTmHeight, err := getTendermintStateHeight(cfg)
+	tmState, err := tmcmd.LoadTendermintState(cfg)
 	require.NoError(t, err)
-	require.Equal(t, targetHeight, finalTmHeight)
+	require.Equal(t, targetHeight, tmState.LastBlockHeight)
 }
 
 // TestRollbackScenario3_TendermintAheadOfApp tests scenario 3: app ahead of tendermint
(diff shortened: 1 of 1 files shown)


sei-tendermint/cmd/tendermint/commands/rollback.go
```diff
@@ -42,6 +42,26 @@ application.
 	return cmd
 }
 
+// LoadTendermintState loads the tendermint state from the database.
+// Returns the state, or an error if loading fails.
+func LoadTendermintState(config *config.Config) (state.State, error) {
+	blockStore, stateStore, err := loadStateAndBlockStore(config)
```
masih, line 48: It is unfortunate that `loadStateAndBlockStore` opens both block and state store. Here we just need state store.

In the future maybe we should refactor this to not unnecessarily open the DB.

sei-cosmos/server/rollback.go
```diff
@@ -7,9 +7,45 @@ import (
 	"github.com/cosmos/cosmos-sdk/server/types"
 	"github.com/spf13/cobra"
 	tmcmd "github.com/tendermint/tendermint/cmd/tendermint/commands"
+	tmcfg "github.com/tendermint/tendermint/config"
 )
 
-var removeBlock = false
+// getTendermintStateHeight gets the current tendermint state height.
+func getTendermintStateHeight(cfg *tmcfg.Config) (int64, error) {
+	tmState, err := tmcmd.LoadTendermintState(cfg)
```
masih: Double checking that this _never_ gets called if the state/block stores are already open by another process?

Depending on the underlying persistent storage they may error out themselves if concurrent attempt is made to open the db. But best to check this to avoid potential corruption.

Ditto for rollback.

sei-cosmos/server/rollback.go
```diff
@@ -41,44 +76,90 @@ application. If you wanna rollback multiple blocks, please add --hard option.
...
+					return err
+				}
+
+				// Verify consistency
+				if app.CommitMultiStore().LastCommitID().Version != newTmHeight {
+					panic("Application state height does not match the tendermint state height")
+				}
+
+				if string(appHash) != string(tmHash) {
+					fmt.Printf("WARNING: Application state hash (%X) does not match the tendermint state hash (%X)\n",
```
masih: What would be the actual course of action for the node operator in this case?

Is "warning" a strong enough signal for this case?

Ditto for the other hash mismatch.

sei-cosmos/server/rollback.go
```diff
@@ -41,44 +76,90 @@ application. If you wanna rollback multiple blocks, please add --hard option.
...
-			if err := app.CommitMultiStore().RollbackToVersion(tmHeight); err != nil {
-				return fmt.Errorf("failed to rollback to version: %w", err)
+				return err
 			}
+			fmt.Printf("Initial Tendermint state height=%d\n", tmHeight)
 
-			lastCommit = app.CommitMultiStore().LastCommitID()
-			fmt.Printf("Rolled back app state to height %d and hash %X\n", lastCommit.GetVersion(), lastCommit.GetHash())
+			// Handle different scenarios based on height comparison
+			if appHeight == tmHeight {
```
masih, line 82: The question for me remains which is: why should these ever be out of sync? Can we engineer it such that the implementation guarantees they get out of sync to begin with, e.g. via a two-stage commit mechanism across the two db.

sei-tendermint/cmd/tendermint/commands/rollback.go
yzang2019: Yeah, that's a bit of extra work and annoying, agree we need a refactor in the future

sei-cosmos/server/rollback.go
yzang2019: Unless they do actually share the same underlying DB and batch commit, otherwise I can't think of any easy way to make it really atomic here. 

The reason why it can get out of sync is because of partial failures in either of the rollback. For example, app rolled back successfully, but tm not, then it will be out of sync, and we don't have a mechanism to roll forward (revert) here in case of partial failures.

For now, I choose to fix this in a straightforward and low risk way, trying to minimize the code we touch by adjusting the rollback ordering to make it idempodent as much as possible upon failures and retries

sei-cosmos/server/rollback.go
yzang2019: Usually this could indicate the operator to rollback more blocks, previously it was a panic, which is bad because that panic would make operator think the rollback actually failed. Here I changed panic to be a warning message, but that's a good point, let me add some more info for what this indicates to the log.

sei-cosmos/server/rollback.go
yzang2019: Yeah, this should never be called when the process is running or by another process. And I think it will actually fail due to resource unavailable error

PR #2569, opened 2025-11-24
Refactor DeliverTx ante

codchen:
Refactor the ante handler logics for DeliverTx into distinct flows. No check is supposed to be added or removed in this PR, but the ordering of checks have changed, which would result in different error codes for transactions that are invalid in multiple ways. As a result, this change is consensus breaking.

The next step is to simplify the actual execution path of DeliverTx

refactor; existing tests

### DIFF
--- a/app/legacyabci/deliver_tx.go
+++ b/app/legacyabci/deliver_tx.go
@@ -93,7 +93,9 @@ func DeliverTx(
 	} else if isEVM {
 		newCtx, err = ante.EvmDeliverTxAnte(anteCtx, txConfig, tx, keepers.UpgradeKeeper, keepers.EvmKeeper)
 		defer func() {
-			newCtx.DeliverTxCallback()(ctx.WithGasMeter(sdk.NewInfiniteGasMeterWithMultiplier(ctx)))
+			if newCtx.DeliverTxCallback() != nil {
+				newCtx.DeliverTxCallback()(ctx.WithGasMeter(sdk.NewInfiniteGasMeterWithMultiplier(ctx)))
+			}
 		}()
 	} else {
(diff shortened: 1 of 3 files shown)


sei-cosmos/baseapp/baseapp.go
```diff
@@ -1012,17 +1012,17 @@ func (app *BaseApp) runTx(ctx sdk.Context, mode runTxMode, tx sdk.Tx, checksum [
...
 }
 
-// runMsgs iterates through a list of messages and executes them with the provided
+// RunMsgs iterates through a list of messages and executes them with the provided
 // Context and execution mode. Messages will only be executed during simulation
 // and DeliverTx. An error is returned if any single message fails or if a
 // Handler does not exist for a given message route. Otherwise, a reference to a
 // Result is returned. The caller must not commit state if an error is returned.
-func (app *BaseApp) runMsgs(ctx sdk.Context, msgs []sdk.Msg, mode runTxMode) (*sdk.Result, error) {
+func (app *BaseApp) RunMsgs(ctx sdk.Context, msgs []sdk.Msg) (*sdk.Result, error) {
```
udpatil, line 1024: Having the ability to bundle messages into one tx from the cosmos side seems like unnecessary complexity given that EVM is one tx == one msg. Is there any existing use case for multiple messages we care to continue supporting, and/or is it fairly straightforward to only allow one msg per tx (eg. adding an antehandler check)? Maybe this is something we don't necessarily need to do in this PR but would be nice to further sumplify execution path and eventually have cosmos features removed.

udpatil (approved):

sei-cosmos/baseapp/baseapp.go
codchen: for EVM txs that is the case. for cosmos txs I'm not sure. We can add a metrics that count num of txs that have >1 messages

PR #2587, opened 2025-12-09
Flatten `sei-db` module

masih:
Remove `sei-db` as a nested go module and instead flatten it into `sei-chain` top level module.

This simplifies the go module structure and removes one degree of complexity from workspace management.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -10,7 +10,6 @@ coverage:
         target: 40%
         paths:
           - "!sei-cosmos/**"
-          - "!sei-db/**"
           - "!sei-ibc-go/**"
           - "!sei-tendermint/**"
           - "!sei-wasmd/**"
@@ -35,12 +34,6 @@ coverage:
           - "sei-wasmd/**"
         flags:
           - sei-wasmd
-      sei-db:
-        target: 30%
-        paths:
-          - "sei-db/**"
-        flags:
-          - sei-db
       sei-ibc-go:
         target: 40%
         paths:


go.work.sum
```diff
@@ -771,9 +771,8 @@ github.com/gonum/floats v0.0.0-20181209220543-c233463c7e82 h1:EvokxLQsaaQjcWVWSV
 github.com/gonum/internal v0.0.0-20181124074243-f884aa714029 h1:8jtTdc+Nfj9AR+0soOeia9UZSvYBvETVHZrugUowJ7M=
 github.com/gonum/lapack v0.0.0-20181123203213-e4cdc5a0bff9 h1:7qnwS9+oeSiOIsiUMajT+0R7HR6hw5NegnKPmn/94oI=
 github.com/gonum/matrix v0.0.0-20181209220409-c518dec07be9 h1:V2IgdyerlBa/MxaEFRbV5juy/C3MGdj4ePi+g6ePIp4=
-github.com/google/btree v1.1.3/go.mod h1:qOPhT0dTNdNzV6Z/lhRX0YXUafgPLFUh+gZMl761Gm4=
+github.com/google/btree v1.0.1/go.mod h1:xXMiIv4Fb/0kKde4SpL7qlzvu5cMJDRkFDxJfI9uaxA=
```
pompon0, line 774: fyi, btree gets reverted to an older version

pompon0 (approved):

go.work.sum
masih: Yeah this also caught my eye thanks for pointing that out. 

This is a transitive dependency. Taking a closer look to see why this happened.

masih: After a closer look, this is in go wor sum, and a result of running `go work sync`. The indirect dependency in `sei-chain` [remains 1.1.3](https://github.com/sei-protocol/sei-chain/blob/9811e0bdb4a4cb4d439b6bb0d64a03dc3b094847/go.mod#L183) which is the one ends up being used in the built binary.

PR #2591, opened 2025-12-10
perf: improve eth_getLogs performance with early rejection and backpressure

blindchaser:
- add early rejection for pruned blocks to avoid wasting resources
- align DB semaphore size with worker pool size
- add backpressure mechanism to reject requests when system is overloaded
- add comprehensive worker pool metrics (Prometheus + optional stdout debug)
- add DB semaphore tracking for I/O monitoring
- add EVM_DEBUG_METRICS env var to enable debug output to stdout(nothing print out by default)

- significant perf improvement on eth_getlog rpc
- the latency average dropped by like 90%+ as we introduced early rejection to avoid resource waste
- the nodes are much more responsive, very few situations where they’re lagging

### DIFF
--- a/evmrpc/server.go
+++ b/evmrpc/server.go
@@ -2,11 +2,15 @@ package evmrpc
 
 import (
 	"context"
+	"strings"
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/rpc"
+
 	"github.com/cosmos/cosmos-sdk/baseapp"
 	"github.com/cosmos/cosmos-sdk/client"
 	sdk "github.com/cosmos/cosmos-sdk/types"
-	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/rpc"
 	"github.com/sei-protocol/sei-chain/app/legacyabci"
 	evmrpcconfig "github.com/sei-protocol/sei-chain/evmrpc/config"
 	"github.com/sei-protocol/sei-chain/evmrpc/stats"
@@ -15,8 +19,6 @@ import (
 	"github.com/sei-protocol/sei-chain/x/evm/keeper"
 	"github.com/tendermint/tendermint/libs/log"
 	rpcclient "github.com/tendermint/tendermint/rpc/client"
-	"strings"
-	"sync"
 )
 
 type ConnectionType string


evmrpc/worker_pool_metrics.go
```diff
@@ -0,0 +1,874 @@
...
+	ErrTypeRateLimited   = "rate_limited"
+	ErrTypeBackpressure  = "backpressure"
+	ErrTypeIOSaturated   = "io_saturated"
+	ErrTypeBlockNotFound = "block_not_found"
+	ErrTypeQueueFull     = "queue_full"
+	ErrTypeOther         = "other"
+)
+
+// WorkerPoolMetrics tracks worker pool performance metrics
+type WorkerPoolMetrics struct {
```
masih, line 40: Nonblocker: Sad that this was not implemented using otel. Is it too late to do that instead of adding to the well out of date telemetry and go-metrics stuff considering this repo is already set up with otel?

evmrpc/server.go
```diff
@@ -49,9 +50,27 @@ func NewEVMHTTPServer(
...
+	workerCount := pool.WorkerCount()
+	queueSize := pool.QueueSize()
+
+	// Set DB semaphore capacity in metrics (aligned with worker count)
+	// Only set once to avoid races when multiple test servers start in parallel.
+	pool.Metrics.DBSemaphoreCapacity.CompareAndSwap(0, int32(workerCount)) //nolint:gosec // G115: safe, max is 64
+
+	// Start metrics printer (every 5 seconds)
+	// Prometheus metrics are always exported; stdout printing requires EVM_DEBUG_METRICS=true
+	StartMetricsPrinter(5 * time.Second)
```
masih: Can we extract all of the hardcoded intervals into one constant?

evmrpc/worker_pool_metrics.go
```diff
@@ -0,0 +1,874 @@
...
+				case <-metricsStopChan:
+					return
+				}
+			}
+		}()
+	})
+}
+
+// StopMetricsPrinter stops the metrics printer
+func StopMetricsPrinter() {
```
masih, line 340: As far as I can see this not called from anywhere, which means there is nothing to stop the goroutine started by `StartMetricsPrinter`. 

This one is a blocker.

evmrpc/worker_pool_metrics.go
```diff
@@ -0,0 +1,874 @@
...
+		m.categorizeError(err)
+	} else {
+		m.GetLogsSuccess.Add(1)
+		m.recordBlockRangeBucketSuccess(blockRange)
+	}
+	// Note: Prometheus export is done in batch via ExportPrometheusMetrics()
+}
+
+// recordBlockRangeBucket records the block range into the appropriate bucket
+func (m *WorkerPoolMetrics) recordBlockRangeBucket(blockRange int64) {
```
masih, line 448: Hm, why not use histogram?

ditto for the other occurrences.

evmrpc/worker_pool_metrics.go
yzang2019: Good call, fixed now

evmrpc/server.go
yzang2019: Fixed

evmrpc/worker_pool_metrics.go
yzang2019: Switched to otel

masih (approved):

PR #2592, opened 2025-12-11
canonical encoding for protobuf

pompon0:
For signing protobuf messages we need a well-defined canonical encoding. This PR introduces such encoding, by introducing the following rules:
* only proto3 is supported
* all fields have to have explicit presence (optional/repeated/oneof)
* float/group/map fields are not supported
* fields are encoded in tag ordering
* unknown fields are ignored when encoding
* repeated fields are packed if possible, except for singletons - this way changing field cardinality optional -> repeated is backward compatible (as usual)
The encoding is using ProtoReflect, which is also used by the standard golang encoding (although with some trickery), therefore it should be efficient enough. We can switch to generated code later if needed.

The rules above do not apply to ALL protobuf messages, just the ones marked with "hashable" option. Also this functionality is incompatible with sei-chain gogo generators (which ignore explicit presence), therefore a new package was introduced, which will use a standard golang generator. This package is currently unused, it will be used for new (autobahn) message types only.


scripts/protoc.sh
```diff
@@ -24,6 +24,7 @@ pushd "$(go env GOMODCACHE)/github.com/regen-network/cosmos-proto@v0.3.1" &&
   popd
 
 go run github.com/bufbuild/buf/cmd/buf@v1.58.0 generate
+go run github.com/bufbuild/buf/cmd/buf@v1.58.0 generate --template buf.gen_v2.yaml
```
masih: The `_v2` might be confusing if it is meant to represent new stuff. Because the term "v2" means something totally different in the context of Sei. 

Separately, I would recommend avoid breaking away from the defaults of buf, in this case the gen file name. 

So what to do instead?

Fork what you need into "hashable" and make it a self contained thing: it's own buf.yaml if it needs it, its own buf.gen.yaml and all the go files that go with it.

sei-tendermint/proto_v2/test/test.hashable.go
```diff
@@ -0,0 +1,5 @@
+// Code generated by buf_plugin. DO NOT EDIT.
+package test
```
masih: Can we modify the generation to make sure these test go files not end up in production scope? i.e. names ending with _test.go.

It would also be nice if the package name is hashable_test, rather than "test". the latter is just far too generic.

sei-tendermint/libs/utils/canonical.go
```diff
@@ -0,0 +1,178 @@
+package utils
```
masih: These "utils" have no business of being in tendermint module at all. Why not move them to where they are used, right next to the `main` that calls them in `proto_plugins/hashable`?

sei-tendermint/internal/hashable/hashable.go
```diff
@@ -0,0 +1,178 @@
...
+	"crypto/sha256"
+	"fmt"
+	"slices"
+
+	"google.golang.org/protobuf/encoding/protowire"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+type Hashable interface {
```
masih, line 14: Name the file `hashable.go` since that's a more relevant/explicit term than "canonical"?

sei-tendermint/proto_v2/hashable/hashable.pb.go
```diff
@@ -0,0 +1,82 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
+// versions:
+// 	protoc-gen-go v1.36.10
+// 	protoc        (unknown)
+// source: hashable/hashable.proto
+
+package hashable
```
masih: In terms of code structure, i think it makes far more sense if we get rid of `proto_v2` folder and move everything related to hashable functionality/testing/etc. under `proto_plugins/hashable`. You can then follow the convension of having a /proto folder in there similar to other modules and do not have to spend time explaining what v2 means in the context of proto.

sei-tendermint/internal/hashable/hashable_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	"google.golang.org/protobuf/proto"
+
+	"github.com/stretchr/testify/require"
+
+	testpb "github.com/tendermint/tendermint/proto_v2/test"
+)
+
+// Test checking that the canonical encoding is a valid proto encoding and that it is stable.
+func TestMarshalCanonicalRoundTrip(t *testing.T) {
```
masih, line 18: would be nice to add some `testdata` proto files and run some tests against those, showcasing the use while at it.

buf.gen_v2.yaml
```diff
@@ -0,0 +1,13 @@
+version: v2
+inputs:
+  - directory: sei-tendermint/proto_v2
+plugins:
+  - remote: buf.build/protocolbuffers/go:v1.36.10
+    out: ./sei-tendermint/proto_v2
+    opt: [paths=source_relative]
```
masih: Keep to existing style in this file and others and use `-` instead of mixed JSON+YAML.

sei-tendermint/internal/hashable/hashable.go
```diff
@@ -0,0 +1,178 @@
...
+		protoreflect.Uint64Kind:
+		b = b.Tag(num, protowire.VarintType)
+	case protoreflect.Fixed32Kind, protoreflect.Sfixed32Kind:
+		b = b.Tag(num, protowire.Fixed32Type)
+	case protoreflect.Fixed64Kind, protoreflect.Sfixed64Kind:
+		b = b.Tag(num, protowire.Fixed64Type)
+	case protoreflect.BytesKind, protoreflect.StringKind, protoreflect.MessageKind:
+		b = b.Tag(num, protowire.BytesType)
+	default:
+		panic(fmt.Errorf("unsupported field kind %s", kind))
```
masih, line 103: The `IsHashable` is not used anywhere in the production path that I can see but if used at runtime, panics here worry me. Maybe something to follow up on in the PR that makes use of hashable.

Ditto for other panics.

sei-tendermint/internal/hashable/hashable.go
```diff
@@ -0,0 +1,178 @@
...
+func (b builder) Varint(v uint64) builder    { return protowire.AppendVarint(b, v) }
+func (b builder) Fixed32(v uint32) builder   { return protowire.AppendFixed32(b, v) }
+func (b builder) Fixed64(v uint64) builder   { return protowire.AppendFixed64(b, v) }
+func (b builder) Bytes(bytes []byte) builder { return protowire.AppendBytes(b, bytes) }
+func (b builder) String(s string) builder    { return protowire.AppendString(b, s) }
+
+func (b builder) Message(msg protoreflect.Message) builder {
+	// NOTE: we ignore unknown fields - we are unable to encode them canonically.
+	// NOTE: we can sort fields on init if needed (in the generated files).
+	for _, fd := range sortedFields(msg.Descriptor().Fields()) {
```
masih, line 56: This allocates and sorts on every message. Worth benchmarking?

sei-tendermint/internal/hashable/hashable.go
```diff
@@ -0,0 +1,178 @@
...
+		if fd.IsList() {
```
masih, line 57: Do we need to cover `map` too?

sei-tendermint/internal/hashable/hashable.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	for i := range list.Len() {
+		b = b.Singular(num, kind, list.Get(i))
+	}
+	return b
+}
+
+func (b builder) Singular(num protoreflect.FieldNumber, kind protoreflect.Kind, value protoreflect.Value) builder {
+	switch kind {
```
masih, line 86: Cover for `protoreflect.GroupKind`? otherwise this would panic.

sei-tendermint/internal/hashable/hashable.go
```diff
@@ -0,0 +1,178 @@
...
+	case protoreflect.BytesKind, protoreflect.StringKind, protoreflect.MessageKind:
+		b = b.Tag(num, protowire.BytesType)
+	default:
+		panic(fmt.Errorf("unsupported field kind %s", kind))
+	}
+	return b.Value(kind, value)
+}
+
+func (b builder) Value(kind protoreflect.Kind, value protoreflect.Value) builder {
+	switch kind {
```
masih, line 109: DItto re `protoreflect.GroupKind`?

sei-tendermint/internal/hashable/hashable.go
```diff
@@ -0,0 +1,178 @@
...
+
+	for i := range list.Len() {
+		b = b.Singular(num, kind, list.Get(i))
+	}
+	return b
+}
+
+func (b builder) Singular(num protoreflect.FieldNumber, kind protoreflect.Kind, value protoreflect.Value) builder {
+	switch kind {
+	case protoreflect.BoolKind,
```
masih, line 87: `protoreflect.FloatKind`, `protoreflect.DoubleKind`?

sei-tendermint/internal/hashable/hashable.go
```diff
@@ -0,0 +1,178 @@
...
+	case protoreflect.Sfixed64Kind:
+		return b.Fixed64(uint64(value.Int()))
+	case protoreflect.BytesKind:
+		return b.Bytes(value.Bytes())
+	case protoreflect.StringKind:
+		return b.String(value.String())
+	case protoreflect.MessageKind:
+		return b.Bytes(builder{}.Message(value.Message()))
+	default:
+		panic(fmt.Errorf("unsupported kind %s", kind))
```
masih, line 145: `protoreflect.FloatKind`, `protoreflect.DoubleKind`?

sei-tendermint/proto_v2/hashable/hashable.pb.go
pompon0: it is the other way round. Plugin is part of the hashable logic. Plugin is just a tool for generating hashable types.

buf.gen_v2.yaml
pompon0: [ ] syntax is yaml syntax all right

sei-tendermint/internal/hashable/hashable.go
pompon0: these panics are protected by the type system - only hashable proto types are allowed here.

sei-tendermint/internal/hashable/hashable.go
pompon0: map is not supported, hashable messages don't have maps

sei-tendermint/internal/hashable/hashable.go
pompon0: panic wai, hashable messages don't use group fields. It is checked in the plugin.

sei-tendermint/internal/hashable/hashable.go
pompon0: ditto

sei-tendermint/internal/hashable/hashable.go
pompon0: hashable messages don't have float and double fields. It is checked in the plugin

sei-tendermint/internal/hashable/hashable.go
pompon0: ditto

sei-tendermint/internal/hashable/hashable.go
pompon0: sure, once we have logic using it in place, so that we know what the target is

sei-tendermint/internal/hashable/hashable_test.go
pompon0: wdym?

sei-tendermint/libs/utils/canonical.go
pompon0: wait, what? They are used nowhere rn.

sei-tendermint/proto_v2/test/test.hashable.go
pompon0: Even if I name the package hashable_test and the file hashable_test.proto, the generator will produce hashabe_test.pb.go file, and golang compiler will consider it production code. Do you have other suggestions?

scripts/protoc.sh
pompon0: the fact that I'm creating a separate proto module doesn't mean that all message types will be hashable. Only the things that we will need commitment for have to be hashable, any new p2p messages/stored messages will not. Therefore I don't think that putting everything under "hashable" namespace makes sense. Neither calling this "autobahn" namespace makes sense. I need a name which will imply "put new proto stuff here, not with the old protos", so sth like "proto_giga"/"proto_v3"?

masih: The plugin "hashable" is a self contained functionality. I do not see any benefit in coupling its naming with what v3/giga/autobahn means in Sei, and I highly recommend avoiding `_v2`.

My comment above suggests making "hashable" concept into its own proto functionality that is the usable what whatever wants to use it.

pompon0: ack

PR #2595, opened 2025-12-11
Add lock to protect SetPrices in price feeder

stevenlanders:
- set prices currently may not have been called by the time reads are attempted
- this can cause a panic due to a race condition

- new unit test


udpatil (approved):

arajasek (approved):
LGTM

PR #2600, opened 2025-12-12
[CON-76] fix: sei-tendermint: include all fields in CommitHash

arajasek:

Pretty much as in the title -- the CommitHash now covers all fields.

Added a few new tests, `make test` passes.

### DIFF
--- a/sei-tendermint/types/block.go
+++ b/sei-tendermint/types/block.go
@@ -845,21 +845,34 @@ func (commit *Commit) ValidateBasic() error {
 	return nil
 }
 
-// Hash returns the hash of the commit
+// Hash returns the hash of the commit.
+// It computes a Merkle tree from all commit fields: Height, Round, BlockID, and Signatures.
 func (commit *Commit) Hash() tmbytes.HexBytes {
 	if commit == nil {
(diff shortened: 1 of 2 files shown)


sei-will (approved):

stevenlanders (approved):

sei-tendermint/types/block.go
```diff
@@ -845,21 +845,34 @@ func (commit *Commit) ValidateBasic() error {
 	return nil
 }
 
-// Hash returns the hash of the commit
+// Hash returns the hash of the commit.
+// It computes a Merkle tree from all commit fields: Height, Round, BlockID, and Signatures.
```
udpatil, line 849: nit: we're not computing a merkle tree here? just concating the fields to the hash

arajasek:
@stevenlanders and I will discuss and perform upgrade testing for this, after which we can land this change.

sei-tendermint/types/block.go
arajasek: The hash is the root of the merkle tree, as far as I understand the implementation of `HashFromByteSlices`.

PR #2602, opened 2025-12-15
Increase codecov change tolerance to 3%

masih:
To reduce noise considering code removal efforts.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -17,7 +17,7 @@ coverage:
         flags:
           - sei-chain
       sei-cosmos:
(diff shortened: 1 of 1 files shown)


masih:
Captured issue for the flaking integration test: https://linear.app/seilabs/issue/SEI-10305/integeration-test-evm-module-flakes

masih:
And another under CON team for tendermint: https://linear.app/seilabs/issue/CON-143/testrouter-dialpeer-reject-flakes

PR #2618, opened 2025-12-16
migrated ed25519 primitives from sei-v3

pompon0:
Copy-pasted ed25519 primitives from sei-v3 (they underlying ed25519 library is the same) and adjusted the codebase accordingly. Since we are migrating to autobahn, the changes below prefer the sei-v3 codebase conventions (lmk if you have concerns). 
In particular:
* signatures are wrapped in a type now: ed25519.Signature
* PrivateKey renamed to SecretKey (and SecretKey.PubKey() to SecretKey.Public())
* added best-effort zeroize for SecretKey, copy-prevention and printing-prevention
* Vote and CommitSig now include Option[Signature], because they occur both in signed and unsigned states for some reason.
* all hashes, signatures, proto, json encodings stay backward compatible
* added change-detector test for json encodings of crypto types
* removed typeRegistry from sei-tendermin/libs/json, because it is effectively unused (only PublicKey was registered AND it was not effective, because everywhere the PublicKey was encoded, it was actually using a custom json marshaler).
* removed sei-cosmos/server/in_plage.go command, because it seems to be stale (lmk if it is actually needed)


sei-tendermint/types/block.go
```diff
@@ -640,10 +640,11 @@ const (
...
 type CommitSig struct {
-	BlockIDFlag      BlockIDFlag `json:"block_id_flag"`
-	ValidatorAddress Address     `json:"validator_address"`
-	Timestamp        time.Time   `json:"timestamp"`
-	Signature        []byte      `json:"signature"`
+	BlockIDFlag BlockIDFlag `json:"block_id_flag"`
+	// WARNING: all fields below should be zeroed if BlockIDFlag == BlockIDFlagAbsent
+	ValidatorAddress Address    `json:"validator_address"`
+	Timestamp        time.Time  `json:"timestamp"`
+	Signature        crypto.Sig `json:"signature"`
```
pompon0: crypto.Sig is of a fixed size. The json marshalling will be backward compatible, as long as it is a non-empty slice.

sei-will (approved):

PR #2631, opened 2025-12-18
[giga] pipelining skeleton

codchen:
Create skeleton code and concurrency control for the following model:
```
1st train: decode (unnecessary if the block fed by consensus is already decoded) and stateless checks
this is not exactly a "train", because the processing of the next block can start even before the current block finishes. However it still needs to send finished blocks to the next train in order
2nd train: I'll just call this ExecuteBlock. It's all stateful (anything stateless should happen in the 1st train) so blocks must be executed one after the other (unless we can establish a dependency graph among blocks)
1. wrap each transaction in the block into a task. A task includes:
a. PreprocessTx - includes fee charging, nonce bumping, value transfer, etc.
b. VMCall/VMCreate - [executes](https://github.com/sei-protocol/sei-chain/pull/2610/files#diff-a3fcc02b6a90cd9324dc5c0aec2f6994195165d9674c7a09e1955499e830837fR5-R6) the contract. Note that this is just the entrypoint into the VM. When VM makes its internal calls, it will use whatever functions it has internally and never calls back to these interfaces
c. PostprocessTx - includes gas refund, and send receipt info to the next train
2. feed the tasks into a [scheduler](https://github.com/sei-protocol/sei-chain/pull/2610/files#diff-5ffdd3eea7ab02b10dc36af3d59ae6d8ee3cf03aeae71d8bad25baf14bf3123bR202-R206)
3. wait for the tasks to finish and commit the changeset
3rd train: writes receipt to disk
also not exactly a "train" and can be as parallelized as resources allow
```
unit test for concurrency control

### DIFF
--- a/giga/executor/executor.go
+++ b/giga/executor/executor.go
@@ -1,6 +1,10 @@
 package executor
 
-import "github.com/sei-protocol/sei-chain/giga/executor/tracks"
+import (
+	"context"
+
+	"github.com/sei-protocol/sei-chain/giga/executor/tracks"
+)
 
 type Config struct {
 	StatelessWorkerCount   int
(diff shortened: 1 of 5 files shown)


masih:
I can't find any `context.Context` usage in this highly concurrent pipeline. That usually means foot guns when it comes to highly concurrent go implementations.

PR #2647, opened 2025-12-30
feat: add live state LtHash library

blindchaser:
- Add standalone LtHash implementation (vector, checksum, serialization)
- Add parallel delta computation helpers + timings

### DIFF
--- a/sei-db/ss/pebbledb/hash_test.go
+++ b/sei-db/ss/pebbledb/hash_test.go
@@ -369,7 +369,6 @@ func TestAsyncComputeMissingRanges(t *testing.T) {
 	// Wait until async computation completes / timeouts
 	// We should now have hashed up to version 30 (3 complete ranges)
 	lastHashed = waitForLastRangeHashed(t, db, 40, 5*time.Second)
-	assert.Equal(t, int64(30), lastHashed)
 	assert.Equal(t, int64(40), lastHashed)
 }
 


sei-db/sc/live_state/lthash/parallel.go
```diff
@@ -0,0 +1,230 @@
...
+type KVPairWithOldValue struct {
+	Key            []byte
+	Value          []byte
+	LastFlushValue []byte // Previous value (for MixOut)
+	Deleted        bool   // If true, only MixOut
+}
+
+// ComputeLtHashDeltaParallel computes the LtHash delta for a changeset.
+// For each KV: MixOut(old) if LastFlushValue set, MixIn(new) if not Deleted.
+func ComputeLtHashDeltaParallel(dbName string, kvPairs []KVPairWithOldValue, numWorkers int) (*LtHash, *LtHashTimings) {
```
yzang2019: Why do we care about dbName when computing ltHash from delta?

sei-db/sc/live_state/lthash/lthash.go
```diff
@@ -0,0 +1,247 @@
...
+	for i := 0; i < LtHashSize; i++ {
+		lth.limbs[i] = binary.LittleEndian.Uint16(output[i*2 : (i+1)*2])
+	}
+	xofBufferPool.Put(bufPtr)
+	return lth
+}
+
+// HashKV creates an LtHash from a key-value pair with domain separation.
+// Returns nil if key or value is empty.
+func HashKV(dbName string, key, value []byte) *LtHash {
```
yzang2019: Same here, why do we need dbName?

PR #2649, opened 2025-12-31
IBC Toggle Inbound + Outbound

Kbhat1:


sei-ibc-go/modules/core/keeper/msg_server.go
```diff
@@ -111,6 +111,10 @@ func (k Keeper) ConnectionOpenInit(goCtx context.Context, msg *connectiontypes.M
 func (k Keeper) ConnectionOpenTry(goCtx context.Context, msg *connectiontypes.MsgConnectionOpenTry) (*connectiontypes.MsgConnectionOpenTryResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
+	if !k.IsInboundEnabled(ctx) {
```
yzang2019, line 119: Is this the right/best place to check for inbound?

Kbhat1: We also do this at the core keeper layer not just msg server

PR #2653, opened 2026-01-02
Seidb restructure

yzang2019:
Refactor and restructure seidb to prepare for v3. The new structure will look like this:
image

This PR made changes of:
1. Move files around to different folder names
2. Update all dependency references based on the new structure
3. Add new placeholders for future preparation
4. Add a new basic pebbledb implementation

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,10 +16,8 @@ RUN mkdir -p /go/lib && \
     cp /tmp/wasmvm-libs/libwasmvm.${ARCH_SUFFIX}.so /go/lib/
 
 COPY go.* ./
-COPY sei-wasmd/go.* ./sei-wasmd/
 COPY sei-cosmos/go.* ./sei-cosmos/
 COPY sei-tendermint/go.* ./sei-tendermint/
-COPY sei-ibc-go/go.* ./sei-ibc-go/
 RUN --mount=type=cache,target=/go/pkg/mod \
     --mount=type=cache,target=/root/.cache/go-build \
     go mod download
--- a/sei-tendermint/internal/mempool/mempool.go
+++ b/sei-tendermint/internal/mempool/mempool.go
@@ -206,7 +206,7 @@ func (txmp *TxMempool) Size() int {
 }
 
 func (txmp *TxMempool) utilisation() float64 {
-	return float64(txmp.Size()) / float64(txmp.config.Size)
+	return float64(txmp.NumTxsNotPending()) / float64(txmp.config.Size)
 }
 
 func (txmp *TxMempool) NumTxsNotPending() int {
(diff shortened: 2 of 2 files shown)


sei-db/db_engine/pebbledb/kv/db.go
```diff
@@ -0,0 +1,95 @@
...
+func (db *Database) Has(key []byte) (bool, error) {
+	val, err := db.Get(key)
+	if err != nil {
+		return false, errors.WithStack(err)
+	}
+	return val != nil, nil
+}
+
+// Get returns value by key.
+func (db *Database) Get(key []byte) ([]byte, error) {
```
blindchaser, line 75: from ai: Pebble Get() returns a value that is only valid until the returned closer is closed.
`value` is returned while `closer.Close()` is also called, which means the caller may observe invalid / corrupted data? We should copy `value` before closing the closer. we can do:

```
func (db *Database) Get(key []byte) ([]byte, error) {
    value, closer, err := db.storage.Get(key)
    if err != nil {
        if errors.Is(err, pebble.ErrNotFound) {
            return nil, nil
        }
        return nil, errors.WithStack(err)
    }
    defer closer.Close()

    out := append([]byte(nil), value...) // copy
    return out, nil
}
```

sei-db/db_engine/pebbledb/kv/db.go
```diff
@@ -0,0 +1,95 @@
...
+// Database represents database.
+type Database struct {
+	storage  *pebble.DB
+	writeOps *pebble.WriteOptions
+}
+
+// OpenDB opens an existing or create a new database.
+func OpenDB(dbPath string) *Database {
+	cache := pebble.NewCache(1024 * 1024 * 512)
+	defer cache.Unref()
```
blindchaser, line 19: `defer cache.Unref()` will run when OpenDB() returns, potentially releasing the cache while the DB is still using it?

sei-db/db_engine/pebbledb/kv/batch.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// DeleteByStore deletes key from the store.
+func (b *Batch) DeleteByStore(storeKey string, key []byte) error {
+	prefixedKey := prependStoreKey(storeKey, key)
+	return b.batch.Delete(prefixedKey, nil)
+}
+
+func getStorePrefix(storeKey string) []byte {
+	return []byte(fmt.Sprintf("s/k:%s/", storeKey))
```
blindchaser: nitpick perf improve:
```

func getStorePrefix(storeKey string) []byte {
    // "s/k:" + storeKey + "/"
    b := make([]byte, 0, len("s/k:/")+len(storeKey))
    b = append(b, "s/k:"...)
    b = append(b, storeKey...)
    b = append(b, '/' )
    return b
}
```

sei-db/state_db/sc/types/store.go
```diff
@@ -31,11 +34,31 @@ type Committer interface {
...
+
+	Version() int64
+
+	RootHash() []byte
+
+	Iterator(start, end []byte, ascending bool) dbm.Iterator
+
+	GetProof(key []byte) *ics23.CommitmentProof
+
+	io.Closer
```
blindchaser, line 63: It seems ModuleStore inherits io.Closer, but commitment.NewStore wraps it and never calls Close. Is ModuleStore.Close a no-op? or we should also have it in `commitment.Store` and call close() on rootmulti

sei-db/db_engine/pebbledb/kv/db.go
yzang2019: That's interesting, I think this is copied from v3 and we didn't really see this issue at all in v3. Will still fix it though

sei-db/db_engine/pebbledb/kv/db.go
yzang2019: We have to keep the defer cache.Unref() - it's the correct pattern recommended by PebbleDB. The idea is to keep reference counting down to 0 when closing the DB.

sei-db/state_db/sc/types/store.go
yzang2019: It's called by CMS I think? But this piece of code need more refactory in the future, I would not really touch it in this PR yet

sei-db/db_engine/pebbledb/kv/batch.go
yzang2019: Nice!

sei-db/db_engine/pebbledb/kv/db.go
blindchaser: i see!

sei-db/state_db/sc/types/store.go
blindchaser: got it sg

blindchaser (approved):

sei-db/db_engine/parquet/placeholder.go
```diff
@@ -0,0 +1 @@
+package parquet
```
jewei1997, line 1: 👍

sei-db/ledger_db/event/placeholder.go
```diff
@@ -0,0 +1 @@
+package event
```
jewei1997, line 1: is this for evm logs?

jewei1997 (approved):

sei-db/db_engine/pebbledb/kv/db.go
blindchaser: i double checked this issue. the `defer` executes immediately when OpenDB returns, not when the DB is closed though

PR #2666, opened 2026-01-06
feat: add generic KV interfaces + Pebble adapter

blindchaser:
- Introduce backend-agnostic DB/Batch/Iterator interfaces and ErrNotFound.
- Add pebbledb adapter implementing the db_engine interfaces.
- Remove legacy pebbledb/kv implementation.

### DIFF
--- a/contracts/src/SstoreGasTest.sol
+++ b/contracts/src/SstoreGasTest.sol
@@ -1,0 +1,106 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+/**
+ * @title SstoreGasTest
+ * @notice Contract to test SSTORE gas accounting differences between receipts and debug traces
+ * @dev This contract performs cold SSTORE operations to reproduce the gas discrepancy issue
+ */
+contract SstoreGasTest {
+    // Storage slots that will be written to (cold SSTORE)
+    uint256 public slot0;
+    uint256 public slot1;
+    uint256 public slot2;
+    uint256 public slot3;
+    uint256 public slot4;
+    
+    // Mapping for additional cold storage writes
+    mapping(uint256 => uint256) public data;
+    
+    event StorageWritten(uint256 indexed count, uint256 gasUsed);
+    
+    /**
+     * @notice Performs a single cold SSTORE operation
+     * @param value The value to store
+     */
+    function singleColdSstore(uint256 value) external {
+        uint256 gasBefore = gasleft();
+        slot0 = value;
+        uint256 gasAfter = gasleft();
+        emit StorageWritten(1, gasBefore - gasAfter);
+    }
+    
+    /**
+     * @notice Performs multiple cold SSTORE operations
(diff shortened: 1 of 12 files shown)


sei-db/db_engine/types.go
```diff
@@ -0,0 +1,83 @@
...
+
+// DB is a low-level KV engine contract (business-agnostic).
+//
+// Zero-copy contract:
+//   - Get may return a view into internal buffers; the returned value is only valid
+//     until closer.Close() is called.
+//   - Callers MUST close the closer when done.
+//   - If the value must outlive the closer, callers must copy it (e.g. bytes.Clone).
+type DB interface {
+	Get(key []byte) (value []byte, closer io.Closer, err error)
```
yzang2019: I think it's better not to return closer to the user? From design perspective, we should probably handle it for the user

sei-db/db_engine/pebbledb/db.go
```diff
@@ -0,0 +1,151 @@
...
+			return nil, fmt.Errorf("OpenOptions.Comparer must be *pebble.Comparer, got %T", opts.Comparer)
+		}
+	}
+
+	// Cache is reference-counted. We keep it alive for the lifetime of the DB and
+	// Unref it in (*pebbleDB).Close(). On failure, the deferred cleanup handles it.
+	cache := pebble.NewCache(1024 * 1024 * 512) // 512MB cache
+
+	var db *pebble.DB
+	defer cleanupOnError(&err,
```
yzang2019: I think the original cache is correct.

- pebble.NewCache() - Creates cache with refcount = 1
- pebble.Open(dbPath, opts) - PebbleDB internally calls cache.Ref(), incrementing refcount to 2
- defer cache.Unref() - Runs when OpenDB returns, decrementing refcount back to 1
- Later: db.Close() - PebbleDB internally calls cache.Unref(), decrementing to 0 → cache is freed

sei-db/db_engine/types.go
blindchaser: the current design is for zero-copy reads. If the caller needs to hold the value beyond the closer's lifetime, they clone it themselves. This avoids unnecessary copies when the caller only needs a quick read. however, if we prefer the simplicity over perf, we can change to always clone internally. 

also we will wrap it on the flatkv level

sei-db/db_engine/pebbledb/db.go
blindchaser: oh pebble.Open() also increase 1, will check thx

blindchaser: will revert the change

sei-db/db_engine/types.go
blindchaser: updated to copy inside without return closer

sei-db/db_engine/pebbledb/db.go
blindchaser: done

PR #2668, opened 2026-01-06
Add new giga executor to sei with config

stevenlanders:
- adds new config for process block
- instruments scripts to use new executor if flag is passed 
- adds 

```
# run giga with OCC
GIGA_EXECUTOR=true GIGA_OCC=true ./scripts/benchmark.sh

# run giga without OCC (direct sequential execution with no scheduler)
GIGA_EXECUTOR=true GIGA_OCC=false ./scripts/benchmark.sh

# run non-giga 
GIGA_EXECUTOR=false ./scripts/benchmark.sh
```

### DIFF
--- a/giga/executor/vm/evmc/host_context.go
+++ b/giga/executor/vm/evmc/host_context.go
@@ -156,7 +156,7 @@ func (h *HostContext) Execute(kind evmc.CallKind, recipient evmc.Address, sender
 		return nil, 0, 0, evmc.Address{}, err
 	}
 
-	// For Create/Create2, calculate the created address
+	// todo(pdrobnjak): calculate/propagate created address
 	var createAddr evmc.Address
 	if kind == evmc.Create || kind == evmc.Create2 {
 		// The created address should be set in the execution result


giga/executor/vm/evmc/interpreter.go
pdrobnjak, line 1: Reminder: `Interpreter` is the one currently doing depth management - check if it is handling depth at the right moments in regards to compatibility with `evmc`/`evmone`.

pdrobnjak, line 1: Reminder: In case of delegate calls we might not be propagating `sender` and `recipient` correctly.

giga/executor/vm/evmc/host_context.go
pdrobnjak, line 1: Reminder: In case of delegate calls we might not be propagating `sender` and `recipient` correctly.

app/app.go
```diff
@@ -1490,6 +1527,365 @@ func (app *App) ProcessBlock(ctx sdk.Context, txs [][]byte, req BlockProcessRequ
...
+	// Prepare context for EVM transaction (set infinite gas meter like original flow)
+	ctx = ctx.WithGasMeter(sdk.NewInfiniteGasMeterWithMultiplier(ctx))
+	ctx = ctx.WithTxIndex(txIndex)
+
+	// Create state DB for this transaction
+	stateDB := evmstate.NewDBImpl(ctx, &app.EvmKeeper, false)
+	defer stateDB.Cleanup()
+
+	// Get EVM message from the transaction using recovered sender
+	evmMsg := app.EvmKeeper.GetEVMMessage(ctx, ethTx, sender)
```
pdrobnjak, line 1664: https://github.com/sei-protocol/go-ethereum/blob/ad46da780dee52820de6ccca91e7d797c00f0c10/core/state_transition.go#L170 we should be using this function instead - we can pass our implementation of signer which already has the senders computed.

Not a blocker for merger of this PR, we should probably have a layer above VM e.g. `ExecuteTransaction` where we pass an Ethereum transaction and it handles the conversion to message and invocation of `vm.ApplyMessage`.

PR #2669, opened 2026-01-06
feat: mempool: return all EVM txs before others when reaping

arajasek:

As a result, validators using this mempool logic will produce blocks with a batch of EVM txs, followed by any non-EVM txs. This will allow us to use future executors that can only run EVM txs to run a bigger batch of EVM txs, falling back to existing executors for all non-EVM and interop behaviour.

Note that:
- this PR is NOT consensus-breaking. this is only a mempool operation, not a change to the rules of the protocol.
- this PR does NOT change which txs the mempool reaps. the exact same txs are selected as before, just ordered to be a contiguous slab of EVM txs, followed by a slab of non-EVM ones.

Added a new test, all tests pass

### DIFF
--- a/sei-tendermint/internal/mempool/mempool.go
+++ b/sei-tendermint/internal/mempool/mempool.go
@@ -540,7 +540,8 @@ func (txmp *TxMempool) Flush() {
 }
 
 // ReapMaxBytesMaxGas returns a list of transactions within the provided size
-// and gas constraints. Transaction are retrieved in priority order.
+// and gas constraints. The returned list starts with EVM transactions (in priority order),
+// followed by non-EVM transactions (in priority order).
 // There are 4 types of constraints.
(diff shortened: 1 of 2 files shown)


sei-tendermint/internal/mempool/mempool.go
```diff
@@ -605,14 +608,19 @@ func (txmp *TxMempool) ReapMaxBytesMaxGas(maxBytes, maxGasWanted, maxGasEstimate
...
 			return false
 		}
 		return true
 	})
 
-	return txs
+	return append(evmTxs, nonEvmTxs...)
 }
 
 // ReapMaxTxs returns a list of transactions within the provided number of
```
arajasek, line 626: Note the `ReapMaxTxs` method's behaviour is _not_ changed by this PR, since that method is used for RPC etc. support (and not for block creation).

PR #2671, opened 2026-01-07
Refactor changelog to generic WAL

yzang2019:
Purpose of this PR:
1. Make WAL generic type, previously it is only able to store changeset. Make changelog a common implementation of WAL[ChangeLogEntry[
2. Bump version of tidywal to latest
3. Refactor MemIAVL to use the new interface

Added and modified unit test


sei-db/db_engine/rocksdb/mvcc/db.go
```diff
@@ -112,10 +112,16 @@ func OpenDB(dataDir string, config config.StateStoreConfig) (*Database, error) {
 	}
 	database.latestVersion.Store(latestVersion)
 
-	streamHandler, _ := changelog.NewStream(
+	streamHandler, _ := generic_wal.NewWAL(
```
masih: Please fix all ignored errors across the PR

masih (commented):
Thanks for this Yiming! Before a full review can I ask you to fix all the ignored errors across the PR please?

I also recommend avoiding the unnecessary package nesting, or the word "generic" in the package name. A typical package structure in Go is typically flatter and less hierarchical than languages like Java. I recommend avoiding common package naming like `types` because there are already many such packages across the codebase and working with the package name conflicts would require alias, which can lead to inconsistent alias naming and ultimately higher cognitive cycles when reading the code.

sei-db/db_engine/rocksdb/mvcc/db.go
yzang2019: Yeah, good call!

sei-db/state_db/sc/memiavl/db.go
```diff
@@ -400,41 +386,24 @@ func (db *DB) ApplyChangeSet(name string, changeSet iavl.ChangeSet) error {
...
-	if err != nil {
-		return 0, err
-	}
-	if lastIndex == 0 {
-		return db.SnapshotVersion(), nil
-	}
-	return utils.IndexToVersion(lastIndex, db.initialVersion.Load()), nil
+// CommittedVersion returns the current version of the MultiTree.
+func (db *DB) CommittedVersion() int64 {
+	return db.MultiTree.Version()
```
blindchaser: `memiavl.DB.CommittedVersion()` now returns `MultiTree.Version()`, and 
`checkBackgroundSnapshotRewrite()` only waits on that. 
In async WAL mode, `MultiTree.Version()` can advance ahead of what’s 
actually persisted in the WAL (writes are buffered and flushed by a goroutine), 
so this barrier does not guarantee the WAL has entries up to `lastCommitInfo.Version`. 
Main previously used “WAL last offset → version” as the barrier specifically to avoid 
this race. 

could this be an issue?

yzang2019: Good point, let me verify and think about it

sei-db/state_db/sc/memiavl/db.go
```diff
@@ -557,32 +641,66 @@ func (db *DB) Commit() (version int64, _err error) {
...
+		if err := wal.Write(entry); err != nil {
+			return 0, fmt.Errorf("failed to write changelog WAL: %w", err)
+		}
+		if err := wal.CheckError(); err != nil {
+			return 0, fmt.Errorf("changelog WAL async write error: %w", err)
+		}
+	}
+	db.pendingLogEntry = proto.ChangelogEntry{}
+
 	v, err := db.MultiTree.SaveVersion(true)
```
blindchaser: should we clear the `db.pendingLogEntry` first or SaveVersion() first? like what happens if SaveVersion fails

sei-db/wal/wal.go
```diff
@@ -0,0 +1,307 @@
...
+		walLog.nextOffset++
+	}
+	return nil
+}
+
+// startWriteGoroutine will start a goroutine to write entries to the log.
+// This should only be called on initialization if async write is enabled
+func (walLog *WAL[T]) startWriteGoroutine() {
+	walLog.writeChannel = make(chan T, walLog.config.WriteBufferSize)
+	walLog.errSignal = make(chan error)
```
blindchaser: the `errSignal` is unbuffered and the worker does errSignal <- err, the worker can block forever when callers keep calling Write() without timely CheckError(), and then stop draining writeChannel. Same risk in processor.go.

sei-db/state_db/sc/memiavl/db.go
yzang2019: Good catch! I think you are right, we should SaveVersion first, and then write to WAL

sei-db/wal/wal.go
yzang2019: It is fine in MemIAVL because MemIAVL will call checkError for every block after commit. But I think since we are making it a WAL library interface, it's better to remove this checkError requirement as you pointed out. Good call!

.github/workflows/integration-test.yml
```diff
@@ -175,6 +175,44 @@ jobs:
           done
           unset IFS  # revert the internal field separator back to default
 
+      # The integration test runner executes commands inside docker containers (see integration_test/scripts/runner.py).
```
masih: hm why such complicated way of getting the logs?

`make docker-cluster-start` mounts `build/generated/logs` inside the containers right?

We could simplify the two build steps added here into one, where _if the build fails_ print whatever lines from the tail of logs.

Optionally you can conditionally upload the entire logs directory as a build artifact in github actions, and that would give you everything. 

Right?

sei-db/db_engine/pebbledb/mvcc/db.go
```diff
@@ -150,16 +151,15 @@ func OpenDB(dataDir string, config config.StateStoreConfig) (*Database, error) {
...
-		},
-	)
+	streamHandler, err := wal.NewChangelogWAL(logger.NewNopLogger(), utils.GetChangelogPath(dataDir), wal.Config{
+		DisableFsync:  true,
+		ZeroCopy:      true,
+		KeepRecent:    uint64(config.KeepRecent),
+		PruneInterval: time.Duration(config.PruneIntervalSeconds) * time.Second,
+	})
+	if err != nil {
+		panic(err)
```
masih: Why panic instead of bubbling up the error? `OpenDB` can return error right?

sei-db/db_engine/pebbledb/mvcc/db.go
```diff
@@ -150,16 +151,15 @@ func OpenDB(dataDir string, config config.StateStoreConfig) (*Database, error) {
...
-		changelog.Config{
-			DisableFsync:  true,
-			ZeroCopy:      true,
-			KeepRecent:    uint64(config.KeepRecent),
-			PruneInterval: time.Duration(config.PruneIntervalSeconds) * time.Second,
-		},
-	)
+	streamHandler, err := wal.NewChangelogWAL(logger.NewNopLogger(), utils.GetChangelogPath(dataDir), wal.Config{
+		DisableFsync:  true,
+		ZeroCopy:      true,
```
masih: I see that `ZeroCopy` is always set to `true` across all instantiations of WAL in the codebase.

I am curious as to why this was made configurable to begin with. Is there an opportunity to simplify the code by implicitly setting it to true until the  actual need for parameterisation arises?

Same goes for `DisableFsync`.

sei-db/db_engine/rocksdb/mvcc/db.go
```diff
@@ -112,16 +113,15 @@ func OpenDB(dataDir string, config config.StateStoreConfig) (*Database, error) {
...
-		},
-	)
+	streamHandler, err := wal.NewChangelogWAL(logger.NewNopLogger(), utils.GetChangelogPath(dataDir), wal.Config{
+		DisableFsync:  true,
+		ZeroCopy:      true,
+		KeepRecent:    uint64(config.KeepRecent),
+		PruneInterval: time.Duration(config.PruneIntervalSeconds) * time.Second,
+	})
+	if err != nil {
+		panic(err)
```
masih: Ditto re panic vs returning error.

sei-db/state_db/sc/memiavl/db.go
```diff
@@ -55,6 +53,13 @@ type DB struct {
 	readOnly bool
 	opts     Options
 
+	// streamHandler is the changelog WAL owned by MemIAVL.
+	// It is opened during OpenDB (if present / allowed) and closed in DB.Close().
+	streamHandler wal.GenericWAL[proto.ChangelogEntry]
```
masih: * Use `wal.ChangelogWAL` type? 
   * Ditto for other occurrences of `wal.GenericWAL[proto.ChangelogEntry]` in the codebase.

Small notes: this is where i would nudge that using generics is an overkill. we seem to only ever use it for one type. I would nudge that even the `wal.GenericWAL` interface is over refactored? if we only ever are going to use single implementation of that then what is the point of the interface? I see no mock testing requirements in the generic wal test package for example. In Go etiquette the choices of refactoring is strongly influenced by the additional LoC it introduces. A core design goal is to reduce LoC and opt for absolutely boring code/ if and elses etc. that is highly readable and easy to maintain.
Unless there are concrete plans to introduce other types?

sei-db/state_db/sc/memiavl/db.go
```diff
@@ -199,20 +203,34 @@ func OpenDB(logger logger.Logger, targetVersion int64, opts Options) (database *
...
+		walIndexDelta = mtree.WorkingCommitInfo().Version - 1
 	}
 
-	if targetVersion == 0 || targetVersion > mtree.Version() {
+	// Replay WAL to catch up to target version (if WAL has entries)
+	if walHasEntries && (targetVersion == 0 || targetVersion > mtree.Version()) {
 		logger.Info("Start catching up and replaying the MemIAVL changelog file")
-		if err := mtree.Catchup(context.Background(), streamHandler, targetVersion); err != nil {
-			return nil, errorutils.Join(err, streamHandler.Close())
+		if err := mtree.Catchup(context.Background(), streamHandler, walIndexDelta, targetVersion); err != nil {
```
masih, line 230: Question please: in a typical runtime environment how long does this take to run? what about the worst case?

sei-db/state_db/sc/memiavl/db.go
```diff
@@ -199,20 +203,34 @@ func OpenDB(logger logger.Logger, targetVersion int64, opts Options) (database *
 		tree.snapshot.leavesMap.PrepareForRandomRead()
 	}
 
-	// Create rlog manager and open the rlog file
-	streamHandler, err := changelog.NewStream(logger, utils.GetChangelogPath(opts.Dir), changelog.Config{
+	// MemIAVL owns changelog lifecycle: always open the WAL here.
+	// Even in read-only mode we may need WAL replay to reconstruct non-snapshot versions.
+	streamHandler, err := wal.NewChangelogWAL(logger, utils.GetChangelogPath(opts.Dir), wal.Config{
```
masih, line 208: The ChangeLogWAL implements `io.Closer` -- It has resources open that need to be closed up/cleaned up, right?

But in the remaining cod, the execution may exit with `error`, and does so without calling `Close` on the opened changelog wal that's no longer needed. This can result in resource leak.

I recommend revisiting the opened resources and closing the wal explicitly before existing with error.

Separately, I see this is also the case for `MultiTree` constructed in earlier lies. I understand this is a pre-existing issue and unmodified in this PR. But while at it we might want to review closing `MultiTree` too since it can also result in resource leak.

sei-db/state_db/sc/memiavl/db.go
```diff
@@ -231,11 +249,16 @@ func OpenDB(logger logger.Logger, targetVersion int64, opts Options) (database *
...
-		// truncate the rlog file
-		logger.Info("truncate rlog after version: %d", targetVersion)
-		truncateIndex := utils.VersionToIndex(targetVersion, mtree.initialVersion.Load())
-		if err := streamHandler.TruncateAfter(truncateIndex); err != nil {
-			return nil, fmt.Errorf("fail to truncate rlog file: %w", err)
+		// truncate the rlog file (if WAL is provided and has entries)
+		if walHasEntries {
+			logger.Info("truncate rlog after version: %d", targetVersion)
+			// Use O(1) conversion: walIndex = version - delta
+			truncateIndex := targetVersion - walIndexDelta
```
masih, line 254: Future work: It would be awesome if we review the "version" type passing across the entire repo. Sometimes we use signed int, sometimes we use unsigned, we then convert from one to another in places etc. etc. This is a security nightmare and a footgun for engineers.

sei-db/state_db/sc/memiavl/db.go
```diff
@@ -400,41 +431,37 @@ func (db *DB) ApplyChangeSet(name string, changeSet iavl.ChangeSet) error {
...
-		return db.SnapshotVersion(), nil
+// CommittedVersion returns the current version of the MultiTree.
+func (db *DB) CommittedVersion() int64 {
+	// Prefer the WAL's last offset, converted via walIndexDelta, to avoid relying
+	// on potentially stale tree metadata. Fall back to the tree version on error
+	// or when WAL is absent/empty.
+	if wal := db.GetWAL(); wal != nil {
+		lastOffset, err := wal.LastOffset()
+		if err != nil {
+			db.logger.Error("failed to read WAL last offset for committed version", "err", err)
```
masih: Question: Could you help me understand how ignoring this error, i.e. logging it and proceeding with the flow, will self correct later? Is the idea that the code would eventually recommit a version and an occasional error is not a show stopper? Thinking out loud, I am trying to see how this error could cause issues down the line. thanks

sei-db/state_db/sc/memiavl/db.go
```diff
@@ -525,19 +552,54 @@ func (db *DB) pruneSnapshots() {
...
+		firstVersion = entry.Version
+		return nil
+	})
+	if err != nil {
+		return 0, false, err
+	}
+
+	// delta = version - index, so for any entry: version = index + delta
+	// #nosec G115 -- WAL indices are always much smaller than MaxInt64 in practice
+	return firstVersion - int64(firstIndex), true, nil
```
masih, line 573: Detto re comment about future work.

sei-db/state_db/sc/memiavl/multitree.go
```diff
@@ -420,11 +450,14 @@ func (t *MultiTree) Catchup(ctx context.Context, stream types.Stream[proto.Chang
 	if err != nil {
```
masih, line 450: Why check error after waiting for all async writes to complete? Is that intended?

sei-db/state_db/sc/memiavl/db_test.go
```diff
@@ -11,12 +11,13 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/require"
```
masih, line 14: I recommend not making more than two import groups: one for go SDK, and the other for everything else.

sei-db/state_db/sc/memiavl/db_test.go
masih: All `t.TempDir`s are automatically deleted. This line can be removed. Ditto for the rest of the tests.

sei-db/state_db/sc/memiavl/db_test.go
```diff
@@ -481,39 +499,250 @@ func TestRlogIndexConversion(t *testing.T) {
...
+// 3. All conversions are then O(1): walIndex = version - delta
+func TestWALIndexDeltaComputation(t *testing.T) {
+	testCases := []struct {
+		name           string
+		initialVersion uint32
+		numVersions    int
+		rollbackTo     int64
+	}{
+		{
+			name:           "delta=0 (version starts at 1)",
```
masih: I recommend avoiding `(){}[]=` symbols in tests and stick to space numbers and characters. It makes the "cleaned" test name more readable where spaces are replaced by `_` and that's it.

sei-db/state_db/sc/memiavl/db_test.go
```diff
@@ -481,39 +499,250 @@ func TestRlogIndexConversion(t *testing.T) {
...
+			db, err := OpenDB(logger.NewNopLogger(), 0, Options{
+				Dir:             dir,
+				CreateIfMissing: true,
+				InitialStores:   initialStores,
+				InitialVersion:  tc.initialVersion,
+			})
+			require.NoError(t, err)
+
+			// Commit multiple versions
+			for i := 0; i < tc.numVersions; i++ {
```
masih, line 550: Try using the newer range over number syntax. Ditto in other places.
```suggestion
			for i := range tc.numVersions {
```

sei-db/state_db/sc/memiavl/db_test.go
```diff
@@ -481,39 +499,250 @@ func TestRlogIndexConversion(t *testing.T) {
...
+
+			// When initialVersion=0, first commit is version 1, so after N commits: version = N
+			// When initialVersion=X, first commit is version X, so after N commits: version = X + N - 1
+			expectedVersion := int64(tc.numVersions)
+			if tc.initialVersion > 0 {
+				expectedVersion = int64(tc.initialVersion) + int64(tc.numVersions) - 1
+			}
+			require.Equal(t, expectedVersion, db.Version())
+
+			require.NoError(t, db.Close())
```
masih, line 574: Great 👍 

Small note: Asserting close does not error is consistently applied in tests here.

sei-db/state_db/sc/store_test.go
```diff
@@ -0,0 +1,1002 @@
...
+
+func TestCommitStoreBasicOperations(t *testing.T) {
+	dir := t.TempDir()
+	cs := NewCommitStore(dir, logger.NewNopLogger(), config.StateCommitConfig{})
+	cs.Initialize([]string{"test"})
+
+	// Load version 0 to initialize the DB
+	_, err := cs.LoadVersion(0, false)
+	require.NoError(t, err)
+	defer cs.Close()
```
masih: Assert no error on close. Ditto in other places.

sei-db/wal/wal.go
```diff
@@ -0,0 +1,309 @@
...
+		logger:    logger,
+		marshal:   marshal,
+		unmarshal: unmarshal,
+		closeCh:   make(chan struct{}),
+		// isClosed is zero-initialized to false (atomic.Bool)
+	}
+	// Finding the nextOffset to write
+	lastIndex, err := log.LastIndex()
+	if err != nil {
+		return nil, err
```
masih: Ditto re resource leak. Do we need to close off the log before returning here? Please also check the reminder of this constructor.

sei-db/wal/wal.go
```diff
@@ -0,0 +1,309 @@
...
+	return w, nil
+
+}
+
+// Write will append a new entry to the end of the log.
+// Whether the writes is in blocking or async manner depends on the buffer size.
+// For async writes, this also checks for any previous async write errors.
+func (walLog *WAL[T]) Write(entry T) error {
+	// Check for any previous async write error
+	if errPtr := walLog.asyncErr.Load(); errPtr != nil {
```
masih: This is unsafe for concurrent use: multiple goroutines could be writing where one errors just after the other one has finished executing this check.

I see that the underlying WAL library used for the core implementation aims to be safe for concurrent use. I see that we use "atomic" reference to provide _some_ safetly, and therefore I assume safety in concurrent use is a requirement.  To correct this, basically the entire Write (and other recievers in this struct) need to be guarded by a mutex (a RW mutex).

sei-db/wal/wal.go
```diff
@@ -0,0 +1,309 @@
...
+	channelBufferSize := walLog.config.WriteBufferSize
+	if channelBufferSize > 0 {
+		if walLog.writeChannel == nil {
+			walLog.logger.Info(fmt.Sprintf("async write is enabled with buffer size %d", channelBufferSize))
+			walLog.startWriteGoroutine()
+		}
+		// async write
+		walLog.writeChannel <- entry
+		walLog.nextOffset++
+	} else {
```
masih, line 105: Style comment: try to avoid nesting in code by using return statements. For example, the if above could add a return nil in which case we will no longer need the `else{}` nesting.

sei-db/wal/wal.go
```diff
@@ -0,0 +1,309 @@
...
+}
+
+// startWriteGoroutine will start a goroutine to write entries to the log.
+// This should only be called on initialization if async write is enabled
+func (walLog *WAL[T]) startWriteGoroutine() {
+	walLog.writeChannel = make(chan T, walLog.config.WriteBufferSize)
+	// Capture the starting offset for the goroutine
+	writeOffset := walLog.nextOffset
+	walLog.wg.Add(1)
+	go func() {
```
masih, line 127: OK write _is_ used concurrently. We have fixes to do in terms of safety I think.

sei-db/wal/wal.go
```diff
@@ -0,0 +1,309 @@
...
+		defer walLog.wg.Done()
+		batch := wal.Batch{}
+		for {
+			entries := channelBatchRecv(walLog.writeChannel)
```
masih: Style comment: I recommend against refactoring `channelBatchRecv` in favour if inline implementation here for better readability, considering the critical role that channel receive plays in metering the rate of this forever loop execution. Also this "util" function is only used once.

By not refactoring that function, the for implementation makes it obvious that channel closure breaks the loop. `Channel` is not utility in this case. It is a first-class vessel that drives the primary function of WAL: writing.

sei-db/wal/wal.go
```diff
@@ -0,0 +1,309 @@
...
+// This should only be called on initialization if async write is enabled
+func (walLog *WAL[T]) startWriteGoroutine() {
+	walLog.writeChannel = make(chan T, walLog.config.WriteBufferSize)
+	// Capture the starting offset for the goroutine
+	writeOffset := walLog.nextOffset
+	walLog.wg.Add(1)
+	go func() {
+		defer walLog.wg.Done()
+		batch := wal.Batch{}
+		for {
```
masih: unconditional forever loops are generally harmful; try to condition them.

Here unfortunately there is no context being passed around and we cannot use the context error being nil as the condition to continue. 

A great fix would be to revisit the recievers in WAL and use Context when appropriate: whenever we want a handle on some background job, or being able to gracefully interrupt the flow.

sei-db/wal/wal.go
```diff
@@ -0,0 +1,309 @@
...
+			entries := channelBatchRecv(walLog.writeChannel)
+			if len(entries) == 0 {
+				// channel is closed
+				break
+			}
+
+			for _, entry := range entries {
```
masih: If we were to not refactor the channel receive, we could have unmarshalled in streaming fashion and removed the need to iterate over entries again. Another added benefit would have been that we could avoid allocating the slice of entries fully in case an error occurs mid entry unmarshal.

sei-db/wal/wal.go
```diff
@@ -0,0 +1,309 @@
...
+func (walLog *WAL[T]) StartPruning(keepRecent uint64, pruneInterval time.Duration) {
+	// Use a minimum interval to avoid tight loops
+	if pruneInterval <= 0 {
+		pruneInterval = time.Minute // Default to 1 minute if not specified
+	}
+
+	ticker := time.NewTicker(pruneInterval)
+	defer ticker.Stop()
+
+	for {
```
masih: Ditto re forever loop and context.

sei-db/wal/wal.go
```diff
@@ -0,0 +1,309 @@
...
+		select {
+		case <-walLog.closeCh:
+			return
+		case <-ticker.C:
+			lastIndex, _ := walLog.log.LastIndex()
+			firstIndex, _ := walLog.log.FirstIndex()
+			if lastIndex > keepRecent && (lastIndex-keepRecent) > firstIndex {
+				prunePos := lastIndex - keepRecent
```
masih: Move this variable definition out the if and reuse? it would make reading the if condition easier on the brain.

sei-db/wal/wal.go
```diff
@@ -0,0 +1,309 @@
...
+	ticker := time.NewTicker(pruneInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-walLog.closeCh:
+			return
+		case <-ticker.C:
+			lastIndex, _ := walLog.log.LastIndex()
+			firstIndex, _ := walLog.log.FirstIndex()
```
masih: Errors ignored here. For example WAL could be corrupt. What would go wrong there?

If safe to ignore, can I ask you to explain why with an inline comment for the future readers of this code please?

sei-db/wal/wal.go
```diff
@@ -0,0 +1,309 @@
...
+	if pruneInterval <= 0 {
+		pruneInterval = time.Minute // Default to 1 minute if not specified
+	}
+
+	ticker := time.NewTicker(pruneInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-walLog.closeCh:
```
masih: you can replace `closeCh` with a running (cancellable) context. But unfortunately the underlying WAL library we use does not take context. Sad face.

sei-db/wal/wal_test.go
```diff
@@ -0,0 +1,484 @@
...
+		{"failure-1", []byte("\n"), 0},
+		{"failure-2", []byte(`{}` + "\n"), 0},
+		{"failure-3", []byte(`{"index":"1"}` + "\n"), 0},
+		{"failure-4", []byte(`{"index":"1","data":"?"}`), 0},
+		{"failure-5", []byte(`{"index":1,"data":"?"}` + "\n" + `{"index":"1","data":"?"}`), 1},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			os.WriteFile(filepath.Join(dir, "00000000000000000001"), tc.logs, 0o600)
```
masih: Assert no error?

sei-db/wal/wal_test.go
```diff
@@ -0,0 +1,484 @@
...
+	require.Equal(t, uint64(0), lastIndex)
+
+	// Clean up
+	err = log.Close()
+	require.NoError(t, err)
+}
+
+func TestTruncateAfter(t *testing.T) {
+	changelog := prepareTestData(t)
+	defer changelog.Close()
```
masih: Assert no error, i.e. graceful closure? Ditto for the rest of the tests.

sei-db/wal/wal_test.go
```diff
@@ -0,0 +1,484 @@
...
+	require.NoError(t, err)
+
+	// Use utility function to get last index without opening stream
+	lastIndex, err := GetLastIndex(dir)
+	require.NoError(t, err)
+	require.Equal(t, uint64(3), lastIndex)
+}
+
+func TestLogPath(t *testing.T) {
+	path := LogPath("/some/dir")
```
masih, line 622: The "util" LogPath is only used once. It adds an additional 3 lines to the codebase and makes the reader jump into another file to follow what the code does. This is another example of over-refactoring that hardly adds any benefit at all I'm afraid.

masih (commented):
Main blockers:
* Potential resource leakage.
* Concurrent use safety of WAL

Other than those I left a bunch of feedback across the code

Thanks for working on this @yzang2019 ! 🍻 This is nearly there

sei-db/db_engine/pebbledb/mvcc/db.go
yzang2019: Yeah I think this was mostly to make the config more explicit but I agree it can be set default to true to avoid unnecessary override

sei-db/db_engine/pebbledb/mvcc/db.go
yzang2019: Yup, we can definitely return error as well

sei-db/wal/wal.go
yzang2019: Yeah I think we do have the close function to interrupt the flow, we don't really have context for WAL since that might introduce extra complexity

.github/workflows/integration-test.yml
yzang2019: This was added by AI to debug the integration test failure running on github, yeah I think we can simplify it a bit

sei-db/db_engine/rocksdb/mvcc/db.go
yzang2019: Should be fixed, thanks!

sei-db/state_db/sc/memiavl/db.go
yzang2019: Yup, will make the changes for type.
We will need to support more types of WAL in the future for v3, for example, we will use a different WAL for receipt. So that's why we want to make it generic first.

sei-db/wal/wal.go
yzang2019: You are right, I think we should check the errors here, gonna fix this, good catch!

sei-db/state_db/sc/memiavl/db.go
yzang2019: It could be ranging from couple of minutes to 1 hour

sei-db/state_db/sc/memiavl/db.go
yzang2019: If the execution exit with error then it means either the process was killed or panic occurs, at that point I don't think closing it or not matter that much since it's not gonna be a resource leak, it would need the operator to restart the process anyway which at that time all resources will be released

sei-db/state_db/sc/memiavl/db_test.go
yzang2019: I'm good with that, previously I think Lint will complain if you don't separate into 3 groups:
1. golang internal
2. other open sources dependency
3. sei specific dependency

I personally feels that is cleaner than just 2 groups

sei-db/state_db/sc/memiavl/db.go
yzang2019: Yeah, it's pretty inconsistent right now, we should in the future unify it for version

sei-db/state_db/sc/memiavl/db.go
yzang2019: Good question, I personally feel lt's still better to return the error and pop it up, so I will change it

sei-db/state_db/sc/memiavl/multitree.go
yzang2019: I believe that's intentional, because the error is only assigned during processing, and we need to wait until all async process finishes incase there's any error happened

sei-db/wal/wal.go
yzang2019: will probably keep closeCh here

sei-db/state_db/sc/memiavl/db_test.go
yzang2019: The canonical (lint-friendly) rule: 3 import groups

Most Go teams and lint setups expect imports to be grouped like this:

Standard library
Third-party (external) packages
Local / project packages

sei-db/state_db/sc/memiavl/db_test.go
yzang2019: Cool, removed

sei-db/state_db/sc/memiavl/db_test.go
yzang2019: Make sense

sei-db/state_db/sc/store_test.go
yzang2019: Fixed

sei-db/wal/wal.go
yzang2019: This should be fixed now, please take another look

sei-db/wal/wal.go
yzang2019: This part is rewritten

sei-db/wal/wal.go
yzang2019: Fixed

sei-db/wal/wal.go
yzang2019: This part is rewritten

blindchaser:
LGTM. not a blocker, but some thoughts on adding test coverage:

Concurrency testing:
- It would be valuable to add a testings that exercises concurrent WAL operations (e.g. async writes while truncation/pruning is running, or Close() racing with in-flight writes) to ensure there are no deadlocks, panics, or log corruption under concurrent access.

Isolation testing:
- Since WAL/changelog is now scoped per DB, it would be good to add a multi-DB isolation test:
  - open two DBs (separate dirs) in the same process,
  - interleave commits on both,
  - trigger snapshot rewrite and/or WAL truncation on DB A,
  - assert DB B’s WAL offsets, CommittedVersion, and state are unaffected.

sei-db/wal/wal_test.go
yzang2019: Fixed all

yzang2019:
> LGTM. not a blocker, but some thoughts on adding test coverage:
> 
> Concurrency testing:
> 
> * It would be valuable to add a testings that exercises concurrent WAL operations (e.g. async writes while truncation/pruning is running, or Close() racing with in-flight writes) to ensure there are no deadlocks, panics, or log corruption under concurrent access.
> 
> Isolation testing:
> 
> * Since WAL/changelog is now scoped per DB, it would be good to add a multi-DB isolation test:
>   
>   * open two DBs (separate dirs) in the same process,
>   * interleave commits on both,
>   * trigger snapshot rewrite and/or WAL truncation on DB A,
>   * assert DB B’s WAL offsets, CommittedVersion, and state are unaffected.

Make sense for concurrency point, for isolation I dont think we need to test that much, since different WAL in different folder should be fully isolated, so I dont think there's any concern around that

sei-db/db_engine/pebbledb/db_test.go
```diff
@@ -14,7 +15,7 @@ func TestDBGetSetDelete(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Open: %v", err)
 	}
-	defer func() { _ = db.Close() }()
+	t.Cleanup(func() { require.NoError(t, db.Close()) })
```
masih, line 18: A small note on `t.Cleanup` vs `defer`: `t.Cleanup` is executed after the test belonging to `t` ends. Whereas, `defer` is executed when the function returns.

We do not always want one or the other. For example, in a test helper function we usually want `t.Cleanup` if it creates resources for the tests that are done after function returns. 

In this specific case and the remaining changes here it makes no difference in using one or the other. But I thought I point out the difference so that we are careful not to always use either of those.

sei-db/state_db/sc/memiavl/db.go
masih: So far probably for the current implementation?

There is one way to make sure resource leaks cannot happen no matter how the constructor is used: On partial instantiation, the constructor should clean up after itself.

sei-db/state_db/sc/memiavl/multitree.go
masih: Sorry, but what async process?

Looking at the code the entire operation is executed sequentially. We just happen to pass "process" as a function. Right?

sei-db/wal/wal.go
```diff
@@ -0,0 +1,332 @@
...
+	dir             string
+	log             *wal.Log
+	config          Config
+	logger          logger.Logger
+	marshal         MarshalFn[T]
+	unmarshal       UnmarshalFn[T]
+	writeChannel    chan T
+	mtx             sync.RWMutex
+	asyncWriteErrCh chan error // buffered=1; async writer reports first error non-blocking
+	isClosed        atomic.Bool
```
masih: Why atomic bool at all? We already have a mutex that guards calls to receivers. We can use it to set the close flag.

sei-db/wal/wal.go
```diff
@@ -0,0 +1,332 @@
...
+// WAL is a generic write-ahead log implementation.
+type WAL[T any] struct {
+	dir             string
+	log             *wal.Log
+	config          Config
+	logger          logger.Logger
+	marshal         MarshalFn[T]
+	unmarshal       UnmarshalFn[T]
+	writeChannel    chan T
+	mtx             sync.RWMutex
```
masih: Style: it is common practice to add a comment above the mutex field explaining what it guards. If it guards fields, it is usually also grouped with the fields it guards in the struct definition.

sei-db/wal/wal.go
```diff
@@ -0,0 +1,332 @@
...
+			}
+
+		}
+	}()
+}
+
+// TruncateAfter will remove all entries that are after the provided `index`.
+// In other words the entry at `index` becomes the last entry in the log.
+func (walLog *WAL[T]) TruncateAfter(index uint64) error {
+	if err := walLog.log.TruncateBack(index); err != nil {
```
masih: To be clear, the call here is protected by a different mutex than the one that protects write for example.

This can result is subtle bugs that will take a long time to debug. Be very careful doing this kind of thing and understand the consequences: It means that WAL does not guarantee atomic operations across all its functions; instead it only guarantees it across Write, and Close.

sei-db/wal/wal.go
```diff
@@ -0,0 +1,332 @@
...
+		}
+		err = processFn(i, entry)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (walLog *WAL[T]) StartPruning(keepRecent uint64, pruneInterval time.Duration) {
```
masih: What happens if this function is called multiple times, for overlapping range when previous pruning has not finished?

This function is exported too. We should guard against this kind of usage.

sei-db/wal/wal_test.go
```diff
@@ -0,0 +1,764 @@
...
+		for !stop.Load() {
+			<-ticker.C
+			first, err1 := changelog.FirstOffset()
+			last, err2 := changelog.LastOffset()
+			if err1 != nil || err2 != nil || first == 0 || last == 0 || first > last {
+				continue
+			}
+			// Read around the midpoint.
+			mid := first + (last-first)/2
+			_, _ = changelog.ReadAt(mid)
```
masih: Please check for all ignored errors and update the AI prompt.

sei-db/wal/wal_test.go
```diff
@@ -0,0 +1,764 @@
...
+
+	writerWg.Wait()
+	stop.Store(true)
+	bgDone := make(chan struct{})
+	go func() { bgWg.Wait(); close(bgDone) }()
+	select {
+	case <-bgDone:
+	case <-time.After(3 * time.Second):
+		t.Fatalf("background workers did not exit (possible deadlock)")
+	}
```
masih: Use `RequireEventually`.

The logic of the test here seems hard to follow honestly and smells like suboptimal AI generation? It can be massively simplified and for specific test cases. I recommend revisiting this if you can.

sei-db/state_db/sc/memiavl/db_test.go
masih: and I specifically advise against the 3 group, in favour of a simpler 2 group. This is based on numerous previous cases and a ton of unnecessary time wasting in lintage.

FWIW either of those paths are inconsistently applied in this repo.

sei-db/wal/wal.go
yzang2019: That's true, will add some checks

sei-db/wal/wal_test.go
yzang2019: Simplified the test

sei-db/wal/wal_test.go
yzang2019: Fixed all, thanks!

sei-db/wal/wal.go
yzang2019: Yeah, unnecessary here, we can simplify it to avoid atomic bool

sei-db/wal/wal.go
yzang2019: Initially I was planning to add the same mutex protection, but considering the wal library itself has locking protection underneath, I think it's probably not needed an extra lock.

sei-db/wal/wal.go
yzang2019: Actually I think it's better to make this a private function, we don't need to expose this any way

sei-db/wal/wal.go
yzang2019: Added a comment

PR #2675, opened 2026-01-08
Adjusted RPC http requests to use POST instead of GET

monty-sei:
RPC calls for the EVM CLI were attempting to send methods via GET instead of POST, resulting in failed extraction of results from the associated RPC calls as none were being returned.

Adjustments to `getChainId()`, `getNonce()` and `CmdAssociateAddress()` were made 

NOTE: the latter `CmdAssociateAddress` had an internal cmd method call to `sei_associate` using GET instead of POST.

Fixes REL-107


x/evm/client/cli/tx_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"math/big"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	EVM_RPC_MAINNET = "https://evm-rpc.sei-apis.com"
+	EVM_RPC_TESTNET = "https://evm-rpc-testnet.sei-apis.com"
```
masih: We should generally avoid tests that assume a remove endpoint is up and available. 

What I recommend doing instead is to:

* make these tests run conditionally/on CI when some env var is set; skipped by default.
* use `httptest` package to assert the behaviour you expect at unit level.

x/evm/client/cli/tx_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"crypto/rand"
+	"math/big"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	EVM_RPC_MAINNET = "https://evm-rpc.sei-apis.com"
```
masih: This naming convention for constants is used in Java but doesn't translate to Golang best practices. For start, it makes the constant exported (more or less equivelant of `public` in Java) because the first letter is a captial letter.

Instead of `CAPITAL_SNAKE_CASE`, we should instead be using `camelCase` for constants that don't need exporting like the ones here, and `CaplitalCamelCase` for the ones that do need exporting.

x/evm/client/cli/tx_test.go
```diff
@@ -0,0 +1,50 @@
...
+	EVM_RPC_TESTNET = "https://evm-rpc-testnet.sei-apis.com"
+)
+
+func TestGetChainId(t *testing.T) {
+	//Run on mainnet RPC url
+	mChainId, err := getChainId(EVM_RPC_MAINNET)
+	require.Nil(t, err)
```
masih: Please use `requireNoError` to assert nil/no error. That way we get nicer test failure logging.

Ditto for the reminder of this PR.

x/evm/client/cli/tx_test.go
```diff
@@ -0,0 +1,50 @@
...
+	//Check nonce on both mainnet and testnet
+	nonce, err := getNonce(EVM_RPC_MAINNET, privateKey.PublicKey)
+	require.Nil(t, err)
+	require.Equal(t, nonce, uint64(0))
+
+	nonce, err = getNonce(EVM_RPC_TESTNET, privateKey.PublicKey)
+	require.Nil(t, err)
+	require.Equal(t, nonce, uint64(0))
+
+	//NOTE: Could add tests for known active public keys for mainnet and testnet
```
masih: I recommend capturing an issue for this if you'd like to break it to its own PR.

Otherwise, I recommend making these tests into a table-test that simply gets a set of URLs and runs. This way we can easily add additional endpoints without the need to rewrite/refactor the test logic.

x/evm/client/cli/tx_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"crypto/ecdsa"
+	"crypto/rand"
+	"math/big"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/stretchr/testify/require"
+)
+
+var (
```
masih: These should be `const` instead.

x/evm/client/cli/tx_test.go
monty-sei: This is done

x/evm/client/cli/tx_test.go
monty-sei: This is done

x/evm/client/cli/tx_test.go
monty-sei: I moved to table-test format but have not supported additional known nonce cases yet

x/evm/client/cli/tx_test.go
monty-sei: This is done

x/evm/client/cli/tx_test.go
monty-sei: This is fair, and I will adjust my morning tomorrow and comment back!

monty-sei: Alright this is now done as well!

x/evm/client/cli/tx_test.go
```diff
@@ -0,0 +1,104 @@
...
+		//Standard method call response from POST request to RPC
+		response := map[string]interface{}{
+			"jsonrpc": "2.0",
+			"id":      "send-cli",
+			"result":  result,
+		}
+
+		//Adjust to default GET response if not POST
+		if r.Method != http.MethodPost {
+			response = map[string]interface{}{
```
masih: Use `any` instead of `interface{}`. The go mod version in this repo is high enough to understand newer syntax. 

Please also add that to your AI prompt since most of the models I have come across are trained on go code old enough that they always use the old syntax.

Ditto for the rest of the PR.

x/evm/client/cli/tx_test.go
```diff
@@ -0,0 +1,104 @@
...
+	}{
+		{"new address", privateKey.PublicKey, "0x0", uint64(0)},
+		{"active address", privateKey.PublicKey, "0x5", uint64(5)},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(st *testing.T) {
+
+			//Setup RPC Server with result and get URL
+			rpcServer := getRPCServer(test.nonceHex)
```
masih: I recommend passing in `st` to assert no errors in writing response etc. from within the server. Right now we are ignoring errors returned by Json marshalling for example.

x/evm/client/cli/tx_test.go
```diff
@@ -0,0 +1,104 @@
...
+		chainId    int64
+		hasURL     bool
+	}{
+		{"mainnet chain id", "0x531", 1329, true},
+		{"testnet chain id", "0x530", 1328, true},
+		{"url error chain id", "", 0, false},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(st *testing.T) {
```
masih: Minor styling comment In table tests `t` is shadowed to avoid accidental use of the parent `t` by naming it to be the same name as parent, i.e. `t` instead of `st`.

x/evm/client/cli/tx_test.go
monty-sei: This is now done!

x/evm/client/cli/tx_test.go
monty-sei: This is now done!

x/evm/client/cli/tx_test.go
monty-sei: This is now done!

PR #2679, opened 2026-01-08
tcp multiplexer for sei giga

pompon0:
TCP Multiplexer for sei giga. It is a replacement for sei-tendermint/internal/p2p/conn.MConnection. It will be used with autobahn only (currently unused). The high level overview of the protocol is described in package comment in mux/mux.go


sei-tendermint/internal/mux/mux.go
```diff
@@ -0,0 +1,487 @@
...
+
+type Mux struct {
+	cfg   *Config
+	kinds map[StreamKind]*kindState
+	queue *utils.Watch[queue]
+}
+
+// NewMux constructs a new multipexer.
+// Remember to spawn Mux.Run() afterwards.
+func NewMux(cfg *Config) *Mux {
```
codchen, line 438: what's the input/output of this multiplexer?

codchen (approved):
logics themselves look good. Have a q about context

sei-tendermint/internal/mux/mux.go
pompon0: Streams of messages of fixed size. This will be used for p2p communication between nodes in giga (mainly consensus messages and block syncing). It is like a gRPC server/client, but with stricter semantics.

PR #2691, opened 2026-01-12
fix double refund

jewei1997:


go.work
```diff
@@ -9,3 +9,5 @@ use (
 	./sei-wasmd
 	./sei-wasmvm
 )
+
+replace github.com/ethereum/go-ethereum => github.com/sei-protocol/go-ethereum v1.15.7-sei-9.0.20260112140943-77a0432740be
```
masih, line 13: Instead of replace directive here, can I ask you to update the actual version in the go mods please?

masih: @jewei1997 we no longer need this replace directive now that the actual version in go.mod is updated. Right?

PR #2695, opened 2026-01-12
Upgrade to PebbleDB v2 + Add DefaultComparer Config Option

Kbhat1:
PebbleDB v2 Upgrade:
- Benchmarks show significant improvements in write/read throughput and latency
- Updated imports across sei-db/db_engine/pebbledb/ and sei-db/db_engine/pebbledb/mvcc/
- Added required v2 comparer functions (ComparePointSuffixes, CompareRangeSuffixes)
- Updated metrics API to use v2 field names

DefaultComparer Option:
- UseDefaultComparer config enables Pebble's default lexicographic comparer instead of custom MVCCComparer
- Only available programmatically via StateStoreConfig - intentionally NOT exposed in TOML
- Useful for modules like EVM where MVCC iteration isn't needed
- MVCC key encoding already uses BigEndian for versions, so ordering is compatible

- Added unit tests
- Tested on node to verify state sync + queries worked

### DIFF
--- a/sei-db/db_engine/rocksdb/mvcc/db_test.go
+++ b/sei-db/db_engine/rocksdb/mvcc/db_test.go
@@ -16,11 +16,13 @@ func TestStorageTestSuite(t *testing.T) {
 	rocksConfig := config.DefaultStateStoreConfig()
 	rocksConfig.Backend = "rocksdb"
 	s := &sstest.StorageTestSuite{
-		NewDB: func(dir string, config config.StateStoreConfig) (types.StateStore, error) {
-			return OpenDB(dir, config)
+		BaseStorageTestSuite: sstest.BaseStorageTestSuite{
+			NewDB: func(dir string, config config.StateStoreConfig) (types.StateStore, error) {
+				return OpenDB(dir, config)
+			},
+			Config:         rocksConfig,
+			EmptyBatchSize: 12,
 		},
-		Config:         rocksConfig,
-		EmptyBatchSize: 12,
 	}
 
 	suite.Run(t, s)


yzang2019 (approved):

sei-db/db_engine/pebbledb/mvcc/db.go
```diff
@@ -83,9 +83,20 @@ func OpenDB(dataDir string, config config.StateStoreConfig) (*Database, error) {
 	cache := pebble.NewCache(1024 * 1024 * 32)
 	defer cache.Unref()
 
+	// Select comparer based on config. Note: UseDefaultComparer is NOT backwards compatible
+	// with existing databases created with MVCCComparer - Pebble will refuse to open due to
+	// comparer name mismatch. Only use UseDefaultComparer for NEW databases.
```
blindchaser, line 90: if mismatch happens, is there a way we can show some errors to the user?

blindchaser: or the pebble already has the error inside?

Kbhat1: It will fail to open the db @blindchaser , so error will show up

Kbhat1: Will paste snippet of the error

sei-db/db_engine/pebbledb/mvcc/db_test.go
```diff
@@ -22,3 +22,49 @@ func TestStorageTestSuite(t *testing.T) {
...
+// TestStorageTestSuiteDefaultComparer runs the storage test suite with Pebble's DefaultComparer
+// instead of MVCCComparer. This is useful for new databases that don't need backwards compatibility.
+// Note: Iterator tests are skipped because DefaultComparer doesn't have the Split function
+// configured for MVCC key encoding, so NextPrefix/SeekLT operations won't work correctly.
+func TestStorageTestSuiteDefaultComparer(t *testing.T) {
+	pebbleConfig := config.DefaultStateStoreConfig()
+	pebbleConfig.Backend = "pebbledb"
+	pebbleConfig.UseDefaultComparer = true
+
+	// Skip all iterator-related tests since DefaultComparer doesn't support MVCC iteration properly
```
yzang2019: Why skipping all iterator related tests here? Can you help me understand why DefaultComparer doesn't support MVCC iteration?

Kbhat1: (Discussed in our sync with yiming) This is specifically because DefaultComparer treates everything as the key (there is no defined notion of prefix / suffix etc in the `Split` method). Iterations thus fails because its unable to properly do NextPrefix / PrevPrefix (these go to the immediate next key / prev key rather than the next prefix / next prefix).

sei-db/db_engine/pebbledb/mvcc/comparator.go
```diff
@@ -121,6 +121,25 @@ var MVCCComparer = &pebble.Comparer{
 		// there is a timestamp we prepend a 0 to the encoded timestamp data.
 		return len(key) + 1
 	},
+
+	// ComparePointSuffixes compares MVCC timestamps (suffixes) for v2.
+	ComparePointSuffixes: func(a, b []byte) int {
```
yzang2019: Why do we add these 2 new overrides?

yzang2019: Seems like they are required for v2, looks good

Kbhat1: Yup

sei-db/db_engine/pebbledb/mvcc/db_test.go
```diff
@@ -22,3 +22,49 @@ func TestStorageTestSuite(t *testing.T) {
 
 	suite.Run(t, s)
 }
+
+// TestStorageTestSuiteDefaultComparer runs the storage test suite with Pebble's DefaultComparer
+// instead of MVCCComparer. This is useful for new databases that don't need backwards compatibility.
+// Note: Iterator tests are skipped because DefaultComparer doesn't have the Split function
+// configured for MVCC key encoding, so NextPrefix/SeekLT operations won't work correctly.
+func TestStorageTestSuiteDefaultComparer(t *testing.T) {
```
Kbhat1, line 33: @masih likely a cleaner way to do this, but we want to eventually remove all iteration from state store. Right now its an artifact of cosmos (iteration) in v2

sei-db/state_db/ss/test/storage_test_suite.go
```diff
@@ -279,6 +279,9 @@ func (s *StorageTestSuite) TestApplyChangesetSyncAtomicAcrossModules() {
 }
 
 func (s *StorageTestSuite) TestDatabaseIteratorEmptyDomain() {
+	if slices.Contains(s.SkipTests, s.T().Name()) {
```
Kbhat1: @masih again, likely a cleaner way to do this, but we want to eventually remove all iteration from state store. Right now its an artifact of cosmos (iteration) in v2

sei-db/db_engine/pebbledb/mvcc/comparator.go
```diff
@@ -121,6 +121,25 @@ var MVCCComparer = &pebble.Comparer{
...
+		if len(a) == 0 {
+			if len(b) == 0 {
+				return 0
+			}
+			return -1
+		}
+		if len(b) == 0 {
+			return 1
+		}
```
masih: bytes compare already handles length and length check potimisation. unless I am missing something?

sei-db/db_engine/pebbledb/mvcc/comparator.go
```diff
@@ -121,6 +121,25 @@ var MVCCComparer = &pebble.Comparer{
...
+		return bytes.Compare(a, b)
+	},
+
+	// CompareRangeSuffixes compares suffixes for range keys (v2).
+	CompareRangeSuffixes: func(a, b []byte) int {
+		return bytes.Compare(a, b)
+	},
```
masih: I think it is clearer to set these to whatever the base comparer value is. Something like this:

```go
	ComparePointSuffixes: pebble.DefaultComparer.ComparePointSuffixes,
	CompareRangeSuffixes: pebble.DefaultComparer.CompareRangeSuffixes,
```

sei-db/db_engine/pebbledb/mvcc/db.go
```diff
@@ -83,15 +83,24 @@ func OpenDB(dataDir string, config config.StateStoreConfig) (*Database, error) {
 	cache := pebble.NewCache(1024 * 1024 * 32)
```
masih, line 85: Unrelated to this PR: we probably should bump this value. It seems small for `seid`?

sei-db/db_engine/pebbledb/mvcc/db.go
```diff
@@ -102,17 +111,10 @@ func OpenDB(dataDir string, config config.StateStoreConfig) (*Database, error) {
 		l.IndexBlockSize = 256 << 10 // 256 KB
 		l.FilterPolicy = bloom.FilterPolicy(10)
 		l.FilterType = pebble.TableFilter
-		// TODO: Consider compression only for specific layers like bottommost
-		l.Compression = pebble.ZstdCompression
```
masih, line 107: Hm, why is this removed? This would be a breaking change regardless of comparator, right?

Side note: the default compression if unspecified is Snappy.

sei-db/db_engine/pebbledb/mvcc/db.go
```diff
@@ -102,17 +111,10 @@ func OpenDB(dataDir string, config config.StateStoreConfig) (*Database, error) {
...
-		if i > 0 {
-			l.TargetFileSize = opts.Levels[i-1].TargetFileSize * 2
-		}
-		l.EnsureDefaults()
```
masih, line 111: Do we need `l.EnsureL1PlusDefaults(&opts.Levels[i-1])` similar to wht [cockroachdb](https://github.com/cockroachdb/cockroach/blob/6bf0f9778f34df3d197c28c7f4b2d1a4a27c3288/pkg/storage/pebble.go#L704) is doing?

Also check table filter policy.

sei-db/db_engine/pebbledb/mvcc/db.go
```diff
@@ -102,17 +111,10 @@ func OpenDB(dataDir string, config config.StateStoreConfig) (*Database, error) {
...
 	}
 
 	opts.Levels[6].FilterPolicy = nil
-	opts.FlushSplitBytes = opts.Levels[0].TargetFileSize
-	opts = opts.EnsureDefaults()
+	opts.EnsureDefaults()
```
masih: This is a redundant call. Pebble does this for any given set of options at the time of instantiation.

sei-db/db_engine/pebbledb/mvcc/db.go
```diff
@@ -83,15 +83,24 @@ func OpenDB(dataDir string, config config.StateStoreConfig) (*Database, error) {
...
+	} else {
+		// TODO: Delete once we remove support
+		comparer = MVCCComparer
+	}
+
 	opts := &pebble.Options{
 		Cache:                       cache,
-		Comparer:                    MVCCComparer,
+		Comparer:                    comparer,
 		FormatMajorVersion:          pebble.FormatNewest,
```
masih: Have you checked if setting this to an alternative value would avoid breaking persistence format? 

We might want to hardcode it to what it _currently_ is rather than implicitly picking the newest format. Because, doing so would avoid accidental breaking changes on dependency updates.

sei-db/db_engine/pebbledb/mvcc/db_test.go
```diff
@@ -22,3 +22,49 @@ func TestStorageTestSuite(t *testing.T) {
...
+		"TestStorageTestSuiteDefaultComparer/TestDatabaseBugInitialReverseIteration",
+		"TestStorageTestSuiteDefaultComparer/TestDatabaseBugInitialReverseIterationHigher",
+		"TestStorageTestSuiteDefaultComparer/TestDatabaseParallelDeleteIteration",
+		"TestStorageTestSuiteDefaultComparer/TestDatabaseParallelIterationVersions",
+		"TestStorageTestSuiteDefaultComparer/TestParallelIterationAndPruning",
+		// Prune tests also use iteration internally
+		"TestStorageTestSuiteDefaultComparer/TestDatabasePrune",
+		"TestStorageTestSuiteDefaultComparer/TestDatabasePruneKeepRecent",
+		"TestStorageTestSuiteDefaultComparer/TestParallelWriteAndPruning",
+	}
```
masih: This is unfortunately error-prone, and can silently result in false negative test failures if we rename any of the tests.

My advice would be to avoid this. I see 2 options:
1.  fork the test sets that we want to have going forward for the default comparator. Then when the MVCC comparator retires remove entire tests for it.

2.  break up the test suits into 2, and have one extend the other where the extended suit runs the current superset of tests for MVCC. This solves your concern about code drift and would avoid the kind of config we see here as well as conditional skip tests. Let me know if you have questions on how to extend the suit while avoiding duplicate code

Kbhat1: Yea I think for (1) I was concerned about the forks not matching. (2) seems reasonable, let me think on this though if there is a more efficient approach

sei-db/db_engine/pebbledb/mvcc/db.go
Kbhat1: During testing I swapped out v1 written db with v2 opendb, but these configs were already removed. Will add this back, I was testing creating in v1 -> opening in v2 and resuming which worked but thats with that layer already removed). There doesn't seem like there is a great way in the code to test this (more of an upgrade?)

sei-db/db_engine/pebbledb/mvcc/db.go
Kbhat1: Its by default set to the MVCCComparer and that is not possible to switch in the app.toml @masih . right now there is NO way the user can specify this, this is is fully in our control

sei-db/db_engine/pebbledb/mvcc/db.go
Kbhat1: Updated

sei-db/db_engine/pebbledb/mvcc/db.go
Kbhat1: Updated the FormatMajorVersion to `pebble.FormatVirtualSSTables` => left a comment on why we did this @masih

sei-db/db_engine/pebbledb/mvcc/comparator.go
Kbhat1: Updated

sei-db/db_engine/pebbledb/mvcc/db.go
Kbhat1: Updated

PR #2701, opened 2026-01-14
[STO-308] New receiptDB receipt-specific interface

jewei1997:

This PR converts the current ReceiptDB interface from StateStore interface to using a receipt-specific interface. In a future PR, we will add in the new Parquet implementation for receipts.

unit tests.

### DIFF
--- a/.github/workflows/forge-test.yml
+++ b/.github/workflows/forge-test.yml
@@ -25,7 +25,12 @@ jobs:
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
         with:
-          version: nightly
+          # Pin to stable for predictable CI. If tests fail after a Foundry update,
+          # either fix compatibility or pin to a specific version (e.g., 1.5.1).
+          # See:
+          #  - https://github.com/foundry-rs/foundry/releases
+          #  - https://github.com/foundry-rs/foundry-toolchain
+          version: stable
 
       - name: Run Forge build
         run: |


sei-db/ledger_db/receipt/receipt_store.go
```diff
@@ -0,0 +1,496 @@
...
+}
+
+type receiptStore struct {
+	db          *mvcc.Database
+	storeKey    sdk.StoreKey
+	stopPruning chan struct{}
+	closeOnce   sync.Once
+}
+
+func NewReceiptStore(log dbLogger.Logger, config dbconfig.StateStoreConfig, storeKey sdk.StoreKey) (ReceiptStore, error) {
```
yzang2019: Would recommend we create a ReceiptStoreConfig instead of reusing the StateStoreConfig

sei-db/ledger_db/receipt/receipt_store.go
```diff
@@ -0,0 +1,496 @@
...
+	}
+	return &r, nil
+}
+
+func (s *receiptStore) SetReceipts(ctx sdk.Context, receipts []ReceiptRecord) error {
+	if s == nil || s.db == nil {
+		return ErrNotConfigured
+	}
+
+	pairs := make([]*iavl.KVPair, 0, len(receipts))
```
yzang2019, line 175: How do we plan to add Parquet based backend logic in the future? These logic are all specific to Pebble backend now, do we plan to do another refactor in the next PR to wrap these logic into a separate function?

evmrpc/filter.go
```diff
@@ -783,7 +783,7 @@ func (f *LogFetcher) GetLogsByFilters(ctx context.Context, crit filters.FilterCr
 		localLogs := f.globalLogSlicePool.Get()
 
 		for _, block := range batch {
-			f.GetLogsForBlockPooled(block, crit, bloomIndexes, &localLogs)
+			f.GetLogsForBlockPooled(block, crit, &localLogs)
```
yzang2019, line 786: Can we explain why we remove  the bloomIndexes here? Would this impact correctness?

jewei1997: we calculate it inside FilterLogs from the filter criteria but it doesn't make sense to include it in the interface.

sei-db/ledger_db/receipt/receipt_store.go
jewei1997: will have parquet implementation implement the same interface

yzang2019 (approved):

PR #2702, opened 2026-01-14
Fix mac local cluster

stevenlanders:
- uses arm64 for macs 

- local mac testing, CI


yzang2019 (approved):
Nice!!

docker/docker-compose.yml
```diff
@@ -57,7 +57,7 @@ services:
         ipv4_address: 192.168.10.11
 
   node2:
-    platform: linux/amd64
```
pompon0, line 60: nit: have you checked, if specifying platform is needed at all? It should be deduced from the image, no?

PR #2704, opened 2026-01-15
Allow unsafe legacy tx when tracing txs before v6.3.0

codchen:
We will disallow unsafe tx starting v6.3.0, but unsafe txs that were sent before v6.3.0 should still be traceable

tested on rpc node

### DIFF
--- a/x/evm/ante/preprocess.go
+++ b/x/evm/ante/preprocess.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"strings"
 
 	"github.com/sei-protocol/sei-chain/utils/helpers"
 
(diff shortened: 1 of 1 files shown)


x/evm/ante/preprocess.go
```diff
@@ -177,7 +178,7 @@ func PreprocessUnpacked(ctx sdk.Context, msgEVMTransaction *evmtypes.MsgEVMTrans
 		V = AdjustV(V, ethTx.Type(), ethCfg.ChainID)
 		txHash = signer.Hash(ethTx)
 	} else {
-		if isBlockTest {
+		if isBlockTest || (ctx.IsTracing() && strings.Compare(ctx.ClosestUpgradeName(), "v6.3.0") < 0) {
```
masih, line 181: Beyond a scope of this PR and a prexisting pattern: I have [raised this issue](https://sei-network-io.slack.com/archives/C06A3D4FWE5/p1767619404649919?thread_ts=1767605984.651449&cid=C06A3D4FWE5) in previous reviews: if we ever hit double digits major/minor `strings.Compare` will become incorrect since it sorts lexicographically.

masih (approved):

PR #2705, opened 2026-01-15
Bootstrap `evmone` integration with build tags

pdrobnjak:
## Summary

- Add evmone library loading infrastructure using Go build tags for platform-specific library selection
- Include pre-built evmone 0.12.0 binaries for darwin/arm64 and linux/amd64
- Load evmone at app startup to avoid first-block latency

### New `giga/executor/lib` package
- `evmlib.go`: `InitEvmoneVM()` function that loads evmone and verifies version compatibility
- `lib_darwin_arm64.go`, `lib_linux_amd64.go`: Build tag files defining platform-specific library names
- Pre-built evmone 0.12.0 binaries (~300KB darwin, ~490KB linux)

### App integration
- `EvmoneVM` field added to `EvmKeeper` to store pre-loaded VM
- `evmone` loaded at app initialization when GigaExecutor is enabled


pdrobnjak (commented):
## Code Review

### Overall
Clean implementation following the pattern from sei-v3. Good use of build tags for platform selection and runtime version verification.

### Strengths
- **Version safety**: The `runtime/debug.ReadBuildInfo()` approach to infer expected evmone version from EVMC dependency is elegant and prevents silent version mismatches
- **No configuration required**: Build tags handle platform selection at compile time
- **Startup loading**: Moving VM initialization to app startup avoids first-block latency

### Questions / Suggestions

1. **`evmlib.go:17-20`** - `runtime.Caller(0)` returns the path of the source file at compile time. This works when running from source, but will this work correctly in a deployed binary? The path will be baked into the binary. Consider adding a test or documenting this behavior.

2. **`evmlib.go:56-57`** - `debug.ReadBuildInfo()` returns `false` when the binary wasn't built with module support or in certain edge cases. The error message is good, but worth noting in docs that this requires Go modules.

3. **`app/app.go:682-685`** - Using `panic()` on load failure is appropriate for startup, but consider whether we want to allow the node to start with GigaExecutor disabled if evmone fails to load (graceful degradation vs fail-fast). Current approach (fail-fast) seems correct for now.

4. **Missing platforms** - Only darwin/arm64 and linux/amd64 are supported. Should we add a `lib_unsupported.go` with a build constraint that provides a clear compile-time error for unsupported platforms? Currently it will fail with "undefined: libName".

5. **Binary verification** - The binaries are downloaded from GitHub releases. For production, consider documenting how to verify the binaries haven't been tampered with (checksums are verified during download but not for committed binaries).

### Minor
- Bootstrap script is well-structured with good error handling
- Library naming convention is consistent and includes version for clarity

giga/executor/lib/evmlib.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/ethereum/evmc/v12/bindings/go/evmc"
+)
+
+// InitVM initializes the EVMC VM by loading the platform-specific evmone library.
+// It verifies that the loaded VM version is compatible with the EVMC bindings version.
+func InitVM() (*evmc.VM, error) {
+	_, path, _, ok := runtime.Caller(0)
+	if !ok {
+		return nil, fmt.Errorf("failed to get caller information")
```
pdrobnjak, line 16: Using `runtime.Caller(0)` returns the source file path at compile time. This path gets baked into the binary. Will this resolve correctly in deployed binaries where the source tree does not exist?

Consider adding a comment explaining this relies on the library files being placed relative to the Go source files at build time.

giga/executor/lib/evmlib.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	return nil
+}
+
+// getExpectedEvmoneVersionPrefix derives the expected evmone version prefix from the EVMC module version.
+// EVMC v12.x.x -> evmone 0.12
+// EVMC v11.x.x -> evmone 0.11
+func getExpectedEvmoneVersionPrefix() (string, error) {
+	info, ok := debug.ReadBuildInfo()
```
pdrobnjak: `debug.ReadBuildInfo()` can return `false` when the binary was not built with module support. Consider documenting this requirement, or adding a fallback/clearer error message for this edge case.

app/app.go
```diff
@@ -678,6 +679,12 @@ func New(
 	app.EvmKeeper.GigaExecutorEnabled = gigaExecutorConfig.Enabled
 	app.EvmKeeper.GigaOCCEnabled = gigaExecutorConfig.OCCEnabled
 	if gigaExecutorConfig.Enabled {
+		// Load evmone VM at startup
+		evmoneVM, err := gigalib.InitVM()
+		if err != nil {
```
pdrobnjak, line 684: Using `panic()` here is a fail-fast approach which seems appropriate for startup. However, should we consider logging a warning and disabling GigaExecutor instead (graceful degradation)? This would allow the node to start with the default geth interpreter if evmone fails to load.

Current approach is fine if we want to enforce evmone availability when GigaExecutor is enabled in config.

giga/executor/lib/lib_linux_amd64.go
```diff
@@ -0,0 +1,5 @@
+//go:build linux && amd64
+
+package lib
+
+const libName = "libevmone.0.12.0_linux_amd64.so"
```
pdrobnjak, line 5: Consider adding a `lib_unsupported.go` file with a build constraint for other platforms that provides a compile-time error with a clear message. Currently, building on an unsupported platform will fail with `undefined: libName` which is not very descriptive.

scripts/bootstrap-evmone.sh
```diff
@@ -0,0 +1,107 @@
...
+
+# Bootstrap script for evmone
+# Downloads pre-built evmone artifacts from GitHub releases
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+LIB_DIR="${PROJECT_ROOT}/giga/executor/lib"
+
+# evmone v0.12.0 is compatible with EVMC v12.x
+EVMONE_VERSION="${EVMONE_VERSION:-0.12.0}"
```
pdrobnjak: nit: Consider adding a comment noting the version coupling: evmone 0.12.x corresponds to EVMC v12.x. This will help future maintainers understand why specific versions are used.

giga/executor/lib/evmlib.go
pdrobnjak: That's good enough for now as we'll be keeping library files relative to the Go source files.

app/app.go
pdrobnjak: Not for now, we are using `giga` only in non-production flows and `evmone` is a requirement for `giga` throughput.

giga/executor/lib/lib_linux_amd64.go
pdrobnjak: Done.

arajasek (approved):

PR #2707, opened 2026-01-15
Add Ethereum state test runner for Giga executor validation

pdrobnjak:
## Summary
- Add comprehensive test runner comparing Giga (evmone) vs V2 (Geth) execution using Ethereum's official GeneralStateTests
- Include 9MB compressed test fixtures that auto-extract on first run
- Add debugging mode to isolate Giga executor from GigaKVStore layer

---

## Important: Non-Test Changes

> **Reviewers**: This PR includes 3 significant changes outside the test infrastructure that require attention.

### 1. GigaBankKeeper Type Change (`app/app.go`)
Changed `GigaBankKeeper` from interface type to concrete pointer:
```go
// Before
GigaBankKeeper   gigabankkeeper.Keeper
// After
GigaBankKeeper   *gigabankkeeper.BaseKeeper
```
**Why**: Needed to expose `UseRegularStore` field for debugging mode.

### 2. CacheKV Store Bug Fix (`giga/deps/store/cachekv.go`)
Fixed cache behavior in `Write()` method:
```go
// Before: cleared the cache entirely
store.cache = &sync.Map{}

// After: marks entries as clean instead of clearing
store.cache.Range(func(key, value any) bool {
    cv := value.(*types.CValue)
    store.cache.Store(key, types.NewCValue(cv.Value(), false))
    return true
})
```
**Why**: The parent store (`commitment.Store`) doesn't make writes visible until `Commit()`. By keeping cache entries (marked clean), subsequent reads still hit the cache instead of falling through to the parent which can't see uncommitted data.

### 3. Test Helper for Debugging (`app/test_helpers.go`)
Added `NewGigaTestWrapperWithRegularStore()` function that runs Giga executor with regular KVStore instead of GigaKVStore, isolating executor logic from store layer issues.

### Related Test Fixes (`giga/deps/xbank/keeper/`)
Fixed type mismatch errors in `deferred_cache_test.go` and `keeper_test.go` caused by the GigaBankKeeper type change above.

---

### New Components
- `TestGigaVsV2_StateTests`: Runs Ethereum state tests through both executors and compares results
- Test harness with types, loader, builder, and skip list support
- Auto-extracts test fixtures from embedded `.tgz` archive

### Test Fixtures
- Embedded `fixtures_general_state_tests.tgz` (9MB, 277MB uncompressed)
- Full Ethereum GeneralStateTests coverage
- `.gitignore` for extracted directory
- Detailed skip list with per-category analysis

## Usage

```bash
# Run tests for a specific category
STATE_TEST_DIR=stChainId go test -v -run TestGigaVsV2_StateTests ./giga/tests/...

# Run with regular store (debugging mode)
USE_REGULAR_STORE=true STATE_TEST_DIR=stChainId go test -v -run TestGigaVsV2_StateTests ./giga/tests/...
```

## Test plan
- [x] Run `TestGigaVsV2_StateTests` with various categories
- [x] Verify auto-extraction from archive works
- [x] Verify skip list filtering works correctly

### DIFF
--- a/.github/workflows/uci-backport.yml
+++ b/.github/workflows/uci-backport.yml
@@ -1,4 +1,5 @@
-name: UCI / Backport
+name: UCI
+run-name: UCI / Backport
 
 on:
   pull_request_target:
@@ -14,4 +15,6 @@ permissions:
 
 jobs:
   backport:
-    uses: sei-protocol/uci/.github/workflows/backport.yml@v0.0.4
+    name: Backport
+    uses: sei-protocol/uci/.github/workflows/backport.yml@v0.0.6
+    secrets: inherit
--- a/.github/workflows/uci-go-lint.yml
+++ b/.github/workflows/uci-go-lint.yml
@@ -1,4 +1,5 @@
 name: UCI
+run-name: UCI / Go Lint
 
 on:
   pull_request:
@@ -18,4 +19,4 @@ concurrency:
 jobs:
   go-lint:
     name: Go
-    uses: sei-protocol/uci/.github/workflows/go-lint.yml@v0.0.1
+    uses: sei-protocol/uci/.github/workflows/go-lint.yml@v0.0.6
--- a/.github/workflows/uci-release-check.yml
+++ b/.github/workflows/uci-release-check.yml
@@ -1,4 +1,5 @@
 name: UCI
+run-name: UCI / Release Check
 
 on:
   pull_request_target:
@@ -17,4 +18,4 @@ concurrency:
 jobs:
   release-check:
     name: Release
-    uses: sei-protocol/uci/.github/workflows/release-check.yml@v0.0.2
+    uses: sei-protocol/uci/.github/workflows/release-check.yml@v0.0.6
--- a/.github/workflows/uci-release-publish.yml
+++ b/.github/workflows/uci-release-publish.yml
@@ -1,4 +1,5 @@
 name: UCI
+run-name: UCI / Release Publish
 
 on:
   push:
@@ -15,4 +16,4 @@ concurrency:
 jobs:
   releaser:
     name: Release
-    uses: sei-protocol/uci/.github/workflows/release-publish.yml@v0.0.2
+    uses: sei-protocol/uci/.github/workflows/release-publish.yml@v0.0.6
--- a/.github/workflows/uci-stale-check.yml
+++ b/.github/workflows/uci-stale-check.yml
@@ -1,4 +1,5 @@
 name: UCI
+run-name: UCI / Stale Check
 
 on:
   workflow_dispatch:
@@ -18,6 +19,6 @@ jobs:
     permissions:
       issues: write
       pull-requests: write
-    uses: sei-protocol/uci/.github/workflows/stale-check.yml@v0.0.3
+    uses: sei-protocol/uci/.github/workflows/stale-check.yml@v0.0.6
     with: 
       days-before-pr-stale: 28
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,119 @@ Ref: https://keepachangelog.com/en/1.0.0/
 -->
 
 # Changelog
+## v6.3
+sei-chain (Note: major repos have been merged into sei-chain)
+* [#2580](https://github.com/sei-protocol/sei-chain/pull/2580) Fix: enforce EIP-6780 selfdestruct for prefunded addresses
+* [#2572](https://github.com/sei-protocol/sei-chain/pull/2572) Extra checks in BitArray methods
+* [#2570](https://github.com/sei-protocol/sei-chain/pull/2570) Strongly typed p2p channels
+* [#2567](https://github.com/sei-protocol/sei-chain/pull/2567) Migrate sei-ibc-go into sei-chain as monorepo
+* [#2563](https://github.com/sei-protocol/sei-chain/pull/2563) Do not return error string on precompile error
+* [#2561](https://github.com/sei-protocol/sei-chain/pull/2561) Make seid rollback idempotent and remove --hard
+* [#2560](https://github.com/sei-protocol/sei-chain/pull/2560) Fix: Resolve data race in parallel snapshot writing
+* [#2558](https://github.com/sei-protocol/sei-chain/pull/2559) Remove custom json encoding of consensus internals and replay command
+* [#2558](https://github.com/sei-protocol/sei-chain/pull/2558) Refactor of consensus reactor task management
+* [#2553](https://github.com/sei-protocol/sei-chain/pull/2553) Refactor CheckTx
+* [#2547](https://github.com/sei-protocol/sei-chain/pull/2547) Deprecate and clean up dbsync code reference
+* [#2543](https://github.com/sei-protocol/sei-chain/pull/2543) Add a benchmark mode
+* [#2542](https://github.com/sei-protocol/sei-chain/pull/2542) Config: Make worker pool configurable and increase default queue size
+* [#2540](https://github.com/sei-protocol/sei-chain/pull/2540) Streamline EndBlock
+* [#2539](https://github.com/sei-protocol/sei-chain/pull/2539) PeerManager rewrite
+* [#2537](https://github.com/sei-protocol/sei-chain/pull/2537) Optimzation: Reduce snapshot creation time
+* [#2534](https://github.com/sei-protocol/sei-chain/pull/2534) Remove ABCI socket/grpc functionality
+* [#2533](https://github.com/sei-protocol/sei-chain/pull/2533) Migrate transaction embedding proto types to Go types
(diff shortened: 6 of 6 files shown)


giga/executor/internal/host_context.go
```diff
@@ -287,3 +289,18 @@ func (h *HostContext) getEVMRevision() evmc.Revision {
...
+func toEvmcError(err error) error {
+	if err == nil {
+		return nil
+	}
+	if _, ok := err.(evmc.Error); ok {
+		return err
+	}
+	if errors.Is(err, vm.ErrExecutionReverted) {
+		return evmc.Revert
+	}
```
pdrobnjak, line 396: **[CRITICAL]** Incomplete Error Mapping

The function only handles `vm.ErrExecutionReverted` and maps all other errors to `evmc.Failure`. This loses important error context.

**Missing mappings:**
- `vm.ErrOutOfGas` → `evmc.OutOfGas`
- `vm.ErrInsufficientBalance` → specific handling
- `vm.ErrDepth` → stack depth exceeded

**Suggestion:**
```go
switch {
case errors.Is(err, vm.ErrExecutionReverted):
    return evmc.Revert
case errors.Is(err, vm.ErrOutOfGas):
    return evmc.OutOfGas
default:
    return evmc.Failure
}
```

giga/tests/state_test.go
```diff
@@ -0,0 +1,582 @@
...
+					break
+				}
+			}
+
+			for i, post := range cancunPosts {
+				subtestName := testName
+				if len(cancunPosts) > 1 {
+					subtestName = testName + "/" + string(rune('0'+i))
+				}
+
```
pdrobnjak: **[SUGGESTION]** Subtest Naming Bug

For indices >= 10, `string(rune('0'+i))` produces non-numeric characters (e.g., ':' for i=10, ';' for i=11).

**Suggestion:** Use `fmt.Sprintf` for subtest naming:
```go
subtestName = fmt.Sprintf("%s/%d", testName, i)
```

giga/tests/state_test.go
```diff
@@ -0,0 +1,582 @@
...
+func parseHexBig(s string) *big.Int {
+	if s == "" {
+		return new(big.Int)
+	}
+	s = strings.TrimPrefix(s, "0x")
+	s = strings.TrimPrefix(s, "0X")
+	result, ok := new(big.Int).SetString(s, 16)
+	if !ok {
+		return new(big.Int)
+	}
```
pdrobnjak, line 96: **[SUGGESTION]** Silent Parse Failures

Parse failures silently return zero, which could mask test data issues.

```go
result, ok := new(big.Int).SetString(s, 16)
if !ok {
    return new(big.Int)  // Silently returns 0 on parse failure
}
```

**Recommendation:** Either log a warning or return an error when parsing fails to avoid masking issues in test fixtures.

giga/tests/harness/loader.go
```diff
@@ -0,0 +1,117 @@
...
+	"sync"
+)
+
+// extractOnce ensures we only extract the archive once per test run
+var extractOnce sync.Once
+
+// LoadStateTest loads a state test from a JSON file
+func LoadStateTest(filePath string) (map[string]*StateTestJSON, error) {
+	file, err := os.Open(filepath.Clean(filePath))
+	defer file.Close()
```
pdrobnjak: The `defer file.Close()` is placed before the error check. If `os.Open` fails, `file` is nil. While Go's `os.File.Close()` handles nil gracefully, this violates idiomatic Go patterns.

Suggest:
```go
file, err := os.Open(filepath.Clean(filePath))
if err != nil {
    return nil, err
}
defer file.Close()
```

giga/tests/harness/loader.go
```diff
@@ -0,0 +1,117 @@
...
+	extractOnce.Do(func() {
+		err = extractArchive(archivePath, "data")
+	})
+	if err != nil {
+		return "", err
+	}
+
+	_, err = os.Stat(dataPath)
+	if err != nil {
+		return "", nil
```
pdrobnjak: Returns `("", nil)` when `os.Stat` fails - an empty path with no error. This could cause confusing downstream errors.

Consider returning the actual error:
```go
if err != nil {
    return "", err
}
```

giga/tests/harness/loader.go
```diff
@@ -0,0 +1,117 @@
...
+		return dataPath, nil
+	}
+
+	archivePath := filepath.Join("data", "fixtures_general_state_tests.tgz")
+	_, err := os.Stat(archivePath)
+	if err != nil {
+		return "", err
+	}
+
+	extractOnce.Do(func() {
```
pdrobnjak, line 95: `sync.Once` captures `err` via closure, but if extraction fails on first call, subsequent calls won't retry and will see a stale/nil `err`. Consider verifying the directory exists after `sync.Once` completes, regardless of whether it was the first invocation.

giga/tests/harness/skip.go
```diff
@@ -0,0 +1,145 @@
...
+	return false
+}
+
+// GetSkipReason returns the skip reason for a test, or empty string if not skipped
+func (sl *SkipList) GetSkipReason(category, testName string) string {
+	_, reason := sl.ShouldSkip(category, testName)
+	return reason
+}
+
+// NormalizeTestName normalizes a test name by removing leading directory components
```
pdrobnjak: `NormalizeTestName` function appears to be unused. Consider removing dead code or adding a comment if it's intended for future use.

giga/tests/state_test.go
```diff
@@ -0,0 +1,195 @@
...
+		sort.Slice(types, func(i, j int) bool {
+			return string(types[i]) < string(types[j])
+		})
+
+		for _, ft := range types {
+			sb.WriteString(fmt.Sprintf("  %s: %d\n", ft, len(ts.FailuresByType[ft])))
+		}
+		sb.WriteString("\n")
+
+		// List failed tests (limit to first 50 to avoid huge output)
```
pdrobnjak: Consider extracting `50` to a named constant for clarity:
```go
const maxFailuresToDisplay = 50
```

giga/tests/harness/loader.go
pdrobnjak: Fixed in c45a914b0 - now returns the actual error instead of `("", nil)`

giga/tests/harness/loader.go
pdrobnjak: Fixed in c45a914b0 - now uses dedicated `extractErr` variable to avoid closure capture issue, and always verifies directory exists after sync.Once completes

giga/tests/harness/skip.go
pdrobnjak: Fixed in c45a914b0 - removed the unused function

giga/tests/state_test.go
pdrobnjak: Fixed in c45a914b0 - extracted to `maxFailuresToDisplay` constant

giga/tests/harness/loader.go
```diff
@@ -0,0 +1,106 @@
...
+	err = decoder.Decode(&tests)
+	if err != nil {
+		return nil, err
+	}
+
+	result := make(map[string]*StateTestJSON)
+	for name, test := range tests {
+		testCopy := test
+		result[name] = &testCopy
+	}
```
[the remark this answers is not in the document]
pdrobnjak: Not addressing - this is a false positive. The code collects map keys into a slice (line 33), sorts them (line 36), then iterates over the sorted slice (line 37). This is the correct pattern for deterministic iteration.

giga/tests/harness/loader.go
```diff
@@ -0,0 +1,106 @@
...
+		if err != nil {
+			return err
+		}
+
+		for name, test := range tests {
+			// Use relative path + test name as key to avoid collisions
+			relPath, _ := filepath.Rel(dirPath, path)
+			fullName := filepath.Join(relPath, name)
+			result[fullName] = test
+		}
```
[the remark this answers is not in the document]
pdrobnjak: Not addressing - this is a false positive. The code collects map keys into a slice (line 61), sorts them (line 65), then iterates over the sorted slice (line 66). This is the correct pattern for deterministic iteration.

giga/tests/harness/skip.go
```diff
@@ -0,0 +1,145 @@
...
+		if len(parts) == 2 && parts[0] == category {
+			shortName := parts[1]
+			// Match if the test name starts with the short name (before any .json or /)
+			if strings.HasPrefix(testName, shortName+".json") ||
+				strings.HasPrefix(testName, shortName+"/") ||
+				testName == shortName {
+				return true, r
+			}
+		}
+	}
```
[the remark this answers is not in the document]
pdrobnjak: Not addressing - the iteration order for pattern matching has minimal practical impact. Skip patterns should be specific enough not to overlap, and the function returns on first match regardless of order.

stevenlanders (approved):

PR #2670, opened 2026-01-07
Add failfast precompile to detect interop

codchen:
## Describe your changes and provide context
Added a failfast precompile to be registered on the giga executor at v2's custom precompile range, so that custom precompile calls can be intercepted and fall back to v2 executor.

## Testing performed to validate your change
to be tested when the executor is ready to wire up

PR #2708, opened 2026-01-15
feat: Giga RPC node (sequential execution)

arajasek:

This PR supports running a Giga node, in consensus with other non-Giga nodes. It only supports sequential execution for Giga, falling back to the v2 logic for cosmos and interop transactions.

TODO:
- [x] Land #2670 
- [x] Land https://github.com/sei-protocol/go-ethereum/pull/78 

- tests pass
- devnet runs with 1/4 nodes using giga, and both evm and cosmos txs hurled at the chain


app/app.go
```diff
@@ -1340,6 +1344,43 @@ func (app *App) ProcessBlockSynchronous(ctx sdk.Context, txs [][]byte, typedTxs
...
+				txResults[i] = res
+				continue
+			}
+			txResults[i] = &abci.ExecTxResult{
+				Code: 1,
+				Log:  fmt.Sprintf("[BUG] giga executor error: %v", execErr),
+			}
+			continue
+		}
+
```
codchen, line 1391: would need to call `ctx.GigaMultiStore().WriteGiga()` here since giga cache isn't visible to v2 executor so need flushing

arajasek: Done, thank you!

PR #2710, opened 2026-01-15
feat(flatkv): introduce interface layer (Store/Iterator) and typed keys

blindchaser:

This PR introduces the FlatKV interface layer, designed to bridge x/evm state access with a multi-DB storage architecture:

- Store Interface: Defines a unified contract for EVM state storage that internally maps logical keys to isolated physical DBs (accounts, code, storage).

- Type-Safe Keys: Adds strong types (Address, CodeHash, AccountKey, etc.) and constructors to prevent raw byte manipulation errors.

- Key Mapping: Implements ParseEVMKey to robustly translate x/evm memiavl keys into internal DB selectors.

- Iterator Semantics: Adopts PebbleDB-style iterator semantics (Next/Prev/SeekGE) with explicit support for multiplexed iteration across the multi-DB layout.


sei-db/state_db/sc/flatkv/keys.go
```diff
@@ -0,0 +1,123 @@
...
+func (k Key) bytes() []byte { return k.b }
+
+// Keyspace prefixes (binary layout, stable across versions).
+//
+//	account: 0x01 || addr(20) || field(1)
+//	code:    0x02 || codehash(32)
+//	storage: 0x03 || addr(20) || slotLen(u16be) || slot
+//	raw:     0x7f || keyLen(u32be) || originalKey
+const (
+	pfxAccount byte = 0x01
```
yzang2019: This might be related to how we wanna split the DB

sei-db/state_db/sc/flatkv/keys.go
```diff
@@ -0,0 +1,123 @@
...
+	pfxCode    byte = 0x02
+	pfxStorage byte = 0x03
+	// pfxRaw is a fallback namespace for keys FlatKV does not model explicitly.
+	// It allows storing unknown/original EVM KV entries without data loss.
+	pfxRaw byte = 0x7f
```
yzang2019: Not sure if we want this, can add when we actually feel we need it, otherwise recommend to remove it

sei-db/state_db/sc/flatkv/keys.go
```diff
@@ -0,0 +1,123 @@
...
+	return keyFromBytes(k)
+}
+
+// StorageKey returns the key for a contract storage slot.
+func StorageKey(addr []byte, slot []byte) Key {
+	k := make([]byte, 0, 1+len(addr)+2+len(slot))
+	k = append(k, pfxStorage)
+	k = append(k, addr...)
+	var sz [2]byte
+	binary.BigEndian.PutUint16(sz[:], uint16(len(slot))) //nolint:gosec
```
yzang2019: Why append a length? Slot should be fixed length, I think it's a fixed length (32 bytes)

sei-db/state_db/sc/flatkv/keys.go
```diff
@@ -0,0 +1,123 @@
...
+	k = append(k, sz[:]...)
+	k = append(k, slot...)
+	return keyFromBytes(k)
+}
+
+// RawKey wraps an unrecognized key (fallback for unknown prefixes).
+func RawKey(original []byte) Key {
```
yzang2019: What would be the usecase for this?

sei-db/state_db/sc/flatkv/keys.go
blindchaser: make sense

sei-db/state_db/sc/flatkv/keys.go
blindchaser: i will add it back when we actually feel we need it

sei-db/state_db/sc/flatkv/keys.go
```diff
@@ -0,0 +1,156 @@
...
+	"bytes"
+	"encoding/binary"
+	"fmt"
+)
+
+const (
+	AddressLen  = 20
+	CodeHashLen = 32
+	SlotLen     = 32
+	WordLen     = 32
```
yzang2019: It seems this is specifically for Balance, shall we just name it BalanceLen to make it easier to understand?

sei-db/state_db/sc/flatkv/keys.go
```diff
@@ -0,0 +1,156 @@
...
+	var nonce [NonceLen]byte
+	binary.BigEndian.PutUint64(nonce[:], v.Nonce)
+	b = append(b, nonce[:]...)
+	b = append(b, v.CodeHash[:]...)
+	return b
+}
+
+// DecodeAccountValue decodes a fixed-length account record.
+func DecodeAccountValue(b []byte) (AccountValue, error) {
+	if len(b) != accountValueEncodedLen {
```
yzang2019: Most Accounts might not have a code hash, do we really want to waste the extra 32 bytes allocated for these accounts? What would the trade off be here to make all account value fixed length vs have different account length for regular account vs contract account

sei-db/state_db/sc/flatkv/evm_memiavl_mapper.go
```diff
@@ -0,0 +1,95 @@
...
+	if len(key) == 0 {
+		return EVMKeyUnknown, Address{}, Slot{}, fmt.Errorf("%w: empty key", ErrMalformedMemIAVLKey)
+	}
+
+	switch {
+	case bytes.HasPrefix(key, evmtypes.NonceKeyPrefix):
+		if len(key) != len(evmtypes.NonceKeyPrefix)+AddressLen {
+			return EVMKeyUnknown, Address{}, Slot{}, fmt.Errorf("%w: nonce key len=%d", ErrMalformedMemIAVLKey, len(key))
+		}
+		a, ok := AddressFromBytes(key[len(evmtypes.NonceKeyPrefix):])
```
yzang2019: Nit: bad naming "a", and also belows

sei-db/state_db/sc/flatkv/evm_memiavl_mapper.go
```diff
@@ -0,0 +1,95 @@
...
+	EVMKeyCode
+	EVMKeyCodeSize
+	EVMKeyStorage
+)
+
+// ParseEVMKey parses a memiavl EVM key (x/evm store keyspace) into a typed selector.
+//
+// For non-storage keys, slot is returned as zero value.
+// For unknown prefixes, returns (EVMKeyUnknown, zero, zero, nil).
+func ParseEVMKey(key []byte) (kind EVMKeyKind, addr Address, slot Slot, err error) {
```
yzang2019: Can we have 2 separate functions instead?
One function to getKeyType which returns EVMKeyKind, and another functions to parseToAccountKey, parseToCodeKey, parseToStorageKey which returns directly the key we defined in the keys.go

yzang2019: Actually it depends on how the caller wanna use it. Let's think more about it

sei-db/state_db/sc/flatkv/keys.go
blindchaser: word is more for evm term and balance is better for code readability. updated

sei-db/state_db/sc/flatkv/evm_memiavl_mapper.go
blindchaser: good catch, fixed

sei-db/state_db/sc/flatkv/evm_memiavl_mapper.go
blindchaser: agree. if callers typically need to switch on kind first, then call type-specific parsers, the split would be cleaner. If most callers need all fields at once, keeping it combined is fine.

sei-db/state_db/sc/flatkv/keys.go
blindchaser: good point. some trade-offs are:
Fixed-length (current, 72 bytes):
- simpler
- predictable memory layout

variable length(40 bytes for eoa, 72 bytes for contract):
- save 32 bytes per EOA (pobably 90%+ accounts are EOA)
- can use length to distinguash eoa and contract?

yzang2019: Yeah, let's do the separation, I think it's a more optimized way to store data

yzang2019 (approved):

sei-db/state_db/sc/flatkv/evm_memiavl_mapper.go
```diff
@@ -0,0 +1,70 @@
...
+	EVMKeyCodeSize
+	EVMKeyStorage
+)
+
+// ParseEVMKey parses a memiavl EVM key (x/evm store keyspace).
+//
+// For non-storage keys, keyBytes is the 20-byte address.
+// For storage keys, keyBytes is addr||slot (20+32 bytes).
+// For unknown or malformed keys, returns (EVMKeyUnknown, nil).
+func ParseEVMKey(key []byte) (kind EVMKeyKind, keyBytes []byte) {
```
Kbhat1: This is quite helpful even for state store, can we make this a generic utility function in sei-db? cc: @yzang2019

PR #2725, opened 2026-01-21
fix: set max packet msg payload default to 1MB

blindchaser:

### DIFF
--- a/.github/workflows/libwasmvm.yml
+++ b/.github/workflows/libwasmvm.yml
@@ -1,0 +1,51 @@
+name: libwasmvm
+on:
+  workflow_call:
+    inputs:
+      ref:
+        description: 'Git ref to checkout (branch, tag, or SHA)'
+        required: false
+        type: string
+        default: ''
+  push:
+    branches:
+      - main
+      - release/**
+
+concurrency:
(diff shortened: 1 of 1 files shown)


sei-tendermint/config/config.go
```diff
@@ -707,7 +707,7 @@ func DefaultP2PConfig() *P2PConfig {
 		// optional header fields are used) and thus the max for (non-Jumbo frame)
 		// Ethernet is 1500 - 20 -20 = 1460
 		// Source: https://stackoverflow.com/a/3074427/820520
-		MaxPacketMsgPayloadSize: 1400,
+		MaxPacketMsgPayloadSize: 1048576,
```
pompon0: how about we hardcode it instead? Changing this value to literally anything else breaks the protocol.

pompon0: also remove the comment, it is not valid any more with this pr.

blindchaser: this is hardcoded in DefaultP2PConfig(), and existing nodes won’t pick up changes unless they update config manually. unless we want to remove this param from config.

blindchaser: good point, removed.

PR #2726, opened 2026-01-21
Exchanging MaxPacketMsgPayloadSize on handshake (CON-186)

pompon0:
A node treats sending a packet larger than locally configured MaxPacketMsgPayloadSize as protocol violation and drops connection. This PR makes nodes exchange MaxPacketMsgPayloadSize on handshake, so that they can avoid that. The fix will work only once both nodes have it. Otherwise they fallback to the old behavior.


sei-tendermint/internal/p2p/conn/connection.go
```diff
@@ -446,15 +447,15 @@ func (c *MConnection) recvRoutine(ctx context.Context) (err error) {
 
 // maxPacketMsgSize returns a maximum size of PacketMsg
 func (c *MConnection) maxPacketMsgSize() uint64 {
-	return uint64(len(protoutils.Marshal(&pb.Packet{
+	return uint64(proto.Size(&pb.Packet{
 		Sum: &pb.Packet_PacketMsg{
 			PacketMsg: &pb.PacketMsg{
-				ChannelId: 0x01,
+				ChannelId: math.MaxInt32,
```
blindchaser, line 453: can you put a short comments why MaxInt32 is used here

pompon0: int32 proto field is encoded as zigzag varint, which takes the most space for maximal/minimal value. Setting it to 1 was technically ok, because all the channel ids that we use are <128, but using maxint is foolproof.

blindchaser: got it, how about add an comments for this

blindchaser (approved):

pompon0:
we will hardcode a constant instead

PR #2731, opened 2026-01-21
Add go bench for sc commit store

yzang2019:
Add go bench for SeiDB CommitStore (Currently only MemIAVL, but will add FlatKV backend in the future)

Manually run long running test:
```
yzang@Yimings-MacBook-Pro sei-chain % go test -tags=enable_bench -bench=BenchmarkWriteThroughput -benchtime=1x -run='^$' ./sei-db/state_db/bench/... -args -keys=2000 -blocks=5000
[Progress] blocks=499/5000, keys=998000/10000000, keys/sec=199568
[Progress] blocks=894/5000, keys=1788000/10000000, keys/sec=178782
[Progress] blocks=1263/5000, keys=2526000/10000000, keys/sec=168388
[Progress] blocks=1615/5000, keys=3230000/10000000, keys/sec=161499
[Progress] blocks=1967/5000, keys=3934000/10000000, keys/sec=157354
[Progress] blocks=2314/5000, keys=4628000/10000000, keys/sec=154261
[Progress] blocks=2657/5000, keys=5314000/10000000, keys/sec=151824
[Progress] blocks=3005/5000, keys=6010000/10000000, keys/sec=150246
[Progress] blocks=3314/5000, keys=6628000/10000000, keys/sec=147286
[Progress] blocks=3631/5000, keys=7262000/10000000, keys/sec=145237
[Progress] blocks=3970/5000, keys=7940000/10000000, keys/sec=144362
[Progress] blocks=4288/5000, keys=8576000/10000000, keys/sec=142931
[Progress] blocks=4577/5000, keys=9154000/10000000, keys/sec=140829
[Progress] blocks=4868/5000, keys=9736000/10000000, keys/sec=139085
[Final] keys=10000000/10000000, keys/sec=137913, elapsed=72.51s
goos: darwin
goarch: arm64
pkg: github.com/sei-protocol/sei-chain/sei-db/state_db/bench
cpu: Apple M2 Pro
BenchmarkWriteThroughput-12    	       1	72509156250 ns/op	    137914 keys/sec	36497844384 B/op	688900873 allocs/op
--- BENCH: BenchmarkWriteThroughput-12
    bench_sc_test.go:248: keysPerBlock=2000, numBlocks=5000, totalKeys=10000000
PASS
ok  	github.com/sei-protocol/sei-chain/sei-db/state_db/bench	78.579s
```


sei-db/state_db/bench/bench_sc_test.go
```diff
@@ -0,0 +1,274 @@
...
+
+		b.StopTimer()
+		elapsed := b.Elapsed().Seconds()
+		b.ReportMetric(float64(totalKeys)/elapsed, "keys/sec")
+	}
+}
+
+// runBenchmarkWithProgress runs the benchmark with periodic progress reporting.
+// Reports keys/sec every 5 seconds to stdout.
+func runBenchmarkWithProgress(b *testing.B, keysPerBlock, numBlocks int) {
```
blindchaser: runBenchmark and runBenchmarkWithProgress share ~80% of the same code, consider refactoring?

sei-db/state_db/bench/bench_sc_test.go
```diff
@@ -0,0 +1,274 @@
...
+	keys := p.keysWritten.Load()
+	fmt.Printf("[Final] keys=%d/%d, keys/sec=%.0f, elapsed=%.2fs\n",
+		keys, p.totalKeys, float64(keys)/elapsed, elapsed)
+}
+
+func (p *ProgressReporter) report() {
+	keys := p.keysWritten.Load()
+	elapsed := time.Since(p.startTime).Seconds()
+	if elapsed > 0 {
+		blocks := keys / int64(p.totalKeys/p.totalBlocks)
```
blindchaser: nit: not likely but edge case when p.totalBlocks > p.totalKeys will panic

keysPerBlock := p.totalKeys / p.totalBlocks
if keysPerBlock == 0 {
    keysPerBlock = 1
}
blocks := keys / int64(keysPerBlock)

sei-db/state_db/bench/bench_sc_test.go
```diff
@@ -0,0 +1,274 @@
...
+	"github.com/sei-protocol/sei-chain/sei-db/common/logger"
+	"github.com/sei-protocol/sei-chain/sei-db/config"
+	"github.com/sei-protocol/sei-chain/sei-db/proto"
+	"github.com/sei-protocol/sei-chain/sei-db/state_db/sc"
+	iavl "github.com/sei-protocol/sei-chain/sei-iavl"
+)
+
+// =============================================================================
+// Flags
+// =============================================================================
```
masih: Do these comments add anything useful? If not I would remove them.

```suggestion

```

sei-db/state_db/bench/bench_sc_test.go
```diff
@@ -0,0 +1,274 @@
...
+
+var (
+	flagKeysPerBlock = flag.Int("keys", 1000, "number of keys per block")
+	flagNumBlocks    = flag.Int("blocks", 100, "number of blocks to commit")
+)
+
+// =============================================================================
+// Constants
+// =============================================================================
```
masih: I generally advise against this type of comment. They are at best subjective, and add very little for me personally in terms of understanding the code better. For example, the executable `const` syntax makes it perfectly clear that these are constants.

If AI generated, I think the model must have been trained on old code-bases probably non go, when screens where small and a 200+ line file was hard to read on small/low-res screens.

Ditto for the reminder of this PR.

sei-db/state_db/bench/bench_sc_test.go
```diff
@@ -0,0 +1,274 @@
...
+	cfg := config.StateCommitConfig{
+		AsyncCommitBuffer: 10,
+		SnapshotInterval:  100,
+	}
+
+	cs := sc.NewCommitStore(dir, logger.NewNopLogger(), cfg)
+	cs.Initialize([]string{EVMStoreName})
+
+	if _, err := cs.LoadVersion(0, false); err != nil {
+		b.Fatalf("failed to init CommitStore: %v", err)
```
masih: You can use testify `require.NoError` it accepts `b` just like `t`.

Ditto for the rest of the PR.

sei-db/state_db/bench/bench_sc_test.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	b.Cleanup(func() {
+		_ = cs.Close()
+	})
+
+	return cs
+}
+
+func generateChangesets(numBlocks, keysPerBlock int) []*proto.NamedChangeSet {
```
masih: Please feel free to do this in future PRs:

You can change the signature of this changeset generator to take a distribution `func` instead that returns keys per block. You could then easily model various random distributions of keys per block that better represent a live network, whilst still being able to represent constant keys per block by simply returning a fixed number in the function.

sei-db/state_db/bench/bench_sc_test.go
```diff
@@ -0,0 +1,274 @@
...
+	cs := make([]*proto.NamedChangeSet, numBlocks)
+	for i := range cs {
+		pairs := make([]*iavl.KVPair, keysPerBlock)
+		for j := range pairs {
+			key := make([]byte, KeySize)
+			val := make([]byte, ValueSize)
+			_, _ = rand.Read(key)
+			_, _ = rand.Read(val)
```
masih: Check error - rarely happens and a quirk of Go SDK but error nevertheless.

sei-db/state_db/bench/bench_sc_test.go
```diff
@@ -0,0 +1,274 @@
...
+	return &ProgressReporter{
+		totalKeys:   totalKeys,
+		totalBlocks: totalBlocks,
+		done:        make(chan struct{}),
+		interval:    interval,
+	}
+}
+
+// Start begins periodic progress reporting in a background goroutine.
+func (p *ProgressReporter) Start() {
```
masih: take a context on start, and use it to stop when we are done. That's clearer than the `done` one off channel.

Whatever `Start`s it, then can simply cancel the start contxt to stop the show.

sei-db/state_db/bench/bench_sc_test.go
```diff
@@ -0,0 +1,274 @@
...
+	fmt.Printf("[Final] keys=%d/%d, keys/sec=%.0f, elapsed=%.2fs\n",
+		keys, p.totalKeys, float64(keys)/elapsed, elapsed)
+}
+
+func (p *ProgressReporter) report() {
+	keys := p.keysWritten.Load()
+	elapsed := time.Since(p.startTime).Seconds()
+	if elapsed > 0 {
+		blocks := keys / int64(p.totalKeys/p.totalBlocks)
+		fmt.Printf("[Progress] blocks=%d/%d, keys=%d/%d, keys/sec=%.0f\n",
```
masih: so this type of reporting would mess up the terminal output, which violates the standard go benchmark output and renders tools like `benchstat` useless.

I recommend instead of terminal output, add metrics. or.. omit progress reporting completely if these benchmarks finish quickly (which they do - ~40s on my machine; the write throughput one that uses reporter finishes in 2s on my machine).

sei-db/state_db/bench/bench_sc_test.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+func runBenchmark(b *testing.B, keysPerBlock, numBlocks int) {
+	changesets := generateChangesets(numBlocks, keysPerBlock)
+	totalKeys := keysPerBlock * numBlocks
+
+	b.ResetTimer()
+	b.ReportAllocs()
+
+	for i := 0; i < b.N; i++ {
```
masih: Use the newer for range syntax.

sei-db/state_db/bench/bench_sc_test.go
```diff
@@ -0,0 +1,274 @@
...
+//
+// Example:
+//
+//	go test -tags=enable_bench -bench=BenchmarkWriteWithDifferentBlockSize -benchtime=1x -run='^$' \
+//	  ./sei-db/state_db/bench/...
+func BenchmarkWriteWithDifferentBlockSize(b *testing.B) {
+	const totalKeys = 1_000_000
+
+	// Different keys per block to test (numBlocks = totalKeys / keysPerBlock)
+	keysPerBlockOptions := []int{10, 20, 100, 200, 1000, 2000}
```
masih: Add 1 and 2 as well for the control case?

sei-db/state_db/bench/bench_sc_test.go
```diff
@@ -0,0 +1,274 @@
+//go:build enable_bench
```
masih: I recommend renaming this flag to something like `slow_bench`. There is no harm (and in fact encouraged) to run fast benchmarks as par of tests. Benchmarks are effectively tests and are run by default.

With the "slow"ness in mind, you can separate slower benchmarks into their own go file with their own flag, and leave the fast ones be -- or better yet, reduce things like block count for the fast ones and run the same as the slow ones.

FWIW running the two benchmarks here takes in total of 39 seconds on my machine. I would be hard pressed to call them slow considering some the tests we have in this repo.

sei-db/state_db/bench/bench_sc_test.go
```diff
@@ -0,0 +1,274 @@
...
+//
+//	-keys     Keys per block (default: 1000)
+//	-blocks   Number of blocks (default: 100)
+//
+// Example:
+//
+//	go test -tags=enable_bench -bench=BenchmarkWriteThroughput -benchtime=1x -run='^$' \
+//	  ./sei-db/state_db/bench/... -args -keys=2000 -blocks=500
+func BenchmarkWriteThroughput(b *testing.B) {
+	b.Logf("keysPerBlock=%d, numBlocks=%d, totalKeys=%d",
```
masih: Why log? Make those param to be in the benchmark name, which makes it clear as well as selectable via `-run` regex when running `go test -bench`

sei-db/state_db/bench/bench_sc_test.go
```diff
@@ -0,0 +1,274 @@
...
+	runBenchmarkWithProgress(b, *flagKeysPerBlock, *flagNumBlocks)
+}
+
+// BenchmarkWriteWithDifferentBlockSize tests throughput with fixed 1M total keys,
+// varying keysPerBlock and numBlocks to find optimal block size.
+//
+// Example:
+//
+//	go test -tags=enable_bench -bench=BenchmarkWriteWithDifferentBlockSize -benchtime=1x -run='^$' \
+//	  ./sei-db/state_db/bench/...
```
masih: Instead of repeating the examples per benchmark, i think it makes a better DX to add a section in README about how to run benchmarks etc. This type of comment tends to rot fast and would be noisy for most devs.

sei-db/state_db/bench/bench_sc_test.go
```diff
@@ -0,0 +1,274 @@
...
+	"github.com/sei-protocol/sei-chain/sei-db/state_db/sc"
+	iavl "github.com/sei-protocol/sei-chain/sei-iavl"
+)
+
+// =============================================================================
+// Flags
+// =============================================================================
+
+var (
+	flagKeysPerBlock = flag.Int("keys", 1000, "number of keys per block")
```
masih: These `flag`s don't do anything right now. Because for them to pick up actual values they need to be parsed via calling `flag.Parse`.

I advise against having these for simple benchmarks like the ones we have here. Instead have a table benchmark that passes in hardcoded values.

Unless the usecase for "adhock benchmark with arbitrary keys per block or number of blocks" is clear I would be hesitant to allow it. The chances are we will have a hand full of statistical distributions that represent the "real network" and that's it (see my comment on taking `func` for load generator). Unless of course I am missing something.

yzang2019: Good point, I'm actually thinking about using this go bench tool to support customized input via parameters to run adhoc benchmarks, for predefined benchmarks, I tend to think they shouldn't be written as a super long running process, like I want to avoid writing to much data in the predefined benchmarks (we have one already).

That's why I wanna introduce a separate go bench which can be run adoc manually, with arbitrary input to define who long we wanna run and how much data we wanna test with

What do you think would be a better way to do that?

sei-db/state_db/bench/bench_sc_test.go
yzang2019: Make sense, will put them into a readme doc

sei-db/state_db/bench/bench_sc_test.go
yzang2019: What do you mean exactly by `Make those param to be in the benchmark name` ? The log is just to print out the value we got from the input flags

sei-db/state_db/bench/bench_sc_test.go
yzang2019: Good point, separating slow vs fast makes lot of sense, will do that, great suggestion!

sei-db/state_db/bench/bench_sc_test.go
yzang2019: The reason I'm adding these is so that we can see intermediate performance numbers and trend when running a really long benchmarks, I'm expecting to use this to run future benchmarks with hours and maybe days to see how the performance varies over time. Storage benchmark is unlike others,  as the DB size grows, as more and more data is being written to, the perf number might change. So I'd leaning towards supporting these long running benchmarks.

Do you have any better advice for how we can do that using go bench?

sei-db/state_db/bench/bench_sc_test.go
yzang2019: Yeah those are AI generated comments, will clean up a little bit

sei-db/state_db/bench/bench_sc_test.go
yzang2019: Good point, will combine into one

sei-db/state_db/bench/bench_sc_test.go
yzang2019: Will think about it, but that's a great suggestion!

sei-db/state_db/bench/README.md
```diff
@@ -0,0 +1,58 @@
+# Benchmarks
```
masih, line 1: nice

PR #2743, opened 2026-01-23
[giga] Add toEvmcError() for Go to EVMC error conversion

pdrobnjak:
## Summary

- Add `toEvmcError()` function to convert geth VM errors to EVMC error types
- Maps `vm.ErrExecutionReverted` → `evmc.Revert`
- All other errors map to `evmc.Failure`

## Context

The EVMC bindings expect `Call()` to return `evmc.Error` type, not standard Go errors. This function provides the conversion layer.

**Limitation:** The evmc Go bindings currently only expose `evmc.Failure` and `evmc.Revert` constants. Additional error codes like `EVMC_OUT_OF_GAS` (3), `EVMC_INVALID_INSTRUCTION` (4), etc. are defined in the C header (`evmc/evmc.h`) but not exported in the Go bindings. A TODO is left to add `OutOfGas` mapping once the bindings are extended.

## Test plan

- [x] Verify code compiles: `go build ./giga/executor/internal/...`
- [x] Function will be exercised by state tests in #2707

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/giga/executor/internal/host_context.go
+++ b/giga/executor/internal/host_context.go
@@ -301,20 +301,19 @@ func (h *HostContext) getEVMRevision() evmc.Revision {
 // To add proper mapping for vm.ErrOutOfGas -> evmc.OutOfGas, the Go bindings in
 // github.com/ethereum/evmc would need to be extended first.
 func toEvmcError(err error) error {
-	if err == nil {
+	switch {
+	case err == nil:
 		return nil
-	}
-	// If it's already an evmc.Error, return as-is
-	if _, ok := err.(evmc.Error); ok {
+	case errors.As(err, new(evmc.Error)):
+		// Already an evmc.Error, return as-is
 		return err
-	}
-	// Map known geth VM errors to EVMC errors
-	if errors.Is(err, vm.ErrExecutionReverted) {
+	case errors.Is(err, vm.ErrExecutionReverted):
 		return evmc.Revert
+	default:
+		// All other errors map to generic failure
+		// TODO: Add evmc.OutOfGas mapping once the Go bindings expose it
+		return evmc.Failure
 	}
-	// All other errors map to generic failure
-	// TODO: Add evmc.OutOfGas mapping once the Go bindings expose it
-	return evmc.Failure
 }
 
 // To be called by an exported EVM create function which knows how to instantiate params like statedb.


giga/executor/internal/host_context.go
```diff
@@ -290,6 +292,31 @@ func (h *HostContext) getEVMRevision() evmc.Revision {
...
+	}
+	// If it's already an evmc.Error, return as-is
+	if _, ok := err.(evmc.Error); ok {
+		return err
+	}
+	// Map known geth VM errors to EVMC errors
+	if errors.Is(err, vm.ErrExecutionReverted) {
+		return evmc.Revert
+	}
+	// All other errors map to generic failure
```
masih: Would it be useful to log the error type in DEBUG for easier debugging? It should also help collating a list of errors over time to map out here.

giga/executor/internal/host_context.go
```diff
@@ -290,6 +292,31 @@ func (h *HostContext) getEVMRevision() evmc.Revision {
...
+	// TODO: Add evmc.OutOfGas mapping once the Go bindings expose it
+	return evmc.Failure
```
masih: Totally a style thing: Compared to if-else, I find switch-case statements to be much more readable and concise in functions like this.

```suggestion
	switch {
	case err == nil:
		return nil
	case errors.As(err, new(evmc.Error)):
		// Already an evmc.Error, return as-is
		return err
	case errors.Is(err, vm.ErrExecutionReverted):
		return evmc.Revert
	default:
		// All other errors map to generic failure
		// TODO: Add evmc.OutOfGas mapping once the Go bindings expose it
		return evmc.Failure
	}
```

giga/executor/internal/host_context.go
pdrobnjak: Probably yes - what do you think about sending metrics instead of DEBUG logging?

giga/executor/internal/host_context.go
pdrobnjak: Done, done.

giga/executor/internal/host_context.go
masih: With metrics, i would worry about the unbound cardinality of possible error types (assuming you are thinking of some sort of tag by error type). it could heavily pollute the metrics.

Honestly, i think simple logging will do, since we have to hand-roll and audit any changes here anyway :)

pdrobnjak: Makes sense, will add in a subsequent PR.

PR #2744, opened 2026-01-23
Add go generate script to download evmone libraries on the fly

masih:
Avoid accidental drift of evmone dynamic libs by fetching them via `go generate`. The script downloads archives from GitHub releases, verifies SHA256 hashes, and extracts the platform-specific shared libraries.

To upgrade evmone, update the version constant and hashes in the script, then re-run `go generate ./...`.

Note that the file permissions for darwin arm64 libs is updated to be `0644`, consistent with the existing linux amd64 libs as a side benefit of changes here: consistent file attributes are enforced.

### DIFF
--- a/giga/deps/xevm/keeper/address.go
+++ b/giga/deps/xevm/keeper/address.go
@@ -8,7 +8,7 @@ import (
 )
 
 func (k *Keeper) SetAddressMapping(ctx sdk.Context, seiAddress sdk.AccAddress, evmAddress common.Address) {
-	store := ctx.GigaKVStore(k.storeKey)
+	store := k.GetKVStore(ctx)
 	store.Set(types.EVMAddressToSeiAddressKey(evmAddress), seiAddress)
 	store.Set(types.SeiAddressToEVMAddressKey(seiAddress), evmAddress[:])
 	if !k.accountKeeper.HasAccount(ctx, seiAddress) {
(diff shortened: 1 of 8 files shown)


giga/executor/lib/gen/main.go
```diff
@@ -0,0 +1,159 @@
...
+
+	// Create context that cancels on SIGINT or SIGTERM
+	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
+	defer cancel()
+
+	for _, p := range platforms {
+		if err := downloadAndExtract(ctx, p, outDir); err != nil {
+			if ctx.Err() != nil {
+				_, _ = fmt.Fprintf(os.Stderr, "\nInterrupted\n")
+				os.Exit(1)
```
pompon0: nit: consider log.Fatalf

giga/executor/lib/gen/main.go
```diff
@@ -0,0 +1,159 @@
...
+			}
+			_, _ = fmt.Fprintf(os.Stderr, "Failed %s-%s: %v\n", p.OS, p.Arch, err)
+			os.Exit(1)
+		}
+	}
+	fmt.Println("All platforms downloaded successfully!")
```
pompon0: nit: log.Print

giga/executor/lib/gen/main.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return fmt.Errorf("gzip: %w", err)
+	}
+	defer func() { _ = gzr.Close() }()
+
+	ext := map[string]string{"darwin": "dylib", "linux": "so"}[p.OS]
+	libPattern := fmt.Sprintf("libevmone.%s.%s", evmoneVersion, ext)
+	if p.OS == "linux" {
+		libPattern = fmt.Sprintf("libevmone.so.%s", evmoneVersion)
+	}
```
pompon0: that looks weird. ptal

PR #2752, opened 2026-01-23
Fix seid version tagging logic for release branch

yzang2019:
Resolve VERSION from branch or tag:
- Extract vX.Y.Z (+ any suffix) from the branch name when present.
- Compare only the base vX.Y.Z (strip any suffix) between branch/tag.
- Prefer tag if bases are equal; otherwise use whichever base is newer.

Tested with a few existing release branches

### DIFF
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -89,19 +89,6 @@ func (app *App) RegisterUpgradeHandlers() {
 				return newVM, err
 			}
 
-			// IBC Toggle migration: initialize InboundEnabled and OutboundEnabled params
-			if upgradeName == "v6.4.0" {
-				newVM, err := app.mm.RunMigrations(ctx, app.configurator, fromVM)
-				if err != nil {
-					return newVM, err
-				}
-
-				// Enable IBC inbound and outbound by default
-				app.IBCKeeper.SetInboundEnabled(ctx, true)
-				app.IBCKeeper.SetOutboundEnabled(ctx, true)
-				return newVM, err
-			}
-
 			return app.mm.RunMigrations(ctx, app.configurator, fromVM)
 		})
 	}
--- a/sei-ibc-go/modules/core/keeper/migrations.go
+++ b/sei-ibc-go/modules/core/keeper/migrations.go
@@ -4,6 +4,7 @@ import (
 	sdk "github.com/sei-protocol/sei-chain/sei-cosmos/types"
 
 	clientkeeper "github.com/sei-protocol/sei-chain/sei-ibc-go/modules/core/02-client/keeper"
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -1,6 +1,26 @@
...
+		echo "$(TAG_VERSION)"; \
+	elif [ -z "$(TAG_VERSION)" ]; then \
+		echo "$(BRANCH_VERSION)"; \
+	elif [ "$$BRANCH_BASE" = "$$TAG_BASE" ]; then \
+		echo "$(TAG_VERSION)"; \
+	elif [ "$$(printf "%s\n%s\n" "$$TAG_BASE" "$$BRANCH_BASE" | sort -V | tail -n 1)" = "$$BRANCH_BASE" ]; then \
+		echo "$(BRANCH_VERSION)"; \
+	else \
+		echo "$(TAG_VERSION)"; \
+	fi)
```
masih: I _think_ this can be simplified to something like this?

```suggestion make
BRANCH_NAME := $(shell git rev-parse --abbrev-ref HEAD)
BRANCH_VERSION := $(shell echo "$(BRANCH_NAME)" | grep -oE 'v[0-9]+\.[0-9]+\.[0-9]+[-A-Za-z0-9._]*')
TAG_VERSION := $(shell git describe --tags 2>/dev/null)

VERSION := $(shell \
	bv="$(BRANCH_VERSION)"; tv="$(TAG_VERSION)"; \
	bb=$${bv%%[-A-Za-z_]*}; tb=$${tv%%[-A-Za-z_]*}; \
	if [ -z "$$bv" ] || [ "$$bb" = "$$tb" ] || \
	   [ "$$(printf '%s\n%s' "$$tb" "$$bb" | sort -V | tail -1)" = "$$tb" ]; then \
		echo "$${tv:-$$bv}"; \
	else \
		echo "$$bv"; \
	fi)
```

yzang2019: Make sense, thanks for that suggestion! I wanna keep the original regex though to cover more edge cases, there's slight difference:

```
Examples:
release/v6.3.9test1
%%[-A-Za-z_]* → v6.3.9 (works)
sed → v6.3.9 (works)

release/v6.3.9_hotfix
%%[-A-Za-z_]* → v6.3.9 (works)
sed → v6.3.9 (works)

release/v6.3.9.1 (suffix starts with .)
%%[-A-Za-z_]* → no change (v6.3.9.1)
sed → v6.3.9 (desired base version)

release/v6.3.9+build1 (suffix starts with +)
%%[-A-Za-z_]* → no change
sed → v6.3.9
So the sed approach handles more edge cases and is safer for comparison.
```

yzang2019: Updated my PR for a simplified version @masih feel free to take another look

PR #2753, opened 2026-01-23
[giga] honor configured sstore gas values

stevenlanders:
- sstore should cost the same on non-giga as giga
- we recently made this a chain param, so it needs to be the same

- new unit tests
- existing integration tests

### DIFF
--- a/giga/executor/internal/host_context_test.go
+++ b/giga/executor/internal/host_context_test.go
@@ -333,7 +333,7 @@ func TestNewHostContextConfig(t *testing.T) {
 			expectedDelta: 0,
 		},
 		{
-			name: "Sei default (72k) - 52k delta",
+			name: "Higher value (72k) - 52k delta",
 			chainConfig: &params.ChainConfig{
 				SeiSstoreSetGasEIP2200: func() *uint64 { v := uint64(72000); return &v }(),
 			},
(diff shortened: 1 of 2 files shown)


giga/executor/executor.go
```diff
@@ -18,7 +18,12 @@ type Executor struct {
 
 func NewEvmoneExecutor(evmoneVM *evmc.VM, blockCtx vm.BlockContext, stateDB vm.StateDB, chainConfig *params.ChainConfig, config vm.Config, customPrecompiles map[common.Address]vm.PrecompiledContract) *Executor {
 	evm := vm.NewEVM(blockCtx, stateDB, chainConfig, config, customPrecompiles)
-	hostContext := internal.NewHostContext(evmoneVM, evm)
+	// Get custom SSTORE gas cost from chain config (Sei's custom parameter)
+	var seiSstoreGas uint64
+	if chainConfig.SeiSstoreSetGasEIP2200 != nil {
+		seiSstoreGas = *chainConfig.SeiSstoreSetGasEIP2200
+	}
+	hostContext := internal.NewHostContext(evmoneVM, evm, seiSstoreGas)
```
yzang2019: Rather than passing the SSTORE into HostContext which requires to change the constructor, would it be easier to just add a function within HostContext to get the SStoreGas from chainConfig.SeiSstoreSetGasEIP2200  directly?

stevenlanders: I gave it some thought - i'm thinking there will likely be other things like this, so i adjusted the constructor to accept a `HostContextConfig` which will pre-render a gas sstore `delta`, avoid the pointer dereferences for SeiSstoreSetGasEIP2200 per tx sstore.

PR #2754, opened 2026-01-23
Composite State Store part 1: EVM config and type definitions

Kbhat1:
- Adds configuration and type definitions for the EVM state store layer:
- EVMStateStoreConfig in config/config.go for enabling/configuring EVM SS
- Type aliases to commonevm.EVMKeyKind for EVM store types
- Helper functions AllEVMStoreTypes() and StoreTypeName()

- Unit tests


sei-db/state_db/ss/evm/config.go
```diff
@@ -0,0 +1,51 @@
...
+
+// EVMStoreKey is the cosmos store key for EVM module
+const EVMStoreKey = "evm"
+
+// EVMStoreType identifies the type of EVM sub-database.
+// Alias to EVMKeyKind from common/evm - use commonevm.ParseEVMKey for routing.
+type EVMStoreType = commonevm.EVMKeyKind
+
+// Re-export EVMKeyKind constants for convenience
+const (
```
yzang2019: Not sure if this belongs to config, maybe we should move the store types to another file?

sei-db/state_db/ss/evm/config.go
```diff
@@ -0,0 +1,51 @@
...
+	StoreUnknown  = commonevm.EVMKeyUnknown
+	StoreNonce    = commonevm.EVMKeyNonce
+	StoreCodeHash = commonevm.EVMKeyCodeHash
+	StoreCode     = commonevm.EVMKeyCode
+	StoreCodeSize = commonevm.EVMKeyCodeSize
```
yzang2019: I don't think we wanna store code size in the new EVM state any more, that's not part of the standard EVM state

sei-db/state_db/ss/evm/config.go
```diff
@@ -0,0 +1,51 @@
...
+// EVMStoreKey is the cosmos store key for EVM module
+const EVMStoreKey = "evm"
+
+// EVMStoreType identifies the type of EVM sub-database.
+// Alias to EVMKeyKind from common/evm - use commonevm.ParseEVMKey for routing.
+type EVMStoreType = commonevm.EVMKeyKind
+
+// Re-export EVMKeyKind constants for convenience
+const (
+	StoreUnknown  = commonevm.EVMKeyUnknown
```
jewei1997: we don't need a key for Balance because we're using tendermint store for balances right?

yzang2019: We should still add it though I think, but won't be used until migration happens

blindchaser (approved):

PR #2755, opened 2026-01-23
Composite State Store Part 2: EVM database implementation

Kbhat1:

- EVMDatabase: Single PebbleDB with DefaultComparer for each EVM data type
- EVMStateStore: Manages multiple EVMDatabase instances (nonce, code, codehash, codesize, storage)
- Versioned key encoding with big-endian version suffix
- ApplyBatch and ApplyChangesetParallel for efficient concurrent writes
- Uses commonevm.ParseMemIAVLEVMKey directly for key routing

- Unit tests

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1444,15 +1444,15 @@ func (app *App) PartitionPrioritizedTxs(_ sdk.Context, txs [][]byte, typedTxs []
 // ExecuteTxsConcurrently calls the appropriate function for processing transacitons
 func (app *App) ExecuteTxsConcurrently(ctx sdk.Context, txs [][]byte, typedTxs []sdk.Tx, absoluteTxIndices []int) ([]*abci.ExecTxResult, sdk.Context) {
 	// Giga only supports synchronous execution for now
-	if app.GigaExecutorEnabled {
-		results := app.ProcessTxsSynchronousGiga(ctx, txs, typedTxs, absoluteTxIndices)
-		return results, ctx
+	if app.GigaExecutorEnabled && app.GigaOCCEnabled {
+		return app.ProcessTXsWithOCCGiga(ctx, txs, typedTxs, absoluteTxIndices)
+	} else if app.GigaExecutorEnabled {
+		return app.ProcessTxsSynchronousGiga(ctx, txs, typedTxs, absoluteTxIndices), ctx
 	} else if !ctx.IsOCCEnabled() {
-		results := app.ProcessTxsSynchronousV2(ctx, txs, typedTxs, absoluteTxIndices)
-		return results, ctx
+		return app.ProcessTxsSynchronousV2(ctx, txs, typedTxs, absoluteTxIndices), ctx
 	}
 
-	return app.ProcessTXsWithOCC(ctx, txs, typedTxs, absoluteTxIndices)
+	return app.ProcessTXsWithOCCV2(ctx, txs, typedTxs, absoluteTxIndices)
 }
 
 func (app *App) GetDeliverTxEntry(ctx sdk.Context, txIndex int, absoluateIndex int, bz []byte, tx sdk.Tx) (res *sdk.DeliverTxEntry) {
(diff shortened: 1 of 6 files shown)


sei-db/state_db/ss/evm/db.go
```diff
@@ -0,0 +1,616 @@
...
+	// Version tracking
+	latestVersion   int64
+	earliestVersion int64
+}
+
+// OpenEVMDB opens a PebbleDB with default comparer for EVM data
+func OpenEVMDB(dir string, storeType EVMStoreType) (*EVMDatabase, error) {
+	opts := &pebble.Options{
+		Comparer:           pebble.DefaultComparer,
+		MaxConcurrentCompactions: func() int { return 2 },
```
yzang2019: Is 2 enough or shall we bump to numCPU?

sei-db/state_db/ss/evm/db.go
```diff
@@ -0,0 +1,616 @@
...
+// GetLatestVersion returns the latest version in this database
+func (db *EVMDatabase) GetLatestVersion() int64 {
+	return db.latestVersion
+}
+
+// SetLatestVersion updates the latest version
+func (db *EVMDatabase) SetLatestVersion(version int64) error {
+	buf := make([]byte, 8)
+	binary.BigEndian.PutUint64(buf, uint64(version))
+	if err := db.storage.Set([]byte(latestVersionKey), buf, syncWriteOpts); err != nil {
```
yzang2019: Do we really wanna sync here? This is gonna trigger fsync which could impact performance

sei-db/state_db/ss/evm/db.go
```diff
@@ -0,0 +1,616 @@
...
+	}
+	it.index++
+}
+
+func (it *EVMIterator) Close() error {
+	return nil
+}
+
+// EVMStateStore manages multiple EVMDatabase instances, one per EVM data type
+type EVMStateStore struct {
```
yzang2019: Should this be moved to a separate file? I feel it would be cleaner if we keep db.go only containing the db related logic. And EVMStateStore seems to be a layer above that, since it needs to manage multiple dbs

sei-db/state_db/ss/evm/db.go
```diff
@@ -0,0 +1,616 @@
...
+	for _, db := range s.databases {
+		if err := db.SetEarliestVersion(version); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Prune removes old versions from all databases
+func (s *EVMStateStore) Prune(version int64) error {
```
yzang2019: Shall we parallelize this or have each DB run the pruning in the background?

sei-db/state_db/ss/evm/db.go
Kbhat1: Yea I think I'll split this out + I'll also move the batch logic out into separate file

sei-db/state_db/ss/evm/db.go
```diff
@@ -0,0 +1,430 @@
...
+	storeType EVMStoreType
+	storage   *pebble.DB
+
+	// Version tracking
+	latestVersion   int64
+	earliestVersion int64
+}
+
+// OpenEVMDB opens a PebbleDB with default comparer for EVM data
+func OpenEVMDB(dir string, storeType EVMStoreType) (*EVMDatabase, error) {
```
yzang2019: suggest rename to OpenDB to keep it consistent

sei-db/state_db/ss/evm/db.go
```diff
@@ -0,0 +1,430 @@
...
+	db, err := pebble.Open(dbPath, opts)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open EVM DB %s: %w", StoreTypeName(storeType), err)
+	}
+
+	evmDB := &EVMDatabase{
+		storeType:       storeType,
+		storage:         db,
+		latestVersion:   0,
+		earliestVersion: 0,
```
yzang2019: Nit: default is 0 anyway so we can omit these 2 lines

sei-db/state_db/ss/evm/db.go
```diff
@@ -0,0 +1,430 @@
...
+	}
+
+	// Load version metadata
+	if err := evmDB.loadVersionMetadata(); err != nil {
+		db.Close()
```
yzang2019: Should check error for close

yzang2019 (approved):

PR #2756, opened 2026-01-23
Composite State Store part 3: Read path implementation

Kbhat1:

- Adds CompositeStateStore with read path routing between Cosmos_SS and EVM_SS:
- Get/Has: Check EVM_SS first for EVM keys, fallback to Cosmos_SS
- Iterator/ReverseIterator: Use Cosmos_SS (source of truth)
- Version methods delegate to Cosmos_SS

- Unit tests

### DIFF
--- a/sei-db/state_db/ss/composite/store.go
+++ b/sei-db/state_db/ss/composite/store.go
@@ -307,7 +307,7 @@ func (s *CompositeStateStore) extractEVMChanges(changesets []*proto.NamedChangeS
 		}
 		for _, kvPair := range changeset.Changeset.Pairs {
 			evmStoreType, keyBytes := commonevm.ParseEVMKey(kvPair.Key)
-			if evmStoreType == evm.StoreUnknown {
+			if evmStoreType == evm.StoreEmpty {
 				continue // Skip zero-length keys
 			}
 			// All EVM keys are routed: optimized keys use stripped key, legacy uses full key
@@ -386,7 +386,7 @@ func (s *CompositeStateStore) Import(version int64, ch <-chan types.SnapshotNode
 		// Route EVM keys to EVM-SS
 		if isEVM {
 			evmStoreType, keyBytes := commonevm.ParseEVMKey(node.Key)
-			if evmStoreType != evm.StoreUnknown {
+			if evmStoreType != evm.StoreEmpty {
 				evmChanges[evmStoreType] = append(evmChanges[evmStoreType], &iavl.KVPair{
 					Key:   keyBytes,
 					Value: node.Value,
@@ -456,7 +456,7 @@ func (s *CompositeStateStore) RawImport(ch <-chan types.RawSnapshotNode) error {
 		// Route EVM keys to EVM-SS
 		if isEVM {
 			evmStoreType, keyBytes := commonevm.ParseEVMKey(node.Key)
-			if evmStoreType != evm.StoreUnknown {
+			if evmStoreType != evm.StoreEmpty {
 				if evmChangesByVersion[node.Version] == nil {
 					evmChangesByVersion[node.Version] = make(map[evm.EVMStoreType][]*iavl.KVPair, evm.NumEVMStoreTypes)
 				}
@@ -688,7 +688,7 @@ func extractEVMChangesFromChangesets(changesets []*proto.NamedChangeSet) map[evm
 		}
 		for _, kvPair := range changeset.Changeset.Pairs {
 			evmStoreType, keyBytes := commonevm.ParseEVMKey(kvPair.Key)
-			if evmStoreType == evm.StoreUnknown {
+			if evmStoreType == evm.StoreEmpty {
 				continue // Skip zero-length keys
 			}
 			evmChanges[evmStoreType] = append(evmChanges[evmStoreType], &iavl.KVPair{
--- a/sei-db/state_db/ss/evm/store.go
+++ b/sei-db/state_db/ss/evm/store.go
@@ -80,7 +80,7 @@ func (s *EVMStateStore) GetDB(storeType EVMStoreType) *EVMDatabase {
 // Get retrieves a value using the full EVM key (with prefix)
 func (s *EVMStateStore) Get(key []byte, version int64) ([]byte, error) {
 	storeType, strippedKey := commonevm.ParseEVMKey(key)
-	if storeType == StoreUnknown {
+	if storeType == StoreEmpty {
 		return nil, nil
 	}
 
@@ -95,7 +95,7 @@ func (s *EVMStateStore) Get(key []byte, version int64) ([]byte, error) {
 // Has checks if a key exists
 func (s *EVMStateStore) Has(key []byte, version int64) (bool, error) {
 	storeType, strippedKey := commonevm.ParseEVMKey(key)
-	if storeType == StoreUnknown {
+	if storeType == StoreEmpty {
 		return false, nil
 	}
 
--- a/sei-db/common/evm/keys.go
+++ b/sei-db/common/evm/keys.go
@@ -21,48 +21,50 @@ var (
 type EVMKeyKind uint8
 
 const (
-	EVMKeyUnknown EVMKeyKind = iota
-	EVMKeyNonce
-	EVMKeyCodeHash
-	EVMKeyCode
-	EVMKeyStorage
+	EVMKeyEmpty    EVMKeyKind = iota // Returned only for zero-length keys
+	EVMKeyNonce                      // Stripped key: 20-byte address
+	EVMKeyCodeHash                   // Stripped key: 20-byte address
+	EVMKeyCode                       // Stripped key: 20-byte address
+	EVMKeyStorage                    // Stripped key: addr||slot (20+32 bytes)
+	EVMKeyLegacy                     // Full original key preserved (address mappings, codesize, etc.)
 )
 
 // ParseEVMKey parses an EVM key from the x/evm store keyspace.
 //
-// For non-storage keys, keyBytes is the 20-byte address.
-// For storage keys, keyBytes is addr||slot (20+32 bytes).
-// For unknown or malformed keys, returns (EVMKeyUnknown, nil).
+// For optimized keys (nonce, code, codehash, storage), keyBytes is the stripped key.
+// For legacy keys (all other EVM data), keyBytes is the full original key.
+// Only returns EVMKeyEmpty for zero-length keys.
 func ParseEVMKey(key []byte) (kind EVMKeyKind, keyBytes []byte) {
 	if len(key) == 0 {
-		return EVMKeyUnknown, nil
+		return EVMKeyEmpty, nil
 	}
 
 	switch {
 	case bytes.HasPrefix(key, evmtypes.NonceKeyPrefix):
 		if len(key) != len(evmtypes.NonceKeyPrefix)+addressLen {
-			return EVMKeyUnknown, nil
+			return EVMKeyLegacy, key // Malformed but still EVM data
 		}
 		return EVMKeyNonce, key[len(evmtypes.NonceKeyPrefix):]
 
 	case bytes.HasPrefix(key, evmtypes.CodeHashKeyPrefix):
 		if len(key) != len(evmtypes.CodeHashKeyPrefix)+addressLen {
-			return EVMKeyUnknown, nil
+			return EVMKeyLegacy, key
 		}
 		return EVMKeyCodeHash, key[len(evmtypes.CodeHashKeyPrefix):]
 
 	case bytes.HasPrefix(key, evmtypes.CodeKeyPrefix):
 		if len(key) != len(evmtypes.CodeKeyPrefix)+addressLen {
-			return EVMKeyUnknown, nil
+			return EVMKeyLegacy, key
 		}
 		return EVMKeyCode, key[len(evmtypes.CodeKeyPrefix):]
 
 	case bytes.HasPrefix(key, evmtypes.StateKeyPrefix):
 		if len(key) != len(evmtypes.StateKeyPrefix)+addressLen+slotLen {
-			return EVMKeyUnknown, nil
+			return EVMKeyLegacy, key
 		}
 		return EVMKeyStorage, key[len(evmtypes.StateKeyPrefix):]
 	}
 
-	return EVMKeyUnknown, nil
+	// All other EVM keys go to legacy store (address mappings, codesize, etc.)
+	return EVMKeyLegacy, key
 }
--- a/sei-db/common/evm/keys_test.go
+++ b/sei-db/common/evm/keys_test.go
@@ -30,6 +30,7 @@ func TestParseEVMKey(t *testing.T) {
 		wantKind  EVMKeyKind
 		wantBytes []byte
 	}{
+		// Optimized keys - stripped
 		{
 			name:      "Nonce",
 			key:       concat(evmtypes.NonceKeyPrefix, addr),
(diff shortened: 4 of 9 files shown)


sei-db/state_db/ss/composite/store.go
```diff
@@ -0,0 +1,211 @@
...
+
+// Get retrieves a value for a key at a specific version
+// For EVM keys: check EVM_SS first, fallback to Cosmos_SS
+// For non-EVM keys: use Cosmos_SS directly
+func (s *CompositeStateStore) Get(storeKey string, version int64, key []byte) ([]byte, error) {
+	// Try EVM store first for EVM keys
+	if s.evmStore != nil && storeKey == evm.EVMStoreKey {
+		storeType, strippedKey := commonevm.ParseEVMKey(key)
+		if storeType != evm.StoreUnknown {
+			db := s.evmStore.GetDB(storeType)
```
yzang2019: Shouldn't we just call EVMStore.Get since EVMStore is supposed to manage all DBs? why do we need to get the actual db from it?

yzang2019: Same for below

sei-db/state_db/ss/composite/store.go
```diff
@@ -0,0 +1,198 @@
...
+	logger      logger.Logger
+}
+
+// NewCompositeStateStore creates a new composite state store
+// cosmosStore: the main state store (required)
+// evmConfig: configuration for EVM state stores (optional - if nil or disabled, only cosmosStore is used)
+// homeDir: base directory for data files
+func NewCompositeStateStore(
+	cosmosStore types.StateStore,
+	evmConfig *config.EVMStateStoreConfig,
```
yzang2019: 1. Why only passing EVMStateStoreConfig here? 
I think either we should pass both CosmosStateStoreConfig and EVMStateStoreConfig, or we should pass in a consolidated config which contains both the cosmos and evm config.

2. Instead of pass in the cosmosStore, we can initialize the cosmos store and evm store in this constructor instead.

sei-db/state_db/ss/composite/store.go
```diff
@@ -0,0 +1,198 @@
...
+
+	return cs, nil
+}
+
+// Get retrieves a value for a key at a specific version
+// For EVM keys: check EVM_SS first, fallback to Cosmos_SS
+// For non-EVM keys: use Cosmos_SS directly
+func (s *CompositeStateStore) Get(storeKey string, version int64, key []byte) ([]byte, error) {
+	// Try EVM store first for EVM keys
+	if s.evmStore != nil && storeKey == evm.EVMStoreKey {
```
yzang2019: We probably need a config check and only read is enabled will we try reading from the evm store.

This config can be something like a read mode

Kbhat1: Added EnableRead config flag. Get/Has now check it before reading from EVM store.

sei-db/state_db/ss/composite/store.go
```diff
@@ -0,0 +1,293 @@
...
+
+// NewCompositeStateStore creates a new composite state store that manages both Cosmos_SS and EVM_SS.
+// It initializes both stores internally and starts pruning on the composite store.
+//
+// ssConfig: configuration for the main Cosmos state store (required)
+// evmConfig: configuration for EVM state stores (optional - if nil or disabled, only Cosmos_SS is used)
+// homeDir: base directory for data files
+func NewCompositeStateStore(
+	ssConfig config.StateStoreConfig,
+	evmConfig *config.EVMStateStoreConfig,
```
yzang2019: Curious why do we use pointer for evm config and non pointer for ss config? Shall we keep both consistent

sei-db/state_db/ss/composite/store.go
```diff
@@ -0,0 +1,731 @@
...
+			_ = cosmosStore.Close()
+			return nil, fmt.Errorf("failed to create EVM store: %w", err)
+		}
+		cs.evmStore = evmStore
+		log.Info("EVM state store enabled", "dir", evmDir, "writeMode", evmConfig.WriteMode, "readMode", evmConfig.ReadMode)
+	}
+
+	// Recover from WAL if needed
+	changelogPath := utils.GetChangelogPath(dbHome)
+	if err := recoverFromWAL(log, changelogPath, cs); err != nil {
```
blindchaser: is there any duplication/overlap between recoverFromWAL vs (replayWAL() + RecoverCompositeStateStore())? if yes consider refactor them

sei-db/state_db/ss/composite/store.go
```diff
@@ -0,0 +1,731 @@
...
+	return streamHandler.Replay(startOffset, lastOffset, func(index uint64, entry proto.ChangelogEntry) error {
+		// Stop if we've reached the end version
+		if toVersion >= 0 && entry.Version > toVersion {
+			return nil
+		}
+		return handler(entry)
+	})
+}
+
+func findReplayStartOffset(streamHandler wal.ChangelogWAL, firstOffset, lastOffset uint64, targetVersion int64) (uint64, error) {
```
blindchaser, line 660: how about using binary search to improve it to O(logN)?

sei-db/state_db/ss/composite/store.go
```diff
@@ -0,0 +1,731 @@
...
+
+	// Start Cosmos import in background
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		cosmosErr = s.cosmosStore.Import(version, cosmosCh)
+	}()
+
+	// Process incoming nodes
+	for node := range ch {
```
blindchaser, line 378: not sure how we do it now. it looks all EVM key-value pairs are buffered into evmChanges in memory, if the machine RAM is low this could leads to OOM. consider periodically flushing to the EVM store? same thing in RawImport

sei-db/state_db/ss/composite/store.go
```diff
@@ -0,0 +1,731 @@
...
+func replayWAL(
+	logger logger.Logger,
+	changelogPath string,
+	fromVersion int64,
+	toVersion int64, // -1 means replay to end of WAL
+	handler WALEntryHandler,
+) error {
+	streamHandler, err := wal.NewChangelogWAL(logger, changelogPath, wal.Config{})
+	if err != nil {
+		return nil // No WAL to recover from
```
blindchaser: in this way, can we catch corrupt WAL errors or IO errors ?

app/seidb.go
```diff
@@ -37,6 +37,13 @@ const (
 	FlagSSPruneInterval     = "state-store.ss-prune-interval"
 	FlagSSImportNumWorkers  = "state-store.ss-import-num-workers"
 
+	// EVM SS Store configs (optimized EVM storage layer)
+	FlagEVMSSEnable     = "state-store.evm-ss-enable"
+	FlagEVMSSDirectory  = "state-store.evm-ss-db-directory"
+	FlagEVMSSKeepRecent = "state-store.evm-ss-keep-recent"
```
yzang2019: Do you think we should share the same keep-recent config for both SS?

app/seidb.go
```diff
@@ -37,6 +37,13 @@ const (
 	FlagSSPruneInterval     = "state-store.ss-prune-interval"
 	FlagSSImportNumWorkers  = "state-store.ss-import-num-workers"
 
+	// EVM SS Store configs (optimized EVM storage layer)
+	FlagEVMSSEnable     = "state-store.evm-ss-enable"
```
yzang2019: I wonder if we need this enable flag or not, or should we just control it via the write mode

sei-db/state_db/ss/store.go
```diff
@@ -31,8 +32,34 @@ func RegisterBackend(backendType BackendType, initializer BackendInitializer) {
...
+// For backward compatibility - use NewStateStoreWithEVM when EVM optimization is needed.
 func NewStateStore(logger logger.Logger, homeDir string, ssConfig config.StateStoreConfig) (types.StateStore, error) {
+	return NewStateStoreWithEVM(logger, homeDir, ssConfig, nil)
+}
+
+// NewStateStoreWithEVM creates a new state store, optionally with EVM optimization layer.
+// When evmConfig is enabled, returns a CompositeStateStore that routes EVM data to
+// optimized separate databases while maintaining full compatibility with the StateStore interface.
+// When evmConfig is nil or disabled, returns a plain state store.
+func NewStateStoreWithEVM(
```
yzang2019: Do we really need a separate function here or can we just combine this with NewStateStore? 
Like in that NewStateStore we always launch a NewCompositeStateStore, the CompositeStateStore should support cosmos only mode which is the same as today, and other modes based on config?

sei-db/state_db/ss/composite/store.go
```diff
@@ -0,0 +1,731 @@
...
+)
+
+// CompositeStateStore routes operations between Cosmos_SS (main state store) and EVM_SS (optimized EVM stores).
+// - Reads check EVM_SS first for EVM keys, then fallback to Cosmos_SS
+// - Writes routing controlled by WriteMode (cosmos_only, dual_write, split_write)
+type CompositeStateStore struct {
+	cosmosStore types.StateStore   // Main MVCC PebbleDB for all modules
+	evmStore    *evm.EVMStateStore // Separate EVM DBs with default comparer (nil if disabled)
+	ssConfig    config.StateStoreConfig
+	evmConfig   config.EVMStateStoreConfig
```
yzang2019: You can also choose to combine and move evmConfig to something inside ssConfig, I end up doing that for sc since I feel it's easier and you don't need to change the upper layer interface

sei-db/state_db/ss/composite/store.go
```diff
@@ -0,0 +1,731 @@
...
+
+	cs := &CompositeStateStore{
+		cosmosStore: cosmosStore,
+		ssConfig:    ssConfig,
+		evmConfig:   evmConfig,
+		logger:      log,
+	}
+
+	// Initialize EVM stores if enabled
+	if evmConfig.Enable {
```
yzang2019: To me this seems a redundant/duplicate config same as write mode, i'm fine to keep it, but can also simplify the config to just use write mode

sei-db/state_db/ss/composite/store.go
```diff
@@ -0,0 +1,731 @@
...
+		cosmosChangesets = stripEVMFromChangesets(changesets)
+	}
+
+	// Write to both stores in parallel - both must succeed
+	// If either fails, return error. Caller can retry safely (idempotent writes).
+	var wg sync.WaitGroup
+	var cosmosErr, evmErr error
+
+	wg.Add(1)
+	go func() {
```
yzang2019: I think we should make ApplyChangesetSync always sync, so here I'd suggest just call s.cosmosStore.ApplyChangesetSync() and s.evmStore.ApplyChangesetSync() sequentially without the extra goroutine. The more performant version should be implemented in ApplyChangesetAsync

sei-db/state_db/ss/composite/store.go
```diff
@@ -0,0 +1,731 @@
...
+		defer wg.Done()
+		cosmosErr = s.cosmosStore.ApplyChangesetAsync(version, cosmosChangesets)
+	}()
+
+	if len(evmChanges) > 0 {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			evmErr = s.evmStore.ApplyChangesetParallel(version, evmChanges)
```
yzang2019: It seems like ApplyChangesetParallel is actually a blocking call, I think what we really need is evmStore.ApplyChangesetAsync, in that function we simply enqueue the changeset into a channel and returns. 

Both evmStore.ApplyChangeset and evmStore.ApplyChangesetAsync eventually should share the same underlying function to persist data in parallel to all DB

sei-db/state_db/ss/composite/store.go
Kbhat1: Exported ReplayWAL from the composite package + RecoverStateStore is just a thin wrapper around composite.ReplayWAL

sei-db/state_db/ss/composite/store.go
Kbhat1: updated

sei-db/state_db/ss/composite/store.go
Kbhat1: Added evmImportFlushThreshold = 10000. Both Import() and RawImport() now flush EVM changes to disk every 10k entries via a flushEVM() closure

app/seidb.go
Kbhat1: Removed the separate EVMStateStoreConfig.KeepRecent. Both Cosmos and EVM stores now use the single StateStoreConfig.KeepRecent.

app/seidb.go
Kbhat1: Removed the Enable flag entirely

sei-db/state_db/ss/store.go
Kbhat1: Yea makes sense. Combined. Removed NewStateStoreWithEVM and newPlainStateStore. NewStateStore now always creates CompositeStateStore

sei-db/state_db/ss/composite/store.go
Kbhat1: Yea discussed this on call. Removed goroutines + WaitGroup from ApplyChangesetSync. It now writes Cosmos first, then EVM, sequentially

Kbhat1: Each EVMDatabase now has a background goroutine consuming from a buffered channel (asyncBufferSize = 100)

sei-db/state_db/ss/evm/store.go
```diff
@@ -30,11 +46,30 @@ func NewEVMStateStore(dir string) (*EVMStateStore, error) {
...
 	return store, nil
 }
 
+// asyncWorker processes writes from a per-DB channel until it's closed.
+func (s *EVMStateStore) asyncWorker(db *EVMDatabase, ch <-chan dbWrite) {
+	defer s.asyncWg.Done()
+	for w := range ch {
+		if err := db.ApplyBatch(w.pairs, w.version); err != nil {
+			// Log error but continue processing; writes are idempotent and
+			// WAL replay will catch up on restart.
```
blindchaser: is the log missing here?

sei-db/state_db/ss/composite/store.go
```diff
@@ -0,0 +1,702 @@
...
+// Only advances EVM version when EVM writes are active (not CosmosOnlyWrite),
+// so that WAL catch-up can later backfill the EVM store if the operator switches modes.
+func (s *CompositeStateStore) SetLatestVersion(version int64) error {
+	if err := s.cosmosStore.SetLatestVersion(version); err != nil {
+		return err
+	}
+	if s.evmStore != nil && s.config.WriteMode != config.CosmosOnlyWrite {
+		if err := s.evmStore.SetLatestVersion(version); err != nil {
+			s.logger.Error("failed to set EVM store latest version", "error", err)
+			// Non-fatal: EVM store is optimization layer
```
blindchaser, line 210: what happens if EVM data writes succeed in ApplyChangesetSync() but SetLatestVersion fails?

sei-db/state_db/ss/evm/store.go
```diff
@@ -198,8 +255,17 @@ func (s *EVMStateStore) Prune(version int64) error {
 	return nil
 }
 
-// Close closes all databases
+// Close closes all async channels, waits for workers to drain, then closes databases.
+// Safe to call multiple times.
 func (s *EVMStateStore) Close() error {
+	// Close all async channels to signal workers to stop (safe: only close once)
```
blindchaser, line 263: comment says "safe to call multiple times" but this is only safe for sequential calls? concurrent calls to Close() can panic on double-close of a channel, not sure is there any conccurent calls cases though

Kbhat1: Close() is only ever called once at shutdown (by the deferred CompositeStateStore.Close())

sei-db/state_db/ss/composite/store.go
Kbhat1: If data is written but the version metadata doesn't advance, then on restart WAL replay sees the EVM store is "behind" (stale version), replays those entries again (writes are idempotent / no-op for already-written data), and sets the version

PR #2757, opened 2026-01-23
Composite State Store part 4: Dual-write path implementation

Kbhat1:

- ApplyChangesetSync / ApplyChangesetAsync: Parallel writes to both stores
- EVM changes extracted via commonevm.ParseMemIAVLEVMKey
- Import/RawImport: Fan out snapshot data to both stores
- Idempotent writes: If either store fails, caller can safely retry

- Unit tests

### DIFF
--- a/sei-db/state_db/ss/composite/store.go
+++ b/sei-db/state_db/ss/composite/store.go
@@ -19,8 +19,8 @@ import (
 )
 
 // CompositeStateStore routes operations between Cosmos_SS (main state store) and EVM_SS (optimized EVM stores).
-// - Reads check EVM_SS first for EVM keys (if EnableRead), then fallback to Cosmos_SS
-// - Writes go to both stores for EVM keys (if EnableWrite), only Cosmos_SS for others
+// - Reads check EVM_SS first for EVM keys, then fallback to Cosmos_SS
+// - Writes routing controlled by WriteMode (cosmos_only, dual_write, split_write)
 type CompositeStateStore struct {
 	cosmosStore types.StateStore   // Main MVCC PebbleDB for all modules
 	evmStore    *evm.EVMStateStore // Separate EVM DBs with default comparer (nil if disabled)
@@ -71,7 +71,7 @@ func NewCompositeStateStore(
 			return nil, fmt.Errorf("failed to create EVM store: %w", err)
 		}
 		cs.evmStore = evmStore
-		log.Info("EVM state store enabled", "dir", evmDir, "read", evmConfig.EnableRead, "write", evmConfig.EnableWrite)
+		log.Info("EVM state store enabled", "dir", evmDir, "writeMode", evmConfig.WriteMode)
 	}
 
 	// Recover from WAL if needed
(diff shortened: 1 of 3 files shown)


sei-db/state_db/ss/composite/store.go
```diff
@@ -160,45 +163,243 @@ func (s *CompositeStateStore) Close() error {
...
+	if evmErr != nil {
+		return fmt.Errorf("evm store failed: %w", evmErr)
+	}
+	return nil
 }
 
-// Import imports initial state
+// extractEVMChanges extracts EVM-routable changes from changesets
+func (s *CompositeStateStore) extractEVMChanges(changesets []*proto.NamedChangeSet) map[evm.EVMStoreType][]*iavl.KVPair {
+	evmChanges := make(map[evm.EVMStoreType][]*iavl.KVPair, 5) // 5 EVM store types
```
jewei1997: should probably make this a constant and document what 5 EVM store types there are.

sei-db/state_db/ss/composite/store.go
```diff
@@ -160,45 +163,243 @@ func (s *CompositeStateStore) Close() error {
...
 func (s *CompositeStateStore) Import(version int64, ch <-chan types.SnapshotNode) error {
-	// TODO: Add dual-write to EVM_SS in next PR
-	return s.cosmosStore.Import(version, ch)
+	if s.evmStore == nil {
+		return s.cosmosStore.Import(version, ch)
+	}
+
+	// Fan out to both stores
+	cosmosCh := make(chan types.SnapshotNode, 100)
+	evmChanges := make(map[evm.EVMStoreType][]*iavl.KVPair, 5)
```
jewei1997: same here

sei-db/state_db/ss/composite/store.go
```diff
@@ -147,45 +151,243 @@ func (s *CompositeStateStore) Close() error {
...
+	return nil
 }
 
-// ApplyChangesetSync applies changeset synchronously (delegates to Cosmos store)
+// ApplyChangesetSync applies changeset synchronously to both stores in parallel.
+// If either fails, returns error - caller should retry (writes are idempotent).
 func (s *CompositeStateStore) ApplyChangesetSync(version int64, changesets []*proto.NamedChangeSet) error {
-	// TODO: Add dual-write to EVM_SS in next PR
-	return s.cosmosStore.ApplyChangesetSync(version, changesets)
+	// Fast path: if no EVM store, just apply to Cosmos
```
yzang2019: We don't always wanna dual write, I feel it's better to have this controlled by a config or write mode in the future.

yzang2019 (approved):

sei-db/state_db/ss/composite/store.go
yzang2019: Actually it might be fine to keep the current logic, ignore the previous comment, we can still have various configs for read and write strategy/mode, which would determine whether we set the store to nil or not.

Kbhat1: Discussed this, I can add this in after your PR with the writeMode goes in

Kbhat1: This was updated

PR #2758, opened 2026-01-26
default SSTORE value back to original 20k

jewei1997:
change SSTORE chain param value back to original 20k. We may change this in the future with a gov proposal.

existing tests


precompiles/pointer/pointer_test.go
```diff
@@ -56,7 +56,7 @@ func TestAddNative(t *testing.T) {
 	evm = vm.NewEVM(*blockCtx, statedb, cfg, vm.Config{}, testApp.EvmKeeper.CustomPrecompiles(ctx))
 	ret, g, err := p.RunAndCalculateGas(evm, caller, caller, append(p.GetExecutor().(*pointer.PrecompileExecutor).AddNativePointerID, args...), suppliedGas, nil, nil, false, false)
 	require.Nil(t, err)
-	require.Equal(t, uint64(8679314), g)
+	require.Equal(t, uint64(8887314), g)
```
masih, line 59: Sanity checking that this is supposed to become 208K less?

PR #2759, opened 2026-01-26
default SSTORE value back to original 20k

jewei1997:

backport of this PR https://github.com/sei-protocol/sei-chain/pull/2758

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -4,4 +4,11 @@ out = "contracts/out"
 test = "contracts/test"
 libs = ["contracts/lib"]
 
+# CI profile mirrors default to satisfy CI runners that select it explicitly.
+[profile.ci]
+src = "contracts/src"
+out = "contracts/out"
+test = "contracts/test"
+libs = ["contracts/lib"]
+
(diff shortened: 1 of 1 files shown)


precompiles/pointer/pointer_test.go
```diff
@@ -56,7 +56,7 @@ func TestAddNative(t *testing.T) {
 	evm = vm.NewEVM(*blockCtx, statedb, cfg, vm.Config{}, testApp.EvmKeeper.CustomPrecompiles(ctx))
 	ret, g, err := p.RunAndCalculateGas(evm, caller, caller, append(p.GetExecutor().(*pointer.PrecompileExecutor).AddNativePointerID, args...), suppliedGas, nil, nil, false, false)
 	require.Nil(t, err)
-	require.Equal(t, uint64(8679314), g)
+	require.Equal(t, uint64(8887314), g)
```
stevenlanders, line 59: thinking - why would this be _higher_ ?

jewei1997: i dont think this is gas used. I think it's gas remaining, so the remaining gas is higher, which means it consumed less gas.

PR #2768, opened 2026-01-26
feat: Giga RPC node (OCC execution)

arajasek:
This is a followup to #2708 adding support for OCC mode to Giga RPC nodes. The logic is:
- Make a list of EVM txs and non-EVM txs
- Run EVM txs in parallel
  - If they fail due to interop, throw away all your work and just use the v2 logic for ALL txs
- Run non-EVM txs in parallel


app/app.go
```diff
@@ -1517,6 +1517,99 @@ func (app *App) ProcessTXsWithOCC(ctx sdk.Context, txs [][]byte, typedTxs []sdk.
...
+		for txIndex, tx := range txs {
+			nonevmEntries[txIndex] = app.GetDeliverTxEntry(ctx, txIndex, absoluteTxIndices[txIndex], tx, typedTxs[txIndex])
+		}
+	} else {
+		// Commit EVM cache to main store before processing non-EVM txs.
+		evmCache.Write()
+		evmCtx.GigaMultiStore().WriteGiga()
+	}
+
+	nonevmScheduler := tasks.NewScheduler(
```
codchen: nit: should we call this v2Scheduler since it may still process evm txs?

codchen (approved):

app/app.go
arajasek: sounds good, thank you

PR #2772, opened 2026-01-27
[giga] Fix gigacachekv.Store.Write() to preserve cache for reads

pdrobnjak:
## Summary
The `Write()` method was clearing the cache after writing dirty entries to the parent store. This caused issues because:

1. The parent store (`commitment.Store`) writes to a changeSet buffer
2. `commitment.Store.Get()` reads from the tree, not the changeSet
3. After clearing the cache, reads fell through to parent which couldn't return uncommitted data

**Fix:** Mark cache entries as clean (non-dirty) instead of clearing them. This preserves readability while still flushing data to parent for eventual commit.

This fixes state test failures where pre-state data was lost after `ProcessBlock` called `WriteGiga()`.

## Test plan
- [x] Existing state tests pass with this fix
- [ ] Verify cache behavior after multiple Write() calls

### DIFF
--- a/giga/deps/store/cachekv.go
+++ b/giga/deps/store/cachekv.go
@@ -110,7 +110,18 @@ func (store *Store) Write() {
 		}
 	}
 
-	store.cache = &sync.Map{}
+	// Mark all entries as clean (not dirty) instead of clearing the cache.
+	// This is important because the parent store (commitment.Store) doesn't make
+	// writes immediately visible until Commit(). By keeping the cache populated
+	// with clean entries, subsequent reads will still hit the cache instead of
+	// falling through to the parent which can't read uncommitted data.
+	store.cache.Range(func(key, value any) bool {
+		cv := value.(*types.CValue)
+		// Replace with a clean (non-dirty) version of the same value
+		store.cache.Store(key, types.NewCValue(cv.Value(), false))
+		return true
+	})
+	// Clear the deleted map since those deletes have been sent to parent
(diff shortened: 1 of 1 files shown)


giga/deps/store/cachekv.go
```diff
@@ -110,7 +110,18 @@ func (store *Store) Write() {
...
+	// Mark all entries as clean (not dirty) instead of clearing the cache.
+	// This is important because the parent store (commitment.Store) doesn't make
+	// writes immediately visible until Commit(). By keeping the cache populated
+	// with clean entries, subsequent reads will still hit the cache instead of
+	// falling through to the parent which can't read uncommitted data.
+	store.cache.Range(func(key, value any) bool {
+		cv := value.(*types.CValue)
+		// Replace with a clean (non-dirty) version of the same value
+		store.cache.Store(key, types.NewCValue(cv.Value(), false))
+		return true
```
pdrobnjak, line 122: **Minor consideration:** This iterates through all cache entries to mark them clean. For very large caches, this could be noticeable. An alternative approach would be to track dirty keys separately (similar to the `deleted` map) and only iterate those, but given this is only called during block processing, the current approach is likely fine.

Also worth noting: after this change, the cache will grow unbounded during a block's execution since entries are never cleared. If memory pressure becomes a concern, a follow-up could implement periodic cleanup of clean entries that match the parent store's committed state.

pdrobnjak:
Addressed by #2780.

PR #2774, opened 2026-01-27
Fixed flaky p2p test

pompon0:
MConnection.sendRoutine() was incorrectly returning DeadlineExceeded in case flushing timer returned at the moment the context was cancelled (context error should be returned instead).

### DIFF
--- a/sei-tendermint/internal/p2p/conn/connection_test.go
+++ b/sei-tendermint/internal/p2p/conn/connection_test.go
@@ -192,8 +192,8 @@ func TestMConnectionReadErrorUnknownChannel(t *testing.T) {
 
 		s.SpawnBg(func() error { return utils.IgnoreCancel(mconnClient.Run(ctx)) })
 		s.Spawn(func() error {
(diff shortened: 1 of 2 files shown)


sei-tendermint/internal/p2p/conn/connection_test.go
```diff
@@ -191,15 +191,12 @@ func TestMConnectionReadErrorUnknownChannel(t *testing.T) {
 		mconnServer := newMConnection(server)
 
 		s.SpawnBg(func() error { return utils.IgnoreCancel(mconnClient.Run(ctx)) })
-
-		// Use a channel to wait for server to complete, ensuring the message
-		// is actually sent before the context is cancelled.
-		serverDone := make(chan error, 1)
-		s.SpawnBg(func() error {
-			serverDone <- mconnServer.Run(ctx)
+		s.Spawn(func() error {
```
masih, line 194: The spawn, receive etc. utilities make it hard for me to reason about the code. I wish we could get rid of all of these micro utilities in favour of boring vanilla Go syntax.

PR #2779, opened 2026-01-27
Add scenario capability to benchmark script

stevenlanders:
## Summary

Adds configurable scenario support to the benchmark script, enabling ERC20 and other contract-based load testing scenarios.  (Fyi the benchmark won't really perform fully until the storage + autobahn changes are in, but this tooling now operates)

### New Benchmark Package (`app/benchmark/`)
- Moved benchmark logic from `app/benchmark.go` to a dedicated package
- Two-phase generator: Warmup → Setup (deploy contracts) → Load (generate transactions)
- Automatic contract deployment with address extraction from receipts
- Configurable via JSON scenario files

### Scenario Configuration (`scripts/scenarios/`)
- `erc20.json` - ERC20 token transfer benchmark (5000 accounts)
- `EVMTransfer.json` - Simple native transfers
- `mixed.json` - Combined workload example
- `README.md` - Documentation

### Mock Balance Fixes (`giga/deps/xbank/keeper/mock_balances.go`)
- Fixed OCC compatibility: mock balances now update supply as normal side effects
- Removed global address tracking that was interfering with OCC re-execution
- Works correctly with parallel transaction execution

### Script Updates (`scripts/benchmark.sh`)
- Added `BENCHMARK_CONFIG` env var to specify scenario file
- Default: `scripts/scenarios/EVMTransfer.json`

## Usage

# ERC20 benchmark with Giga Executor
GIGA_EXECUTOR=true BENCHMARK_CONFIG=scripts/scenarios/erc20.json ./scripts/benchmark.sh

### DIFF
--- a/app/benchmark/generator.go
+++ b/app/benchmark/generator.go
@@ -320,16 +320,8 @@ func (g *Generator) generateLoadBlock() []*abci.TxRecord {
 	}
 
 	loadTxs := loadGen.GenerateN(g.txsPerBatch)
-	if len(loadTxs) == 0 {
-		return nil
-	}
-
 	txRecords := make([]*abci.TxRecord, 0, len(loadTxs))
 	for _, loadTx := range loadTxs {
-		if loadTx.EthTx == nil {
-			continue
-		}
-
 		txRecord, err := g.ethTxToTxRecord(loadTx.EthTx)
 		if err != nil {
 			panic(fmt.Sprintf("benchmark: Failed to convert load tx: %v", err))


arajasek (approved):

app/benchmark/config.go
```diff
@@ -0,0 +1,45 @@
...
+// LoadConfig reads a sei-load config file or returns a default config.
+// The chainID and seiChainID are always overridden with actual values from the running chain.
+func LoadConfig(configPath string, evmChainID int64, seiChainID string) (*config.LoadConfig, error) {
+	if configPath == "" {
+		// Return default config (EVMTransfer scenario)
+		return &config.LoadConfig{
+			ChainID:    evmChainID,
+			SeiChainID: seiChainID,
+			MockDeploy: true, // We handle deployment in-process
+			Accounts:   &config.AccountConfig{Accounts: 5000},
```
wen-coding, line 20: nit: I assume number of Accounts can maybe be configured later?

app/benchmark/generator.go
```diff
@@ -0,0 +1,553 @@
...
+	PhaseWarmup Phase = iota
+	// PhaseSetup is the phase where contracts are deployed.
+	PhaseSetup
+	// PhaseLoad is the main phase where load transactions are generated.
+	PhaseLoad
+)
+
+// WarmupBlocks is the number of blocks to wait before starting setup.
+// This allows the chain to fully initialize (EVM genesis, fee collector address, etc.)
+const WarmupBlocks = 3
```
wen-coding, line 38: Is it possible you need a different value here when the number of accounts increase?

app/benchmark/generator.go
```diff
@@ -0,0 +1,553 @@
...
+
+	// Create shared account pool
+	if cfg.Accounts != nil {
+		accounts := loadtypes.GenerateAccounts(cfg.Accounts.Accounts)
+		bg.sharedAccounts = loadtypes.NewAccountPool(&loadtypes.AccountConfig{
+			Accounts:       accounts,
+			NewAccountRate: cfg.Accounts.NewAccountRate,
+		})
+		bg.accountPools = append(bg.accountPools, bg.sharedAccounts)
+	}
```
wen-coding, line 105: Sorry for a naive question, when would cfg.Accounts be nil? How do you generate traffic with no shared account pool?

app/benchmark/generator.go
```diff
@@ -0,0 +1,553 @@
...
+		"chainID", g.chainID.String(),
+		"dataLen", len(deployData))
+
+	// Create deployment transaction with chain ID set
+	tx := &ethtypes.DynamicFeeTx{
+		ChainID:   g.chainID,
+		Nonce:     nonce,
+		To:        nil, // nil To address indicates contract creation
+		Value:     big.NewInt(0),
+		Gas:       5000000,                   // 5M gas limit for deployment (increased)
```
wen-coding, line 219: Would you ever need to change this constant and the constants below?

app/benchmark/generator.go
```diff
@@ -0,0 +1,553 @@
...
+		GasFeeCap: big.NewInt(1000000000000), // 1000 gwei (high to ensure inclusion)
+		Data:      deployData,
+	}
+
+	// Sign the transaction (use CancunSigner to match sei-load scenarios)
+	signer := ethtypes.NewCancunSigner(g.chainID)
+	signedTx, err := ethtypes.SignTx(ethtypes.NewTx(tx), signer, g.deployer.PrivKey)
+	if err != nil {
+		g.logger.Error("benchmark: Failed to sign deployment tx", "error", err)
+		return nil
```
wen-coding: I assume signing of transaction should never fail since you set up all the accounts yourself. Is this worth a panic rather than nil (confused with not needing deployment).

app/benchmark/generator.go
```diff
@@ -0,0 +1,553 @@
...
+		}
+
+		// Create deployment transaction (only happens once per scenario)
+		deployTx := g.createDeploymentTx(state)
+		if deployTx == nil {
+			// Scenario doesn't need deployment (e.g., EVMTransfer)
+			g.logger.Info("benchmark: Scenario doesn't need deployment, attaching with zero address",
+				"scenario", state.scenario.Name())
+			state.deployed = true
+			if err := state.scenario.Attach(g.cfg, common.Address{}); err != nil {
```
wen-coding, line 269: Is this a transient error meaning you should try again?

app/benchmark/generator.go
```diff
@@ -0,0 +1,553 @@
...
+		state.deployTx = deployTx
+		state.deployTxHash = deployTx.Hash()
+		g.pendingDeploys[deployTx.Hash()] = state
+
+		g.logger.Info("benchmark: Created deployment transaction (will only deploy once)",
+			"scenario", state.config.Name,
+			"txHash", deployTx.Hash().Hex())
+
+		// Convert to Cosmos SDK tx
+		txRecord, err := g.ethTxToTxRecord(deployTx)
```
wen-coding, line 284: Is this a transient error that when you try again you will succeed?

app/benchmark/generator.go
```diff
@@ -0,0 +1,553 @@
...
+	if len(g.pendingDeploys) == 0 {
+		allHandled := true
+		for _, state := range g.scenarios {
+			if !state.deployed && state.deployTxHash == (common.Hash{}) {
+				allHandled = false
+				break
+			}
+		}
+		if allHandled {
+			g.transitionToLoadPhase()
```
wen-coding: This means if one scenario failed we will not transition to load phase? Then I think if any of the above error is not transient we should panic immediately.

app/benchmark/generator.go
```diff
@@ -0,0 +1,553 @@
...
+	g.mu.RLock()
+	loadGen := g.loadGenerator
+	g.mu.RUnlock()
+
+	if loadGen == nil {
+		return nil
+	}
+
+	loadTxs := loadGen.GenerateN(g.txsPerBatch)
+	if len(loadTxs) == 0 {
```
wen-coding: Is this expected? There are no loadTxs for some scenario?

app/benchmark/generator.go
```diff
@@ -0,0 +1,553 @@
...
+		return nil
+	}
+
+	txRecords := make([]*abci.TxRecord, 0, len(loadTxs))
+	for _, loadTx := range loadTxs {
+		if loadTx.EthTx == nil {
```
wen-coding: When could this happen?

app/benchmark/config.go
stevenlanders: yeah later on in this file it overwrites this structure from the json (if it's defined).  5000 is the default
```
	var cfg config.LoadConfig
	if err := json.Unmarshal(data, &cfg); err != nil {
		return nil, err
	}
```

app/benchmark/generator.go
stevenlanders: so the accounts get created kind of lazily as txs get sent - this is just to mitigate a few of the other startup services (like rpc/ports)

app/benchmark/generator.go
stevenlanders: by default the load generator library generates a new account on every send if for some reason this is nil

app/benchmark/generator.go
stevenlanders: if we have a contract deploy scenario (right now we have ERC20, noop, and some basic examples) so this is just 'large enough' - this is more just for setup (it shouldn't generally change, because a 5M contract is kinda big)

app/benchmark/generator.go
stevenlanders: converting to panic

app/benchmark/generator.go
```diff
@@ -0,0 +1,553 @@
...
+		return nil, fmt.Errorf("failed to encode tx: %w", err)
+	}
+
+	return &abci.TxRecord{
+		Action: abci.TxRecord_UNMODIFIED,
+		Tx:     txbz,
+	}, nil
+}
+
+// ProcessReceipts handles receipts from FinalizeBlock to extract deployed addresses.
```
wen-coding, line 364: nit: FinalizeBlock sounds like it's the block containing results at the end...

app/benchmark/generator.go
stevenlanders: refactoring to make this more clear - this if a fast path exit scenario when there's no pending deploy txs

app/benchmark/generator.go
```diff
@@ -0,0 +1,553 @@
...
+func (g *Generator) ProcessReceipts(receipts map[common.Hash]*evmtypes.Receipt) {
```
wen-coding, line 365: nit: Can rename to something else which makes it obvious we are extracting deployed addresses?

app/benchmark/generator.go
stevenlanders: yeah this is a defensive AI apparently

app/benchmark/generator.go
```diff
@@ -0,0 +1,553 @@
...
+
+		addr := common.HexToAddress(receipt.ContractAddress)
+		g.logger.Info("benchmark: Contract deployed successfully",
+			"scenario", state.config.Name,
+			"address", addr.Hex(),
+			"txHash", txHash.Hex(),
+			"gasUsed", receipt.GasUsed)
+
+		// Attach the deployed address to the scenario
+		if err := state.scenario.Attach(g.cfg, addr); err != nil {
```
wen-coding, line 420: Can we set deployed to true if this happens?

app/benchmark/generator.go
```diff
@@ -0,0 +1,553 @@
...
+
+		if receipt.Status != uint32(ethtypes.ReceiptStatusSuccessful) {
+			g.logger.Error("benchmark: Deployment failed",
+				"scenario", state.config.Name,
+				"txHash", txHash.Hex(),
+				"status", receipt.Status,
+				"vmError", receipt.VmError,
+				"gasUsed", receipt.GasUsed)
+			// Remove from pending but don't mark as deployed
+			delete(g.pendingDeploys, txHash)
```
wen-coding, line 408: Will this make allDeployed always fail?

app/benchmark/generator.go
```diff
@@ -0,0 +1,553 @@
...
+// transitionToLoadPhase switches from setup to load generation.
+func (g *Generator) transitionToLoadPhase() {
+	g.logger.Info("benchmark: All scenarios deployed, transitioning to load phase")
+	g.phase = PhaseLoad
+
+	// Create weighted generator from deployed scenarios
+	weightedConfigs := make([]*generator.WeightedCfg, 0, len(g.scenarios))
+	for _, state := range g.scenarios {
+		if !state.deployed {
+			g.logger.Info("benchmark: Scenario not deployed, skipping", "scenario", state.config.Name)
```
wen-coding, line 458: I guess you could panic here as well, because we checked that all scenarios are deployed before entering this function.

app/benchmark/generator.go
stevenlanders: it cannot - just removed the check (i suspect the ai thought a *EthTx was worth checking)

app/benchmark/generator.go
```diff
@@ -0,0 +1,553 @@
...
+	hashes := make([]common.Hash, 0, len(g.pendingDeploys))
+	for hash := range g.pendingDeploys {
+		hashes = append(hashes, hash)
+	}
+	return hashes
+}
+
+// StartProposalChannel creates a channel that generates proposals.
+func (g *Generator) StartProposalChannel(ctx context.Context, logger *Logger) <-chan *abci.ResponsePrepareProposal {
+	ch := make(chan *abci.ResponsePrepareProposal, 100)
```
wen-coding, line 526: Why 100?

app/benchmark/logger.go
```diff
@@ -0,0 +1,227 @@
...
+	defer l.mx.Unlock()
+
+	// Initialize lastFlushTime on first increment (when blocks actually start processing)
+	if l.lastFlushTime.IsZero() {
+		l.lastFlushTime = time.Now()
+	}
+
+	l.txCount += count
+	l.blockCount++
+	if height > l.latestHeight {
```
wen-coding, line 53: what is height? When would height stay the same when blockCount incements?

app/benchmark/logger.go
```diff
@@ -0,0 +1,227 @@
...
+		l.latestHeight = height
+	}
+
+	// Calculate time difference between consecutive blocks
+	if !l.prevBlockTime.IsZero() {
```
wen-coding, line 58: Does this mean you will miss blockTime for first block?

app/benchmark/logger.go
```diff
@@ -0,0 +1,227 @@
...
+	defer l.mx.Unlock()
+	l.blockProcessStartTime = time.Now()
+}
+
+// EndBlockProcessing marks the end of block processing (at FinalizeBlock end) and records the duration.
+func (l *Logger) EndBlockProcessing() {
+	l.mx.Lock()
+	defer l.mx.Unlock()
+
+	if l.blockProcessStartTime.IsZero() {
```
wen-coding, line 93: This shouldn't normally happen right?

app/benchmark/logger.go
```diff
@@ -0,0 +1,227 @@
...
+		return
+	}
+
+	duration := time.Since(l.blockProcessStartTime)
+	if duration > l.maxBlockProcessTime {
+		l.maxBlockProcessTime = duration
+	}
+	l.totalBlockProcessTime += duration
+	l.blockProcessCount++
+	l.blockProcessStartTime = time.Time{} // Reset for next block
```
wen-coding, line 103: But this will be reset in StartBlockProcessing, do you mean some blocks will not call StartBlockProcessing ?

app/benchmark/logger.go
```diff
@@ -0,0 +1,227 @@
...
+type flushStats struct {
+	txCount           int64
+	blockCount        int64
+	latestHeight      int64
+	maxBlockTimeMs    int64
+	avgBlockTimeMs    int64
+	maxCommitTimeMs   int64
+	avgCommitTimeMs   int64
+	maxBlockProcessMs int64
+	avgBlockProcessMs int64
```
wen-coding, line 133: Should any of this ever be negative?

app/invariance.go
```diff
@@ -241,9 +246,13 @@ func (app *App) LightInvarianceTotalSupply(cms sdk.CommitMultiStore) {
 	if !weiDiffRemainder.IsZero() {
 		panic(fmt.Sprintf("non-zero wei diff found! Pre-block wei total %s, post-block wei total %s", weiPreTotal, weiPostTotal))
 	}
+	// Formula: useiDiff = useiPreTotal - useiPostTotal - weiDiffInUsei + supplyChanged
+	// If money is conserved, this should be zero
+	// useiPreTotal - useiPostTotal = how much usei left balances (negative means usei entered balances)
+	// weiDiffInUsei = how much usei was moved to wei balances
+	// supplyChanged = how much new usei was minted
```
wen-coding, line 253: How is this comment related tothis PR?

app/invariance.go
```diff
@@ -241,9 +246,13 @@ func (app *App) LightInvarianceTotalSupply(cms sdk.CommitMultiStore) {
...
 	useiDiff := useiPreTotal.Sub(useiPostTotal).Sub(weiDiffInUsei).Add(supplyChanged)
 	if !useiDiff.IsZero() {
 		panic(fmt.Sprintf("unexpected usei balance total found! Pre-block usei total %s wei total %s total supply %s, post-block usei total %s wei total %s total supply %s", useiPreTotal, weiPreTotal, preTotalSupply, useiPostTotal, weiPostTotal, preTotalSupply.Add(supplyChanged)))
 	}
-	app.Logger().Info("successfully verified supply light invariance")
```
wen-coding, line 248: Removed because this isn't a useful log?

giga/deps/xevm/state/mock_balances.go
```diff
@@ -24,177 +20,54 @@ import (
...
-
-	ctx := s.ctx
-
-	// this avoids emitting cosmos events for ephemeral bookkeeping transfers like send_native
-	if s.eventsSuppressed {
-		ctx = ctx.WithEventManager(sdk.NewEventManager())
+// ensureMinimumBalance tops off the account if balance is low.
+// Called from GetBalance to ensure preCheck passes in StateTransition.
+func (s *DBImpl) ensureMinimumBalance(evmAddr common.Address) {
+	if s.ctx.ChainID() == "pacific-1" {
```
wen-coding, line 32: I think you use `evmcfg.IsLiveEVMChainID(evmChainID)` elsewhere?

giga/deps/xevm/state/mock_balances.go
```diff
@@ -24,177 +20,54 @@ import (
...
-	return s.k.GetSeiAddressOrDefault(s.ctx, evmAddr)
-}
-
-func (s *DBImpl) send(from sdk.AccAddress, to sdk.AccAddress, amt *big.Int) {
-	usei, wei := SplitUseiWeiAmount(amt)
-	err := s.k.BankKeeper().SendCoinsAndWei(s.ctx, from, to, usei, wei)
-	if err != nil {
-		s.err = err
-	}
+	_ = s.k.BankKeeper().SendCoinsAndWei(ctx, moduleAddr, seiAddr, usei, wei)
```
wen-coding, line 72: Should you log this error maybe?

scripts/benchmark.sh
```diff
@@ -211,4 +220,10 @@ echo "To capture heap profile:"
 echo "  go tool pprof http://localhost:6060/debug/pprof/heap"
 echo "============================================================"
 echo ""
-BENCHMARK_TXS_PER_BATCH=$BENCHMARK_TXS_PER_BATCH ~/go/bin/seid start --chain-id sei-chain 2>&1 | grep benchmark
+if [ "$DEBUG" = true ]; then
+  # Debug mode: print all output
+  BENCHMARK_CONFIG=$BENCHMARK_CONFIG BENCHMARK_TXS_PER_BATCH=$BENCHMARK_TXS_PER_BATCH ~/go/bin/seid start --chain-id sei-chain
```
wen-coding, line 225: nit: could chain-id uses something more obvious it's a benchmark?

app/benchmark/generator.go
wen-coding: That seems a bit inefficient, is that really frequently in use?

PR #2785, opened 2026-01-28
Composite State Store Part 5: WAL integration + Composite Pruning Manager

Kbhat1:

- Add `RecoverCompositeStateStore` to handle WAL replay for both stores on startup
- Refactor pruning to go through CompositeStateStore so both Cosmos_SS and EVM_SS are pruned correctly
- If `EVM_SS` is behind (e.g., newly enabled or crash), sync it from WAL before standard recovery
- Standard WAL replay uses `ApplyChangesetSync` which dual-writes to both stores automatically
- Backward search for replay offset matches existing `RecoverStateStore` pattern
- Both stores stay consistent after crashes/restarts

- Unit tests
- Verifying on node

### DIFF
--- a/sei-db/state_db/ss/evm/store.go
+++ b/sei-db/state_db/ss/evm/store.go
@@ -123,15 +123,23 @@ func (s *EVMStateStore) ApplyChangesetParallel(version int64, changes map[EVMSto
 	return nil
 }
 
-// GetLatestVersion returns the maximum latest version across all databases
+// GetLatestVersion returns the minimum latest version across all databases.
+// Using min ensures crash-recovery correctness: if a crash interrupts
+// SetLatestVersion mid-loop, some DBs may be one version ahead of others.
+// min() guarantees WAL replay starts from the furthest-behind DB, and
+// replaying to an already-current DB is harmless (writes are idempotent).
+// Under normal operation all DBs are at the same version, so min == max.
 func (s *EVMStateStore) GetLatestVersion() int64 {
-	var maxVersion int64
+	var minVersion int64 = -1
 	for _, db := range s.databases {
-		if v := db.GetLatestVersion(); v > maxVersion {
-			maxVersion = v
+		if v := db.GetLatestVersion(); minVersion < 0 || v < minVersion {
+			minVersion = v
 		}
 	}
-	return maxVersion
+	if minVersion < 0 {
+		return 0
+	}
+	return minVersion
 }
 
 // SetLatestVersion sets the latest version on all databases


sei-db/state_db/ss/composite/store.go
```diff
@@ -492,3 +492,202 @@ func (s *CompositeStateStore) Prune(version int64) error {
...
+
+	if startOffset > lastOffset {
+		logger.Info("No WAL entries to replay for EVM sync")
+		return nil
+	}
+
+	logger.Info("Replaying WAL to EVM store", "startOffset", startOffset, "endOffset", lastOffset)
+
+	return streamHandler.Replay(startOffset, lastOffset, func(index uint64, entry proto.ChangelogEntry) error {
+		if entry.Version > cosmosVersion {
```
blindchaser: optional: logging wal replay progress something like: 

`if index % 1000 == 0 {
        logger.Info("WAL replay progress", "offset", index, "version", entry.Version)
    }`

sei-db/state_db/ss/composite/store.go
```diff
@@ -492,3 +492,202 @@ func (s *CompositeStateStore) Prune(version int64) error {
...
+
+	evmVersion := compositeStore.evmStore.GetLatestVersion()
+	logger.Info("Recovering CompositeStateStore",
+		"cosmosVersion", cosmosVersion,
+		"evmVersion", evmVersion,
+		"changelogPath", changelogPath,
+	)
+
+	// Sync EVM_SS if behind
+	if evmVersion < cosmosVersion {
```
yzang2019: I think we should not compare EVM version with cosmos Version to determine behind or not, because both SS could be falling behind at the same time even if they are at the same height

sei-db/state_db/ss/composite/store.go
```diff
@@ -492,3 +492,202 @@ func (s *CompositeStateStore) Prune(version int64) error {
...
+			logger.Error("Failed to sync EVM store, continuing with degraded performance",
+				"error", err, "evmVersion", evmVersion, "cosmosVersion", cosmosVersion)
+		}
+	}
+
+	return recoverStandardWAL(logger, changelogPath, compositeStore, cosmosVersion)
+}
+
+// syncEVMStore replays WAL entries to catch up EVM_SS to Cosmos_SS version
+func syncEVMStore(
```
yzang2019: Can we consolidate the changelog restore logic into one function?

sei-db/state_db/ss/composite/store.go
Kbhat1: @yzang2019 The logic handles both cases:

1. `syncEVMStore` only runs if EVM_SS < Cosmos_SS (catches EVM up to Cosmos)
2. `recoverStandardWAL` always runs after and replays through `CompositeStateStore`, which dual-writes to both stores

So if both are at version 80 and WAL is at 100, we skip step 1 (they're equal), then step 2 replays 81-100 to both stores via dual-write. The two-phase approach covers all scenarios.

yzang2019: Hmm why do we need separate steps here? Would it be easier to just recover both SS till latest height?

PR #2786, opened 2026-01-29
Consolidate SC configurations and interface for Giga

yzang2019:
This PR unifies and consolidate the configuration and interface for committer (CommitStore) by:
- Introduce MemIAVLConfig and put that under memIAVL package
- Introduce FlatKVConfig and put that under flatkv package
- Consolidate them into one StateCommitConfig and would be used by composite store
- Added a new CompositeCommitStore which handles both cosmos and evm backends based on read and write mode
- Moved original cosmos CommitStore to memIAVL package (Still using this one for root multistore)
- Added bunch of todos

### DIFF
--- a/sei-db/state_db/sc/memiavl/snapshot_catchup_test.go
+++ b/sei-db/state_db/sc/memiavl/snapshot_catchup_test.go
@@ -16,7 +16,7 @@ func TestSnapshotTimeThrottling(t *testing.T) {
 
 	db, err := OpenDB(logger.NewNopLogger(), 0, Options{
 		Config: Config{
-			SnapshotInterval:        100,        // Small interval for testing
+			SnapshotInterval:        100,     // Small interval for testing
 			SnapshotMinTimeInterval: 60 * 60, // 1 hour minimum time interval (in seconds)
 		},
 		Dir:             dir,


sei-db/config/read_mode.go
```diff
@@ -0,0 +1,40 @@
...
+	// This is the default/legacy behavior.
+	CosmosOnlyRead ReadMode = "cosmos_only"
+
+	// EVMFirstRead reads EVM data from EVM backend first, falls back to Cosmos if not found.
+	// Use during migration to test EVM backend reads while maintaining Cosmos as fallback.
+	EVMFirstRead ReadMode = "evm_first"
+
+	// SplitRead reads EVM data from EVM backend and non-EVM data from Cosmos.
+	// Use when migration is complete and backends are fully separated.
+	SplitRead ReadMode = "split_read"
```
blindchaser, line 19: how about naming it SplitByKeyRead. 

and do we still have split read use case if we choose migrating ALL EVM state over to Giga DB?

sei-db/config/write_mode.go
```diff
@@ -0,0 +1,40 @@
...
+		return false
+	}
+}
+
+// ParseWriteMode converts a string to a WriteMode, returning an error if invalid
+func ParseWriteMode(s string) (WriteMode, error) {
+	m := WriteMode(s)
+	if !m.IsValid() {
+		return "", fmt.Errorf("invalid write mode: %q, valid modes: %s, %s, %s, %s",
+			s, CosmosOnlyWrite, DualWrite, SplitWrite)
```
blindchaser: this s a good catch from ai

sei-db/state_db/sc/memiavl/store.go
```diff
@@ -1,52 +1,37 @@
...
-		PrefetchThreshold:                config.SnapshotPrefetchThreshold,
-		CreateIfMissing:                  true,
-		OnlyAllowExportOnSnapshotVersion: config.OnlyAllowExportOnSnapshotVersion,
+func NewCommitStore(homeDir string, logger logger.Logger, config Config) *CommitStore {
+	commitDBPath := utils.GetCommitStorePath(homeDir)
+	opts := Options{
+		Config:          config, // Embed the config directly
+		Dir:             commitDBPath,
+		CreateIfMissing: true,
+		ZeroCopy:        true,
```
blindchaser, line 29: confirm that we want the ZeroCopy always to be true?

sei-db/state_db/sc/composite/store.go
```diff
@@ -0,0 +1,235 @@
...
+// Being used for two scenario:
+// ReadOnly: Either for state sync or for historical proof
+// Writable: Opened during initialization for root multistore
+func (cs *CompositeCommitStore) LoadVersion(targetVersion int64, readOnly bool) (types.Committer, error) {
+	cosmosSC, err := cs.cosmosCommitter.LoadVersion(targetVersion, readOnly)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load cosmos version: %w", err)
+	}
+
+	return cosmosSC, nil
```
blindchaser: the return type of `func  LoadVersion` is `types.Committer`, and we always return a cosmosCommitter here?

sei-db/config/read_mode.go
yzang2019: Still need split read, it means EVM goes from EVM and cosmos goes to cosmos

sei-db/state_db/sc/memiavl/store.go
yzang2019: Yup

sei-db/config/write_mode.go
yzang2019: True, should be fixed

sei-db/state_db/sc/composite/store.go
yzang2019: You are right, here we should return a composite store

yzang2019:
> Configs: StateCommitConfig no longer has AsyncCommitBuffer, SnapshotKeepRecent, etc. (they moved under MemIAVLConfig), but the template still renders .StateCommit.*. WriteConfigFile will panic at runtime when rendering.
> 
> need to update the toml.go, app/seidb.go / sei-cosmos/server/config/config.go etc

We don't need to change the template because it's rendered by cosmos, and cosmos config is updated to use the new config 

> Configs: StateCommitConfig no longer has AsyncCommitBuffer, SnapshotKeepRecent, etc. (they moved under MemIAVLConfig), but the template still renders .StateCommit.*. WriteConfigFile will panic at runtime when rendering.
> 
> need to update the toml.go, app/seidb.go / sei-cosmos/server/config/config.go etc

That's a good catch! Will fix the toml config name

Kbhat1 (approved):

PR #2788, opened 2026-02-02
feat: add ledger cache layer for receipt store

jewei1997:
## Summary

This PR introduces an in-memory cache layer on top of the pebbledb receipt store backend, improving `GetReceipt` performance for recently accessed receipts.

- Add `cachedReceiptStore` wrapper that caches receipts in rotating chunks
- Simplify `FilterLogs` API: change from per-block signature to range-based `(fromBlock, toBlock, crit)`
- Pebble backend's `FilterLogs` returns `ErrRangeQueryNotSupported`, signaling callers to fetch receipts individually
- Update `evmrpc/filter.go` with fallback logic for backends that don't support range queries
- Simplify `ReceiptStoreConfig` by removing unused pebble options

The cache rotates every 500 blocks (configurable) and keeps 3 chunks, providing a sliding window of recent receipts for fast lookups.

## Test plan

- [x] Receipt store unit tests pass
- [x] Cached receipt store tests verify cache hits
- [x] FilterLogs returns `ErrRangeQueryNotSupported` for pebble backend

PR #2789, opened 2026-02-02
feat: add parquet receipt store with DuckDB range queries

jewei1997:
## Summary

This PR adds a parquet-based receipt storage backend with DuckDB for efficient range queries on logs, enabling fast `eth_getLogs` queries across block ranges.

- Add parquet backend option (`Backend: "parquet"` in config)
- Parquet files rotate every 500 blocks
- DuckDB queries across closed parquet files for efficient log filtering
- WAL for crash recovery of in-progress parquet files
- Pruning of old parquet files based on `KeepRecent` config
- Build tag support: use `-tags duckdb` to enable parquet backend

The parquet backend supports the new `FilterLogs` range query API introduced in #2788, enabling efficient cross-block log queries without falling back to per-receipt fetching.

## Dependencies

- Depends on #2788 (ledger cache layer)

## Test plan

- [x] Receipt store unit tests pass (without duckdb tag)
- [ ] Parquet store tests pass with `-tags duckdb`
- [ ] Integration testing with full node using parquet backend

### DIFF
--- a/sei-db/ledger_db/parquet/reader.go
+++ b/sei-db/ledger_db/parquet/reader.go
@@ -43,13 +43,21 @@ func NewReader(basePath string) (*Reader, error) {
 	db.SetMaxOpenConns(numCPU * 2)
 	db.SetMaxIdleConns(numCPU)
 
-	_, _ = db.Exec(fmt.Sprintf("SET threads TO %d", numCPU))
-	_, _ = db.Exec("SET memory_limit = '1GB'")
-	_, _ = db.Exec("SET enable_object_cache = true")
-	_, _ = db.Exec("SET parquet_metadata_cache_size = 500")
-	_, _ = db.Exec("SET access_mode = 'READ_ONLY'")
-	_, _ = db.Exec("SET enable_progress_bar = false")
-	_, _ = db.Exec("SET preserve_insertion_order = false")
+	settings := []string{
+		fmt.Sprintf("SET threads TO %d", numCPU),
+		"SET memory_limit = '1GB'",
+		"SET enable_object_cache = true",
+		"SET parquet_metadata_cache_size = 500",
+		"SET access_mode = 'READ_ONLY'",
+		"SET enable_progress_bar = false",
+		"SET preserve_insertion_order = false",
+	}
+	for _, statement := range settings {
+		if _, err = db.Exec(statement); err != nil {
+			_ = db.Close()
+			return nil, fmt.Errorf("failed to configure duckdb (%s): %w", statement, err)
+		}
+	}
 
 	reader := &Reader{
 		db:       db,


sei-db/ledger_db/receipt/receipt_store.go
```diff
@@ -86,6 +91,11 @@ func newReceiptBackend(log dbLogger.Logger, config dbconfig.ReceiptStoreConfig,
 
 	backend := normalizeReceiptBackend(config.Backend)
 	switch backend {
+	case receiptBackendParquet:
+		if !ParquetEnabled() {
```
yzang2019: do we need this extra check at all? Wonder in what situation will we disable Parquet but also set backend to parquet? I think we should only need to keep one

sei-db/ledger_db/receipt/parquet_store.go
```diff
@@ -0,0 +1,817 @@
+//go:build duckdb
```
yzang2019: Curious why do we need to add this build flag? Is there any extra steps for setting up duckdb? Can we just enable Parquet via config?

jewei1997: fixed

sei-db/ledger_db/receipt/receipt_store.go
jewei1997: fixed

sei-db/ledger_db/parquet/reader.go
```diff
@@ -0,0 +1,373 @@
...
+	db.SetMaxOpenConns(numCPU * 2)
+	db.SetMaxIdleConns(numCPU)
+
+	_, _ = db.Exec(fmt.Sprintf("SET threads TO %d", numCPU))
+	_, _ = db.Exec("SET memory_limit = '1GB'")
+	_, _ = db.Exec("SET enable_object_cache = true")
+	_, _ = db.Exec("SET parquet_metadata_cache_size = 500")
+	_, _ = db.Exec("SET access_mode = 'READ_ONLY'")
+	_, _ = db.Exec("SET enable_progress_bar = false")
+	_, _ = db.Exec("SET preserve_insertion_order = false")
```
blindchaser: do we want to ignore all errors here?

sei-db/ledger_db/receipt/receipt_cache.go
```diff
@@ -214,3 +215,68 @@ func (c *ledgerCache) AddLogsForBlock(blockNumber uint64, logs []*ethtypes.Log)
...
+					result = append(result, &logCopy)
+				}
+			}
+		}
+	}
+	return result
+}
+
+// matchLog checks if a log matches the filter criteria.
+func matchLog(lg *ethtypes.Log, crit filters.FilterCriteria) bool {
```
blindchaser, line 246: are matchLog / matchesCriteria / isLogExactMatch having overlap or dup?

sei-db/ledger_db/parquet/store.go
```diff
@@ -0,0 +1,600 @@
...
+	s.receiptsBuffer = s.receiptsBuffer[:0]
+	s.logsBuffer = s.logsBuffer[:0]
+	return nil
+}
+
+func (s *Store) closeWritersLocked() error {
+	var errs []error
+
+	if s.receiptWriter != nil {
+		if err := s.receiptWriter.Close(); err != nil {
```
blindchaser, line 498: from ai: closeWritersLocked() calls file.Close() but not file.Sync(). Go's os.File.Close() does not guarantee fsync — data may still be in the OS page cache. Since ClearWAL() runs immediately after in rotateFileLocked(), a crash between close and OS writeback would lose the rotated parquet file with no WAL to recover from.

Suggest adding file.Sync() before file.Close() in closeWritersLocked():

`
func (s *Store) closeWritersLocked() error {
    // ...close writers...
    if s.receiptFile != nil {
        _ = s.receiptFile.Sync()
        if err := s.receiptFile.Close(); err != nil { ... }
    }
    if s.logFile != nil {
        _ = s.logFile.Sync()
        if err := s.logFile.Close(); err != nil { ... }
    }
    // ...
}
`

blindchaser:
it seems every build pulls in DuckDB native libraries even when only the pebble backend is needed, is this expected?

PR #2790, opened 2026-02-03
giga allocation optimizations

codchen:
- optimize allocation in scheduler.go
- use byte array instead of calling Hex() every time for map keys in state/

refactor - existing tests should hold

### DIFF
--- a/giga/deps/tasks/scheduler.go
+++ b/giga/deps/tasks/scheduler.go
@@ -163,9 +163,22 @@ func (s *scheduler) DoExecute(work func()) {
 	s.executeCh <- work
 }
 
+// Pool for reusing conflict dedup maps
+var conflictMapPool = sync.Pool{
+	New: func() interface{} {
+		return make(map[int]struct{})
+	},
+}
+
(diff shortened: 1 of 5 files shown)


giga/deps/tasks/scheduler.go
```diff
@@ -163,9 +163,22 @@ func (s *scheduler) DoExecute(work func()) {
...
+}
+
 func (s *scheduler) findConflicts(task *deliverTxTask) (bool, []int) {
 	var conflicts []int
-	uniq := make(map[int]struct{})
+	uniq := conflictMapPool.Get().(map[int]struct{})
+	// Clear the map for reuse
+	for k := range uniq {
+		delete(uniq, k)
+	}
```
masih: You can simply use the built-in `clear` function, like this: `clear(uniq)`. 

I recommend doing a quick benchmark to gauge improvements.

stevenlanders: Nice!  I didn't know about that ^

arajasek (approved):

PR #2791, opened 2026-02-03
Autobahn migrated from sei-v3

pompon0:
Migrated consensus,avail,data,producer and types modules. Migrated corresponding tests. Replaced gRPC usage with custom rpc layer on top of mux. GigaRouter is still hardcoded to be disabled.

NOTE: there is a lot of missing parts and coverage, but this pr will allow us to stop development of autobahn in sei-v3, so that we can start making it production-ready in sei-chain.


sei-tendermint/light/example_test.go
```diff
@@ -47,17 +46,12 @@ func TestExampleClient(t *testing.T) {
...
-		block, err = primary.LightBlock(ctx, 2)
-		if err == nil {
-			break
-		}
-		if !errors.Is(err, provider.ErrHeightTooHigh) {
-			t.Fatal(err)
-		}
-		time.Sleep(time.Second)
+	t.Logf("wait for the blocks used in test")
+	utils.OrPanic1(waitForBlock(ctx, primary, 3))
```
wen-coding, line 50: Just curious, why do we need to change this test?

sei-tendermint/internal/autobahn/autobahn.proto
wen-coding, line 13: nit: comment what this does?

sei-tendermint/internal/autobahn/autobahn.proto
```diff
@@ -213,42 +204,3 @@ message ConsensusReq {
...
-message AppVote {
-  SignedMsg app_vote = 1;
-}
-
-message ConsensusResp {}
-
-message PingReq {}
-message PingResp {}
-
-message StreamLaneProposalsReq {
```
wen-coding: Hmm, I think this one is still used? Do you plan to do that in another PR? https://github.com/sei-protocol/sei-v3/blob/c7bb4236eb15014d7283103529632930a69fb619/stream/consensus/avail/client.go#L54

sei-tendermint/internal/p2p/mux/mux.go
```diff
@@ -202,12 +202,19 @@ func (r *runner) runSend(ctx context.Context, conn conn.Conn) error {
...
-						ctrl.Updated()
+					if s, ok := inner.streams[id]; ok {
+						for sInner, ctrl := range s.inner.Lock() {
+							sInner.send.bufBegin += 1
+							ctrl.Updated()
+						}
 					}
 				}
 			}
+			// TODO(gprusak): this is counterintuitive asymmetric behavior:
```
wen-coding, line 213: Sorry I'm confused, didn't find this in v3 code, Is this migrated from some other repo?

sei-tendermint/internal/p2p/giga/consensus.go
```diff
@@ -0,0 +1,140 @@
...
+		s.Spawn(func() error { return sendUpdates(ctx, c, x.state.SubscribeCommitVote()) })
+		s.Spawn(func() error { return sendUpdates(ctx, c, x.state.SubscribeTimeoutVote()) })
+		s.Spawn(func() error { return sendUpdates(ctx, c, x.state.SubscribeTimeoutQC()) })
+		return nil
+	})
+}
+
+// Ping implements pb.StreamAPIServer.
+// Note that we use streaming RPC, because unary RPC apparently causes 10ms extra delay on avg (empirically tested).
+func (x *Service) serverPing(ctx context.Context, server rpc.Server[API]) error {
```
wen-coding, line 90: And from here on is from stream/consensus/server.go?

Why don't we keep the original client/server separation?

sei-tendermint/internal/p2p/giga/avail.go
```diff
@@ -0,0 +1,255 @@
...
+	"errors"
+	"fmt"
+	"github.com/tendermint/tendermint/internal/autobahn/data"
+	apb "github.com/tendermint/tendermint/internal/autobahn/pb"
+	"github.com/tendermint/tendermint/internal/autobahn/types"
+	"github.com/tendermint/tendermint/internal/p2p/giga/pb"
+	"github.com/tendermint/tendermint/internal/p2p/rpc"
+)
+
+func (x *Service) serverStreamLaneProposals(ctx context.Context, server rpc.Server[API]) error {
```
wen-coding, line 14: So from here on it's stream/consensus/avail/server.go?

sei-tendermint/internal/p2p/giga/avail.go
```diff
@@ -0,0 +1,255 @@
...
+			}
+			next = qc.Index() + 1
+			if err := stream.Send(ctx, types.CommitQCConv.Encode(qc)); err != nil {
+				return fmt.Errorf("stream.Send(): %w", err)
+			}
+		}
+	})
+}
+
+func (x *Service) clientStreamLaneProposals(ctx context.Context, c rpc.Client[API]) error {
```
wen-coding, line 123: nit: If we have to put client and server in the same file, can we always do client first or always do server first maybe?

sei-tendermint/internal/autobahn/avail/queue.go
```diff
@@ -0,0 +1,33 @@
+package avail
+
```
wen-coding, line 2: Since queue sounds like a general enough data structure, can you maybe comment why we need it defined in avail/, not at some more generic location?

sei-tendermint/internal/autobahn/avail/subscriptions.go
```diff
@@ -0,0 +1,103 @@
+package avail
+
```
wen-coding, line 2: I believe this file is new, can you maybe explain a little bit why it's organized this way? How many receiver instances do we plan to have for each kind?

sei-tendermint/internal/autobahn/consensus/commit_votes.go
```diff
@@ -0,0 +1,66 @@
...
+	// Construct a CommitQC from the votes.
+	old := cv.qc.Load()
+	if old, ok := old.Get(); ok && !old.Proposal().View().Less(view) {
+		return
+	}
+	var votes []*types.Signed[*types.CommitVote]
+	for _, v := range cv.byHash[h] {
+		votes = append(votes, v)
+	}
+	cv.qc.Store(utils.Some(types.NewCommitQC(votes)))
```
wen-coding, line 65: Why are we changing from cv.qc.Update to cv.qc.Load and cv.qc.Store now?

sei-tendermint/internal/autobahn/data/testonly.go
```diff
@@ -0,0 +1,183 @@
...
+func TestLaneQC(keys []types.SecretKey, header *types.BlockHeader) *types.LaneQC {
+	vote := types.NewLaneVote(header)
+	votes := []*types.Signed[*types.LaneVote]{}
+	for _, k := range keys {
+		votes = append(votes, types.Sign(k, vote))
+	}
+	return types.NewLaneQC(votes)
+}
+
+func TestCommitQC(
```
wen-coding, line 33: Nice, I like that the helpers are all moved here.

sei-tendermint/internal/autobahn/producer/state.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/tendermint/tendermint/libs/utils/scope"
+	"golang.org/x/time/rate"
+)
+
+// Config is the config of the block scope.
+type Config struct {
+	MaxGasPerBlock   uint64
+	MaxTxsPerBlock   uint64
+	MaxTxsPerSecond  utils.Option[uint64]
+	MempoolSize      uint64
```
wen-coding, line 21: Do we still have a mempool in Giga?

sei-tendermint/internal/p2p/giga/consensus.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/tendermint/tendermint/internal/autobahn/types"
+	"github.com/tendermint/tendermint/internal/p2p/giga/pb"
+	"github.com/tendermint/tendermint/internal/p2p/rpc"
+	"github.com/tendermint/tendermint/libs/utils"
+	"github.com/tendermint/tendermint/libs/utils/scope"
+	"time"
+)
+
+// Sends a consensus message to the peer whenever atomic watch is updated.
+func sendUpdates[T interface {
```
wen-coding, line 16: Okay, this one is moved from stream/consensus/client.go?

sei-tendermint/light/example_test.go
pompon0: The test is flaky, it is not related to the autobahn in any sense.

sei-tendermint/internal/autobahn/autobahn.proto
pompon0: this is documented on the hashable package.

sei-tendermint/internal/p2p/mux/mux.go
pompon0: No this is my own code. I've found a bug when plugging the mux into autobahn, which is a result of this inconsistency.

sei-tendermint/internal/p2p/giga/consensus.go
pompon0: the client and server implementation are tightly coupled, given that we use streaming RPCs. However, the way I've distributed the code across the files is kinda arbitrary, we can cleanup it later. A lot of it will need to be moved around anyway.

sei-tendermint/internal/p2p/giga/avail.go
pompon0: yeah

sei-tendermint/internal/p2p/giga/avail.go
pompon0: sure, this is a work-in-progress layout, definitely will need a cleanup.

sei-tendermint/internal/autobahn/avail/queue.go
pompon0: my rule of thumb is to write general enough abstractions to avoid tangling logic. The question of visibility (where the code should be located, which packages should have access to it) is a totally separate question. So far this struct was under a lot of code churn, that's why it is in the package that is making use of it.

sei-tendermint/internal/autobahn/avail/subscriptions.go
pompon0: There is 1 receiver per connection currently. I wanted to separate autobahn from the transport layer. Autobahn api surface is way too large atm (number of receivers, custom receivers for each consensus message type, proper buffer sizes in rpcs, etc), it will require some work to simplify it. Or we can fallback to merge back the transport layer into autobahn (which is currently under p2p/giga). This is TBD.

sei-tendermint/internal/autobahn/producer/state.go
pompon0: the block production pipeline in giga needs to be designed and implemented. I've copied over the producer for reference.

sei-tendermint/internal/autobahn/consensus/commit_votes.go
pompon0: thank for reminding me. The AtomicWatch has evolved since I implemented it in sei-v3. Now it is separated into AtomicSend and AtomicRecv and does not provide atomic updated by default - you need to wrap AtomicSend into a mutex - it simplifies the AtomicSend api AND allows AtomicSend to share mutex with other data, which is usually very convenient.

pompon0: I've left a bunch of TODOs, because this change introduces some race conditions. Let me address them.

sei-tendermint/internal/p2p/giga/consensus.go
pompon0: yes

sei-tendermint/internal/autobahn/consensus/commit_votes.go
pompon0: fixed

sei-tendermint/internal/autobahn/consensus/inner.go
```diff
@@ -23,22 +23,23 @@ func (i *inner) View() types.View {
...
 		return nil
 	}
 	if err := qc.Verify(s.Data().Committee()); err != nil {
 		return fmt.Errorf("qc.Verify(): %w", err)
 	}
-	i := s.inner.Load()
-	if qc.Proposal().Index() < types.NextIndexOpt(i.viewSpec.CommitQC) {
-		return nil
+	for i := range s.inner.Lock() {
+		if qc.Proposal().Index() < types.NextIndexOpt(i.Load().viewSpec.CommitQC) {
```
wen-coding: nit: is this the same check as the one on line 26:
`i.viewSpec.View().Index > qc.Proposal().Index()`

Would help readability if we use the same expression to show that we are just rechecking the same condition.

sei-tendermint/internal/autobahn/consensus/state.go
```diff
@@ -61,11 +62,13 @@ func (s *State) SubscribeTimeoutQC() utils.AtomicRecv[utils.Option[*types.Timeou
 
 // NewState constructs a new state.
 func NewState(cfg *Config, data *data.State) *State {
+	inner := utils.Alloc(utils.NewAtomicSend(inner{}))
 	return &State{
 		cfg: cfg,
 		// metrics: NewMetrics(),
```
wen-coding, line 68: We do plan to add metrics back right?

pompon0: yes, we need to adapt them to the sei-chain metrics

sei-tendermint/internal/autobahn/consensus/inner.go
pompon0: done

PR #2792, opened 2026-02-03
CompositeStateStore Part 6: StateStore legacy db

Kbhat1:

Add a catch-all `LegacyDB` to EVM-SS that stores all EVM module keys that don't fit into the optimized databases (Nonce, CodeHash, Code, Storage).

This ensures ALL EVM data is routed to EVM-SS:
- Optimized keys (nonce, code, codehash, storage) go to specialized DBs
- All other EVM keys (address mappings, receipts, pointers, bloom, etc.) go to LegacyDB with the full original key preserve

### DIFF
--- a/sei-db/common/evm/keys.go
+++ b/sei-db/common/evm/keys.go
@@ -21,48 +21,50 @@ var (
 type EVMKeyKind uint8
 
 const (
-	EVMKeyUnknown EVMKeyKind = iota
-	EVMKeyNonce
-	EVMKeyCodeHash
-	EVMKeyCode
-	EVMKeyStorage
+	EVMKeyEmpty    EVMKeyKind = iota // Returned only for zero-length keys
+	EVMKeyNonce                      // Stripped key: 20-byte address
+	EVMKeyCodeHash                   // Stripped key: 20-byte address
+	EVMKeyCode                       // Stripped key: 20-byte address
+	EVMKeyStorage                    // Stripped key: addr||slot (20+32 bytes)
+	EVMKeyLegacy                     // Full original key preserved (address mappings, codesize, etc.)
 )
 
 // ParseEVMKey parses an EVM key from the x/evm store keyspace.
 //
-// For non-storage keys, keyBytes is the 20-byte address.
-// For storage keys, keyBytes is addr||slot (20+32 bytes).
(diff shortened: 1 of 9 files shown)


sei-db/common/evm/keys.go
```diff
@@ -25,14 +25,16 @@ const (
 	EVMKeyNonce
 	EVMKeyCodeHash
 	EVMKeyCode
+	EVMKeyCodeSize
```
yzang2019: CodeSize should always goes to legacy because we do not want to store it in EVM eventually right?

sei-db/common/evm/keys.go
```diff
@@ -25,14 +25,16 @@ const (
...
 //
-// For non-storage keys, keyBytes is the 20-byte address.
-// For storage keys, keyBytes is addr||slot (20+32 bytes).
-// For unknown or malformed keys, returns (EVMKeyUnknown, nil).
+// For optimized keys (nonce, code, codehash, storage), keyBytes is the stripped key.
+// For legacy keys (all other EVM data), keyBytes is the full original key.
+// Only returns EVMKeyUnknown for empty keys.
 func ParseEVMKey(key []byte) (kind EVMKeyKind, keyBytes []byte) {
 	if len(key) == 0 {
 		return EVMKeyUnknown, nil
```
yzang2019: I think it would be better to return EVMEmptyKey rather than Unknown here?

sei-db/state_db/ss/composite/store.go
```diff
@@ -661,14 +663,15 @@ func extractEVMChangesFromChangesets(changesets []*proto.NamedChangeSet) map[evm
...
-			evmStoreType, strippedKey := commonevm.ParseEVMKey(kvPair.Key)
-			if evmStoreType != evm.StoreUnknown {
-				evmChanges[evmStoreType] = append(evmChanges[evmStoreType], &iavl.KVPair{
-					Key:    strippedKey,
-					Value:  kvPair.Value,
-					Delete: kvPair.Delete,
-				})
+			evmStoreType, keyBytes := commonevm.ParseEVMKey(kvPair.Key)
+			if evmStoreType == evm.StoreUnknown {
+				continue // Empty key
```
yzang2019: Same comment as above, being more explicit about empty key

PR #2793, opened 2026-02-03
feat(sei-db): add flatkv store implementation

blindchaser:
- Multi-DB storage: storageDB, accountDB, codeDB, metadataDB
- LtHash-based state commitment using internal key formats
- Configurable write toggles per DB type
- Iterator support for storage keys


sei-db/config/flatkv_config.go
```diff
@@ -0,0 +1,44 @@
+package config
+
+// FlatKVConfig defines configuration for the FlatKV (EVM) commit store.
+type FlatKVConfig struct {
```
yzang2019: Let's put this under flatkv package?

sei-db/config/flatkv_config.go
```diff
@@ -0,0 +1,44 @@
...
+	// EnableStorageWrites enables writes to storageDB and its LtHash contribution.
+	// When false, storage data is skipped entirely (no DB writes, no LtHash updates).
+	// Default: true
+	EnableStorageWrites bool `mapstructure:"enable-storage-writes"`
```
yzang2019: Not sure why we want these configs, as long as dual write is true, all EVM state should be written to flatkv, and we don't need to control which key to write or not

sei-db/config/flatkv_config.go
```diff
@@ -0,0 +1,44 @@
...
+	// When false, code data is skipped entirely (no DB writes, no LtHash updates).
+	// Default: true
+	EnableCodeWrites bool `mapstructure:"enable-code-writes"`
+
+	// AsyncWrites enables async writes to data DBs for better performance.
+	// When true: data DBs use Sync=false, then Flush() at FlushInterval.
+	// When false (default): all writes use Sync=true for maximum durability.
+	// WAL and metaDB always use sync writes regardless of this setting.
+	// Default: false
+	AsyncWrites bool `mapstructure:"async-writes"`
```
yzang2019: Probably better to split into Fsync: bool and AsyncWriteBuffer: int

sei-db/config/flatkv_config.go
```diff
@@ -0,0 +1,44 @@
...
+
+	// FlushInterval controls how often to flush data DBs and update metaDB.
+	// Only applies when AsyncWrites=true.
+	// - 0 or 1: flush every block (safest, slowest)
+	// - N > 1: flush every N blocks (faster, recovers up to N blocks from WAL)
+	// Default: 100
+	FlushInterval int `mapstructure:"flush-interval"`
```
yzang2019: I think we don't need this config today, since we have to flush every block, we can add in the future if we really need it

sei-db/state_db/sc/composite/store.go
```diff
@@ -36,23 +40,29 @@ type CompositeCommitStore struct {
...
-		cosmosCommitter: cosmosSC,
+		cosmosCommitter: cosmosCommitter,
 		homeDir:         homeDir,
 		config:          cfg,
 	}
 
-	// TODO: initialize FlatKV store for evmSC when cfg.WriteMode != config.CosmosOnlyWrite
+	// Initialize FlatKV store struct if write mode is not cosmos_only
+	// Note: DB is NOT opened here, will be opened in LoadVersion
+	if cfg.WriteMode != config.CosmosOnlyWrite {
```
yzang2019: Better to add a function of shouldWriteToFlatKV() to check config.SplitWrite || config.DualWrite in case we add some new enums in the future

yzang2019: Would recommend checking DualWrite or SplitWrite to determine whether we should initialize EMVCommitter

sei-db/state_db/sc/composite/store.go
```diff
@@ -81,14 +91,33 @@ func (cs *CompositeCommitStore) LoadVersion(targetVersion int64, readOnly bool)
...
 		logger:          cs.logger,
 		cosmosCommitter: cosmosCommitter,
 		homeDir:         cs.homeDir,
 		config:          cs.config,
-		// TODO: Also load evmCommitter for readOnly if enabled
-	}, nil
+	}
+
+	// Also load evmCommitter if enabled
+	if cs.config.WriteMode != config.CosmosOnlyWrite && cs.evmCommitter != nil {
```
yzang2019: We can remove this check `cs.config.WriteMode != config.CosmosOnlyWrite` since evmCommitter will be nil if write is not enabled

sei-db/state_db/sc/composite/store.go
```diff
@@ -81,14 +91,33 @@ func (cs *CompositeCommitStore) LoadVersion(targetVersion int64, readOnly bool)
...
+		// Use LoadVersion on existing evmCommitter (matches memiavl pattern)
+		// This properly handles readOnly flag and avoids resource leaks
+		evmStore, err := cs.evmCommitter.LoadVersion(targetVersion, readOnly)
```
yzang2019, line 106: Not sure if we wanna call LoadVersion here, CMS will call LoadVersion as well, we need to make sure it's only called in one place

sei-db/state_db/sc/flatkv/store.go
```diff
@@ -0,0 +1,681 @@
...
+	"github.com/sei-protocol/sei-chain/sei-db/config"
+	db_engine "github.com/sei-protocol/sei-chain/sei-db/db_engine"
+	"github.com/sei-protocol/sei-chain/sei-db/db_engine/pebbledb"
+	"github.com/sei-protocol/sei-chain/sei-db/proto"
+	"github.com/sei-protocol/sei-chain/sei-db/state_db/sc/flatkv/lthash"
+	"github.com/sei-protocol/sei-chain/sei-db/wal"
+)
+
+const (
+	// DB subdirectories
```
yzang2019, line 18: We need a legacyDB as well

sei-db/state_db/sc/flatkv/store.go
```diff
@@ -0,0 +1,681 @@
...
+	accountDBDir = "account"
+	codeDBDir    = "code"
+	storageDBDir = "storage"
+	metadataDir  = "metadata"
+
+	// Metadata DB keys
+	MetaGlobalVersion = "v" // Global committed version watermark (8 bytes)
```
yzang2019: suggest metadata key stored with _meta/version, _meta/hash

sei-db/state_db/sc/flatkv/store_read.go
```diff
@@ -0,0 +1,296 @@
...
+	"fmt"
+
+	"github.com/sei-protocol/sei-chain/sei-db/common/evm"
+	db_engine "github.com/sei-protocol/sei-chain/sei-db/db_engine"
+	evmtypes "github.com/sei-protocol/sei-chain/x/evm/types"
+)
+
+// Get returns the value for the given memiavl key.
+// Returns (value, true) if found, (nil, false) if not found.
+func (s *CommitStore) Get(key []byte) ([]byte, bool) {
```
yzang2019, line 14: Why do we have store_read.go but not store_write.go? If we want to split, I was expecting to see read and write. Otherwise maybe we should just combine

sei-db/config/flatkv_config.go
blindchaser: need to decouple common/evm from x/evm/types (inlined the 5 key-prefix constants) to break the transitive import cycle.

moved to flatkv.Config / flatkv.DefaultConfig(), same pattern as memiavl.Config now.

sei-db/config/flatkv_config.go
blindchaser: Removed EnableStorageWrites (and the other per-DB toggle configs). All EVM state is written when dual-write is on, no per-key-type toggles.

sei-db/config/flatkv_config.go
blindchaser: done, replaced AsyncWrites with Fsync bool (default true) and AsyncWriteBuffer int (default 0 = synchronous).

sei-db/config/flatkv_config.go
blindchaser: Removed

sei-db/state_db/sc/composite/store.go
blindchaser: changed to explicit `DualWrite || SplitWrite` check.

sei-db/state_db/sc/composite/store.go
blindchaser: removed the redundant WriteMode check. All call sites now just guard on `cs.evmCommitter != nil`

sei-db/state_db/sc/composite/store.go
blindchaser: evmCommitter.LoadVersion is only called inside `CompositeCommitStore.LoadVersion()`. added a comment clarifying this is the single entry point

sei-db/state_db/sc/flatkv/store.go
blindchaser: added legacyDBDir = "legacy" for the legacy DB.

sei-db/state_db/sc/flatkv/store.go
blindchaser: addressed

sei-db/state_db/sc/flatkv/store_read.go
blindchaser: good point, both store_read.go and store_write.go now exist

sei-db/state_db/sc/composite/store.go
yzang2019: I see, that make sense

sei-db/config/toml.go
```diff
@@ -41,6 +41,9 @@ sc-snapshot-interval = {{ .StateCommit.MemIAVLConfig.SnapshotInterval }}
 # to allow more frequent snapshots during normal operation.
 sc-snapshot-min-time-interval = {{ .StateCommit.MemIAVLConfig.SnapshotMinTimeInterval }}
 
+# SnapshotWriterLimit defines the max concurrency for taking commit store snapshot
+sc-snapshot-writer-limit = {{ .StateCommit.MemIAVLConfig.SnapshotWriterLimit }}
```
yzang2019: Is this intentional?

sei-db/state_db/sc/flatkv/api.go
```diff
@@ -90,7 +104,13 @@ type Iterator interface {
 	Next() bool
 	Prev() bool
 
-	// Key returns the current key (valid until next move).
+	// Kind returns the type of the current key (Storage, Nonce, Code, CodeHash).
+	Kind() evm.EVMKeyKind
```
yzang2019: This seems a little bit weird to me that existing in the interface of iterator, do we have to expose it at the interface level or is there other way to get around?

sei-db/state_db/sc/flatkv/store.go
```diff
@@ -0,0 +1,289 @@
...
+	accountDB  db_engine.DB // addr(20) → AccountValue (40 or 72 bytes)
+	codeDB     db_engine.DB // addr(20) → bytecode
+	storageDB  db_engine.DB // addr(20)||slot(32) → value(32)
+	legacyDB   db_engine.DB // Legacy data for backward compatibility
+
+	// Per-DB local metadata (stored inside each DB at 0x00)
+	// Tracks committed version for recovery and consistency checks
+	storageLocalMeta *LocalMeta
+	accountLocalMeta *LocalMeta
+	codeLocalMeta    *LocalMeta
```
yzang2019: Would it be easier to track them via a map[string][DBMeta] ?

yzang2019: In the future if we have sharding, there will be even more DBs right?

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -0,0 +1,378 @@
...
+			// Parse memiavl key to determine type
+			kind, keyBytes := evm.ParseEVMKey(pair.Key)
+			if kind == evm.EVMKeyUnknown {
+				// Skip non-EVM keys silently
+				continue
+			}
+
+			// Route to appropriate DB based on key type
+			switch kind {
+			case evm.EVMKeyStorage:
```
yzang2019, line 46: Can we use a map to avoid switch case boilerplate here?

yzang2019 (approved):

sei-db/config/toml.go
blindchaser: removed. rebase wrong code here

sei-db/state_db/sc/flatkv/store.go
blindchaser: yeah good point. addressed

sei-db/state_db/sc/flatkv/store_write.go
blindchaser: the handlers have different signatures/dependencies (storage/code are direct KV writes, nonce/codehash merge into AccountValue). a map dispatch would need closures capturing different locals, making it less readable for 4 cases. happy to refactor if we add more key types (e.g. sharding), but for now the switch reads more clearly.

sei-db/state_db/sc/flatkv/api.go
blindchaser: agreed it's a bit awkward. moving move Kind() off the interface onto the concrete type

sei-db/state_db/sc/flatkv/iterator.go
```diff
@@ -0,0 +1,247 @@
...
+		internalStart = metaKeyLowerBound()
+	}
+
+	iter, err := db.NewIter(&db_engine.IterOptions{
+		LowerBound: internalStart,
+		UpperBound: internalEnd,
+	})
+	if err != nil {
+		return &emptyIterator{}
+	}
```
cody-littley, line 65: Nit, would it be worth logging something here to help debug iterator creation failures in pebble? Or are these types of errors common and/or expected?

sei-db/state_db/sc/flatkv/iterator.go
```diff
@@ -0,0 +1,247 @@
...
+		internalPrefix = metaKeyLowerBound()
+	}
+
+	iter, err := db.NewIter(&db_engine.IterOptions{
+		LowerBound: internalPrefix,
+		UpperBound: internalEnd,
+	})
+	if err != nil {
+		return &emptyIterator{}
+	}
```
cody-littley, line 90: Also something potentially worth logging (depending on what you decide to do for my similar comment above).

sei-db/state_db/sc/flatkv/iterator.go
```diff
@@ -0,0 +1,247 @@
...
+	return it.iter.Valid()
+}
+
+func (it *dbIterator) Error() error {
+	if it.err != nil {
+		return it.err
+	}
+	return it.iter.Error()
+}
```
cody-littley, line 118: A potential alternative to logging might be to make it so that any error returned during iterator construction can be surfaced here (e.g. by adding an err field to `emptyIterator`).

sei-db/state_db/sc/flatkv/store.go
```diff
@@ -0,0 +1,278 @@
...
+		codeDBDir:    codeDB,
+		storageDBDir: storageDB,
+	}
+	for name, db := range dataDBs {
+		meta, err := loadLocalMeta(db)
+		if err != nil {
+			return fmt.Errorf("failed to load %s local meta: %w", name, err)
+		}
+		s.localMeta[name] = meta
+	}
```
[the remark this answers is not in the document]
cody-littley: I don't think determinism is important here.

sei-db/state_db/sc/flatkv/store_meta.go
```diff
@@ -0,0 +1,81 @@
...
+	db_engine "github.com/sei-protocol/sei-chain/sei-db/db_engine"
+	"github.com/sei-protocol/sei-chain/sei-db/state_db/sc/flatkv/lthash"
+)
+
+// loadLocalMeta loads the local metadata from a DB, or returns default if not present.
+func loadLocalMeta(db db_engine.DB) (*LocalMeta, error) {
+	val, err := db.Get(DBLocalMetaKey)
+	// Check for real errors first to avoid masking I/O issues
+	if err != nil && !db_engine.IsNotFound(err) {
+		return nil, err
```
cody-littley: Consider wrapping this error (feel free to ignore if wrapping errors isn't a pattern the team currently follows)
```suggestion
if err != nil && !db_engine.IsNotFound(err) {
		return nil, fmt.Errorf("could not get DBLocalMetaKey: %v", err)
```

sei-db/state_db/sc/flatkv/store_read.go
```diff
@@ -0,0 +1,271 @@
...
+	"encoding/binary"
+	"fmt"
+
+	"github.com/sei-protocol/sei-chain/sei-db/common/evm"
+	db_engine "github.com/sei-protocol/sei-chain/sei-db/db_engine"
+)
+
+// Get returns the value for the given memiavl key.
+// Returns (value, true) if found, (nil, false) if not found.
+func (s *CommitStore) Get(key []byte) ([]byte, bool) {
```
cody-littley, line 14: Nit, it might be nice to split out the logic in the switch statement case blocks into small helper functions.

sei-db/state_db/sc/flatkv/store_read.go
```diff
@@ -0,0 +1,271 @@
...
+		// Read from codeDB
+		value, err := s.codeDB.Get(keyBytes)
+		if err != nil {
+			return nil, false
+		}
+		return value, true
+
+	default:
+		return nil, false
+	}
```
cody-littley, line 105: Do we expect to hit the default case during normal operation? If not, would it make sense to return an error in case we accidentally end up in that scenario?
```suggestion
	default:
		return nil, fmt.Errorf("unhandled kind: %d", kind)
	}
```

sei-db/state_db/sc/flatkv/store_write.go
cody-littley, line 46: Also potentially a good candidate to split switch case blocks into helper functions.

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -0,0 +1,378 @@
...
+			continue
+		}
+
+		for _, pair := range namedCS.Changeset.Pairs {
+			// Parse memiavl key to determine type
+			kind, keyBytes := evm.ParseEVMKey(pair.Key)
+			if kind == evm.EVMKeyUnknown {
+				// Skip non-EVM keys silently
+				continue
+			}
```
cody-littley, line 42: Do we expect to find non-EVM keys during regular operations? If not, would it make sense to log a warning?

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -0,0 +1,378 @@
...
+		return fmt.Errorf("accountDB flush: %w", err)
+	}
+	if err := s.codeDB.Flush(); err != nil {
+		return fmt.Errorf("codeDB flush: %w", err)
+	}
+	if err := s.storageDB.Flush(); err != nil {
+		return fmt.Errorf("storageDB flush: %w", err)
+	}
+	return nil
+}
```
cody-littley, line 258: Since commit/flush is probably doing nontrivial IO work, I wonder if we could get some meaningful performance gain by committing/flushing each DB on a parallel goroutine.
```suggestion
func (s *CommitStore) flushAllDBs() error {

	errChan := make(chan error, 3)

	go func() {
		if err := s.accountDB.Flush(); err != nil {
			errChan <- fmt.Errorf("accountDB flush: %w", err)
		} else {
			errChan <- nil
		}
	}()

	go func() {
		if err := s.codeDB.Flush(); err != nil {
			errChan <- fmt.Errorf("codeDB flush: %w", err)
		} else {
			errChan <- nil
		}
	}()

	go func() {
		if err := s.storageDB.Flush(); err != nil {
			errChan <- fmt.Errorf("storageDB flush: %w", err)
		} else {
			errChan <- nil
		}
	}()

	for i := 0; i < 3; i++ {
		err := <-errChan
		if err != nil {
			return fmt.Errorf("failed to flush DB: %v", err)
		}
	}

	return nil
}
```

cody-littley (approved):
My comments are mostly nitpicks, overall LGTM. Please take my review with a grain of salt, since I'm still building context.

sei-db/state_db/sc/flatkv/iterator.go
blindchaser: good point, will add a log.Error here

sei-db/state_db/sc/flatkv/store_meta.go
blindchaser: addressed

sei-db/state_db/sc/flatkv/store_read.go
blindchaser: acknowledged, the cases are ~10-15 lines each so I lean towards keeping them inline for now to avoid function-per-case overhead. happy to revisit if the cases grow in complexity.

sei-db/state_db/sc/flatkv/store_read.go
blindchaser: the Get() signature returns ([]byte, bool) per the Store interface matching the memiavl, the default case is expected to fire for some EVMKeyLegacy, but im not 100% sure during operations though. returning (nil, false) is the correct "not found" semantics here.

sei-db/state_db/sc/flatkv/store_write.go
blindchaser: will add a log.Warn as good defensive measure.

sei-db/state_db/sc/flatkv/store_write.go
blindchaser: good thought! i think memtable → L0 is probably fast enough, If profiling shows flush is a bottleneck we can revisit. will keep this in mind

sei-db/state_db/sc/flatkv/store_write.go
blindchaser: agreed it would improve readability. I'll keep this as-is for now since the cases are tightly coupled with the local variables (modifiedAccounts, storagePairs, codePairs, s.storageWrites, etc.), happy to revisit if the function grows further.

sei-db/state_db/sc/flatkv/iterator.go
blindchaser: done

app/params/config.go
```diff
@@ -107,8 +107,6 @@ func setValidatorTypeTendermintConfig(config *tmcfg.Config) {
 // setFullnodeTypeTendermintConfig sets common Tendermint config for fullnode-like nodes
 func setFullnodeTypeTendermintConfig(config *tmcfg.Config) {
 	config.TxIndex.Indexer = []string{"kv"} // Full nodes need tx indexing for queries
-	config.RPC.ListenAddress = "tcp://0.0.0.0:26657"
```
mojtaba-esk, line 110: Hey @blindchaser 
These lines were added by #2903 Can you please explain why they are removed? or was it an unresolved conflict issue?

mojtaba-esk: This PR is a large one and such accident could happen. A new PR is opened to put them back #2913 would you please have a look and see if it is alright? thanks

PR #2794, opened 2026-02-03
go bench read + write receipts/logs for parquet vs pebble

jewei1997:
## Summary

  - Add benchmarks comparing pebble vs parquet receipt store read/write performance across realistic workloads (100 blocks with 10-100 receipts per block)
 - using 3000 receipts/block and 5k blocks for writes and 2k blocks for reads (block range is ~100-500 blocks) though. 

Writes
 - both pebble and parquet do sync writes to WAL + async writes to underlying store
 - receipt cache is not used in benchmark for more accurate comparison
 - command `go test -bench=BenchmarkReceiptWriteAsync -benchtime=1x -run=^$ ./sei-db/ledger_db/receipt/`

Reads
  - pebble iterates over each receipt (still uses bloom filters) while parquet issues a SQL-like filter query
  - when hitting ledger cache both are fast as expected but outside of ledger cache, pebble super slow due to need to iterate over each receipt
  - tested over varying concurrency, block range, diff filter criteria (topics, address)
  - command `go test -run='^$' -bench 'BenchmarkReceiptReadStore100' -benchtime=1x -count=1 -timeout=30m ./sei-db/ledger_db/receipt/ -v`

## Benchmark Results

Parquet faster for writes and reads 👍 

Writes: 
On detached pacific-1 RPC node (EC2 instance with EBS), getting 347,112.00 receipts/s writes using parquet, but only "only" 143,000 receipts/s using pebble

Reads:
Not even comparable. Parquet way faster. 20ms per log for pebble vs 0.05918 ms per log for parquet


blindchaser (approved):

yzang2019 (approved):

evmrpc/ethbloom/bloom.go
```diff
@@ -0,0 +1,126 @@
+package ethbloom
```
jewei1997, line 1: just a refactor, not new code added

PR #2799, opened 2026-02-04
made the peer dialing less aggressive

pompon0:
* DialRate is now exposed in config.toml. It defaults to 10s (instead of previous 1.5s).
* Reduced MaxAcceptRate from 0.5s to 1s
* Fixed NewConnections metric (outgoing dials were reported twice). Added success/failure label.

### DIFF
--- a/sei-cosmos/storev2/rootmulti/store_test.go
+++ b/sei-cosmos/storev2/rootmulti/store_test.go
@@ -118,7 +118,7 @@ func TestCacheMultiStoreWithVersion_OnlyUsesSSStores(t *testing.T) {
 			ssCfg.Enable = tc.ssEnabled
 			ssCfg.AsyncWriteBuffer = 0
 
-			store := NewStore(home, log.NewNopLogger(), scCfg, ssCfg, false, []string{})
+			store := NewStore(home, log.NewNopLogger(), scCfg, ssCfg, false)
 			defer func() { _ = store.Close() }()
 
 			iavlKey1 := types.NewKVStoreKey("iavl_store1")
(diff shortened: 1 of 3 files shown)


sei-tendermint/config/config.go
```diff
@@ -679,6 +679,9 @@ type P2PConfig struct { //nolint: maligned
 	// Peer connection configuration.
 	HandshakeTimeout time.Duration `mapstructure:"handshake-timeout"`
 	DialTimeout      time.Duration `mapstructure:"dial-timeout"`
+	
+	// How often node should dial a new peer.
+	DialRate         time.Duration `mapstructure:"dial-rate"`
```
yzang2019: Shall we also add to config.toml?

pompon0: you mean the template?

philipsu522 (approved):

sei-tendermint/config/config.go
yzang2019: yeah

masih: * Do we have a metric for for dials? a simple counter would do. If not worth adding I think

* "Rate" suggest a ratio; x over time. But the unit is time.Duration, which implies this is dial "interval"? If so, perhaps `DialInterval` is a better name.

masih (approved):

sei-tendermint/config/config.go
pompon0: we do, it is called NewConnections

pompon0: will add it in another pr

PR #2801, opened 2026-02-04
fix: prevent nil pointer panic during pruning on shutdown

blindchaser:

- Remove redundant stateStore.Close() from HandleClose() since cms.Close() already handles it via rootmulti.Store.Close() -> ssStore.Close()
- Add defensive nil checks in pebbledb/rocksdb Prune() methods to gracefully return error instead of panic if db is already closed


Kbhat1 (approved):

sei-db/ss/pebbledb/db.go
```diff
@@ -187,6 +188,9 @@ func New(dataDir string, config config.StateStoreConfig) (*Database, error) {
 }
 
 func (db *Database) Close() error {
+	// Mark as closed first to signal pruning to skip/return early
+	db.closed.Store(true)
```
masih: This should do a compare and swap and proceed if not closed already

blindchaser: good point

PR #2807, opened 2026-02-05
fix(test): force GC behaviour for determinism

arajasek:
Cursor-supplied change, aims to prevent failures like this one https://github.com/sei-protocol/sei-chain/actions/runs/21715802000/job/62631307540?pr=2805

### DIFF
--- a/sei-tendermint/internal/libs/clist/clist_test.go
+++ b/sei-tendermint/internal/libs/clist/clist_test.go
@@ -157,7 +157,9 @@ func TestGCRandom(t *testing.T) {
 		el := els[i]
 		l.Remove(el)
 		_ = el.Next()
+		els[i] = nil // Clear reference to allow GC
 	}
+	els = nil // Clear the slice to allow GC of all elements
 
 	tickerQuitCh := make(chan struct{})
 	tickerDoneCh := make(chan struct{})


pompon0 (approved):

sei-tendermint/internal/libs/clist/clist_test.go
```diff
@@ -157,7 +157,9 @@ func TestGCRandom(t *testing.T) {
 		el := els[i]
 		l.Remove(el)
 		_ = el.Next()
+		els[i] = nil // Clear reference to allow GC
 	}
+	els = nil // Clear the slice to allow GC of all elements
```
masih, line 162: I would probably rewrite this entire test to use `runtime.ReadMemStats`. That should be more reliable and inherently less flaky. Because, it won't depend on the finalizer timing.

Something like this:
```go
func TestGCRandom(t *testing.T) {
	const numElements = 10000

	// Force GC and get baseline
	runtime.GC()
	runtime.GC() // twice to clean up finalizers from previous tests
	var before runtime.MemStats
	runtime.ReadMemStats(&before)

	l := New()
	type value struct {
		Data [128]byte // Make it large enough to measure
	}

	for i := 0; i < numElements; i++ {
		v := new(value)
		l.PushBack(v)
	}

	els := make([]*CElement, 0, numElements)
	for el := l.Front(); el != nil; el = el.Next() {
		els = append(els, el)
	}

	// Remove in random order
	for _, i := range mrand.Perm(numElements) {
		l.Remove(els[i])
		els[i] = nil
	}
	els = nil

	// Force GC and measure
	runtime.GC()
	runtime.GC()
	var after runtime.MemStats
	runtime.ReadMemStats(&after)

	// Check that memory was reclaimed (allow some slack for runtime overhead)
	leaked := int64(after.HeapAlloc) - int64(before.HeapAlloc)
	maxAcceptableLeak := int64(numElements * 64) // Much less than numElements * sizeof(value)

	if leaked > maxAcceptableLeak {
		t.Errorf("potential memory leak: %d bytes not reclaimed", leaked)
	}
}
```

PR #2809, opened 2026-02-05
feat: turn snapshot prune async

blindchaser:

### DIFF
--- a/sei-db/sc/memiavl/db_test.go
+++ b/sei-db/sc/memiavl/db_test.go
@@ -153,8 +153,10 @@ func TestRewriteSnapshotBackground(t *testing.T) {
 	close(stopCh)
 	wg.Wait()
 
-	db.pruneSnapshotLock.Lock()
-	defer db.pruneSnapshotLock.Unlock()
+	// Wait for any ongoing prune to finish
+	require.Eventually(t, func() bool {
+		return !db.pruningInProgress.Load()
+	}, time.Second, 10*time.Millisecond, "prune should complete")
 
 	entries, err := os.ReadDir(db.dir)
 	require.NoError(t, err)


sei-db/sc/memiavl/db.go
```diff
@@ -88,6 +89,9 @@ type DB struct {
 	mtx sync.Mutex
 	// worker goroutine IdleTimeout = 5s
 	snapshotWriterPool *pond.WorkerPool
+
+	// closed indicates the DB is shutting down
+	closed uint32
```
yzang2019: Can we rename this to pruningInProgress?

sei-db/state_db/sc/flatkv/catchup.go
```diff
@@ -0,0 +1,511 @@
+package flatkv
```
yzang2019: Don't think we need this file

sei-db/sc/memiavl/db.go
```diff
@@ -532,6 +542,9 @@ func (db *DB) pruneSnapshots() {
 		db.logger.Error("failed to find first snapshot", "err", err)
 	}
 
+	if db.streamHandler == nil {
+		return
+	}
 	if err := db.streamHandler.TruncateBefore(utils.VersionToIndex(earliestVersion+1, db.initialVersion.Load())); err != nil {
```
yzang2019, line 547: Can we add a log here for how long the whole pruning process take?

sei-db/sc/memiavl/db.go
```diff
@@ -491,9 +495,15 @@ func (db *DB) checkBackgroundSnapshotRewrite() error {
 
 // pruneSnapshot prune the old snapshots
 func (db *DB) pruneSnapshots() {
+	if atomic.LoadUint32(&db.pruningInProgress) == 1 {
```
yzang2019: Maybe use compareAndSet?

sei-db/sc/memiavl/db.go
```diff
@@ -532,6 +541,9 @@ func (db *DB) pruneSnapshots() {
 		db.logger.Error("failed to find first snapshot", "err", err)
 	}
 
+	if db.streamHandler == nil {
+		return
```
Kbhat1, line 545: should we log something here?

sei-db/sc/memiavl/db.go
```diff
@@ -491,9 +493,16 @@ func (db *DB) checkBackgroundSnapshotRewrite() error {
 
 // pruneSnapshot prune the old snapshots
 func (db *DB) pruneSnapshots() {
-	// wait until last prune finish
-	db.pruneSnapshotLock.Lock()
-	defer db.pruneSnapshotLock.Unlock()
+	// CAS: only one prune can run at a time
+	if !db.pruningInProgress.CompareAndSwap(false, true) {
```
masih, line 497: What if a prune takes just long enough to overlap with the next prune? 
In that case the new prune cycle will be skipped and we end up with 2 cycles of unpruned state in the next prune

masih (commented):
As is this is fine to test the theories discussed in the sync.But for a deeper fix i would like to understand the edge cases this change rises.

sei-db/sc/memiavl/db.go
blindchaser: yup, looking into the code. we will need to optimize some pruning logic and where/when pruning happens. creating a ticket for that.

yzang2019: Each prune should prune all snapshots that need to be pruned, not just delete one snapshot. We also should make the pruning logic to be triggered after each commit, instead of only after snapshot creation

PR #2810, opened 2026-02-05
fix(giga): check whether txs follow Giga ordering

arajasek:

Giga's OCC executor needs transactions to be "all EVM, then all Cosmos". If a set of txs aren't in this order, fall back to V2.

Also avoid unnecessary scheduling if there aren't any txs.

Tests pass.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -749,7 +749,7 @@ func New(
 		if gigaExecutorConfig.OCCEnabled {
 			logger.Info("benchmark: Giga Executor with OCC is ENABLED - using new EVM execution path with parallel execution")
 		} else {
-			logger.Info("benchmark: Giga Executor (evmone-based) is ENABLED - using new EVM execution path (sequential)")
+			logger.Info("benchmark: Giga Executor is ENABLED - using new EVM execution path (sequential)")
 		}
 	} else {
 		logger.Info("benchmark: Giga Executor is DISABLED - using default GETH interpreter")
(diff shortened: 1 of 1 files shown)


app/app.go
```diff
@@ -721,7 +721,7 @@ func New(
 		if gigaExecutorConfig.OCCEnabled {
 			logger.Info("benchmark: Giga Executor with OCC is ENABLED - using new EVM execution path with parallel execution")
 		} else {
-			logger.Info("benchmark: Giga Executor (evmone-based) is ENABLED - using new EVM execution path (sequential)")
+			logger.Info("benchmark: Giga Executor is ENABLED - using new EVM execution path (sequential)")
```
arajasek, line 752: Unrelated fix

app/app.go
```diff
@@ -1537,62 +1537,78 @@ func (app *App) ProcessTXsWithOCCV2(ctx sdk.Context, txs [][]byte, typedTxs []sd
 func (app *App) ProcessTXsWithOCCGiga(ctx sdk.Context, txs [][]byte, typedTxs []sdk.Tx, absoluteTxIndices []int) ([]*abci.ExecTxResult, sdk.Context) {
 	evmEntries := make([]*sdk.DeliverTxEntry, 0, len(txs))
 	v2Entries := make([]*sdk.DeliverTxEntry, 0, len(txs))
+	firstCosmosSeen := false
 	for txIndex, tx := range txs {
 		if app.GetEVMMsg(typedTxs[txIndex]) != nil {
+			if firstCosmosSeen {
+				// Oops! This isn't "all EVM txs, then all Cosmos txs" - we need to fallback to V2.
```
codchen, line 1554: can we add a log here if this isn't expected? (like if the ordering is supposed to be all evm then cosmos)

arajasek: I'm happy to do that, but the ordering isn't _currently_ unexpected, because the mempool ordering change will only go live with 6.4.0. It'll only become unexpected when validators are running 6.4.

arajasek: (Done.)

PR #2812, opened 2026-02-05
Adjust fallback logic

stevenlanders:
Refactor ExecuteTxsConcurrently to centralize giga→v2 fallback. Giga functions now return a bool instead of handling fallback internally.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1359,9 +1359,9 @@ func (app *App) ProcessTxsSynchronousV2(ctx sdk.Context, txs [][]byte, typedTxs 
 }
 
 // ProcessTxsSynchronousGiga executes transactions synchronously using the Giga executor.
-// Returns ErrGigaFallbackToV2 if any transaction requires fallback to v2 execution.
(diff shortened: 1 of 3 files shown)


app/app.go
```diff
@@ -1441,14 +1441,35 @@ func (app *App) PartitionPrioritizedTxs(_ sdk.Context, txs [][]byte, typedTxs []
...
+		if app.GigaOCCEnabled {
+			results, newCtx, err = app.ProcessTXsWithOCCGiga(ctx.WithGiga(true), txs, typedTxs, absoluteTxIndices)
+		} else {
+			results, err = app.ProcessTxsSynchronousGiga(ctx.WithGiga(true), txs, typedTxs, absoluteTxIndices)
+			newCtx = ctx
+		}
+
+		if err == nil {
+			return results, newCtx
+		}
```
arajasek, line 1463: let's check the error and at least scream if the error _isn't_ ErrGigaFallbackToV2

stevenlanders: changed to `bool` (meaning ok)

PR #2814, opened 2026-02-05
feat: make snapshot prune async

blindchaser:
- Run pruneSnapshots() asynchronously (go db.pruneSnapshots()) to avoid blocking Commit
- Use TryLock instead of Lock so overlapping prunes are skipped rather than queued
- Add logging for prune start/skip/completion with duration
- Add idempotent Close() guard to prevent double-close
- Update test to use Eventually on directory state

### DIFF
--- a/sei-db/state_db/sc/memiavl/db.go
+++ b/sei-db/state_db/sc/memiavl/db.go
@@ -11,7 +11,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/alitto/pond"
@@ -76,9 +75,9 @@ type DB struct {
 	// Protected by db.mtx (only accessed in Commit call chain)
 	lastSnapshotTime time.Time
 
-	// pruningInProgress guards concurrent prune operations (CAS-based)
-	pruningInProgress atomic.Bool
-	// closed ensures Close() is idempotent, protected by db.mtx
+	// pruneSnapshotLock guards concurrent prune operations; use TryLock in pruneSnapshots
+	pruneSnapshotLock sync.Mutex
+	// closed guards against double Close(), protected by db.mtx
 	closed bool
 
 	// walIndexDelta is the difference: version - walIndex for any entry.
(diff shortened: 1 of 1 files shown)


sei-db/state_db/sc/memiavl/db.go
```diff
@@ -510,9 +514,18 @@ func (db *DB) checkBackgroundSnapshotRewrite() error {
...
+	if !db.pruningInProgress.CompareAndSwap(false, true) {
+		db.logger.Info("pruneSnapshots skipped, previous prune still in progress")
+		return
+	}
+	defer db.pruningInProgress.Store(false)
+
+	db.logger.Info("pruneSnapshots started")
+	startTime := time.Now()
+	defer func() {
+		db.logger.Info("pruneSnapshots completed", "duration", fmt.Sprintf("%.2fs", time.Since(startTime).Seconds()))
```
yzang2019: duration_sec

sei-db/state_db/sc/memiavl/db.go
```diff
@@ -510,9 +514,18 @@ func (db *DB) checkBackgroundSnapshotRewrite() error {
 // pruneSnapshots prunes old snapshots, keeping only snapshotKeepRecent recent ones.
 // Note: WAL truncation is now handled by CommitStore after each commit.
 func (db *DB) pruneSnapshots() {
-	// wait until last prune finish
-	db.pruneSnapshotLock.Lock()
-	defer db.pruneSnapshotLock.Unlock()
+	// CAS: only one prune can run at a time
+	if !db.pruningInProgress.CompareAndSwap(false, true) {
```
stevenlanders: this works, but fwiw, there's also a TryLock option (fyi):
```
    if !db.pruningMu.TryLock() {
        db.logger.Info("pruneSnapshots skipped, previous prune still in progress")
        return
    }
    defer db.pruningMu.Unlock()
```

sei-db/state_db/sc/memiavl/db.go
```diff
@@ -929,9 +942,16 @@ func (db *DB) Close() error {
 	db.logger.Info("Closing memiavl db...")
 	db.mtx.Lock()
 	defer db.mtx.Unlock()
+	// Idempotent: prevent double-close spinning on pruningInProgress CAS
+	if db.closed {
+		return nil
+	}
+	db.closed = true
+	// Wait for any ongoing prune to finish, then block new prunes
+	for !db.pruningInProgress.CompareAndSwap(false, true) {
```
stevenlanders: and if we used a mutex, this could be
```
   db.pruningMu.Lock() // blocks until pruning finishes, no spinning needed
   db.pruningMu.Unlock()
```

sei-db/state_db/sc/memiavl/db.go
blindchaser: this seems even simpler than cas way

sei-db/state_db/sc/memiavl/db.go
blindchaser: updated

sei-db/state_db/sc/memiavl/db.go
blindchaser: will take this approach

PR #2817, opened 2026-02-06
Flatten `sei-tendermint` go mod into `sei-chain`

masih:
Flatten the sei-tendermint go module into sei-chain tin reduce dependency management complexity and error-prone multi layer replace directives.

Part of PLT-6

### DIFF
--- a/sei-cosmos/store/multiversion/mvkv.go
+++ b/sei-cosmos/store/multiversion/mvkv.go
@@ -82,9 +82,6 @@ type VersionIndexedStore struct {
 	iterateset Iterateset
 	// TODO: need to add iterateset here as well
 
-	// used for iterators - populated at the time of iterator instantiation
-	// TODO: when we want to perform iteration, we need to move all the dirty keys (writeset and readset) into the sortedTree and then combine with the iterators for the underlying stores
-	sortedStore *dbm.MemDB // always ascending sorted
 	// parent stores (both multiversion and underlying parent store)
 	multiVersionStore MultiVersionStore
 	parent            types.KVStore
@@ -104,7 +101,6 @@ func NewVersionIndexedStore(parent types.KVStore, multiVersionStore MultiVersion
 		readset:           make(map[string][][]byte),
 		writeset:          make(map[string][]byte),
 		iterateset:        []*iterationTracker{},
-		sortedStore:       dbm.NewMemDB(),
 		parent:            parent,
 		multiVersionStore: multiVersionStore,
 		transactionIndex:  transactionIndex,


sei-tendermint/cmd/tendermint/commands/debug/dump.go
```diff
@@ -75,12 +75,18 @@ if enabled.`,
 			}
 			dumpDebugData(ctx, logger, rpc, dumpArgs)
 
-			ticker := time.NewTicker(time.Duration(frequency) * time.Second)
-			for range ticker.C {
-				dumpDebugData(ctx, logger, rpc, dumpArgs)
-			}
+			ticker := time.NewTicker(time.Duration(frequency) * time.Second) //nolint: gosec
+			defer ticker.Stop()
```
pompon0, line 79: nit: not needed since 1.23

sei-tendermint/crypto/merkle/proof.go
```diff
@@ -153,19 +153,19 @@ func ProofFromProto(pb *tmcrypto.Proof) (*Proof, error) {
 // Recursive impl.
 func computeHashFromAunts(index, total int64, leafHash []byte, innerHashes [][]byte) ([]byte, error) {
 	if index >= total || index < 0 || total <= 0 {
-		return nil, fmt.Errorf("Calling computeHashFromAunts() with invalid index (%d) and total (%d)", index, total)
+		return nil, fmt.Errorf("calling computeHashFromAunts() with invalid index (%d) and total (%d)", index, total)
```
pompon0, line 156: fyi, the lint which checks that the first letter of the error is lowercase is imprecise and super annoying - sometimes the first letter simply has to be uppercase (like when the first word is a name of a public type, or sth). I'd recommend to disable it.

sei-tendermint/internal/consensus/msgs.go
```diff
@@ -336,7 +336,7 @@ func newValidBlockMessageFromProto(pb *tmcons.NewValidBlock) (*NewValidBlockMess
 func proposalMessageFromProto(pb *tmcons.Proposal) (*ProposalMessage, error) {
 	proposal, err := types.ProposalFromProto(&pb.Proposal)
 	if err != nil {
-		return nil, fmt.Errorf("Proposal: %w", err)
```
pompon0, line 339: like here: the field is "Proposal", not "proposal". Lowercasing is not helpful.

sei-tendermint/internal/consensus/peer_state.go
```diff
@@ -394,7 +395,7 @@ func (ps *PeerState) ApplyNewRoundStepMessage(msg *NewRoundStepMessage) {
 	defer ps.mtx.Unlock()
 
 	// ignore duplicates or decreases
-	if msg.HRS.Cmp(ps.PRS.HRS) <= 0 {
+	if msg.Cmp(ps.PRS.HRS) <= 0 {
```
pompon0, line 398: this lint does something weird - we compare HRS embedded fields here. Hiding that fact harms readability.

sei-tendermint/internal/consensus/peer_state.go
```diff
@@ -298,7 +298,8 @@ func (ps *PeerState) EnsureVoteBitArrays(height int64, numValidators int) {
 }
 
 func (ps *PeerState) ensureVoteBitArrays(height int64, numValidators int) {
-	if ps.PRS.Height == height {
+	switch ps.PRS.Height {
```
pompon0, line 301: @masih wdyt about this if -> switch lint? Do you find it more readable?

sei-tendermint/internal/mempool/tx.go
```diff
@@ -352,7 +352,7 @@ func (p *PendingTxs) popTxsAtIndices(indices []int) {
 	if len(indices) == 0 {
 		return
 	}
-	newTxs := []TxWithResponse{}
+	newTxs := make([]TxWithResponse, 0, len(p.txs)-len(indices))
```
pompon0: is this safe? what if indices is larger than txs?

sei-tendermint/internal/p2p/mux/stream_state.go
```diff
@@ -119,7 +120,7 @@ func (s *streamState) RemoteMsgEnd() error {
 		if inner.recv.used == inner.recv.end {
 			return fmt.Errorf("buffer full")
 		}
-		inner.recv.used += 1
+		inner.recv.used++
```
pompon0: do we want this lint? I find it rather annoying. It is not like "++" is more readable and it compiles to the same thing.

sei-tendermint/light/rpc/client.go
```diff
@@ -54,8 +54,6 @@ type Client struct {
 	closers []func()
 }
 
-var _ rpcclient.Client = (*Client)(nil)
```
pompon0, line 57: why is this assertion removed?

sei-tendermint/test/e2e/app/app.go
```diff
@@ -215,7 +215,7 @@ func (app *Application) FinalizeBlock(_ context.Context, req *abci.RequestFinali
 					},
 					{
 						Key:   []byte("height"),
-						Value: []byte(strconv.Itoa(int(req.Height))),
+						Value: []byte(strconv.FormatInt(req.Height, 10)),
```
pompon0, line 218: huh, how is this better?

sei-tendermint/cmd/tendermint/commands/debug/dump.go
masih: Although it is not strictly needed after 1.23, there are a few reasons to still call stop: communication of intent, defensive programming in case of refactoring that may move the ticker and so on. 
But the main functional reason to call it is that it would stop the ticker immediately, instead of leaving it to GC. It makes up a more predictable code behaviour.

sei-tendermint/crypto/merkle/proof.go
masih: Agreed that ST1005 can be annoying, but I'd push back on disabling it entirely; the consistency alone is worth it. The general practice is to instead capitalise the first letter in logs, because they typically are a list of sentences. See:
* https://go.dev/wiki/CodeReviewComments#error-strings

The "(unless beginning with proper nouns or acronyms)" is loose. As a reference, I very rarely come across Go SDK implementation that falls in that exceptional category. 

In this specific case, the error should indeed start with lower case.

sei-tendermint/internal/consensus/peer_state.go
masih: This is from QF1008 staticcheck linter. Its rationale is rooted in [Effective Go](https://go.dev/doc/effective_go#embedding), where the entire point of embedding is promotion; to avoid bookkeeping and decouple from internals.

I'm against always using the explicit form, because that defeats the purpose of the language feature. In this case, HRS is a core type with no clashing receiver functions on both embedder and embedded types in this repo.
If HRS is more than an implementation detail, then my recommendation is to not embed it; use a named field instead. That forces callers to always spell out "I want to compare X and Y by their progress in terms of height, round, step". What harms readability most is allowing two ways of doing the same thing when the intent is to always use one. Keeping QF1008 enabled helps enforce that consistency.

Happy to revert this change in any case.

sei-tendermint/internal/consensus/peer_state.go
masih: This wasn't a linter suggestion; it was a style choice on my part. The switch makes the two height cases immediately visible as the top-level branching structure, with the nil checks as secondary logic within each. With nested if/else, the height comparison and nil checks sit at the same visual level.

Happy to revert if you prefer the original. You'll be staring at this code more than I will.

sei-tendermint/internal/mempool/tx.go
masih: It is safe because [popped indices are never more than transactions](https://github.com/sei-protocol/sei-chain/blob/f3984c44e651fe6ba998d4c09373043f8b9255be/sei-tendermint/internal/mempool/tx.go#L336). But certainly worth writing it such that nothing can possibly go wrong 👍

sei-tendermint/internal/p2p/mux/stream_state.go
masih: This wasn't the linter, this was me. Happy to revert if you find it annoying.

FWIW `++` is idiomatic Go; the language deliberately only has post-increment as a statement, not an expression, so it's unambiguous.

sei-tendermint/light/rpc/client.go
masih: By accident; apologies.

sei-tendermint/test/e2e/app/app.go
masih: It is better because, `Itoa` literally calls `FormatInt` under the hood. The issue is correctness:

`req.Height` is `int64`. The original code does `strconv.Itoa(int(req.Height))`, which casts `int64` → `int`. On 64-bit platforms `int` is 64 bits so this is fine, but `int` is only guaranteed to be platform-sized; on a 32-bit platform it's 32 bits, and that cast silently truncates any height above ~2.1 billion. `FormatInt` takes `int64` directly, so there's no cast and no truncation risk.

Even if this will never run on 32-bit in practice, the cast is doing nothing useful; it's an extra conversion that only exists because `Itoa` takes `int` instead of `int64`. 

Simply put, `FormatInt` is the right function when you already have an `int64`.

sei-tendermint/internal/consensus/msgs.go
masih: Sure, we can keep it uppercase and add a lint ignore. But looking at the repo, the error strings are already inconsistent. Rather than adding lint ignores case by case, I'd suggest adopting a consistent `"verb noun/Noun: %w"` pattern; example

```go
fmt.Errorf("converting Proposal: %w", err)
```

This reads well when errors chain and eventually logged; e.g. :
```
Failed to process peer message: converting Proposal: invalid signature length
```
vs.
```
Failed to process peer message: Proposal: invalid signature length
```
...and naturally avoids the capitalization lint issue entirely. It also matches the Go stdlib pattern (e.g. `"tls: failed to find certificate..."`) where error strings compose mid-sentence as described in the [Go Code Review Comments](https://go.dev/wiki/CodeReviewComments#error-strings).

This repo doesn't have a logging convention yet. Adopting `"verb noun/Noun/WhatEver: %w"` now means fewer lint ignores going forward and clearer log lines for debugging.

sei-tendermint/light/rpc/client.go
masih: _Actually_, it was removed because it's asserted twice. The duplicate assertion is a few lines above this one. Removed it again.

masih:
Landing this to avid further merge conflicts with `main`. Happy to discuss any changes retroactively, specifically [this](https://github.com/sei-protocol/sei-chain/pull/2817#discussion_r2787318314) one considering logging revamp is on the horizon.

sei-tendermint/internal/consensus/msgs.go
pompon0: as far as I'm concerned verbs are almost always meaningless - all we care in practice is a stack trace, and mangling it into a human readable message is just another layer of obfuscation that we have to get through to actually locate where the error comes from.

sei-tendermint/internal/consensus/peer_state.go
pompon0: here embedding is used to group a bunch of fields, especially so that you can grab them and compare them against a group of fields of another struct. The point is that this call is asymmetric now - we promote the receiver but not the argument. And it will stop working if we define another Cmp on the msg itself.

sei-tendermint/internal/p2p/mux/stream_state.go
pompon0: ack, as long as it is not linter, its fine by me.

sei-tendermint/crypto/merkle/proof.go
pompon0: I disagree that it is worth it, but I can work with that - in worst case I can prefix every message with "q".

sei-tendermint/internal/consensus/msgs.go
masih: > as far as I'm concerned verbs are almost always meaningless - 

A tad over generalised 😅 

> actually locate where the error comes from.

Yep; logging is a mess all over the repo and this is a great test to make sure logging is useful.

sei-tendermint/crypto/merkle/proof.go
masih: Glad to hear that. Also Go community is pretty active, maybe you could submit a proposal and hear what the community has to say on this.

sei-tendermint/internal/consensus/peer_state.go
masih: > And it will stop working if we define another Cmp on the msg itself.

Correct. emphasis on _if_.

as I pointed out, defining HRS as a named struct field is the concrete way to make sure that can't happen in that the compiler will error out. Having it embedded, and worrying about overrides seems like a racy way to assert intent. I would encourage thinking deeper about the intent of embedding.

PR #2823, opened 2026-02-06
fix to halt due to reconstructing block from bad proposal

pompon0:

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"math"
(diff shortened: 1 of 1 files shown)


sei-tendermint/internal/consensus/state.go
```diff
@@ -1522,43 +1521,19 @@ func (cs *State) defaultDoPrevote(ctx context.Context, height int64, round int32
...
-				cs.signAddVote(ctx, tmproto.PrevoteType, nil, types.PartSetHeader{})
-				return
-			}
-			if block == nil {
-				logger.Error("prevote step: ProposalBlock is nil")
-				cs.signAddVote(ctx, tmproto.PrevoteType, nil, types.PartSetHeader{})
-				return
-			}
-			cs.roundState.SetProposalBlock(block)
+		if block == nil {
```
masih: Under what condition `block` and `err` could both be `nil`? I think this path in the current implementation is unreachable?

Makes more sense to panic/scream loudly here if block ends up being `nil` after `err` check.

pompon0: true, fixed.

PR #2835, opened 2026-02-11
feat: embed genesis for well-known chains

mojtaba-esk:

This PR proposes embedding genesis file for the well-known networks (pacific-1, arctic-1, atlantic-2) so users only need the seid binary. 
Running seid init <moniker> --chain-id pacific-1 (or arctic-1 / atlantic-2) writes the correct genesis from the binary; no separate genesis download. Adds an app/genesis package with go:embed and tests that assert SHA256 digests of the embedded files to avoid accidental changes.

Fixes PLT-63

### DIFF
--- a/evmrpc/filter.go
+++ b/evmrpc/filter.go
@@ -18,6 +18,7 @@ import (
 	ethrpc "github.com/ethereum/go-ethereum/rpc"
 	"github.com/hashicorp/golang-lru/v2/expirable"
 	evmrpcconfig "github.com/sei-protocol/sei-chain/evmrpc/config"
+	"github.com/sei-protocol/sei-chain/sei-db/ledger_db/receipt"
 	rpcclient "github.com/sei-protocol/sei-chain/sei-tendermint/rpc/client"
 	"github.com/sei-protocol/sei-chain/sei-tendermint/rpc/coretypes"
 	tmtypes "github.com/sei-protocol/sei-chain/sei-tendermint/types"
@@ -761,6 +762,16 @@ func (f *LogFetcher) GetLogsByFilters(ctx context.Context, crit filters.FilterCr
 		return nil, 0, fmt.Errorf("block range too large (%d), maximum allowed is %d blocks", blockRange, f.filterConfig.maxBlock)
 	}
 
+	// Try efficient range query first (supported by parquet/DuckDB backend)
+	// #nosec G115 -- begin and end are validated to be positive block heights above
+	if logs, rangeErr := f.tryFilterLogsRange(ctx, uint64(begin), uint64(end), crit); rangeErr == nil {
+		return logs, end, nil
+	} else if !errors.Is(rangeErr, receipt.ErrRangeQueryNotSupported) {
+		// If it's a real error (not just unsupported), return it
+		return nil, 0, rangeErr
+	}
+	// Fall back to block-by-block querying for backends that don't support range queries
+
 	bloomIndexes := EncodeFilters(crit.Addresses, crit.Topics)
 	blocks, end, applyOpenEndedLogLimit, err := f.fetchBlocksByCrit(ctx, crit, lastToHeight, bloomIndexes)
 	if err != nil {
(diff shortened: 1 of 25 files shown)


cmd/seid/cmd/init.go
```diff
@@ -144,26 +145,37 @@ func InitCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
...
-				return errors.Wrap(err, "Failed to marshall default genesis state")
-			}
 
-			genDoc := &types.GenesisDoc{}
-			if _, err := os.Stat(genFile); err != nil {
-				if !os.IsNotExist(err) {
-					return err
+			var genDoc *types.GenesisDoc
+			var appState json.RawMessage
+			if genesis.IsWellKnown(chainID) {
```
masih, line 151: I think we want the genesis substitution order to be:
1. `genesis.json` in config folder, if present.
2. if not -->   is chain ID well known ?
   * yes: load from embedded genesis.
   * no: error out with genesis missing

Bonus logic (please feel free to do in a separate PR): 
  * if explicit is present but for a well known chain ID:
     * Check that it is right ( note that the format is JSON and we need to sanitise formatting before checking hash, or... load as struct and compare types equality)
        * if good: log in debug that "hey I know my networks -- you don't have to have the genesis file"
         * if not good: panic out with appropriate messaging. Most likely accidentally edited genesis; invalid for the network in any case.

This would follow the general operational best practice of: "Explicit over implicit", and "Implicit over unspecified".

Does that make sense @mojtaba-esk? WDYT?

cmd/seid/cmd/init.go
```diff
@@ -144,26 +145,37 @@ func InitCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
...
+				appState = genDoc.AppState
 			} else {
-				genDoc, err = types.GenesisDocFromFile(genFile)
+				var err error
+				appState, err = json.MarshalIndent(mbm.DefaultGenesis(cdc), "", " ")
 				if err != nil {
-					return errors.Wrap(err, "Failed to read genesis doc from file")
+					return errors.Wrap(err, "Failed to marshall default genesis state")
+				}
+				genDoc = &types.GenesisDoc{}
```
masih, line 164: Move closer to where it is first used?

cmd/seid/cmd/init.go
```diff
@@ -144,26 +145,37 @@ func InitCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
...
+				if _, err := os.Stat(genFile); err != nil {
```
masih: Also check that `genFile` is not a directory?

app/genesis/genesis_test.go
```diff
@@ -0,0 +1,51 @@
...
+		require.NoError(t, err)
+		require.NotNil(t, genDoc)
+		require.Equal(t, chainID, genDoc.ChainID)
+	}
+}
+
+func TestEmbeddedGenesisUnknownChain(t *testing.T) {
+	_, err := EmbeddedGenesis("unknown")
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "unknown chain-id")
```
masih:
```suggestion
	require.ErrorContains(t, err, "unknown chain-id")
```

app/genesis/genesis_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// SHA-256
+var expectedGenesisDigests = map[string]string{
+	"arctic-1":   "a8b60161e345d8afb0d9b0c524b6541e6135f5eda4092fae287d8496e14554d8",
+	"atlantic-2": "3c135db9177a428893353d7889149ca2ed9c075d6846be07af60354022b81318",
+	"pacific-1":  "4304cf1c7f46d153b79f1195b2d334f7f7cf02f26e02a3bb77c544a4987c1432",
```
masih: These are the hashes of the indented JSON, and technically the same JSON values in different order/indentation is still valid. 

But then we will/should never edit those JSON files anyway so hardcoding the hashes of pretty JSON is probably fine 👍

app/genesis/genesis_test.go
```diff
@@ -0,0 +1,51 @@
...
+func TestWellKnownChainIDs(t *testing.T) {
+	ids := WellKnownChainIDs()
+	require.EqualValues(t, wellKnownChainIDs, ids)
+}
+
+func TestEmbeddedGenesisDoc(t *testing.T) {
+	for _, chainID := range WellKnownChainIDs() {
+		genDoc, err := EmbeddedGenesisDoc(chainID)
+		require.NoError(t, err)
+		require.NotNil(t, genDoc)
```
masih, line 77: not nil is implied by the check in the next line, right?

app/genesis/genesis.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/tendermint/tendermint/types"
+)
+
+//go:embed chains/*.json
+var embeddedGenesisFS embed.FS
+
+var wellKnownChainIDs = []string{"arctic-1", "atlantic-2", "pacific-1"}
+
+func WellKnownChainIDs() []string {
```
masih: This is used for testing only -- I would recommend hardcoding known chain IDs in tests instead and getting rid of this function. Alternatively, in tests you can get the list by `ls`ing the `chains/*.json` files which has an added benfit that the code _does_ include all json files put in there.

app/genesis/genesis.go
```diff
@@ -0,0 +1,46 @@
...
+	return append([]string(nil), wellKnownChainIDs...)
```
masih: If you decide to keep this function, `slices.Clone` is more idiomatic.

app/genesis/genesis.go
```diff
@@ -0,0 +1,46 @@
...
+	"fmt"
+	"slices"
+
+	"github.com/tendermint/tendermint/types"
+)
+
+//go:embed chains/*.json
+var embeddedGenesisFS embed.FS
+
+var wellKnownChainIDs = []string{"arctic-1", "atlantic-2", "pacific-1"}
```
masih: An edge-case can occure here where someone adds the JSON but forgets to update this. 

We can avoid it by loading the known IDs via package `Init`, which would get rid of this struct entirely and makes the JSON files in chains directory the sole source of truth.

app/genesis/genesis.go
```diff
@@ -0,0 +1,46 @@
...
+func EmbeddedGenesisDoc(chainID string) (*types.GenesisDoc, error) {
+	data, err := EmbeddedGenesis(chainID)
+	if err != nil {
+		return nil, err
+	}
+	genDoc, err := types.GenesisDocFromJSON(data)
+	if err != nil {
+		return nil, fmt.Errorf("parsing embedded genesis for %s: %w", chainID, err)
+	}
+	return genDoc, nil
```
masih, line 72: Check that genesis doc chain ID matches the given chain ID in case JSON files are fat fingered by accident?

app/genesis/genesis.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+	return data, nil
+}
+
+func EmbeddedGenesisDoc(chainID string) (*types.GenesisDoc, error) {
+	data, err := EmbeddedGenesis(chainID)
+	if err != nil {
+		return nil, err
+	}
+	genDoc, err := types.GenesisDocFromJSON(data)
```
masih, line 65: Thinking out loud: Unfortunately `ValidateAndComplete` are coupled. Maybe worth refactoring that function into `Validate` and `ValidateAndComplete`, then calling `Validate` here after the files are loaded. 

It is a one off validation, and would avoid having faulty nodes running on any network by accident.

app/genesis/genesis.go
mojtaba-esk: Good idea. will apply it

app/genesis/genesis.go
mojtaba-esk: Thanks for the hint

app/genesis/genesis.go
mojtaba-esk: That would be safer. `ValidateAndComplete` currently fills in some default values if they are empty like `ConsensusParams`, `GenesisTime`, `InitialHeight` and these values are not supposed to be empty on a genesis file of a running network. So I think it is fine to keep it as it is for the moment. WDYT?
We could refactor it in a followup.

app/genesis/genesis.go
mojtaba-esk: Good idea, will do it.

mojtaba-esk: added, although it is evaluated in the tests as well.

app/genesis/genesis_test.go
mojtaba-esk: This is a critical point. I will see if I can hash it without considering whitespaces. Like the gendoc itself. if that works, the indents are fine as on some IDEs it is configured as one or two tabs one some it uses spaces... etc. Also the ordering should be fine as long as the gendoc structure is not touched.

app/genesis/genesis_test.go
mojtaba-esk: it is implied for the `genDoc.ChainID` but for the `genDoc` if under any circumstance it returns as a nil without an error, the next line throws a panic.

cmd/seid/cmd/init.go
mojtaba-esk: Thanks for the detailed comment. 
> genesis.json in config folder, if present. 
This is checked a few lines above so it does not reach to this point when genesis.json exists.

Will apply the `2` as this makes sense.

Let's do the rest (Bonus logic) in a followup.

mojtaba-esk: oh I see what you mean now.  Created a followup for it: https://linear.app/seilabs/issue/PLT-116/validate-explicit-genesis-file-for-well-known-chains-in-seid-init

cmd/seid/cmd/init.go
mojtaba-esk: we can't move this as it needs to have an empty gendoc in case the file does not exist and after the follwoing if statement, the given ChainID and the default values like `appState` is assigned to it.

cmd/seid/cmd/init.go
mojtaba-esk: We could do that sure.

app/genesis/genesis_test.go
mojtaba-esk: Applied

app/genesis/genesis_test.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+// SHA-256(genDoc)
+var expectedGenesisDigests = map[string]string{
+	"arctic-1":   "6d091f04f578537a1715d605ec2efc120f743a0d26a2ae3738bbe2ffbab652c3",
+	"atlantic-2": "d9291825bcdc6c333dbcb2232d38bfb89cbd03f75932bbf1c9738842e34a2315",
+	"pacific-1":  "3a1f5d87df75f4fdb85eaf1b506e080cbcee3a748048e1e80721f68eb2193e43",
+}
+
+func genesisDocDigest(chainID string) ([]byte, error) {
```
mojtaba-esk, line 21: Here the limitation is that we do not order the `appState` to have a canonical value of it as it would have added extra complexity to the code that seems not be required at this stage.

app/genesis/genesis.go
masih: SGTM 👍

cmd/seid/cmd/init.go
masih: @mojtaba-esk I think we are still missing item No. 1 in my comment right? 

We want to check if the genesis is explicitly set first, and only implicitly load for well known chains if it is _absent_.

masih (approved):

PR #2850, opened 2026-02-11
perf: set GOGC=200 to reduce GC frequency

pdrobnjak:
## Summary
Increase `GOGC` to 200 to reduce garbage collection frequency

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -440,6 +440,8 @@ func New(
 	appOptions []AppOption,
 	baseAppOptions ...func(*baseapp.BaseApp),
 ) *App {
+	debug.SetGCPercent(200)
+
 	appCodec := encodingConfig.Marshaler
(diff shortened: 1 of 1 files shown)


app/app.go
```diff
@@ -440,6 +440,8 @@ func New(
 	appOptions []AppOption,
 	baseAppOptions ...func(*baseapp.BaseApp),
 ) *App {
+	debug.SetGCPercent(200)
```
masih, line 443: Hm... we shouldn't do this on the main app path.

_If_ a node operator wants this, they can always set it as env var. 

For testing purposes I also recommend using env var.  

In any case, this diff should not land on `main`

pdrobnjak: Ye, aware, I'm just getting my perf changes in order.

Don't review anything for now, I'm still checking on them.

PR #2857, opened 2026-02-11
fix: use MADV_RANDOM during loadtree

blindchaser:

- NewMmap applies MADV_RANDOM at creation
- Remove dead code: NewMmapRandom, PrepareForSequentialRead
- Remove redundant post-load PrepareForRandomRead loops
- Fix FillDefaults: treat threshold<=0 as unset (default 0.8)
- Fix log format: remove trailing \n causing blank lines

### DIFF
--- a/sei-db/state_db/sc/memiavl/tree.go
+++ b/sei-db/state_db/sc/memiavl/tree.go
@@ -331,13 +331,6 @@ func (t *Tree) ReplaceWith(other *Tree) error {
 	t.cowVersion = other.cowVersion
 	t.zeroCopy = other.zeroCopy
 
-	// Safety net: ensure MADV_RANDOM is set on the new snapshot's mmap files.
-	if t.snapshot != nil {
-		t.snapshot.nodesMap.PrepareForRandomRead()
-		t.snapshot.leavesMap.PrepareForRandomRead()
-		t.snapshot.kvsMap.PrepareForRandomRead()
-	}
-
 	if snapshot != nil {
 		return snapshot.Close()
 	}


sei-db/state_db/sc/memiavl/mmap.go
```diff
@@ -19,9 +19,31 @@ type MmapFile struct {
...
+	if err != nil {
+		return nil, err
+	}
+	mmapFile.PrepareForSequentialRead()
+	return mmapFile, nil
+}
+
+// NewMmapRandom opens the file and creates a read-only memory mapping with MADV_RANDOM.
+// Does not apply MADV_WILLNEED; pages are loaded on-demand.
+func NewMmapRandom(path string) (*MmapFile, error) {
```
yzang2019: Where is this being called?

sei-db/state_db/sc/memiavl/snapshot.go
```diff
@@ -180,13 +180,15 @@ func OpenSnapshot(snapshotDir string, opts Options) (*Snapshot, error) {
 		return errors.Join(errs...)
 	}
 
-	if nodesMap, err = NewMmap(filepath.Join(snapshotDir, FileNameNodes)); err != nil {
+	// Load snapshot mmap files with MADV_RANDOM.
+	// Snapshot prefetch is handled separately by prefetchSnapshot() at the end of this function.
+	if nodesMap, err = NewMmapRandom(filepath.Join(snapshotDir, FileNameNodes)); err != nil {
```
yzang2019: I thought we should only do that when we open snapshot for creation not for restart?

Kbhat1:
is there any clean way to unit test this?

sei-db/state_db/sc/memiavl/mmap.go
blindchaser: in sc/memiavl/snapshot.go -> OpenSnapshot()

sei-db/state_db/sc/memiavl/snapshot.go
blindchaser: this won't impact prefetchSnapshot() during restart, prefetch uses separate file I/O (not mmap), so it's unaffected by the mmap's madvise hint. The difference between background creation and restart is whether we run prefetchSnapshot(): on restart we do (cold cache), on background reload we don't (data is still hot from the write).

blindchaser: feels MADV_RANDOM is the correct hint for later WAL replay/catchup, and normal serving?

yzang2019: Discussed offline, let's fix it and move this logic to the correct location

philipsu522 (approved):

PR #2865, opened 2026-02-11
Fix: validate AppQC/CommitQC index alignment

wen-coding:
## Summary
- Reject mismatched AppQC/CommitQC indices and surface errors.
- Propagate prune errors to avoid silent state corruption.
- Add regression tests for mismatched QC handling.
- Port of sei-v3 PR #509.

## Test Plan
- `go test ./sei-tendermint/internal/autobahn/...`


sei-tendermint/internal/autobahn/avail/state_test.go
```diff
@@ -183,3 +184,92 @@ func TestState(t *testing.T) {
...
+
+		// Push first 2 votes (no quorum yet for 4 nodes, quorum is 3)
+		require.NoError(state.PushAppVote(ctx, votes[0]))
+		require.NoError(state.PushAppVote(ctx, votes[1]))
+
+		// 3. Now manually corrupt the internal state to force a mismatch.
+		// We replace the CommitQC at index 0 with a CommitQC that has a different index (index 1).
+		qc1 := makeQC(utils.Some(qc0), nil)
+
+		for inner, ctrl := range state.inner.Lock() {
```
pompon0: I haven't previously paid attention to this test. Why are we corrupting the internal state here? Can we somehow create a realistic scenario instead?

wen-coding: That's a good point, my first time using Cursor and I found out AI can add unnecessary tests. This should be covered under QC verify, removing these tests.

PR #2866, opened 2026-02-12
skip LastResultsHash check if giga executor is on

codchen:
Giga executors may yield different gas usage than v2 executors due to:
- more parity with Ethereum
- optimized state access pattern
Since we only care about state consistency, which is covered by AppHash, we don't need to require lastResultsHash on nodes that enabled giga.

patch on giga-enabled RPC nodes

### DIFF
--- a/app/seidb.go
+++ b/app/seidb.go
@@ -37,6 +37,11 @@ const (
 	FlagSSPruneInterval     = "state-store.ss-prune-interval"
 	FlagSSImportNumWorkers  = "state-store.ss-import-num-workers"
 
+	// EVM SS optimization (embedded in SS config, controlled via write/read mode)
+	FlagEVMSSDirectory = "state-store.evm-ss-db-directory"
+	FlagEVMSSWriteMode = "state-store.evm-ss-write-mode"
+	FlagEVMSSReadMode  = "state-store.evm-ss-read-mode"
+
 	// Other configs
 	FlagSnapshotInterval = "state-sync.snapshot-interval"
 	FlagMigrateIAVL      = "migrate-iavl"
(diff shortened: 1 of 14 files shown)


sei-tendermint/internal/state/execution.go
```diff
@@ -215,7 +215,7 @@ func (blockExec *BlockExecutor) ValidateBlock(ctx context.Context, state State,
 	err := validateBlock(state, block)
 	if err != nil {
 		// Check if this is a LastResultsHash mismatch and log detailed info
-		if !bytes.Equal(block.LastResultsHash, state.LastResultsHash) {
+		if !types.SkipLastResultsHashValidation.Load() && !bytes.Equal(block.LastResultsHash, state.LastResultsHash) {
```
arajasek, line 218: Does this actually work? I think you're just skipping the log, but still returning the error here, so I don't think `SkipLastResultsHashValidation` actually does anything.

codchen: `err` comes from `validateBlock`, which wouldn't return a non-nil value because of https://github.com/sei-protocol/sei-chain/pull/2866/changes#diff-d3412156734f105e2fc032cd6850dd192eb7f1a039872a127df968c1bffde8f9R65

arajasek: That makes sense, but then why are we changing anything here at all? I think we can leave this file as-is.

arajasek: I think there's value in containing a change like this if possible.

codchen: on a giga node, if there is indeed a validation error that's not LastResultsHash mismatch, but LastResultsHash does mismatch, then the log here would say "LastResultsHash error" instead of the real error.

it's a rare case though

PR #2868, opened 2026-02-12
fix: respect existing genesis file

mojtaba-esk:
This PR is a followup for #2835 to handle a case where the user already has their own genesis file and does not want to remove them or overwrite them. Here is the slack comment by @masih that suggested this fix: https://sei-network-io.slack.com/archives/C06A3D4FWE5/p1770833867974669?thread_ts=1770803742.497719&cid=C06A3D4FWE5


cmd/seid/cmd/init.go
```diff
@@ -233,3 +200,67 @@ func InitCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
...
+		}
+		return genDoc, nil
+	}
+
+	appState, err := json.MarshalIndent(mbm.DefaultGenesis(cdc), "", " ")
+	if err != nil {
+		return nil, errors.Wrap(err, "Failed to marshall default genesis state")
+	}
+	genDoc := &types.GenesisDoc{ChainID: chainID, AppState: appState}
+	if fi, err := os.Stat(genFile); err == nil {
```
masih: Ditto re DIR check. We probably can refactor the existing file checking to use here as well as above.

cmd/seid/cmd/init_test.go
```diff
@@ -230,3 +235,111 @@ func TestModeConfigurationMatrix(t *testing.T) {
 		})
 	}
 }
+
+// TestLoadOrWriteGenesis_ExplicitConfigWins: existing genesis + no overwrite leaves file unchanged.
+func TestLoadOrWriteGenesis_ExplicitConfigWins(t *testing.T) {
+	dir := t.TempDir()
+	genFile := filepath.Join(dir, "genesis.json")
+	chainID := "atlantic-2"
```
masih: I would define this as `const` instead of `var`? Ditto for other such names constant in this file.

masih (approved):
Nice thank you for getting this sorted so quickly 🚀 

No blockers; just minor comments.

sei-will (approved):

PR #2873, opened 2026-02-12
fix to halt due to reconstructing block from bad proposal (backported #2823)

pompon0:


sei-tendermint/internal/consensus/state.go
```diff
@@ -827,15 +827,12 @@ func (cs *State) handleMsg(ctx context.Context, mi msgInfo, fsyncUponCompletion
...
-		if err = cs.setProposal(msg.Proposal, mi.ReceiveTime); err == nil {
-			if key, ok := cs.privValidatorPubKey.Get(); ok && cs.config.GossipTransactionKeyOnly {
-				if !cs.isProposer(key.Address()) && cs.roundState.ProposalBlock() == nil {
-					created := cs.tryCreateProposalBlock(spanCtx, msg.Proposal.Height, msg.Proposal.Round, msg.Proposal.Header, msg.Proposal.LastCommit, msg.Proposal.Evidence, msg.Proposal.ProposerAddress)
-					if created {
-						cs.fsyncAndCompleteProposal(ctx, fsyncUponCompletion, msg.Proposal.Height, span, true)
-					}
-				}
-			}
+		err = cs.setProposal(msg.Proposal, mi.ReceiveTime)
```
masih, line 830: Avoid re-declaring `err` considering the `break`?

pompon0: the error is logged after the switch clause. Don't ask me why.

PR #2879, opened 2026-02-12
Add more metrics for snapshot and state sync 

yzang2019:
This PR is adding more visibility around MemIAVL snapshot creation + replay + pruning, as well as state sync snapshot creation process.

With these metrics, we should have better visibility to correlate some timing for performance changes in relate to the snapshot behavior

Tested locally and verified the metrics works


sei-db/state_db/sc/memiavl/db.go
```diff
@@ -541,10 +546,12 @@ func (db *DB) pruneSnapshots() {
 		}
 
 		name := snapshotName(version)
-		db.logger.Info("prune snapshot", "name", name)
 
 		if err := atomicRemoveDir(filepath.Join(db.dir, name)); err != nil {
 			db.logger.Error("failed to prune snapshot", "err", err)
+		} else {
+			db.logger.Info("successfully pruned snapshot", "name", name)
+			otelMetrics.SnapshotPruneCount.Add(context.Background(), 1)
```
masih: We probably want to measure the failure rate too right? 

In which case, you can use the same metric and tag by `status`?

yzang2019: Good idea!

sei-db/state_db/sc/memiavl/snapshot.go
```diff
@@ -864,22 +856,6 @@ func (w *snapshotWriter) writeKeyValueDirect(key, value []byte) error {
 
 // writeLeaf sends leaf and KV write operations to the pipeline
 func (w *snapshotWriter) writeLeaf(version uint32, key, value, hash []byte) error {
-	// Track channel fill metrics for all channels
```
yzang2019, line 867: Removing these since it seems they are not being used

sei-db/state_db/sc/memiavl/metrics.go
```diff
@@ -9,24 +9,64 @@ var (
...
-			"memiavl_snapshot_creation_latency",
-			metric.WithDescription("Time taken to create memiavl snapshot"),
+		SnapshotRewriteLatency: must(meter.Float64Histogram(
+			"memiavl_snapshot_rewrite_latency",
+			metric.WithDescription("Time taken to write to the new memiavl snapshot"),
 			metric.WithUnit("s"),
 		)),
+		SnapshotRewriteCount: must(meter.Int64Counter(
+			"memiavl_snapshot_rewrite_count",
+			metric.WithDescription("Total num of times memiavl snapshot rewrite attempts"),
```
blindchaser: nit: grammar seems a little weird, how about: "Total number of memiavl snapshot rewrite attempts"

yzang2019: Yup, that looks better!

go.work.sum
```diff
@@ -0,0 +1,1684 @@
+bazil.org/fuse v0.0.0-20200407214033-5883e5a4b512 h1:SRsZGA7aFnCZETmov57jwPrWuTmaZK6+4R4v5FUe1/c=
```
masih, line 1: Sigh; looks like I missed this being added back in @yzang2019 .... Going to delete it

yzang2019: ah sorry, add back accidently, this was an old PR that I forgot to merge, thanks!

masih: No worries, sorry for missing it in my review of the PR

PR #2881, opened 2026-02-13
Add unjail to staking precompile

codchen:
Add `unjail` function to the staking precompile

unit test


precompiles/staking/staking_test.go
```diff
@@ -2603,3 +2615,111 @@ func TestStakingPrecompileStaticCallPrevention(t *testing.T) {
 		_ = err
 	})
 }
+
+func TestUnjail(t *testing.T) {
```
philipsu522, line 2619: should we verify events are emitted?

should we add some more test cases? thinking:
- calling unjail on a non-validator
- calling unjail with improper params (not enough fees, wrong key...etc)

PR #2887, opened 2026-02-13
fixed autorestart cooldown

pompon0:
There was a semantic mismatch between blocksync sending at most 1 message to the restartCh, and WaitForQuitSignals which ignored messages until cooldown has passed. I think that cooldown in blocksync logic was exactly trying to avoid sending messages when WaitForQuitSignals could have ignored them, which is a very fragile way of doing things. This PR removes cooldown from WaitForQuitSignals, simply delegating the cooldown logic to blocksync and makes the signal nonblocking and idempotent.

### DIFF
--- a/sei-tendermint/internal/autobahn/autobahn.proto
+++ b/sei-tendermint/internal/autobahn/autobahn.proto
@@ -154,6 +154,26 @@ message FullTimeoutVote {
   optional PrepareQC latest_prepare_qc = 2;
 }
 
+// Persisted consensus state (for crash recovery).
+// Only persist votes and QCs here - these are the canonical consensus artifacts.
+// Do NOT persist internal derived fields (e.g., cached computations, runtime state).
+// Derived fields are implementation-dependent and should be recomputed on load.
+message PersistedInner {
+  optional CommitQC commit_qc = 1;
+  optional PrepareQC prepare_qc = 2;
+  optional TimeoutQC timeout_qc = 3;
+
+  optional SignedMsg commit_vote = 4;
+  optional SignedMsg prepare_vote = 5;
+  optional FullTimeoutVote timeout_vote = 6;
+}
+
+// Wrapper for persisted data with sequence number.
+message PersistedWrapper {
+  optional uint64 seq = 1;
+  optional bytes data = 2;
+}
+
 message AppQC {
   AppProposal vote = 1;
   repeated Signature sigs = 2;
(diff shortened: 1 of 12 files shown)


sei-tendermint/config/config.go
```diff
@@ -1398,6 +1398,7 @@ type SelfRemediationConfig struct {
 	BlocksBehindCheckIntervalSeconds uint64 `mapstructure:"blocks-behind-check-interval"`
 
 	// Cooldown between each restart
+	// DEPRECATED
```
masih: The predominant pattern in this repo (and other libraries I have come across) is Upper D instead of ALL CAPS.
Bonus: add some extra info for the users. Something like this:

```suggestion
	// Deprecated: The cooldown mechanism is completely removed. As a result this configuration has no effect.
```

sei-cosmos/server/util.go
```diff
@@ -408,26 +407,15 @@ func TrapSignal(cleanupFunc func()) {
...
-				}
-				return ErrorCode{Code: RestartErrorCode}
-			}
-		}
-	} else {
-		sig := <-sigs
+	select {
+	case sig := <-sigs:
+		// TODO: why +128?
 		return ErrorCode{Code: int(sig.(syscall.Signal)) + 128}
```
masih: This error gets bubbled up to the surface, which ultimately results in node existing with a non-zero exit code. 

The non-zero exit code as a result of `SIGINT?SIGTERM` seems unreasonable? It makes it impossible (i think?) to "gracefully" stop the node such that it exits with error code 0.

sei-cosmos/server/util.go
```diff
@@ -408,26 +407,15 @@ func TrapSignal(cleanupFunc func()) {
 }
 
 // WaitForQuitSignals waits for SIGINT and SIGTERM and returns.
-func WaitForQuitSignals(ctx *Context, restartCh chan struct{}, canRestartAfter time.Time) ErrorCode {
+func WaitForQuitSignals(ctx *Context, restartCh chan struct{}) error {
 	sigs := make(chan os.Signal, 1)
 	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
```
masih: Within "soft-restart" codepath execution, this registeration will happen every time `WaitForQuitSignals` is called.

We should defer signal stop.
```suggestion
	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
	defer signal.Stop(sigs)
```

PR #2892, opened 2026-02-13
Fix potential auto restart failure due to blocksync always true

yzang2019:
Problem identified:
- In autoRestartIfBehind we use the blockSync as a flag to determine if already lag behind to avoid additional restart signals.
- At the same time, the restart has a "cool down" period -- meaning, restart signals during cool down period are silently ignored.
- Now, if a node restarts during cool down period, the signal gets ignored and the blockSync flag will remain indefinitely true effectively disabling the self-remediation process

Fix:
- Do not set blocksync mode in auto restart handler
Added unit test


masih:
@yzang2019 OK to close this?

PR #2896, opened 2026-02-16
consensus: persist AppQC, blocks, and CommitQCs with async persistence

wen-coding:
## Summary

Replace file-per-entry persistence for blocks and CommitQCs with WAL-based persistence (`sei-db/wal`), wrapped in a generic `indexedWAL[T]` that tracks monotonic indices and typed entries.

Ref: sei-protocol/sei-v3#512

### Generic WAL wrapper (`persist/wal.go`)

- **`indexedWAL[T]`**: Wraps `sei-db/wal` with monotonic index tracking (`firstIdx`, `nextIdx`) and typed codec-based serialization. Provides `Write`, `ReadAll`, `TruncateBefore` (with verify callback), `TruncateAll`, `Count`, and `FirstIdx`. Configured with synchronous unbatched writes and fsync.
- **`codec[T]` interface**: Marshal/unmarshal pair satisfied by `protoutils.Conv[T, P]`.

### Block persistence (`persist/blocks.go`)

Three-layer architecture:

- **`laneWALState`** (lock-free): Low-level WAL operations — `persistBlock` (contiguity-checked append), `truncateForAnchor` (truncate WAL so a given block number becomes the oldest entry), `loadAll` (replay with gap detection). Tracks `nextBlockNum` derived from WAL count.
- **`laneWAL`**: Wraps `laneWALState` with a `utils.Mutex` for per-lane write serialization.
- **`BlockPersister`**: Top-level API managing a `utils.RWMutex[map[LaneID]*laneWAL]`. Each lane gets its own WAL in a hex-encoded subdirectory, so truncation is independent per lane. Double-checked locking in `getOrCreateLane`. No-op mode when `dir` is `None`.

**`MaybePruneAndPersistLane`** is the single public write API:
- anchor set + proposals non-empty: truncate then append (runtime path)
- anchor set + proposals empty: truncate only (startup prune)
- anchor empty + proposals non-empty: append only
- anchor empty + proposals empty: no-op

The per-lane lock is held for the entire truncate-then-append sequence. `afterEach` callback (as `utils.Option[func(T)]`) is invoked under lock after each successful append.

`truncateForAnchor` takes `BlockNumber` directly (not `CommitQC`), so `testDeleteBefore` calls it without duplicating logic.

**Fix**: `truncateForAnchor` on an empty WAL now advances `nextBlockNum` when the anchor is ahead, preventing "out of sequence" errors on the next `persistBlock`.

### CommitQC persistence (`persist/commitqcs.go`)

Two-layer architecture:

- **`commitQCState`** (lock-free): `persistCommitQC` (duplicate-tolerant, gap-rejecting append), `deleteBefore` (truncate + optional anchor re-persist for crash recovery).
- **`CommitQCPersister`**: Wraps state with `utils.Mutex`. No-op mode when `iw` is `None`.

**`MaybePruneAndPersist`** is the single public write API (same four-case matrix as blocks). Lock held for entire truncate-then-append. `afterEach` invoked under lock.

### Caller changes (`avail/state.go`, `avail/inner.go`)

- `NewState` startup prune calls `MaybePruneAndPersistLane` / `MaybePruneAndPersist` instead of the removed `DeleteBefore`.
- `runPersist` uses `scope.Parallel` to fan out per-lane block writes and commitQC writes concurrently via the new APIs. `afterEach` callbacks (`markBlockPersisted`, `markCommitQCsPersisted`) advance in-memory cursors under the WAL lock.
- Minor comment clarifications in `inner.go`.

## Test plan

- [x] `persist/wal_test.go`: open/reopen, Write + ReadAll, TruncateBefore with verify, TruncateAll, Count/FirstIdx tracking, empty WAL behavior, out-of-range truncation, verify callback rejection, reopen after truncation
- [x] `persist/blocks_test.go`: empty dir, load existing lanes, persist + reload, gap detection on load, truncate (partial, past-all, on empty WAL cursor advancement), stale block rejection, lane mismatch, out-of-sequence/duplicate rejection, no-op persister, empty lane survives reopen, concurrent distinct lanes, crash recovery round-trip
- [x] `persist/commitqcs_test.go`: empty dir, persist + reload, gap detection, truncate (partial, past-all, with/without anchor re-persist), duplicate tolerance, out-of-sequence rejection, no-op persister, LoadNext tracking, close idempotency, crash recovery round-trip
- [x] `avail/state_test.go`: fresh start, load blocks, load commitQCs, load both, load with AppQC anchor, non-contiguous error, corrupt data error, startup prune, empty WAL cursor fix scenario

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,4 @@ integration_test/dapp_tests/artifacts
 
 # DB benchmark data
 sei-db/state_db/bench/cryptosim/data/**
+sei-db/state_db/bench/cryptosim/bin/
--- a/app/ante/evm_delivertx.go
+++ b/app/ante/evm_delivertx.go
@@ -90,6 +90,7 @@ func DecorateNonceCallback(ctx sdk.Context, ek *evmkeeper.Keeper, evmAddr common
 		// bump nonce if it is for some reason not incremented (e.g. ante failure)
 		if ek.GetNonce(callCtx, evmAddr) == startingNonce {
 			ek.SetNonce(callCtx, evmAddr, startingNonce+1)
+			ek.SetNonceBumped(callCtx)
 		}
 	})
 }
--- a/docker/localnode/Dockerfile
+++ b/docker/localnode/Dockerfile
@@ -1,10 +1,22 @@
+FROM golang:1.25.6-bookworm AS go-dist
+
 FROM ubuntu:latest
 ENV HOME="/root" PATH="/root/go/bin:/sei-protocol/sei-chain/integration_test/upgrade_module/scripts/:$PATH"
 RUN apt-get update && apt-get install -y make build-essential git jq python3 curl vim uuid-runtime
-RUN curl -L https://go.dev/dl/go1.25.6.linux-amd64.tar.gz | tar xvzf - -C /usr/local/
-RUN curl -L https://foundry.paradigm.xyz | bash
+COPY --from=go-dist /usr/local/go /usr/local/go
+RUN set -eux; \
+    curl -fsSL https://foundry.paradigm.xyz | bash; \
+    for attempt in 1 2 3 4 5; do \
+      if /root/.foundry/bin/foundryup; then \
+        break; \
+      fi; \
+      if [ "${attempt}" -eq 5 ]; then \
+        echo "foundryup failed after retries"; \
+        exit 1; \
+      fi; \
+      sleep 2; \
+    done
 RUN curl -sL https://deb.nodesource.com/setup_16.x | bash
-RUN /root/.foundry/bin/foundryup
 RUN mkdir -p /root/go/pkg/mod && \
     mkdir -p /root/.cache && \
     chmod -R a+rwX /root
--- a/docker/rpcnode/Dockerfile
+++ b/docker/rpcnode/Dockerfile
@@ -1,8 +1,10 @@
+FROM golang:1.25.6-bookworm AS go-dist
+
 FROM ubuntu:latest
 ENV HOME="/root" PATH="/root/go/bin:$PATH"
 RUN apt-get update && \
     apt-get install -y make git build-essential jq python3 curl vim uuid-runtime
-RUN curl -L https://go.dev/dl/go1.25.6.linux-amd64.tar.gz | tar xvzf - -C /usr/local/
+COPY --from=go-dist /usr/local/go /usr/local/go
 RUN mkdir -p /root/go/pkg/mod && \
     mkdir -p /root/.cache && \
     chmod -R a+rwX /root
--- a/evmrpc/block.go
+++ b/evmrpc/block.go
@@ -342,7 +342,11 @@ func EncodeTmBlock(
 		case *types.MsgEVMTransaction:
 			ethtx, _ := m.AsTransaction()
 			hash := ethtx.Hash()
-			receipt, _ := k.GetReceipt(latestCtx, hash)
+			receipt, err := k.GetReceipt(latestCtx, hash)
+			if err != nil {
+				// tx doesn't have a receipt because of nonce mismatch
+				continue
+			}
 			if !fullTx {
 				transactions = append(transactions, hash.Hex())
 			} else {
--- a/evmrpc/info.go
+++ b/evmrpc/info.go
@@ -292,7 +292,8 @@ func (i *InfoAPI) getRewards(block *coretypes.ResultBlock, baseFee *big.Int, rew
 		// okay to get from latest since receipt is immutable
 		receipt, err := i.keeper.GetReceipt(i.ctxProvider(LatestCtxHeight), ethtx.Hash())
 		if err != nil {
-			return nil, err
+			// tx doesn't have a receipt because of nonce mismatch
+			continue
 		}
 		receiptEffectiveGasPrice := new(big.Int).SetUint64(receipt.EffectiveGasPrice)
 		if receiptEffectiveGasPrice.Cmp(baseFee) < 0 {
--- a/giga/deps/xevm/state/accesslist.go
+++ b/giga/deps/xevm/state/accesslist.go
@@ -64,7 +64,6 @@ func (s *DBImpl) AddSlotToAccessList(addr common.Address, slot common.Hash) {
 }
 
 func (s *DBImpl) Prepare(_ params.Rules, sender, coinbase common.Address, dest *common.Address, precompiles []common.Address, txAccesses ethtypes.AccessList) {
-	s.Snapshot()
 	s.AddAddressToAccessList(sender)
 	if dest != nil {
 		s.AddAddressToAccessList(*dest)
--- a/sei-db/common/metrics/pebble_metrics.go
+++ b/sei-db/common/metrics/pebble_metrics.go
@@ -158,6 +158,63 @@ type PebbleMetrics struct {
 	fileCacheHits          metric.Int64Counter
 	fileCacheMisses        metric.Int64Counter
 
+	// prev* track last scraped cumulative values so we Add(delta) not Add(total).
+	prevCompactionCount                 int64
+	prevCompactionCancelledCount        int64
+	prevCompactionCancelledBytes        int64
+	prevCompactionFailedCount           int64
+	prevCompactionDefaultCount          int64
+	prevCompactionDeleteOnlyCount       int64
+	prevCompactionElisionOnlyCount      int64
+	prevCompactionCopyCount             int64
+	prevCompactionMoveCount             int64
+	prevCompactionReadCount             int64
+	prevCompactionTombstoneDensityCount int64
+	prevCompactionRewriteCount          int64
+	prevCompactionMultiLevelCount       int64
+	prevCompactionBlobFileRewriteCount  int64
+	prevCompactionCounterLevelCount     int64
+	prevIngestCount                     int64
+	prevFlushCount                      int64
+	prevFlushBytesWritten               int64
+	prevFlushAsIngestCount              int64
+	prevFlushAsIngestTableCount         int64
+	prevFlushAsIngestBytes              int64
+	prevFilterHits                      int64
+	prevFilterMisses                    int64
+	prevWalBytesIn                      int64
+	prevWalBytesWritten                 int64
+	prevWalFailoverDirSwitchCount       int64
+	prevKeysMissizedTombstonesCount     int64
+	prevSnapshotPinnedKeys              int64
+	prevSnapshotPinnedSize              int64
+	prevFileCacheHits                   int64
+	prevFileCacheMisses                 int64
+	prevCacheHits                       int64
+	prevCacheMisses                     int64
+
+	// prev*ByLevel hold previous cumulative values per level (index = level).
+	prevCompactionBytesReadByLevel            []int64
+	prevCompactionBytesWrittenByLevel         []int64
+	prevSstableBytesIngestedByLevel           []int64
+	prevSstableBytesMovedByLevel              []int64
+	prevSstableBytesReadByLevel               []int64
+	prevSstableBytesFlushedByLevel            []int64
+	prevSstableTablesCompactedByLevel         []int64
+	prevSstableTablesFlushedByLevel           []int64
+	prevSstableTablesIngestedByLevel          []int64
+	prevSstableTablesMovedByLevel             []int64
+	prevSstableTablesDeletedByLevel           []int64
+	prevSstableTablesExcisedByLevel           []int64
+	prevSstableBlobBytesReadEstimateByLevel   []int64
+	prevSstableBlobBytesCompactedByLevel      []int64
+	prevSstableBlobBytesFlushedByLevel        []int64
+	prevSstableMultiLevelBytesInTopByLevel    []int64
+	prevSstableMultiLevelBytesInByLevel       []int64
+	prevSstableMultiLevelBytesReadByLevel     []int64
+	prevSstableBytesWrittenDataBlocksByLevel  []int64
+	prevSstableBytesWrittenValueBlocksByLevel []int64
+
 	walFailoverDirSwitchCount    metric.Int64Counter
 	walFailoverPrimaryDuration   metric.Float64Gauge
 	walFailoverSecondaryDuration metric.Float64Gauge
@@ -1120,6 +1177,17 @@ func uint64ToInt64Clamped(v uint64) int64 {
 	return int64(v)
 }
 
+// addDelta computes the difference between current and prev, updates prev to current,
+// and adds the positive delta to the counter. Used to convert cumulative scraped
+// values into rate/counter increments.
+func addDelta(ctx context.Context, counter metric.Int64Counter, current int64, prev *int64, opts ...metric.AddOption) {
+	delta := current - *prev
+	*prev = current
+	if delta > 0 {
+		counter.Add(ctx, delta, opts...)
+	}
+}
+
 // recordFromPebble fetches the current metrics from the Pebble DB via Metrics(), then
 // records compaction, flush, level, memtable, WAL, and cache metrics with the configured
 // database name as the "db" attribute.
@@ -1131,7 +1199,8 @@ func (pm *PebbleMetrics) recordFromPebble(ctx context.Context) {
 	dbAttr := attribute.String("db", pm.databaseName)
 
 	if pm.compactionCount != nil {
-		pm.compactionCount.Add(ctx, m.Compact.Count, metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.compactionCount, m.Compact.Count,
+			&pm.prevCompactionCount, metric.WithAttributes(dbAttr))
 	}
 	if pm.compactionDuration != nil {
 		pm.compactionDuration.Record(ctx, m.Compact.Duration.Seconds(), metric.WithAttributes(dbAttr))
@@ -1147,46 +1216,60 @@ func (pm *PebbleMetrics) recordFromPebble(ctx context.Context) {
 		pm.compactionNumInProgress.Record(ctx, m.Compact.NumInProgress, metric.WithAttributes(dbAttr))
 	}
 	if pm.compactionCancelledCount != nil {
-		pm.compactionCancelledCount.Add(ctx, m.Compact.CancelledCount, metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.compactionCancelledCount, m.Compact.CancelledCount,
+			&pm.prevCompactionCancelledCount, metric.WithAttributes(dbAttr))
 	}
 	if pm.compactionCancelledBytes != nil {
-		pm.compactionCancelledBytes.Add(ctx, m.Compact.CancelledBytes, metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.compactionCancelledBytes, m.Compact.CancelledBytes,
+			&pm.prevCompactionCancelledBytes, metric.WithAttributes(dbAttr))
 	}
 	if pm.compactionFailedCount != nil {
-		pm.compactionFailedCount.Add(ctx, m.Compact.FailedCount, metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.compactionFailedCount, m.Compact.FailedCount,
+			&pm.prevCompactionFailedCount, metric.WithAttributes(dbAttr))
 	}
 	if pm.compactionDefaultCount != nil {
-		pm.compactionDefaultCount.Add(ctx, m.Compact.DefaultCount, metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.compactionDefaultCount, m.Compact.DefaultCount,
+			&pm.prevCompactionDefaultCount, metric.WithAttributes(dbAttr))
 	}
 	if pm.compactionDeleteOnlyCount != nil {
-		pm.compactionDeleteOnlyCount.Add(ctx, m.Compact.DeleteOnlyCount, metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.compactionDeleteOnlyCount, m.Compact.DeleteOnlyCount,
+			&pm.prevCompactionDeleteOnlyCount, metric.WithAttributes(dbAttr))
 	}
 	if pm.compactionElisionOnlyCount != nil {
-		pm.compactionElisionOnlyCount.Add(ctx, m.Compact.ElisionOnlyCount, metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.compactionElisionOnlyCount, m.Compact.ElisionOnlyCount,
+			&pm.prevCompactionElisionOnlyCount, metric.WithAttributes(dbAttr))
 	}
 	if pm.compactionCopyCount != nil {
-		pm.compactionCopyCount.Add(ctx, m.Compact.CopyCount, metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.compactionCopyCount, m.Compact.CopyCount,
+			&pm.prevCompactionCopyCount, metric.WithAttributes(dbAttr))
 	}
 	if pm.compactionMoveCount != nil {
-		pm.compactionMoveCount.Add(ctx, m.Compact.MoveCount, metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.compactionMoveCount, m.Compact.MoveCount,
+			&pm.prevCompactionMoveCount, metric.WithAttributes(dbAttr))
 	}
 	if pm.compactionReadCount != nil {
-		pm.compactionReadCount.Add(ctx, m.Compact.ReadCount, metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.compactionReadCount, m.Compact.ReadCount,
+			&pm.prevCompactionReadCount, metric.WithAttributes(dbAttr))
 	}
 	if pm.compactionTombstoneDensityCount != nil {
-		pm.compactionTombstoneDensityCount.Add(ctx, m.Compact.TombstoneDensityCount, metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.compactionTombstoneDensityCount, m.Compact.TombstoneDensityCount,
+			&pm.prevCompactionTombstoneDensityCount, metric.WithAttributes(dbAttr))
 	}
 	if pm.compactionRewriteCount != nil {
-		pm.compactionRewriteCount.Add(ctx, m.Compact.RewriteCount, metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.compactionRewriteCount, m.Compact.RewriteCount,
+			&pm.prevCompactionRewriteCount, metric.WithAttributes(dbAttr))
 	}
 	if pm.compactionMultiLevelCount != nil {
-		pm.compactionMultiLevelCount.Add(ctx, m.Compact.MultiLevelCount, metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.compactionMultiLevelCount, m.Compact.MultiLevelCount,
+			&pm.prevCompactionMultiLevelCount, metric.WithAttributes(dbAttr))
 	}
 	if pm.compactionBlobFileRewriteCount != nil {
-		pm.compactionBlobFileRewriteCount.Add(ctx, m.Compact.BlobFileRewriteCount, metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.compactionBlobFileRewriteCount, m.Compact.BlobFileRewriteCount,
+			&pm.prevCompactionBlobFileRewriteCount, metric.WithAttributes(dbAttr))
 	}
 	if pm.compactionCounterLevelCount != nil {
-		pm.compactionCounterLevelCount.Add(ctx, m.Compact.CounterLevelCount, metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.compactionCounterLevelCount, m.Compact.CounterLevelCount,
+			&pm.prevCompactionCounterLevelCount, metric.WithAttributes(dbAttr))
 	}
 	if pm.compactionNumProblemSpans != nil {
 		pm.compactionNumProblemSpans.Record(ctx, int64(m.Compact.NumProblemSpans), metric.WithAttributes(dbAttr))
@@ -1196,32 +1279,35 @@ func (pm *PebbleMetrics) recordFromPebble(ctx context.Context) {
 	}
 
 	if pm.ingestCount != nil {
-		pm.ingestCount.Add(ctx, uint64ToInt64Clamped(m.Ingest.Count), metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.ingestCount, uint64ToInt64Clamped(m.Ingest.Count),
+			&pm.prevIngestCount, metric.WithAttributes(dbAttr))
 	}
 
 	if pm.flushCount != nil {
-		pm.flushCount.Add(ctx, m.Flush.Count, metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.flushCount, m.Flush.Count, &pm.prevFlushCount, metric.WithAttributes(dbAttr))
 	}
 	if pm.flushDuration != nil {
 		pm.flushDuration.Record(ctx,
 			m.Flush.WriteThroughput.WorkDuration.Seconds(), metric.WithAttributes(dbAttr))
 	}
 	if pm.flushBytesWritten != nil {
-		pm.flushBytesWritten.Add(ctx, m.Flush.WriteThroughput.Bytes, metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.flushBytesWritten, m.Flush.WriteThroughput.Bytes,
+			&pm.prevFlushBytesWritten, metric.WithAttributes(dbAttr))
 	}
 	if pm.flushNumInProgress != nil {
 		pm.flushNumInProgress.Record(ctx, m.Flush.NumInProgress, metric.WithAttributes(dbAttr))
 	}
 	if pm.flushAsIngestCount != nil {
-		pm.flushAsIngestCount.Add(ctx, uint64ToInt64Clamped(m.Flush.AsIngestCount), metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.flushAsIngestCount, uint64ToInt64Clamped(m.Flush.AsIngestCount),
+			&pm.prevFlushAsIngestCount, metric.WithAttributes(dbAttr))
 	}
 	if pm.flushAsIngestTableCount != nil {
-		pm.flushAsIngestTableCount.Add(ctx, uint64ToInt64Clamped(m.Flush.AsIngestTableCount),
-			metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.flushAsIngestTableCount, uint64ToInt64Clamped(m.Flush.AsIngestTableCount),
+			&pm.prevFlushAsIngestTableCount, metric.WithAttributes(dbAttr))
 	}
 	if pm.flushAsIngestBytes != nil {
-		pm.flushAsIngestBytes.Add(ctx,
-			uint64ToInt64Clamped(m.Flush.AsIngestBytes), metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.flushAsIngestBytes, uint64ToInt64Clamped(m.Flush.AsIngestBytes),
+			&pm.prevFlushAsIngestBytes, metric.WithAttributes(dbAttr))
 	}
 	if pm.flushIdleDuration != nil {
 		pm.flushIdleDuration.Record(ctx,
@@ -1229,16 +1315,40 @@ func (pm *PebbleMetrics) recordFromPebble(ctx context.Context) {
 	}
 
 	if pm.filterHits != nil {
-		pm.filterHits.Add(ctx, m.Filter.Hits, metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.filterHits, m.Filter.Hits, &pm.prevFilterHits, metric.WithAttributes(dbAttr))
 	}
 	if pm.filterMisses != nil {
-		pm.filterMisses.Add(ctx, m.Filter.Misses, metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.filterMisses, m.Filter.Misses, &pm.prevFilterMisses, metric.WithAttributes(dbAttr))
 	}
 
 	for level := 0; level < len(m.Levels); level++ {
 		lm := m.Levels[level]
 		levelAttr := attribute.Int("level", level)
 		attrs := metric.WithAttributes(dbAttr, levelAttr)
+
+		// Grow prev slices if needed.
+		for level >= len(pm.prevCompactionBytesReadByLevel) {
+			pm.prevCompactionBytesReadByLevel = append(pm.prevCompactionBytesReadByLevel, 0)
+			pm.prevCompactionBytesWrittenByLevel = append(pm.prevCompactionBytesWrittenByLevel, 0)
+			pm.prevSstableBytesIngestedByLevel = append(pm.prevSstableBytesIngestedByLevel, 0)
+			pm.prevSstableBytesMovedByLevel = append(pm.prevSstableBytesMovedByLevel, 0)
+			pm.prevSstableBytesReadByLevel = append(pm.prevSstableBytesReadByLevel, 0)
+			pm.prevSstableBytesFlushedByLevel = append(pm.prevSstableBytesFlushedByLevel, 0)
+			pm.prevSstableTablesCompactedByLevel = append(pm.prevSstableTablesCompactedByLevel, 0)
+			pm.prevSstableTablesFlushedByLevel = append(pm.prevSstableTablesFlushedByLevel, 0)
+			pm.prevSstableTablesIngestedByLevel = append(pm.prevSstableTablesIngestedByLevel, 0)
+			pm.prevSstableTablesMovedByLevel = append(pm.prevSstableTablesMovedByLevel, 0)
+			pm.prevSstableTablesDeletedByLevel = append(pm.prevSstableTablesDeletedByLevel, 0)
+			pm.prevSstableTablesExcisedByLevel = append(pm.prevSstableTablesExcisedByLevel, 0)
+			pm.prevSstableBlobBytesReadEstimateByLevel = append(pm.prevSstableBlobBytesReadEstimateByLevel, 0)
+			pm.prevSstableBlobBytesCompactedByLevel = append(pm.prevSstableBlobBytesCompactedByLevel, 0)
+			pm.prevSstableBlobBytesFlushedByLevel = append(pm.prevSstableBlobBytesFlushedByLevel, 0)
+			pm.prevSstableMultiLevelBytesInTopByLevel = append(pm.prevSstableMultiLevelBytesInTopByLevel, 0)
+			pm.prevSstableMultiLevelBytesInByLevel = append(pm.prevSstableMultiLevelBytesInByLevel, 0)
+			pm.prevSstableMultiLevelBytesReadByLevel = append(pm.prevSstableMultiLevelBytesReadByLevel, 0)
+			pm.prevSstableBytesWrittenDataBlocksByLevel = append(pm.prevSstableBytesWrittenDataBlocksByLevel, 0)
+			pm.prevSstableBytesWrittenValueBlocksByLevel = append(pm.prevSstableBytesWrittenValueBlocksByLevel, 0)
+		}
 
 		if pm.sstableCount != nil {
 			pm.sstableCount.Record(ctx, lm.TablesCount, attrs)
@@ -1262,34 +1372,44 @@ func (pm *PebbleMetrics) recordFromPebble(ctx context.Context) {
 			pm.sstableVirtualSize.Record(ctx, uint64ToInt64Clamped(lm.VirtualTablesSize), attrs)
 		}
 		if pm.compactionBytesRead != nil {
-			pm.compactionBytesRead.Add(ctx, uint64ToInt64Clamped(lm.TableBytesIn), attrs)
+			addDelta(ctx, pm.compactionBytesRead, uint64ToInt64Clamped(lm.TableBytesIn),
+				&pm.prevCompactionBytesReadByLevel[level], attrs)
 		}
 		if pm.compactionBytesWritten != nil {
-			pm.compactionBytesWritten.Add(ctx, uint64ToInt64Clamped(lm.TableBytesCompacted), attrs)
+			addDelta(ctx, pm.compactionBytesWritten, uint64ToInt64Clamped(lm.TableBytesCompacted),
+				&pm.prevCompactionBytesWrittenByLevel[level], attrs)
 		}
 		if pm.sstableBytesIngested != nil {
-			pm.sstableBytesIngested.Add(ctx, uint64ToInt64Clamped(lm.TableBytesIngested), attrs)
+			addDelta(ctx, pm.sstableBytesIngested, uint64ToInt64Clamped(lm.TableBytesIngested),
+				&pm.prevSstableBytesIngestedByLevel[level], attrs)
 		}
 		if pm.sstableBytesMoved != nil {
-			pm.sstableBytesMoved.Add(ctx, uint64ToInt64Clamped(lm.TableBytesMoved), attrs)
+			addDelta(ctx, pm.sstableBytesMoved, uint64ToInt64Clamped(lm.TableBytesMoved),
+				&pm.prevSstableBytesMovedByLevel[level], attrs)
 		}
 		if pm.sstableBytesRead != nil {
-			pm.sstableBytesRead.Add(ctx, uint64ToInt64Clamped(lm.TableBytesRead), attrs)
+			addDelta(ctx, pm.sstableBytesRead, uint64ToInt64Clamped(lm.TableBytesRead),
+				&pm.prevSstableBytesReadByLevel[level], attrs)
 		}
 		if pm.sstableBytesFlushed != nil {
-			pm.sstableBytesFlushed.Add(ctx, uint64ToInt64Clamped(lm.TableBytesFlushed), attrs)
+			addDelta(ctx, pm.sstableBytesFlushed, uint64ToInt64Clamped(lm.TableBytesFlushed),
+				&pm.prevSstableBytesFlushedByLevel[level], attrs)
 		}
 		if pm.sstableTablesCompacted != nil {
-			pm.sstableTablesCompacted.Add(ctx, uint64ToInt64Clamped(lm.TablesCompacted), attrs)
+			addDelta(ctx, pm.sstableTablesCompacted, uint64ToInt64Clamped(lm.TablesCompacted),
+				&pm.prevSstableTablesCompactedByLevel[level], attrs)
 		}
 		if pm.sstableTablesFlushed != nil {
-			pm.sstableTablesFlushed.Add(ctx, uint64ToInt64Clamped(lm.TablesFlushed), attrs)
+			addDelta(ctx, pm.sstableTablesFlushed, uint64ToInt64Clamped(lm.TablesFlushed),
+				&pm.prevSstableTablesFlushedByLevel[level], attrs)
 		}
 		if pm.sstableTablesIngested != nil {
-			pm.sstableTablesIngested.Add(ctx, uint64ToInt64Clamped(lm.TablesIngested), attrs)
+			addDelta(ctx, pm.sstableTablesIngested, uint64ToInt64Clamped(lm.TablesIngested),
+				&pm.prevSstableTablesIngestedByLevel[level], attrs)
 		}
 		if pm.sstableTablesMoved != nil {
-			pm.sstableTablesMoved.Add(ctx, uint64ToInt64Clamped(lm.TablesMoved), attrs)
+			addDelta(ctx, pm.sstableTablesMoved, uint64ToInt64Clamped(lm.TablesMoved),
+				&pm.prevSstableTablesMovedByLevel[level], attrs)
 		}
 		if pm.sstableCompensatedFillFactor != nil {
 			pm.sstableCompensatedFillFactor.Record(ctx, lm.CompensatedFillFactor, attrs)
@@ -1298,39 +1418,49 @@ func (pm *PebbleMetrics) recordFromPebble(ctx context.Context) {
 			pm.sstableEstimatedReferencesSize.Record(ctx, uint64ToInt64Clamped(lm.EstimatedReferencesSize), attrs)
 		}
 		if pm.sstableTablesDeleted != nil {
-			pm.sstableTablesDeleted.Add(ctx, uint64ToInt64Clamped(lm.TablesDeleted), attrs)
+			addDelta(ctx, pm.sstableTablesDeleted, uint64ToInt64Clamped(lm.TablesDeleted),
+				&pm.prevSstableTablesDeletedByLevel[level], attrs)
 		}
 		if pm.sstableTablesExcised != nil {
-			pm.sstableTablesExcised.Add(ctx, uint64ToInt64Clamped(lm.TablesExcised), attrs)
+			addDelta(ctx, pm.sstableTablesExcised, uint64ToInt64Clamped(lm.TablesExcised),
+				&pm.prevSstableTablesExcisedByLevel[level], attrs)
 		}
 		if pm.sstableBlobBytesReadEstimate != nil {
-			pm.sstableBlobBytesReadEstimate.Add(ctx, uint64ToInt64Clamped(lm.BlobBytesReadEstimate), attrs)
+			addDelta(ctx, pm.sstableBlobBytesReadEstimate, uint64ToInt64Clamped(lm.BlobBytesReadEstimate),
+				&pm.prevSstableBlobBytesReadEstimateByLevel[level], attrs)
 		}
 		if pm.sstableBlobBytesCompacted != nil {
-			pm.sstableBlobBytesCompacted.Add(ctx, uint64ToInt64Clamped(lm.BlobBytesCompacted), attrs)
+			addDelta(ctx, pm.sstableBlobBytesCompacted, uint64ToInt64Clamped(lm.BlobBytesCompacted),
+				&pm.prevSstableBlobBytesCompactedByLevel[level], attrs)
 		}
 		if pm.sstableBlobBytesFlushed != nil {
-			pm.sstableBlobBytesFlushed.Add(ctx, uint64ToInt64Clamped(lm.BlobBytesFlushed), attrs)
+			addDelta(ctx, pm.sstableBlobBytesFlushed, uint64ToInt64Clamped(lm.BlobBytesFlushed),
+				&pm.prevSstableBlobBytesFlushedByLevel[level], attrs)
 		}
 		if pm.sstableMultiLevelBytesInTop != nil {
-			pm.sstableMultiLevelBytesInTop.Add(ctx, uint64ToInt64Clamped(lm.MultiLevel.TableBytesInTop), attrs)
+			addDelta(ctx, pm.sstableMultiLevelBytesInTop, uint64ToInt64Clamped(lm.MultiLevel.TableBytesInTop),
+				&pm.prevSstableMultiLevelBytesInTopByLevel[level], attrs)
 		}
 		if pm.sstableMultiLevelBytesIn != nil {
-			pm.sstableMultiLevelBytesIn.Add(ctx, uint64ToInt64Clamped(lm.MultiLevel.TableBytesIn), attrs)
+			addDelta(ctx, pm.sstableMultiLevelBytesIn, uint64ToInt64Clamped(lm.MultiLevel.TableBytesIn),
+				&pm.prevSstableMultiLevelBytesInByLevel[level], attrs)
 		}
 		if pm.sstableMultiLevelBytesRead != nil {
-			pm.sstableMultiLevelBytesRead.Add(ctx, uint64ToInt64Clamped(lm.MultiLevel.TableBytesRead), attrs)
+			addDelta(ctx, pm.sstableMultiLevelBytesRead, uint64ToInt64Clamped(lm.MultiLevel.TableBytesRead),
+				&pm.prevSstableMultiLevelBytesReadByLevel[level], attrs)
 		}
 		if pm.sstableValueBlocksSize != nil {
 			pm.sstableValueBlocksSize.Record(ctx, uint64ToInt64Clamped(lm.Additional.ValueBlocksSize), attrs)
 		}
 		if pm.sstableBytesWrittenDataBlocks != nil {
-			pm.sstableBytesWrittenDataBlocks.Add(ctx,
-				uint64ToInt64Clamped(lm.Additional.BytesWrittenDataBlocks), attrs)
+			addDelta(ctx, pm.sstableBytesWrittenDataBlocks,
+				uint64ToInt64Clamped(lm.Additional.BytesWrittenDataBlocks),
+				&pm.prevSstableBytesWrittenDataBlocksByLevel[level], attrs)
 		}
 		if pm.sstableBytesWrittenValueBlocks != nil {
-			pm.sstableBytesWrittenValueBlocks.Add(ctx,
-				uint64ToInt64Clamped(lm.Additional.BytesWrittenValueBlocks), attrs)
+			addDelta(ctx, pm.sstableBytesWrittenValueBlocks,
+				uint64ToInt64Clamped(lm.Additional.BytesWrittenValueBlocks),
+				&pm.prevSstableBytesWrittenValueBlocksByLevel[level], attrs)
 		}
 	}
 
@@ -1365,10 +1495,12 @@ func (pm *PebbleMetrics) recordFromPebble(ctx context.Context) {
 		pm.walPhysicalSize.Record(ctx, uint64ToInt64Clamped(m.WAL.PhysicalSize), metric.WithAttributes(dbAttr))
 	}
 	if pm.walBytesIn != nil {
-		pm.walBytesIn.Add(ctx, uint64ToInt64Clamped(m.WAL.BytesIn), metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.walBytesIn,
+			uint64ToInt64Clamped(m.WAL.BytesIn), &pm.prevWalBytesIn, metric.WithAttributes(dbAttr))
 	}
 	if pm.walBytesWritten != nil {
-		pm.walBytesWritten.Add(ctx, uint64ToInt64Clamped(m.WAL.BytesWritten), metric.WithAttributes(dbAttr))
+		addDelta(ctx, pm.walBytesWritten, uint64ToInt64Clamped(m.WAL.BytesWritten),
+			&pm.prevWalBytesWritten, metric.WithAttributes(dbAttr))
 	}
 
 	if pm.tableObsoleteSize != nil {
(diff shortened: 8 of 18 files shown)


sei-tendermint/internal/autobahn/consensus/persist/blocks.go
```diff
@@ -0,0 +1,192 @@
...
+					Str("lane", lane.String()).
+					Uint64("gapAt", uint64(sorted[i-1]+1)).
+					Int("skipped", len(sorted)-i).
+					Msg("truncating loaded blocks at gap; remaining will be re-fetched")
+				break
+			}
+			contiguous = append(contiguous, LoadedBlock{Number: n, Proposal: bs[n]})
+		}
+		result[lane] = contiguous
+	}
```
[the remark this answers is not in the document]
wen-coding: iterating over all items on startup.

sei-tendermint/internal/autobahn/avail/inner.go
```diff
@@ -16,21 +16,84 @@
...
+			}
+			q.q[q.next] = b
+			q.next++
+		}
+		// Advance the votes queue to match so headers() returns ErrPruned
+		// for already-committed blocks instead of blocking forever.
+		vq := i.votes[lane]
+		vq.first = minN
+		vq.next = minN
+	}
```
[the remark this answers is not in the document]
wen-coding: Iterating over all items on startup.

sei-tendermint/internal/autobahn/avail/inner.go
```diff
@@ -16,21 +17,67 @@
...
+		for _, b := range bs {
+			q.q[q.next] = b.Proposal
+			q.next++
+		}
+		// Advance the votes queue to match so headers() returns ErrPruned
+		// for already-committed blocks instead of blocking forever.
+		vq := i.votes[lane]
+		vq.first = first
+		vq.next = first
+	}
```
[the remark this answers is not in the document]
wen-coding: ditto

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -198,18 +261,48 @@
...
+		s.blockPersist.DeleteBefore(laneFirsts)
+	}
 	return nil
 }
 
+func snapshotLaneFirsts(inner *inner) map[types.LaneID]types.BlockNumber {
+	m := make(map[types.LaneID]types.BlockNumber, len(inner.blocks))
+	for lane, q := range inner.blocks {
+		m[lane] = q.first
+	}
```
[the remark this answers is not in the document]
wen-coding: The order here doesn't matter.

sei-tendermint/internal/autobahn/consensus/persist/blocks.go
```diff
@@ -0,0 +1,256 @@
...
+	blocks, err := bp.loadAll()
+	if err != nil {
+		return nil, nil, err
+	}
+	for lane, bs := range blocks {
+		if len(bs) > 0 {
+			bp.persisted[lane] = bs[len(bs)-1].Number + 1
+		}
+	}
```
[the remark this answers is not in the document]
wen-coding: Iterating over all items on startup.

sei-tendermint/internal/autobahn/consensus/persist/blocks.go
```diff
@@ -0,0 +1,267 @@
...
+		ch:        make(chan persistJob, persistQueueSize),
+		persisted: make(map[types.LaneID]types.BlockNumber),
+	}
+	blocks, err := bp.loadAll()
+	if err != nil {
+		return nil, nil, err
+	}
+	for lane, bs := range blocks {
+		if len(bs) > 0 {
+			bp.persisted[lane] = bs[len(bs)-1].Number + 1
```
pompon0: persisted, or next-to-persist?

sei-tendermint/internal/autobahn/consensus/persist/blocks.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+	n, err := strconv.ParseUint(parts[1], 10, 64)
+	if err != nil {
+		return types.PublicKey{}, 0, fmt.Errorf("bad block number in %q: %w", name, err)
+	}
+	return lane, types.BlockNumber(n), nil
+}
+
+// PersistBlock writes a signed lane proposal to its own file.
+func (bp *BlockPersister) PersistBlock(lane types.LaneID, n types.BlockNumber, proposal *types.Signed[*types.LaneProposal]) error {
```
pompon0: lane and n are contained in LaneProposal

sei-tendermint/internal/autobahn/consensus/persist/blocks.go
```diff
@@ -0,0 +1,267 @@
...
+		}
+	}
+}
+
+// Queue enqueues a block for async persistence. Blocks if the queue is full
+// until space is available or ctx is cancelled. We must not drop blocks because
+// the blockPersisted cursor advances sequentially — a hole would stall voting
+// on the affected lane until restart (which reconstructs the cursor from disk).
+// TODO: add retry on persistence failure to avoid restart-only recovery.
+func (bp *BlockPersister) Queue(ctx context.Context, lane types.LaneID, n types.BlockNumber, proposal *types.Signed[*types.LaneProposal]) error {
```
pompon0: ditto

sei-tendermint/internal/autobahn/consensus/persist/blocks.go
```diff
@@ -0,0 +1,267 @@
...
+		}
+		path := filepath.Join(bp.dir, entry.Name())
+		if err := os.Remove(path); err != nil && !os.IsNotExist(err) {
+			log.Warn().Err(err).Str("path", path).Msg("failed to delete block file")
+		}
+	}
+}
+
+// Queue enqueues a block for async persistence. Blocks if the queue is full
+// until space is available or ctx is cancelled. We must not drop blocks because
```
pompon0: It should be easy to check here that blocks are received in order and return an error if that's not the case (holes are possible, because inner state can skip forward)

sei-tendermint/internal/autobahn/consensus/persist/blocks.go
```diff
@@ -0,0 +1,267 @@
...
+	select {
+	case bp.ch <- persistJob{lane: lane, number: n, proposal: proposal}:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+// Run drains the internal queue, fsyncs each block to disk, and calls
+// onPersisted with the exclusive upper bound of contiguously persisted blocks
```
pompon0: "contiguously" is misleading, because holes are expected. More like highest.

sei-tendermint/internal/autobahn/consensus/persist/blocks.go
```diff
@@ -0,0 +1,267 @@
...
+// for the lane. Blocks until ctx is cancelled.
+func (bp *BlockPersister) Run(ctx context.Context, onPersisted func(types.LaneID, types.BlockNumber)) error {
```
pompon0: nit: how about a Watch/AtomicWatch on the persisted lane tips instead (you can use immutable map for this purpose)? If you don't want to, document how onPersisted will be called sequentially and that it shouldn't block (since it doesn't accept a context).

sei-tendermint/internal/autobahn/consensus/persist/persist.go
```diff
@@ -55,9 +46,10 @@ import (
 	"github.com/sei-protocol/sei-chain/sei-tendermint/libs/utils"
 )
 
+// A/B file suffixes.
 const (
-	suffixA = "_a.pb"
-	suffixB = "_b.pb"
+	SuffixA = "_a.pb"
```
pompon0: the suffixes are leaking again. Please move any logic which needs them to the persist package.

sei-tendermint/internal/autobahn/consensus/inner_test.go
```diff
@@ -920,3 +925,33 @@ func TestPushTimeoutQCClearsStaleState(t *testing.T) {
...
+
+	cs, err := NewState(&Config{
+		Key:         keys[0],
+		ViewTimeout: func(types.View) time.Duration { return time.Hour },
+	}, ds)
+	require.NoError(t, err)
+
+	// Inject a persister that always fails.
+	wantErr := errors.New("disk on fire")
+	cs.persister = utils.Some[persist.Persister](failPersister{err: wantErr})
```
pompon0: not related to this pr: don't substitute the persister on the flight in test - it is fragile to assume that constructor doesn't make any assumptions about this field being constant. Instead, you can replace `Config.PersistentStateDir Optional[string]` with `Config.Persister Optional[Persister]`. If you don't want to make the the persister configurable, then just add a private constructor (like `NewStateWithPersister`) which will be called from `NewState()` and which you will be able to use in tests.

sei-tendermint/internal/autobahn/avail/inner.go
```diff
@@ -14,23 +15,87 @@ type inner struct {
...
+	i := &inner{
 		latestAppQC:    utils.None[*types.AppQC](),
 		latestCommitQC: utils.NewAtomicSend(utils.None[*types.CommitQC]()),
 		appVotes:       newQueue[types.GlobalBlockNumber, appVotes](),
 		commitQCs:      newQueue[types.RoadIndex, *types.CommitQC](),
 		blocks:         blocks,
 		votes:          votes,
 	}
+
+	if loaded == nil {
```
pompon0: use Optional for optional values, not nil

sei-tendermint/internal/autobahn/avail/inner.go
```diff
@@ -14,23 +15,87 @@ type inner struct {
...
+		return i
+	}
+
+	// Restore AppQC and advance queues past already-processed indices.
+	i.latestAppQC = loaded.appQC
+	if aq, ok := loaded.appQC.Get(); ok {
+		// CommitQCs through this index have been processed; skip them.
+		i.commitQCs.first = aq.Proposal().RoadIndex() + 1
```
pompon0: why +1? we need the corresponding commitQC, since they constitute the justification for pruning. Also we need to persist all the commitQCs as well. Prefer to call "prune" instead of manually setting internals of the queue.

pompon0: in fact, it should be possible to initialize commitQC and appVotes queues by just calling inner.prune().

sei-tendermint/internal/autobahn/avail/inner.go
```diff
@@ -14,23 +15,87 @@ type inner struct {
 	commitQCs      *queue[types.RoadIndex, *types.CommitQC]
 	blocks         map[types.LaneID]*queue[types.BlockNumber, *types.Signed[*types.LaneProposal]]
 	votes          map[types.LaneID]*queue[types.BlockNumber, blockVotes]
+	// blockPersisted tracks per-lane how far block persistence has progressed.
+	// RecvBatch only yields blocks below this cursor for voting.
+	// nil when persistence is disabled (testing); RecvBatch then uses q.next.
+	// blockPersisted itself is not persisted to disk: on restart it is
+	// reconstructed from the blocks already on disk (see newInner).
+	blockPersisted map[types.LaneID]types.BlockNumber
```
pompon0: nextToPersist?

sei-tendermint/internal/autobahn/avail/inner.go
```diff
@@ -14,23 +15,87 @@ type inner struct {
...
+	// blockPersisted gates RecvBatch: only blocks below this cursor are
+	// eligible for voting. Lanes without loaded blocks start at 0, which
+	// is safe — an empty queue has nothing to vote on. New blocks must
+	// arrive in order via PushBlock, get persisted, and the callback will
+	// advance blockPersisted accordingly.
+	i.blockPersisted = make(map[types.LaneID]types.BlockNumber, c.Lanes().Len())
+
+	// Restore persisted blocks into their lane queues.
+	for lane, bs := range loaded.blocks {
+		q, ok := i.blocks[lane]
```
pompon0, line 110: ditto - don't mix internals of queue here - use queue.pushBack.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -38,15 +40,68 @@ type State struct {
...
-func NewState(key types.SecretKey, data *data.State) *State {
-	return &State{
-		key:   key,
-		data:  data,
-		inner: utils.NewWatch(newInner(data.Committee())),
+// stateDir is None when persistence is disabled (testing only).
+func NewState(key types.SecretKey, data *data.State, stateDir utils.Option[string]) (*State, error) {
+	var loaded *loadedAvailState
+	var p utils.Option[persist.Persister]
+	var bp *persist.BlockPersister
```
pompon0: Option

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -38,15 +40,68 @@ type State struct {
...
 
 // NewState constructs a new availability state.
-func NewState(key types.SecretKey, data *data.State) *State {
-	return &State{
-		key:   key,
-		data:  data,
-		inner: utils.NewWatch(newInner(data.Committee())),
+// stateDir is None when persistence is disabled (testing only).
+func NewState(key types.SecretKey, data *data.State, stateDir utils.Option[string]) (*State, error) {
+	var loaded *loadedAvailState
```
pompon0: Option

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -38,15 +40,68 @@ type State struct {
...
+	var p utils.Option[persist.Persister]
```
pompon0: nit: it is kind of inconsistent that block persister is strongly typed, and appqc persister is not. Can we make it persister[T proto.Message]?

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -176,9 +232,16 @@ func (s *State) PushAppVote(ctx context.Context, v *types.Signed[*types.AppVote]
...
 		if updated {
+			if err := s.persistAppQC(appQC); err != nil {
+				return err
+			}
+			laneFirsts = snapshotLaneFirsts(inner)
 			ctrl.Updated()
 		}
 	}
+	if s.blockPersist != nil {
+		s.blockPersist.DeleteBefore(laneFirsts)
```
pompon0: nit: for consistency this also should be async, and should terminate Run() in case of problems.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -176,9 +232,16 @@ func (s *State) PushAppVote(ctx context.Context, v *types.Signed[*types.AppVote]
...
 		}
 		if updated {
+			if err := s.persistAppQC(appQC); err != nil {
+				return err
+			}
+			laneFirsts = snapshotLaneFirsts(inner)
 			ctrl.Updated()
 		}
 	}
+	if s.blockPersist != nil {
```
pompon0: use Option of optional fields.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -176,9 +232,16 @@ func (s *State) PushAppVote(ctx context.Context, v *types.Signed[*types.AppVote]
 			return err
 		}
 		if updated {
+			if err := s.persistAppQC(appQC); err != nil {
```
pompon0: it is not the PushAppVote caller's fault that persistance of AppQC failed (i.e. terminating a connection would be misleading). It should be done from another task. It will require reorganizing some code, so you can add a TODO for now.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -200,18 +263,48 @@ func (s *State) PushAppQC(appQC *types.AppQC, commitQC *types.CommitQC) error {
 	if appQC.Proposal().RoadIndex() != commitQC.Proposal().Index() {
 		return fmt.Errorf("mismatched QCs: appQC index %v, commitQC index %v", appQC.Proposal().RoadIndex(), commitQC.Proposal().Index())
 	}
+	var laneFirsts map[types.LaneID]types.BlockNumber
 	for inner, ctrl := range s.inner.Lock() {
 		updated, err := inner.prune(appQC, commitQC)
```
pompon0, line 337: nit: it seems to me that prune should just return the lane snapshot instead of "updated" at this point.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -200,18 +263,48 @@ func (s *State) PushAppQC(appQC *types.AppQC, commitQC *types.CommitQC) error {
...
+	m := make(map[types.LaneID]types.BlockNumber, len(inner.blocks))
+	for lane, q := range inner.blocks {
+		m[lane] = q.first
+	}
+	return m
+}
+
+// persistAppQC writes the AppQC to the A/B file. Called under the inner lock
+// so the checkpoint is durable before ctrl.Updated() notifies watchers.
+func (s *State) persistAppQC(appQC *types.AppQC) error {
```
pompon0: this shouldn't be called while holding a lock on inner - an async pruning task would be more appropriate.

pompon0: in fact you can just wrap latestAppQC into an AtomicSend and prune based on that.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -293,6 +386,11 @@ func (s *State) PushBlock(ctx context.Context, p *types.Signed[*types.LanePropos
 		q.pushBack(p)
 		ctrl.Updated()
 	}
+	if s.blockPersist != nil {
+		// Blocking: called outside the inner lock so only this goroutine stalls.
+		// See Queue() for why we must not drop blocks.
+		return utils.IgnoreAfterCancel(ctx, s.blockPersist.Queue(ctx, h.Lane(), h.BlockNumber(), p))
```
pompon0: imo, this should be nonblocking, and we should just make the persistance queue smarter, for example (not necessarily the best choice, just an idea) you can treat the inner state as a queue itself - it autoprunes itself, and an async persistance task will be just trying to keep up with it.

pompon0: I understand your confusion around cancellation handling, so let me explain the situation here. It is perfectly fine for a function to return "Canceled" in case the context was canceled. In particular "Canceled" is the only error Queue() can return. IgnoreAfterCancel is for situations where you have a function which runs a bunch of tasks in the background that might terminate with an error, even if the function itself succeeded. Even so the properly implemented background task should usually return nil or Canceled upon termination (so IgnoreCanceled should be more adequate). It MIGHT happen though that tasks are not isolated enough and termination of one task will cause error for another (horizontal error propagation), in which case IgnoreAfterCancel is the last resort (which hopefully should only happen in tests).

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -443,18 +543,35 @@ func (s *State) produceBlock(ctx context.Context, key types.SecretKey, payload *
...
+		q.q[q.next] = result
 		q.next += 1
 		ctrl.Updated()
-		return p, nil
 	}
-	panic("unreachable")
+	if s.blockPersist != nil {
+		// Blocking: called outside the inner lock so only this goroutine stalls.
+		// See Queue() for why we must not drop blocks.
+		if err := utils.IgnoreAfterCancel(ctx, s.blockPersist.Queue(ctx, lane, blockNum, result)); err != nil {
```
pompon0: ditto

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -293,6 +386,11 @@ func (s *State) PushBlock(ctx context.Context, p *types.Signed[*types.LanePropos
 		q.pushBack(p)
 		ctrl.Updated()
 	}
+	if s.blockPersist != nil {
+		// Blocking: called outside the inner lock so only this goroutine stalls.
+		// See Queue() for why we must not drop blocks.
```
pompon0: IgnoreAfterCancel will not protect the block from being dropped.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -443,18 +543,35 @@ func (s *State) produceBlock(ctx context.Context, key types.SecretKey, payload *
 		if q.first < q.next {
 			parent = q.q[q.next-1].Msg().Block().Header().Hash()
 		}
-		p := types.Sign(key, types.NewLaneProposal(types.NewBlock(lane, q.next, parent, payload)))
-		q.q[q.next] = p
+		blockNum = q.next
```
pompon0: ok, I see that there is preexisting code which updates queue internals directly. Can you fix that and use pushBack instead? I'm fine with leaving read-only access to internals though (for now).

sei-tendermint/internal/autobahn/avail/inner.go
wen-coding: renamed nextBlockToPersist

sei-tendermint/internal/autobahn/avail/inner.go
wen-coding: done.

sei-tendermint/internal/autobahn/avail/inner.go
wen-coding: done.

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: done.

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: done.

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: done.

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: done.

sei-tendermint/internal/autobahn/consensus/persist/blocks.go
wen-coding: done

sei-tendermint/internal/autobahn/consensus/persist/blocks.go
wen-coding: done

sei-tendermint/internal/autobahn/consensus/persist/blocks.go
wen-coding: done.

sei-tendermint/internal/autobahn/consensus/persist/blocks.go
wen-coding: done

sei-tendermint/internal/autobahn/consensus/persist/blocks.go
wen-coding: done

sei-tendermint/internal/autobahn/consensus/persist/persist.go
wen-coding: done

sei-tendermint/internal/autobahn/consensus/inner_test.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: ah I see, removed

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: removed

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: we don't use queue any more.

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: changed, does this look better?

sei-tendermint/internal/autobahn/avail/inner.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/inner.go
```diff
@@ -14,23 +15,105 @@ type inner struct {
...
+	l, ok := loaded.Get()
+	if !ok {
+		return i, nil
+	}
+
+	// Restore persisted CommitQCs into the queue.
+	if len(l.commitQCs) > 0 {
+		i.commitQCs.reset(l.commitQCs[0].Index)
+		for _, lqc := range l.commitQCs {
+			if lqc.Index == i.commitQCs.next {
```
pompon0: return error in case loaded state is not a list of consecutive commitQCs.

sei-tendermint/internal/autobahn/avail/queue.go
```diff
@@ -16,6 +16,12 @@ func (q *queue[I, T]) Len() uint64 {
 	return uint64(q.next) - uint64(q.first)
 }
 
+// reset sets the starting position of an empty queue.
```
pompon0: nit: you should just recreate the queue if you want to reset the state. Or you can call prune(start) if you want to move 0 -> start.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -38,15 +40,86 @@ type State struct {
 	key   types.SecretKey
 	data  *data.State
 	inner utils.Watch[*inner]
+
+	// persister writes avail inner state to disk using A/B files; None when persistence is disabled.
+	persister utils.Option[persist.Persister[*apb.AppQC]]
```
pompon0: nit: group the persisters under a single Option. It would be kind of weird to allow for an arbitrary subset of persisters.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -38,15 +40,86 @@ type State struct {
...
+	blockPersist utils.Option[*persist.BlockPersister]
+	// commitQCPersist writes/deletes individual CommitQC files; None when persistence is disabled.
+	commitQCPersist utils.Option[*persist.CommitQCPersister]
+}
+
+// innerFile is the A/B file prefix for avail inner state persistence.
+const innerFile = "avail_inner"
+
+// loadPersistedState loads persisted avail state from disk and creates persisters for ongoing writes.
+func loadPersistedState(dir string) (*loadedAvailState, persist.Persister[*apb.AppQC], *persist.BlockPersister, *persist.CommitQCPersister, error) {
```
pompon0: oof, this result type deserves a struct. It will combine nicely with the previous comment.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -171,11 +244,12 @@ func (s *State) PushAppVote(ctx context.Context, v *types.Signed[*types.AppVote]
 		if !ok {
 			return nil
 		}
-		updated, err := inner.prune(appQC, qc)
+		laneFirsts, err := inner.prune(appQC, qc)
```
pompon0: ugh, so the laneFirsts are not always used (wasted computation), and now we check for non-nil map (which are proper empty maps). Perhaps it was not such a good idea :/ . Up to you, if you want to revert it to the previous version.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -443,18 +519,37 @@ func (s *State) produceBlock(ctx context.Context, key types.SecretKey, payload *
...
+						log.Error().Err(err).Msg("failed to persist AppQC")
+					}
+					return nil
+				})
+			})
+		}
+		if bp, ok := s.blockPersist.Get(); ok {
+			cp, _ := s.commitQCPersist.Get()
+			scope.SpawnNamed("persist", func() error {
+				return s.persistLoop(ctx, bp, cp)
```
pompon0: once you get all persisterts into a single struct, it can have its own Run() method to encapsulate these.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -490,3 +585,97 @@ func (s *State) Run(ctx context.Context) error {
...
+	commitQCs     []*types.CommitQC
+	laneFirsts    map[types.LaneID]types.BlockNumber
+	commitQCFirst types.RoadIndex
+	commitQCCur   types.RoadIndex // clamped cursor (may have advanced past pruned entries)
+}
+
+func (s *State) persistLoop(ctx context.Context, bp *persist.BlockPersister, cp *persist.CommitQCPersister) error {
+	blockCur := bp.LoadTips()
+	var commitQCCur types.RoadIndex
+	if cp != nil {
```
pompon0: either assume cp is non-nil, or make it Optional

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -490,3 +585,97 @@ func (s *State) Run(ctx context.Context) error {
...
+		}
+		if len(batch.commitQCs) > 0 {
+			if err := cp.DeleteBefore(batch.commitQCFirst); err != nil {
+				return fmt.Errorf("commitqc deleteBefore: %w", err)
+			}
+		}
+
+		// Update nextBlockToPersist under lock.
+		if len(batch.blocks) > 0 {
+			for inner, ctrl := range s.inner.Lock() {
```
pompon0: nit: I would make it a method of State, MarkAsPersisted, or similar.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -490,3 +585,97 @@ func (s *State) Run(ctx context.Context) error {
...
+) (persistBatch, error) {
+	var b persistBatch
+	for inner, ctrl := range s.inner.Lock() {
+		if err := ctrl.WaitUntil(ctx, func() bool {
+			for lane, q := range inner.blocks {
+				if blockCur[lane] < q.next {
+					return true
+				}
+			}
+			return cp != nil && commitQCCur < inner.commitQCs.next
```
pompon0: ditto

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -490,3 +585,97 @@ func (s *State) Run(ctx context.Context) error {
...
+				ctrl.Updated()
+			}
+		}
+	}
+}
+
+// collectPersistBatch waits for new blocks or commitQCs and collects them under lock.
+func (s *State) collectPersistBatch(
+	ctx context.Context,
+	cp *persist.CommitQCPersister,
```
pompon0: cp is only passed to check whether commitQCs update should also be considered. This should be a given always (see comments above)

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -443,18 +519,37 @@ func (s *State) produceBlock(ctx context.Context, key types.SecretKey, payload *
...
 	return scope.Run(ctx, func(ctx context.Context, scope scope.Scope) error {
+		if p, ok := s.persister.Get(); ok {
+			appQCRecv := s.appQCSend.Subscribe()
+			scope.SpawnNamed("appQCPersist", func() error {
+				return appQCRecv.Iter(ctx, func(_ context.Context, v utils.Option[*types.AppQC]) error {
+					appQC, ok := v.Get()
+					if !ok {
+						return nil
+					}
+					if err := p.Persist(types.AppQCConv.Encode(appQC)); err != nil {
```
pompon0: If I understand correctly, it is possible that appQC persistance will outdistance commitQCs persistance, in which case we will end up with inconsistent state (AppQC for a future CommitQC).

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -490,3 +585,97 @@ func (s *State) Run(ctx context.Context) error {
...
+			if err := cp.DeleteBefore(batch.commitQCFirst); err != nil {
+				return fmt.Errorf("commitqc deleteBefore: %w", err)
+			}
+		}
+
+		// Update nextBlockToPersist under lock.
+		if len(batch.blocks) > 0 {
+			for inner, ctrl := range s.inner.Lock() {
+				for lane, next := range blockCur {
+					inner.nextBlockToPersist[lane] = next
```
pompon0: we need a similar mechanism for commitQCs - in particular consensus.State should not start voting for height H+1 until commitQC for H is persisted. Otherwise we can lose commitQCs.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -490,3 +585,97 @@ func (s *State) Run(ctx context.Context) error {
...
+// persistBatch holds the data collected under lock for one persist iteration.
+type persistBatch struct {
+	blocks        []*types.Signed[*types.LaneProposal]
+	commitQCs     []*types.CommitQC
+	laneFirsts    map[types.LaneID]types.BlockNumber
+	commitQCFirst types.RoadIndex
+	commitQCCur   types.RoadIndex // clamped cursor (may have advanced past pruned entries)
+}
+
+func (s *State) persistLoop(ctx context.Context, bp *persist.BlockPersister, cp *persist.CommitQCPersister) error {
```
pompon0: this loop looks very similar to lanevote subscription (which is intentional). Perhaps we could deduplicate this logic in an elegant way? (this is not a request for change, it is just a thought).

sei-tendermint/internal/autobahn/avail/queue.go
wen-coding: Yeah I originally did prune(start), but I found the code hard to understand, why do I need pruning when the queue is empty?
I can also do newQueueFrom(start I), that's basically reset(), would you prefer that?

sei-tendermint/internal/autobahn/avail/inner.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: no problem, reverted

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: did that, but now we rely on methods of State to collectPersistBatch and markBlocksPersisted, seems cleaner to actually put them in State Run(), so I reverted the change

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: removed

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: gone

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: gone

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: It's true. Changed it so that we write AppQC in the same loop as commitQC, in the future they can be written to the same WAL file.

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: It is very similar, but they are handling quite different data, so not totally sure generic abstraction is worth it here. We can maybe rethink about it after the WAL change.

pompon0: ok

sei-tendermint/internal/autobahn/avail/queue.go
pompon0: I'm fine with reset, but it should also reset q then

sei-tendermint/internal/autobahn/avail/state.go
pompon0: still returns 5 values though

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -134,7 +202,12 @@ func (s *State) PushCommitQC(ctx context.Context, qc *types.CommitQC) error {
 			return nil
 		}
 		inner.commitQCs.pushBack(qc)
-		inner.latestCommitQC.Store(utils.Some(qc))
+		// When persistence is disabled, publish immediately.
+		// When enabled, the persist goroutine publishes after writing to disk,
+		// so consensus won't advance until the CommitQC is durable.
+		if inner.nextBlockToPersist == nil {
```
pompon0: don't compare to nil, either make nextBlockToPersist into an Option, or check persisters field presence instead.

pompon0: alternatively, you can move branching to the persisters logic - i.e. have a dummy persister task which will just bump nextBlockToPersist and latestCommitQC without persisting anything - it would make the control flow in tests and in prod more similar imo.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -443,18 +517,71 @@ func (s *State) produceBlock(ctx context.Context, key types.SecretKey, payload *
...
 	}
-	panic("unreachable")
+	return result, nil
 }
 
 // Run runs the background tasks of the state.
 func (s *State) Run(ctx context.Context) error {
 	return scope.Run(ctx, func(ctx context.Context, scope scope.Scope) error {
+		if pers, ok := s.persisters.Get(); ok {
+			scope.SpawnNamed("persist", func() error {
```
pompon0: nit: this is a rather large task, I feel like it deserves having its own method.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -443,18 +517,71 @@ func (s *State) produceBlock(ctx context.Context, key types.SecretKey, payload *
...
+						return err
+					}
+
+					for _, qc := range batch.commitQCs {
+						if err := pers.commitQCs.PersistCommitQC(qc); err != nil {
+							return fmt.Errorf("persist commitqc %d: %w", qc.Index(), err)
+						}
+						commitQCCur = qc.Index() + 1
+					}
+					if len(batch.commitQCs) > 0 {
```
pompon0: we have now similar problem here - you delete commitQCs before bumping AppQC, therefore we can end up in a state where persisted AppQC does not have a corresponding AppQC.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -443,18 +517,71 @@ func (s *State) produceBlock(ctx context.Context, key types.SecretKey, payload *
...
 // Run runs the background tasks of the state.
 func (s *State) Run(ctx context.Context) error {
 	return scope.Run(ctx, func(ctx context.Context, scope scope.Scope) error {
+		if pers, ok := s.persisters.Get(); ok {
+			scope.SpawnNamed("persist", func() error {
+				blockCur := pers.blocks.LoadTips()
+				commitQCCur := pers.commitQCs.LoadNext()
+				var lastPersistedAppQC utils.Option[*types.AppQC]
+				for {
+					batch, err := s.collectPersistBatch(ctx, blockCur, commitQCCur)
```
pompon0: Couple of things I feel that will be important in the final version, consider adding a todo:
* I suspect that  having block persistence block commitQC persistence will degrade consensus latency significantly (imo they should be fsynced independently)
* let's consider a situation in which our nodes' disk is (just slightly) slower than other nodes' disks. In such a situation our batches will grow (network will move slightly faster). With larger batches, PersistBatch will slow down even more and making us fall behind even more, making the batches larger, etc (i.e. our node will choke on persistence). What I mean is that the larger the batch, the larger the avg latency from receiving a proposal to voting for availability. It would be nice to make this latency independent from the amount of blocks to be persisted at any given time. I'm afraid that if for some reason majority of nodes will end up with large batches just once (let's say because of a network hiccup), they will never recover from having large batches (batched blocks -> batched votes -> batched tipcut -> batch of new capacity -> and so on). I'm not sure if it is clear what I'm trying to say here, we can discuss offline.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -490,3 +617,66 @@ func (s *State) Run(ctx context.Context) error {
...
+			inner.latestCommitQC.Store(utils.Some(qc))
+		}
+		ctrl.Updated()
+	}
+}
+
+// collectPersistBatch waits for new blocks or commitQCs and collects them under lock.
+func (s *State) collectPersistBatch(
+	ctx context.Context,
+	blockCur map[types.LaneID]types.BlockNumber,
```
pompon0: nit: I feel like passing blockCur and commitQCCur is redundant here - you have lastQC and nextBlockToPersist in the inner state already.

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: Good catch, changed it so that we persist AppQC before delete, hopefully in WAL we don't need delete any more.

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: Added dummy persisters.

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: Yeah I understand your concern. Several observations:
1. We do eventually need to keep up with the incoming blocks, if any of the pipeline can't keep up, then we are in trouble and can't support this flow, this may be persistence right now, but the same argument can be applied to execution and storage and other components. I think we do need real benchmarking to see where we are
2. Currently we don't set a limit on number of blocks we persist, if turns out persistence is the bottleneck after benchmarking, then we can be smarter in handling batches. Maybe instead of trying to save everything, we can sometimes wait a bit so we don't have to persist blocks which are quickly pruned

Adding a TODO for the concern now, this is definitely valid concern but probably too big to address in this PR.

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: Yeah I think we don't have to pass them in here, changed.

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: Err, I thought I pushed but apparently lost it, really pushed this time

sei-tendermint/internal/autobahn/avail/inner.go
```diff
@@ -14,23 +15,101 @@ type inner struct {
...
+
+	// Restore persisted blocks into their lane queues.
+	for lane, bs := range l.blocks {
+		q, ok := i.blocks[lane]
+		if !ok || len(bs) == 0 {
+			continue
+		}
+		first := bs[0].Number
+		q.reset(first)
+		for _, b := range bs {
```
pompon0: nit: for consistency with populating CommitQCs, you might want to check the block numbers here as well. Maybe also put parent hash checking here?

sei-tendermint/internal/autobahn/avail/queue.go
pompon0: please address the comment above

sei-tendermint/internal/autobahn/avail/queue_test.go
```diff
@@ -0,0 +1,87 @@
+package avail
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
```
pompon0: libs/utils/require

sei-tendermint/internal/autobahn/avail/queue_test.go
```diff
@@ -0,0 +1,87 @@
...
+	require.Equal(t, uint64(0), q.first)
+	require.Equal(t, uint64(3), q.next)
+	require.Equal(t, uint64(3), q.Len())
+	require.Equal(t, "a", q.q[0])
+	require.Equal(t, "b", q.q[1])
+	require.Equal(t, "c", q.q[2])
+}
+
+func TestQueueReset(t *testing.T) {
+	q := newQueue[uint64, string]()
```
pompon0: add a test for reset() after populating the queue.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -38,15 +40,96 @@ type State struct {
...
+		blocks:    persist.NewNoOpBlockPersister(),
+		commitQCs: persist.NewNoOpCommitQCPersister(),
+	}
+}
+
+// innerFile is the A/B file prefix for avail inner state persistence.
+const innerFile = "avail_inner"
+
+// loadPersistedState loads persisted avail state from disk and creates persisters for ongoing writes.
+func loadPersistedState(dir string) (*loadedAvailState, persisters, error) {
```
pompon0: nit: for consistency, you might want to accept dir Option[string] and pass it to each persister constructor, which will act as a noop in case no dir is provided (given that you have added a noop mode to each of them anyway).

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -490,3 +577,129 @@ func (s *State) Run(ctx context.Context) error {
...
+		}
+
+		for _, proposal := range batch.blocks {
+			h := proposal.Msg().Block().Header()
+			if err := pers.blocks.PersistBlock(proposal); err != nil {
+				return fmt.Errorf("persist block %s/%d: %w", h.Lane(), h.BlockNumber(), err)
+			}
+			s.markBlockPersisted(h.Lane(), h.BlockNumber()+1)
+		}
+		if err := pers.blocks.DeleteBefore(batch.laneFirsts); err != nil {
```
pompon0: Deleting blocks before persisting AppQC may lead to weird avail State, where we have a gap between the AppQC and the first persisted block of the lane. I don't see any obvious reason why it would be a bug in consensus protocol, but it is a nice invariant to have.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -490,3 +577,129 @@ func (s *State) Run(ctx context.Context) error {
...
+				}
+			}
+			lastPersistedAppQC = batch.appQC
+		}
+
+		if len(batch.commitQCs) > 0 {
+			if err := pers.commitQCs.DeleteBefore(batch.commitQCFirst); err != nil {
+				return fmt.Errorf("commitqc deleteBefore: %w", err)
+			}
+			s.markCommitQCsPersisted(commitQCCur, utils.Some(batch.commitQCs[len(batch.commitQCs)-1]))
```
pompon0: nit: that can happen just after persisting commitQCs. CommitQCCur seems redundant - NextOpt(persistedCommitQC) should do.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -490,3 +577,129 @@ func (s *State) Run(ctx context.Context) error {
...
+}
+
+// persistBatch holds the data collected under lock for one persist iteration.
+type persistBatch struct {
+	blocks        []*types.Signed[*types.LaneProposal]
+	commitQCs     []*types.CommitQC
+	appQC         utils.Option[*types.AppQC]
+	laneFirsts    map[types.LaneID]types.BlockNumber
+	commitQCFirst types.RoadIndex
+	commitQCCur   types.RoadIndex // snapshot of nextCommitQCToPersist (clamped)
```
pompon0: nit: commitQCNext for consistency?

sei-tendermint/internal/autobahn/avail/subscriptions.go
```diff
@@ -48,10 +48,11 @@ func (r *LaneVotesRecv) RecvBatch(ctx context.Context) ([]*types.Signed[*types.L
 	for inner, ctrl := range r.state.inner.Lock() {
 		for {
 			for lane, bq := range inner.blocks {
-				for i := max(bq.first, r.next[lane]); i < bq.next; i++ {
+				upperBound := min(bq.next, inner.nextBlockToPersist[lane])
```
pompon0, line 51: nit: add a TODO, that nextBlockToPersist might deserve a separate Watch to avoid waking up too often too many tasks (a potential optimization).

sei-tendermint/internal/autobahn/avail/state.go
pompon0: nit: the caveat with how noop mode currently works is that you enable noop mode for each persister separately. Which is fine as long as fields of persisters type are private, and therefore it is not possible to set a combination of noop and real persisters (which would result in an invalid avail State on restart). Just a FYI

sei-tendermint/internal/autobahn/consensus/persist/blocks.go
```diff
@@ -0,0 +1,263 @@
...
+			log.Warn().Err(err).Str("path", path).Msg("failed to delete block file")
+		}
+	}
+	return nil
+}
+
+// PersistBatch persists a batch of blocks to disk, updates tips after each
+// successful write, and cleans up old files below laneFirsts.
+// Returns the updated tips snapshot.
+func (bp *BlockPersister) PersistBatch(
```
pompon0: unused?

sei-tendermint/internal/autobahn/consensus/persist/blocks.go
```diff
@@ -0,0 +1,263 @@
...
+		raw[lane][n] = proposal
+		log.Info().Str("lane", lane.String()).Uint64("block", uint64(n)).Msg("loaded persisted block")
+	}
+
+	result := map[types.LaneID][]LoadedBlock{}
+	for lane, bs := range raw {
+		sorted := slices.Sorted(maps.Keys(bs))
+		var contiguous []LoadedBlock
+		for i, n := range sorted {
+			if i > 0 && n != sorted[i-1]+1 {
```
pompon0: ~~I think here we should load the LAST contiguous range, no?~~ Actually, this is more complicated. We need pruning to be consistent across the CommitQCs, blocks and AppQC. In particular which range is the one to be loaded depends on the AppQC. Also the blocks which do not belong to the relevant range should be pruned, so that they are not loaded later by accident - let's say that we have a gap, and but during loading we figured that we did not persist the AppQC for the last range, so we loaded the earlier range. Then after a while, we persisted the blocks from the gap and restarted again. Suddenly there is no gap any more, however the old blocks after a gap do not match the blocks before the gap.

sei-tendermint/internal/autobahn/consensus/persist/blocks_test.go
```diff
@@ -0,0 +1,309 @@
...
+	corruptName := blockFilename(lane, 1)
+	require.NoError(t, os.WriteFile(filepath.Join(dir, "blocks", corruptName), []byte("corrupt"), 0600))
+
+	_, blocks, err := NewBlockPersister(dir)
+	require.NoError(t, err)
+	require.Equal(t, 1, len(blocks[lane]), "should only load the valid block")
+	require.NoError(t, utils.TestDiff(b0, blocks[lane][0].Proposal))
+}
+
+func TestLoadCorruptMidSequenceTruncatesAtGap(t *testing.T) {
```
pompon0: if it is not too complicated, it would be nice to have a test which makes avail.State to generate a gap (i.e. persist some stuff, then suddenly receive future data, then persist it) and try to restart the state from it, ensuring that all the latest data is actually loaded back.

sei-tendermint/internal/autobahn/consensus/persist/commitqcs.go
```diff
@@ -0,0 +1,192 @@
...
+	n, err := strconv.ParseUint(name, 10, 64)
+	if err != nil {
+		return 0, fmt.Errorf("bad commitqc filename %q: %w", name, err)
+	}
+	return types.RoadIndex(n), nil
+}
+
+// PersistCommitQC writes a CommitQC to its own file.
+func (cp *CommitQCPersister) PersistCommitQC(qc *types.CommitQC) error {
+	idx := qc.Index()
```
pompon0, line 103: nit: you might want to add a check that qc.Index() >= cp.next and return an error otherwise.

sei-tendermint/internal/autobahn/consensus/persist/commitqcs.go
```diff
@@ -0,0 +1,192 @@
...
+
+	if len(raw) == 0 {
+		return nil, nil
+	}
+
+	sorted := slices.Sorted(maps.Keys(raw))
+
+	var contiguous []LoadedCommitQC
+	for i, idx := range sorted {
+		if i > 0 && idx != sorted[i-1]+1 {
```
pompon0: ~~ditto, we need the LAST contiguous range.~~

sei-tendermint/internal/autobahn/consensus/persist/blocks.go
pompon0: also, since persisting may fail before AppQC is persisted, we might end up with a block range which exceeds the lane capacity - this also needs to be taken care of during loading.

sei-tendermint/internal/autobahn/avail/inner.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/queue_test.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/queue_test.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: moved all deletion to after AppQC is persisted

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: done

sei-tendermint/internal/autobahn/consensus/persist/blocks.go
wen-coding: removed

sei-tendermint/internal/autobahn/consensus/persist/commitqcs.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/subscriptions.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: done

sei-tendermint/internal/autobahn/consensus/persist/blocks.go
wen-coding: That's a good point. I think we can use persisted AppQC as the anchor, because now we persist AppQC first before removing old ones, we just load the contiguous blocks/commitQCs starting from the AppQC we know and ignore everything else, don't crash if there is a gap.

Also added check for block range which exceeds the lane capacity.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -38,15 +40,91 @@ type State struct {
...
+		return utils.None[*loadedAvailState](), persisters{}, fmt.Errorf("NewCommitQCPersister: %w", err)
+	}
+
+	pers := persisters{appQC: appQCPersister, blocks: bp, commitQCs: cp}
+
+	if _, ok := dir.Get(); !ok {
+		return utils.None[*loadedAvailState](), pers, nil
+	}
+
+	var appQC utils.Option[*types.AppQC]
```
pompon0: nit: you might want to filter out blocks and commitQCs already here.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -38,15 +40,91 @@ type State struct {
...
 	}
+
+	inner, err := newInner(data.Committee(), loaded)
+	if err != nil {
+		return nil, err
+	}
+
+	// newInner may have filtered the loaded CommitQCs (e.g., skipping stale
+	// orphans from a crashed fast-forward). Reset the persister's cursor to
+	// match the actual loaded range so it doesn't reject valid new QCs.
```
pompon0: nit: how about strictly pruning the files which were filtered out, for both commitQCs and blocks?

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -490,3 +572,130 @@ func (s *State) Run(ctx context.Context) error {
...
+	var lastPersistedAppQC utils.Option[*types.AppQC]
+	for {
+		batch, err := s.collectPersistBatch(ctx)
+		if err != nil {
+			return err
+		}
+
+		for _, proposal := range batch.blocks {
+			h := proposal.Msg().Block().Header()
+			if err := pers.blocks.PersistBlock(proposal); err != nil {
```
pompon0: this now conflicts with the capacity bound on the lane blocks :/. Consider a scenario:
* new AppQC Y arrives
* we have more capacity for new blocks of a lane X
* a new block B arrives to lane X
* we persist B here and mark as persisted
* a LaneVote is sent for this block
* node is restarted
* B is filtered out because it exceeded capacity, since we haven't persisted Y yet.

pompon0: We indeed want to mark blocks as persisted ASAP, but just modifying avail.State to admit more blocks than capacity on restart won't work, because we might have a gap in blocks.

I suggest the following order of persisting stuff:
* First we persist AppQC and the corresponding CommitQC together - this is a complete anchor.
* Once it is persisted, we do the following in parallel (or arbitrary order for now):
  * persist the commitQCs in order, then we update the LastCommitQC (or after every commitQC, I'm not sure it matters)
  * for every lane, persist lane blocks in order (and mark as persisted after every block).
  * prune old data in whatever order.

sei-tendermint/internal/autobahn/avail/queue.go
wen-coding: Oops, somehow missed this one, done.

sei-tendermint/internal/autobahn/consensus/persist/blocks_test.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: Changed and we now gate pushing new block on persisted AppQC/CommitQC pair.

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -490,3 +620,163 @@ func (s *State) Run(ctx context.Context) error {
...
+
+		// 3. Prune old data.
+		if err := pers.blocks.DeleteBefore(batch.laneFirsts); err != nil {
+			return fmt.Errorf("block deleteBefore: %w", err)
+		}
+		if len(batch.commitQCs) > 0 {
+			if err := pers.commitQCs.DeleteBefore(batch.commitQCFirst); err != nil {
+				return fmt.Errorf("commitqc deleteBefore: %w", err)
+			}
+			s.markCommitQCsPersisted(batch.commitQCs[len(batch.commitQCs)-1])
```
pompon0: markCommitQCsPersisted can be called directly after PersistCommitQC loop.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -38,15 +40,135 @@ type State struct {
...
+type persisters struct {
+	pruneAnchor persist.Persister[*pb.PersistedAvailPruneAnchor]
+	blocks      *persist.BlockPersister
+	commitQCs   *persist.CommitQCPersister
+}
+
+// innerFile is the A/B file prefix for avail inner state persistence.
+const innerFile = "avail_inner"
+
+func decodePruneAnchor(a *pb.PersistedAvailPruneAnchor) (*types.AppQC, *types.CommitQC, error) {
```
pompon0: nit: you might want to add Anchor type and define a full proto.Conv for it.

sei-tendermint/internal/autobahn/avail/inner.go
```diff
@@ -14,23 +17,149 @@ type inner struct {
...
+		nextBlockToPersist:  make(map[types.LaneID]types.BlockNumber, c.Lanes().Len()),
+		persistedBlockStart: make(map[types.LaneID]types.BlockNumber, c.Lanes().Len()),
+	}
+
+	l, ok := loaded.Get()
+	if !ok {
+		return i, nil
+	}
+
+	if l.pruneAppQC != nil {
```
pompon0: Option[Anchor]

sei-tendermint/internal/autobahn/avail/inner.go
```diff
@@ -14,23 +17,149 @@ type inner struct {
...
+	}
+
+	// Restore persisted CommitQCs into the queue. Stale entries below the
+	// prune anchor have already been filtered by loadPersistedState.
+	if len(l.commitQCs) > 0 {
+		i.commitQCs.reset(l.commitQCs[0].Index)
+		i.commitQCs.pushBack(l.commitQCs[0].QC)
+		for _, lqc := range l.commitQCs[1:] {
+			if lqc.Index != i.commitQCs.next {
+				log.Warn().
```
pompon0: Given that we persist anchor first now, this should never happen.

sei-tendermint/internal/autobahn/avail/inner.go
```diff
@@ -14,23 +17,149 @@ type inner struct {
...
+		q, ok := i.blocks[lane]
+		if !ok || len(bs) == 0 {
+			continue
+		}
+		first := bs[0].Number
+		q.reset(first)
+		var lastHash types.BlockHeaderHash
+		for j, b := range bs {
+			if q.Len() >= BlocksPerLane {
+				log.Warn().
```
pompon0: this should never happen

sei-tendermint/internal/autobahn/avail/inner.go
```diff
@@ -14,23 +17,149 @@ type inner struct {
...
+					Stringer("lane", lane).
+					Uint64("loaded", q.Len()).
+					Msg("capping loaded blocks at lane capacity")
+				break
+			}
+			if b.Number != q.next {
```
pompon0, line 119: should never happen

sei-tendermint/internal/autobahn/avail/inner.go
```diff
@@ -14,23 +17,149 @@ type inner struct {
...
+				log.Warn().
+					Stringer("lane", lane).
+					Uint64("expected", uint64(q.next)).
+					Uint64("got", uint64(b.Number)).
+					Msg("skipping non-contiguous persisted blocks (orphans will be cleaned up)")
+				break
+			}
+			if j > 0 {
```
pompon0, line 122: ditto

sei-tendermint/internal/autobahn/avail/inner.go
pompon0: i.e. return error instead of logging

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -490,3 +620,163 @@ func (s *State) Run(ctx context.Context) error {
...
+		if err != nil {
+			return err
+		}
+
+		// 1. Persist prune anchor first — establishes the crash-recovery watermark.
+		if anchor, ok := batch.pruneAnchor.Get(); ok {
+			if err := pers.pruneAnchor.Persist(anchor); err != nil {
+				return fmt.Errorf("persist prune anchor: %w", err)
+			}
+			commitQC, err := types.CommitQCConv.Decode(anchor.CommitQc)
```
pompon0: you shouldn't need to re-decode it. It shouldn't be encoded within the batch.

sei-tendermint/internal/autobahn/avail/inner.go
```diff
@@ -14,23 +17,149 @@ type inner struct {
...
+	// goroutine bumps it immediately). Not persisted to disk: on restart it is
+	// reconstructed from the blocks already on disk (see newInner).
+	//
+	// TODO: consider giving this its own AtomicSend to avoid waking unrelated
+	// inner waiters (PushVote, PushCommitQC, etc.) on every markBlockPersisted
+	// call. Only RecvBatch needs to be notified of cursor changes;
+	// collectPersistBatch is in the same goroutine and reads it directly.
+	nextBlockToPersist map[types.LaneID]types.BlockNumber
+
+	// persistedBlockStart is the per-lane block number derived from the last
```
pompon0, line 31: nit: this is fine (as we discussed yesterday), but without persistedBlockStart it would work as well, which now I see that it would be strictly less logic. This is just observation, no changes requested.

sei-tendermint/internal/autobahn/avail/inner.go
```diff
@@ -14,23 +17,149 @@ type inner struct {
...
+			Msg("loaded persisted prune anchor")
+		for lane := range i.blocks {
+			i.persistedBlockStart[lane] = l.pruneCommitQC.LaneRange(lane).First()
+		}
+	}
+
+	// Restore persisted CommitQCs into the queue. Stale entries below the
+	// prune anchor have already been filtered by loadPersistedState.
+	if len(l.commitQCs) > 0 {
+		i.commitQCs.reset(l.commitQCs[0].Index)
```
pompon0: nit: single i.prune(anchor) call will move all the lane/vote and commitqc queues to the right position (no resets needed) and imo it would clarify the logic here.

sei-tendermint/internal/autobahn/avail/state_test.go
```diff
@@ -193,14 +213,127 @@ func TestState(t *testing.T) {
...
+func TestStateRestartFromPersisted(t *testing.T) {
+	rng := utils.TestRng()
+	committee, keys := types.GenCommittee(rng, 3)
+	dir := t.TempDir()
+
+	// Phase 1: Run state with persistence through 2 iterations.
+	var wantAppQCIdx types.RoadIndex
+	var wantNextBlocks map[types.LaneID]types.BlockNumber
+
+	require.NoError(t, scope.Run(context.Background(), func(ctx context.Context, s scope.Scope) error {
```
pompon0: t.Context()

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/inner.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/inner.go
wen-coding: done

sei-tendermint/internal/autobahn/avail/inner.go
wen-coding: changed

sei-tendermint/internal/autobahn/avail/inner.go
wen-coding: changed

sei-tendermint/internal/autobahn/avail/inner.go
wen-coding: changed

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: Changed

sei-tendermint/internal/autobahn/avail/inner.go
wen-coding: changed

sei-tendermint/internal/autobahn/avail/state_test.go
wen-coding: done

PR #2903, opened 2026-02-17
fix: config issue since v6.3.0 upgrade

mojtaba-esk:
Closes PLT-131

There was also an issue with config files in the `init` cmd, that is resolved here.

### DIFF
--- a/scripts/benchmark.sh
+++ b/scripts/benchmark.sh
@@ -77,7 +77,7 @@ else
   make install-bench
 fi
 # initialize chain with chain ID and add the first key
-~/go/bin/seid init demo --chain-id sei-chain
+~/go/bin/seid init demo --chain-id sei-chain --overwrite
 ~/go/bin/seid keys add $keyname --keyring-backend test
 # add the key as a genesis account with massive balances of several different tokens
 ~/go/bin/seid add-genesis-account $(~/go/bin/seid keys show $keyname -a --keyring-backend test) 100000000000000000000usei,100000000000000000000uusdc,100000000000000000000uatom --keyring-backend test


cmd/seid/cmd/root.go
```diff
@@ -95,6 +96,12 @@ func NewRootCmd() (*cobra.Command, params.EncodingConfig) {
 				return err
 			}
 
+			// Skip creating config.toml/app.toml when running "init"; init creates them itself.
```
mojtaba-esk, line 99: The `init` issue  with configs is here.

docs/migration/seiv2_config_migration.md
```diff
@@ -557,8 +557,8 @@ statesync-no-peers-available-window-seconds = 0
 # Set to 0 to disable
 blocks-behind-threshold = 0
 
-# How often to check if node is behind
-blocks-behind-check-interval = 60
+# How often to check if node is behind (in seconds)
+blocks-behind-check-interval-seconds = 60
```
masih: The usage of `blocks-behind-check-interval` goes right back to 3 years ago when self remediation was added. 

Considering we were using the wrong config consistently everywhere, I don't think this is a refactor worth doing. Because: we con't have a good way to evolve the config (yet), and it would render all existing config out there ineffective. See: https://github.com/sei-protocol/sei-chain/pull/2883

WDYT? does that make sense?

app/params/config.go
```diff
@@ -107,6 +107,8 @@ func setValidatorTypeTendermintConfig(config *tmcfg.Config) {
 // setFullnodeTypeTendermintConfig sets common Tendermint config for fullnode-like nodes
 func setFullnodeTypeTendermintConfig(config *tmcfg.Config) {
 	config.TxIndex.Indexer = []string{"kv"} // Full nodes need tx indexing for queries
+	config.RPC.ListenAddress = "tcp://0.0.0.0:26657"
+	config.P2P.ListenAddress = "tcp://0.0.0.0:26656"
```
masih, line 111: What's the default bind address? Checking that this does not override it (e.g. binding to local host vs all interfaces)

docs/migration/seiv2_config_migration.md
mojtaba-esk: oh I see. will revert this change to avoid breaking it.

app/params/config.go
mojtaba-esk: The default bind address is `127.0.0.1` and this function is called once the mode is `full` or `archive`. However, there is a caveat which is the default mode is also `full` which means if the user explicitly does not pass the `--mode` flag set to `validator` or `seed`, the address will be `0.0.0.0` and they have to later update it.
WTDY we should do? update the docs for validators? or explore other options?

yzang2019: That's a good question, is there a way to dynamically determine this by the actual mode config?

masih: @mojtaba-esk could you help me wrap my head around why the values are being explicitly set here?

mojtaba-esk: @masih The values are set in the issue itself: https://linear.app/seilabs/issue/PLT-131/fix-config-issue-since-v630-upgrade so we can either use some constants here. We could probably find the ports from some already defined constants somewhere in the codebase.

@yzang2019 that would be ideal. We could change the default mode to something else than `full`.
Currently in this PR only full and archive nodes get this config.

My thought is to update docs as it is more frequent that users intiate a full node for RPC and Validators are expected to read the docs although this is not the case usually :D

PR #2908, opened 2026-02-17
fix(consensus): harden FullProposal.Verify and NewProposal against malicious proposals

wen-coding:
## Summary

- Verify proposer signature in `FullProposal.Verify` (was only checking key identity, not the actual sig)
- Verify LaneQC header hash matches the proposal's `LaneRange.LastHash`
- Reject proposals containing non-committee lanes via new `Proposal.Verify` method
- Add leader validation to `NewProposal` constructor (fail-fast on wrong key)
- Use `utils.OrPanic1` for compact error handling in tests
- Comprehensive test coverage for all verification paths

## Test plan

- [x] All `TestProposalVerify*` tests pass
- [x] Full `autobahn/...` test suite passes (avail, consensus, data, types)


sei-tendermint/internal/autobahn/types/proposal.go
```diff
@@ -304,6 +308,10 @@ func (m *FullProposal) Verify(c *Committee, vs ViewSpec) error {
 				return nil
 			}
 		}
+		// Verify the proposal contains exactly the committee lanes (no extra non-committee lanes).
```
pompon0: I've read this code again. It won't work because the LaneRange() returns [0,0) if the lane range is missing in the proposal.

sei-tendermint/internal/autobahn/types/proposal_test.go
```diff
@@ -0,0 +1,662 @@
...
+
+func TestVerifyFreshProposalWithBlocks(t *testing.T) {
+	rng := utils.TestRng()
+	committee, keys := GenCommittee(rng, 4)
+	vs := ViewSpec{}
+	proposerKey := leaderKey(committee, keys, vs.View())
+
+	// Produce a LaneQC for the proposer's lane.
+	lane := proposerKey.Public()
+	laneQC, _ := makeLaneQC(rng, committee, keys, lane, 0, BlockHeaderHash{})
```
pompon0: nit: don't use the zero hash value, just generate a proper block instead - unfortunately golang does not support types without zero values.

sei-tendermint/internal/autobahn/types/proposal_test.go
```diff
@@ -0,0 +1,662 @@
...
+func TestVerifyRejectsViewMismatch(t *testing.T) {
+	rng := utils.TestRng()
+	committee, keys := GenCommittee(rng, 4)
+
+	// Create a proposal at view (5, 0) but verify against genesis ViewSpec (0, 0).
+	vs := ViewSpec{}
+	otherView := View{Index: 5, Number: 0}
+	otherKey := leaderKey(committee, keys, otherView)
+
+	otherProposal := newProposal(otherView, time.Now(), nil, utils.None[*AppProposal]())
```
pompon0: use NewProposal which is the SoT for constructing FullProposals.

sei-tendermint/internal/autobahn/types/proposal_test.go
```diff
@@ -0,0 +1,662 @@
...
+	fp := &FullProposal{
+		proposal: Sign(otherKey, otherProposal),
+	}
+	err := fp.Verify(committee, vs)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "view")
```
pompon0: don't do stringly typed assertions. If you want to distinguish errors, define a custom error and use errors.As/Is.

sei-tendermint/internal/autobahn/types/proposal_test.go
```diff
@@ -0,0 +1,662 @@
...
+	differentProposal := newProposal(vs.View(), time.Now().Add(time.Hour), nil, utils.None[*AppProposal]())
+	forgedFP := &FullProposal{
+		proposal: &Signed[*Proposal]{
+			hashed: NewHashed(differentProposal),
+			sig:    fp.proposal.sig, // reuse leader's sig from the original proposal
+		},
+	}
+	err = forgedFP.Verify(committee, vs)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "proposal signature")
```
pompon0: ditto

sei-tendermint/internal/autobahn/types/proposal_test.go
```diff
@@ -0,0 +1,662 @@
...
+	committee, keys := GenCommittee(rng, 4)
+	vs := ViewSpec{}
+	proposerKey := leaderKey(committee, keys, vs.View())
+
+	// Build a valid proposal.
+	fp, err := NewProposal(proposerKey, committee, vs, time.Now(), nil, utils.None[*AppQC]())
+	require.NoError(t, err)
+
+	// Tamper: build a different proposal and attach the original signature.
+	// The claimed key is still the leader's, but the signature doesn't match.
```
pompon0: nit: prefer malforming a correctly constructed proposal - i.e. use NewProposal and then substitute the signature.

sei-tendermint/internal/autobahn/types/proposal_test.go
```diff
@@ -0,0 +1,662 @@
...
+
+	// Re-sign the same proposal with a different (non-leader) key.
+	var wrongKey SecretKey
+	for _, k := range keys {
+		if k.Public() != correctLeader.Public() {
+			wrongKey = k
+			break
+		}
+	}
+	tamperedFP := &FullProposal{
```
pompon0, line 135: ditto

sei-tendermint/internal/autobahn/types/proposal_test.go
```diff
@@ -0,0 +1,662 @@
...
+			hashed: NewHashed(differentProposal),
+			sig:    fp.proposal.sig, // reuse leader's sig from the original proposal
+		},
+	}
+	err = forgedFP.Verify(committee, vs)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "proposal signature")
+}
+
+func TestVerifyRejectsWrongProposer(t *testing.T) {
```
pompon0: nit: since proposal_test.go belongs to a single types package, prefer prefixing the test names with TestProposal, so that it is clear which type is being tested.

sei-tendermint/internal/autobahn/types/proposal.go
pompon0: perhaps Proposal type also deserves its own Verify method?

wen-coding: Good catch about the LaneRange(), changed.

Also added Verify to Proposal.

sei-tendermint/internal/autobahn/types/proposal_test.go
wen-coding: done

sei-tendermint/internal/autobahn/types/proposal_test.go
wen-coding: done

sei-tendermint/internal/autobahn/types/proposal_test.go
wen-coding: Actually we don't need to verify what error is returned, removed

sei-tendermint/internal/autobahn/types/proposal_test.go
wen-coding: removed

sei-tendermint/internal/autobahn/types/proposal_test.go
wen-coding: done.

sei-tendermint/internal/autobahn/types/proposal_test.go
wen-coding: done

sei-tendermint/internal/autobahn/types/proposal_test.go
wen-coding: done

sei-tendermint/internal/autobahn/types/proposal.go
```diff
@@ -169,6 +169,18 @@
...
 
+// Verify checks that every present lane range belongs to the committee
+// and is internally valid. Lanes may be omitted — omitted lanes are
+// treated as implicit empty ranges by FullProposal.Verify.
+func (m *Proposal) Verify(c *Committee) error {
+	for _, r := range m.laneRanges {
+		if err := r.Verify(c); err != nil {
+			return fmt.Errorf("laneRange[%v]: %w", r.Lane(), err)
+		}
+	}
```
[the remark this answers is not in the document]
wen-coding: We don't care about the order here, need to verify all lanes.

sei-tendermint/internal/autobahn/types/proposal_test.go
```diff
@@ -0,0 +1,639 @@
...
+	rng := utils.TestRng()
+	committee, keys := GenCommittee(rng, 4)
+	vs := ViewSpec{}
+	proposerKey := leaderKey(committee, keys, vs.View())
+
+	// Produce a LaneQC for the proposer's lane.
+	lane := proposerKey.Public()
+	laneQC, _ := makeLaneQC(rng, committee, keys, lane, 0, GenBlockHeaderHash(rng))
+
+	fp, _ := NewProposal(proposerKey, committee, vs, time.Now(),
```
pompon0: nit: you can use utils.OrPanic1() to check for errors in a compact way.

sei-tendermint/internal/autobahn/types/proposal_test.go
```diff
@@ -0,0 +1,639 @@
...
+	committee, keys := GenCommittee(rng, 4)
+	vs := ViewSpec{}
+	proposerKey := leaderKey(committee, keys, vs.View())
+
+	// Build two valid proposals with different timestamps.
+	fp1, _ := NewProposal(proposerKey, committee, vs, time.Now(), nil, utils.None[*AppQC]())
+	fp2, _ := NewProposal(proposerKey, committee, vs, time.Now().Add(time.Hour), nil, utils.None[*AppQC]())
+
+	// Swap: fp2's content with fp1's signature.
+	forgedFP := &FullProposal{
```
pompon0: fyi, I meant more sth like `fp2.proposal.sig = fp1.proposal.sig`, but not a strong opinion.

sei-tendermint/internal/autobahn/types/proposal_test.go
```diff
@@ -0,0 +1,639 @@
...
+
+	// Re-sign the same proposal with a different (non-leader) key.
+	var wrongKey SecretKey
+	for _, k := range keys {
+		if k.Public() != correctLeader.Public() {
+			wrongKey = k
+			break
+		}
+	}
+	tamperedFP := &FullProposal{
```
pompon0, line 135: nit: I think that NewProposal does not verify whether leader is correct, so it can be called directly

sei-tendermint/internal/autobahn/types/proposal_test.go
```diff
@@ -0,0 +1,639 @@
...
+	lane := keys[0].Public()
+	var tamperedRanges []*LaneRange
+	for _, r := range origP.laneRanges {
+		if r.Lane() == lane {
+			tamperedRanges = append(tamperedRanges, &LaneRange{lane: lane, first: 5, next: 5})
+		} else {
+			tamperedRanges = append(tamperedRanges, r)
+		}
+	}
+	tamperedProposal := newProposal(origP.view, origP.createdAt, tamperedRanges, origP.app)
```
pompon0, line 283: NewProposal?

sei-tendermint/internal/autobahn/types/proposal_test.go
```diff
@@ -0,0 +1,639 @@
...
+	var keptRanges []*LaneRange
+	i := 0
+	for _, r := range origP.laneRanges {
+		if i%2 == 0 {
+			keptRanges = append(keptRanges, r)
+		}
+		i++
+	}
+
+	shortProposal := newProposal(origP.view, origP.createdAt, keptRanges, origP.app)
```
pompon0, line 257: NewProposal? Especially the happy path test (no tampering) shouldn't need to access private constructors.

sei-tendermint/internal/autobahn/types/proposal_test.go
```diff
@@ -0,0 +1,639 @@
...
+	leader1 := leaderKey(committee, keys, vs1.View())
+	fp1, _ := NewProposal(leader1, committee, vs1, time.Now(), nil, utils.None[*AppQC]())
+
+	// Tamper: downgrade app to None (lower than commitQC0's app at RoadIndex 0).
+	origP := fp1.Proposal().Msg()
+	var ranges []*LaneRange
+	for _, r := range origP.laneRanges {
+		ranges = append(ranges, r)
+	}
+	tamperedProposal := newProposal(origP.view, origP.createdAt, ranges, utils.None[*AppProposal]())
```
pompon0, line 388: NewProposal?

sei-tendermint/internal/autobahn/types/proposal_test.go
wen-coding: Done

sei-tendermint/internal/autobahn/types/proposal_test.go
wen-coding: Done.

sei-tendermint/internal/autobahn/types/proposal_test.go
wen-coding: Hmm, but maybe we should check?
Since NewProposal has everything to verify the leader, we can check at construction time, instead of sending the wrong proposal across the wires and getting it dropped.

Adding the check now and updating tests.

sei-tendermint/internal/autobahn/types/proposal_test.go
wen-coding: Can't use NewProposal here, because NewProposal's construction logic always produces all lanes and signs over the content. If we use NewProposal and replaces signature, it would fail at "bad signature" in Verify.

sei-tendermint/internal/autobahn/types/proposal_test.go
wen-coding: Same here, NewProposal would set first to a sane value, we need to tamper and sign the wrong value.

sei-tendermint/internal/autobahn/types/proposal_test.go
wen-coding: We force app to be None here, NewProposal will reset app to be old appQC I believe.

PR #2909, opened 2026-02-17
removed tendermint binary and abciclient.

pompon0:
tendermint binary was unused (only some of its commands are imported by seid binary). abciclient was an unncessary abstraction because seid works only with LocalClient which is a noop wrapper.


sei-tendermint/cmd/tendermint/commands/rollback.go
```diff
@@ -3,45 +3,10 @@ package commands
...
 	"github.com/sei-protocol/sei-chain/sei-tendermint/config"
 	"github.com/sei-protocol/sei-chain/sei-tendermint/internal/state"
 )
 
-var removeBlock bool = false
-
-func MakeRollbackStateCommand(conf *config.Config) *cobra.Command {
-	cmd := &cobra.Command{
-		Use:   "rollback",
-		Short: "rollback tendermint state by one height",
```
wen-coding, line 17: Why are we removing rollback commands as well? This only works on tendermint binary?

pompon0: there are 2 rollback commands and seid is using just 1 of them

masih (approved):
Yay; less is more 🚀

PR #2910, opened 2026-02-17
fix: harden PushQC against stale QCs and unverified headers

wen-coding:
## Summary

- Guard QC insertion and `ctrl.Updated()` with `needQC` to prevent stale QCs from advancing `nextQC` or triggering spurious updates.
- Match blocks against stored (already verified) QC headers instead of the unverified incoming QC's headers. Uses per-block stored QC lookup to handle ranges spanning multiple QCs.
- Cap block insertion loop at `inner.nextQC` to prevent nil-pointer dereference when a malicious QC's range extends beyond stored QCs.

## Test plan

- `TestPushQCStaleQCDoesNotCorruptState`: pushes a stale malicious QC with blocks extending beyond stored range. Verifies no panic (min cap), no state corruption (QCs, blocks, nextQC intact), and subsequent valid QC is accepted.
- `TestPushQCIgnoresBlocksMatchingUnverifiedHeaders`: pushes a tampered QC (same range, different headers) with blocks matching the tampered headers. Verifies fake blocks are rejected and real blocks are accepted afterward.
- Both tests verified to catch their respective bugs when the fix is reverted.


sei-tendermint/internal/autobahn/data/state.go
```diff
@@ -115,25 +115,19 @@ func (s *State) Committee() *types.Committee { return s.cfg.Committee }
 func (s *State) PushQC(ctx context.Context, qc *types.FullCommitQC, blocks []*types.Block) error {
 	// Wait until QC is needed.
 	gr := qc.QC().GlobalRange()
-	needQC, err := func() (bool, error) {
```
pompon0, line 118: if that's not too much trouble, I still think that it would be useful to skip verifying the QC if it is not needed. Then the fix would look as follows: if we already have the qc, then just drop the unverified qc and use the one we already have instead.

wen-coding: I can do that, but why prefer skipping verification? Is QC verification very expensive?

sei-tendermint/internal/autobahn/data/state_test.go
```diff
@@ -108,6 +110,76 @@ func TestState(t *testing.T) {
...
+
+	votes := make([]*types.Signed[*types.CommitVote], 0, len(badKeys))
+	for _, k := range badKeys {
+		votes = append(votes, types.Sign(k, types.NewCommitVote(proposal.Proposal().Msg())))
+	}
+	maliciousQC := types.NewFullCommitQC(types.NewCommitQC(votes), headers)
+
+	// Stale QC is silently ignored (verification is skipped, insert guard prevents
+	// advancing nextQC). No error is returned.
+	require.NoError(t, state.PushQC(ctx, maliciousQC, nil))
```
pompon0: the fact that the maliciousQC is ignored here is an implementation detail. Both error and no error are acceptable results. What you should test against is that the state did not change: call QC() on all previous heights - it should return the correct QC (and perhaps check the nextQC, since the api does not expose it currently in any way).

sei-tendermint/internal/autobahn/data/state_test.go
```diff
@@ -108,6 +110,76 @@ func TestState(t *testing.T) {
...
+
+	// Verify state is not corrupted: the next valid QC should still be accepted.
+	qc2, blocks2 := TestCommitQC(rng, committee, keys, utils.Some(qc1.QC()))
+	require.NoError(t, state.PushQC(ctx, qc2, blocks2))
```
pompon0, line 210: in particular this call returning no error does not guarantee that the qc2 was not ignored.

sei-tendermint/internal/autobahn/data/state_test.go
wen-coding: good point, changed.

sei-tendermint/internal/autobahn/data/state_test.go
wen-coding: changed.

PR #2919, opened 2026-02-18
feat(sei-db): add flatkv store implementation

blindchaser:

- Multi-DB storage: storageDB, accountDB, codeDB, metadataDB
- LtHash-based state commitment using internal key formats
- Configurable write toggles per DB type
- Iterator support for storage keys


cody-littley (approved):
LGTM. I only skimmed this, as it is a reconstruction of the PR I reviewed yesterday (of which I did a deep review).

PR #2922, opened 2026-02-18
Add go bench for WAL

yzang2019:
Add write performance go bench for the tidywal as well as our own WAL wrapper


sei-db/wal/wal_bench_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+	"github.com/sei-protocol/sei-chain/sei-db/common/logger"
+)
+
+func makePayload(size int) []byte {
+	buf := make([]byte, size)
+	if _, err := rand.Read(buf); err != nil {
+		panic(err)
+	}
+	return buf
```
cody-littley, line 19: If we want to benchmark large block sizes, we'd probably want to reuse randomness. What I've done before is pre-generate a 1 gigabyte slice of random data, then just choose random sub slices for each operation.

Perhaps not necessary to do in this PR though, since the data sizes you are dealing with are fairly small.

sei-db/wal/wal_bench_test.go
```diff
@@ -0,0 +1,118 @@
...
+	"time"
+
+	"github.com/tidwall/wal"
+
+	"github.com/sei-protocol/sei-chain/sei-db/common/logger"
+)
+
+func makePayload(size int) []byte {
+	buf := make([]byte, size)
+	if _, err := rand.Read(buf); err != nil {
```
blindchaser, line 16: the rand func may introduce significant latency to the metrics, how about reset timer before Write()?

sei-db/wal/wal_bench_test.go
```diff
@@ -0,0 +1,118 @@
...
+				})
+				if err != nil {
+					b.Fatal(err)
+				}
+
+				b.ResetTimer()
+				start := time.Now()
+
+				for i := 0; i < b.N; i++ {
+					payload := makePayload(es)
```
blindchaser: same here

sei-db/wal/wal_bench_test.go
yzang2019: Yeah good point, maybe we can just generate the payload once and reuse it, I just don't know whether it will cause any difference when we are always writing the same data

sei-db/wal/wal_bench_test.go
```diff
@@ -0,0 +1,118 @@
...
+						b.Fatal(err)
+					}
+				}
+
+				b.StopTimer()
+				if err := w.Close(); err != nil {
+					b.Fatal(err)
+				}
+
+				elapsed := time.Since(start)
```
blindchaser, line 110: ns/op stops at b.StopTimer() before Close() and
us/write is computed from time.Since(start) after Close(), so it includes the cost of draining the channel and flushing all pending writes to disk, is it ok?

yzang2019: True, will put StopTimer to after close()

sei-db/wal/wal_bench_test.go
```diff
@@ -0,0 +1,118 @@
...
+			b.Run(name, func(b *testing.B) {
+				dir := b.TempDir()
+				log, err := wal.Open(dir, &wal.Options{
+					NoSync: noSync,
+					NoCopy: true,
+				})
+				if err != nil {
+					b.Fatal(err)
+				}
+				b.Cleanup(func() { _ = log.Close() })
```
masih, line 47: Note that cleanup is run _after_ the benchmark finishes. This is inconsistent with how the wrapper benchmark is implemented.

Meaning, the `BenchmarkWALWrapperWrite` includes the teardown in the benchmarks.

The code should be consistent across the two.

sei-db/wal/wal_bench_test.go
```diff
@@ -0,0 +1,118 @@
...
+				b.ResetTimer()
+				start := time.Now()
+
+				for i := 0; i < b.N; i++ {
+					if err := w.Write(payload); err != nil {
+						b.Fatal(err)
+					}
+				}
+
+				if err := w.Close(); err != nil {
```
masih, line 105: Ditto re taking teardown into account. See my earlier comment.

masih (commented):
Left a comment that I recommend addressing in a follow up PR considering this PR is merged.

Separate from the teardown time issue, the wrapper benchmark matrix varies buffer size, while tidwall varies fsync mode. If the goal is direct comparability, the dimensions differ. Note that wrapper currently opens tidwall WAL with `NoSync: true`.

PR #2924, opened 2026-02-18
Fix off-by-one in PushBlock that causes nil dereference panic

wen-coding:
## Summary

- **Fix off-by-one in `PushBlock`**: The `WaitUntil` condition used `n <= inner.nextQC`, which allows `PushBlock` to proceed when `n == nextQC`. Since `inner.qcs` stores entries for the half-open range `[first, nextQC)`, the map lookup returns `nil`, and the subsequent `qc.Headers()` call panics with a nil pointer dereference. Changed to `n < inner.nextQC` to match every other waiter in the file (`QC`, `Block`, `GlobalBlock`, `AppProposal`).

### DIFF
--- a/sei-tendermint/internal/autobahn/data/state_test.go
+++ b/sei-tendermint/internal/autobahn/data/state_test.go
@@ -301,6 +301,27 @@ func TestExecution(t *testing.T) {
 	}
 }
 
+func TestPushBlockAcceptsBlockWithQC(t *testing.T) {
+	ctx := t.Context()
+	rng := utils.TestRng()
+	committee, keys := types.GenCommittee(rng, 3)
+
+	state := NewState(&Config{
+		Committee: committee,
+	}, utils.None[BlockStore]())
+
+	// Push QC without blocks.
+	qc, blocks := TestCommitQC(rng, committee, keys, utils.None[*types.CommitQC]())
+	require.NoError(t, state.PushQC(ctx, qc, nil))
+	gr := qc.QC().GlobalRange()
+
+	// PushBlock for a block whose QC is already present succeeds immediately.
+	require.NoError(t, state.PushBlock(ctx, gr.First, blocks[0]))
+	got, err := state.TryBlock(gr.First)
+	require.NoError(t, err)
+	require.Equal(t, blocks[0], got)
+}
+
 func TestPushBlockWaitsForQC(t *testing.T) {
 	synctest.Test(t, func(t *testing.T) {
 		ctx := t.Context()


sei-tendermint/internal/autobahn/data/state_test.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/sei-protocol/sei-chain/sei-tendermint/internal/autobahn/types"
 	"github.com/sei-protocol/sei-chain/sei-tendermint/libs/utils"
 	"github.com/sei-protocol/sei-chain/sei-tendermint/libs/utils/scope"
+	"github.com/stretchr/testify/require"
```
pompon0: libs/utils/require. It contains a strongly typed wrappers of testify/require. If you are missing any wrappers, you can add them there.

sei-tendermint/internal/autobahn/data/state_test.go
```diff
@@ -146,3 +147,42 @@ func TestExecution(t *testing.T) {
...
+
+	// Prepare second QC covering [N, M) but don't push it yet.
+	qc2, blocks2 := TestCommitQC(rng, committee, keys, utils.Some(qc1.QC()))
+	gr2 := qc2.QC().GlobalRange()
+
+	// Block gr2.First should not be in state yet.
+	_, err := state.TryBlock(gr2.First)
+	require.ErrorIs(t, err, ErrNotFound)
+
+	// PushBlock for a block in qc2's range. With the off-by-one bug
```
pompon0: nit: scope.Run?

sei-tendermint/internal/autobahn/data/state_test.go
```diff
@@ -146,3 +147,42 @@ func TestExecution(t *testing.T) {
...
+	// (n <= inner.nextQC), this would immediately dereference a nil QC
+	// pointer and panic. With the fix, it waits for the QC.
+	errc := make(chan error, 1)
+	go func() {
```
pompon0: this looks like a great use case for https://go.dev/blog/synctest . Do you want to give it a try?

sei-tendermint/internal/autobahn/data/state_test.go
wen-coding: Changed

sei-tendermint/internal/autobahn/data/state_test.go
wen-coding: done.

sei-tendermint/internal/autobahn/data/state_test.go
wen-coding: done.

PR #2925, opened 2026-02-18
Verify LaneProposal signer and parent hash chain in PushBlock

wen-coding:
## Summary

- **Signer == lane check**: `PushBlock` did not verify that the signer of a `LaneProposal` matches the block's lane. A malicious validator could craft blocks on another validator's lane, bypassing `VerifySig` (which only checks the signer is a committee replica, not the lane producer). Added a cheap signer-lane comparison before expensive crypto verification.
- **Parent hash chain verification**: `PushBlock` did not verify that the incoming block's `parentHash` matches the previous block's header hash. A malicious lane proposer could send a block with a fake `parentHash`, which honest validators would accept and vote on. When `headers()` later tries to reconstruct the block chain by walking `parentHash` links backwards, it cannot find a matching vote and waits forever — deadlocking the node. Added parent hash verification when the previous block is available (after pruning, the check is safely skipped since `headers()` never follows the first block's `parentHash` in a `LaneRange`). A mismatch silently drops the block rather than returning an error, since equivocation (same key producing different chains) is a legitimate scenario.


sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -264,6 +267,20 @@ func (s *State) PushBlock(ctx context.Context, p *types.Signed[*types.LanePropos
 		if q.next != h.BlockNumber() {
 			return nil
 		}
+		// Verify parent hash chain to prevent a malicious proposer from
+		// breaking the block chain, which would deadlock header reconstruction.
+		// A mismatch is not an error: it can happen legitimately when a
+		// producer equivocates (restarts and produces a different chain).
```
pompon0: equivocation of producer is as malicious behavior as equivocation of proposer. It does not endanger the security, but imo it would be worth escalating as an error, so that it gets logged. It will help debugging stalled lanes, if nothing more.

wen-coding: Hmm, if we return an error, and the producer keep pushing the same equivocating block, then every receiver will repeatedly reconnect?
I've changed it to log an error but return nil for now

PR #2930, opened 2026-02-19
Use composite SC for RootMultistore

yzang2019:
The purpose of this PR is to make sure sei-chain will use the new CompositeSC which manages both MemIAVL and FlatKV backend:
- Root multistore (CMS) now creates composite sc instead of memiavl sc
- Fixed a few functions for composite sc to make cosmos only mode behaves the same as before
- Adjust flatkv LoadVersion interface to remove readonly

### DIFF
--- a/app/abci.go
+++ b/app/abci.go
@@ -128,7 +128,7 @@ func (app *App) DeliverTx(ctx sdk.Context, req abci.RequestDeliverTxV2, tx sdk.T
 		telemetry.SetGauge(float32(gInfo.GasUsed), "tx", "gas", "used")
 		telemetry.SetGauge(float32(gInfo.GasWanted), "tx", "gas", "wanted")
 	}()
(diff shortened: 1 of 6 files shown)


sei-db/state_db/sc/composite/store.go
```diff
@@ -92,32 +91,30 @@ func (cs *CompositeCommitStore) LoadVersion(targetVersion int64, readOnly bool)
...
+			logger:          cs.logger,
+			cosmosCommitter: cosmosCommitter,
+			homeDir:         cs.homeDir,
+			config:          cs.config,
+		}
+		// TODO: Support loading FlatKV at target version for read only
+		return newStore, nil
 	}
 
+	cs.cosmosCommitter = cosmosCommitter
```
blindchaser, line 106: it looks before we always create new store and now we mutates self if it's non readOnly.

can you explain a bit why

PR #2932, opened 2026-02-19
chore: improve integration test runs in the CI (build once, reuse across matrix)

mojtaba-esk:

This PR proposes to build the localnode image once in a dedicated job (base from GHCR or build+push, app with GHA cache, then push to ttl.sh with commit-SHA tag). All 14 matrix test jobs pull that image and run the cluster/tests instead of building. Makefile adds build-docker-node-app and drops proxy/local-only logic; Dockerfiles use ubuntu:25.04.

### Comparison Scope
This compares the **Baseline CI architecture** vs the **Run after improvements applied (Latest run)* *.

- Baseline: Each of the 14 jobs builds its own image.
- Latest: One shared image build, then 14 jobs run in parallel.
- Total time is measured as wall-clock duration (critical path).

---

### Wall Clock Time

| Run | Total Time |
|------|------------|
| Baseline | 1h 56m 27s |
| Latest | 1h 05m 50s |

**Total reduction: 50m 37s (~43.5% faster)**

---

### Critical Path (Normalized Comparison)

Since baseline jobs each built their own image, we compare:

- Baseline bottleneck:  
  `EVM Interoperability = 14m 17s`

- Latest bottleneck (Build + EVM Interoperability):  
  `2m 17s + 10m 19s = 12m 36s`

| Metric | Baseline | Latest |
|--------|----------|--------|
| Critical Path | 14m 17s | 12m 36s |

**Net improvement on true per-job cost: 1m 41s (~12% faster)**

---

### Resource Usage Impact

Baseline:
- 14 jobs each built their own image.
- Build step duplicated 14 times.
- Significant redundant compute usage.

Latest:
- Image built once (2m 17s).
- Eliminates 13 redundant builds.
- Reduces overall runner CPU time and resource contention.

This reduces total runner-minute consumption while also improving wall-clock time.

---

### Summary

- CI wall-clock time reduced by ~50 minutes.
- ~43% faster pipeline execution.
- Shared image build removes redundant compute work.
- Critical path is now ~12% faster than baseline.
- Lower compute waste and more scalable architecture.

While the per-job runtime gain is modest (~1m 41s), the overall CI latency and resource efficiency improvements are substantial.

**Note:** _Timings were collected from the GitHub Actions usage page for this PR only and may vary due to runner load, caching state, and other environmental factors._

### DIFF
--- a/sei-db/ledger_db/parquet/wal.go
+++ b/sei-db/ledger_db/parquet/wal.go
@@ -1,6 +1,7 @@
 package parquet
 
 import (
+	"context"
 	"encoding/binary"
 	"fmt"
 	"os"
@@ -109,6 +110,7 @@ func NewWAL(logger dbLogger.Logger, dir string) (dbwal.GenericWAL[WALEntry], err
 		return nil, err
 	}
 	return dbwal.NewWAL(
+		context.Background(),
 		encodeWALEntry,
 		decodeWALEntry,
 		logger,
--- a/sei-db/wal/changelog.go
+++ b/sei-db/wal/changelog.go
@@ -1,6 +1,8 @@
 package wal
 
 import (
+	"context"
+
 	"github.com/sei-protocol/sei-chain/sei-db/common/logger"
 	"github.com/sei-protocol/sei-chain/sei-db/proto"
 )
@@ -12,6 +14,7 @@ type ChangelogWAL = GenericWAL[proto.ChangelogEntry]
 // This is a convenience wrapper that handles serialization automatically.
 func NewChangelogWAL(logger logger.Logger, dir string, config Config) (ChangelogWAL, error) {
 	return NewWAL(
+		context.Background(),
 		func(e proto.ChangelogEntry) ([]byte, error) { return e.Marshal() },
 		func(data []byte) (proto.ChangelogEntry, error) {
 			var e proto.ChangelogEntry
--- a/sei-db/wal/wal.go
+++ b/sei-db/wal/wal.go
@@ -1,11 +1,11 @@
 package wal
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
-	"sync"
 	"time"
 
 	"github.com/tidwall/wal"
@@ -13,26 +13,77 @@ import (
 	"github.com/sei-protocol/sei-chain/sei-db/common/logger"
 )
 
+// The size of internal channel buffers if the provided buffer size is less than 1.
+const defaultBufferSize = 1024
+
+// The size of write batches if the provided write batch size is less than 1.
+const defaultWriteBatchSize = 64
+
 // WAL is a generic write-ahead log implementation.
 type WAL[T any] struct {
-	dir             string
-	log             *wal.Log
-	config          Config
-	logger          logger.Logger
-	marshal         MarshalFn[T]
-	unmarshal       UnmarshalFn[T]
-	writeChannel    chan T
-	mtx             sync.RWMutex // guards WAL state: lazy init/close of writeChannel, isClosed checks
-	asyncWriteErrCh chan error   // buffered=1; async writer reports first error non-blocking
-	isClosed        bool
-	closeCh         chan struct{}  // signals shutdown to background goroutines
-	wg              sync.WaitGroup // tracks background goroutines (pruning)
+	ctx    context.Context
+	cancel context.CancelFunc
+
+	dir       string
+	log       *wal.Log
+	config    Config
+	logger    logger.Logger
+	marshal   MarshalFn[T]
+	unmarshal UnmarshalFn[T]
+
+	// The size of write batches.
+	writeBatchSize int
+	asyncWrites    bool
+
+	writeChan    chan *writeRequest[T]
+	truncateChan chan *truncateRequest
+	closeReqChan chan struct{}
+	closeErrChan chan error
 }
 
+// A request to truncate the log.
+type truncateRequest struct {
+	// If true, truncate before the provided index. Otherwise, truncate after the provided index.
+	before bool
+	// The index to truncate at.
+	index uint64
+	// Errors are returned over this channel, nil is written if completed with no error
+	errChan chan error
+}
+
+// A request to write to the WAL.
+type writeRequest[T any] struct {
+	// The data to write
+	entry T
+	// Errors are returned over this channel, nil is written if completed with no error
+	errChan chan error
+}
+
+// Configuration for the WAL.
 type Config struct {
+	// The number of recent entries to keep in the log.
+	KeepRecent uint64
+
+	// The interval at which to prune the log.
+	PruneInterval time.Duration
+
+	// The size of internal buffers. Also controls whether or not the Write method is asynchronous.
+	//
+	// If BufferSize is greater than 0, then the Write method is asynchronous, and the size of internal
+	// buffers is set to the provided value. If Buffer size is less than 1, then the Write method is synchronous,
+	// and any internal buffers are set to a default size.
 	WriteBufferSize int
-	KeepRecent      uint64
-	PruneInterval   time.Duration
+
+	// The size of write batches. If less than or equal to 0, a default of 64 is used.
+	// If 1, no batching is done.
+	WriteBatchSize int
+
+	// If true, do an fsync after each write.
+	FsyncEnabled bool
+
+	// If true, make a deep copy of the data for every write. If false, then it is not safe to modify the data after
+	// reading/writing it.
+	DeepCopyEnabled bool
 }
 
 // NewWAL creates a new generic write-ahead log that persists entries.
@@ -49,6 +100,7 @@ type Config struct {
 //	    logger, dir, config,
 //	)
 func NewWAL[T any](
+	ctx context.Context,
 	marshal MarshalFn[T],
 	unmarshal UnmarshalFn[T],
 	logger logger.Logger,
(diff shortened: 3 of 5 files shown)


docker/localnode/Dockerfile.base
```diff
@@ -0,0 +1,16 @@
+# Base image with Go, Foundry, and Node only. Rebuild when tool versions change.
+# Tag: sei-chain/localnode-base:go1.25.6
+FROM ubuntu:25.04
+ENV HOME="/root" PATH="/root/go/bin:$PATH"
+RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
+    --mount=type=cache,target=/var/lib/apt/lists,sharing=locked \
+    apt-get update && apt-get install -y make build-essential git jq python3 curl vim uuid-runtime
+RUN curl -L https://go.dev/dl/go1.25.6.linux-amd64.tar.gz | tar xvzf - -C /usr/local/
+RUN curl -L https://foundry.paradigm.xyz | bash
```
bdchatham: nit: may be worthwhile to pin this to a specific release. As is, this will take whatever that request returns, which could vary from run to run. Can also validate the checksum for integrity. Reproducibility in practice being the more concerning point to me.

Makefile
```diff
@@ -212,14 +212,75 @@ build-price-feeder-linux:
...
-# Build docker image for detected platform
-build-docker-node:
-	@echo "Building for $(DOCKER_PLATFORM)..."
-	@cd docker && docker build --tag sei-chain/localnode localnode --platform $(DOCKER_PLATFORM)
+# Optional: set DOCKER_BUILD_CACHE_FROM/DOCKER_BUILD_CACHE_TO (e.g. type=gha) to use Buildx layer cache
+DOCKER_BUILD_CACHE_FROM ?=
+DOCKER_BUILD_CACHE_TO ?=
+LOCALNODE_BASE_TAG := sei-chain/localnode-base:go1.25.6
+
+# Base image (Go, Foundry, Node). Rebuild only when tool versions or Dockerfile.base change.
```
bdchatham: Just a matter of preference: this seems a little misleading on first read. I initially interpreted this as the target enforcing this diff check.

Makefile
```diff
@@ -212,14 +212,75 @@ build-price-feeder-linux:
...
+			--tag $(LOCALNODE_BASE_TAG) \
+			--platform $(DOCKER_PLATFORM) \
+			-f localnode/Dockerfile.base \
+			localnode; \
+	else \
+		docker build --tag $(LOCALNODE_BASE_TAG) -f localnode/Dockerfile.base localnode --platform $(DOCKER_PLATFORM); \
+	fi
+.PHONY: build-docker-node-base
+
+# Build only the app image (requires sei-chain/localnode-base:go1.25.6 to exist locally; used in CI when base is pulled or pre-built)
```
bdchatham: Instead of commenting this dependency, consider codifying it and making the build of the localnode-base a prerequisite or adding a log to make the outcome clear.

.github/workflows/integration-test.yml
```diff
@@ -17,8 +17,85 @@ defaults:
   run:
     shell: bash
 
+env:
+  GHCR_BASE_IMAGE: ghcr.io/${{ github.repository_owner }}/sei-chain-localnode-base:go1.25.6
```
masih: Which container registry used to host the image is probably a minor detail, right? In that spirit, I would rename this env to be more representative of what it does. Maybe `IT_BASE_IMAGE` or `INTEGRATION_TEST_BASE_IMAGE`?

Makefile
```diff
@@ -212,14 +212,75 @@ build-price-feeder-linux:
...
+	fi
+.PHONY: build-docker-node-app
+
+# Build app image with prebuilt seid binary (context = repo root; requires build/seid and build/price-feeder to exist).
+# Used in CI: build seid in a container first, then run this target so the image contains the binary.
+build-docker-node-app-prebuilt:
+	@echo "Building localnode app image (with prebuilt binary)..."
+	@if [ ! -f build/seid ] || [ ! -f build/price-feeder ]; then \
+		echo "ERROR: build/seid and build/price-feeder must exist. Run the build step first."; exit 1; \
+	fi
```
bdchatham, line 259: Nice!

Makefile
```diff
@@ -212,14 +212,75 @@ build-price-feeder-linux:
...
+			--platform $(DOCKER_PLATFORM) \
+			-f localnode/Dockerfile \
+			localnode; \
+	else \
+		docker build --cache-from $(LOCALNODE_BASE_TAG) --tag sei-chain/localnode -f localnode/Dockerfile localnode --platform $(DOCKER_PLATFORM); \
+	fi
+.PHONY: build-docker-node-app
+
+# Build app image with prebuilt seid binary (context = repo root; requires build/seid and build/price-feeder to exist).
+# Used in CI: build seid in a container first, then run this target so the image contains the binary.
```
bdchatham: nit: we're building `seid` in the runner I believe.

.github/workflows/integration-test.yml
```diff
@@ -17,8 +17,85 @@ defaults:
...
+          DOCKER_BUILDKIT: 1
+          DOCKER_BUILD_CACHE_FROM: type=gha
+          DOCKER_BUILD_CACHE_TO: type=gha,mode=max
+          GHCR_ORG: ${{ github.repository_owner }}
+        run: make build-docker-node-app-prebuilt
+
+      - name: Push app image to GHCR
+        run: |
+          docker tag sei-chain/localnode ${{ steps.image_name.outputs.image_name }}
+          docker push ${{ steps.image_name.outputs.image_name }}
```
masih, line 94: This builds and pushes the image _per commit_. We would be hitting the max size limit on GHCR pretty quickly? But then these images are very short-lived. Right?

Non blocker but I wonder if we should use ECR with a propepr lifecycle policy instead. _Or_ build the image as a tar ball, upload it to GHA as an artifact, then download it in other jobs. This would save the push and pull dance.

.github/workflows/integration-test.yml
```diff
@@ -167,16 +244,52 @@ jobs:
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Pull localnode image
+        run: |
+          docker pull ${{ needs.build-localnode.outputs.image_name }}
+          docker tag ${{ needs.build-localnode.outputs.image_name }} sei-chain/localnode
```
masih, line 251: Why tag again? we know what we just pulled right?

.github/workflows/integration-test.yml
```diff
@@ -167,16 +244,52 @@ jobs:
...
       - name: Start 4 node docker cluster
-        run: make clean && DOCKER_DETACH=true INVARIANT_CHECK_INTERVAL=10 ${{matrix.test.env}} make docker-cluster-start
+        run: DOCKER_DETACH=true INVARIANT_CHECK_INTERVAL=10 ${{matrix.test.env}} make docker-cluster-up
```
masih: Move env vars to step `env` ?

Makefile
```diff
@@ -212,14 +212,75 @@ build-price-feeder-linux:
...
+			--platform $(DOCKER_PLATFORM) \
+			-f docker/localnode/Dockerfile.prebuilt \
+			.; \
+	else \
+		docker build --build-arg GHCR_ORG=$(GHCR_ORG) --tag sei-chain/localnode -f docker/localnode/Dockerfile.prebuilt . --platform $(DOCKER_PLATFORM); \
+	fi
+.PHONY: build-docker-node-app-prebuilt
+
+# Build docker image for detected platform (depends on base; app layer is quick)
+build-docker-node: build-docker-node-base
```
bdchatham: This is better since you're explicitly orchestrating dependencies rather than them being in a comment most people (or agents) won't immediately see. 

Also, not immediately clear why you'd use a sub-make target for `build-docker-node-app`. If you do: `build-docker-node: build-docker-node-base build-docker-node-app` then Makefile analysis tools will be able to better understand make target relationships.

Let me know if I've missed something.

docker/localnode/Dockerfile.base
mojtaba-esk: Very good point. I applied the changes and will be pushing it ASAP. Pinned the installer to the `v1.5.1` tag and run `foundryup -i v1.5.1`; skipping checksum verification for now since we're pulling it from the official github tag over https.

Makefile
mojtaba-esk: I see, made it clear now.

Makefile
mojtaba-esk: Nice catch! added.

Makefile
mojtaba-esk: oh yeah the comment needs to be updated or removed. good catch!

Makefile
mojtaba-esk: Good point on the sub-make. I've switched to listing both as prerequisites:
`build-docker-node: build-docker-node-base build-docker-node-app`
so the dependency graph is explicit and tools (and readers) can see it without running the recipe. No need for a submake here since the target only needs those two up to date.

docker/localnode/Dockerfile
```diff
@@ -1,21 +1,12 @@
...
-RUN curl -sL https://deb.nodesource.com/setup_16.x | bash
-RUN /root/.foundry/bin/foundryup
-RUN mkdir -p /root/go/pkg/mod && \
-    mkdir -p /root/.cache && \
-    chmod -R a+rwX /root
-SHELL ["/bin/bash", "-c"]
+# Application image: base + scripts. Rebuild when scripts or base tag change.
 
+ARG GHCR_ORG=sei-protocol
+FROM ghcr.io/${GHCR_ORG}/sei-chain-localnode-base:go1.25.6
```
masih: This might be misleading, in that the local docker cluster start now always pulls from remote, regardless of make target executed. Right?

docker/localnode/Dockerfile
```diff
@@ -1,21 +1,12 @@
...
-RUN curl -L https://foundry.paradigm.xyz | bash
-RUN curl -sL https://deb.nodesource.com/setup_16.x | bash
-RUN /root/.foundry/bin/foundryup
-RUN mkdir -p /root/go/pkg/mod && \
-    mkdir -p /root/.cache && \
-    chmod -R a+rwX /root
-SHELL ["/bin/bash", "-c"]
+# Application image: base + scripts. Rebuild when scripts or base tag change.
 
+ARG GHCR_ORG=sei-protocol
```
masih: I doubt we would ever use any other org. Better to set an ARG for FROM image?

.github/workflows/integration-test.yml
mojtaba-esk: delicate point! ofc, renamed it to INTEGRATION_TEST_BASE_IMAGE so the name reflects what it's for rather than the registry.

.github/workflows/integration-test.yml
mojtaba-esk: Yes, it's one image per commit. I originally used `ttl.sh` (images expire after a time window, e.g. 24h), but push time was ~50s there vs ~12–15s on GHCR, so I switched to GHCR for speed.
ECR with a lifecycle policy makes sense. Another option I'm considering is deleting the image from GHCR after the workflow run finishes (on success or failure), so we don't accumulate tags. WDYT?

.github/workflows/integration-test.yml
mojtaba-esk: it is b/c downstream steps expect the image as `sei-chain/localnode`, not the one with GHCR name.

.github/workflows/integration-test.yml
mojtaba-esk: you're right. nice one. organized it.

docker/localnode/Dockerfile
mojtaba-esk: sure. keep it simple :)

docker/localnode/Dockerfile
mojtaba-esk: Nice catch. if user tries to build it the image tag might be different and if this image does not exist in the registry, it cannot pull it. as we already tag it in the ci once we pull it, we can simply rename the tag to a local name. so in that case if user runs local cluster start, base and app are built on the machine; nothing is pulled from GHCR.

.github/workflows/integration-test.yml
mojtaba-esk: A cleanup job is added.

PR #2934, opened 2026-02-19
WAL wrapper improvements

cody-littley:

Ticket: https://linear.app/seilabs/issue/STO-361/improve-wal-library-correctness

This PR fixes several issues identified in this thread: https://sei-network-io.slack.com/archives/C08KPKN3SLT/p1771432603483039?thread_ts=1771432091.719279&cid=C08KPKN3SLT

- Async vs. non-async write mode is now an explicit toggle in settings
    - legacy code that expects the write buffer size to control async behavior has been modified to maintain legacy behavior
- Added toggles for fsync mode and shallow copy mode. 
- Threading model altered. Instead of using mutexes, all work is performed on a background goroutine, and communication with this goroutine is mediated with channels.
    - Fixes identified thread safety issues.
    - This refactor is not the only way to fix the issues identified in the thread, and so if people don't like this change I'm willing to roll to the old mutex-based implementation.
    - My personal opinion: when it comes to reasoning about multi-threaded code, this pattern is easier to reason about than using mutexes. Since there is only one thread "doing things", it's trivial to convince yourself of its safety.
    - This threading change can sometimes lead to performance improvements. But for this specific case, I think it highly unlikely that this will have a performance impact. It's very unlikely that we will see mutex contention impact performance in this code, and so performance will likely be near identical to a mutex-based implementation.
- Added behind-the-scenes batching. Top level API remains identical. But if there is a bunch of work that builds up in the queue, the worker thread aggregates multiple writes into larger batches.
    - Large impact for small writes in async mode. 2-3x throughput improvement with writes less than 10kb.
    - Impact tapers off the larger the writes get. This is expected.
    - Extreme impact (orders of magnitude) when writing with fsync mode enabled, since fsync is super expensive and this lets us do fewer fsync operations. 
        - For the consensus team's use case, this is unlikely to help much though, since they will be doing one WAL per lane and blocking on each write before proceeding to the next write.
        - If we move to all lanes in a single WAL, then consensus team will likely realize a large speedup from batching.
        - It's possible to make the consensus team's use case much faster, but to do so I'd need to completely change `wal.go`'s API and semantics. I didn't think that appropriate in this PR.

I utilized @yzang2019's benchmark. My changes had no impact on performance in some cases, and had a strong positive impact in others.

 
Screenshot 2026-02-18 at 3 54 48 PM

### DIFF
--- a/app/seidb.go
+++ b/app/seidb.go
@@ -27,6 +27,9 @@ const (
 	FlagSCSnapshotWriteRateMBps            = "state-commit.sc-snapshot-write-rate-mbps"
 	FlagSCCacheSize                        = "state-commit.sc-cache-size"
 	FlagSCOnlyAllowExportOnSnapshotVersion = "state-commit.sc-only-allow-export-on-snapshot-version"
+	FlagSCHistoricalProofMaxInFlight       = "state-commit.sc-historical-proof-max-inflight"
+	FlagSCHistoricalProofRateLimit         = "state-commit.sc-historical-proof-rate-limit"
+	FlagSCHistoricalProofBurst             = "state-commit.sc-historical-proof-burst"
 
 	// SS Store configs
 	FlagSSEnable            = "state-store.ss-enable"
@@ -110,6 +113,17 @@ func parseSCConfigs(appOpts servertypes.AppOptions) config.StateCommitConfig {
 	scConfig.MemIAVLConfig.SnapshotWriterLimit = cast.ToInt(appOpts.Get(FlagSCSnapshotWriterLimit))
 	scConfig.MemIAVLConfig.SnapshotPrefetchThreshold = cast.ToFloat64(appOpts.Get(FlagSCSnapshotPrefetchThreshold))
 	scConfig.MemIAVLConfig.SnapshotWriteRateMBps = cast.ToInt(appOpts.Get(FlagSCSnapshotWriteRateMBps))
+
+	if v := appOpts.Get(FlagSCHistoricalProofMaxInFlight); v != nil {
+		scConfig.HistoricalProofMaxInFlight = cast.ToInt(v)
+	}
+	if v := appOpts.Get(FlagSCHistoricalProofRateLimit); v != nil {
+		scConfig.HistoricalProofRateLimit = cast.ToFloat64(v)
+	}
+	if v := appOpts.Get(FlagSCHistoricalProofBurst); v != nil {
+		scConfig.HistoricalProofBurst = cast.ToInt(v)
+	}
+
 	return scConfig
 }
 
--- a/app/seidb_test.go
+++ b/app/seidb_test.go
@@ -64,3 +64,24 @@ func TestNewDefaultConfig(t *testing.T) {
 	assert.Equal(t, scConfig, config.DefaultStateCommitConfig())
 	assert.Equal(t, ssConfig, config.DefaultStateStoreConfig())
 }
+
+type mapAppOpts map[string]interface{}
+
+func (m mapAppOpts) Get(s string) interface{} {
+	return m[s]
+}
+
+func TestParseSCConfigs_HistoricalProofFlags(t *testing.T) {
+	appOpts := mapAppOpts{
+		FlagSCEnable: true,
+
+		FlagSCHistoricalProofMaxInFlight: 7,
+		FlagSCHistoricalProofRateLimit:   12.5,
+		FlagSCHistoricalProofBurst:       3,
+	}
+
+	scConfig := parseSCConfigs(appOpts)
+	assert.Equal(t, 7, scConfig.HistoricalProofMaxInFlight)
+	assert.Equal(t, 12.5, scConfig.HistoricalProofRateLimit)
+	assert.Equal(t, 3, scConfig.HistoricalProofBurst)
+}
--- a/codecov.yml
+++ b/codecov.yml
@@ -12,7 +12,9 @@ coverage:
           - sei-chain
     patch:
       default:
-        target: auto
+        informational: true   # Show patch coverage, but do not fail CI for it.
+        target: 50%           # Keep displayed goal stable across PRs.
+        threshold: 5%         # Toleration for noisy small diffs.
 
 comment:
   layout: "reach,diff,flags,tree,betaprofiling"
@@ -45,6 +47,7 @@ ignore:
   # Project-specific exclusions
   - "precompiles/**/legacy/**/*"
   - "**/cli/**/*"
+  - "sei-db/state_db/bench"
 
 flags:
   sei-chain:
--- a/sei-cosmos/storev2/rootmulti/store.go
+++ b/sei-cosmos/storev2/rootmulti/store.go
@@ -11,6 +11,7 @@ import (
 
 	"cosmossdk.io/errors"
 	"github.com/armon/go-metrics"
+	"golang.org/x/time/rate"
 
 	protoio "github.com/gogo/protobuf/io"
 	snapshottypes "github.com/sei-protocol/sei-chain/sei-cosmos/snapshots/types"
@@ -52,6 +53,9 @@ type Store struct {
 	ckvStores      map[types.StoreKey]types.CommitKVStore
 	gigaKeys       []string
 	pruningManager *pruning.Manager
+
+	histProofSem     chan struct{}
+	histProofLimiter *rate.Limiter
 }
 
 type VersionedChangesets struct {
@@ -72,14 +76,30 @@ func NewStore(
 	if scConfig.Directory != "" {
 		scDir = scConfig.Directory
 	}
+	maxInFlight := scConfig.HistoricalProofMaxInFlight
+	if maxInFlight <= 0 {
+		maxInFlight = 1
+	}
+
+	burst := scConfig.HistoricalProofBurst
+	if burst <= 0 {
+		burst = 1
+	}
+
+	var limiter *rate.Limiter
+	if scConfig.HistoricalProofRateLimit > 0 {
+		limiter = rate.NewLimiter(rate.Limit(scConfig.HistoricalProofRateLimit), burst)
+	}
 	scStore := composite.NewCompositeCommitStore(scDir, logger, scConfig)
 	store := &Store{
-		logger:       logger,
-		scStore:      scStore,
-		storesParams: make(map[types.StoreKey]storeParams),
-		storeKeys:    make(map[string]types.StoreKey),
-		ckvStores:    make(map[types.StoreKey]types.CommitKVStore),
-		gigaKeys:     gigaKeys,
+		logger:           logger,
+		scStore:          scStore,
+		storesParams:     make(map[types.StoreKey]storeParams),
+		storeKeys:        make(map[string]types.StoreKey),
+		ckvStores:        make(map[types.StoreKey]types.CommitKVStore),
+		gigaKeys:         gigaKeys,
+		histProofSem:     make(chan struct{}, maxInFlight),
+		histProofLimiter: limiter,
 	}
 	if ssConfig.Enable {
 		ssStore, err := ss.NewStateStore(logger, homeDir, ssConfig)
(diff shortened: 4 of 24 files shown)


sei-db/wal/wal.go
```diff
@@ -49,249 +81,502 @@
...
+		closeReqChan:   make(chan struct{}),
+		closeErrChan:   make(chan error, 1),
+		writeChan:      make(chan *writeRequest[T], bufferSize),
+		truncateChan:   make(chan *truncateRequest, bufferSize),
+		getOffsetChan:  make(chan *getOffsetRequest, bufferSize),
+		readAtChan:     make(chan *readAtRequest[T], bufferSize),
+		replayChan:     make(chan *replayRequest[T], bufferSize),
 	}
+
+	go w.mainLoop()
```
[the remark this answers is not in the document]
cody-littley: not a problem here

sei-db/wal/changelog.go
```diff
@@ -11,7 +13,15 @@ type ChangelogWAL = GenericWAL[proto.ChangelogEntry]
 // NewChangelogWAL creates a new WAL for ChangelogEntry.
 // This is a convenience wrapper that handles serialization automatically.
 func NewChangelogWAL(logger logger.Logger, dir string, config Config) (ChangelogWAL, error) {
+
+	// Legacy compatibility: originally, BufferSize > 0 implied async writes. Preserve that for
+	// callers who set BufferSize without explicitly setting AsyncWrites.
+	if config.BufferSize > 0 {
+		config.AsyncWrites = true
```
bdchatham: nit: putting a method on the config like `useAsyncWriter` that does this check rather than doing this backwards translation seems more straightforward.

sei-db/wal/changelog.go
```diff
@@ -11,7 +13,15 @@ type ChangelogWAL = GenericWAL[proto.ChangelogEntry]
 // NewChangelogWAL creates a new WAL for ChangelogEntry.
 // This is a convenience wrapper that handles serialization automatically.
 func NewChangelogWAL(logger logger.Logger, dir string, config Config) (ChangelogWAL, error) {
+
+	// Legacy compatibility: originally, BufferSize > 0 implied async writes. Preserve that for
+	// callers who set BufferSize without explicitly setting AsyncWrites.
+	if config.BufferSize > 0 {
```
yzang2019: If already we have BufferSize to indicate async or not, why do we still need to add another config for AsyncWrites (true/false)  here? Isn't that a redundant config?

cody-littley: Previously, we would only use the buffer when running in async mode. After my refactor, we always use the buffer regardless of whether we are running in async mode. This means we need to configure buffer size independently of async mode configuration.

My reasoning behind always using the buffer is twofold. It reduces complexity, since there is one fewer code pathway. Additionally, it allows for batching (internal) even when running in sync mode (although you need multiple writer goroutines to take advantage of batching in sync mode).

sei-db/wal/wal.go
```diff
@@ -49,249 +81,502 @@ type Config struct {
...
-		config:          config,
-		logger:          logger,
-		marshal:         marshal,
-		unmarshal:       unmarshal,
-		closeCh:         make(chan struct{}),
-		asyncWriteErrCh: make(chan error, 1),
+
+	bufferSize := config.BufferSize
+	if config.BufferSize == 0 {
+		bufferSize = 1024
```
yzang2019: Maybe have a const for these default value? Same as writeBatchSize

sei-db/ledger_db/parquet/wal.go
```diff
@@ -109,6 +110,7 @@ func NewWAL(logger dbLogger.Logger, dir string) (dbwal.GenericWAL[WALEntry], err
 		return nil, err
 	}
 	return dbwal.NewWAL(
+		context.Background(),
```
bdchatham, line 113: Currently not handling caller's context. Coming in a later PR?

sei-db/wal/changelog.go
cody-littley: Let me know if you feel strongly about this. I can revert back to the old pattern of configuration without a huge rewrite.

sei-db/wal/wal.go
```diff
@@ -15,24 +15,56 @@ import (
...
-	KeepRecent      uint64
-	PruneInterval   time.Duration
+	// The number of recent entries to keep in the log.
+	KeepRecent uint64
+
+	// The interval at which to prune the log.
+	PruneInterval time.Duration
+
+	// If true, the writes are asynchronous, and will return immediately without waiting for the write to be durable
+	AsyncWrites bool
```
yzang2019: I feel it sometimes could be confusing to have two configs to control one feature, user don't know which one dominates without looking into the code. Would suggest still keep only one here which is the bufferSize > 0 to indicate whether we are doing async writes or not

sei-db/wal/changelog.go
yzang2019: Let me take a look at the new code path first for how buffer is being used now, thanks for explanation!

sei-db/wal/wal.go
```diff
@@ -325,3 +610,48 @@ func open(dir string, opts *wal.Options) (*wal.Log, error) {
 	}
 	return rlog, err
 }
+
+// The main loop doing work in the background.
+func (walLog *WAL[T]) mainLoop() {
+
+	pruneInterval := walLog.config.PruneInterval
+	if pruneInterval < time.Second {
```
yzang2019: What about the case where we want to disable auto pruning? I think previously we have assumption if pruneInterval <= 0 then we will disable pruning, and we still need to support that case right?

sei-db/wal/wal.go
cody-littley: Const added for both.

sei-db/wal/wal.go
cody-littley: Config reverted to a single setting, as suggested.

sei-db/wal/wal.go
cody-littley: This is still supported, but I can see how it is confusing in its current form. Currently, `prune()` becomes a noop if it is disabled:

```
func (walLog *WAL[T]) prune() {
	keepRecent := walLog.config.KeepRecent
	if keepRecent <= 0 || walLog.config.PruneInterval <= 0 {
		// pruning is disabled
		return
	}
```

I've refactored the code to make it a more obvious that pruning is disabled.

```
	var pruneChan <-chan time.Time
	if walLog.config.PruneInterval > 0 && walLog.config.KeepRecent > 0 {
		pruneTicker := time.NewTicker(walLog.config.PruneInterval)
		defer pruneTicker.Stop()
		pruneChan = pruneTicker.C
	}

// ...

		case <-pruneChan:
			walLog.prune()
```

The `pruneChan` is nil if pruning is not configured. `<- nil` is legal in go, it will just never be selected/executed.

sei-db/wal/wal.go
```diff
@@ -49,249 +81,502 @@ type Config struct {
...
+// include them in the batch.
+func (walLog *WAL[T]) handleBatchedWrite(req *writeRequest[T]) {
+
+	requests := make([]*writeRequest[T], 0)
+	requests = append(requests, req)
+
+	keepLooking := true
+	for keepLooking && len(requests) < walLog.writeBatchSize {
+		select {
+		case req := <-walLog.writeChan:
```
bdchatham: nit: shadowing is misleading here.

sei-db/wal/changelog.go
cody-littley: I already pushed a change to revert to the previous config pattern. We can still use the old pattern, you'd just lose the ability to configure internal buffer sizes when you are running in sync mode. Perhaps not too much of a problem? Since internal buffer size is unlikely to be important super important when running in sync mode.

sei-db/wal/wal.go
```diff
@@ -49,249 +85,506 @@ type Config struct {
...
+		req.errChan <- fmt.Errorf("failed to write: %w", err)
+		return
+	}
+
+	req.errChan <- nil
+}
+
+// handleBatchedWrite is called when batching is enabled. This method may pop pending writes from the writeChan and
+// include them in the batch.
+func (walLog *WAL[T]) handleBatchedWrite(req *writeRequest[T]) {
```
bdchatham, line 219: nit: I'd probably break this method up a bit given it has multiple independent steps.

sei-db/wal/wal.go
```diff
@@ -49,249 +85,506 @@ type Config struct {
...
+		return
+	}
+
+	for _, r := range requests {
+		r.errChan <- nil
+	}
+}
+
+// A request to truncate the log.
+type truncateRequest struct {
```
bdchatham: Can we put this at the top of the file?

sei-db/wal/wal.go
```diff
@@ -325,3 +618,48 @@ func open(dir string, opts *wal.Options) (*wal.Log, error) {
...
+			walLog.handleReplay(req)
+		case <-pruneChan:
+			walLog.prune()
+		case <-walLog.closeReqChan:
+			running = false
+		}
+	}
+
+	// drain pending work, then tear down
+	walLog.drain()
```
yzang2019, line 523: Shall we do this instead to be safer?
walLog.cancel()     // 1. stop accepting new requests
walLog.drain()      // 2. finish in-flight work already in the buffer

yzang2019: Between drain() returning and cancel() being called , the context is still active. A concurrent caller (e.g. Write()) can successfully send a request into the buffered channel because the context check in Write() passes. But mainLoop has already stopped reading — that request is silently dropped and its errChan will never receive a response.

sei-db/wal/wal.go
```diff
@@ -49,135 +82,276 @@ type Config struct {
...
+// sendTruncate sends a truncate request to the main loop and waits for completion.
+func (walLog *WAL[T]) sendTruncate(before bool, index uint64) error {
+	req := &truncateRequest{
+		before:  before,
+		index:   index,
+		errChan: make(chan error, 1),
+	}
+
+	select {
+	case _, ok := <-walLog.ctx.Done():
```
yzang2019: I think we usually just do this:
case <-walLog.ctx.Done():
  return err

sei-db/wal/wal.go
```diff
@@ -49,249 +85,506 @@ type Config struct {
...
+
+// A request to read an entry at a specific index.
+type readAtRequest[T any] struct {
+	index        uint64
+	responseChan chan *readAtResponse[T]
 }
 
-// ReadAt will read the log entry at the provided index
+// A response to a read-at request.
+type readAtResponse[T any] struct {
```
bdchatham: All structs in a separate file or at the top of this file would be helpful.

sei-db/wal/wal.go
cody-littley: Note that even if a caller manages to submit a message to the buffer after `Close()` is called but before the context is cancelled, they will not block forever waiting for a response. When waiting for a response, we abort early if the context is cancelled:

```
	select {
	case _, ok := <-walLog.ctx.Done():
		if !ok {
			return fmt.Errorf("WAL was closed after write was submitted but before write was finalized, " +
				"write may or may not be durable")
		}
	case err := <-req.errChan:
```

cody-littley: That being said, I don't think there is any harm in cancelling the context before draining additional work. Change made.

yzang2019 (approved):
Overall LGTM!

sei-db/wal/wal.go
cody-littley: I'm not sure I follow. In this specific case, there is no pre-existing `err` to return.

In order to make this code easier to read, I created some helper methods for inserting and removing things from channels. Eventually I'll probably extract these methods to a utility package, but for the time being I put them inside `wal.go` since there are no other callers.

```
// Push to a channel, returning an error if the context is cancelled before the value is pushed.
func interuptablePush[T any](ctx context.Context, ch chan T, value T) error {
	select {
	case <-ctx.Done():
		return fmt.Errorf("context cancelled: %w", ctx.Err())
	case ch <- value:
		return nil
	}
}

// Pull from a channel, returning an error if the context is cancelled before the value is pulled.
func interuptablePull[T any](ctx context.Context, ch <-chan T) (T, error) {
	var zero T
	select {
	case <-ctx.Done():
		return zero, fmt.Errorf("context cancelled: %w", ctx.Err())
	case value, ok := <-ch:
		if !ok {
			return zero, fmt.Errorf("channel closed")
		}
		return value, nil
	}
}
```

sei-db/wal/wal.go
```diff
@@ -49,249 +85,506 @@ type Config struct {
...
 		}
-		return err
-	default:
-		return nil
 	}
 }
 
+// Shut down the WAL. Sends a close request to the main loop so in-flight writes (and other work)
+// can complete before teardown. Idempotent.
+func (walLog *WAL[T]) Close() error {
```
bdchatham, line 444: nit: this isn't really thread safe although the current call pattern is safe. Could be more defensive/explicit to use sync.once. Two racing threads could try to put multiple entries into the channel. One will hang.

sei-db/wal/wal.go
```diff
@@ -49,135 +82,261 @@ type Config struct {
...
-		log:             log,
-		config:          config,
-		logger:          logger,
-		marshal:         marshal,
-		unmarshal:       unmarshal,
-		closeCh:         make(chan struct{}),
-		asyncWriteErrCh: make(chan error, 1),
+
+	bufferSize := config.WriteBufferSize
+	if config.WriteBufferSize == 0 {
```
blindchaser: nit: should this be config.WriteBufferSize <= 0? I think memiavl sometimes passes -1 to mean sync mode. With == 0, negative values can still flow into make(chan, size) and panic.

sei-db/wal/wal.go
```diff
@@ -49,249 +85,506 @@ type Config struct {
...
+
+	if walLog.asyncWrites {
+		// Do not wait for the write to be durable
+		return nil
+	}
+
+	select {
+	case _, ok := <-walLog.ctx.Done():
+		if !ok {
+			return fmt.Errorf("WAL was closed after write was submitted but before write was finalized, " +
```
bdchatham: I think you'd want to have a non-blocking check for the error channel here. Otherwise, you are swallowing any errors on a standard done() scenario.

sei-db/wal/wal.go
yzang2019: Yeah seems like it's breaking some unit test

sei-db/wal/wal.go
```diff
@@ -49,249 +85,506 @@ type Config struct {
...
+			for _, req := range requests {
+				req.errChan <- err
 			}
+			return
+		}
+		batch.Write(lastOffset+1, bz)
+		lastOffset++
+	}
 
+	if err := walLog.log.WriteBatch(batch); err != nil {
```
bdchatham, line 240: Seems problematic to return the error to all error channels since some of them may have been successfully written. In the case the wal implementation spans multiple segments then you could write sub-batches but return the error only for the last segment.

sei-db/wal/wal.go
cody-littley: Good catch, fixed

sei-db/wal/changelog.go
cody-littley: This has been reverted, config behavior now mimics old config behavior.

sei-db/ledger_db/parquet/wal.go
cody-littley: Yes, I intentionally left this for a latter PR. Ideally we'd be plumbing a context down to this level. But in the interest of keeping this PR's change set small, I'd prefer to leave that to a future PR.

sei-db/wal/wal.go
cody-littley: renamed to avoid shadowing

sei-db/wal/wal.go
cody-littley: Broke it out into a few helper methods.

sei-db/wal/wal.go
cody-littley: done

sei-db/wal/wal.go
cody-littley: Moved to top of file

sei-db/wal/wal.go
cody-littley: I think this is thread safe even if you have concurrent `Close()` calls. The first caller will send a message to the background thread. Other callers may become temporarily blocked. But once the background thread cancels the context, all other callers to `Close()` will stop waiting.

I added a unit test to verify this behavior.

```
	// Calling close lots of times shouldn't cause any problems.
	for i := 0; i < 10; i++ {
		require.NoError(t, changelog.Close())
	}
```

sei-db/wal/wal.go
cody-littley: This has been refactored since you initially made this comment.

sei-db/wal/wal.go
cody-littley: I'm not sure it is important to differentiate which writes became durable on disk and which ones didn't within a batch. In general, I'm not sure that this is a recoverable error if the WAL stops being able to accept new writes, since we're probably bumping into something like a full disk.

sei-db/wal/wal.go
bdchatham: Calling it serially is definitely no problem. 

The callout is that thread safety of Close() is implicitly dependent on properly handling `closeErrChan`. If you protect it differently you don't have to have this implicit dependency. 

Not a critical problem but wanted to mention the alternative.

sei-db/wal/wal.go
bdchatham: Hmm yeah. I don't have as much context as I'd like. Idk entirely how the WAL is used, but my main concern was the possibility of writing something to it but reporting to the caller that there was actually an error and what that caller will do with that signal

yzang2019: In sei-chain currently if we did fail write to the WAL, e.g. write() returns an error, it will panic and cause the whole process to exit

sei-db/wal/wal.go
cody-littley: The main challenge that `sync.Once()` doesn't resolve in this scenario is ensuring that `Close()` blocks all callers until everything is fully closed. We don't want to just block the first caller while things are being closed, we want to block all callers. So we'd still have to either block all callers on `closeErrChan`, or we'd have to introduce an additional synchronization primitive like a semaphore.

bdchatham: Yeah, it's a good point. The race between the removal and replacement was just the smell I wanted to dig into more. Appreciate the discussion on it

sei-db/wal/wal.go
bdchatham: Got it - thanks for the context and discussion on this

blindchaser (approved):

PR #2935, opened 2026-02-19
replacing abci protobuf types with plain golang types (part 1)

pompon0:
I'm working on reducing the API surface between consensus and Application for autobahn migration. This PR replaces protobuf types with plain golang types to reduce compatibility requirements. Will continue in the next pr.

### DIFF
--- a/sei-tendermint/abci/types/plain_types.go
+++ b/sei-tendermint/abci/types/plain_types.go
@@ -45,9 +45,9 @@ const (
 func (r ResponseOfferSnapshot_Result) String() string {
 	switch r {
 	case ResponseOfferSnapshot_UNKNOWN:
-		return "UNKNOWN"
+		return "UNKNOWN" //nolint:goconst
 	case ResponseOfferSnapshot_ACCEPT:
-		return "ACCEPT"
+		return "ACCEPT" //nolint:goconst
 	case ResponseOfferSnapshot_ABORT:
 		return "ABORT"
 	case ResponseOfferSnapshot_REJECT:
@@ -57,7 +57,7 @@ func (r ResponseOfferSnapshot_Result) String() string {
 	case ResponseOfferSnapshot_REJECT_SENDER:
 		return "REJECT_SENDER"
 	default:
-		return "UNKNOWN"
+		return "UNKNOWN" //nolint:goconst
 	}
 }
 
@@ -106,9 +106,9 @@ const (
 func (r ResponseApplySnapshotChunk_Result) String() string {
 	switch r {
 	case ResponseApplySnapshotChunk_UNKNOWN:
-		return "UNKNOWN"
+		return "UNKNOWN" //nolint:goconst
 	case ResponseApplySnapshotChunk_ACCEPT:
-		return "ACCEPT"
+		return "ACCEPT" //nolint:goconst
 	case ResponseApplySnapshotChunk_ABORT:
 		return "ABORT"
 	case ResponseApplySnapshotChunk_RETRY:
@@ -118,7 +118,7 @@ func (r ResponseApplySnapshotChunk_Result) String() string {
 	case ResponseApplySnapshotChunk_REJECT_SNAPSHOT:
 		return "REJECT_SNAPSHOT"
 	default:
-		return "UNKNOWN"
+		return "UNKNOWN" //nolint:goconst
 	}
 }
 
@@ -145,13 +145,13 @@ const (
 func (s ResponseProcessProposal_ProposalStatus) String() string {
 	switch s {
 	case ResponseProcessProposal_UNKNOWN:
-		return "UNKNOWN"
+		return "UNKNOWN" //nolint:goconst
 	case ResponseProcessProposal_ACCEPT:
-		return "ACCEPT"
+		return "ACCEPT" //nolint:goconst
 	case ResponseProcessProposal_REJECT:
 		return "REJECT"
 	default:
-		return "UNKNOWN"
+		return "UNKNOWN" //nolint:goconst
 	}
 }
 


sei-tendermint/abci/types/plain_types.go
```diff
@@ -0,0 +1,409 @@
...
+type ResponseOfferSnapshot_Result int32
+
+const (
+	ResponseOfferSnapshot_UNKNOWN       ResponseOfferSnapshot_Result = 0
+	ResponseOfferSnapshot_ACCEPT        ResponseOfferSnapshot_Result = 1
+	ResponseOfferSnapshot_ABORT         ResponseOfferSnapshot_Result = 2
+	ResponseOfferSnapshot_REJECT        ResponseOfferSnapshot_Result = 3
+	ResponseOfferSnapshot_REJECT_FORMAT ResponseOfferSnapshot_Result = 4
+	ResponseOfferSnapshot_REJECT_SENDER ResponseOfferSnapshot_Result = 5
+)
```
masih, line 43: Use `iota`?
```suggestion
const (
	ResponseOfferSnapshot_UNKNOWN ResponseOfferSnapshot_Result = iota
	ResponseOfferSnapshot_ACCEPT
	ResponseOfferSnapshot_ABORT
	ResponseOfferSnapshot_REJECT
	ResponseOfferSnapshot_REJECT_FORMAT
	ResponseOfferSnapshot_REJECT_SENDER
)
```

sei-tendermint/abci/types/plain_types.go
```diff
@@ -0,0 +1,409 @@
...
+	Result        ResponseApplySnapshotChunk_Result
+	RefetchChunks []uint32
+	RejectSenders []string
+}
+
+// ResponseApplySnapshotChunk_Result mirrors the historical protobuf enum.
+type ResponseApplySnapshotChunk_Result int32
+
+const (
+	ResponseApplySnapshotChunk_UNKNOWN         ResponseApplySnapshotChunk_Result = 0
```
masih: Ditto re using `iota`.

sei-tendermint/rpc/client/interface.go
```diff
@@ -28,8 +28,6 @@ import (
 	"github.com/sei-protocol/sei-chain/sei-tendermint/types"
 )
 
-//go:generate ../../scripts/mockery_generate.sh Client
-
```
masih, line 32: 🎉

sei-tendermint/abci/types/plain_types.go
```diff
@@ -0,0 +1,409 @@
...
+	TxResults             []*ExecTxResult
+	ValidatorUpdates      []ValidatorUpdate
+	ConsensusParamUpdates *tmproto.ConsensusParams
+}
+
+// ResponseProcessProposal_ProposalStatus mirrors the historical protobuf enum.
+type ResponseProcessProposal_ProposalStatus int32
+
+const (
+	ResponseProcessProposal_UNKNOWN ResponseProcessProposal_ProposalStatus = 0
```
masih: Ditto re using `iota`.

sei-tendermint/abci/types/plain_types.go
```diff
@@ -0,0 +1,409 @@
...
+
+// ResponseApplySnapshotChunk is the plain Go replacement for the legacy protobuf message.
+type ResponseApplySnapshotChunk struct {
+	Result        ResponseApplySnapshotChunk_Result
+	RefetchChunks []uint32
+	RejectSenders []string
+}
+
+// ResponseApplySnapshotChunk_Result mirrors the historical protobuf enum.
+type ResponseApplySnapshotChunk_Result int32
```
masih, line 95: Implement `Stringer` like we have for status?

sei-tendermint/abci/types/plain_types.go
```diff
@@ -0,0 +1,409 @@
...
+	AppHash  []byte
+}
+
+// ResponseOfferSnapshot is the plain Go replacement for the legacy protobuf message.
+type ResponseOfferSnapshot struct {
+	Result ResponseOfferSnapshot_Result
+}
+
+// ResponseOfferSnapshot_Result mirrors the historical protobuf enum.
+type ResponseOfferSnapshot_Result int32
```
masih, line 34: Implement `Stringer` like we have for status?

sei-tendermint/abci/types/plain_types.go
```diff
@@ -0,0 +1,409 @@
...
+		return "ACCEPT"
+	case ResponseProcessProposal_REJECT:
+		return "REJECT"
+	default:
+		return "UNKNOWN"
+	}
+}
+
+// RequestProcessProposal is the plain Go replacement for the legacy protobuf message.
+type RequestProcessProposal struct {
```
masih, line 161: You can extract the common fileds across this type and `RequestFinalizeBlock` into an _unexported_ type, e.g. blockRequestBase to reduce boiler plating. Something like this:


```
type blockRequestBase struct {
    Txs                   [][]byte
    ByzantineValidators   []Misbehavior
    Hash                  []byte
    Height                int64
    Time                  time.Time
    NextValidatorsHash    []byte
    ProposerAddress       []byte
    AppHash               []byte
    ValidatorsHash        []byte
    ConsensusHash         []byte
    DataHash              []byte
    EvidenceHash          []byte
    LastBlockHash         []byte
    LastBlockPartSetTotal int64
    LastBlockPartSetHash  []byte
    LastCommitHash        []byte
    LastResultsHash       []byte
}
```

Then these types would become:
```
type RequestProcessProposal struct {
    blockRequestBase
    ProposedLastCommit CommitInfo
}

type RequestFinalizeBlock struct {
    blockRequestBase
    DecidedLastCommit CommitInfo
}
```
Which makes semantic differences obvious/easy to see.

sei-tendermint/abci/types/plain_types.go
```diff
@@ -0,0 +1,409 @@
...
+	ValidatorsHash        []byte
+	ConsensusHash         []byte
+	DataHash              []byte
+	EvidenceHash          []byte
+	LastBlockHash         []byte
+	LastBlockPartSetTotal int64
+	LastBlockPartSetHash  []byte
+	LastCommitHash        []byte
+	LastResultsHash       []byte
+}
```
masih, line 180: Thinking out loud: Do we need validator for these?

sei-tendermint/abci/types/plain_types.go
```diff
@@ -0,0 +1,409 @@
+package types
+
+import (
+	"time"
+
+	tmproto "github.com/sei-protocol/sei-chain/sei-tendermint/proto/tendermint/types"
+)
+
+// RequestLoadLatest is the plain Go replacement for the legacy protobuf message.
```
masih: Would be nice to also document the purpose of the type as well as its origin. Ditto for the rest.

sei-tendermint/abci/types/plain_types.go
```diff
@@ -0,0 +1,409 @@
...
+	DataHash              []byte
+	EvidenceHash          []byte
+	LastBlockHash         []byte
+	LastBlockPartSetTotal int64
+	LastBlockPartSetHash  []byte
+	LastCommitHash        []byte
+	LastResultsHash       []byte
+}
+
+func (m *RequestProcessProposal) GetTxs() [][]byte {
```
masih, line 182: Beyond the scope of this PR:

Do we still need these getters? Aside from >400 lies of bilerplate they add, these have a hidden cost in that they encourage pointer passing. They make sense for a one-size-cut-all Proto generator but for hand-rolled types we could be more concise/robust.

I understand that removing these needs a careful review to make sure we are not missing any nil checks. But we all will be better off for it: in that it forces us to validate earlier, and interrupt the execution flow when it makes no sense to proceed. e.g. when `*RequestProcessProposal` is nil.

sei-tendermint/abci/example/kvstore/kvstore.go
```diff
@@ -408,13 +409,13 @@ func (app *Application) updateValidator(v types.ValidatorUpdate) *types.ExecTxRe
 		delete(app.valAddrToPubKeyMap, string(pubkey.Address()))
 	} else {
 		// add or update validator
-		value := bytes.NewBuffer(make([]byte, 0))
-		if err := types.WriteMessage(&v, value); err != nil {
+		bz, err := proto.Marshal(&v)
```
masih, line 412: This is a breaking change, i.e. not backward compatible. Because types.WriteMessage [was varint delimited](https://github.com/sei-protocol/sei-chain/blob/55c02544fbb8f12f43c248a02a78df1b1e4995dc/sei-tendermint/internal/libs/protoio/writer.go#L55).

Do we want to break backward compatibility?

sei-tendermint/abci/types/plain_types.go
pompon0: the plan is to use block header, because that's where all these fields come from, and it doesn't seem feasible to make block header an implementation detail any time soon (there are APIs and storage depending on it). Refactoring this will be the next step.

sei-tendermint/abci/types/plain_types.go
pompon0: wdym?

sei-tendermint/abci/types/plain_types.go
pompon0: I don't like the getters either, although they do not block my goal of reducing the api surface. +1 for getting rid of them though

sei-tendermint/abci/example/kvstore/kvstore.go
pompon0: this is an "example" app, so no compatibility guarantees are needed (as long as it works).

sei-tendermint/abci/types/plain_types.go
masih: I mean: do any of the types here need validation? We might have validation already somewhere. Checking that we do validate fields if they come off the wire.

pompon0: these are passed from consensus to application. Application does not interpret most of these, doesn't even access many of those, but still persists it at places. Presumably so that the header hash can be computed by light clients or sth.

PR #2936, opened 2026-02-19
Add FlatKV database to the benchmarking utility.

cody-littley:

Ticket: https://linear.app/seilabs/issue/STO-358/state-db-go-bench-should-testing-all-sc-backends

Add support to the DB benchmarking utility for different DB implementations.

I ran the benchmarks locally and verified that they worked as expected.


sei-db/state_db/bench/wrappers/db_implementations.go
```diff
@@ -0,0 +1,106 @@
...
+	return NewMemIAVLWrapper(cs)
+}
+
+func newFlatKVCommitStore(b *testing.B) DBWrapper {
+	b.Helper()
+	dir := b.TempDir()
+	cfg := flatkv.DefaultConfig()
+	cs := flatkv.NewCommitStore(dir, logger.NewNopLogger(), cfg)
+
+	_, err := cs.LoadVersion(0, false)
```
yzang2019: Need to remove the extra arg

sei-db/state_db/bench/bench_sc_test.go
```diff
@@ -1,52 +1,194 @@
...
 
-// BenchmarkWriteWithDifferentBlockSize tests throughput with fixed 1M total keys,
+	"github.com/sei-protocol/sei-chain/sei-db/state_db/bench/wrappers"
+)
+
+// Using MemIAVL, Tests throughput with fixed total keys,
 // varying keysPerBlock and numBlocks to find optimal block size.
-func BenchmarkWriteWithDifferentBlockSize(b *testing.B) {
+func BenchmarkMemIAVLWriteWithDifferentBlockSize(b *testing.B) {
 	const totalKeys int64 = 1_000_000
```
yzang2019: Let's change this to 100k as well?

yzang2019: Oh nvm, it seems 100k isn't accurate enough, we probably wanna test more

yzang2019: For the total testing time perspective, maybe still reduce to 100k level, and use a smaller batch size like 1,2,10,20 and up till 100,200?

sei-db/state_db/bench/wrappers/db_implementations.go
cody-littley: fixed

sei-db/state_db/bench/bench_sc_test.go
cody-littley: change made

yzang2019:
> seems like the current benchmarks have high variance, I get different numbers each time. Is there a way to reduce or eliminate this variability?

We can have a separate PR to address that issue, I don't think it's a blocker for this PR

PR #2940, opened 2026-02-20
flaky test fix

pompon0:
Added more precise handling of background tasks in tests which are allowed to fail during test cleanup.


sei-tendermint/internal/p2p/router_test.go
```diff
@@ -312,7 +312,7 @@ func TestRouter_FilterByIP(t *testing.T) {
 			if err != nil {
 				return fmt.Errorf("peerTransport.dial(): %w", err)
 			}
-			s.SpawnBg(func() error { return tcpConn.Run(ctx) })
+			s.SpawnBg(func() error { return utils.IgnoreCancel(tcpConn.Run(ctx)) })
```
pompon0, line 315: fyi, this one is intentional, because peer is not expected to end the connection first.

sei-tendermint/libs/utils/tcp/tcp.go
```diff
@@ -69,7 +69,7 @@ func (c Conn) Flush(_ context.Context) error { return nil }
 func (c Conn) Close()                        { _ = c.conn.Close() }
 
 func (c Conn) Run(ctx context.Context) error {
-	return utils.IgnoreCancel(scope.Run(ctx, func(ctx context.Context, s scope.Scope) error {
+	return scope.Run(ctx, func(ctx context.Context, s scope.Scope) error {
```
wen-coding, line 72: Why are we removing this IgnoreCancel?

pompon0: to handle the error in a more robust way at the call site

sei-tendermint/libs/utils/testonly.go
```diff
@@ -219,3 +220,14 @@ func (c *ProtoConv[T, P]) Test(want T) error {
 	}
 	return TestDiff(want, got)
 }
+
+// IgnoreAfterCancel silently drops the error if the context is already canceled.
+// Should be used for background tasks in tests, which cannot be guaranteed to exit gracefully.
+// For example - if you have a tcp connection, then during cleanup one end will disconnect faster than the other,
+// causing a race condition between context cancellation and disconnection error.
+func IgnoreAfterCancel(ctx context.Context, err error) error {
+	if ctx.Err() != nil {
```
masih, line 229: > // IgnoreAfterCancel silently drops the error if the context is already canceled.

Does it? It drops any old error in context. There could be different types of error here.

I am sorry I sound like a broken record: These utility functions are anti-pattern: they add more LoC than they save; and more cognitive load than it should take to read simple go code.

pompon0: > Does it? It drops any old error in context. There could be different types of error here.

True, context may contain various errors as well, I use word "canceled" to not say "context is done" or "context.Done() channel is closed" because it sounds weird and I've never heard anyone saying this. Also given that context is propagated to arbitrary subroutines, allowing to terminate/cancell/close done channel of context with an error different Canceled was a design mistake, because it makes literally almost ALL of golang code need to handle arbitrary errors they are not even aware of.

> These utility functions are anti-pattern: they add more LoC than they save; and more cognitive load than it should take to read simple go code.

Well, that's like your opinion, man. I find your select clauses hardly readable, if you really would like to know.

PR #2941, opened 2026-02-20
bugfix: make router load addresses from peerdb

pompon0:
Even though router was correctly storing the addresses of connected peers to db, it was not loading them back on startup. Therefore router was behaving as if peerdb did not exist in the first place. I have added a test for the fix.

### DIFF
--- a/sei-tendermint/internal/p2p/router_test.go
+++ b/sei-tendermint/internal/p2p/router_test.go
@@ -308,7 +308,7 @@ func TestRouter_FilterByIP(t *testing.T) {
 			if err != nil {
 				return fmt.Errorf("peerTransport.dial(): %w", err)
 			}
-			s.SpawnBg(func() error { return tcpConn.Run(ctx) })
+			s.SpawnBg(func() error { return utils.IgnoreCancel(tcpConn.Run(ctx)) })
 			if _, _, err := r2.handshakeV2(ctx, tcpConn, utils.Some(addr)); err != nil {
 				return fmt.Errorf("handshake(): %w", err)
 			}
@@ -343,8 +343,6 @@ func TestRouter_FilterByIP(t *testing.T) {
 		t.Fatal(err)
 	}
 }
-
-func ignore(error) error { return nil }
 
 func blindHandshake(ctx context.Context, c tcp.Conn, key NodeSecretKey, info types.NodeInfo) error {
 	return utils.IgnoreCancel(scope.Run(ctx, func(ctx context.Context, s scope.Scope) error {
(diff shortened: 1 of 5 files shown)


sei-tendermint/internal/p2p/router.go
```diff
@@ -76,6 +76,11 @@ func NewRouter(
 	if err != nil {
 		return nil, fmt.Errorf("newPeerDB(): %w", err)
 	}
+	for addr := range peerDB.All() {
+		if err := peerManager.AddAddrs(utils.Slice(addr)); err != nil {
+			logger.Error("peerDB: bad address", "addr", addr.String(), "err", err)
```
wen-coding, line 81: How often do the addresses change? Is it possible that these addresses are outdated after you restart?

sei-tendermint/internal/p2p/router.go
```diff
@@ -300,12 +305,16 @@ func (r *Router) dialPeersRoutine(ctx context.Context) error {
...
+	storeInterval := r.options.peerStoreInterval()
 	for {
-		for db := range r.peerDB.Lock() {
+		for db, ctrl := range r.peerDB.Lock() {
 			// Mark connections as still available.
 			now := time.Now()
-			for _, conn := range r.peerManager.Conns().All() {
+			conns := r.peerManager.Conns()
+			if conns.Len() > 0 {
+				ctrl.Updated()
```
wen-coding, line 315: nit: should we call this only if any address changed?

sei-tendermint/internal/p2p/router_test.go
```diff
@@ -747,3 +743,78 @@ func TestRouter_DontSendOnInvalidChannel(t *testing.T) {
...
+		t.Logf("start the first node")
+		r := makeRouter(logger, rng)
+		addr := TestAddress(r)
+		s.SpawnBg(func() error { return utils.IgnoreCancel(r.Run(ctx)) })
+
+		db := dbm.NewMemDB()
+		key := makeKey(rng)
+		info := makeInfo(key)
+		options := makeRouterOptions()
+		options.PeerStoreInterval = utils.Some(time.Second)
```
wen-coding, line 761: nit: Could we set it to even a bit lower to make the test run faster?

sei-tendermint/internal/p2p/router.go
yzang2019, line 81: If we detect bad address, shall we remove them from peer db?

pompon0: they might be outdated. We don't know how long the downtime was. We also don't know if the peers are down. We make no assumptions here. The amount of stored addresses is small. It is not problem to just dial them all.

sei-tendermint/internal/p2p/router.go
pompon0: this Watch was added only for the test purposes. It is fine to call Updated even of nothing changed. The point is to call it rarely enough to not waste cpu cycles. Here we call update ~10s by default, which is fine.

sei-tendermint/internal/p2p/router_test.go
pompon0: we could, currently 1s is nowhere close the latency of many other tests, so it shouldn't hurt.

sei-tendermint/internal/p2p/router.go
pompon0: nah, peer db is a cyclic buffer in a sense, it will clean itself

yzang2019: Yeah, the peer can be outdated but usually wouldn't with just a restart. Even if there are outdated peers, those will get evicted from the DB and be replaced by valid peers right?

pompon0: yes, we only insert entries to db if we successfully connect to a peer and evict entries of peers that we haven't seen the longest

PR #2946, opened 2026-02-21
Fix flaky `TestClientMethodCalls` by using eventual time assertion

masih:
Switch to using eventual arrival of events instead of a fixed timeout.

Also refine the coverage config to avoid miss leading patch coverage diff on small changes.

Flaked on [main](https://github.com/sei-protocol/sei-chain/actions/runs/22244063952/job/64354114511).

### DIFF
--- a/sei-tendermint/rpc/jsonrpc/jsonrpc_test.go
+++ b/sei-tendermint/rpc/jsonrpc/jsonrpc_test.go
@@ -8,7 +8,7 @@ import (
 	"errors"
 	mrand "math/rand"
 	"net/http"
-	"os/exec"
+	"path/filepath"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


bdchatham (approved):

sei-tendermint/rpc/client/helpers.go
```diff
@@ -57,7 +57,7 @@ func WaitForHeight(ctx context.Context, c StatusClient, h int64, waiter Waiter)
 // until ctx ends. It reports an error if ctx ends before a matching event is
 // received.
 func WaitForOneEvent(ctx context.Context, c EventsClient, query string) (types.EventData, error) {
-	for {
+	for ctx.Err() == nil {
```
philipsu522, line 60: do we need to timeout after a while to avoid infinite running?

masih: Yes, and we do [here](https://github.com/sei-protocol/sei-chain/pull/2946/changes#diff-6928f36338e2c6212f68d172ff1d320945bb1a97f84ddcd0f1309617f9f64979R36). i.e.the context passed in here has a timeout: `attemptCtx, cancel := context.WithTimeout(ctx, waitForEventAttemptTimeout)`

PR #2949, opened 2026-02-22
Do not support historical proof on SeiDB

yzang2019:
Problem: 
SeiDB memIAVL is very inefficient at serving historical queries when state is large, since it needs to call LoadVersion each time to open the DB at an old height in order to serve a single query. 

Solution:
Currently, as a short term mitigation, we will stop historical proof support until we have a better solution. Any historical proof request should probably goes to an IAVL archive node, or once Giga is out, we will find some other solution for how to efficiently serve this.

### DIFF
--- a/sei-cosmos/storev2/rootmulti/store_test.go
+++ b/sei-cosmos/storev2/rootmulti/store_test.go
@@ -85,15 +85,25 @@ func TestSCSS_WriteAndHistoricalRead(t *testing.T) {
 	require.EqualValues(t, 0, resp.Code)
 	require.Equal(t, valV1, resp.Value)
 
-	// Query API with proof at v1 should still return v1 (served by SC historical)
+	// Query API with proof at historical height should fail (only latest supported)
 	resp = store.Query(abci.RequestQuery{
 		Path:   "/store1/key",
 		Data:   keyBytes,
 		Height: c1.Version,
 		Prove:  true,
 	})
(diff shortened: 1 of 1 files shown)


sei-cosmos/storev2/rootmulti/store.go
```diff
@@ -556,35 +556,34 @@ func (rs *Store) Query(req abci.RequestQuery) abci.ResponseQuery {
...
+			return sdkerrors.QueryResult(errors.Wrap(sdkerrors.ErrInvalidRequest, "proof is unexpectedly empty; ensure height has not been pruned"))
+		}
 	}
 
-	// trim the path and execute the query
-	req.Path = subPath
-	res := store.Query(req)
+	// We don't support historical proofs until we have a better solution
+	rs.logger.Error("unable to serve ABCI request with proof=true", "request_height", req.Height, "latest_height", rs.scStore.Version())
+	return sdkerrors.QueryResult(errors.Wrap(sdkerrors.ErrInvalidRequest, "only support proof on latest height"))
```
masih:
```suggestion
	return sdkerrors.QueryResult(errors.Wrap(sdkerrors.ErrInvalidRequest, "proof only supported on latest height"))
```

philipsu522 (approved):

blindchaser (approved):

yzang2019:
Not moving forward with this approach since it is too risky, might break TM light client such as IBC

PR #2956, opened 2026-02-23
CompositeStateStore: Add RocksDB backend support 

Kbhat1:

- Adds RocksDB backend support for and change default PebbleDB in the composite state store
- Introduces `EVMDBEngine` interface + backend registry so both Cosmos_SS and EVM_SS respect `ssConfig.Backend` for engine selection
- `NewCompositeStateStore` now accepts a pre-created cosmos store from the existing backend registry instead of importing `pebbledb/mvcc` directly

- Unit tests
- verifying on node


sei-db/state_db/ss/evm/db_rocksdb.go
```diff
@@ -0,0 +1,340 @@
...
+
+	return evmDB, nil
+}
+
+func newRocksDBEVMOpts() *grocksdb.Options {
+	opts := grocksdb.NewDefaultOptions()
+	opts.SetCreateIfMissing(true)
+	opts.SetComparator(createTimestampComparator())
+	opts.IncreaseParallelism(runtime.NumCPU())
+	opts.OptimizeLevelStyleCompaction(512 * 1024 * 1024)
```
Kbhat1, line 108: Will have to tune these opts more

sei-db/state_db/ss/composite/store.go
```diff
@@ -24,32 +23,30 @@ import (
...
 
-// NewCompositeStateStore creates a new composite state store that manages both Cosmos_SS and EVM_SS.
-// It initializes both stores internally and starts pruning on the composite store.
+// NewCompositeStateStore wraps a pre-created Cosmos store with optional EVM stores.
+// The cosmosStore is created by the caller (ss.NewStateStore) using the backend registry,
+// so both PebbleDB and RocksDB are supported without a direct import here.
 // EVM stores are opened when ssConfig.EVMEnabled() returns true (derived from WriteMode/ReadMode).
+// The same backend (ssConfig.Backend) is used for EVM sub-databases.
 func NewCompositeStateStore(
+	cosmosStore types.StateStore,
```
yzang2019, line 41: Having to pass in cosmosStore seems a bit weird to me, why can't we keep the previous interface but just move backend to top level ssconfig?

sei-db/state_db/ss/evm/db_rocksdb.go
```diff
@@ -0,0 +1,340 @@
...
+	opts.SetBottommostCompressionOptionsZstdMaxTrainBytes(compressOpts.MaxDictBytes*100, true)
+
+	return opts
+}
+
+func (db *EVMDatabaseRocksDB) Get(key []byte, version int64) ([]byte, error) {
+	if version < db.earliestVersion.Load() {
+		return nil, nil
+	}
+	slice, err := db.storage.GetCF(rocksNewTSReadOpts(version), db.cfHandle, key)
```
blindchaser: rocksNewTSReadOpts is called but not destroyed. 

consider:
`ro := rocksNewTSReadOpts(version)
 defer ro.Destroy()`

sei-db/state_db/ss/evm/db_rocksdb.go
```diff
@@ -0,0 +1,340 @@
...
+	binary.LittleEndian.PutUint64(ts[:], uint64(version))
+
+	batch := grocksdb.NewWriteBatch()
+	batch.DeleteCFWithTS(db.cfHandle, key, ts[:])
+	defer batch.Destroy()
+
+	if err := db.storage.Write(rocksDefaultWriteOpts, batch); err != nil {
+		return err
+	}
+	return nil
```
blindchaser, line 190: it loosk PebbleDB's Delete calls updateLatestVersion(version), but the RocksDB implementation drops the version update, is this expected?

sei-db/state_db/ss/evm/db_rocksdb.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+	db.tsLow = tsLow
+	return db.SetEarliestVersion(tsLow)
+}
+
+func (db *EVMDatabaseRocksDB) Close() error {
+	if db.storage == nil {
+		return nil
+	}
+	db.cfHandle = nil
```
blindchaser: can you double check if `db.cfHandle=nil` is good to release the resource as it's C-allocated. may need a `db.cfHandle.Destroy()` before `db.cfHandle = nil`

sei-db/state_db/ss/evm/db_rocksdb.go
```diff
@@ -0,0 +1,340 @@
...
+}
+
+// OpenRocksDB opens a RocksDB instance for one EVM sub-database.
+func OpenRocksDB(dir string, storeType EVMStoreType) (*EVMDatabaseRocksDB, error) {
+	dbPath := filepath.Join(dir, StoreTypeName(storeType))
+
+	defaultOpts := grocksdb.NewDefaultOptions()
+	defaultOpts.SetCreateIfMissing(true)
+	defaultOpts.SetCreateIfMissingColumnFamilies(true)
+
```
blindchaser, line 58: same here for c resources leaks. need 

defer defaultOpts.Destroy()
defer cfOpts.Destroy()

sei-db/state_db/ss/evm/store.go
```diff
@@ -18,62 +18,71 @@ type dbWrite struct {
...
 }
 
-// EVMStateStore manages multiple EVMDatabase instances, one per EVM data type.
+// EVMStateStore manages multiple EVMDBEngine instances, one per EVM data type.
 // Each database has an optional background goroutine for async writes.
 type EVMStateStore struct {
-	databases map[EVMStoreType]*EVMDatabase
+	databases map[EVMStoreType]EVMDBEngine
 	dir       string
+	backend   string
```
blindchaser: how does the backend field being used?

sei-db/state_db/ss/evm/db_rocksdb.go
```diff
@@ -0,0 +1,340 @@
...
+var (
+	rocksDefaultWriteOpts = grocksdb.NewDefaultWriteOptions()
+	rocksDefaultReadOpts  = grocksdb.NewDefaultReadOptions()
+)
+
+// EVMDatabaseRocksDB is a RocksDB-backed versioned KV store for a single EVM data type.
+// Uses user-defined timestamps for MVCC versioning and a column family for state data.
+type EVMDatabaseRocksDB struct {
+	storeType EVMStoreType
```
blindchaser: how does the storeType being used?

jewei1997 (approved):
similar comments to Yiren

once those are fixed, lgtm

sei-db/state_db/ss/evm/db_rocksdb.go
Kbhat1: Updated

Kbhat1: Needs to be updated on original rocksdb as well will make a PR

sei-db/state_db/ss/composite/store.go
Kbhat1: The `ss` package imports `composite`, so composite can't import ss back to use the backend registry (would be a circular import). The old code addressed by hardcoding pebbledb/mvcc directly, but now that we support both PebbleDB and RocksDB (with RocksDB behind a build tag), we need the registry-based selection that lives in ss. Passing in the pre-created store is the standard dependency injection pattern to avoid the cycle

yzang2019: Let me think about whether there's other way to address this, thanks for the explanation!

sei-db/state_db/ss/evm/db_rocksdb.go
```diff
@@ -0,0 +1,351 @@
...
+)
+
+var (
+	rocksDefaultWriteOpts = grocksdb.NewDefaultWriteOptions()
+	rocksDefaultReadOpts  = grocksdb.NewDefaultReadOptions()
+)
+
+// EVMDatabaseRocksDB is a RocksDB-backed versioned KV store for a single EVM data type.
+// Uses user-defined timestamps for MVCC versioning and a column family for state data.
+type EVMDatabaseRocksDB struct {
```
yzang2019, line 41: Why do we implement a separate RocksDB instead of reusing the existing RocksDB in db engine?

PR #2957, opened 2026-02-23
[giga] remove Snapshot() call in Prepare()

codchen:
The Snapshot() call would return a `rev` number for potential `Revert(rev)` calls to revert to. However the `Snapshot` call in `Prepare` has its `rev` value discarded, so it isn't useful from a reverting perspective. The other side effects, like creating another nested layer of CacheKV, is also meaningless under the context of EVM (there will just be a layer of CacheKV with zero dirty values). So it can be removed in theory.

Removing it resulted in a 1k increase in theoretical tps (from 20k to 21k) 

existing tests should still stand

### DIFF
--- a/giga/deps/xevm/state/statedb.go
+++ b/giga/deps/xevm/state/statedb.go
@@ -54,9 +54,7 @@ func NewDBImpl(ctx sdk.Context, k EVMKeeper, simulation bool) *DBImpl {
 		journal:            []journalEntry{},
 		coinbaseEvmAddress: feeCollector,
 	}
-	if k.ShouldUseRegularStore() {
-		s.Snapshot()
-	}
+	s.Snapshot() // take an initial snapshot for GetCommitted
 	return s
 }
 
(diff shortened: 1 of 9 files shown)


giga/deps/xevm/state/statedb.go
```diff
@@ -54,7 +54,9 @@ func NewDBImpl(ctx sdk.Context, k EVMKeeper, simulation bool) *DBImpl {
 		journal:            []journalEntry{},
 		coinbaseEvmAddress: feeCollector,
 	}
-	s.Snapshot() // take an initial snapshot for GetCommitted
+	if k.ShouldUseRegularStore() {
```
stevenlanders: eventually for performance/clarity it might be good to have a way to only check these kind of flags initially when setting handlers (like set a s.Snapshot handler at init/creation).  This is fine - just thinking out loud.

codchen: @stevenlanders i think that change was due to my stale local main. Rebased and this PR is just a one-liner

PR #2958, opened 2026-02-23
Add metrics for historical proof success/failure rate

yzang2019:
Add metrics to indicate historical proof success/failure rate due to rate limiting.

### DIFF
--- a/sei-cosmos/storev2/rootmulti/store.go
+++ b/sei-cosmos/storev2/rootmulti/store.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"math"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -598,17 +599,19 @@ func (rs *Store) Query(req abci.RequestQuery) abci.ResponseQuery {
 		// historical path (this is where RPC pressure happens)
 		if err := rs.tryAcquireHistProofPermit(); err != nil {
 			rs.logger.Debug("Failed to acquire historical proof permit", "err", err)
-			telemetry.IncrCounterWithLabels([]string{"historical", "proof"},
+			telemetry.IncrCounterWithLabels([]string{"historical", "abci", "query"},
 				1,
 				[]metrics.Label{
 					telemetry.NewLabel("success", "false"),
+					telemetry.NewLabel("proof", strconv.FormatBool(needProof)),
 				})
 			return sdkerrors.QueryResult(err)
 		} else {
-			telemetry.IncrCounterWithLabels([]string{"historical", "proof"},
+			telemetry.IncrCounterWithLabels([]string{"historical", "abci", "query"},
 				1,
 				[]metrics.Label{
 					telemetry.NewLabel("success", "true"),
+					telemetry.NewLabel("proof", strconv.FormatBool(needProof)),
 				})
 		}
 		defer rs.releaseHistProofPermit()


sei-cosmos/storev2/rootmulti/store.go
```diff
@@ -615,27 +620,23 @@ func (rs *Store) Query(req abci.RequestQuery) abci.ResponseQuery {
...
-	}
-
-	if !needProof {
+	// If underlying query failed (e.g. invalid height/path) or doesn' need proof, return as-is.
+	if res.Code != 0 || !needProof {
 		return res
 	}
 
 	// Must have proof ops from underlying store query before appending commit proof.
-	if res.ProofOps == nil {
```
Kbhat1, line 628: Was removing this intentional? We check res.ProofOps.Ops right after this before checking res.ProofOps is not nil

yzang2019: We didn't remove this, just combining it with the other if to remove duplicate code

sei-cosmos/storev2/rootmulti/store.go
```diff
@@ -615,27 +620,23 @@ func (rs *Store) Query(req abci.RequestQuery) abci.ResponseQuery {
...
-		return sdkerrors.QueryResult(errors.Wrap(sdkerrors.ErrInvalidRequest, "proof is unexpectedly empty; ensure height has not been pruned"))
-	}
-
 	if commitInfo != nil {
 		res.ProofOps.Ops = append(res.ProofOps.Ops, commitInfo.ProofOp(storeName))
 	}
-
-	if len(res.ProofOps.Ops) == 0 {
+	if res.ProofOps == nil || len(res.ProofOps.Ops) == 0 {
```
masih: Hmm, this is a panic waiting to happen 😱 : Note that in line 630, couple of lines above, we are appending to `ProofOps.Ops`. But the nil check happens _after_ it.

sei-cosmos/storev2/rootmulti/store.go
```diff
@@ -615,27 +620,23 @@ func (rs *Store) Query(req abci.RequestQuery) abci.ResponseQuery {
...
 		return sdkerrors.QueryResult(errors.Wrap(sdkerrors.ErrInvalidRequest, "proof is unexpectedly empty; ensure height has not been pruned"))
 	}
+	telemetry.IncrCounterWithLabels([]string{"historical", "proof"},
```
masih: This includes latest as well as historical. The label suggests it should measure historical only. As is, i think this is missleading. We should either update the label, or exclude latest.

sei-cosmos/storev2/rootmulti/store.go
yzang2019: If you check the original code, I think that should be the correct order: https://github.com/sei-protocol/sei-chain/blob/v6.3.2/sei-cosmos/storev2/rootmulti/store.go#L568

sei-cosmos/storev2/rootmulti/store.go
yzang2019: Good point

sei-cosmos/storev2/rootmulti/store.go
masih: I don't follow, sorry. If commitInfo is non nil, we [end up appending to `ProofOps`](https://github.com/sei-protocol/sei-chain/blob/v6.3.2/sei-cosmos/storev2/rootmulti/store.go#L570) without having checked if it is non nil. 

Prior to this PR, we correctly checked that it is non nil _before_ appending anything, whether commitInfo was nil or not. 

Does that make sense? What am I missing?

Kbhat1: Yea had the same comment above

sei-cosmos/storev2/rootmulti/store.go
Kbhat1: But we check res.ProofOps.Ops right after this before checking res.ProofOps is not nil

sei-cosmos/storev2/rootmulti/store.go
yzang2019: Yup, resolved on slack

sei-cosmos/storev2/rootmulti/store.go
```diff
@@ -598,7 +598,18 @@ func (rs *Store) Query(req abci.RequestQuery) abci.ResponseQuery {
...
+				[]metrics.Label{
+					telemetry.NewLabel("success", "false"),
+				})
 			return sdkerrors.QueryResult(err)
+		} else {
+			telemetry.IncrCounterWithLabels([]string{"historical", "proof"},
+				1,
+				[]metrics.Label{
+					telemetry.NewLabel("success", "true"),
+				})
```
masih, line 615: These metrics still end up ticking regardless of whether the query was ~historical~ with proof=true or not I'm afraid.

yzang2019: This is under the section of historical query already, can you double check?

masih: Apologies for confusion; corrected my comment.

masih: Discussed in slack; path forward:
* rename metric name to not have "proof" in it since it can capture info for non-proof calls.
* tag the metric by "proof" instead, which may be true or false.

PR #2963, opened 2026-02-23
fix(giga): bail on wrong nonce as v2 does

arajasek:

Reopening #2877 against main

fixes mismatch of [this block](https://seiscan.io/txs?block=193119398). there are 2 txs (don't show up on seiscan) that have invalid nonces:
```
V2's log: "nonce too high: address 0x6a9dE37D180C56b7E9362d7688ACa8B7acEb425C, tx: 23213 state: 23211"

V2's log: "nonce too high: address 0x4A7d9e41340d1dba4d157fdE37Ed1697E42CBC7b, tx: 7423188 state: 7423187"
```

In this case we just bail WITHOUT bumping the nonce.

tests pass, get us past the block in question.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1747,6 +1747,24 @@ func (app *App) executeEVMTxWithGigaExecutor(ctx sdk.Context, msg *evmtypes.MsgE
 	_, isAssociated := app.GigaEvmKeeper.GetEVMAddress(ctx, seiAddr)
 
 	// ============================================================================
+	// Nonce validation (mirrors V2's ante handler check in x/evm/ante/sig.go)
+	// V2 rejects with ErrWrongSequence if txNonce != expectedNonce, with NO state changes.
+	// ============================================================================
+	expectedNonce := app.GigaEvmKeeper.GetNonce(ctx, sender)
+	txNonce := ethTx.Nonce()
+	if txNonce != expectedNonce {
+		nonceDirection := "too high"
+		if txNonce < expectedNonce {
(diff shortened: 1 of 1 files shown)


app/app.go
```diff
@@ -1755,6 +1755,28 @@ func (app *App) executeEVMTxWithGigaExecutor(ctx sdk.Context, msg *evmtypes.MsgE
...
 	}
 
+	// ============================================================================
+	// Nonce validation (mirrors V2's ante handler check in x/evm/ante/sig.go)
+	// V2 rejects with ErrWrongSequence if txNonce != expectedNonce, with NO state changes.
+	// ============================================================================
+	expectedNonce := app.GigaEvmKeeper.GetNonce(ctx, sender)
+	txNonce := ethTx.Nonce()
+	if txNonce != expectedNonce {
+		// Calculate intrinsic gas for reporting (V2 reports this as gasUsed)
```
codchen: can you link where this happens for v2? from what I see the v2 ante handler would just exit with 0 gas used.

not blocking though since gas used no longer affects consensus

arajasek: Thanks, you're right. I confused myself with this case below when it _does_ report a value: https://github.com/sei-protocol/sei-chain/pull/2963/changes#diff-0f1d2976054440336a576d47a44a37b80cdf6701dd9113012bce0e3c425819b7R1861

arajasek: (Fixed.)

PR #2970, opened 2026-02-24
Add snapshot import for Giga Live State

yzang2019:
This PR add feature to import snapshot
- Add feature for FlatKV import from snapshot
- Make go benchmark able to load from local snapshot

### DIFF
--- a/sei-db/state_db/sc/composite/store.go
+++ b/sei-db/state_db/sc/composite/store.go
@@ -262,9 +262,12 @@ func (cs *CompositeCommitStore) Importer(version int64) (types.Importer, error) 
 	if err != nil {
 		return nil, err
 	}
-	evmImporter, err := cs.evmCommitter.Importer(version)
-	if err != nil {
-		return nil, err
+	var evmImporter types.Importer
+	if cs.evmCommitter != nil {
+		evmImporter, err = cs.evmCommitter.Importer(version)
+		if err != nil {
+			return nil, err
+		}
 	}
 	compositeImporter := NewImporter(cosmosImporter, evmImporter)
 	return compositeImporter, nil


sei-db/state_db/bench/wrappers/db_implementations.go
cody-littley, line 1: It's funny, I have almost this exact change set in my branch where I'm working on the crypto transfer simulation. 😅

sei-db/state_db/bench/helper.go
```diff
@@ -251,6 +260,127 @@ func keyFromIndex(index int64) []byte {
...
+				Height:  int8(i.IAVL.Height), //nolint:gosec
+				Version: i.IAVL.Version,
+			}
+			if node.Height == 0 && node.Value == nil {
+				node.Value = []byte{}
+			}
+			importer.AddNode(node)
+			if node.Height == 0 {
+				totalKeys++
+				if totalKeys%1_000_000 == 0 {
```
cody-littley, line 366: Should number of keys before a snapshot be configurable?

sei-db/state_db/sc/flatkv/store_lifecycle.go
```diff
@@ -54,9 +56,9 @@ func (s *CommitStore) Close() error {
 // Exporter creates an exporter for the given version.
 // NOTE: Not yet implemented. Will be added with state-sync support.
 // The future implementation will export each DB separately with internal key format.
-func (s *CommitStore) Exporter(version int64) (Exporter, error) {
-	// Return a placeholder exporter that indicates not implemented
-	return &notImplementedExporter{}, nil
+func (s *CommitStore) Exporter(version int64) (types.Exporter, error) {
+	// TODO: Implement exporter
+	return nil, nil
 }
```
cody-littley, line 61: Nit, should this panic or return something like `fmt.Errof("not implemented")`?

sei-db/state_db/sc/composite/importer.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/sei-protocol/sei-chain/sei-db/state_db/sc/types"
+)
+
+var _ types.Importer = (*SnapshotImporter)(nil)
+
+type SnapshotImporter struct {
+	cosmosImporter types.Importer
+	evmImporter    types.Importer
+	currentModule  string
```
blindchaser, line 14: it seems currentModule is declared but not set anywhere

sei-db/state_db/sc/flatkv/importer.go
```diff
@@ -0,0 +1,83 @@
...
+	if len(imp.batch) == 0 {
+		return
+	}
+
+	cs := []*proto.NamedChangeSet{{
+		Name:      "evm",
+		Changeset: iavl.ChangeSet{Pairs: imp.batch},
+	}}
+	if err := imp.store.ApplyChangeSets(cs); err != nil {
+		imp.err = fmt.Errorf("import apply changesets: %w", err)
```
blindchaser, line 55: how about adding some logs here like

`imp.store.log.Error("import flush failed, remaining nodes will be dropped", "err", imp.err)`

so we can see error log output with context. otherwise we only see errors when Close()

sei-db/state_db/sc/flatkv/importer.go
```diff
@@ -0,0 +1,83 @@
...
+		return
+	}
+	if err := imp.store.commitBatches(imp.version); err != nil {
+		imp.err = fmt.Errorf("import commit batches: %w", err)
```
blindchaser, line 60: same as above

sei-db/state_db/bench/helper.go
yzang2019: We usually want to finish loading all keys in a snapshot so that we can do hash verification, but I'm open to that feature in a future PR

sei-db/state_db/sc/composite/importer.go
yzang2019: good catch!

sei-db/state_db/sc/flatkv/store_lifecycle.go
yzang2019: Good call, will do that instead

PR #2971, opened 2026-02-24
Fix Rocksdb MVCC read timestamp lifetime for iterators

masih:
RocksDB ReadOptions timestamp handling in MVCC reads had unsafe lifetime semantics:

* newTSReadOptions used a local [8]byte timestamp and passed it to SetTimestamp.
* Under concurrent iterator creation, the timestamp backing memory could be reused, causing versioned reads to drift to the wrong version.
* Iterator read options were not owned/cleaned up with iterator lifetime, making behavior fragile and cleanup inconsistent.

The changes here use heap-backed timestamp slices
(`make([]byte, TimestampSize)`) when calling `SetTimestamp` / `SetIterStartTimestamp`, and create `ReadOptions` once per iterator in `Iterator` and `ReverseIterator`, passing them into the iterator wrapper. The `mvcc.iterator` is extended to own `*grocksdb.ReadOptions` and destroy them in `Close()`, while `getSlice` is updated to destroy per-read `ReadOptions` after `GetCF`. `RawIterate` is also updated to pass owned read options into the iterator lifecycle. As a minor cleanup, `Next()` is made a pointer receiver for consistency.

This should also fix the flaky rocksdb tests, flaked most recently on [main](https://github.com/sei-protocol/sei-chain/actions/runs/22270452280/job/64424250807) among other PRs with unrelated changes.

### DIFF
--- a/sei-db/db_engine/pebbledb/mvcc/iterator.go
+++ b/sei-db/db_engine/pebbledb/mvcc/iterator.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"sync"
 
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric"
@@ -32,6 +33,8 @@ type iterator struct {
 	reverse            bool
 	iterationCount     int64
 	storeKey           string
+
+	closeSync sync.Once
 }
 
 func newPebbleDBIterator(src *pebble.Iterator, prefix, mvccStart, mvccEnd []byte, version int64, earliestVersion int64, reverse bool, storeKey string) *iterator {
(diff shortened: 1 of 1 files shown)


sei-db/db_engine/rocksdb/mvcc/iterator.go
```diff
@@ -157,6 +160,10 @@ func (itr *iterator) Error() error {
 func (itr *iterator) Close() error {
 	itr.source.Close()
 	itr.source = nil
+	if itr.readOpts != nil {
```
bdchatham: nit: small concern is a race on this if-statement and a second call to Destroy occurring.

masih: Thanks for pointing that out Brandon

PR #2972, opened 2026-02-24
feat(flatkv): add snapshot, WAL catchup, and rollback support

blindchaser:
Introduce a snapshot-based lifecycle for FlatKV so that restarts replay
only from the nearest PebbleDB checkpoint instead of the full WAL.

Key changes:
- Snapshot management: immutable PebbleDB checkpoints created via
  Checkpoint(), managed through a "current" symlink and atomic
  directory operations. Configurable interval, retention, and
  minimum time between snapshots.
- Working directory: mutable clone of the baseline snapshot (hardlinks
  for .sst files) so writes never mutate snapshot dirs.
- WAL catchup: on open, replay changelog entries from the snapshot
  version to the target version using O(1) arithmetic + O(log N)
  binary search for offset resolution.
- Rollback: rewind to the best snapshot <= target, truncate WAL,
  replay to exact version, and prune future snapshots.
- File lock: prevent concurrent access from multiple processes.
- Migration: automatically move pre-snapshot flat layout into a
  versioned snapshot directory on first open.
- Auto WAL truncation: periodically discard WAL entries older than
  the earliest retained snapshot.
- Fix account LtHash baseline capture to use pre-batch state when
  multiple ApplyChangeSets calls precede a single Commit.
- Add legacyDB to flushAllDBs.
- Mark Iterator/IteratorByPrefix as EXPERIMENTAL (unused in production).


sei-db/db_engine/types.go
```diff
@@ -55,6 +55,12 @@ type Batch interface {
 	io.Closer
 }
 
+// Checkpointable is an optional capability for DB engines that support
+// efficient point-in-time snapshots via filesystem hardlinks.
+type Checkpointable interface {
```
cody-littley, line 69: Can you augment the godoc with information about concurrency? For example, is it safe to call this method concurrently with updates in other threads? When this method returns, is the checkpoint capable of surviving a host OS crash?

sei-db/state_db/sc/flatkv/snapshot.go
```diff
@@ -0,0 +1,626 @@
...
+	"github.com/sei-protocol/sei-chain/sei-db/db_engine/pebbledb"
+	"github.com/sei-protocol/sei-chain/sei-db/proto"
+)
+
+const (
+	snapshotPrefix = "snapshot-"
+	snapshotDirLen = len(snapshotPrefix) + 20
+
+	currentLink    = "current"
+	currentTmpLink = "current-tmp"
```
cody-littley, line 42: Short descriptions for what each constant is for might be helpful.

sei-db/state_db/sc/flatkv/snapshot.go
```diff
@@ -0,0 +1,626 @@
...
+	}
+	return nil
+}
+
+// updateCurrentSymlink atomically updates the current symlink to point at snapshotDir.
+// snapshotDir should be the bare directory name (e.g. "snapshot-00000000000000000100"),
+// not a full path.
+func updateCurrentSymlink(root, snapshotDir string) error {
+	tmpPath := filepath.Join(root, currentTmpLink)
+	_ = os.Remove(tmpPath)
```
cody-littley: What if this removal fails due to invalid file permissions? I'm guessing you aren't checking the error in case it fails due to the path not existing. Perhaps this can first check if the file exists, delete it if it exists, and then return an error if that deletion fails.

sei-db/state_db/sc/flatkv/snapshot.go
cody-littley: One way to document this, feel free to push back if you think it's too much overhead.

At past companies, when documenting this sort of file layout, I sometimes find it useful to do the following:

- write a simple unit test that generates a basic file structure
- pause the unit test before it deletes its data
- run `tree` on the directory created by the test
- edit the result for readability copy-paste the rest somewhere

Here's an example: https://github.com/Layr-Labs/eigenda/blob/master/litt/docs/filesystem_layout.md

Image

If the file layout is to big, it might make sense to split it out into a markdown file that you can just reference in the godoc.

sei-db/state_db/sc/flatkv/store.go
```diff
@@ -106,59 +113,190 @@ func NewCommitStore(homeDir string, log logger.Logger, cfg Config) *CommitStore
...
+
+// open opens all database instances.
+//
+// Layout:
+//
+//	flatkv/
+//	  current -> snapshot-NNNNN
+//	  snapshot-NNNNN/{account,code,...}/  (immutable)
+//	  working/{account,code,...}/          (mutable clone)
+//	  changelog/                           (WAL, shared)
```
cody-littley, line 200: Ah, didn't see this when I made my comment above. It might be nice to replicate this to the other file that deals with directory names.

sei-db/db_engine/types.go
blindchaser: sure, added concurrency and durability documentation to the Checkpointable

sei-db/state_db/sc/flatkv/snapshot.go
blindchaser: good suggestion. adding inline godoc with the directory layout (ASCII tree) above the constants in snapshot.go. this covers the logical structure (which directories exist and what they store).If we add more complexity later (e.g. sharded storage), we may revist and split it out into a dedicated doc.

sei-db/state_db/sc/flatkv/snapshot.go
blindchaser: adding error handling here

sei-db/config/toml.go
```diff
@@ -65,15 +65,26 @@ sc-snapshot-write-rate-mbps = {{ .StateCommit.MemIAVLConfig.SnapshotWriteRateMBp
...
 # AsyncWriteBuffer defines the size of the async write buffer for data DBs.
 # Set <= 0 for synchronous writes.
 async-write-buffer = {{ .StateCommit.FlatKVConfig.AsyncWriteBuffer }}
+
+# SnapshotInterval defines how often (in blocks) a PebbleDB checkpoint is taken.
+# 0 disables auto-snapshots. Default: 10000.
+snapshot-interval = {{ .StateCommit.FlatKVConfig.SnapshotInterval }}
+
+# SnapshotKeepRecent defines how many old snapshots to keep besides the latest one.
+# 0 = keep only the current snapshot. Default: 1.
```
yzang2019: We probably need more than 1 snapshot here in order to do state export

sei-db/config/toml.go
```diff
@@ -65,15 +65,26 @@ sc-snapshot-write-rate-mbps = {{ .StateCommit.MemIAVLConfig.SnapshotWriteRateMBp
...
+snapshot-keep-recent = {{ .StateCommit.FlatKVConfig.SnapshotKeepRecent }}
+
+# SnapshotMinTimeInterval is the minimum wall-clock seconds between consecutive
+# auto-snapshots. Prevents dense snapshots during catch-up. Default: 3600 (1 hour).
+snapshot-min-time-interval = {{ .StateCommit.FlatKVConfig.SnapshotMinTimeInterval }}
```
yzang2019: I think this is a memiavl specific optimization, and not needed by flatKV, since FlatKV snapshot is so cheap, would rather remove the config for simplicity

sei-db/state_db/sc/flatkv/snapshot.go
yzang2019: Yeah, I think that layer document might help people understand FlatKV better, but agree we can work on a ReadMe of this whole FlatKV and include there in the future

sei-db/state_db/sc/flatkv/store_catchup.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("catchup replay: %w", err)
+	}
+
+	if replayed > 0 {
+		if !s.config.Fsync {
+			if err := s.flushAllDBs(); err != nil {
```
yzang2019, line 166: Why do we flush here?

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -232,10 +229,45 @@ func (s *CommitStore) Commit() (int64, error) {
...
+	if version%1000 == 0 {
+		s.tryTruncateWAL()
+	}
+
 	s.log.Info("Committed version", "version", version)
 	return version, nil
 }
 
+// flushAllDBs flushes all data DBs to ensure data is on disk.
+func (s *CommitStore) flushAllDBs() error {
```
yzang2019, line 274: I think the only place where need to call flush is right before we take a new snapshot?

sei-db/config/toml.go
blindchaser: agree. I removed snapshot-min-time-interval from FlatKV config/template/code, and now auto-snapshot is triggered only by snapshot-interval.

sei-db/state_db/sc/flatkv/store_catchup.go
blindchaser: during catch-up with NoSync, per-entry batch commits may still be in OS/page cache. If we advance global metadata immediately, a crash can leave the global watermark ahead of durable data.

so we do a single flush before committing global metadata to preserve durability ordering. I added this rationale as an inline comment.

sei-db/state_db/sc/flatkv/store_write.go
blindchaser: comments above

sei-db/config/toml.go
blindchaser: i see. bump the default to be 2

PR #2974, opened 2026-02-24
added a config parameter to limit outbound p2p connections.

pompon0:
For a healthy p2p network nodes should provide more capacity than they consume. This PR adds a limit on the number of outbound connections, so that a node does not try to fill up ALL of the available connection slots that it has, but rather to establish just a reasonable number of connections and let other nodes connect to it. Additionally:
* logic for managing a pool of connections has been extracted from peerManagerInner to a separate type (we have regular and persistent pools)
* wired up the HandshakeTimeout, DialTimeout and a couple other P2P config fields which were not properly copied to routeroptions.

### DIFF
--- a/sei-tendermint/libs/utils/math.go
+++ b/sei-tendermint/libs/utils/math.go
@@ -1,8 +1,9 @@
 package utils
 
 import (
-	"golang.org/x/exp/constraints"
 	"unsafe"
+
+	"golang.org/x/exp/constraints"
 )
 
 // Bits returns the number of bits of the given integer type.
--- a/sei-tendermint/libs/utils/tcp/tcp_test.go
+++ b/sei-tendermint/libs/utils/tcp/tcp_test.go
@@ -3,10 +3,11 @@ package tcp
 import (
 	"context"
 	"fmt"
-	"github.com/sei-protocol/sei-chain/sei-tendermint/libs/utils"
-	"github.com/sei-protocol/sei-chain/sei-tendermint/libs/utils/scope"
 	"net/netip"
 	"testing"
+
+	"github.com/sei-protocol/sei-chain/sei-tendermint/libs/utils"
+	"github.com/sei-protocol/sei-chain/sei-tendermint/libs/utils/scope"
 )
 
 func TestTestReservePort(t *testing.T) {


sei-tendermint/config/config.go
```diff
@@ -647,10 +647,14 @@ type P2PConfig struct {
 	// UPNP port forwarding. UNUSED
 	UPNP bool `mapstructure:"upnp"`
 
-	// MaxConnections defines the maximum number of connected peers (inbound and
-	// outbound).
+	// MaxConnections limits the number of connected peers (inbound and outbound).
 	MaxConnections uint16 `mapstructure:"max-connections"`
 
+	// MaxConnections limits the number of outbound peers.
```
wen-coding: MaxConnections -> MaxOutboundConnections ?

sei-tendermint/config/config.go
```diff
@@ -647,10 +647,14 @@ type P2PConfig struct {
...
+	// It should be significantly lower than MaxConnections, unless
```
wen-coding, line 654: nit: maybe we should give a guideline how much lower this should be compared to MaxConnections

sei-tendermint/config/config.go
```diff
@@ -647,10 +647,14 @@ type P2PConfig struct {
...
+	// the node is supposed to have a small number of connections altogether.
+	MaxOutboundConnections int
```
wen-coding: nit: why int instead of uint16 like MaxConnections?

sei-tendermint/internal/p2p/peermanager.go
```diff
@@ -16,254 +14,84 @@ const maxAddrsPerPeer = 10
...
-		return utils.Some(v)
-	}
-	return utils.None[V]()
+	return i.regular.AddAddr(addr)
 }
 
 func (i *peerManagerInner[C]) TryStartDial(persistentPeer bool) (NodeAddress, bool) {
 	// Check concurrent dials limit.
-	if len(i.dialing) >= i.options.maxDials() {
+	if len(i.regular.dialing)+len(i.persistent.dialing) >= i.options.maxDials() {
```
wen-coding, line 39: should we cleanly split everything (including maxdials) between the two pool?

sei-tendermint/config/config.go
```diff
@@ -647,10 +647,14 @@ type P2PConfig struct {
...
 	UPNP bool `mapstructure:"upnp"`
 
-	// MaxConnections defines the maximum number of connected peers (inbound and
-	// outbound).
+	// MaxConnections limits the number of connected peers (inbound and outbound).
 	MaxConnections uint16 `mapstructure:"max-connections"`
 
+	// MaxConnections limits the number of outbound peers.
+	// It should be significantly lower than MaxConnections, unless
+	// the node is supposed to have a small number of connections altogether.
```
wen-coding, line 655: Hmm, should we say only regular pool is subject to this limit, it doesn't limit persistent peers?

sei-tendermint/internal/p2p/peermanager.go
```diff
@@ -16,254 +14,84 @@ const maxAddrsPerPeer = 10
...
-		peerAddrs.fails[addr] = now
+		i.regular.DialFailed(addr)
 	}
 }
 
 func (i *peerManagerInner[C]) Evict(id types.NodeID) {
-	if !i.isPersistent(id) {
-		delete(i.addrs, id)
-	}
-	if !i.isUnconditional[id] {
```
wen-coding: originally even when id is persistent we will close connection if it's not unconditional, we don't need this any more?

sei-tendermint/libs/utils/im/im.go
```diff
@@ -28,10 +29,25 @@ func (m Map[K, V]) Get(key K) (V, bool) {
...
+		return utils.Some(v)
+	}
+	return utils.None[V]()
+}
+
 func (m Map[K, V]) Set(key K, value V) Map[K, V] {
 	return Map[K, V]{m.m.Set(key, value)}
 }
 
+func (m Map[K, V]) SetOpt(key K, mvalue utils.Option[V]) Map[K, V] {
```
wen-coding, line 45: nit: SetOpt sounds like it will do nothing if the key doesn't exist, how about SetOrDelete or something like that.

sei-tendermint/internal/p2p/peermanager_pool.go
```diff
@@ -0,0 +1,242 @@
...
+
+func getOpt[K comparable, V any](m map[K]V, k K) utils.Option[V] {
+	if v, ok := m[k]; ok {
+		return utils.Some(v)
+	}
+	return utils.None[V]()
+}
+
+type poolConfig struct {
+	selfID          types.NodeID
```
wen-coding, line 33: nit: would you use different selfID on different pools?

sei-tendermint/internal/p2p/peermanager.go
```diff
@@ -349,39 +177,38 @@ func (m *peerManager[C]) Subscribe() *peerUpdatesRecv[C] {
...
-
-		conns:   utils.NewAtomicSend(im.NewMap[types.NodeID, C]()),
-		addrs:   map[types.NodeID]*peerAddrs{},
-		dialing: map[types.NodeID]NodeAddress{},
+		selfID:       selfID,
+		options:      options,
+		isPersistent: map[types.NodeID]bool{},
+		conns:        utils.NewAtomicSend(im.NewMap[types.NodeID, C]()),
+
+		persistent: newPool[C](poolConfig{selfID: selfID}),
```
wen-coding, line 185: Where are the metrics? Should we split metrics by pool as well?

sei-tendermint/config/config.go
pompon0: well, I don't know that yet :). I estimate 10 outbound, 50 total, but this is tbd.

sei-tendermint/internal/p2p/peermanager.go
pompon0: probably, but that's more configuration. We can reconsider that later.

sei-tendermint/config/config.go
pompon0: changed both to uint.

sei-tendermint/libs/utils/im/im.go
pompon0: It is dual to getopt. I've added documentation. I can change to SetOrDelete if you feel strongly about this.

sei-tendermint/internal/p2p/peermanager_pool.go
pompon0: No, but I want the pools to be self contained.

sei-tendermint/internal/p2p/peermanager.go
pompon0: there is a metric for the total number of connections, there is a log printed every 10s with number of connections (regular vs persistent) and there is network state exposed via net_info rpc endpoint. Metrics and the rpc endpoint need a revamp indeed but it is out of scope of this pr.

sei-tendermint/internal/p2p/peermanager.go
pompon0: all persistent peers are unconditional (see the old newPeerManager impl). In this pr I've merged the unconditional and persistent to be the same concept (see pr description) to simplify the situation a bit.

PR #2976, opened 2026-02-25
Fix hanging upgrade tests by adding timeouts to wait_for_height

masih:
The minor upgrade integration test was occasionally hanging until the workflow timeout killed it, which cascaded into failing Integration Test Check.

The culprit was `wait_for_height.sh`: it used an unbounded while true loop, so if a node stayed alive but stopped producing blocks, the script would spin forever.

Rewrote `wait_for_height.sh` to fail fast instead of hanging. The script now has configurable timeouts via environment variables: a per-call timeout for seid status (default 5s), a max total wait (default 180s), and a max stall duration when no new blocks appear (default 30s). It then exits with a clear error if any of those thresholds are breached.

Stall tracking resets when the node advances past target-1, and logs now include elapsed time for easier debugging.

Flaked on [main](https://github.com/sei-protocol/sei-chain/actions/runs/22335750775/job/64627784198).

### DIFF
--- a/integration_test/upgrade_module/scripts/wait_for_height.sh
+++ b/integration_test/upgrade_module/scripts/wait_for_height.sh
@@ -12,12 +12,26 @@ fi
 # Fetch the node ID from the environment or default to 0
 NODE_ID=${ID:-0}
 
+# Keep this script bounded to avoid hanging CI jobs indefinitely.
+STATUS_TIMEOUT_SECONDS=${WAIT_FOR_HEIGHT_STATUS_TIMEOUT_SECONDS:-5}
+MAX_WAIT_SECONDS=${WAIT_FOR_HEIGHT_MAX_WAIT_SECONDS:-180}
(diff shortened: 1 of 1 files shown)


pompon0 (approved):

integration_test/upgrade_module/scripts/wait_for_height.sh
```diff
@@ -12,20 +12,54 @@ fi
...
+
    # Check if the service is running (it might panic at the height and not let us reach it)
    if ! pgrep -f "seid start --chain-id sei" > /dev/null; then
       echo "Seid no longer running (panic)"
       break
    fi
 
    # Query the current block height of the node
-   CURRENT_BLOCK_HEIGHT=$(seid status | jq '.SyncInfo.latest_block_height' -r)
+   CURRENT_BLOCK_HEIGHT=$(timeout "$STATUS_TIMEOUT_SECONDS" seid status 2>/dev/null | jq '.SyncInfo.latest_block_height' -r 2>/dev/null)
```
mojtaba-esk, line 42: love this line

PR #2979, opened 2026-02-25
ERC20 simulation benchmark

cody-littley:

Adds a new database benchmark suite for simulating ERC20 workload. This benchmark is intended to run at large scale and over large durations.

See `README.md` for information about how to run this tool locally.

Ran the benchmark locally.

### DIFF
--- a/sei-db/config/toml.go
+++ b/sei-db/config/toml.go
@@ -65,15 +65,22 @@ sc-snapshot-write-rate-mbps = {{ .StateCommit.MemIAVLConfig.SnapshotWriteRateMBp
 ###############################################################################
 
 [state-commit.flatkv]
-# Fsync controls whether data DB writes use fsync for durability.
-# When true (default): all data DB writes use Sync=true for maximum durability.
-# When false: data DBs use Sync=false for better performance.
-# WAL and metaDB always use sync writes regardless.
+# Fsync controls whether PebbleDB writes (data DBs + metadataDB) use fsync.
+# WAL always uses NoSync (matching memiavl); crash recovery relies on
+# WAL catchup, which is idempotent. Default: false.
 fsync = {{ .StateCommit.FlatKVConfig.Fsync }}
 
 # AsyncWriteBuffer defines the size of the async write buffer for data DBs.
 # Set <= 0 for synchronous writes.
 async-write-buffer = {{ .StateCommit.FlatKVConfig.AsyncWriteBuffer }}
+
+# SnapshotInterval defines how often (in blocks) a PebbleDB checkpoint is taken.
+# 0 disables auto-snapshots. Default: 10000.
+snapshot-interval = {{ .StateCommit.FlatKVConfig.SnapshotInterval }}
+
+# SnapshotKeepRecent defines how many old snapshots to keep besides the latest one.
+# 0 = keep only the current snapshot. Default: 2.
+snapshot-keep-recent = {{ .StateCommit.FlatKVConfig.SnapshotKeepRecent }}
 `
 
 // StateStoreConfigTemplate defines the configuration template for state-store
--- a/sei-db/db_engine/pebbledb/db.go
+++ b/sei-db/db_engine/pebbledb/db.go
@@ -123,6 +123,15 @@ func (p *pebbleDB) Flush() error {
 	return p.db.Flush()
 }
 
+func (p *pebbleDB) Checkpoint(destDir string) error {
+	if p.db == nil {
+		return errors.New("pebbleDB: checkpoint on closed database")
+	}
+	return p.db.Checkpoint(destDir, pebble.WithFlushedWAL())
+}
+
+var _ db_engine.Checkpointable = (*pebbleDB)(nil)
+
 func (p *pebbleDB) Close() error {
 	// Make Close idempotent: Pebble panics if Close is called twice.
 	if p.db == nil {
--- a/sei-db/db_engine/types.go
+++ b/sei-db/db_engine/types.go
@@ -55,6 +55,21 @@ type Batch interface {
 	io.Closer
 }
 
+// Checkpointable is an optional capability for DB engines that support
+// efficient point-in-time snapshots via filesystem hardlinks.
+//
+// Concurrency: Checkpoint is safe to call concurrently with reads and writes
+// on the same DB instance. The resulting snapshot reflects a consistent
+// point-in-time view; concurrent writes may or may not be included.
+//
+// Durability: When Checkpoint returns nil, the destination directory is a
+// complete, crash-safe copy of the database state. It survives host OS
+// crashes because it consists of hardlinks to already-fsynced SST files
+// plus a flushed manifest.
+type Checkpointable interface {
+	Checkpoint(destDir string) error
+}
+
 // Iterator provides ordered iteration over keyspace with seek primitives.
 //
 // Zero-copy contract:
--- a/sei-db/state_db/sc/flatkv/api.go
+++ b/sei-db/state_db/sc/flatkv/api.go
@@ -9,12 +9,7 @@ import (
 
 // Options configures a FlatKV store.
 type Options struct {
-	// Dir is the base directory containing
-	// account/,
-	// code/,
-	// storage/,
-	// changelog/,
-	// __metadata.
+	// Dir is the base directory containing snapshot dirs, working/, and changelog/.
 	Dir string
 }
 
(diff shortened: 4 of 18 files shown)


sei-db/state_db/bench/cryptosim/database.go
```diff
@@ -0,0 +1,210 @@
...
+
+// Creates a new database for the cryptosim benchmark.
+func NewDatabase(
+	config *CryptoSimConfig,
+	db wrappers.DBWrapper,
+) *Database {
+	return &Database{
+		config: config,
+		db:     db,
+		batch:  NewSyncMap[string, *proto.NamedChangeSet](),
```
yzang2019: It seems like the way we use batch is for pending updates, so the actual value just need to be []byte right? Why do we need to store it as proto.NamedChangeSet?

sei-db/state_db/bench/cryptosim/config/basic-config.json
```diff
@@ -0,0 +1,6 @@
+{
+    "Comment": "Basic configuration for the cryptosim benchmark. Intended for basic correctness/sanity testing.",
+    "DataDir": "data",
+    "MinimumNumberOfAccounts": 100000
```
yzang2019: Shall we list all the configurable options in basic-config and set a default for them? Which should make it easier to override in the future for a different testing setup

sei-db/state_db/bench/cryptosim/database.go
cody-littley: I originally did it this way because this is the type that `ApplyChangeSets()`, but I agree that it's a lot easier to read if I store it as `[]byte` data, and only convert to `proto.NamedChangeSet` when it's time to finalize a block. Change made.

sei-db/state_db/bench/cryptosim/config/basic-config.json
cody-littley: Added. I also added a comment in the code reminding people to keep the `basic-config.json` file in sync with default values.

For more specialized configs, we can probably drop fields that shadow the default values. But as an example config file, I can see how having all the options already in the file might be helpful.

sei-db/state_db/bench/cryptosim/database.go
```diff
@@ -0,0 +1,196 @@
...
+	return nil
+}
+
+// Retrieve a value from the database/cache.
+//
+// This method is safe to call concurrently with other calls to Put() and Get(). Is not thread
+// safe with FinalizeBlock().
+func (d *Database) Get(key []byte) ([]byte, bool, error) {
+	if value, found := d.batch.Get(string(key)); found {
+		return value, true, nil
```
blindchaser, line 64: is there data race here if we return underlying bytes arrray directly withou copy?

sei-db/state_db/bench/cryptosim/cryptosim.go
```diff
@@ -0,0 +1,367 @@
...
+		select {
+		case <-c.ctx.Done():
+			if c.database.TransactionCount() > 0 {
+				c.generateConsoleReport(true)
+				fmt.Printf("\nTransaction workload halted.\n")
+			}
+			return
+		default:
+
+			txn, err := BuildTransaction(c.dataGenerator)
```
blindchaser, line 293: when BuildTransaction() returns error, the txn is nil. would this panic executor?

sei-db/state_db/bench/cryptosim/database.go
```diff
@@ -0,0 +1,196 @@
...
+		d.flushFunc()
+	}
+
+	if d.transactionsInCurrentBlock == 0 {
+		return nil
+	}
+
+	d.transactionsInCurrentBlock = 0
+
+	changeSets := make([]*proto.NamedChangeSet, 0, d.transactionsInCurrentBlock+2)
```
blindchaser, line 129: it looks slice always starts with capacity 2 regardless of batch size, is this on purpose?

sei-db/state_db/bench/cryptosim/data_generator.go
```diff
@@ -0,0 +1,257 @@
...
+	if err != nil {
+		return 0, nil, fmt.Errorf("failed to put account: %w", err)
+	}
+
+	return accountID, address, nil
+}
+
+// Creates a new ERC20 contract and optionally writes it to the database. Returns the address of the new ERC20 contract.
+func (d *DataGenerator) CreateNewErc20Contract(
+	cryptosim *CryptoSim,
```
blindchaser: how's the cryptosim param being used?

sei-db/state_db/bench/cryptosim/cryptosim_config.go
```diff
@@ -0,0 +1,211 @@
...
+
+	// The size of the queue for each transaction executor.
+	ExecutorQueueSize int
+}
+
+// Returns the default configuration for the cryptosim benchmark.
+func DefaultCryptoSimConfig() *CryptoSimConfig {
+
+	// Note: if you add new fields or modify default values, be sure to keep config/basic-config.json in sync.
+	// That file should conatain every available config set to its default value, as a reference.
```
blindchaser: nit: contain

blindchaser:
Is there a way to configure how long the benchmark runs?

sei-db/state_db/bench/cryptosim/database.go
cody-littley: Not a data race unless we attempt to modify the returned slices.

Although generally making a defensive copy is not a bad idea, I'm especially concerned about benchmark overhead. I'd rather have a footgun here than risk unnecessary overhead.

In order to make this function's behavior more obvious, I added a comment in the godoc mentioning that modification of values was unsafe without first copying the data.

sei-db/state_db/bench/cryptosim/cryptosim.go
cody-littley: You are correct, a nill transaction would cause problems. Added a `continue` if the err is not nil.

sei-db/state_db/bench/cryptosim/database.go
cody-littley: Intentional. 😄 

Every time the benchmark finalizes a batch, it also updates two benchmark metadata keys: one for max account ID, and another for max ERC20 contract ID. So we need capacity for two extra change sets to avoid resizing the slice.

sei-db/state_db/bench/cryptosim/data_generator.go
cody-littley: It used to be used, but now it's no longer used. Removed.

sei-db/state_db/bench/cryptosim/cryptosim_config.go
cody-littley: fixed

cody-littley:
@blindchaser 

> Is there a way to configure how long the benchmark runs?

There wasn't a way, but I added one. By default it will run forever, but you can update a config flag to specify a max runtime.

sei-db/state_db/bench/cryptosim/database.go
blindchaser: binary.BigEndian.PutUint64(srcAccountValue[:8], uint64(txn.newSrcBalance))
binary.BigEndian.PutUint64(dstAccountValue[:8], uint64(txn.newDstBalance))
binary.BigEndian.PutUint64(feeValue[:8], uint64(txn.newFeeBalance))

these are modifying the returned bytes directly, right? 

if we don't want to copy in the Get(), maybe we can copy during Execute(), the caller side.

sei-db/state_db/bench/cryptosim/database.go
blindchaser: got it

sei-db/state_db/bench/helper.go
```diff
@@ -435,8 +439,8 @@ func runBenchmark(b *testing.B, scenario TestScenario, withProgress bool) {
 					progress.Add(len(changeset.Changeset.Pairs))
 				}
 			}
-			err := cs.Close() // close to make sure all data got flushed
-			require.NoError(b, err)
+			closeErr := cs.Close() // close to make sure all data got flushed
```
blindchaser, line 438: looks double close() with one in defer

sei-db/state_db/bench/cryptosim/README.md
```diff
@@ -0,0 +1,52 @@
...
+# Data Persistence
+
+The cryptosim benchmark writes database files. If provided with those database files at startup time, the cryptosim
+benchmark can continue running with that database. This may be useful when operating on large data sets that require
+significant setup time.
+
+Note that if continuing from an earlier run, you may not alter the `Seed`, `CannedRandomSize`, or 
+`AccountKeySize` parameters, since keys are deterministically derived from these configuration values.
+
+# Transaciton Model
```
blindchaser: nit: "Transaction"

sei-db/state_db/bench/helper.go
cody-littley: removed deferred close

sei-db/state_db/bench/cryptosim/README.md
cody-littley: fixed

sei-db/state_db/bench/cryptosim/database.go
cody-littley: Good point, I missed this. Fixed by adding copy before mutation in execution. (Aside: footgun doing what footguns do 😅)

PR #2981, opened 2026-02-26
Parquet remove last file if corrupted

jewei1997:

If the node suddenly crashes and the parquet receiptdb writer is not able to write the last "magic bytes footer" at the end of the last parquet that makes the parquet file readable then DuckDB will not able to read that file upon startup. This fix will examine the last file upon startup. If it is unreadable due to magic bytes missing, we will delete it and restore it using the WAL.

There's another fix here to not create a file that ends in *_receipt_0.parquet and *_log_0.parquet. The first file should be *_receipt_N.parquet and *_log_N.parquet where N is the current block number. To do this we need to lazily create the parquet file upon the first receipt/log being written to it.

unti tests


sei-db/ledger_db/parquet/reader.go
yzang2019: Nit: would be cleaner if we can add and use a helper functions here:

receiptFiles := r.getAllParquetFilesByPrefix("receipt")

logFiles := r.getAllParquetFilesByPrefix("log")

sei-db/ledger_db/parquet/reader.go
```diff
@@ -112,17 +113,22 @@ func (r *Reader) scanExistingFiles() {
...
+	r.closedLogFiles = r.validateAndCleanFiles(logFiles, "receipts")
 }
 
-func (r *Reader) validateFiles(files []string) []string {
+// validateAndCleanFiles checks the last file for readability. If it is corrupt
+// (e.g. missing parquet footer from an unclean shutdown), the file and its
+// counterpart (identified by counterpartPrefix) are deleted from disk so they
+// cannot poison future DuckDB queries. Only the last file needs checking
+// because all previously rotated files had their writers properly closed.
+func (r *Reader) validateAndCleanFiles(files []string, counterpartPrefix string) []string {
```
yzang2019, line 129: How about changing to this:

func (r *Reader) removeCorruptedFiles(receiptFiles []string, logFiles []string) []string {

yzang2019 (approved):

Kbhat1 (approved):

PR #2984, opened 2026-02-26
Refactor StateStore for better readability and Giga support

yzang2019:
This PR mainly add RocksDB support for SS and refactor for better readability and code structure:
- Add New MvccDB interface for DB engine to unify pebbledb mvcc and rocksdb mvcc
- StateStore interface unchanged and now extends MvccDB
- Add a new CosmosStateStore: Thin wrapper around a single DB that implements StateStore. 
- EVMStateStore refactored: Sub-DBs typed as types.MvccDB (was types.StateStore)
- CompositeStateStore simplified (ss/composite/store.go): Both cosmosStore and evmStore fields are now types.StateStore
- not found errors moved to common/errors
- RawImport and migrationKey interface removed
- SS Migrator removed
- db init static function removed, replaced by resolveBackend function
- Fixed recoverFromWAL logic

New structure:
image

### DIFF
--- a/sei-db/state_db/ss/composite/store.go
+++ b/sei-db/state_db/ss/composite/store.go
@@ -315,51 +315,58 @@ func RecoverCompositeStateStore(
 	changelogPath string,
 	compositeStore *CompositeStateStore,
 ) error {
-	// TODO: Remove this piece once cosmos is fully deprecated
+	var cosmosVersion int64
 	if compositeStore.cosmosStore != nil {
-		cosmosVersion := compositeStore.cosmosStore.GetLatestVersion()
-		logger.Info("Recovering Cosmos state store",
-			"cosmosVersion", cosmosVersion,
-			"changelogPath", changelogPath,
-		)
-		if err := ReplayWAL(logger, changelogPath, cosmosVersion, -1, func(entry proto.ChangelogEntry) error {
+		cosmosVersion = compositeStore.cosmosStore.GetLatestVersion()
+	}
+
+	var evmVersion int64
+	if compositeStore.evmStore != nil {
+		evmVersion = compositeStore.evmStore.GetLatestVersion()
+	}
+
+	startVersion := cosmosVersion
+	if compositeStore.evmStore != nil && evmVersion < startVersion {
+		startVersion = evmVersion
+	}
+
+	splitWrite := compositeStore.config.WriteMode == config.SplitWrite
+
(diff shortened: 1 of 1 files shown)


sei-db/state_db/ss/composite/store.go
```diff
@@ -348,259 +253,117 @@ func stripEVMFromChangesets(changesets []*proto.NamedChangeSet) []*proto.NamedCh
...
+		}
+	}
+
+	if compositeStore.evmStore != nil {
+		evmVersion := compositeStore.evmStore.GetLatestVersion()
+		logger.Info("Recovering EVM state store",
+			"evmVersion", evmVersion,
+			"changelogPath", changelogPath,
+		)
+		if err := ReplayWAL(logger, changelogPath, evmVersion, -1, func(entry proto.ChangelogEntry) error {
```
Kbhat1: Curious, we do 2 separate replayWAls one for evmstore and cosmosStore. Wondering if we could just do 1 and do both inside each? 

Not blocking, just curious

yzang2019: It is definitely doable, I feel separating them is cleaner in the code and easier to remove later on, but the side effect is that it would require double read the WAL for some entries. Given that SS shouldn't fall behind too much (limited by buffer size), I think this should be fine to start with?

Kbhat1: Yea definitely fine to start with, could we just add a TODO to reconsider in future @yzang2019 ?

cmd/seid/cmd/root.go
```diff
@@ -311,19 +310,6 @@ func newApp(
 		baseapp.SetOccEnabled(cast.ToBool(appOpts.Get(baseapp.FlagOccEnabled))),
 	)
 
-	// Start migration if --migrate flag is set
-	if cast.ToBool(appOpts.Get("migrate-iavl")) {
```
blindchaser, line 315: it looks `--migrate-iavl` still changes runtime behavior?

blindchaser (approved):

sei-db/state_db/ss/composite/store.go
yzang2019: Added

cmd/seid/cmd/root.go
yzang2019: Yeah pretty sure no one is using that, it's for archive migration

PR #2985, opened 2026-02-26
feat(evm): EVM RPC .io/.iox integration tests (spec fixtures)

mojtaba-esk:
### Summary
This PR proposes to add a runner that executes execution-apis-style .io/.iox fixtures (255 tests) against Sei EVM RPC with spec-only checks. Script seeds block and deploy tx; fixtures use `__SEED__` and bindings (`${txHash}`, `${deployTxHash}`). 
Also a job ( "EVM RPC .io/.iox (spec fixtures)") is added to integration CI with `continue-on-error` so PRs can merge while endpoint coverage is incomplete; failures still show in the run.

### Why

- Track JSON-RPC conformance; many tests currently fail (not implemented / behavior differences).
- Non-blocking CI until endpoints are fixed.

See the [RPC_IO_README.md](https://github.com/sei-protocol/sei-chain/blob/8da453b5115ada53ffa0c9ece0fd86dc189e72de/integration_test/evm_module/rpc_io_test/RPC_IO_README.md) file for more details and the latest test run


integration_test/evm_module/rpc_io_test/rpc_io_test.go
```diff
@@ -0,0 +1,256 @@
...
+		}
+		if len(filtered) > 0 {
+			files = filtered
+		}
+	}
+
+	url := rpcURL()
+	client := &rpcClient{URL: url}
+	if !nodeReachable(client) {
+		t.Skipf("EVM RPC node not reachable at %s", url)
```
masih: Hm this means the tests pass if RPC is not reachable. That means the CI will never be red?

integration_test/evm_module/scripts/evm_rpc_tests.sh
```diff
@@ -0,0 +1,42 @@
...
+run() {
+  docker exec "$CONTAINER" /bin/bash -c "export PATH=\$PATH:/root/go/bin && printf '%s\n' '$PASSWORD' | $*"
+}
+
+# Seed chain with block/tx/contract; export SEI_EVM_IO_SEED_BLOCK so .iox __SEED__ tag resolves to deploy block.
+# CLI deploy expects hex file with no whitespace; write trimmed hex to a temp path in the container.
+docker exec "$CONTAINER" /bin/bash -c "tr -d '[:space:]' < \"$CONTRACT_HEX\" > /tmp/minimal_contract.hex"
+run seid tx evm associate-address --from "$FROM" $KEYRING_FLAG --chain-id sei -b block -y 2>/dev/null || true
+run seid tx evm send "$RECIPIENT" 1 --from "$FROM" $KEYRING_FLAG --chain-id sei --evm-rpc http://localhost:8545 -b sync -y
+DEPLOY_OUT=$(run seid tx evm deploy /tmp/minimal_contract.hex --from "$FROM" $KEYRING_FLAG --chain-id sei --evm-rpc http://localhost:8545 -b block -y 2>&1) || true
```
masih: Should this be using `SEI_EVM_RPC_URL` instead of hardcoding localhost?

integration_test/evm_module/scripts/evm_rpc_tests.sh
```diff
@@ -0,0 +1,42 @@
...
+DEPLOY_TX=$(echo "$DEPLOY_OUT" | grep -oE '0x[a-fA-F0-9]{64}' | head -1)
+if [[ -n "$DEPLOY_TX" ]]; then
+  sleep 2
+  for _ in 1 2 3 4 5 6 7 8 9 10; do
+    RESP=$(docker exec "$CONTAINER" curl -s -X POST -H "Content-Type: application/json" -d "{\"jsonrpc\":\"2.0\",\"id\":1,\"method\":\"eth_getTransactionReceipt\",\"params\":[\"$DEPLOY_TX\"]}" http://localhost:8545 2>/dev/null) || true
+    SEED=$(echo "$RESP" | jq -r '.result.blockNumber // empty' 2>/dev/null)
```
masih: Small note that this would silently fail if `jq` is not in path.

integration_test/evm_module/rpc_io_test/rpc_io_test.go
mojtaba-esk: that's true. Unfortunately since the Coverage fast test in the CI failed because this test fails, I had to make it skip. As the coverage test does not start a cluster nor an RPC endpoint. What do you think we should do about it?

integration_test/evm_module/scripts/evm_rpc_tests.sh
mojtaba-esk: nice catch! initially I thought to make it to be able to run tests against a running network like testnet or mainnet and I actually did it. but later as we needed particular data to exist on a known block for a number of tests to be passed. I left that idea. but I am a fan of having it as a variable to have more flexibility so we might still want to run it against another RPC server. will fix it.

integration_test/evm_module/scripts/evm_rpc_tests.sh
mojtaba-esk: Correct, let me see how to fix it.

integration_test/evm_module/rpc_io_test/rpc_io_test.go
masih: Ah I see; easy fix:
* Use an env var instead to skip the tests, instead of RPC unavailability. Then set that env var in CI for the integration tests.

This way we make sure that tests don't accidentally get skipped if there is a genuine issue standing up RPC as part of actual RPC testing.

An alternative is build flags but i think env var is simpler

mojtaba-esk: ah I see, good hint, thanks man. will do it rn.

mojtaba-esk:
> Awesome! glad to see so many tests. This is going to up our game in terms of EVM API compatibility.
> 
> Question: of the 255 tests how many actually pass and what is the plan for those that do not?

Thanks. Here is a report on how many tests are used from ETH vanilla, how many modified slightly to match the data (blockhash, tx,...) and how many are actually generated copying the same style: https://github.com/sei-protocol/sei-chain/blob/8da453b5115ada53ffa0c9ece0fd86dc189e72de/integration_test/evm_module/rpc_io_test/RPC_IO_README.md

20 endpoints are not covered by Eth Exec API so we generated tests for them + 18 endpoints which we updated .io files to .iox (extended io) which has more stuff to handle some stateful test cases

integration_test/evm_module/rpc_io_test/io.go
```diff
@@ -0,0 +1,292 @@
...
+			}
+
+			pairs = append(pairs, ioxPair{Request: curReq, Expected: []byte(strings.TrimPrefix(trimmed, respPrefix))})
+			lastIdx = len(pairs) - 1
+			inBinding = true
+			curReq = nil
+			continue
+		}
+
+		if inBinding && lastIdx >= 0 && strings.HasPrefix(trimmed, directivePrefix) {
```
masih, line 91: Small point that `parseIOFile` doesn't seem to return error for invalid fixture directives. A malformed directive would then result in assertions getting dropped scilently.

.github/workflows/integration-test.yml
```diff
@@ -184,7 +190,9 @@ jobs:
       - name: Verify Sei Chain is running
         run: python3 integration_test/scripts/runner.py integration_test/startup/startup_test.yaml
 
+      # EVM execution-apis .io is allowed to fail until endpoint coverage is fixed (non-blocking for PR merge)
       - name: ${{ matrix.test.name }}
+        continue-on-error: ${{ matrix.test.name == 'EVM RPC .io/.iox (spec fixtures)' }}
```
masih, line 195: Have you had a chance to dig through the ones that are failing to see why? it would be awesome to get them all to pass. Exiting to see so many passing already 🙌

integration_test/evm_module/rpc_io_test/RPC_IO_README.md
```diff
@@ -0,0 +1,333 @@
...
+
+
+### Skipped tests (0)
+
+With the script setting **SEI_EVM_IO_SEED_BLOCK** and **SEI_EVM_IO_DEPLOY_TX_HASH**, no tests are skipped in the latest run. If you run `go test` without the script, some tests may skip for missing `${txHash}` or `${deployTxHash}`. When a test skips, the runner logs **[SKIP]** lines with bindings and placeholders so you can see why.
+
+**Debug one or a few SEED tests:** Run only specific files with extra per-pair logging (request after substitution, bindings, whether `result.transactions` is present):
+
+```bash
+SEI_EVM_IO_DEBUG_FILES="debug_getRawTransaction/get-tx.iox" go test ./integration_test/evm_module/rpc_io_test/ -v -run TestEVMRPCSpec
```
masih: Also mention `SEI_EVM_IO_RUN_INTEGRATION=1`?

integration_test/evm_module/rpc_io_test/io.go
mojtaba-esk: oops, correct. nice one. will add some validations to it.

.github/workflows/integration-test.yml
mojtaba-esk: Not yet, but plan to  do it.

integration_test/evm_module/rpc_io_test/RPC_IO_README.md
mojtaba-esk: sure, missed here.

PR #2987, opened 2026-02-27
Migrate tendermint logging to `slog` and remove go-kit/log dependency

masih:
Remove dependency to go-kit/log and logfmt, and adapt existing logging utility libraries to use Go SDK's `slog`

The changes here intentionally aim to introduce as little changes as possible, and are part of the larger integration of `seilog`, an extended `slog` library into sei-chain.


masih:
Flaking test fixed [here](https://github.com/sei-protocol/sei-chain/pull/2998).

PR #2988, opened 2026-02-27
Fix flaky `TestPeerManager_MaxOutboundConnectionsForDialing`

masih:
The test's atomic counter (dialsAndConns) was decremented via defer, creating a race window where the peer manager had already freed a slot (allowing a new StartDial) but the test counter still included the old dial/conn. This caused transient "dialing + outbound = 4, want <= 3" assertion failures, especially under -race in CI.

Decrement dialsAndConns immediately before DialFailed/Disconnected (and in early error paths) so test accounting stays synchronized with slot release. No changes to peer-manager production logic.

Flaked on [main](https://github.com/sei-protocol/sei-chain/actions/runs/22451497719/job/65020933922).

### DIFF
--- a/sei-tendermint/internal/p2p/peermanager_test.go
+++ b/sei-tendermint/internal/p2p/peermanager_test.go
@@ -495,22 +495,28 @@ func TestPeerManager_MaxOutboundConnectionsForDialing(t *testing.T) {
 				return fmt.Errorf("dialing + outbound = %v, want <= %v", got, maxOutbound)
 			}
 			s.Spawn(func() error {
-				defer dialsAndConns.Add(-1)
 				if i%2 == 0 {
 					if err := utils.Sleep(ctx, 10*time.Millisecond); err != nil {
 						return err
 					}
+					// Keep accounting in sync with slot release: decrement before
+					// unblocking StartDial to avoid transient overcount in this test.
+					dialsAndConns.Add(-1)
 					m.DialFailed(addr)
 					return nil
 				}
 				conn := makeConnTo(addr)
 				if err := m.Connected(conn); err != nil {
+					dialsAndConns.Add(-1)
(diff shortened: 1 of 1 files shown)


sei-tendermint/internal/p2p/peermanager_test.go
```diff
@@ -495,22 +495,28 @@ func TestPeerManager_MaxOutboundConnectionsForDialing(t *testing.T) {
 				return fmt.Errorf("dialing + outbound = %v, want <= %v", got, maxOutbound)
 			}
 			s.Spawn(func() error {
-				defer dialsAndConns.Add(-1)
 				if i%2 == 0 {
 					if err := utils.Sleep(ctx, 10*time.Millisecond); err != nil {
 						return err
```
mojtaba-esk, line 500: Q: why don't we need `dialsAndConns.Add(-1)` here?

mojtaba-esk: is it missing or on purpose?

masih: The utils.Sleep obfuscates the fact that it only returns error when ctx errors; not in normal success flow.
The dialsAndConns counter is only used for the success-path invariant check. So if it returns error due to context timeout or cancellation then the test has failed.

masih: Does that make sense Mojtaba? I am happy to add it anyway. It just makes no difference I think.

pompon0: I'd add it

mojtaba-esk: ah I see, that's probably why the defer thing was messing with it.

PR #2990, opened 2026-02-27
Reduced number of addresses per NodeID in peermanager to 1

pompon0:
With 6.3 upgrade, invalid/outdated addresses are no longer gossiped by honest nodes (malicious nodes can poison the addresses buffer anyway). With that we can simplify the logic to handle just 1 address per peer. I've also added an additional mechanism to prefer public ips over private ips, to account for nodes gossiping private addresses (that they have configured as persistent peers for example).

### DIFF
--- a/sei-tendermint/internal/p2p/conn_tracker_test.go
+++ b/sei-tendermint/internal/p2p/conn_tracker_test.go
@@ -1,27 +1,21 @@
 package p2p
 
 import (
-	"math"
-	"math/rand"
 	"net/netip"
+	"sync/atomic"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/require"
 )
 
-func randByte() byte {
-	return byte(rand.Intn(math.MaxUint8))
-}
-
-func randPort() uint16 {
-	return uint16(rand.Intn(math.MaxUint16))
-}
+var localAddrCounter atomic.Uint32
 
 func randLocalAddr() netip.AddrPort {
+	n := localAddrCounter.Add(1)
 	return netip.AddrPortFrom(
-		netip.AddrFrom4([4]byte{127, randByte(), randByte(), randByte()}),
-		randPort(),
+		netip.AddrFrom4([4]byte{127, byte(n >> 16), byte(n >> 8), byte(n)}),
+		uint16(n),
 	)
 }
 


sei-tendermint/internal/p2p/peermanager_pool.go
```diff
@@ -81,61 +71,73 @@ type peerConn interface {
...
 		}
-		toPrune, ok := failedAddr.Get()
-		if !ok {
-			return false
+		// Prune private address, if we insert a public address
+		if !old.isPublic && addr.IsPublic() {
+			p.addrs[pa.addr.NodeID] = pa
+			return nil
+		}
+		// Otherwise, do not replace
```
wen-coding, line 101: Hmm, what if my legitimate peer is moving to a new data center and i do want to replace it? I need to shut down the old one and wait for the address to be valid?

sei-tendermint/internal/p2p/peermanager_pool.go
```diff
@@ -81,61 +71,73 @@ type peerConn interface {
...
+	// It doesn't matter what was the time of the failure,
+	// since lastFail time is used just for round robin ordering.
+	for id, old := range p.addrs {
+		if old.lastFail.IsPresent() {
+			delete(p.addrs, id)
+			p.addrs[pa.addr.NodeID] = pa
+			return nil
+		}
+	}
+	// If the new address is public, find a private address to prune.
```
wen-coding, line 119: could it be the private address is actually a high staked validator I don't want to lose?

sei-tendermint/internal/p2p/peermanager_pool.go
pompon0: You can start a new node, then shut down the old node and wait for the network to prune the old address in favor of the new one. If you change the addresses regularly, you are most likely to use DNS anyway. There is little incentive to reuse node keys - persistent peers configuration requires people to configure the new address manually anyway.

sei-tendermint/internal/p2p/peermanager_pool.go
pompon0: If there is ANY peer address that you don't want to lose for ANY reason, then you should put it in persistent peers list. And the peer should add you to the unconditional list, otherwise there is no guarantee that it will have capacity for you.

PR #2992, opened 2026-02-27
Pex during handshake

pompon0:
This PR allows nodes to learn more node addresses, even if the peer they dial is out of capacity for new connections. This works by making listener node send pex batch as part of the handshake (it might discard the connection just after handshake, in case it decides it does not have capacity for this connection).

This new "pex in handshake" is enabled only if reactor pex is enabled in the node - disabling pex reactor prevents a node from learning new addresses and is currently mainly used to prevent node from connecting to random peers (which is misleading indirect use of pex flag), and this pr maintains this semantics to avoid distruptions. I think we should change this semantics and require people to just set MaxConnected to 0 instead (which is a direct way to say: connect only to persistent peers).

Additionally a SelfAddress is added to the handshake message: nodes advertise addresses of nodes they are connected to. Until now they could only advertise the addresses of outbound connections (i.e. verified addresses), but with this PR also SelfAddress of inbound connections is included (each node declares just their own up to date address, so it is fine to gossip it).

Note that SelfAddress could have been also be extracted from the pex response (it is always included), but I wanted to make it more explicit that it is special.


sei-tendermint/internal/p2p/pex/reactor.go
```diff
@@ -31,20 +31,13 @@ const (
...
 	maxAddressSize = 256
 
-	// max addresses returned by GetSelection
-	// NOTE: this must match "maxMsgSize"
-	maxGetSelection = 250
-
 	// NOTE: amplification factor!
 	// small request results in up to maxMsgSize response
-	maxMsgSize = maxAddressSize * maxGetSelection
+	maxMsgSize = 1000 + maxAddressSize*p2p.MaxPexAddrs
```
wen-coding, line 36: How big is the average msg now? Where does the constant 1000 come from?

sei-tendermint/internal/p2p/conv.go
```diff
@@ -48,16 +48,33 @@ var nodePublicKeyConv = utils.ProtoConv[NodePublicKey, *pb.NodePublicKey]{
...
+type handshakeMsg struct {
+	NodeAuth NodeChallengeSig
+	handshakeSpec
+}
+
 var handshakeMsgConv = protoutils.Conv[*handshakeMsg, *pb.Handshake]{
 	Encode: func(m *handshakeMsg) *pb.Handshake {
+		var selfAddr *string
+		if addr, ok := m.SelfAddr.Get(); ok {
+			selfAddr = utils.Alloc(addr.String())
```
wen-coding, line 66: When would it happen that you can't get selfAddr here?

sei-tendermint/internal/p2p/conv.go
```diff
@@ -70,9 +87,29 @@ var handshakeMsgConv = protoutils.Conv[*handshakeMsg, *pb.Handshake]{
 		if err != nil {
 			return nil, fmt.Errorf("NodeAuthSig: %w", err)
 		}
+		var selfAddr utils.Option[NodeAddress]
+		if p.SelfAddr != nil {
+			addr, err := ParseNodeAddress(*p.SelfAddr)
+			if err != nil {
+				return nil, fmt.Errorf("SelfAddr: %w", err)
```
wen-coding, line 94: Would this ever happen during normal operations? DNS failures?

sei-tendermint/internal/p2p/conv.go
```diff
@@ -70,9 +87,29 @@ var handshakeMsgConv = protoutils.Conv[*handshakeMsg, *pb.Handshake]{
...
+			}
+			selfAddr = utils.Some(addr)
+		}
+		pexAddrs := make([]NodeAddress, len(p.PexAddrs))
+		for i, addrString := range p.PexAddrs {
+			addr, err := ParseNodeAddress(addrString)
+			if err != nil {
+				return nil, fmt.Errorf("PexAddrs[%v]: %w", i, err)
```
wen-coding, line 102: If this may happen on a valid address, should we just ignore that one address and keep the others?

sei-tendermint/internal/p2p/conv.go
wen-coding, line 86: Can anyone with zero stake send us address updates?

sei-tendermint/internal/p2p/router.go
```diff
@@ -144,7 +147,8 @@ func (r *Router) Addresses(id types.NodeID) []NodeAddress {
 }
 
 func (r *Router) Advertise(maxAddrs int) []NodeAddress {
-	return r.peerManager.Advertise(maxAddrs)
+	addrs := r.peerManager.Advertise()
+	return addrs[:min(len(addrs), maxAddrs)]
```
wen-coding, line 151: nit: would we ever want randomly pick instead of always the front?

sei-tendermint/internal/p2p/router.go
```diff
@@ -217,7 +221,20 @@ func (r *Router) acceptPeersRoutine(ctx context.Context) error {
...
+					if r.options.PexOnHandshake {
+						pexAddrs = r.Advertise(MaxPexAddrs)
+					}
+					hConn, err := handshake(handshakeCtx, tcpConn, r.privKey, handshakeSpec{
+						SelfAddr: r.options.SelfAddress,
+						// Listener has to send pex data, so that dialer can learn about more peers in
+						// case listener does not have capacity for new connections.
+						// Dialer also could potentially send pex data, but there is no benefit from doing so:
+						// - if listener is full, then it won't use the new data. Listener also will not broadcast unverified data to anyone.
+						// - if it is not full, then the connection will be established and pex data will be sent the regular way using PEX protocol.
```
wen-coding, line 234: I'm confused, do you mean, if it  is not full, Listener may broadcast unverified data to others?

sei-tendermint/internal/p2p/conv.go
pompon0: node doesn't have to configure an external (public) address in case it doesn't have one.

sei-tendermint/internal/p2p/conv.go
pompon0: yes, node keys do not have stake assigned. They are not validator keys

sei-tendermint/internal/p2p/conv.go
pompon0: Adversary node can send broken data. This is just a proto converter though, it doesn't assume the proto to be valid in any sense, other than specified by the proto message definiton

sei-tendermint/internal/p2p/conv.go
pompon0: wdym? Valid address is parseable. This is not a dynamic property.

sei-tendermint/internal/p2p/router.go
pompon0: pex data is exchanged periodically both ways over every connection. But if the connection dies immediately after handshake, then there is no exchange obviously. Connection may die immediately after handshake if listener rejects it (maxconnected capacity reached), in which case it won't try to establish new connections any time soon anyway, so it doesn't need new peer addresses. Noone is sending unverified data to anyone.

sei-tendermint/internal/p2p/conv.go
wen-coding: Ah okay, if this is a static parser this should be safe.

sei-tendermint/internal/p2p/router.go
wen-coding: Okay, I mean, "Listener also will not broadcast unverified data to anyone." should always apply, not just the case "if listener is full". I was just confused by the comments.

sei-tendermint/internal/p2p/pex/reactor.go
pompon0: let's say 50 addresses * (42B for NodeID and ~20B for ip/dns address + 6B of protobuf overhead) = 3.5kB. This 1kB extra is arbitrary for whatever constant overhead. This estimation is not very precise.

sei-tendermint/internal/p2p/router.go
pompon0: We might want to. Currently the earlier addresses are of outbound connections, and they are preferred rn (by a rather weak argument that these are valid because we dialed them, while addresses of inbound connections are self-declared, and therefore potentially misconfigured). We can revisit later.

sei-tendermint/internal/p2p/router.go
pompon0: true. What I meant is that because listener is full, it won't connect to the new addresses, and therefore it won't gossip it further (listener won't use it, and he won't tell anyone about the new addresses either). I've updated the comment.

PR #2995, opened 2026-02-28
Extend DB benchmark to support state store (SS) backends

Kbhat1:
- Extend cryptosim DB benchmark to support state store (SS) backends via a lightweight StateStore → DBWrapper adapter

- Adds SSPebbleDB (raw MVCC) and SSComposite (full CompositeStateStore with EVM sub-DB routing) so the same ERC20 changeset workload can benchmark the historical state store

- Suppports mode for SC + SS

- running bench against various different configs

### DIFF
--- a/sei-db/state_db/bench/wrappers/composite_wrapper.go
+++ b/sei-db/state_db/bench/wrappers/composite_wrapper.go
@@ -34,7 +34,7 @@ func (c *compositeWrapper) LoadVersion(version int64) error {
 }
 
 func (c *compositeWrapper) Version() int64 {
-	return c.base.WorkingCommitInfo().Version
+	return c.base.Version()
 }
 
 func (c *compositeWrapper) Importer(version int64) (types.Importer, error) {


sei-db/state_db/bench/cryptosim/config/ss-composite-config.json
```diff
@@ -0,0 +1,27 @@
+{
+  "Comment": "State store benchmark using composite SS with EVM sub-stores (dual write + EVM-first read).",
+  "Backend": "SSComposite",
+  "BlocksPerCommit": 32,
```
cody-littley: I suggest removing the config entries that mirror the default configuration. I included all of the options in `basic-config.json` as an example of what was possible to set. But if we shadow default values elsewhere, we risk config drift when the default values inevitably change over time.

Currently, the only required field is `DataDir`, and any of the others will revert to the default value if not defined.

Kbhat1: Makes sense, updated

PR #3001, opened 2026-03-02
Fix lock issue in cacheMultiStore

yzang2019:
This PR is to fix a locking issue in CacheMultiStore

Added a unit test

### DIFF
--- a/sei-cosmos/storev2/rootmulti/store_test.go
+++ b/sei-cosmos/storev2/rootmulti/store_test.go
@@ -287,7 +287,7 @@ func TestCacheMultiStoreWithVersion_NoReentrantRLockDeadlock(t *testing.T) {
 	ssCfg := config.DefaultStateStoreConfig()
 	ssCfg.Enable = false
 
-	store := NewStore(home, log.NewNopLogger(), scCfg, ssCfg, false, []string{})
+	store := NewStore(home, log.NewNopLogger(), scCfg, ssCfg, []string{})
 	defer func() { _ = store.Close() }()
 
 	key := types.NewKVStoreKey("store1")


sei-cosmos/storev2/rootmulti/store_test.go
```diff
@@ -267,6 +267,62 @@ func TestQuery_HistoricalNoProofWithoutSS_UsesPermit(t *testing.T) {
...
+	<-writerReady
+
+	done := make(chan struct{})
+	go func() {
+		defer close(done)
+		// version=0 with SS disabled takes the else-if branch that previously
+		// called the public CacheMultiStore (which re-acquired RLock).
+		cms, err := store.CacheMultiStoreWithVersion(0)
+		require.NoError(t, err)
+		require.NotNil(t, cms)
```
masih: Avoid using `require` inside go routines. Instead use `assert`; the latter does not panic and let the test proceed and fail.

sei-cosmos/storev2/rootmulti/store_test.go
```diff
@@ -267,6 +267,62 @@ func TestQuery_HistoricalNoProofWithoutSS_UsesPermit(t *testing.T) {
 	require.Contains(t, resp.Log, "historical proof busy")
 }
 
+func TestCacheMultiStoreWithVersion_NoReentrantRLockDeadlock(t *testing.T) {
```
masih, line 283: I don't think this test actually exercises the deadlock:

The writer goroutine acquires store.mtx.Lock(), holds it for 50ms, then calls CacheMultiStoreWithVersion(0). With this ordering, the reader blocks on the first RLock() until the writer releases. 

So by the time the old code would attempt its second RLock() (via CacheMultiStore()), there's no pending writer left to cause a deadlock. This means the test can pass even with the buggy (double-lock) implementation.
To actually exercise the deadlock scenario, the writer needs to be waiting on Lock() while the reader already holds RLock(); i.e., the writer must be queued between the reader's first and second read-lock acquisitions.

yzang2019: Ah good point! Fixed the unit test and confirms the test will fail/deadlock when I use the old code

PR #3002, opened 2026-03-02
Add opts for scdual write

yzang2019:
Add flags/opts to allow setting dual write and dual read for SC, remove migration flag as well

Tested on RPC node

### DIFF
--- a/sei-db/state_db/bench/bench_ss_test.go
+++ b/sei-db/state_db/bench/bench_ss_test.go
@@ -1,0 +1,45 @@
+package bench
+
+import (
+	"testing"
+
+	"github.com/sei-protocol/sei-chain/sei-db/state_db/bench/wrappers"
+)
+
+func BenchmarkSSCompositeWrite(b *testing.B) {
+	const totalKeys int64 = 10_000
+
+	scenarios := []TestScenario{
+		{
+			Name:      "100_keys_per_block",
+			TotalKeys: totalKeys,
+			NumBlocks: totalKeys / 100,
+			Backend:   wrappers.SSComposite,
+		},
+	}
+
+	for _, scenario := range scenarios {
+		b.Run(scenario.Name, func(b *testing.B) {
+			runBenchmark(b, scenario, false)
+		})
+	}
+}
+
+func BenchmarkCombinedCompositeDualSSCompositeWrite(b *testing.B) {
+	const totalKeys int64 = 10_000
+
+	scenarios := []TestScenario{
+		{
+			Name:      "100_keys_per_block",
+			TotalKeys: totalKeys,
+			NumBlocks: totalKeys / 100,
+			Backend:   wrappers.CompositeDual_SSComposite,
+		},
+	}
+
+	for _, scenario := range scenarios {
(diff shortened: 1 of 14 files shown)


app/seidb.go
```diff
@@ -114,6 +104,8 @@ func parseSCConfigs(appOpts servertypes.AppOptions) config.StateCommitConfig {
 	scConfig.MemIAVLConfig.SnapshotWriterLimit = cast.ToInt(appOpts.Get(FlagSCSnapshotWriterLimit))
 	scConfig.MemIAVLConfig.SnapshotPrefetchThreshold = cast.ToFloat64(appOpts.Get(FlagSCSnapshotPrefetchThreshold))
 	scConfig.MemIAVLConfig.SnapshotWriteRateMBps = cast.ToInt(appOpts.Get(FlagSCSnapshotWriteRateMBps))
+	scConfig.WriteMode = config.WriteMode(cast.ToString(appOpts.Get(FlagSCWriteMode)))
```
blindchaser: considering learning from ss config parsing string: only overwrite if non-empty, validate, and panic on invalid values
```

if wm := cast.ToString(appOpts.Get(FlagEVMSSWriteMode)); wm != "" {
		parsedWM, err := config.ParseWriteMode(wm)
		if err != nil {
			panic(fmt.Sprintf("invalid EVM SS write mode %q: %s", wm, err))
		}
		ssConfig.WriteMode = parsedWM
	}
	if rm := cast.ToString(appOpts.Get(FlagEVMSSReadMode)); rm != "" {
		parsedRM, err := config.ParseReadMode(rm)
		if err != nil {
			panic(fmt.Sprintf("invalid EVM SS read mode %q: %s", rm, err))
		}
		ssConfig.ReadMode = parsedRM
	}
```

app/seidb.go
```diff
@@ -48,8 +47,6 @@ const (
 
 	// Other configs
 	FlagSnapshotInterval = "state-sync.snapshot-interval"
-	FlagMigrateIAVL      = "migrate-iavl"
```
blindchaser, line 51: in `cmd/seid/cmd/root.go -> addModuleInitFlags()`, --migrate-iavl and --migrate-height CLI flags orphaned

app/seidb.go
```diff
@@ -16,21 +16,20 @@ import (
 
 const (
 	// SC Store configs
-	FlagSCEnable                           = "state-commit.sc-enable"
-	FlagSCDirectory                        = "state-commit.sc-directory"
-	FlagSCAsyncCommitBuffer                = "state-commit.sc-async-commit-buffer"
-	FlagSCZeroCopy                         = "state-commit.sc-zero-copy"
```
blindchaser, line 22: docker config files (docker/localnode/config/app.toml, docker/rpcnode/config/app.toml) still reference sc-zero-copy and sc-cache-size. these are likely pre-existing and out of scope for this PR, but worth noting for cleanup

app/seidb.go
yzang2019: Good call, removing

app/seidb.go
yzang2019: Good point

PR #3003, opened 2026-03-02
fix(flatkv): fix state sync panic on nil DB handles during snapshot restore

blindchaser:
Issue:

rootmulti.Restore closes the SC store before calling Importer(), leaving all DB handles nil. The old code created a KVImporter against the closed store, which panicked in commitGlobalMetadata() on the nil metadataDB.

A second bug in composite.SnapshotImporter.AddModule() never set currentModule, so the EVM importer received zero nodes — masking the nil-DB crash until Close() reached commitGlobalMetadata().

Fixes:
- Reopen the store in Importer() when isClosed() is true
- Set currentModule in composite SnapshotImporter.AddModule()
- Move flatkv data dir under data/ to align with evm_ss convention
- Add isClosed() helper to avoid spreading the nil-check assumption

### DIFF
--- a/sei-db/state_db/sc/flatkv/store.go
+++ b/sei-db/state_db/sc/flatkv/store.go
@@ -57,9 +57,9 @@ type pendingAccountWrite struct {
 // CommitStore implements flatkv.Store for EVM state storage.
 // NOT thread-safe; callers must serialize all operations.
 type CommitStore struct {
-	log     logger.Logger
-	config  Config
-	dbDir string
+	log    logger.Logger
+	config Config
+	dbDir  string
 
 	// Five separate PebbleDB instances
 	metadataDB seidbtypes.KeyValueDB // Global version + LtHash watermark


sei-db/state_db/sc/flatkv/store.go
```diff
@@ -118,7 +118,7 @@ func NewCommitStore(homeDir string, log logger.Logger, cfg Config) *CommitStore
 }
 
 func (s *CommitStore) flatkvDir() string {
-	return filepath.Join(s.homeDir, flatkvRootDir)
+	return filepath.Join(s.homeDir, "data", flatkvRootDir)
```
yzang2019: We should do the appending to the dir when we initialize FlatKV in CompositeSC layer instead of changing the function here. We probably don't wanna change the test as well, just use whatever is passed in should be good ?

sei-db/state_db/sc/flatkv/store_lifecycle.go
```diff
@@ -7,6 +7,12 @@ import (
 	"github.com/sei-protocol/sei-chain/sei-db/state_db/sc/types"
 )
 
+// isClosed reports whether the store's DB handles have been released.
+func (s *CommitStore) isClosed() bool {
```
yzang2019, line 11: Shall we add an atomic bool just to make sure this is thread safe in the future?

blindchaser: closed flag alone is probably not sufficient since other store fields are still unsynchronized. I’d avoid that for now and revisit when/if we decide to make CommitStore concurrent-safe.

sei-db/state_db/sc/flatkv/store.go
blindchaser: thx updated

PR #3006, opened 2026-03-03
WAL utility tears itself down after the first error

cody-littley:

https://linear.app/seilabs/issue/STO-397/address-wal-feedback

Per feedback, the desired behavior of the WAL utility is that it should tear itself down after it encounters the first error.

Unit test coverage.

### DIFF
--- a/sei-db/state_db/sc/flatkv/lthash/lthash.go
+++ b/sei-db/state_db/sc/flatkv/lthash/lthash.go
@@ -76,6 +76,14 @@ func (l *LtHash) MixOut(other *LtHash) {
 	}
 }
 
+// Equal returns true if both LtHash vectors are identical.
+func (l *LtHash) Equal(other *LtHash) bool {
+	if other == nil {
+		return false
+	}
+	return l.limbs == other.limbs
+}
+
 // Clone returns a deep copy.
 func (l *LtHash) Clone() *LtHash {
 	clone := &LtHash{}
--- a/sei-db/state_db/sc/flatkv/lthash_correctness_test.go
+++ b/sei-db/state_db/sc/flatkv/lthash_correctness_test.go
@@ -1,0 +1,671 @@
+package flatkv
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"testing"
+
+	"github.com/sei-protocol/sei-chain/sei-db/common/evm"
+	"github.com/sei-protocol/sei-chain/sei-db/db_engine/types"
+	"github.com/sei-protocol/sei-chain/sei-db/proto"
+	"github.com/sei-protocol/sei-chain/sei-db/state_db/sc/flatkv/lthash"
+	iavl "github.com/sei-protocol/sei-chain/sei-iavl/proto"
+	"github.com/stretchr/testify/require"
+)
+
+// fullScanLtHash computes an LtHash from scratch by iterating every KV pair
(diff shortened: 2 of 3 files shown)


sei-db/wal/wal.go
```diff
@@ -341,13 +372,23 @@ func (walLog *WAL[T]) handleTruncate(req *truncateRequest) {
 		err = walLog.log.TruncateBack(req.index)
 	}
 	if err != nil {
-		req.errChan <- fmt.Errorf("failed to truncate: %w", err)
+		err = fmt.Errorf("failed to truncate: %w", err)
+		if strings.Contains(err.Error(), "out of range") {
```
blindchaser, line 376: in case tidwall/wal library changes its error message wording in the future, how about: 

```
if strings.Contains(err.Error(), "out of range") {
        req.errChan <- fmt.Errorf("failed to truncate: %w", err)
        return
    }
    walLog.reportFatalError(fmt.Errorf("failed to truncate: %w", err), req.errChan)
    return
```

sei-db/wal/wal.go
```diff
@@ -184,6 +196,18 @@ func (walLog *WAL[T]) Write(entry T) error {
...
+// reportFatalError records a fatal error and shuts down the WAL. err is the fatal error to report and store.
+// chanErr, if non-nil, is used to notify the caller of the request that triggered the error.
+func (walLog *WAL[T]) reportFatalError(err error, chanErr chan error) {
+	if chanErr != nil {
+		chanErr <- err
+	}
+	// Store on heap so the pointer remains valid after this function returns.
+	p := new(error)
+	*p = err
+	walLog.asyncError.Store(p)
```
blindchaser, line 208: should we call walLog.cancel() in the reportFatalError()?

sei-db/wal/wal.go
cody-littley: I wrapped the error inside the "out of range" block, but I'm not sure if that is what you were asking me to do. New code is below:


```
	if err != nil {
		err = fmt.Errorf("failed to truncate: %w", err)
		if strings.Contains(err.Error(), "out of range") {
			err = fmt.Errorf("out of range truncate error: %w", err)
			req.errChan <- err
			return
		}
		walLog.reportFatalError(err, req.errChan)
		return
	}
```

sei-db/wal/wal.go
cody-littley: Explicitly calling `cancel()` is not required. When `asyncError` is set, the loop exits, and immediately after the loop exits the context is cancelled.

```
	for running && walLog.asyncError.Load() == nil {
		select {
		case <-walLog.ctx.Done():
			running = false
		case req := <-walLog.writeChan:
			walLog.handleWrite(req)
		case req := <-walLog.truncateChan:
			walLog.handleTruncate(req)
		case <-pruneChan:
			walLog.prune()
		case <-walLog.closeReqChan:
			running = false
		}
	}

	walLog.cancel()
```

yzang2019, line 208: If we have multiple errors, will the later errors replace the previous one here? And are we OK with that?

cody-littley: Went back and checked, and there was one edge case where this was possible (i.e. during drain() when we are tearing down the system). I fixed this issue, and so now it should never be possible for `asyncError` to be set more than once.

PR #3010, opened 2026-03-03
Cody littley/db metrics

cody-littley:

Add lots of PebbleDB Metrics

image

Ran locally


sei-cosmos/storev2/rootmulti/store.go
```diff
@@ -90,7 +91,8 @@ func NewStore(
 	if scConfig.HistoricalProofRateLimit > 0 {
 		limiter = rate.NewLimiter(rate.Limit(scConfig.HistoricalProofRateLimit), burst)
 	}
-	scStore := composite.NewCompositeCommitStore(scDir, logger, scConfig)
+	ctx := context.Background()
```
yzang2019, line 94: Any specific reason for adding this context?

cody-littley: The new metrics utility collects metrics periodically in a background goroutine. I use the context to manage that goroutine's lifecycle, so it can tear itself down that goroutine when it is time to do so (e.g. at the end of a unit test).

sei-db/db_engine/pebbledb/db.go
```diff
@@ -80,7 +87,10 @@ func Open(path string, opts types.OpenOptions) (_ types.KeyValueDB, err error) {
 		return nil, err
 	}
 
-	return &pebbleDB{db: db}, nil
+	ctx, cancel := context.WithCancel(context.Background())
```
blindchaser: it looks we want to pass context.Context down through the store constructors, do we want to use `context.Background()` here?

cody-littley: Good point, I've piped a context into this method.

PR #3016, opened 2026-03-04
Deflake TestNodeStartStop under CI coverage load

masih:
- bind RPC to a random free port to avoid cross-test port collisions
- subscribe to NewBlock before starting the node to avoid missing early events
- increase block wait timeout from 10s to 30s for heavily loaded CI
- guard cleanup behind successful start to avoid cleanup deadlocks on start failure

### DIFF
--- a/sei-tendermint/node/node_test.go
+++ b/sei-tendermint/node/node_test.go
@@ -54,6 +54,7 @@ func newLocalNodeService(ctx context.Context, cfg *config.Config, logger log.Log
 func TestNodeStartStop(t *testing.T) {
 	cfg, err := config.ResetTestRoot(t.TempDir(), "node_node_test")
 	require.NoError(t, err)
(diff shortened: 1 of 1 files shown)


sei-tendermint/node/node_test.go
```diff
@@ -64,26 +65,33 @@ func TestNodeStartStop(t *testing.T) {
 
 	n, ok := ns.(*nodeImpl)
 	require.True(t, ok)
+	t.Cleanup(leaktest.CheckTimeout(t, time.Second))
+
+	started := false
```
pompon0, line 70: nit: you can just delay scheduling the cleanup instead.

masih: Aye; there will be more flaky test fixes. I will address this then. Thanks @pompon0

PR #3017, opened 2026-03-04
fix(giga): order fee validation checks correctly

arajasek:

This PR improves and cleans up the validation performed by the Giga executor before running a tx:
- Moves all validation into its own method
- Bumps nonce if and only if the nonce is correct
  - Basically "if nonce is correct, it will get bumped, regardless of any other potential failures"
- Orders the various validation checks to match v2, so that in the edge-case where multiple things fail, we respect v2's behaviour.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -19,6 +19,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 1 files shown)


arajasek:
Draft until tested on pacific today

PR #3023, opened 2026-03-05
Fix cryptosim metrics, QOL scripting upgrades

cody-littley:

Fix invalid pebble cryptosim metrics. We were incorrectly adding the total to the counter each iteration, as opposed to adding the change in the total to the counter.

This PR also contains several opportunistic quality of life improvements to the top level scripts (these were convenient when running local testing on this branch).

Ran locally.


sei-db/common/metrics/pebble_metrics.go
```diff
@@ -1131,7 +1188,11 @@ func (pm *PebbleMetrics) recordFromPebble(ctx context.Context) {
 	dbAttr := attribute.String("db", pm.databaseName)
 
 	if pm.compactionCount != nil {
-		pm.compactionCount.Add(ctx, m.Compact.Count, metric.WithAttributes(dbAttr))
+		delta := m.Compact.Count - pm.prevCompactionCount
```
blindchaser: how about have a `func addDelta()` to reuse code, something like

```
func addDelta(counter metric.Int64Counter, ctx context.Context, current int64, prev *int64, opts ...metric.AddOption) {
    delta := current - *prev
    *prev = current
    if delta > 0 {
        counter.Add(ctx, delta, opts...)
    }
}
```

cody-littley: Good idea, change made.

PR #3024, opened 2026-03-05
Remove dependency to zerolog in favour of slog

masih:
Redirect all usages of zerolog to tendermint logger interface temporarily which wraps slog.

Wherever zerolog default logger was used directly replace it with seilog


sei-tendermint/internal/autobahn/avail/app_votes.go
```diff
@@ -1,11 +1,14 @@
...
 
 import (
-	"github.com/rs/zerolog/log"
+	"log/slog"
 
 	"github.com/sei-protocol/sei-chain/sei-tendermint/internal/autobahn/types"
+	"github.com/sei-protocol/seilog"
 )
 
+var logger = seilog.NewLogger("tendermint", "internal", "autobahn", "avail")
```
wen-coding, line 10: nit: this might be a bit brittle if we somehow move things around later, is it possible to auto derive the file path in Go?

masih: It is possible but I would not recommend it. Because:
* dynamic inference would mean extra allocations at runtime, and often includes ugly package level init with panic on error or worse silent errors
* There may be operational scripts that depend on logger names which would break if logger names change or worse failures go un noticed.

The concept of "logger name" does not have to depend on Go package structure _at all_. It makes sense for it to be an operational overlay for the implementation picked based on architecturual components instead of implementation detail like go package structure. I have picked package structure here as the v0 just to avoid accidental duplicate loggers that may not belong.

My hope is that experts like yourself would change the logger names as you see fit to mean something architecturally rather than at package level.

For example, for this one I would probably pick the name `consensus/autobahn`.  The key thing to consider in picking names is: how much control is needed in setting logging level for all logs that use the same name. Because level is settable _per logger name_. So I would recommend finer grain naming if you ever envision the need to fine-tune logging levels to reduce noise.

PR #3026, opened 2026-03-05
Update trace queue wait by timeout

Kbhat1:
- Make trace RPC timeout apply to semaphore wait as well as trace execution
- Prevent queued trace requests from hanging indefinitely by making semaphore acquisition context-aware.
- `Add regression tests covering timeout while waiting for a slot and proper semaphore cleanup.

- Verifying on node

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,4 @@ integration_test/dapp_tests/artifacts
 
 # DB benchmark data
 sei-db/state_db/bench/cryptosim/data/**
+sei-db/state_db/bench/cryptosim/bin/
--- a/app/ante/evm_delivertx.go
+++ b/app/ante/evm_delivertx.go
@@ -90,6 +90,7 @@ func DecorateNonceCallback(ctx sdk.Context, ek *evmkeeper.Keeper, evmAddr common
 		// bump nonce if it is for some reason not incremented (e.g. ante failure)
 		if ek.GetNonce(callCtx, evmAddr) == startingNonce {
 			ek.SetNonce(callCtx, evmAddr, startingNonce+1)
+			ek.SetNonceBumped(callCtx)
 		}
 	})
 }
--- a/docker/localnode/Dockerfile
+++ b/docker/localnode/Dockerfile
@@ -1,10 +1,22 @@
+FROM golang:1.25.6-bookworm AS go-dist
+
 FROM ubuntu:latest
 ENV HOME="/root" PATH="/root/go/bin:/sei-protocol/sei-chain/integration_test/upgrade_module/scripts/:$PATH"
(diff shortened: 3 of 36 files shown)


evmrpc/tracers.go
```diff
@@ -48,12 +48,32 @@ type DebugAPI struct {
...
 // If the semaphore is nil (unlimited concurrency), it does nothing and returns a no-op release function.
-func (api *DebugAPI) acquireTraceSemaphore() func() {
+// The acquisition respects the provided context so callers do not block forever waiting for a slot.
+func (api *DebugAPI) acquireTraceSemaphore(ctx context.Context) (func(), error) {
 	if api.traceCallSemaphore != nil {
-		api.traceCallSemaphore <- struct{}{}
-		return func() { <-api.traceCallSemaphore }
+		select {
+		case api.traceCallSemaphore <- struct{}{}:
+			return func() { <-api.traceCallSemaphore }, nil
```
masih, line 64: We should check if the context is _not_ cancelled again here before returning. It is possible that we end up here after context cancellation since there is no guaranteed ordering for select statements in go.

Kbhat1: Updated, added a second context check after acquiring the slot so canceled requests do not slip through that race

evmrpc/tracers.go
```diff
@@ -48,12 +48,40 @@ type DebugAPI struct {
...
+			// If cancellation won the race at the same time as semaphore acquisition,
+			// release the slot and surface the context error.
+			select {
+			case <-ctx.Done():
+				<-api.traceCallSemaphore
+				return func() {}, ctx.Err()
+			default:
+			}
+			return func() { <-api.traceCallSemaphore }, nil
+		case <-ctx.Done():
```
yzang2019, line 65: It seems we are still not failing fast here?
In case where the semaphore is filled up with maxConcurrency, here we will blocks until the context times out, it is not fail-fast.

What we want here is to immediately return error without blocking/waiting when maxConcurrency is reached

evmrpc/tracers.go
masih: @Kbhat1 That change works but I think a simple if statement to check of ctx.Err is nil would be better. something like this:

 ```go
if ctx.Err() != nil {
   <-api.traceCallSemaphore
   return func() {}, ctx.Err()
}
return func() { <-api.traceCallSemaphore }, nil
```

evmrpc/tracers.go
Kbhat1: Discussed in a meeting. Updated

yzang2019 (approved):

evmrpc/tracers.go
Kbhat1: Updated

PR #3027, opened 2026-03-05
flatkv cache

cody-littley:

Add a caching layer to FlatKV, more than doubling performance in cryptosim benchmarks.

Unit tests, ran benchmark over several days.

### DIFF
--- a/sei-cosmos/server/config/config.go
+++ b/sei-cosmos/server/config/config.go
@@ -317,7 +317,7 @@ func DefaultConfig() *Config {
 			SnapshotDirectory:  "",
 		},
 		StateCommit: config.DefaultStateCommitConfig(),
-		StateStore:  *config.DefaultStateStoreConfig(),
+		StateStore:  config.DefaultStateStoreConfig(),
 		Genesis: GenesisConfig{
 			StreamImport:      false,
 			GenesisStreamFile: "",
--- a/sei-db/config/ss_config.go
+++ b/sei-db/config/ss_config.go
@@ -94,8 +94,8 @@ func (c StateStoreConfig) EVMEnabled() bool {
 }
 
 // DefaultStateStoreConfig returns the default StateStoreConfig
-func DefaultStateStoreConfig() *StateStoreConfig {
-	return &StateStoreConfig{
+func DefaultStateStoreConfig() StateStoreConfig {
+	return StateStoreConfig{
 		Enable:               true,
 		Backend:              DefaultSSBackend,
 		AsyncWriteBuffer:     DefaultSSAsyncBuffer,
--- a/sei-db/config/toml_test.go
+++ b/sei-db/config/toml_test.go
@@ -21,7 +21,7 @@ func TestStateCommitConfigTemplate(t *testing.T) {
 
 	cfg := TemplateConfig{
 		StateCommit: DefaultStateCommitConfig(),
-		StateStore:  *DefaultStateStoreConfig(),
+		StateStore:  DefaultStateStoreConfig(),
 	}
 
 	// Parse and execute the StateCommit template
@@ -64,7 +64,7 @@ func TestStateStoreConfigTemplate(t *testing.T) {
 
 	cfg := TemplateConfig{
 		StateCommit: DefaultStateCommitConfig(),
-		StateStore:  *DefaultStateStoreConfig(),
+		StateStore:  DefaultStateStoreConfig(),
 	}
 
 	// Parse and execute the StateStore template
@@ -127,7 +127,7 @@ func TestDefaultConfigTemplate(t *testing.T) {
 
 	cfg := TemplateConfig{
 		StateCommit:  DefaultStateCommitConfig(),
-		StateStore:   *DefaultStateStoreConfig(),
+		StateStore:   DefaultStateStoreConfig(),
 		ReceiptStore: DefaultReceiptStoreConfig(),
 	}
 
@@ -293,7 +293,7 @@ func TestTemplateFieldPathsExist(t *testing.T) {
 
 	cfg := TemplateConfig{
 		StateCommit:  DefaultStateCommitConfig(),
-		StateStore:   *DefaultStateStoreConfig(),
+		StateStore:   DefaultStateStoreConfig(),
 		ReceiptStore: DefaultReceiptStoreConfig(),
 	}
 
--- a/sei-db/db_engine/pebbledb/mvcc/bench_test.go
+++ b/sei-db/db_engine/pebbledb/mvcc/bench_test.go
@@ -11,7 +11,7 @@ import (
 func BenchmarkDBBackend(b *testing.B) {
 	s := &sstest.StorageBenchSuite{
 		NewDB: func(dir string) (types.StateStore, error) {
-			return OpenDB(dir, *config.DefaultStateStoreConfig())
+			return OpenDB(dir, config.DefaultStateStoreConfig())
 		},
 		BenchBackendName: "PebbleDB",
 	}
--- a/sei-db/db_engine/pebbledb/mvcc/db_test.go
+++ b/sei-db/db_engine/pebbledb/mvcc/db_test.go
@@ -11,7 +11,7 @@ import (
 )
 
 func TestStorageTestSuite(t *testing.T) {
-	pebbleConfig := *config.DefaultStateStoreConfig()
+	pebbleConfig := config.DefaultStateStoreConfig()
 	pebbleConfig.Backend = "pebbledb"
 	s := &sstest.StorageTestSuite{
 		BaseStorageTestSuite: sstest.BaseStorageTestSuite{
@@ -32,7 +32,7 @@ func TestStorageTestSuite(t *testing.T) {
 // configured for MVCC key encoding, so NextPrefix/SeekLT operations won't work correctly.
 // BaseStorageTestSuite contains only tests that work with both comparers.
 func TestStorageTestSuiteDefaultComparer(t *testing.T) {
-	pebbleConfig := *config.DefaultStateStoreConfig()
+	pebbleConfig := config.DefaultStateStoreConfig()
 	pebbleConfig.Backend = "pebbledb"
 	pebbleConfig.UseDefaultComparer = true
 
--- a/sei-db/ledger_db/receipt/receipt_store.go
+++ b/sei-db/ledger_db/receipt/receipt_store.go
@@ -124,7 +124,7 @@ func newReceiptBackend(config dbconfig.ReceiptStoreConfig, storeKey sdk.StoreKey
 		ssConfig.KeepLastVersion = false
 		ssConfig.Backend = "pebbledb"
 
-		db, err := mvcc.OpenDB(ssConfig.DBDirectory, *ssConfig)
+		db, err := mvcc.OpenDB(ssConfig.DBDirectory, ssConfig)
 		if err != nil {
 			return nil, err
 		}
--- a/sei-db/ledger_db/receipt/receipt_store_test.go
+++ b/sei-db/ledger_db/receipt/receipt_store_test.go
@@ -205,7 +205,7 @@ func TestRecoverReceiptStoreReplaysChangelog(t *testing.T) {
 	cfg := dbconfig.DefaultReceiptStoreConfig()
 	cfg.DBDirectory = dir
 	cfg.KeepRecent = 0
-	ssConfig := *dbconfig.DefaultStateStoreConfig()
+	ssConfig := dbconfig.DefaultStateStoreConfig()
 	ssConfig.DBDirectory = cfg.DBDirectory
 	ssConfig.KeepRecent = cfg.KeepRecent
 	if cfg.PruneIntervalSeconds > 0 {
--- a/sei-db/state_db/bench/wrappers/db_implementations.go
+++ b/sei-db/state_db/bench/wrappers/db_implementations.go
@@ -34,7 +34,7 @@ func DefaultBenchStateStoreConfig() *config.StateStoreConfig {
 	cfg.AsyncWriteBuffer = config.DefaultSSAsyncBuffer
 	cfg.WriteMode = config.SplitWrite
 	cfg.ReadMode = config.EVMFirstRead
-	return cfg
+	return &cfg
 }
 
 func newMemIAVLCommitStore(dbDir string) (DBWrapper, error) {
--- a/sei-db/state_db/ss/composite/recovery_test.go
+++ b/sei-db/state_db/ss/composite/recovery_test.go
@@ -35,7 +35,7 @@ func TestRecoverCompositeStateStore(t *testing.T) {
 	require.NoError(t, err)
 	defer os.RemoveAll(dir)
 
-	ssConfig := *config.DefaultStateStoreConfig()
+	ssConfig := config.DefaultStateStoreConfig()
 	ssConfig.Backend = "pebbledb"
 	dbHome := utils.GetStateStorePath(dir, ssConfig.Backend)
 	mvccDB, err := backend.ResolveBackend(ssConfig.Backend)(dbHome, ssConfig)
@@ -107,7 +107,7 @@ func TestSyncEVMStoreBehind(t *testing.T) {
 	require.NoError(t, err)
 	defer os.RemoveAll(dir)
 
-	ssConfig := *config.DefaultStateStoreConfig()
+	ssConfig := config.DefaultStateStoreConfig()
 	ssConfig.Backend = "pebbledb"
 	dbHome := utils.GetStateStorePath(dir, ssConfig.Backend)
 	mvccDB, err := backend.ResolveBackend(ssConfig.Backend)(dbHome, ssConfig)
@@ -219,7 +219,7 @@ func TestConstructorRecoversStalEVM(t *testing.T) {
 	require.NoError(t, err)
 	defer os.RemoveAll(dir)
 
-	ssConfig := *config.DefaultStateStoreConfig()
+	ssConfig := config.DefaultStateStoreConfig()
 	ssConfig.Backend = "pebbledb"
 	dbHome := utils.GetStateStorePath(dir, ssConfig.Backend)
 
--- a/sei-db/tools/cmd/seidb/benchmark/iteration.go
+++ b/sei-db/tools/cmd/seidb/benchmark/iteration.go
@@ -62,7 +62,7 @@ func DBIteration(inputKVDir string, numVersions int, outputDir string, dbBackend
 	fmt.Printf("Iterating Over DB at  %s\n", outputDir)
 	ssConfig := config.DefaultStateStoreConfig()
 	ssConfig.Backend = dbBackend
-	backend, err := ss.NewStateStore(outputDir, *ssConfig)
+	backend, err := ss.NewStateStore(outputDir, ssConfig)
 	if err != nil {
 		panic(err)
 	}
--- a/sei-db/tools/cmd/seidb/benchmark/random_read.go
+++ b/sei-db/tools/cmd/seidb/benchmark/random_read.go
@@ -67,7 +67,7 @@ func DBRandomRead(inputKVDir string, numVersions int, outputDir string, dbBacken
 	fmt.Printf("Reading Raw Keys and Values from %s\n", inputKVDir)
 	ssConfig := config.DefaultStateStoreConfig()
 	ssConfig.Backend = dbBackend
-	backend, err := ss.NewStateStore(outputDir, *ssConfig)
+	backend, err := ss.NewStateStore(outputDir, ssConfig)
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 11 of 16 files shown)


sei-db/state_db/sc/composite/store.go
```diff
@@ -68,11 +68,15 @@ func NewCompositeCommitStore(
...
+		cfg.FlatKVConfig.DataDir = filepath.Join(homeDir, "data", "flatkv")
+		var err error
+		store.evmCommitter, err = flatkv.NewCommitStore(ctx, cfg.FlatKVConfig)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create FlatKV commit store: %w", err)
+		}
 	}
 
-	return store
+	return store, nil
```
yzang2019: Instead of changing the interface to return an error, wouldn't it be easier to just panic inside this constructor? Is there any benefit to return the error if we failed to create the store?

sei-db/config/sc_config.go
```diff
@@ -46,7 +46,7 @@ type StateCommitConfig struct {
 	MemIAVLConfig memiavl.Config
 
 	// FlatKVConfig is the configuration for the FlatKV (EVM) backend
-	FlatKVConfig flatkv.Config
+	FlatKVConfig *flatkv.Config
```
yzang2019: Why do we need to make this a pointer?

sei-cosmos/storev2/rootmulti/store.go
```diff
@@ -92,10 +92,16 @@ func NewStore(
...
 	ctx := context.Background()
-	scStore := composite.NewCompositeCommitStore(ctx, scDir, scConfig)
+	scStore, err := composite.NewCompositeCommitStore(ctx, scDir, scConfig)
+	if err != nil {
+		panic(err)
+	}
 	if err := scStore.CleanupCrashArtifacts(); err != nil {
 		panic(err)
 	}
+	if err := scStore.CleanupCrashArtifacts(); err != nil {
```
yzang2019, line 96: Duplicate CleanupCrashArtifacts call here, we should remove the one above?

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -24,6 +24,16 @@ func (s *CommitStore) ApplyChangeSets(cs []*proto.NamedChangeSet) error {
...
 
+	s.phaseTimer.SetPhase("apply_change_sets_batch_read")
+
+	// Batch read all old values from DBs in parallel.
+	storageOld, accountOld, codeOld, legacyOld, err := s.batchReadOldValues(cs)
+	if err != nil {
+		return fmt.Errorf("failed to batch read old values: %w", err)
+	}
+
+	s.phaseTimer.SetPhase("apply_change_sets_prepare")
```
yzang2019: Duplicate code s.phaseTimer.SetPhase("apply_change_sets_prepare")

sei-db/db_engine/pebbledb/pebbledb_config.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/sei-protocol/sei-chain/sei-db/common/unit"
+)
+
+// Configuration for the PebbleDB database.
+type PebbleDBConfig struct {
```
yzang2019, line 9: These configs are not specific to PebbleDB, instead, they are specific to flatkv, I think we should move them to flatkv instead of put them here, we could change the db engine in the future for flatkv, but these configs are universal to all db engines

sei-db/db_engine/pebbledb/pebbledb_config.go
```diff
@@ -0,0 +1,56 @@
...
+	// The directory to store the database files. This has no default value and must be provided.
+	DataDir string
+	// The size of key-value cache, in bytes.
+	CacheSize uint64
+	// The number of shards in the key-value cache. Must be a power of two and greater than 0.
+	CacheShardCount uint64
+	// The size of pebbleDB's internal block cache, in bytes.
+	BlockCacheSize int
```
yzang2019: Would recommend remove this BlockCacheSize  from the config, we can set a tuned default value for now, we can add it in the future if we are 100% sure we need a different specific value for each DB.
In any way, we are not going rely on the PebbleDB internal cache for read optimization, so would recommend maybe just set a default number

sei-db/db_engine/pebbledb/db.go
```diff
@@ -27,29 +25,23 @@ type pebbleDB struct {
...
 
-// Open opens (or creates) a Pebble-backed DB at path, returning the DB interface.
+// Open opens (or creates) a Pebble-backed DB at path, returning a KeyValueDB
 func Open(
 	ctx context.Context,
-	path string,
-	opts types.OpenOptions,
-	enableMetrics bool,
+	config *PebbleDBConfig,
+	comparer *pebble.Comparer,
```
yzang2019: Do we have to pass the comparer in? Under what scenario will we pass in a custom comparer?

sei-db/state_db/sc/flatkv/config.go
```diff
@@ -33,15 +45,133 @@ type Config struct {
...
+// Validate checks that the configuration is sane and returns an error if it is not.
+func (c *Config) Validate() error {
+	if c.DataDir == "" {
+		return fmt.Errorf("data dir is required")
+	}
+	if c.AccountDBConfig.Validate() != nil {
+		return fmt.Errorf("account db config is invalid: %w", c.AccountDBConfig.Validate())
+	}
+	if c.CodeDBConfig.Validate() != nil {
+		return fmt.Errorf("code db config is invalid: %w", c.CodeDBConfig.Validate())
```
yzang2019: Nit: We shouldn't need to call Validate() twice here

sei-db/state_db/sc/flatkv/config.go
```diff
@@ -33,15 +45,133 @@ type Config struct {
...
+		return fmt.Errorf("storage db config is invalid: %w", c.StorageDBConfig.Validate())
+	}
+	if c.LegacyDBConfig.Validate() != nil {
+		return fmt.Errorf("legacy db config is invalid: %w", c.LegacyDBConfig.Validate())
+	}
+	if c.MetadataDBConfig.Validate() != nil {
+		return fmt.Errorf("metadata db config is invalid: %w", c.MetadataDBConfig.Validate())
+	}
+
+	if c.ReaderThreadsPerCore < 0 {
```
yzang2019: If ReaderThreadsPerCore == 0, and ReaderConstantThreadCount == 0, it will create a pool with 0 workers, causing deadlock issue?

sei-db/db_engine/pebbledb/pebbledb_config.go
```diff
@@ -0,0 +1,56 @@
...
+		EstimatedOverheadPerEntry: 256,
+	}
+}
+
+// Validates the configuration (basic sanity checks).
+func (c *PebbleDBConfig) Validate() error {
+	if c.DataDir == "" {
+		return fmt.Errorf("data dir is required")
+	}
+	if c.CacheSize > 0 && (c.CacheShardCount&(c.CacheShardCount-1)) != 0 {
```
yzang2019: Shall we also validate and make sure CacheShardCount > 0?

yzang2019:
One blocker comment: I think we should decouple cache from db_engine as much as possible, which allows FlatKV to switch different db_engine easily without reimplmenting the cache for each engine in the future

sei-db/state_db/sc/composite/store.go
cody-littley: Converted back to a `panic()`.

As a general pattern, I prefer to return errors when possible instead of using panic, except in cases where I'm just doing sanity/invariant testing. But this isn't something I feel strongly enough to argue about, so I don't mind switching this one back. 😜

sei-db/config/sc_config.go
cody-littley: Reverted back to a direct reference.

No need to make this a pointer, except that I prefer pointer types for any struct that has more than one or two fields. It's unlikely that this is going to be a major performance factor in this case, so I don't mind using a direct reference if that's your preferred pattern.

sei-cosmos/storev2/rootmulti/store.go
cody-littley: Removed duplicate call. This was probably happened when I merged upstream into my branch.

sei-db/state_db/sc/flatkv/store_write.go
cody-littley: Fixed.

sei-db/db_engine/pebbledb/pebbledb_config.go
cody-littley: Extracted configuration.

sei-db/db_engine/pebbledb/pebbledb_config.go
cody-littley: Removed in this PR and set to previous default, but I'd like to discuss adding it back in a future.

The block cache still serves an important role in database performance. It's used by pebble to cache LSM blocks, and we can't just disable it.

The reason why I added this config was because I thought the default value of 0.5gb was way too large for unit tests, which I generally prefer to use very little resources. Eventually I'd like to run unit tests in parallel, that will be tricky if each FlatKV instance utilizes 2.5gb memory.

Although I haven't done so yet, I'd actually like to enable configuration of ALL pebbleDB settings (although we can give them sane defaults). Right now it's tricky to experiment with pebble configuration because it's all hard coded.

Perhaps good for a quick in-person discussion.

sei-db/db_engine/pebbledb/db.go
cody-littley: Removed comparer parameter.

sei-db/state_db/sc/flatkv/config.go
cody-littley: fixed

sei-db/state_db/sc/flatkv/config.go
cody-littley: Good point. I've updated the pool constructor to more elegantly handle this case:

```
	if workers <= 0 {
		workers = 1
	}
```

sei-db/db_engine/pebbledb/pebbledb_config.go
cody-littley: Good point. Validation added.

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -430,10 +439,13 @@ func (s *CommitStore) commitBatches(version int64) error {
 	var wg sync.WaitGroup
 	wg.Add(len(pending))
 	for i, p := range pending {
-		go func(idx int, b types.Batch) {
-			defer wg.Done()
-			errs[idx] = b.Commit(syncOpt)
-		}(i, p.batch)
+		err := s.miscPool.Submit(s.ctx, func() {
```
yzang2019: This could a deadlock issue under load.

In commitBatches(), tasks are submitted to miscPool. Each task calls cachedBatch.Commit(), which calls cache.BatchSet(), which also submits to miscPool. If the pool queue is saturated (all workers executing batch commits), the inner BatchSet submissions will block waiting for a free slot — but slots can't free up until workers finish, and workers can't finish until BatchSet completes.

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -319,10 +325,13 @@ func (s *CommitStore) flushAllDBs() error {
 	var wg sync.WaitGroup
 	wg.Add(4)
 	for i, db := range []types.KeyValueDB{s.accountDB, s.codeDB, s.storageDB, s.legacyDB} {
-		go func(idx int, db types.KeyValueDB) {
-			defer wg.Done()
-			errs[idx] = db.Flush()
-		}(i, db)
+		err := s.miscPool.Submit(s.ctx, func() {
+			errs[i] = db.Flush()
+			wg.Done()
```
yzang2019: we should add defer for wg.Done() to avoid panic hangs forever

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -89,6 +95,7 @@ func (s *CommitStore) ApplyChangeSets(cs []*proto.NamedChangeSet) error {
 					return fmt.Errorf("invalid address length %d for key kind %d", len(keyBytes), kind)
 				}
 				addrStr := string(addr[:])
+				addrKey := string(AccountKey(addr))
```
yzang2019, line 97: If anyone ever changes AccountKey to add a prefix, or any transformation (which is a natural evolution for a "DB key builder" function), batchReadOldValues would silently fail to find pending writes in s.accountWrites

Suggest to pick one canonical key representation and use it everywhere — either always string(addr[:]) or always string(AccountKey(addr)).

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -452,3 +464,186 @@ func (s *CommitStore) commitBatches(version int64) error {
...
+				} else {
+					storageBatch[k] = types.BatchGetResult{}
+				}
+
+			case evm.EVMKeyNonce, evm.EVMKeyCodeHash:
+				addr, ok := AddressFromBytes(keyBytes)
+				if !ok {
+					continue
+				}
+				k := string(AccountKey(addr))
```
blindchaser: consider using `string(addr[:])` for the `s.accountWrites` lookup (matching ApplyChangeSets and store_read.go), and a separate `addrKey := string(AccountKey(addr))` for the `accountOld`/`accountBatch` maps. Today AccountKey returns addr[:] so the result is the same, but using two different derivations for the same map key is fragile if AccountKey ever changes.

sei-db/state_db/sc/flatkv/config.go
```diff
@@ -33,15 +46,148 @@ type Config struct {
...
+		return fmt.Errorf("storage db config is invalid: %w", err)
+	}
+	if err := c.LegacyDBConfig.Validate(); err != nil {
+		return fmt.Errorf("legacy db config is invalid: %w", err)
+	}
+	if err := c.MetadataDBConfig.Validate(); err != nil {
+		return fmt.Errorf("metadata db config is invalid: %w", err)
+	}
+
+	if c.ReaderThreadsPerCore < 0 {
```
blindchaser: nit: <=0 to match error text

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -452,3 +464,186 @@ func (s *CommitStore) commitBatches(version int64) error {
...
+	var wg sync.WaitGroup
+	var storageErr, accountErr, codeErr, legacyErr error
+
+	if len(storageBatch) > 0 {
+		wg.Add(1)
+		err = s.miscPool.Submit(s.ctx, func() {
+			defer wg.Done()
+			storageErr = s.storageDB.BatchGet(storageBatch)
+		})
+		if err != nil {
```
blindchaser: If a later Submit fails and we return early, previously submitted tasks may still be running and writing to their batch maps? how about calling wg.Wait() before returning on submit error to avoid a race.

sei-db/state_db/bench/cryptosim/cmd/configure-logger/main.go
```diff
@@ -38,6 +38,21 @@ func run() error {
 		return fmt.Errorf("load config: %w", err)
 	}
 
+	if cfg.LogDir == "" {
+		return fmt.Errorf("LogDir is empty, refusing to proceed")
+	}
+
+	if cfg.DeleteDataDirOnStartup {
```
blindchaser: In cmd/cryptosim DeleteDataDirOnStartup deletes DataDir, and in cmd/configure-logger it deletes LogDir, enabling this flag now has inconsistent

sei-db/db_engine/dbcache/cache_config.go
```diff
@@ -0,0 +1,43 @@
...
+func DefaultCacheConfig() CacheConfig {
+	return CacheConfig{
+		ShardCount:                8,
+		MaxSize:                   512 * unit.MB,
+		EstimatedOverheadPerEntry: DefaultEstimatedOverheadPerEntry,
+	}
+}
+
+// Validate checks that the configuration is sane and returns an error if it is not.
+func (c *CacheConfig) Validate() error {
```
blindchaser, line 35: is this function being used now?

sei-db/state_db/sc/flatkv/store_write.go
cody-littley: This is a good observation, but I've already taken precautions against this exact scenario. ;)

There several pool implementations:

- fixed: N worker threads, work only runs on these threads
- elastic: N worker threads, if worker thread is not immediately available then spin up new goroutine
- adhoc: each job gets its own goroutine (for unit tests, mostly)

The misc pool is using an elastic pool. This means that it's safe to send tasks with blocking dependencies, without fear of deadlock.

```
miscPool := threading.NewElasticPool(ctx, "flatkv-misc", miscPoolSize)
```

In the elastic pool's implementation, the work queue is a channel of size 0, meaning that if there isn't a worker currently sitting idle, we will immediately fall through to the default case in the select statement.

```
func (ep *elasticPool) Submit(ctx context.Context, task func()) error {
	if task == nil {
		return fmt.Errorf("elastic pool: nil task")
	}
	select {
	case <-ctx.Done():
		return ctx.Err()
	case <-ep.ctx.Done():
		return fmt.Errorf("elastic pool is shut down")
	case ep.workQueue <- task:
		return nil
	default:
		// All warm workers are busy; spawn a temporary goroutine.
		go task()
		return nil
	}
}
```

sei-db/state_db/sc/flatkv/store_write.go
cody-littley: Change made.

sei-db/state_db/sc/flatkv/store_write.go
cody-littley: Fixed.

sei-db/state_db/sc/flatkv/store_write.go
cody-littley: change made

sei-db/state_db/sc/flatkv/config.go
cody-littley: fixed

sei-db/state_db/sc/flatkv/store_write.go
cody-littley: Good point, although I think we can fix this with a slightly simpler solution.

Submit can only fail when contexts get cancelled, i.e. we should only expect to encounter this sort of error during system teardown workflows. So I don't think it's important to optimize performance here, just to make sure it's functionally correct.

The problem is that when we return, we're returning garbage map data. And even worse, we're returning garbage data that isn't threadsafe. 

I don't think we need to block until all goroutines are finished. I think the important part is to just always return nil values if we are returning an error. It's ok of the goroutine doesn't immediately stop, as long as the caller isn't receiving unsafe/invalid data.

I've converted error return cases to use the following form:

```
		if err != nil {
			return nil, nil, nil, nil, fmt.Errorf("failed to submit batch get: %w", err)
		}
```

sei-db/state_db/bench/cryptosim/cmd/configure-logger/main.go
cody-littley: Blame @masih for this one, configuring the logger at init time makes this sort of workflow wonky. 😜 

I've added seperate configurations for deleting log dirs, so now it should be easier to grok how the settings control the workflow.

sei-db/db_engine/dbcache/cache_config.go
cody-littley: Good catch, they weren't being called! I've added calls to these methods inside flatKV config.

```
	if err := c.AccountCacheConfig.Validate(); err != nil {
		return fmt.Errorf("account cache config is invalid: %w", err)
	}
	if err := c.CodeCacheConfig.Validate(); err != nil {
		return fmt.Errorf("code cache config is invalid: %w", err)
	}
	if err := c.StorageCacheConfig.Validate(); err != nil {
		return fmt.Errorf("storage cache config is invalid: %w", err)
	}
	if err := c.LegacyCacheConfig.Validate(); err != nil {
		return fmt.Errorf("legacy cache config is invalid: %w", err)
	}
	if err := c.MetadataCacheConfig.Validate(); err != nil {
		return fmt.Errorf("metadata cache config is invalid: %w", err)
	}
```

yzang2019 (approved):

sei-db/state_db/sc/flatkv/store_write.go
blindchaser: Agreed on the batchReadOldValues fix.

in func commitBatches(), there is a similar issue and I’d still lean toward switching back to plain goroutines, since there’s no returned map state to null out there. the main issue in that path is just the partial-submit/unwind interaction, and direct goroutines seem like the simplest way to avoid that.

cody-littley: As we discussed offline, I've changed the function signature so that `Submit()` never returns an error.

PR #3028, opened 2026-03-05
Parquet crash testing unit testing hooks

jewei1997:
This PR adds test-only fault-injection hooks to the parquet receipt store so we can simulate crashes at specific points in the write pipeline and validate recovery behavior. The hooks cover the key stages of persistence: after WAL write, before parquet flush, after parquet flush, after closing writers during file rotation, and after WAL clear during rotation.

It also adds a SimulateCrash() helper that intentionally abandons the store without the normal flush/finalization path, which lets the tests mimic abrupt process termination and then reopen the same store directory to verify recovery.

On top of that, this PR adds parquet receipt crash-recovery coverage that:

verifies recovery at each hook point, including file-rotation scenarios
runs randomized multi-crash stress tests to ensure WAL-committed blocks remain readable after reopen
verifies concurrent readers can still read committed receipts and logs while writes are artificially slowed
The goal is to increase confidence in parquet receipt durability and crash recovery behavior without changing normal production behavior outside of tests.

go test ./sei-db/ledger_db/receipt -run 'TestCrashRecoveryAtEachHookPoint|TestCrashRecoveryStress|TestSlowFlushWithConcurrentReads' -count=1

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -23,7 +23,7 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
-      - uses: actions/setup-go@v3
+      - uses: actions/setup-go@v6
         with:
           go-version: '1.25.6'
       # Initializes the CodeQL tools for scanning.
--- a/.github/workflows/eth_blocktests.yml
+++ b/.github/workflows/eth_blocktests.yml
@@ -44,7 +44,7 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Set up Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v6
         with:
           go-version: '1.25.6'
 
--- a/.github/workflows/golangci.yml
+++ b/.github/workflows/golangci.yml
@@ -16,7 +16,7 @@ jobs:
     name: lint
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/setup-go@v3
+      - uses: actions/setup-go@v6
         with:
           go-version: '1.25.6'
       - uses: actions/checkout@v3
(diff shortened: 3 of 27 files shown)


sei-db/ledger_db/parquet/store.go
```diff
@@ -240,6 +262,39 @@ func (s *Store) UpdateLatestVersion(version int64) {
...
+//   - flushLocked(): in-memory buffered receipts are lost
+//   - receiptWriter.Close() / logWriter.Close(): parquet footers are never
+//     written, leaving on-disk files corrupt/unreadable
+//   - receiptFile.Sync() / logFile.Sync(): OS-buffered writes may be lost
+//
+// The raw os.File.Close() and wal.Close() calls below exist only to release
+// file descriptors and locks so the test process can reopen the same directory.
+func (s *Store) SimulateCrash() {
+	if s.pruneStop != nil {
+		close(s.pruneStop)
```
blindchaser, line 277: set pruneStop to nil after close? otherwise Close() will do double close on a closed channel

sei-db/ledger_db/receipt/parquet_crash_test.go
```diff
@@ -0,0 +1,408 @@
...
+}
+
+// TestCrashRecoveryStress runs 30 randomized trials, each with 1-3 chained
+// crash-recover cycles. Each cycle writes a random number of blocks, crashes
+// at a randomly chosen non-rotation hook point, then reopens the store. After
+// all cycles complete, every block ever written (including crash blocks) must
+// be recoverable via WAL replay.
+func TestCrashRecoveryStress(t *testing.T) {
+	seed := int64(42)
+	t.Logf("random seed: %d (change to reproduce a specific run)", seed)
```
blindchaser, line 203: it looks the seed is always 42 not random?

PR #3029, opened 2026-03-05
fix: carry PrepareQC lock across consecutive timeouts in voteTimeout

### LINKED ISSUE #1: Node system requirements
Hello. What are the system requirements for installing this node? In the documentation https://docs.seinetwork.io/nodes-and-validators/joining-testnets this column is empty.

wen-coding:
## Summary

- **Safety bug fix**: `voteTimeout` only attached `i.PrepareQC` (current view) to TimeoutVotes. Since `pushTimeoutQC` clears `i.PrepareQC` on view change, consecutive timeouts (e.g. offline leader) would lose the PrepareQC lock. This produced a TimeoutQC with no `latestPrepareQC`, allowing a later leader to propose a conflicting block instead of being forced to repropose — breaking the safety invariant (two CommitQCs possible at the same RoadIndex).
- **Fix**: If no PrepareQC was observed in the current view (`!pqc.IsPresent()`), `voteTimeout` now inherits the one from `i.TimeoutQC.LatestPrepareQC()`. The lock propagates through the TimeoutQC chain indefinitely until a CommitQC advances the RoadIndex.
- **No new state or messages**: the fix reads an existing field (`TimeoutQC.LatestPrepareQC()`) that was already persisted; no schema or protocol changes needed.

### Attack scenario (before fix)

1. View (M, 0): PrepareQC forms for P. CommitVotes sent. Timer fires → TimeoutQC(M,0) carries the PrepareQC.
2. View (M, 1): `pushTimeoutQC` clears `i.PrepareQC`. Leader is offline. Timer fires → TimeoutVotes carry **no** PrepareQC. TimeoutQC(M,1) has `latestPrepareQC = None`.
3. View (M, 2): No reproposal forced. Leader proposes conflicting P'. CommitQC for P' forms. Old CommitVotes from step 1 also form CommitQC for P. **Two CommitQCs at RoadIndex M.**

### Why fix #1 alone is sufficient (no lock check in pushProposal needed)

With the fix, every honest validator's TimeoutVote carries the inherited PrepareQC. Any quorum of 2f+1 timeout votes includes ≥ f+1 honest validators carrying it, so `NewTimeoutQC` always selects it. `FullProposal.Verify` already enforces that if `TimeoutQC.reproposal()` returns a proposal, the leader must repropose it — so conflicting proposals are rejected by all honest validators.

## Test plan

Tests exercise the real `State.pushTimeoutQC`, `State.pushPrepareQC`, and `State.voteTimeout` methods via `scope.Run` + `s.Run(ctx)` (no mirrored production logic):

- [x] `TestVoteTimeoutPrepareQC_BothNone` — both None → None
- [x] `TestVoteTimeoutPrepareQC_OnlyCurrentView` — current view PrepareQC, no TimeoutQC → uses current
- [x] `TestVoteTimeoutPrepareQC_InheritedFromTimeoutQC` — core safety test: None + inherited → uses inherited, chains through 3 consecutive timeouts
- [x] `TestVoteTimeoutPrepareQC_CurrentViewHigherThanInherited` — reproposal succeeds (both present, current higher) → uses current
- [x] `TestVoteTimeoutPrepareQC_CurrentViewPresentInheritedNone` — current present, TimeoutQC has no PrepareQC → uses current
- [x] `TestVoteTimeoutPrepareQC_PersistedRestart` — restart regression: persisted TimeoutQC survives restart, voteTimeout inherits PrepareQC from it
- [x] `TestNewTimeoutQC_MixedPrepareQCs` — quorum-intersection: single vote carries PrepareQC among None votes → selected and Verify passes
- [x] `TestNewTimeoutQC_AllNone` — all votes carry None → LatestPrepareQC is None, Verify passes
- [x] `TestTimeoutQCVerify_HighestPrepareQCSelected` — votes carry PrepareQCs at different views → highest selected, Verify passes
- [x] Verified test fails when fix is reverted (`InheritedFromTimeoutQC` catches the bug)
- [x] All existing consensus and types tests pass

### DIFF
--- a/sei-tendermint/internal/autobahn/consensus/state_test.go
+++ b/sei-tendermint/internal/autobahn/consensus/state_test.go
@@ -259,3 +259,76 @@ func TestVoteTimeoutPrepareQC_CurrentViewPresentInheritedNone(t *testing.T) {
 	})
 	require.NoError(t, err)
 }
+
+// TestVoteTimeoutPrepareQC_PersistedRestart verifies that after a restart,
+// voteTimeout still inherits the PrepareQC from the persisted TimeoutQC.
+func TestVoteTimeoutPrepareQC_PersistedRestart(t *testing.T) {
+	rng := utils.TestRng()
+	_, keys := types.GenCommittee(rng, 3)
+	dir := t.TempDir()
+
+	makeCfg := func() *Config {
+		return &Config{
+			Key:                keys[0],
+			ViewTimeout:        func(types.View) time.Duration { return time.Hour },
+			PersistentStateDir: utils.Some(dir),
+		}
+	}
+	makeDataState := func() *data.State {
+		return data.NewState(
+			&data.Config{Committee: testCommittee(keys...)},
+			utils.None[data.BlockStore](),
+		)
+	}
+
+	view0 := types.View{Index: 0, Number: 0}
+	pqc0 := makePrepareQC(keys, types.GenProposalAt(rng, view0))
+
+	// Session 1: push PrepareQC + TimeoutQC, let runOutputs persist.
+	err := scope.Run(t.Context(), func(ctx context.Context, sc scope.Scope) error {
+		s, err := NewState(makeCfg(), makeDataState())
+		if err != nil {
+			return fmt.Errorf("NewState: %w", err)
+		}
+		sc.SpawnBg(func() error { return utils.IgnoreCancel(s.Run(ctx)) })
+
+		if err := s.pushPrepareQC(ctx, pqc0); err != nil {
+			return fmt.Errorf("pushPrepareQC: %w", err)
+		}
+		tqc0 := makeTimeoutQC(keys, view0, utils.Some(pqc0))
+		if err := s.pushTimeoutQC(ctx, tqc0); err != nil {
+			return fmt.Errorf("pushTimeoutQC: %w", err)
+		}
+		// Wait until runOutputs has processed the state change (and persisted it).
+		if _, err := s.myView.Wait(ctx, func(vs types.ViewSpec) bool {
+			return vs.TimeoutQC.IsPresent()
+		}); err != nil {
+			return fmt.Errorf("wait for persist: %w", err)
+		}
+		return nil
+	})
+	require.NoError(t, err)
+
+	// Session 2: restart from persisted state, verify PrepareQC inheritance.
+	err = scope.Run(t.Context(), func(ctx context.Context, sc scope.Scope) error {
+		s2, err := NewState(makeCfg(), makeDataState())
+		if err != nil {
+			return fmt.Errorf("NewState (restart): %w", err)
+		}
+		sc.SpawnBg(func() error { return utils.IgnoreCancel(s2.Run(ctx)) })
+
+		view1 := types.View{Index: 0, Number: 1}
+		if err := s2.voteTimeout(ctx, view1); err != nil {
+			return fmt.Errorf("voteTimeout: %w", err)
+		}
+		tv, ok := s2.innerRecv.Load().TimeoutVote.Get()
+		if !ok {
+			return fmt.Errorf("TimeoutVote not present after restart")
+		}
+		if !testTimeoutVotePrepareQC(tv).IsPresent() {
+			return fmt.Errorf("PrepareQC must be inherited from persisted TimeoutQC after restart")
+		}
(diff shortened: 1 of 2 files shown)


sei-tendermint/internal/autobahn/consensus/inner.go
```diff
@@ -155,7 +155,17 @@ func (s *State) voteTimeout(ctx context.Context, view types.View) error {
 		if i.View() != view || i.TimeoutVote.IsPresent() {
 			return nil
 		}
-		v := types.NewFullTimeoutVote(s.cfg.Key, view, i.PrepareQC)
+		// Attach the highest PrepareQC between the current view's and the one
+		// inherited from the TimeoutQC that justified entering this view.
+		// Without this, consecutive timeouts (e.g. offline leader) would lose
+		// the lock and allow a conflicting proposal to be committed.
+		pqc := i.PrepareQC
+		if tqc, ok := i.TimeoutQC.Get(); ok {
```
pompon0: nit
```
if tqc,ok := i.TimeoutQC.Get(); ok && !pqc.IsPresent() {
  pqc = tqc.LastestPrepareQC
}
```

sei-tendermint/internal/autobahn/consensus/inner_test.go
```diff
@@ -920,3 +920,151 @@ func TestPushTimeoutQCClearsStaleState(t *testing.T) {
...
+// attach the given PrepareQC.
+func makeTimeoutQC(keys []types.SecretKey, view types.View, pqc utils.Option[*types.PrepareQC]) *types.TimeoutQC {
+	votes := make([]*types.FullTimeoutVote, len(keys))
+	for i, k := range keys {
+		votes[i] = types.NewFullTimeoutVote(k, view, pqc)
+	}
+	return types.NewTimeoutQC(votes)
+}
+
+// simulatePushTimeoutQC mirrors the production pushTimeoutQC: keeps CommitQC
```
pompon0: can you simulate it on State rather than inner, so that we don't have to "mirror" anything?

sei-tendermint/internal/autobahn/consensus/inner.go
wen-coding: done

sei-tendermint/internal/autobahn/consensus/inner_test.go
wen-coding: done

sei-tendermint/internal/autobahn/consensus/state_test.go
```diff
@@ -0,0 +1,197 @@
...
+		PersistentStateDir: utils.None[string](),
+	}, dataState)
+	require.NoError(t, err)
+	return s
+}
+
+// startRunOutputs launches runOutputs in the background and returns
+// a context that is cancelled when the test finishes.
+func startRunOutputs(t *testing.T, s *State) context.Context {
+	ctx, cancel := context.WithCancel(t.Context())
```
pompon0: this doesn't make sense. t.Context() is always canceled when the test finishes.
Also, why not s.Run()? The fact that s.runOutputs is suffcient rn is an implementation detail that the test should not need to rely on.
Also the s.Run() output should be checked for (non-cancellation) error, so I'd recommend `scope.Run(func(){ s.SpawnBg(s.Run()) })`

wen-coding: changed

sei-tendermint/internal/autobahn/consensus/state_test.go
```diff
@@ -0,0 +1,261 @@
...
+			return fmt.Errorf("PrepareQC should be present")
+		}
+		pqc, _ := gotPQC.Get()
+		if pqc.Proposal().View() != view1 {
+			return fmt.Errorf("expected PrepareQC at view %v, got %v", view1, pqc.Proposal().View())
+		}
+		return nil
+	})
+	require.NoError(t, err)
+}
```
masih, line 261: Not a blocker: It would be great to also add a test cases for 
* the exact quorum-intersection behavior in `NewTimeoutQC` plus `TimeoutQC.Verify`.
* the restart regression test with persisted `TimeoutQC` just to verify `voteTimeout` is carried over to prepare.

wen-coding: done

PR #3031, opened 2026-03-06
Source Go toolchain from official image and harden foundry install

masih:
- Remove inline curl/tar Go download from localnode and rpcnode Dockerfiles
- Use multi-stage build with golang:1.25.6-bookworm and copy /usr/local/go
- Eliminates flaky external tarball fetch during docker build
- Harden foundry download to retry on failures as it often flakes

Flaked on [main](https://github.com/sei-protocol/sei-chain/actions/runs/22683804651/job/65761178703).

### DIFF
--- a/docker/localnode/Dockerfile
+++ b/docker/localnode/Dockerfile
@@ -4,9 +4,19 @@ FROM ubuntu:latest
 ENV HOME="/root" PATH="/root/go/bin:/sei-protocol/sei-chain/integration_test/upgrade_module/scripts/:$PATH"
 RUN apt-get update && apt-get install -y make build-essential git jq python3 curl vim uuid-runtime
 COPY --from=go-dist /usr/local/go /usr/local/go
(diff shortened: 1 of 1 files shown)


docker/rpcnode/Dockerfile
```diff
@@ -1,8 +1,10 @@
+FROM golang:1.25.6-bookworm AS go-dist
```
mojtaba-esk, line 1: Great idea

docker/rpcnode/Dockerfile
```diff
@@ -1,8 +1,10 @@
...
+
 FROM ubuntu:latest
```
mojtaba-esk, line 3: I know it is out of the scope of this PR, just want to share that I am not sure if using `latest` tag is the best idea here. To keep away from flakiness, we probably should use a fixed working version.

masih: yup, also i do not love piping whatever fundary hosts into `bash` 😅

PR #3095, opened 2026-03-20
fix: accept legacy CommitHash and prevent mmap use-after-free for v6.3->v6.4 upgrade

masih:

Two fixes for deploying release/v6.4 on nodes previously running release/v6.3:

### Fix 1: LastCommitHash mismatch panic on startup

The `Commit.Hash()` algorithm was changed in CON-76 (#2600) to include Height, Round, and BlockID in the Merkle tree alongside signatures. Blocks stored by v6.3 nodes have `LastCommitHash` computed with the old signatures-only algorithm. When v6.4 loads these blocks during ABCI handshake replay, `ValidateBasic` recomputes the hash with the new algorithm, producing a different value and causing a panic:

```
panic: error from proto block: wrong Header.LastCommitHash. Expected 75AA2EF6..., got E3CCCADC...
```

Adds `Commit.LegacyHash()` that uses the old (signatures-only) algorithm, and updates `Block.ValidateBasic()` to fall back to it when the new hash doesn't match.

### Fix 2: SIGSEGV in historical RPC queries (use-after-free of mmap'd memory)

When serving historical queries, `rootmulti.Store.Query()` opens a read-only memiavl instance (mmap'd), queries it, and defers `Close()`. The response byte slices (`Key`, `Value`) may point directly into the mmap'd region. When the deferred `Close()` unmaps the memory before JSON marshaling reads the response, the process segfaults:

```
fatal error: fault
[signal SIGSEGV: segmentation violation code=0x1 addr=0x7fd224098d0a]
encoding/base64.(*Encoding).Encode(...)
```

Copies `res.Key` and `res.Value` for non-latest queries so they survive the mmap unmap.

### Note on PR #3093

The IBC Upgrade Params fix from #3093 is already included in this branch since it was merged into `release/v6.4` before this branch was created.

- `go test ./sei-tendermint/types/` — all tests pass including `TestCommitHash` and `TestBlockValidateBasic`
- `go test ./sei-tendermint/internal/store/` — all store tests pass
- Verified that tampered commits are still correctly rejected (neither new nor legacy hash matches)
- Fix 1 confirmed working on arctic-1 RPC node (handshake completed successfully)

<div><a href="https://cursor.com/agents/bc-e0b2fa4c-bfe9-4cfc-8420-d63c3ac1c87b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-e0b2fa4c-bfe9-4cfc-8420-d63c3ac1c87b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/sei-cosmos/storev2/rootmulti/store.go
+++ b/sei-cosmos/storev2/rootmulti/store.go
@@ -638,6 +638,19 @@ func (rs *Store) Query(req abci.RequestQuery) abci.ResponseQuery {
 
 	res := store.Query(req)
 
+	// When serving from a historical read-only memiavl instance, the response
+	// byte slices (Key, Value) may point directly into mmap'd memory that will
+	// be unmapped when the deferred scStore.Close() runs. Copy them so the
+	// caller doesn't hit a use-after-free segfault during JSON marshaling.
+	if !latest {
+		if len(res.Key) > 0 {
+			res.Key = append([]byte{}, res.Key...)
+		}
+		if len(res.Value) > 0 {
+			res.Value = append([]byte{}, res.Value...)
+		}
+	}
+
 	// If underlying query failed (e.g. invalid height/path) or doesn' need proof, return as-is.
 	if res.Code != 0 || !needProof {
 		return res
--- a/sei-tendermint/types/block.go
+++ b/sei-tendermint/types/block.go
@@ -89,7 +89,9 @@ func (b *Block) ValidateBasic() error {
 	}
 
 	if w, g := b.LastCommit.Hash(), b.LastCommitHash; !bytes.Equal(w, g) {
-		return fmt.Errorf("wrong Header.LastCommitHash. Expected %X, got %X", w, g)
+		if wLegacy := b.LastCommit.LegacyHash(); !bytes.Equal(wLegacy, g) {
+			return fmt.Errorf("wrong Header.LastCommitHash. Expected %X, got %X", w, g)
+		}
 	}
 
 	// NOTE: b.Data.Txs may be nil, but b.Data.Hash() still works fine.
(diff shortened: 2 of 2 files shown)


jsand11:
// File: sei-db/common/safe.go  (recommended location)
// Package common provides shared utility functions used across sei-db modules.

package common

import "github.com/cosmos/cosmos-sdk/store/types"

// SafeCopySlice creates a new independent copy of the input byte slice.
// Returns nil if the input is nil.
//
// This function should be used when dealing with byte slices that originate
// from mmap'd memory regions (e.g. MemIAVL / state store queries) to prevent
// use-after-free bugs when the underlying memory is unmapped (e.g. after
// defer store.Close()).
//
// Example usage in Query methods:
//
//	res.Key   = common.SafeCopySlice(res.Key)
//	res.Value = common.SafeCopySlice(res.Value)
func SafeCopySlice(b []byte) []byte {
	if b == nil {
		return nil
	}
	c := make([]byte, len(b))
	copy(c, b)
	return c
}

// SafeCopyStoreValue is a convenience wrapper that copies both Key and Value
// fields of a types.StoreKVPair (used in some iterator / proof paths).
func SafeCopyStoreValue(kv types.StoreKVPair) types.StoreKVPair {
	return types.StoreKVPair{
		Key:   SafeCopySlice(kv.Key),
		Value: SafeCopySlice(kv.Value),
	}
}

// SafeCopyBytes is an alias for SafeCopySlice — use whichever name feels
// more readable in context.
func SafeCopyBytes(b []byte) []byte {
	return SafeCopySlice(b)
}

PR #3155, opened 2026-03-31
Restructure sei data folder for Giga

yzang2019:
The current data folder is not intuitive and not very clean for Giga storage. It currently looks like this:
image

The goal of this PR is to restructure the data folder for Giga into this new structure while also making it backward compatible:
image

Added unit test to make sure it works for both new nodes and existing nodes which has the old path


sei-tendermint/config/db.go
```diff
@@ -14,8 +17,42 @@ type DBContext struct {
...
+		return baseDir
+	}
+	legacyPath := filepath.Join(baseDir, dbID+".db")
+	if pathExists(legacyPath) {
+		return baseDir
+	}
+	return filepath.Join(baseDir, subDir)
+}
+
+func pathExists(path string) bool {
```
blindchaser: dup `func pathExists`

sei-tendermint/cmd/tendermint/commands/reindex_event.go
```diff
@@ -149,23 +149,23 @@ func loadEventSinks(cfg *tmcfg.Config) ([]indexer.EventSink, error) {
 func loadStateAndBlockStore(cfg *tmcfg.Config) (*store.BlockStore, state.Store, error) {
 	dbType := dbm.BackendType(cfg.DBBackend)
 
-	if !os.FileExists(filepath.Join(cfg.DBDir(), "blockstore.db")) {
+	blockstoreDir := tmcfg.ResolveDBDir("blockstore", cfg.DBDir())
+	if !os.FileExists(filepath.Join(blockstoreDir, "blockstore.db")) {
 		return nil, nil, fmt.Errorf("no blockstore found in %v", cfg.DBDir())
```
blindchaser: the error msg seems to should be against blockstoreDir instead of cfg.DBDir()

sei-tendermint/cmd/tendermint/commands/reindex_event.go
```diff
@@ -149,23 +149,23 @@ func loadEventSinks(cfg *tmcfg.Config) ([]indexer.EventSink, error) {
...
 	if err != nil {
 		return nil, nil, err
 	}
 	blockStore := store.NewBlockStore(blockStoreDB)
 
-	if !os.FileExists(filepath.Join(cfg.DBDir(), "state.db")) {
-		return nil, nil, fmt.Errorf("no blockstore found in %v", cfg.DBDir())
+	stateDir := tmcfg.ResolveDBDir("state", cfg.DBDir())
+	if !os.FileExists(filepath.Join(stateDir, "state.db")) {
+		return nil, nil, fmt.Errorf("no state store found in %v", cfg.DBDir())
```
blindchaser: same as above, error msg against stateDir instead of cfg.DBDir()

sei-tendermint/cmd/tendermint/commands/reset.go
```diff
@@ -110,53 +110,34 @@ func ResetAll(dbDir, privValKeyFile, privValStateFile string, keyType string, ho
...
-	if tmos.FileExists(txIndex) {
-		if err := os.RemoveAll(txIndex); err == nil {
-			logger.Info("Removed tx_index.db", "dir", txIndex)
-		} else {
-			logger.Error("error removing tx_index.db", "dir", txIndex, "err", err)
-		}
-	}
+// ResetState removes all blocks, tendermint state, indexed transactions and evidence.
+// It handles both the legacy flat layout and the new subdirectory layout.
+func ResetState(dbDir string) error {
```
blindchaser, line 126: consider updating reset related unit testings

app/receipt_store_config.go
```diff
@@ -20,7 +20,7 @@ func readReceiptStoreConfig(homePath string, appOpts seidbconfig.AppOptions) (se
 		return receiptConfig, err
 	}
 	if receiptConfig.DBDirectory == "" {
-		receiptConfig.DBDirectory = filepath.Join(homePath, "data", "receipt.db")
+		receiptConfig.DBDirectory = utils.GetReceiptStorePath(homePath)
```
blindchaser: readReceiptStoreConfig() was changed to default to data/ledger/receipt.db, but the emitted config template still says <home>/data/receipt.db

need to update the related toml.go

blindchaser (approved):

Kbhat1 (approved):

sei-tendermint/config/db.go
yzang2019: You mean this one and the one in sei-db? I think this is fine we don't wanna couple sei-tendemrint and sei-db. Tendermint manages their own db separately than sei-db

sei-tendermint/cmd/tendermint/commands/reindex_event.go
yzang2019: good catch!

sei-tendermint/cmd/tendermint/commands/reindex_event.go
yzang2019: good catch!

sei-tendermint/cmd/tendermint/commands/reset.go
yzang2019: Added

app/receipt_store_config.go
yzang2019: Fixed

sei-db/common/utils/path.go
```diff
@@ -1,13 +1,69 @@
...
+	return filepath.Join(homePath, "data", "state_commit", "flatkv")
+}
+
+// GetStateStorePath returns the path for the Cosmos state store (SS).
+// New nodes use data/state_store/cosmos/{backend}; existing nodes with
+// data/{backend} continue using the legacy path for backward compatibility.
 func GetStateStorePath(homePath string, backend string) string {
-	return filepath.Join(homePath, "data", backend)
+	legacyPath := filepath.Join(homePath, "data", backend)
+	if PathExists(legacyPath) {
```
masih: Minor: Is there ever a case where an abrupt node failure could cause the creation of an empty directory here?

If so, for robustness i recommend checking if this dir is empty and proceed with legacy path iff it is not empty.

sei-db/common/utils/path.go
```diff
@@ -1,13 +1,69 @@
...
+import (
+	"os"
+	"path/filepath"
+)
 
-func GetCommitStorePath(homePath string) string {
-	return filepath.Join(homePath, "data", "committer.db")
+// PathExists returns true if the given path exists on disk.
+func PathExists(path string) bool {
+	_, err := os.Stat(path)
```
masih: Harden this by requiring the type of path? e.g. directory? file etc.

This is racy in that it doesn't atomically check and it repeatedly checks but hopefully the probability of race is low enough that we can accept the race condition as negligible.

masih (approved):
No blockers 🚀

cody-littley (approved):
Does this PR handle migration (i.e. moving memiavl data from original location to new location)?

jewei1997 (approved):

sei-db/common/utils/path.go
yzang2019: If the creation of empty dir happens, then it will continue proceed with the old path, which is fine. I don't think there's any case where it is started with the new path first and then suddenly crash and create an old path

sei-db/common/utils/path.go
yzang2019: Make sense, will do

PR #3165, opened 2026-04-01
perf: Build the Ethereum chain config once per fee check

amir-deris:
There are areas we build the ChainConfig twice and it can simply be built once and reused to improve performance.


masih (approved):

bdchatham (approved):
LGTM

PR #3216, opened 2026-04-08
Include receipt store earliest height in watermark bounds to prevent silent undercount on pruned blocks

amir-deris:
## Problem

There was a fix implemented in [PR 3125](https://github.com/sei-protocol/sei-chain/pull/3125) in order to make `eth_getBlockTransactionCountByNumber == len(eth_getBlockByNumber.transactions)` by using same filtering mechanism for evm transactions to consider availability of receipts. That fix raised the following concern:

This change makes [GetBlockTransactionCountByNumber](https://github.com/sei-protocol/sei-chain/blob/292f78c514eadbdb610b0797f114cae0b8b65b28/evmrpc/block.go#L167) and [GetBlockTransactionCountByHash](https://github.com/sei-protocol/sei-chain/blob/292f78c514eadbdb610b0797f114cae0b8b65b28/evmrpc/block.go#L184) results dependent on the receipt store due to the [receipt query](https://github.com/sei-protocol/sei-chain/blob/292f78c514eadbdb610b0797f114cae0b8b65b28/evmrpc/block.go#L544) leading to inconsistencies in historical results.
Those functions do [ensure](https://github.com/sei-protocol/sei-chain/blob/292f78c514eadbdb610b0797f114cae0b8b65b28/evmrpc/block.go#L190) that the block is available based on [watermarks](https://github.com/sei-protocol/sei-chain/blob/292f78c514eadbdb610b0797f114cae0b8b65b28/evmrpc/watermark_manager.go#L267). However, the receipt store [does not contribute](https://github.com/sei-protocol/sei-chain/blob/292f78c514eadbdb610b0797f114cae0b8b65b28/evmrpc/watermark_manager.go#L113-L118) to the earliest height in the watermark manager, only the latest. So even if [EnsureBlockHeightAvailable](https://github.com/sei-protocol/sei-chain/blob/292f78c514eadbdb610b0797f114cae0b8b65b28/evmrpc/watermark_manager.go#L214) returns that the block is available that does not guarantee all receipts for that height are available. This will occur in the cases when [ReceiptStore's KeepRecent](https://github.com/sei-protocol/sei-chain/blob/292f78c514eadbdb610b0797f114cae0b8b65b28/sei-db/config/receipt_config.go#L46) is less than the [StateStore's KeepRecent](https://github.com/sei-protocol/sei-chain/blob/292f78c514eadbdb610b0797f114cae0b8b65b28/sei-db/config/ss_config.go#L42) and [MinRetainBlocks](https://github.com/sei-protocol/sei-chain/blob/292f78c514eadbdb610b0797f114cae0b8b65b28/sei-cosmos/server/config/config.go#L89). In such cases the returned transaction count may diverge from the true count due to pruned receipts.

## Analysis

`eth_getBlockTransactionCountByNumber` and `eth_getBlockTransactionCountByHash` silently return an incorrect (lower) transaction count for historical blocks when the receipt store has been pruned further back than the block or state stores.

The watermark check in `EnsureBlockHeightAvailable` uses `blockEarliest` (derived from Tendermint) to gate requests, but the receipt store's earliest version never contributed to that bound. So a block could pass the availability check while its receipts were already pruned — causing `countBlockTxsLikeEncodeTmBlock` to silently skip EVM transactions whose receipts are missing rather than returning an error.

This scenario occurs when `ReceiptStoreConfig.KeepRecent` is smaller than `StateStoreConfig.KeepRecent` / `MinRetainBlocks`.

## Fix

- Added `EarliestVersion() int64` to the `ReceiptStore` interface and implemented it across all backends (`receiptStore`/pebble, `parquetReceiptStore`, `cachedReceiptStore`) and the underlying `parquet.Store`.
- Added `WatermarkManager.EnsureReceiptHeightAvailable` — a targeted check that returns an explicit "receipts have been pruned" error when the requested height is below the receipt store's earliest retained version. Returns `nil` when no receipt store is configured or no pruning has occurred.
- Called `EnsureReceiptHeightAvailable` in both `GetBlockTransactionCountByNumber` and `GetBlockTransactionCountByHash` after block resolution, so callers receive a clear error instead of a silently wrong count.

## Result

- Within receipt retention window: correct count (consistent with `eth_getBlockByNumber`)
- Outside receipt retention window: explicit pruned error (instead of silent undercount)

Fixes: PLT-256


evmrpc/watermark_manager.go
```diff
@@ -219,6 +219,24 @@ func (m *WatermarkManager) EnsureBlockHeightAvailable(ctx context.Context, heigh
...
+// EnsureBlockHeightAvailable because the receipt store can be configured with a
+// smaller KeepRecent than the block or state stores.
+func (m *WatermarkManager) EnsureReceiptHeightAvailable(_ context.Context, height int64) error {
+	if m.receiptStore == nil {
+		return nil
+	}
+	earliest := m.receiptStore.EarliestVersion()
+	if earliest <= 0 {
+		return nil
+	}
```
bdchatham: Trying to wrap my head around this. Is there something special about this? Is it not generalized in the if-statement below it?

amir-deris: Thanks for feedback. That is a good point as the if statement below will not cause error since height is non-negative. I will remove it.

bdchatham (approved):

sei-db/ledger_db/parquet/store.go
```diff
@@ -181,6 +181,11 @@ func (s *Store) SetEarliestVersion(version int64) {
 	s.earliestVersion.Store(version)
 }
 
+// EarliestVersion returns the earliest version retained in the store.
+func (s *Store) EarliestVersion() int64 {
+	return s.earliestVersion.Load()
```
masih, line 186: Looks like `earliestVersion` is never changed in parquet implementation even after pruning?



CC @jewei1997 for context.

masih (commented):
* We also need to update GetBlockByNumber/Hash otherwise the behaviour between GetBlockTransactionCountByNumber/Hash and GetBlockByNumber/Hash would be inconsistent.

* Left a comment on earliest version in parquet store; looks like it is broken?

Thanks for picking this work up @amir-deris 🍻

sei-db/ledger_db/parquet/store.go
amir-deris: @masih  That is a nice observation. The earliest version in parquet store remains zero even after pruning, unlike other store implementations like `pebbledb/mvcc/db.go`. Should I fix that for Parquet store in another PR?

masih: Yup, it would make sense to fix this issue first in a separate PR, before landing this one. To be clear, `earliestVersion` looks like a pre-existing issue.

@jewei1997 would love your eyes on this; would you be able to take on the fix for `earliestVersion` please?

jewei1997: yes can do that.

jewei1997 (approved):

amir-deris:
As discussed, I will wait for the other fix to go out for `earliestVersion` in Parquet store, before merging this one.

amir-deris:
> As discussed, I will wait for the other fix to go out for `earliestVersion` in Parquet store, before merging this one.

Since Parquet store is going away, not waiting for that work anymore.

PR #3222, opened 2026-04-09
Tx index for Parquet receipt store

jewei1997:

Create a tx index to store (tx hash -> block number) to allow parquet to quickly serve getReceiptByHash queries.

unit tests


sei-db/config/receipt_config.go
```diff
@@ -20,8 +20,23 @@ const (
...
+
+	ReceiptTxIndexBackendNone   = ""
+	ReceiptTxIndexBackendPebble = "pebble"
 )
 
+func NormalizeReceiptTxIndexBackend(backend string) string {
+	switch strings.ToLower(strings.TrimSpace(backend)) {
+	case "", "none", "disabled":
+		return ReceiptTxIndexBackendNone
+	case "pebble", "pebbledb":
```
yzang2019: Would recommend just allow "pebbledb" and all others will default to empty

sei-db/ledger_db/parquet/reader.go
```diff
@@ -321,26 +321,67 @@ func (r *Reader) MaxReceiptBlockNumber(ctx context.Context) (uint64, bool, error
...
+	return r.getReceiptByTxHashFromFiles(ctx, txHash, nil)
+}
+
+// fileForBlock returns the receipt parquet file path that should contain
+// blockNumber, using the sorted tracked file list. Returns "" if no
+// candidate is found.
+func (r *Reader) fileForBlock(blockNumber uint64) string {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+
```
yzang2019, line 361: fileForBlock() acquires r.mu.RLock() to read closedReceiptFiles, but it does not hold pruneMu. getReceiptByTxHashFromFiles acquires pruneMu.RLock() separately. 

Between these two locks, a concurrent prune could delete the file that fileForBlock just returned. The query would then fail on a missing file. The existing GetReceiptByTxHash avoids this by acquiring pruneMu first, then snapshotting files — GetReceiptByTxHashInBlock should follow the same pattern

sei-db/ledger_db/receipt/parquet_store.go
```diff
@@ -190,13 +234,52 @@ func (s *parquetReceiptStore) SetReceipts(ctx sdk.Context, receipts []ReceiptRec
 		return err
 	}
 
+	if s.txHashIndex != nil {
```
yzang2019, line 240: Index update happens after WriteReceipts happens here. If crash happens in between, it could lead to the parquet data exists but the index doesn't. And WAL replay doesn't seems to reindex theses receipts as well

sei-db/ledger_db/receipt/tx_hash_index.go
```diff
@@ -0,0 +1,261 @@
...
+	blockVal := encodeBlockNumber(blockNumber)
+	for _, txHash := range txHashes {
+		if err := batch.Set(makeTxHashKey(txHash), blockVal); err != nil {
+			return err
+		}
+		if err := batch.Set(makeBlockTxKey(blockNumber, txHash), nil); err != nil {
+			return err
+		}
+	}
+	return batch.Commit(dbtypes.WriteOptions{Sync: true})
```
yzang2019: Avoid using Sync:true for perf reasons, this will add a significant latency to each block

sei-db/ledger_db/parquet/store.go
```diff
@@ -156,7 +154,9 @@ func resolveStoreConfig(cfg StoreConfig) StoreConfig {
 	resolved.DBDirectory = cfg.DBDirectory
 	resolved.KeepRecent = cfg.KeepRecent
 	resolved.PruneIntervalSeconds = cfg.PruneIntervalSeconds
-	resolved.DisableTxIndexLookup = cfg.DisableTxIndexLookup
+	if cfg.TxIndexBackend != "" {
+		resolved.TxIndexBackend = dbconfig.NormalizeReceiptTxIndexBackend(cfg.TxIndexBackend)
+	}
```
yzang2019, line 159: If a user explicitly sets TxIndexBackend = "" to disable the index, this condition check will skips it and the resolved config keeps the default (pebble). The user's intent to disable is silently ignored.

jewei1997: I think we do want the default to be pebble. Unless they explicitly chose to disable it.

jewei1997: hm actually this may be an issue, fixing

sei-db/ledger_db/parquet/reader.go
```diff
@@ -321,26 +321,81 @@ func (r *Reader) MaxReceiptBlockNumber(ctx context.Context) (uint64, bool, error
...
+// GetReceiptByTxHash queries for a receipt by transaction hash across all
+// closed parquet files (full scan).
 func (r *Reader) GetReceiptByTxHash(ctx context.Context, txHash common.Hash) (*ReceiptResult, error) {
-	// Hold pruneMu first to prevent file deletion, then snapshot the list.
+	return r.getReceiptByTxHashFromFiles(ctx, txHash, nil)
+}
+
+// GetReceiptByTxHashInBlock narrows the search to the parquet file that
+// should contain blockNumber, falling back to a full scan on miss.
+func (r *Reader) GetReceiptByTxHashInBlock(ctx context.Context, txHash common.Hash, blockNumber uint64) (*ReceiptResult, error) {
```
cody-littley, line 332: > falling back to a full scan on miss.

Could this be a potential DOS attack vector? If somebody is sending you requests for transactions that don't exist, will that cause us to do lots of full scans?

sei-db/ledger_db/parquet/reader.go
```diff
@@ -321,26 +321,81 @@ func (r *Reader) MaxReceiptBlockNumber(ctx context.Context) (uint64, bool, error
...
+		}
+	}
+	return best
+}
+
+func (r *Reader) getReceiptByTxHashFromFiles(ctx context.Context, txHash common.Hash, files []string) (*ReceiptResult, error) {
+	r.pruneMu.RLock()
+	defer r.pruneMu.RUnlock()
+	return r.getReceiptByTxHashFromFilesLocked(ctx, txHash, files)
+}
```
cody-littley, line 378: The only places where I see this method called pass in `nil` for the list of files. Is the `files` parameter actually needed?

sei-db/ledger_db/receipt/parquet_store.go
cody-littley: Was this discussed on a call? I'm also curious about how we handle a crash here.

sei-db/ledger_db/receipt/tx_hash_index.go
```diff
@@ -0,0 +1,261 @@
...
+			if pruneBefore > 0 {
+				if err := p.index.PruneBefore(context.Background(), uint64(pruneBefore)); err != nil {
+					logger.Error("failed to prune tx hash index", "err", err)
+				}
+			}
+			select {
+			case <-p.stopCh:
+				return
+			case <-time.After(p.interval):
+			}
```
cody-littley: nit, consider using a ticker instead

cody-littley (commented):
LLM review left the following commnets:

# Bug Report: Branch vs Main

**Branch commits:** 5 commits (tx hash index for parquet receipt store + pruning race fixes)  
**Scope:** 21 changed files in `sei-db/`  
**Clean areas:** `parquet/reader.go`, `parquet/store.go`, config changes, and test changes — no bugs found.

---

## Bug 1 (High): Iterator leak in `PruneBefore`

**File:** `sei-db/ledger_db/receipt/tx_hash_index.go`, lines 146–190

The Pebble iterator opened at line 146 has no `defer` close. The batch gets a deferred `Close()` (line 155), but the iterator is only closed explicitly at line 188 on the happy path. There are **four** early-return error paths (lines 169, 173, 178, 185) that all leak it.

A leaked Pebble iterator pins its memtable/sstable snapshot, blocking compaction and growing memory. Since the pruner retries on a timer, repeated transient I/O errors would accumulate leaked iterators.

**Fix:** Add `defer iter.Close()` right after the `NewIter` nil-error check at line 152, mirroring the batch pattern.

---

## Bug 2 (High): Resource leak when `replayWAL()` fails during construction

**File:** `sei-db/ledger_db/receipt/parquet_store.go`, lines 70–72

If the tx-hash index backend is Pebble, the constructor opens the Pebble DB (line 51) and starts the pruner goroutine (line 63) *before* calling `replayWAL()` at line 70. If `replayWAL()` fails, the error return leaks all three resources:

1. `store` (parquet.Store) — open DuckDB connection, parquet writers, its own prune goroutine
2. `idx` (PebbleTxHashIndex) — open Pebble database with file locks
3. `pruner` goroutine — running in background, holding a reference to the index

Note the contrast with the error paths *inside* the switch (lines 53, 66) which correctly call `store.Close()`. The Pebble file lock prevents re-opening the index on retry.

**Fix:** Call `wrapper.Close()` before returning the error (it already handles all three teardowns).

---

## Bug 3 (Medium): `IndexBlock` overwrite leaves a stale reverse-index entry

**File:** `sei-db/ledger_db/receipt/tx_hash_index.go`, lines 119–139

When the same tx hash is indexed at a new block number, `IndexBlock` overwrites the primary key (`h + txHash → newBlock`) and writes a new reverse key (`b + newBlock + txHash`), but never deletes the old reverse key (`b + oldBlock + txHash`).

**Scenario:**

1. `IndexBlock(100, [A])` → writes `h+A → 100` and `b+100+A → []`
2. `IndexBlock(200, [A])` → writes `h+A → 200` and `b+200+A → []` — but `b+100+A` remains
3. `PruneBefore(150)` → scans `[b+0, b+150)`, finds `b+100+A`, deletes `h+A`
4. `GetBlockNumber(A)` → returns `(0, false, nil)` even though the receipt lives at block 200

The receipt is still in parquet, so the query falls back to a full DuckDB scan — no data loss, but the index silently degrades to full-scan performance for every overwritten-then-pruned hash.

**Fix:** In `IndexBlock`, read the existing value of `h + txHash` before overwriting. If it exists and differs, delete the old `b + oldBlock + txHash` entry in the same batch.

---

## Bug 4 (Medium): `txHashIndexPruner.Stop()` panics on double call

**File:** `sei-db/ledger_db/receipt/tx_hash_index.go`, lines 252–254

`Stop()` calls `close(p.stopCh)` with no `sync.Once` guard. A second call panics with `"close of closed channel"`. This is reachable because `parquetReceiptStore.Close()` checks `s.indexPruner != nil` but never nils it out after stopping. A deferred cleanup + explicit `Close()` (a normal Go pattern) would crash the node.

Contrast with `PebbleTxHashIndex.Close()` at line 199, which correctly uses `closeOnce`.

**Fix:** Add a `sync.Once` to `Stop()`, consistent with the rest of the file.

---

## Bug 5 (Low): Pruner starts before WAL replay, creating a race window

**File:** `sei-db/ledger_db/receipt/parquet_store.go`, lines 63 and 70

The pruner goroutine is started at line 63 and immediately executes a prune cycle (the sleep comes *after* the prune in the loop body). WAL replay doesn't start until line 70. If the WAL replays index entries for blocks that the pruner concurrently decides are pruneable, the two goroutines race on the same Pebble keys.

In practice the window is narrow (the pruner targets old blocks while the WAL replays recent ones), but it's an unnecessary race that's trivially avoided.

**Fix:** Move `pruner.Start()` to after `replayWAL()` returns successfully.

sei-db/ledger_db/receipt/parquet_store.go
Kbhat1: +1 Can we make the index update part of the same durable/recoverable flow, or otherwise rebuild missing index entries on startup from closed parquet files?

jewei1997: We do WAL crash recovery here https://github.com/sei-protocol/sei-chain/blob/b62c9aa9722eb8c64f97dae51c02a2721a5255bd/sei-db/ledger_db/receipt/parquet_store.go#L448.

sei-db/ledger_db/parquet/reader.go
jewei1997: yeah i think we should do limits on the RPC layer similar to what we do for eth_getLogs today

sei-db/ledger_db/receipt/parquet_store.go
```diff
@@ -37,6 +45,28 @@ func newParquetReceiptStore(cfg dbconfig.ReceiptStoreConfig, storeKey sdk.StoreK
...
+		)
+		pruner.Start()
+		wrapper.indexPruner = pruner
+	default:
+		_ = store.Close()
+		return nil, fmt.Errorf("unsupported receipt tx index backend: %s", txIndexBackend)
+	}
+
 	if err := wrapper.replayWAL(); err != nil {
 		return nil, err
```
Kbhat1, line 70: Think there needs to be a `wrapper.Close()` before returning here

yzang2019 (approved):

Kbhat1 (approved):

PR #3223, opened 2026-04-09
fix reconcile order, pruning spin, and WAL lock retention

wen-coding:
## Summary

Follow-up fixes to #3126 based on reviewer feedback:

1. **Fix reconcile order**: trim tail before prefix. When QCs WAL is empty (corruption), `Reset` blocks cursor to `committee.FirstBlock()`. Without this, prefix reconciliation fast-forwards both block and QC cursors to stale positions, causing "out of sequence" errors on new data.

2. **Fix runPruning spin**: `WaitUntil` now uses `inner.first+1 < inner.nextAppProposal` so the loop blocks when only one entry remains (kept by the `+1` guard) instead of spinning.

3. **Move WAL ops outside locks**: `dataWAL.TruncateBefore` in `PruneBefore` and `runPruning` is now called after releasing the inner lock, avoiding lock retention during disk I/O.

4. **Add `GlobalBlockPersister.Reset`**: Clears WAL and sets cursor to a given position. Used by reconcile for the corruption recovery case.

5. **Rename and reorder tests**: All reconcile tests renamed to `TestReconcileCaseN` matching the numbered case table in the reconcile comment. Added case 1 (empty/empty) and case 5 (blocks ahead of QCs).

## Reconcile case table

| Case | Blocks | QCs | Scenario | Action |
|------|--------|-----|----------|--------|
| 1 | empty | empty | Fresh start | no-op |
| 2 | [a,b] | empty | QCs lost (corruption) | Reset blocks to fb |
| 3 | empty | [X,Y) | Blocks lost (crash) | Prefix: fast-forward blocks.next |
| 4 | [a,b] | [X,Y) | Normal (a=X, b<Y) | no-op |
| 5 | [a,b] | [X,Y) | Prune crash: blocks ahead (a>X) | Prefix: truncate QCs |
| 6 | [a,b] | [X,Y) | Prune crash: QCs ahead (a<X) | Prefix: truncate blocks |
| 7 | [a,b] | [X,Y) | Persist crash: blocks past (b>=Y) | Tail: truncate blocks |
| 8 | [a,b] | [X,Y) | QCs ahead (normal, b<Y) | Tail: no-op |

## Test plan

- [x] `TestReconcileCase1Empty` — fresh start
- [x] `TestReconcileCase2QCsLost` — QCs lost, blocks re-pushed
- [x] `TestReconcileCase2ResetsCursor` — cursor reset verified
- [x] `TestReconcileCase3BlocksLost` — blocks lost, QCs survive
- [x] `TestReconcileCase4Normal` — normal recovery
- [x] `TestReconcileCase4AfterPruning` — after pruning
- [x] `TestReconcileCase5BlocksAhead` — blocks ahead of QCs
- [x] `TestReconcileCase6QCsAhead` — QCs ahead of blocks
- [x] `TestReconcileCase7BlocksPastQCs` — blocks past QC range
- [x] `TestReconcileCase7BlocksTail` — tail truncation with re-push
- [x] `TestReconcileCase8BlocksBehind` — QCs ahead normal
- [x] `TestReconcilePartialQCPrefix` — partial QC from pruning
- [x] `TestReconcileBlockGap` — block gap detection
- [x] `TestRunPruningEmptyState` — no panic on empty state
- [x] All existing tests pass
- [x] gofmt and go vet clean

### DIFF
--- a/sei-tendermint/internal/autobahn/data/state.go
+++ b/sei-tendermint/internal/autobahn/data/state.go
@@ -83,7 +83,7 @@ func (dw *DataWAL) TruncateBefore(n types.GlobalBlockNumber) error {
 //
 //	Case  Blocks    QCs       Scenario                          Action
 //	1     empty     empty     Fresh start                       no-op
-//	2     [a,b]     empty     QCs lost (corruption)             Reset blocks to fb
+//	2     [a,b]     empty     QCs lost (corruption)             error (statesync needed)
 //	3     empty     [X,Y)     Blocks lost (crash)               Prefix: fast-forward blocks.next to X
 //	4     [a,b]     [X,Y)     Normal (a=X, b<Y)                 no-op
 //	5     [a,b]     [X,Y)     Prune crash: blocks ahead (a>X)   Prefix: truncate QCs to a
@@ -92,14 +92,13 @@ func (dw *DataWAL) TruncateBefore(n types.GlobalBlockNumber) error {
 //	8     [a,b]     [X,Y)     QCs ahead (normal, b<Y)           Tail: no-op (blocks catch up)
 func (dw *DataWAL) reconcile(committee *types.Committee) error {
 	fb := committee.FirstBlock()
-	// Fix tail first: when QCs are empty (corruption), reset blocks
-	// before prefix gets a chance to fast-forward cursors to stale values.
+	// Fix tail: remove blocks past QC range.
 	qcNext := dw.CommitQCs.Next()
-	if qcNext == fb {
-		if err := dw.Blocks.Reset(fb); err != nil {
-			return err
-		}
-	} else if err := dw.Blocks.TruncateAfter(qcNext); err != nil {
+	if qcNext == fb && dw.Blocks.Next() > fb {
+		// Blocks exist but QCs WAL is empty — data is corrupted.
+		return fmt.Errorf("corrupted WAL: blocks exist but QCs WAL is empty; statesync required to recover")
+	}
+	if err := dw.Blocks.TruncateAfter(qcNext); err != nil {
 		return fmt.Errorf("truncate blocks tail: %w", err)
 	}
 	// Fix prefix: align both WALs to the later start.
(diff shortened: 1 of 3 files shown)


sei-tendermint/internal/autobahn/data/state.go
```diff
@@ -79,19 +79,30 @@ func (dw *DataWAL) TruncateBefore(n types.GlobalBlockNumber) error {
...
-// to match completes the interrupted prune.
+// The possible WAL states on startup and how they are handled:
 //
-// Tail: if blocks WAL extends past QCs WAL, the excess blocks were
-// persisted without their corresponding QCs (crash during parallel
-// persistence in runPersist). These blocks are untrustworthy — a
-// different QC could arrive for those positions — so they are removed.
+//	Case  Blocks    QCs       Scenario                          Action
+//	1     empty     empty     Fresh start                       no-op
+//	2     [a,b]     empty     QCs lost (corruption)             Reset blocks to fb
```
pompon0: what is fb?

pompon0: in case of corruption we should statesync, no? It feels weird imo to treat this specific corruption scenario differently.

sei-tendermint/internal/autobahn/data/state_test.go
```diff
@@ -499,10 +523,80 @@ func TestStateRecoveryQCsOnly(t *testing.T) {
 	require.NoError(t, dw2.Close())
 }
 
+func TestReconcileCase4Normal(t *testing.T) {
+	// Reconcile case 4: Normal (a=X, b<Y)
```
pompon0: nit: comment + log of the same message is redundant, just t.Log would work.

sei-tendermint/internal/autobahn/data/state.go
pompon0: i.e. imo we should return an error "corrupted" instead of silently resetting to first block

pompon0 (approved):

sei-tendermint/internal/autobahn/data/state.go
wen-coding: Yeah, now I think about it, not very comfortable going forward with no QC, changed to an error.

sei-tendermint/internal/autobahn/data/state_test.go
wen-coding: done

arajasek (approved):

PR #3224, opened 2026-04-10
Made autobahn producer use TxMempool

pompon0:
Also
* disabled bunch of reactors unused in autobahn mode.
* fixed race conditions in mempool processPeerUpdates


sei-tendermint/internal/mempool/mempool.go
```diff
@@ -492,6 +499,31 @@ func (txmp *TxMempool) Flush() {
...
+
+// ReapMaxTxsBytesMaxGas returns a list of transactions within the provided tx,
+// byte, and gas constraints together with the total estimated gas for the
+// returned transactions.
+//
+// NOTE: Gas limits are enforced using int64 running totals. If those totals
+// overflow, gas limit enforcement no longer works correctly. This preserves the
+// historical behavior for backward compatibility.
+func (txmp *TxMempool) ReapMaxTxsBytesMaxGas(maxTxs int, maxBytes, maxGasWanted, maxGasEstimated int64) (types.Txs, int64) {
+	if maxTxs < 0 {
```
wen-coding: it's okay for maxTxs to be equal to 0?

sei-tendermint/internal/mempool/mempool.go
```diff
@@ -492,6 +499,31 @@ func (txmp *TxMempool) Flush() {
...
+}
+
+// ReapMaxTxsBytesMaxGas returns a list of transactions within the provided tx,
+// byte, and gas constraints together with the total estimated gas for the
+// returned transactions.
+//
+// NOTE: Gas limits are enforced using int64 running totals. If those totals
+// overflow, gas limit enforcement no longer works correctly. This preserves the
+// historical behavior for backward compatibility.
+func (txmp *TxMempool) ReapMaxTxsBytesMaxGas(maxTxs int, maxBytes, maxGasWanted, maxGasEstimated int64) (types.Txs, int64) {
```
wen-coding: just curious, why is maxTxs int instead of uint?

sei-tendermint/internal/mempool/mempool.go
```diff
@@ -514,7 +546,7 @@ func (txmp *TxMempool) ReapMaxBytesMaxGas(maxBytes, maxGasWanted, maxGasEstimate
 		size := types.ComputeProtoSizeForTxs([]types.Tx{wtx.tx})
 
 		// bytes limit is a hard stop
-		if maxBytes > -1 && totalSize+size > maxBytes {
+		if totalSize+size > maxBytes {
```
wen-coding: Just curious: any reason the bytes and gas checks happen before append but numtx check happens after append?

sei-tendermint/internal/autobahn/producer/state.go
```diff
@@ -54,21 +53,32 @@ func NewState(cfg *Config, consensus *consensus.State) *State {
...
+		select {
+		case <-ctx.Done():
+		case <-s.txMempool.TxsAvailable():
 		}
-		txs = append(txs, tx.Payload)
-		totalGas += tx.GasUsed
+	}
+
+	txs, totalGas := s.txMempool.ReapMaxTxsBytesMaxGas(
+		int(s.cfg.maxTxsPerBlock()), // nolint:gosec // config values fit into int on supported platforms.
```
wen-coding: nit: Should we just make the first arg of ReapMaxTxsBytesMaxGas uint64 so we don't need a conversion?

sei-tendermint/internal/autobahn/producer/state.go
```diff
@@ -54,21 +53,32 @@ func NewState(cfg *Config, consensus *consensus.State) *State {
...
+		utils.Max[int64](),
+		int64(s.cfg.MaxGasPerBlock), // nolint:gosec // config values stay within int64 range.
+		int64(s.cfg.MaxGasPerBlock), // nolint:gosec // config values stay within int64 range.
```
wen-coding: So we won't have different maxGasWanted and maxGasEstimated in the future?

sei-tendermint/internal/autobahn/producer/state.go
```diff
@@ -54,21 +53,32 @@ func NewState(cfg *Config, consensus *consensus.State) *State {
...
+		case <-ctx.Done():
+		case <-s.txMempool.TxsAvailable():
 		}
-		txs = append(txs, tx.Payload)
-		totalGas += tx.GasUsed
+	}
+
+	txs, totalGas := s.txMempool.ReapMaxTxsBytesMaxGas(
+		int(s.cfg.maxTxsPerBlock()), // nolint:gosec // config values fit into int on supported platforms.
+		utils.Max[int64](),
```
wen-coding: We probably should have maxBytes as well?

sei-tendermint/internal/autobahn/producer/state.go
```diff
@@ -54,21 +53,32 @@ func NewState(cfg *Config, consensus *consensus.State) *State {
...
+		int64(s.cfg.MaxGasPerBlock), // nolint:gosec // config values stay within int64 range.
+		int64(s.cfg.MaxGasPerBlock), // nolint:gosec // config values stay within int64 range.
+	)
+	s.txMempool.RemoveTxs(txs)
+	payloadTxs := make([][]byte, 0, len(txs))
```
wen-coding, line 70: Can we do a cast instead?
`payloadTxs := [][]byte(txs)`

sei-tendermint/internal/autobahn/producer/state.go
```diff
@@ -54,21 +53,32 @@ func NewState(cfg *Config, consensus *consensus.State) *State {
...
+	for _, tx := range txs {
+		payloadTxs = append(payloadTxs, tx)
 	}
 	return types.PayloadBuilder{
 		CreatedAt: time.Now(),
-		TotalGas:  totalGas,
-		Txs:       txs,
+		// TODO: ReapMaxTxsBytesMaxGas does not handle corner cases correctly rn, which actually
+		// can produce negative total gas. Fixing it right away might be backward incompatible afaict,
+		// so we leave it as is for now.
```
wen-coding, line 78: Should I be worried because I see:
TotalGas: uint64(totalGas), // nolint:gosec // **guaranteed to be positive**
in the next line?

sei-tendermint/internal/mempool/reactor/reactor.go
```diff
@@ -23,39 +23,38 @@ var (
 	_ service.Service = (*Reactor)(nil)
 )
 
+const MempoolChannel p2p.ChannelID = 0x30
```
wen-coding, line 26: Just curious, how was 0x30 decided?

sei-tendermint/internal/autobahn/producer/state.go
```diff
@@ -54,21 +53,32 @@ func NewState(cfg *Config, consensus *consensus.State) *State {
...
-		totalGas += tx.GasUsed
+	}
+
+	txs, totalGas := s.txMempool.ReapMaxTxsBytesMaxGas(
+		int(s.cfg.maxTxsPerBlock()), // nolint:gosec // config values fit into int on supported platforms.
+		utils.Max[int64](),
+		int64(s.cfg.MaxGasPerBlock), // nolint:gosec // config values stay within int64 range.
+		int64(s.cfg.MaxGasPerBlock), // nolint:gosec // config values stay within int64 range.
+	)
+	s.txMempool.RemoveTxs(txs)
```
wen-coding: Just an observation: I think it's okay here to acquire mutex twice, since if the tx disappeared in between RemoveTxs still work.
But maybe some day we should re-think the mutex use in mempool, it looks correct and efficient to me in the current context, but if we add more and more features then it might be harder to reason.

sei-tendermint/node/node.go
```diff
@@ -187,22 +189,38 @@ func makeNode(
...
-	mpReactor, err := mempool.NewReactor(mp, router)
-	if err != nil {
-		return nil, fmt.Errorf("mempool.NewReactor(): %w", err)
-	}
 	node.router = router
+	node.mempool = mp
 	node.rpcEnv.Router = router
 	node.shutdownOps = makeCloser(closers)
 
+	if !gigaEnabled {
```
wen-coding, line 217: nit: I think it might help readability to add comment here why we disable the reactor for Autobahn.

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -163,10 +161,27 @@ func (r *GigaRouter) runExecute(ctx context.Context) error {
...
+		for i, tx := range b.Payload.Txs() {
+			blockTxs[i] = tx
+		}
+		r.cfg.TxMempool.Lock()
+		err = r.cfg.TxMempool.Update(
+			ctx,
+			int64(n), // nolint:gosec // autobahn block numbers fit in int64.
+			blockTxs,
+			resp.TxResults,
+			mempool.NopTxConstraintsFetcher,
```
wen-coding: Does this mean we don't set constraints for tx any more?

sei-tendermint/internal/mempool/mempool.go
pompon0: I suppose so

sei-tendermint/internal/autobahn/producer/state.go
pompon0: I don't know, probably we will

sei-tendermint/internal/autobahn/producer/state.go
pompon0: eventually yes

sei-tendermint/internal/autobahn/producer/state.go
pompon0: golang does not support such casting

sei-tendermint/internal/mempool/reactor/reactor.go
pompon0: I don't know, here I'm just moving the constant to another location

sei-tendermint/internal/autobahn/producer/state.go
pompon0: agreed, Mempool handling is already very fragile imo

sei-tendermint/internal/p2p/giga_router.go
pompon0: we need the constraints to be fixed per epoch, because we don't know where the lane blocks will be sequenced. For now we disable them

sei-tendermint/internal/autobahn/producer/state.go
pompon0: yeah, that's misleading. I'll remove the comment after nolint, and just leave the TODO.

sei-tendermint/internal/mempool/mempool.go
pompon0: no reason, moved the check

sei-will (approved):

sei-tendermint/internal/p2p/giga_router.go
wen-coding: Let's add a TODO then with reasoning.

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -94,8 +90,74 @@ func NewGigaRouter(cfg *GigaRouterConfig, key NodeSecretKey) (*GigaRouter, error
...
+		// We need it so that app does not emit error logs.
+		proposer := slices.MinFunc(vals, func(a, b abci.ValidatorUpdate) int { return a.PubKey.Compare(b.PubKey) })
+		key, err := crypto.PubKeyFromProto(proposer.PubKey)
+		if err != nil {
+			return nil, fmt.Errorf("crypto.PubKeyFromProto(): %w", err)
+		}
+		proposerAddress = key.Address()
+	}
+
+	r.cfg.TxMempool.Lock()
```
wen-coding, line 110: Do we need to add a simple metrics on how long we hold this lock and whether it affects RPC?

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -126,47 +188,7 @@ func (r *GigaRouter) runExecute(ctx context.Context) error {
...
-		}
-		// TODO: we need the block to be persisted before we vote for apphash.
-		if err := r.data.PushAppHash(ctx, n, resp.AppHash); err != nil {
-			return fmt.Errorf("r.data.PushAppHash(%v): %w", n, err)
-		}
-		commitResp, err := r.cfg.App.Commit(ctx)
-		if err != nil {
-			return fmt.Errorf("r.cfg.App.Commit(): %w", err)
-		}
+		commitResp, err := r.executeBlock(ctx, b)
```
wen-coding, line 198: If there is error here, should we early exit?

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -94,8 +90,74 @@ func NewGigaRouter(cfg *GigaRouterConfig, key NodeSecretKey) (*GigaRouter, error
...
+			Time:    b.Timestamp,
+			// WARNING: the reward distribution has corner cases where it forgets the proposer,
+			// because reward is distributed with a delay. This is not our problem here though.
+			ProposerAddress: proposerAddress,
+		}).ToProto(),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("r.cfg.App.FinalizeBlock(): %w", err)
+	}
+	// TODO: we need the block to be persisted before we vote for apphash.
```
wen-coding: hmm, is this not covered yet?

sei-tendermint/internal/p2p/giga_router_test.go
```diff
@@ -77,6 +78,15 @@ func (a *testApp) Info(_ context.Context, _ *abci.RequestInfo) (*abci.ResponseIn
 	panic("unreachable")
 }
 
+func (a *testApp) CheckTx(context.Context, *abci.RequestCheckTxV2) (*abci.ResponseCheckTxV2, error) {
```
wen-coding, line 81: Do we have a test where a block failed to execute and commit somewhere?

sei-tendermint/internal/p2p/giga_router.go
wen-coding, line 216: I was trying this PR on https://github.com/sei-protocol/sei-chain/pull/3234. There was one error and all Giga Router stopped, no retries.

AI explains:
Spawn uses SpawnBg which calls s.Cancel(err) if the task returns an error. This cancels the entire scope — all other goroutines get cancelled.

So if runExecute (spawned with SpawnNamed which uses Spawn) returns an error early, the whole scope is cancelled including the dial retry loops.

Does that make sense?

wen-coding:
## Testing PR #3224 with autobahn integration tests

Tested this PR using the autobahn docker cluster setup (PR #3220 + integration tests from #3234). Found the following issues:

### 1. Panic in `executeBlock` — staking params not initialized

```
panic: UnmarshalJSON cannot decode empty bytes

goroutine 329 [running]:
github.com/sei-protocol/sei-chain/sei-cosmos/x/params/types.Subspace.Get(...)
    sei-cosmos/x/params/types/subspace.go:109
github.com/sei-protocol/sei-chain/sei-cosmos/x/staking/keeper.Keeper.MaxValidators(...)
    sei-cosmos/x/staking/keeper/params.go:18
github.com/sei-protocol/sei-chain/sei-cosmos/x/staking/keeper.Keeper.GetBondedValidatorsByPower(...)
    sei-cosmos/x/staking/keeper/validator.go:221
github.com/sei-protocol/sei-chain/app.(*App).GetValidators(...)
    app/app.go:2192
github.com/sei-protocol/sei-chain/sei-tendermint/internal/p2p.(*GigaRouter).executeBlock(...)
    sei-tendermint/internal/p2p/giga_router.go:98
```

`executeBlock` calls `app.GetValidators()` which reads `MaxValidators` from the staking params store, but the params haven't been populated yet at the time the first block is being finalized after `InitChain`. This causes a panic on the very first block.

### 2. RPC `/status` nil pointer panic

```
panic: runtime error: invalid memory address or nil pointer dereference

types.(*ValidatorSet).Copy(...)
    types/validator_set.go:261
consensus.(*State).GetValidators(...)
    internal/consensus/state.go:310
rpc/core.(*Environment).validatorAtHeight(...)
    internal/rpc/core/status.go:120
```

When autobahn is enabled, the CometBFT consensus reactor is skipped, so `ValidatorSet` is nil. The RPC `/status` endpoint calls `GetValidators()` which tries to `Copy()` a nil set.

### Reproduction

```bash
# On the wen/test-3224-with-autobahn branch (PR #3224 + #3220 cherry-picked)
AUTOBAHN=true make docker-cluster-start
# Check logs:
grep "panic" build/generated/logs/seid-0.log
# Try RPC:
curl http://localhost:26657/status
```

Note: The autobahn config also needs `allow_empty_blocks: true` for the cluster to produce blocks without submitted transactions. This is being fixed on our side.

wen-coding:
### Follow-up on Bug #1: Missing Commit after InitChain

Dug deeper into the root cause. The panic in `executeBlock` calling `app.GetValidators()` is because `runExecute` calls `InitChain` but never calls `Commit` after it:

```go
// giga_router.go runExecute()
if last == 0 {
    if _, err := app.InitChain(ctx, r.cfg.GenDoc.ToRequestInitChain()); err != nil {
        return fmt.Errorf("App.InitChain(): %w", err)
    }
    // No Commit() here!
}

for n := next; ; n += 1 {
    b, err := r.data.GlobalBlock(ctx, n)
    // ...
    if vals := r.cfg.App.GetValidators(); len(vals) > 0 {  // PANIC
```

`InitChain` populates the staking params in the deliverState, but `GetValidators()` (in `app.go:2192`) creates a new context via `NewUncachedContext` which reads from the **committed** store — which is still empty because `Commit` was never called.

In CometBFT's normal flow, the handshaker (`replay.go`) calls `Commit` after `InitChain` before any blocks are finalized. The fix is likely to add `app.Commit()` after `InitChain` in `runExecute`, similar to what CometBFT does.

sei-tendermint/internal/p2p/giga_router.go
pompon0: added a TODO

sei-tendermint/internal/p2p/giga_router.go
pompon0: done

sei-tendermint/internal/p2p/giga_router.go
pompon0: true

sei-tendermint/internal/p2p/giga_router.go
pompon0: execution failure is a valid termination reason, because it means that we are not able to execute a finalized block.

sei-tendermint/internal/p2p/giga_router_test.go
pompon0: only happy path tests for now until we get this to work somehow.

sei-will (approved):

wen-coding:
### Follow-up: Height mismatch after InitChain + Commit fix

Tested the latest `5ed9b0686` (with the `Commit` after `InitChain` fix). New panic:

```
panic: invalid height: 1; expected: 2
```

at `app/abci.go:33` in `BeginBlock`.

The issue is in `runExecute`:

```go
if last == 0 {
    if _, err := app.InitChain(ctx, ...); err != nil { ... }
    if _, err := app.Commit(ctx); err != nil { ... }
    next, ok = utils.SafeCast[atypes.GlobalBlockNumber](r.cfg.GenDoc.InitialHeight)  // next = 1
}

for n := next; ; n += 1 {
    // executeBlock at height 1, but app already committed height 1 via InitChain+Commit
    // app expects height 2
```

`InitChain` + `Commit` advances the app to height 1. Then `next` is set to `GenDoc.InitialHeight` (1), so `executeBlock` tries to finalize height 1 again. The app rejects it because it expects height 2.

Fix: after `InitChain` + `Commit`, set `next = InitialHeight + 1` (or query `app.Info()` again to get the actual height).

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -168,7 +172,10 @@ func (r *GigaRouter) runExecute(ctx context.Context) error {
 	next := last + 1
 	if last == 0 {
 		if _, err := app.InitChain(ctx, r.cfg.GenDoc.ToRequestInitChain()); err != nil {
-			return fmt.Errorf("App.InitChain(): %w", err)
+			return fmt.Errorf("app.InitChain(): %w", err)
+		}
+		if _, err := app.Commit(ctx); err != nil {
+			return fmt.Errorf("app.Commit(): %w", err)
```
wen-coding, line 178: Tried again and another panic:
`panic: invalid height: 1; expected: 2`

`BeginBlock` at `app/abci.go:33` expects height 2 but got height 1. This is because `InitChain` + `Commit` already advanced the app state to height 1, so when `executeBlock` tries to finalize block 1, the app expects block 2.

wen-coding (approved):

PR #3227, opened 2026-04-10
Giga store migration guide

Kbhat1:
- Giga Store migration guide + safety checks for EVM SS write/read modes 
- Validates config-level mode combinations, rejects startup when the EVM SS DB is missing or empty while Cosmos SS has history, and rejects mismatched earliest versions post-recovery
- Together these catch the case where an operator flips `evm-ss-write-mode/evm-ss-read-mode` off `cosmos_only` without state syncing
- Docs also updated to cover the full migration flow

- Unit tests
- Verifying on node


sei-db/state_db/ss/composite/store.go
```diff
@@ -77,19 +85,116 @@ func NewCompositeStateStore(
...
+// has history. WAL replay only covers the last KeepRecent blocks, so it cannot
+// rebuild a fresh EVM SS from scratch.
+func (s *CompositeStateStore) validateEVMStateStorePreRecovery() error {
+	if s.evmStore == nil {
+		return nil
+	}
+
+	cosmosLatest := s.cosmosStore.GetLatestVersion()
+	evmLatest := s.evmStore.GetLatestVersion()
+	if cosmosLatest > 0 && evmLatest == 0 {
```
yzang2019, line 165: After state sync, it seems we will only set earliestVersion for SS: 

Image

But what about latestVersion? Is that set as well?

yzang2019 (approved):

jewei1997 (approved):

PR #3228, opened 2026-04-10
fix ss import for raw flatkv modules

Kbhat1:
- Update SS import to recognize raw FlatKV module names and decode those low-level values with vtype before routing them into the logical EVM state store

- Unit tests
- State syncing fully on node


Kbhat1:
Closed in favor of https://github.com/sei-protocol/sei-chain/pull/3250

PR #3229, opened 2026-04-10
feat(flatkv): introduce module-prefix physical keys across all FlatKV DBs

blindchaser:
## Summary

Introduces module-prefixed physical keys across all FlatKV data DBs, enabling non-EVM Cosmos modules to be stored in legacyDB and preparing for future DB consolidation by ensuring key uniqueness across namespaces.

Physical key format: `"module/" + type_prefix + stripped_key`. EVM keys become `"evm/" + kind_byte + addr_or_slot`, non-EVM Cosmos keys become `"<module>/" + original_key`.

- `ktype/ktype.go` (new package): Defines physical key encoding — `ModulePhysicalKey`, `EVMPhysicalKey`, `StripModulePrefix`, `StripEVMPhysicalKey`, `PrefixEnd`. Defines `EVMKeyAccount` as the canonical kind for merged account rows. Replaces the old `keys.go` in the flatkv package.
- `common/evm/keys.go`: Adds `EVMKeyPrefixByte` helper. `BuildMemIAVLEVMKey` refactored to use it, eliminating a redundant allocation.
- `store_apply.go`: `ApplyChangeSets` now splits incoming changesets into EVM vs non-EVM; non-EVM modules are routed into legacyDB under their module prefix. `classifyAndPrefix` replaces the former two-pass `sortChangeSets`+`prefixModuleKeys`, converting keys to physical format via `ktype` helpers in a single pass.
- `store_read.go`: Extracts generic `readFromDB` that checks pending writes then falls back to DB, replacing four per-DB getter implementations. `getLegacyData`/`getLegacyValue` now require a `moduleName` parameter. `getAccountData`, `getStorageData`, `getCodeData` validate key length before constructing physical keys.
- `store_write.go`: Replaces `prepareBatchCodeDB`, `prepareBatchStorageDB`, `prepareBatchLegacyDB` with a single generic `prepareBatch[T]`. Extracts `collectPendingReads` and `deserializeBatchResults` generics to deduplicate `batchReadOldValues`. Commit ordering fix: global metadata is now persisted to metadataDB before updating in-memory `committedVersion`/`committedLtHash`, so a metadata write failure cannot leave the store in an inconsistent state.
- `api.go`: `Store` interface updated — `Get`, `Has`, `GetBlockHeightModified` now take a `moduleName` parameter. `Iterator` docs updated to reflect physical key format for `Key()` and dual-format acceptance for `SeekGE`/`SeekLT`.
- `vtype/legacy_data.go`: Replaces value-length inference for deletion with an explicit `MarkDeleted()` flag and `isDelete` field. The previous implementation treated `len(value) == 0` as a deletion, but Cosmos SDK modules (bank, staking, etc.) legitimately write empty values (`[]byte{}`) to the store — this is standard Cosmos convention and not a deletion. Now that non-EVM modules are routed into legacyDB, the old heuristic would silently drop valid empty-value writes. `MarkDeleted()` makes the delete intent explicit, and `SetValue` (even with `nil` or `[]byte{}`) clears the delete flag, preserving Cosmos semantics.


sei-db/state_db/sc/flatkv/vtype/legacy_data.go
```diff
@@ -97,19 +98,34 @@ func (l *LegacyData) GetValue() []byte {
 }
 
 // Set the legacy value. Returns self (or a new LegacyData if nil).
+// Clears the delete flag — an explicit SetValue is a write, not a deletion,
+// even when value is empty ([]byte{} is a valid Cosmos module value).
```
cody-littley, line 102: Ah, I didn't know this. Good thing to catch!

sei-db/state_db/sc/flatkv/exporter.go
```diff
@@ -229,12 +244,21 @@ func (e *KVExporter) storageToNodes(key, value []byte) ([]*types.SnapshotNode, e
...
+
 	legacyData, err := vtype.DeserializeLegacyData(value)
 	if err != nil {
-		return nil, fmt.Errorf("corrupt legacy entry key=%x: %w", key, err)
+		return nil, fmt.Errorf("corrupt legacy entry module=%s key=%x: %w", moduleName, originalKey, err)
 	}
+
+	// Emit the original key (without module prefix). The importer feeds nodes
+	// back through ApplyChangeSets which re-applies the prefix via
+	// prefixModuleKeys, avoiding double-prefixing.
```
cody-littley: Not a request to change this PR, since this preserves the existing behavior. But we should discuss import/export as a larger group.

I think we need to change the way we export data. Instead of exporting the top level application key-value pairs, I think we need to export the raw keys-value pairs including all of the metadata. This is because we need to pass metadata fields such as block height modified and serialization version during a state sync if we want to calculate the same state hash, and these values are not currently contained within the data we currently export.

sei-db/state_db/sc/flatkv/keys.go
```diff
@@ -85,6 +100,60 @@ func SlotFromBytes(b []byte) (Slot, bool) {
...
+	result := make([]byte, n+1+len(key))
+	copy(result, moduleName)
+	result[n] = '/'
+	copy(result[n+1:], key)
+	return result
+}
+
+// StripModulePrefix splits a module-prefixed physical key into its module name
+// and original key. Returns ok=false if no "/" separator is found.
+func StripModulePrefix(physicalKey []byte) (moduleName string, originalKey []byte, ok bool) {
```
cody-littley: Is there a reason why we don't return an error if we don't find a `"/"` character? Should this be something that bubbles up as an error to the user if there is a malformed key?

sei-db/state_db/sc/flatkv/keys.go
```diff
@@ -85,6 +100,60 @@ func SlotFromBytes(b []byte) (Slot, bool) {
...
+	if memiavlKey == nil {
+		return nil
+	}
+	return ModulePhysicalKey(evm.EVMStoreKey, memiavlKey)
+}
+
+// StripEVMPhysicalKey extracts the EVM key kind and stripped key from a
+// physical DB key. This is the inverse of EVMPhysicalKey for export paths.
+// For account keys the returned kind is EVMKeyAccount (evm.EVMKeyNonce).
+func StripEVMPhysicalKey(physicalKey []byte) (kind evm.EVMKeyKind, strippedKey []byte, ok bool) {
```
cody-littley: If you promote `StripModulePrefix()` to return an error, then you should also promote this function to return an error.

sei-db/state_db/sc/flatkv/store_apply.go
```diff
@@ -19,10 +20,38 @@ func (s *CommitStore) ApplyChangeSets(changeSets []*proto.NamedChangeSet) error
...
+			}
+			for _, pair := range cs.Changeset.Pairs {
+				if pair.Delete {
+					modMap[string(pair.Key)] = nil
+				} else {
+					modMap[string(pair.Key)] = pair.Value
+				}
+			}
+		}
+	}
```
cody-littley: Could we pull this logic into `sortChangeSets()`? Perhaps it would make sense to also add modules to keys, so we'd want to rename the function to `sortAndAddPrefix()` or something.

sei-db/state_db/sc/flatkv/store_apply.go
```diff
@@ -19,10 +20,38 @@ func (s *CommitStore) ApplyChangeSets(changeSets []*proto.NamedChangeSet) error
...
+
+	changesByType, err := sortChangeSets(evmChangeSets)
 	if err != nil {
 		return err
 	}
+
+	prefixModuleKeys(changesByType, nonEVMByModule)
```
cody-littley: It's not obvious that this function also merges non`EVMByModule` down into `changesByType`. Perhaps we could rename this to something like `prefixAndMerge()`?

sei-db/state_db/sc/flatkv/store_read.go
```diff
@@ -195,40 +196,41 @@ func (s *CommitStore) IteratorByPrefix(prefix []byte) Iterator {
...
 }
 
-func (s *CommitStore) getStorageData(keyBytes []byte) (*vtype.StorageData, error) {
-	pendingWrite, hasPending := s.storageWrites[string(keyBytes)]
-	if hasPending {
-		return pendingWrite, nil
+func (s *CommitStore) getAccountData(keyBytes []byte) (*vtype.AccountData, error) {
+	if len(keyBytes) != AddressLen {
+		return nil, nil
 	}
```
cody-littley: Shouldn't we return an error if there is a malformed key? IMO, silently ignoring malformed keys risks making other bugs more tricky to detect.

sei-db/state_db/sc/flatkv/store_read.go
```diff
@@ -195,40 +196,41 @@ func (s *CommitStore) IteratorByPrefix(prefix []byte) Iterator {
...
+	return readFromDB(EVMPhysicalKey(EVMKeyAccount, keyBytes), s.accountWrites, s.accountDB, vtype.DeserializeAccountData, "accountDB")
```
cody-littley: nit, my personal preference is to always wrap errors as I pass them up the stack. I find it tends to make it easier to trace an error to its origin if each step wraps the error.

sei-db/state_db/sc/flatkv/store_read.go
```diff
@@ -195,40 +196,41 @@ func (s *CommitStore) IteratorByPrefix(prefix []byte) Iterator {
...
-	value, err := s.storageDB.Get(keyBytes)
-	if err != nil {
-		if errorutils.IsNotFound(err) {
-			return nil, nil
-		}
-		return nil, fmt.Errorf("storageDB I/O error for key %x: %w", keyBytes, err)
+func (s *CommitStore) getStorageData(keyBytes []byte) (*vtype.StorageData, error) {
+	if len(keyBytes) != AddressLen+SlotLen {
+		return nil, nil
 	}
```
cody-littley: We probably shouldn't ignore malformed keys.

sei-db/state_db/sc/flatkv/store_read.go
```diff
@@ -243,19 +245,10 @@ func (s *CommitStore) getStorageValue(key []byte) ([]byte, error) {
...
-	}
-
-	value, err := s.codeDB.Get(keyBytes)
-	if err != nil {
-		if errorutils.IsNotFound(err) {
-			return nil, nil
-		}
-		return nil, fmt.Errorf("codeDB I/O error for key %x: %w", keyBytes, err)
+	if len(keyBytes) != AddressLen {
+		return nil, nil
```
cody-littley: we shouldn't ignore malformed keys

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -195,270 +187,148 @@ func (s *CommitStore) commitBatches(version int64) error {
...
-			return nil, nil, nil, nil, fmt.Errorf("storageDB batch read error for key %x: %w", k, v.Error)
-		}
-		if v.IsFound() {
-			storageOld[k], err = vtype.DeserializeStorageData(v.Value)
-			if err != nil {
-				return nil, nil, nil, nil, fmt.Errorf("failed to deserialize storage data: %w", err)
-			}
-		}
+	if err = deserializeBatchResults(storageBatch, storageOld, vtype.DeserializeStorageData, "storageDB"); err != nil {
+		return nil, nil, nil, nil, err
```
cody-littley: nit, perhaps wrap error

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -195,270 +187,148 @@ func (s *CommitStore) commitBatches(version int64) error {
...
-			return nil, nil, nil, nil, fmt.Errorf("accountDB batch read error for key %x: %w", k, v.Error)
-		}
-		if v.IsFound() {
-			accountOld[k], err = vtype.DeserializeAccountData(v.Value)
-			if err != nil {
-				return nil, nil, nil, nil, fmt.Errorf("failed to deserialize account data: %w", err)
-			}
-		}
+	if err = deserializeBatchResults(accountBatch, accountOld, vtype.DeserializeAccountData, "accountDB"); err != nil {
+		return nil, nil, nil, nil, err
```
cody-littley: nit, perhaps wrap error

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -195,270 +187,148 @@ func (s *CommitStore) commitBatches(version int64) error {
...
-			return nil, nil, nil, nil, fmt.Errorf("codeDB batch read error for key %x: %w", k, v.Error)
-		}
-		if v.IsFound() {
-			codeOld[k], err = vtype.DeserializeCodeData(v.Value)
-			if err != nil {
-				return nil, nil, nil, nil, fmt.Errorf("failed to deserialize code data: %w", err)
-			}
-		}
+	if err = deserializeBatchResults(codeBatch, codeOld, vtype.DeserializeCodeData, "codeDB"); err != nil {
+		return nil, nil, nil, nil, err
```
cody-littley: nit, perhaps wrap error

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -195,270 +187,148 @@ func (s *CommitStore) commitBatches(version int64) error {
...
-			return nil, nil, nil, nil, fmt.Errorf("legacyDB batch read error for key %x: %w", k, v.Error)
-		}
-		if v.IsFound() {
-			legacyOld[k], err = vtype.DeserializeLegacyData(v.Value)
-			if err != nil {
-				return nil, nil, nil, nil, fmt.Errorf("failed to deserialize legacy data: %w", err)
-			}
-		}
+	if err = deserializeBatchResults(legacyBatch, legacyOld, vtype.DeserializeLegacyData, "legacyDB"); err != nil {
+		return nil, nil, nil, nil, err
```
cody-littley: nit, perhaps wrap error

cody-littley (approved):

yzang2019:
Some suggestions:
1. Follow the vtype Package Pattern — Create a ktype Package
The repo already has a vtype/ package that cleanly encapsulates value-type serialization and deserialization (AccountData, StorageData, LegacyData, etc.). The key serialization logic introduced in this PR — ModulePhysicalKey, StripModulePrefix, EVMPhysicalKey, StripEVMPhysicalKey, PrefixEnd, plus the Address, Slot, and LocalMeta types — would be better housed in an analogous ktype/ package. Right now these are all in keys.go in the root flatkv package alongside the store implementation. Following Cody's pattern with vtype would keep the codebase consistent and make the key format a first-class abstraction with its own tests.

2. Consolidate Sorting and Prefixing
As cody-littley suggested, sortChangeSets and prefixModuleKeys could be merged into a single pass. Currently sortChangeSets classifies EVM pairs by kind using memiavl keys, and then prefixModuleKeys iterates again to prepend the prefix. Doing the prefix during classification avoids the extra iteration and the temporary non-prefixed maps.

sei-db/state_db/sc/flatkv/store_apply.go
```diff
@@ -163,12 +193,41 @@ func sortChangeSets(changeSets []*proto.NamedChangeSet) (map[evm.EVMKeyKind]map[
...
 
+// prefixModuleKeys prepends "module/" to every key for physical DB namespacing.
+// EVM keys get "evm/" prepended (same layout as ModulePhysicalKey).
+// Non-EVM cosmos keys are merged into the legacy map with their own module prefix.
+func prefixModuleKeys(
+	changesByType map[evm.EVMKeyKind]map[string][]byte,
+	nonEVMByModule map[string]map[string][]byte,
+) {
+	evmPrefix := evm.EVMStoreKey + "/"
+	for kind, m := range changesByType {
```
yzang2019: This allocates a new map and string-concatenates every key on every block. For storage-heavy blocks with tens of thousands of entries, this is a non-trivial allocation. Consider prepending the prefix earlier in the pipeline (inside sortChangeSets) to avoid the extra copy pass, or using a key builder that avoids repeated string concatenation.

sei-db/state_db/sc/flatkv/keys.go
blindchaser: good call. changed to return error instead of ok bool

sei-db/state_db/sc/flatkv/keys.go
blindchaser: done. propagates the error from StripModulePrefix

sei-db/state_db/sc/flatkv/store_write.go
blindchaser: agreed. getAccountData, getStorageData, and getCodeData now return fmt.Errorf with expected vs actual key length instead of nil, nil

sei-db/state_db/sc/flatkv/store_write.go
blindchaser: wrapped all four deserializeBatchResults calls with context

sei-db/state_db/sc/flatkv/store_apply.go
blindchaser: good point

yzang2019 (approved):

PR #3231, opened 2026-04-12
fix: Prometheus metrics prefix

m3diumrare:
Most default Prometheus or Prometheus-like systems reject metrics names that don't match `[a-zA-Z_:][a-zA-Z0-9_:]*`


m3diumrare:
@philipsu522 is there a tag to add that can make sure this gets included in the next release? Looks like the CI test flaked unrelated to this change

m3diumrare:
What feedback do I need to address?

m3diumrare:
2 months and one claude token burn later and this PR got replaced: https://github.com/m3diumrare/sei-chain/commit/c82f9f4e04083360074b1b7154f98ce7e9df2ab1

Closing.

PR #3234, opened 2026-04-13
feat: add autobahn integration tests - batch 1 (CON-249)

wen-coding:
## Summary
Add `make autobahn-integration-test` target with 5 integration tests for the autobahn consensus mode:

1. **Block production** — verify height is advancing (via `abci_info`)
2. **Bank transfer** — send usei via `-b sync`, poll for balance
3. **Liveness under max faults** — kill `f` nodes (BFT bound), chain continues
4. **Halts beyond max faults** — kill `f + 1`, chain halts
5. **Recovery** — SKIPPED: autobahn node restart not yet supported (CometBFT handshaker can't reconcile app height vs empty block store)

Test harness is a Go test binary (`integration_test/autobahn/autobahn_test.go`) guarded by the `autobahn_integration` build tag, so it's excluded from the default `go test ./...` run and only invoked via `make autobahn-integration-test`.

The cluster size is discovered at runtime via `docker ps`, and `maxFaults = (clusterSize - 1) / 3` derives the BFT bound — fault-tolerance tests auto-scale to whatever cluster size docker-compose.yml declares. Log checks run via `docker exec grep` inside each running container, so stale host-side logs from prior runs can't false-positive.

Every test is guarded by `assertAutobahnEnabled` which checks `GigaRouter initialized` in every running node's log.

### Known workarounds (with TODOs)
- Uses `abci_info` instead of `/status` for height because `/status` reads from the CometBFT block store, which autobahn does not populate
- Uses `-b sync` instead of `-b block` because CometBFT consensus is disabled in autobahn mode
- 30s stabilization wait for autobahn giga connections to establish

## Usage
```bash
make autobahn-integration-test
```

## Test plan
- [x] `make autobahn-integration-test` passes tests 1–4 locally
- [x] Cluster starts, tests run, cluster stops cleanly (on both success and failure)


integration_test/autobahn/scripts/autobahn_tests.sh
```diff
@@ -0,0 +1,131 @@
...
+# TODO: Re-enable once autobahn supports node restart. Currently, a restarted seid fails
+# because autobahn writes to the app state but not to the CometBFT block/state store.
+# On restart, the CometBFT handshaker sees appHeight >> storeHeight and cannot reconcile.
+echo ""
+echo "=== Test 5: Recovery (SKIPPED — autobahn node restart not yet supported) ==="
+
+echo ""
+echo "============================================"
+echo "  All Autobahn Integration Tests PASSED"
+echo "============================================"
```
masih: Any chance we can write these in go?

pompon0: +1

wen-coding: Changed.

integration_test/autobahn/autobahn_test.go
```diff
@@ -0,0 +1,265 @@
...
+var (
+	clusterSize int
+	maxFaults   int
+)
+
+// listRunningNodes returns the container names of currently-running
+// sei-node-* containers.
+func listRunningNodes(t *testing.T) []string {
+	t.Helper()
+	out, err := exec.Command("docker", "ps",
```
pompon0, line 56: nit: given that we already run docker commands from here, would it make sense to spawn & teardown the docker cluster from go as well?

integration_test/autobahn/autobahn_test.go
```diff
@@ -0,0 +1,265 @@
...
+	if err != nil {
+		return 0, err
+	}
+	defer resp.Body.Close()
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return 0, err
+	}
+	var parsed struct {
+		Response struct {
```
pompon0: is this type exported by sei-tendermint by any chance?

integration_test/autobahn/autobahn_test.go
```diff
@@ -0,0 +1,265 @@
...
+
+func TestAutobahn(t *testing.T) {
+	// Discover cluster size once, before any test kills nodes.
+	names := listRunningNodes(t)
+	if len(names) == 0 {
+		t.Fatalf("no running sei-node-* containers")
+	}
+	clusterSize = len(names)
+	// BFT tolerates f faults in a cluster of n = 3f + 1.
+	maxFaults = (clusterSize - 1) / 3
```
pompon0, line 279: here we hardcode weights of the cluster. Add a TODO to address that, once custom weights are supported (or skip logging it)

pompon0 (approved):

integration_test/autobahn/autobahn_test.go
wen-coding: done

integration_test/autobahn/autobahn_test.go
wen-coding: Changed

integration_test/autobahn/autobahn_test.go
wen-coding: Good point, changed

arajasek (approved):

PR #3236, opened 2026-04-13
feat(loadtest): migrate metrics to OpenTelemetry

amir-deris:
## Summary

Replaces the loadtest package's legacy `utils/metrics` and `sei-cosmos/telemetry` instrumentation with OpenTelemetry (OTel), exporting metrics via the standard Prometheus bridge.

- Adds `otel_metrics.go`: initializes a process-scoped OTel `MeterProvider` backed by a dedicated Prometheus registry. Instruments three metrics: `sei_loadtest_produce_count`, `sei_loadtest_consume_count`, and `sei_loadtest_tps_tps`.
- Replaces `metrics.IncrProducerEventCount` / `IncrConsumerEventCount` / `SetThroughputMetricByType` calls in `loadtest_client.go` and `main.go` with package-local wrappers over the OTel counters/gauge.
- Simplifies `metrics.go`: removes the old `telemetry.Metrics`-based handler; the `/metrics` endpoint now serves from the OTel-registered Prometheus gatherer via `promhttp`.
- `run()` initializes the OTel provider on startup and defers a clean shutdown.
- Adds scrape-level tests (`otel_metrics_scrape_test.go`) to verify metrics are exposed correctly.

## Test plan

- [x] `go test ./loadtest/...` passes, including new scrape tests
- [x] Loadtest binary runs and `/metrics` endpoint exposes `sei_loadtest_*` metrics with `msg_type`, `service`, and `host` labels

### DIFF
--- a/loadtest/main.go
+++ b/loadtest/main.go
@@ -39,7 +39,6 @@ import (
 	"golang.org/x/time/rate"
 
 	"github.com/sei-protocol/sei-chain/app"
-	"github.com/sei-protocol/sei-chain/utils/metrics"
 	tokenfactorytypes "github.com/sei-protocol/sei-chain/x/tokenfactory/types"
 )
 
@@ -139,6 +138,12 @@ func deployEvmContracts(config *Config) {
 }
 
 func run(config *Config) {
+	shutdownOtel, err := initLoadtestOtelMetrics()
+	if err != nil {
+		panic(err)
+	}
+	defer func() { _ = shutdownOtel(context.Background()) }()
+
 	config.EVMAddresses = &EVMAddresses{}
 	// Start metrics collector in another thread
 	metricsServer := MetricsServer{}
@@ -273,7 +278,7 @@ func printStats(
 		//nolint:gosec
 		tps := float64(sentCount-prevTotalSent) / elapsed.Seconds()
 		totalTps += tps
-		defer metrics.SetThroughputMetricByType("tps", float32(tps), msgType)
+		defer setThroughputMetricByType("tps", float32(tps), msgType)
 	}
 
 	var totalDuration time.Duration
--- a/loadtest/otel_metrics.go
+++ b/loadtest/otel_metrics.go
@@ -1,0 +1,121 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"sync"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/otlptranslator"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	otelprometheus "go.opentelemetry.io/otel/exporters/prometheus"
+	"go.opentelemetry.io/otel/metric"
+	otelmetric "go.opentelemetry.io/otel/sdk/metric"
+)
+
+const loadtestTelemetryServiceName = "loadtest-client"
+
+var (
+	loadtestOtelErr error
+
+	loadtestPromReg    *prometheus.Registry
+	loadtestShutdownMP func(context.Context) error
+
+	loadtestProduceCounter metric.Int64Counter
+	loadtestConsumeCounter metric.Int64Counter
+	loadtestTpsGauge       metric.Float64Gauge
+
+	loadtestHost string
+)
+
+func loadtestMetricOpts(msgType string) metric.MeasurementOption {
+	var once sync.Once
+	once.Do(func() {
+		loadtestHost, _ = os.Hostname()
+		if loadtestHost == "" {
+			loadtestHost = "unknown"
+		}
+	})
+	return metric.WithAttributes(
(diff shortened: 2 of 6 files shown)


loadtest/otel_metrics.go
```diff
@@ -0,0 +1,121 @@
...
+		attribute.String("msg_type", msgType),
+		attribute.String("service", loadtestTelemetryServiceName),
+		attribute.String("host", loadtestHost),
+	)
+}
+
+// initLoadtestOtelMetrics configures the process-wide OTel MeterProvider and Prometheus
+// registry used by the loadtest metrics HTTP server. Call once from run() before any
+// producer/consumer records metrics.
+func initLoadtestOtelMetrics() (func(context.Context) error, error) {
```
bdchatham, line 51: nit: You specify context but don't use it. It's useful for managing the lifecycle of your process, like `incrConsumerEventCount`.

loadtest/main.go
```diff
@@ -139,6 +138,12 @@ func deployEvmContracts(config *Config) {
 }
 
 func run(config *Config) {
+	shutdownOtel, err := initLoadtestOtelMetrics()
+	if err != nil {
+		panic(err)
+	}
+	defer func() { _ = shutdownOtel(context.Background()) }()
```
bdchatham, line 145: You'd want to create a context here that respects termination signals. This relates to the comment below on contexts. Example:

```
ctx, cancel := signal.NotifyContext(parent, os.Interrupt, syscall.SIGTERM)
```

This gives you a signal-aware context and a cancel method you can call if you want to manually trigger the threads using that context to gracefully stop.

loadtest/otel_metrics.go
```diff
@@ -0,0 +1,121 @@
...
+		)
+		if err != nil {
+			loadtestOtelErr = fmt.Errorf("create otel prometheus exporter: %w", err)
+			return
+		}
+
+		provider := otelmetric.NewMeterProvider(otelmetric.WithReader(exporter))
+		otel.SetMeterProvider(provider)
+
+		meter := provider.Meter("loadtest")
```
bdchatham, line 70: **Definitely not blocking**

It may be useful to support dimensions parameters so that you can run a load test with some ID and you are able to filter metrics across load test data by it. You could pass this in via parameter or env.

I could see this helping users maintain their sanity when comparing between load test results. I sadly know from painful experience.

loadtest/otel_metrics.go
```diff
@@ -0,0 +1,121 @@
...
+func setThroughputMetricByType(metricName string, value float32, msgType string) {
+	ctx := context.Background()
+	// Legacy keys were sei, loadtest, tps, <metricName>. Loadtest only passes "tps".
+	if metricName != "tps" {
+		return
+	}
+	loadtestTpsGauge.Record(ctx, float64(value), loadtestMetricOpts(msgType))
+}
+
+func loadtestPrometheusGatherer() prometheus.Gatherer {
```
bdchatham, line 119: nit: is this needed?

loadtest/otel_metrics.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	return loadtestShutdownMP, nil
+}
+
+func incrProducerEventCount(msgType string) {
+	ctx := context.Background()
+	loadtestProduceCounter.Add(ctx, 1, loadtestMetricOpts(msgType))
+}
+
+func incrConsumerEventCount(msgType string) {
```
bdchatham, line 105: For my own understanding, what does `msgType` usually contain? Any examples?

loadtest/otel_metrics.go
```diff
@@ -0,0 +1,121 @@
...
+	ctx := context.Background()
+	loadtestConsumeCounter.Add(ctx, 1, loadtestMetricOpts(msgType))
+}
+
+func setThroughputMetricByType(metricName string, value float32, msgType string) {
+	ctx := context.Background()
+	// Legacy keys were sei, loadtest, tps, <metricName>. Loadtest only passes "tps".
+	if metricName != "tps" {
```
bdchatham, line 113: Should this just return an error since loadtest only supports tps? This will fail silently which might be intended so double checking

loadtest/otel_metrics.go
amir-deris: I believe `context` here is part of the return type, specifying a function that takes `context` and returning error, so I don't think I am supposed to use the `context` here. It just specifies the return function signature.

bdchatham: Ah, good point. Resolving this one

loadtest/main.go
amir-deris: That is a great feedback. I will work on fixing the context usage.

amir-deris:
Since the `loadtest` package is deprecated and we are going to improve `sei-load` repository instead, declining this PR.

PR #3243, opened 2026-04-14
fix: autobahn InitChain, GetValidators, and mempool TTL (CON-249)

wen-coding:
## Summary

Fixes autobahn `runExecute` so the docker cluster (`AUTOBAHN=true`) starts and produces blocks, and fixes a test hang in `TestGigaRouter_FinalizeBlocks`.

**Root cause 1 — double InitChain**: The CometBFT handshaker (`consensus/replay.go`) always calls `InitChain` when `appHeight==0`. The autobahn `runExecute` was also calling `InitChain` — this double-init corrupted app state. Additionally, `GetValidators` panicked reading staking params from the empty committed store before the first `Commit`.

**Root cause 2 — mempool TTL eviction**: The mempool starts at `height=-1`. Txs submitted before the first block inherit `height=-1`. With `TTLNumBlocks=10`, `purgeExpiredTxs` instantly evicts all txs on the first `Update(blockHeight=InitialHeight)` because `-1 < InitialHeight-10` when `InitialHeight` is large (random up to 100k in the test). This is a race: the producer must drain the mempool before the first block executes, otherwise txs are silently lost.

### What changed

1. **`giga_router.go` — remove InitChain/Commit from `runExecute`**: The CometBFT handshaker owns `InitChain`. `runExecute` now just sets `next = InitialHeight` when `last == 0`, relying on the `deliverState` the handshaker already set up. Includes a `WARNING` comment about state sync incompatibility.

2. **`app/app.go` — `GetValidators` uses `DeliverContext` at height 0**: After the handshaker's `InitChain` but before the first `Commit`, the committed store is empty (staking params not persisted). Reading from it panics in `MaxValidators` with "UnmarshalJSON cannot decode empty bytes". At `LastBlockHeight() == 0`, we read from `DeliverContext` instead, which has the uncommitted staking state.

3. **`baseapp/test_helpers.go` — add `DeliverContext()` accessor**: Returns the current `deliverState` context, or nil if not set.

4. **`giga_router_test.go` — align testApp with real SDK**:
   - Adapt `Committed` field: `InitChain` now sets `Committed=true` so `FinalizeBlock` can follow without an intermediate `Commit` (matching the CometBFT handshaker flow). Double-commit is still caught.
   - `Info()` returns `height=0` before first Commit (matches real SDK).
   - `TestGigaRouter_FinalizeBlocks` calls `InitChain` before `router.Run()` to simulate the CometBFT handshaker.
   - Add `TestInitChainCommitThenFinalize` contract test for testApp.

5. **`mempool/testonly.go` — disable TTL in `TestConfig()`**: Sets `TTLNumBlocks=0` and `TTLDuration=0` to prevent spurious tx eviction in all mempool tests.

## Test plan

- [x] `TestInitChainCommitThenFinalize` passes
- [x] `TestGigaRouter_FinalizeBlocks` passes (was hanging in CI)
- [x] Docker `AUTOBAHN=true` cluster: all 4 nodes produce blocks in sync
- [x] Bank transfer finalizes
- [x] Fault tolerance: 1 node down → chain continues (3/4 quorum); 2 nodes down → chain halts


pompon0:
this pr seems to be a noop, no?

wen-coding:
> this pr seems to be a noop, no?

Sorry, still working on it, Claude pushed before it's ready. I'll ping you when it's ready for review.

sei-will (approved):

sei-tendermint/internal/p2p/giga_router_test.go
```diff
@@ -223,7 +267,19 @@ func TestGigaRouter_FinalizeBlocks(t *testing.T) {
...
+			// In production, the handshaker always runs before GigaRouter.Run().
+			if _, err := app.InitChain(ctx, genDoc.ToRequestInitChain()); err != nil {
+				return fmt.Errorf("app.InitChain(): %w", err)
+			}
+			mempoolCfg := mempool.TestConfig()
+			// Disable TTL purging: txs enter the mempool at height=-1 (the mempool's
+			// initial height) but the first block is at InitialHeight (random, up to
+			// 100001). With TTLNumBlocks=10, purgeExpiredTxs would instantly evict all
+			// txs on the first Update call because -1 < InitialHeight-10.
+			mempoolCfg.TTLNumBlocks = 0
```
pompon0: we need purging disabled by default for all mempool tests, so move these to TestConfig()

sei-tendermint/internal/p2p/giga_router_test.go
```diff
@@ -187,6 +182,55 @@ func (c *testNodeCfg) GigaNodeAddr() GigaNodeAddr {
 	}
 }
 
+// TestInitChainCommitThenFinalize verifies the autobahn block execution flow:
```
pompon0: Afaiu this is a contract test for testApp itself, right? Please document it somehow.

sei-tendermint/internal/p2p/giga_router_test.go
```diff
@@ -36,7 +36,6 @@ type testAppState struct {
 	Blocks     []*abci.RequestFinalizeBlock
 	Txs        map[shaHash]bool
 	AppHash    shaHash
-	Committed  bool
```
pompon0, line 39: why do you remove Committed field? It was introduced to verify that GigaRouter actually calls Commit at appropriate moments.

wen-coding: Originally testApp: InitChain → Commit → FinalizeBlock → Commit → ...
But in real SDK flow (handshake): InitChain → FinalizeBlock → Commit → ...
The handshaker calls `InitChain `but does not call `Commit` after it. Then `runExecute` calls `FinalizeBlock` at `InitialHeight` using the `deliverState` that `InitChain` set up. The old testApp rejected this because `Committed` was false after `InitChain` (only `Commit` set it to true).

sei-tendermint/internal/p2p/giga_router_test.go
wen-coding: done

sei-tendermint/internal/p2p/giga_router_test.go
wen-coding: done

sei-tendermint/internal/p2p/giga_router_test.go
wen-coding: Okay, changed it to set Committed to true in InitChain so we can still check for double commit.

sei-tendermint/internal/mempool/testonly.go
```diff
@@ -4,5 +4,11 @@ func TestConfig() *Config {
 	cfg := DefaultConfig()
 	cfg.CacheSize = 1000
 	cfg.DropUtilisationThreshold = 0.0
+	// Disable TTL purging in tests: the mempool starts at height=-1, so txs
```
pompon0: nit: this test config is used in various context, so the comment doesn't seem relevant.

pompon0 (approved):

sei-tendermint/internal/mempool/testonly.go
wen-coding: Changed

PR #3246, opened 2026-04-14
fix(evmrpc): omit notifications from legacy JSON-RPC batch responses per spec

amir-deris:
## Summary

Fixes the `sei-legacy-http` batch gateway to correctly handle JSON-RPC 2.0 **notifications** (requests with no `id` field).

**Previous behavior:**
- Notifications were assigned "leftover" inner responses in FIFO order, producing spurious response entries
- When all items in a batch produced no responses, the gateway returned `[]` (an empty JSON array)

**New behavior per JSON-RPC 2.0 §6:**
- Notifications are silently omitted from the batch response array
- If every item in a batch is a notification (or produces no response), the server sends an **empty HTTP body** rather than `[]`
- Batch response array length is no longer 1:1 with the request batch when notifications are present

**Changes:**
- Switched from fixed-size indexed output arrays to `append`-based slices so response count naturally matches non-notification items
- Extracted `seiLegacyBatchResponsesNoForward` helper for the no-forward path
- Extracted `appendMergeFailure` closure in `mergeSeiLegacyHTTPBatch` to eliminate repeated merge-failure blocks
- Added `writeJSONRPCBatchResponse` which handles the empty-body case (replaces direct `writeJSONArrayResponse` calls)
- Updated `AGENTS.md` to document the notification-omission and empty-body behaviour
- Added four unit tests covering: notification omitted from merged response, notification + blocked call, invalid + notification, all-notifications → empty body

## Test plan
- [x] `go test ./evmrpc/... -run TestWrapSeiLegacyHTTP_Batch`
- [x] Existing legacy batch tests continue to pass


evmrpc/sei_legacy_http.go
```diff
@@ -194,15 +187,38 @@ func (g *seiLegacyHTTPGate) handleBatch(w http.ResponseWriter, r *http.Request,
...
+func seiLegacyBatchResponsesNoForward(
+	invalidReq []bool,
+	blockedErr []error,
+	ids []json.RawMessage,
+	lenMsgs int,
+) []json.RawMessage {
+	out := make([]json.RawMessage, 0, lenMsgs)
+	for i := 0; i < lenMsgs; i++ {
+		if invalidReq[i] {
+			out = append(out, marshalJSONRPCError(orNullID(ids[i]), -32600, seiLegacyBatchInvalidReqMsg))
```
bdchatham: nit: can you set a constant with a self-describing name for `-32600`? I had to look this up. Without context, it's a magic number to a reader.

amir-deris: That is a good point. Will make the fix for it.

evmrpc/sei_legacy_http.go
```diff
@@ -13,6 +13,11 @@ import (
 // read more than the inner JSON-RPC stack will accept (see rpc.Server.SetHTTPBodyLimit).
 const seiLegacyHTTPMaxBody = 5 * 1024 * 1024
 
+const (
```
bdchatham, line 16: Nice 👍🏻

bdchatham (approved):

evmrpc/sei_legacy_http.go
```diff
@@ -194,15 +192,38 @@ func (g *seiLegacyHTTPGate) handleBatch(w http.ResponseWriter, r *http.Request,
...
+// inner server: invalid slots yield -32600, blocked gated methods yield gate errors, and notifications
+// are omitted (JSON-RPC 2.0: no response for notifications, including in batches).
+func seiLegacyBatchResponsesNoForward(
+	invalidReq []bool,
+	blockedErr []error,
+	ids []json.RawMessage,
+	lenMsgs int,
+) []json.RawMessage {
+	out := make([]json.RawMessage, 0, lenMsgs)
+	for i := 0; i < lenMsgs; i++ {
```
masih: Use the newer for range syntax
```suggestion
	for i := range lenMsgs {
```

masih (approved):

PR #3249, opened 2026-04-15
strongly typed clist

pompon0:
Made clist generic. Also
* removed unused maxLength
* made blocking methods of clist respect ctx
* simplified CheckTx by removing the callback


sei-will (approved):

sei-tendermint/internal/evidence/reactor.go
```diff
@@ -229,50 +230,39 @@ func (r *Reactor) broadcastEvidenceLoop(ctx context.Context, peerID types.NodeID
...
-				return
-			}
-		}
-
-		ev := next.Value.(types.Evidence)
-		evProto, err := types.EvidenceToProto(ev)
+		next, err := r.evpool.WaitEvidenceFront(ctx)
 		if err != nil {
-			panic(fmt.Errorf("failed to convert evidence: %w", err))
+			return
```
wen-coding, line 239: nit: Should we return error some day?

sei-tendermint/internal/evidence/reactor.go
```diff
@@ -229,50 +230,39 @@ func (r *Reactor) broadcastEvidenceLoop(ctx context.Context, peerID types.NodeID
...
-		case <-timer.C:
-			// start from the beginning after broadcastEvidenceIntervalS seconds
-			timer.Reset(time.Second * broadcastEvidenceIntervalS)
-			next = nil
+		err = utils.WithTimeout(ctx, broadcastEvidenceInterval, func(ctx context.Context) error {
+			for {
+				ev := next.Value()
+				evProto, err := types.EvidenceToProto(ev)
+				if err != nil {
+					panic(fmt.Errorf("failed to convert evidence: %w", err))
```
wen-coding, line 247: nit: add a comment what this panic will do? (I guess it will close our Gossip to this peer?)

sei-tendermint/internal/libs/clist/clist.go
```diff
@@ -12,14 +12,14 @@ to ensure garbage collection of removed elements.
...
 import (
-	"fmt"
+	"context"
+	"errors"
 	"sync"
 )
 
 // MaxLength is the max allowed number of elements a linked list is
 // allowed to contain.
 // If more elements are pushed to the list it will panic.
```
wen-coding, line 22: Should you move this comment as well?

sei-tendermint/internal/libs/clist/clist.go
```diff
@@ -39,69 +39,76 @@ the for-loop. Use sync.Cond when you need serial access to the
...
 	e.mtx.RLock()
 	isRemoved := e.removed
 	e.mtx.RUnlock()
 	return isRemoved
 }
 
-func (e *CElement) detachNext() {
+func (e *CElement[T]) Value() T {
+	e.mtx.RLock()
+	value := e.value
```
wen-coding, line 106: value in e never changes right? Do we need this lock?

sei-tendermint/internal/rpc/core/mempool.go
```diff
@@ -23,7 +23,7 @@ import (
 // https://docs.tendermint.com/master/rpc/#/Tx/broadcast_tx_async
 // Deprecated and should be removed in 0.37
 func (env *Environment) BroadcastTxAsync(ctx context.Context, req *coretypes.RequestBroadcastTx) (*coretypes.ResultBroadcastTx, error) {
-	go func() { _ = env.Mempool.CheckTx(ctx, req.Tx, nil, mempool.TxInfo{}) }()
+	go func() { _, _ = env.Mempool.CheckTx(ctx, req.Tx, mempool.TxInfo{}) }()
```
wen-coding, line 26: This ctx is the request context right? Will it be cancelled once the request is done? Then we run CheckTX in a canclled context?

sei-tendermint/internal/mempool/reactor/reactor.go
```diff
@@ -241,43 +242,28 @@ func (r *Reactor) processPeerUpdates(ctx context.Context) error {
 
 func (r *Reactor) broadcastTxRoutine(ctx context.Context, peerID types.NodeID) {
 	peerMempoolID := r.ids.GetForPeer(peerID)
-	// TODO: this function does not call any external code, so panics should not be expected.
-	defer func() {
-		if e := recover(); e != nil {
```
wen-coding, line 246: Do we not need this recover() any more?

sei-tendermint/internal/rpc/core/mempool.go
pompon0: CheckTx in sei-chain/app/abci.go is not blocking, so it ignores the fact that context is cancelled. Context is used for tracing though.

sei-tendermint/internal/mempool/reactor/reactor.go
pompon0: afaict this code is safe now. There is no interface calls which would be allowed to panic.

sei-tendermint/internal/libs/clist/clist.go
pompon0: we don't

sei-tendermint/internal/evidence/reactor.go
pompon0: done

sei-tendermint/internal/evidence/reactor.go
pompon0: not necessarily, since the error will be always context.Canceled/DeadlineExceeded here. TBD

wen-coding (approved):

PR #3250, opened 2026-04-15
FlatKV refactor for state sync import + export

yzang2019:

Refactor the FlatKV export/import pipeline to use raw physical keys and vtype-serialized values end-to-end, and teach the SS (State Store) composite importer to handle FlatKV snapshot data.

### SC (State Commitment) Layer

**FlatKV Exporter — rewritten to raw export**
- Replaced the old parsing/conversion exporter with a raw iterator approach: `KVExporter` now uses `RawGlobalIterator()` to emit physical key/value pairs directly from the DBs without any transformation
- Removed all per-type conversion functions (`accountToNodes`, `codeToNodes`, `storageToNodes`, `legacyToNodes`, `pendingNodes`)

**FlatKV Importer — channel-based concurrent pipeline**
- Rewrote `KVImporter` as a concurrent pipeline with per-DB worker goroutines (`dbWorker`) writing batches and computing LtHash in parallel
- `AddNode` sends to a shared `ingestCh`; a `dispatch` goroutine routes keys to the correct DB worker via `routePhysicalKey`
- Fail-fast error propagation using `atomic.Pointer[error]` and a `done` channel — `AddNode` and `dispatch` exit immediately on first error
- New `FinalizeImport(version)` method persists LtHash and version metadata after the entire import completes
- Fsync is always disabled during import for performance

**FlatKV RawGlobalIterator — simplified**
- Reimplemented as a `sequentialIterator` that iterates DBs in fixed order (EVM sub-DBs then legacyDB) using `dataDBs()`
- Iterator errors are now checked and propagated (previously silently swallowed)
- Removed `EmptyIterator` (dead code)

**FlatKV Store refactoring**
- Added `dataDBs()` and `namedDataDBs()` helpers in `store.go` to centralize the list of data DBs, replacing inline DB lists throughout the codebase
- Added `routePhysicalKey(key)` to route a physical key to the correct DB
- Removed all memiavl references from FlatKV code and comments

**Composite SC Store**
- Fixed `flatkvCommiter` typo → `flatkvCommitter` (used ~10 times)
- Moved `FlatKVExportModuleName` to `common/keys/keys.go` as `FlatKVStoreKey`
- Fixed composite importer routing bug: FlatKV nodes now return early if `flatkvImporter` is nil instead of falling through to Cosmos importer
- Fixed error message: "failed to create evm importer" → "failed to create flatkv importer"

### SS (State Store) Layer
**SS Composite Import — rewritten and simplified**
- Rewrote `Import` from ~135 lines with two separate code paths into a unified ~75-line function:
  - Single routing loop for all write modes (DualWrite, SplitWrite, CosmosOnlyWrite, no-EVM)
  - `evmCh` is only created when EVM store is available and active
  - Lightweight `done` channel + `sync.Once` pattern replaces complex `drainImportErr`/`sendNode` retry closures
- Added `convertFlatKVNodes()` function that transforms FlatKV physical-key snapshot nodes into SS-compatible format:
  - Strips module prefix (`ktype.StripModulePrefix`) and parses key kind (`keys.ParseEVMKey`)
  - **Account keys**: deserializes `vtype.AccountData`, emits separate nonce (8-byte BE) and codeHash (32-byte) nodes; skips zero-nonce and zero-codeHash (EOA accounts)
  - **Storage/Code keys**: deserializes the corresponding vtype and emits raw value with `StoreKey = "evm"`
  - **Legacy keys**: deserializes `vtype.LegacyData` and preserves the original module name as `StoreKey` so keys route back to the correct Cosmos SS module (e.g. `"bank"`, `"staking"`)
- Updated import routing: when EVM SS is enabled, EVM keys go exclusively to EVM store, non-EVM keys go to Cosmos store
- Removed unused `normalizeSnapshotNode` function and `splitWrite` variable from Import

### Config & Misc

- Fixed config validation error messages: `ReaderConstantThreadCount` / `ReaderPoolQueueSize` / `MiscConstantThreadCount` now say "must not be negative" (previously said "must be greater than 0" while allowing 0)
- Fixed duplicate "for" typo in `api.go` comment

- **`ss/composite/store_test.go`**: Updated `TestImport_OnlyEvmModule` to verify EVM-only routing (no cosmos duplication); rewrote `TestImport_OnlyEvmFlatkvModule` and `TestImport_BothEvmAndEvmFlatkv` with proper FlatKV physical keys and vtype-serialized values; rewrote `TestImport_CosmosOnlyWrite_ConvertsFlatkvToCosmos` to verify conversion in no-EVM path
- **`sc/flatkv/import_export_test.go`**: Updated round-trip tests, exporter tests for raw key/value output, importer tests for physical key routing, corrupt data propagation tests
- **`sc/composite/store_test.go`**: Added `RawGlobalIterator()` to mock, updated all `flatkvCommitter` references, replaced `FlatKVExportModuleName` with `keys.FlatKVStoreKey`
- **`sc/flatkv/store_write_test.go`**, **`store_read_test.go`**, **`snapshot_test.go`**, **`perdb_lthash_test.go`**, **`lthash_correctness_test.go`**: Fixed compile errors, updated to use centralized DB helpers and physical key formats


sei-db/state_db/sc/flatkv/importer.go
```diff
@@ -2,89 +2,230 @@ package flatkv
...
+		ltPairs: make([]lthash.KVPairWithLastValue, 0, importBatchSize),
+		ltHash:  ltHash,
+	}
+}
+
+// run drains the worker channel until closed, flushing whenever the
+// buffer reaches importBatchSize. If done fires, the worker abandons
+// remaining work and exits immediately.
+func (w *dbWorker) run(done <-chan struct{}) error {
+	for {
```
blindchaser, line 54: consider to add  

```
defer func() {
        if w.batch != nil {
            _ = w.batch.Close()
 }()
```

to avoid resource leak in some edge cases

sei-db/state_db/ss/composite/store.go
```diff
@@ -258,124 +261,168 @@ func stripEVMFromChangesets(changesets []*proto.NamedChangeSet) []*proto.NamedCh
...
+	kind, strippedKey := keys.ParseEVMKey(innerKey)
+
+	switch kind {
+	case keys.EVMKeyNonce:
+		acct, err := vtype.DeserializeAccountData(node.Value)
+		if err != nil {
+			return nil, fmt.Errorf("convertFlatKVNodes account: %w", err)
+		}
+		var nodes []types.SnapshotNode
+		if nonce := acct.GetNonce(); nonce != 0 {
```
blindchaser: when nonce == 0 and codeHash is zero (e.g., an EOA that has received Sei but never sent a transaction), the SS store should still keep the account right?

@Kbhat1 @jewei1997

sei-db/state_db/sc/flatkv/importer.go
```diff
@@ -2,89 +2,230 @@ package flatkv
...
+	}
+}
+
+// flush commits the current PebbleDB batch and updates the running LtHash.
+func (w *dbWorker) flush() error {
+	if len(w.ltPairs) == 0 {
+		return nil
+	}
+
+	newHash, _ := lthash.ComputeLtHash(w.ltHash, w.ltPairs)
```
cody-littley, line 85: In theory, we could offload lattice hash calculation to a work pool and get parallelism between DB operations and hash calculations. Cryptosim performance makes me think we could probably get a 2-3x speedup from this, assuming receiving data from the network isn't the bottleneck.

Writing this so we remember it in the future. Not a blocker for this PR (let's not prematurely optimize until we have benchmarks and actionable data).

sei-db/state_db/sc/flatkv/store_iterator.go
```diff
@@ -0,0 +1,144 @@
...
+// in fixed order (account → code → storage → legacy). Within each DB the
+// keys are returned in PebbleDB's natural order. Per-DB _meta/* keys are
+// skipped. Pending writes are not visible. metadataDB is not included.
+func (s *CommitStore) RawGlobalIterator() Iterator {
+	return &sequentialIterator{dbs: s.dataDBs()}
+}
+
+// sequentialIterator iterates through a slice of DBs one at a time.
+// It fully drains the current DB before moving to the next.
+type sequentialIterator struct {
```
cody-littley, line 12: Nit, I think putting these before structs that implement an interface makes the code easier to grok.

```suggestion
var _ Iterator = (*sequentialIterator)(nil)

// sequentialIterator iterates through a slice of DBs one at a time.
// It fully drains the current DB before moving to the next.
type sequentialIterator struct {
```

sei-db/state_db/ss/composite/store.go
```diff
@@ -258,124 +261,168 @@ func stripEVMFromChangesets(changesets []*proto.NamedChangeSet) []*proto.NamedCh
...
+		ld, err := vtype.DeserializeLegacyData(node.Value)
+		if err != nil {
+			return nil, fmt.Errorf("convertFlatKVNodes legacy: %w", err)
+		}
+		return []types.SnapshotNode{
+			{StoreKey: moduleName, Key: innerKey, Value: ld.GetValue()},
+		}, nil
+
+	default:
+		return nil, nil
```
cody-littley: Should this return an error? If we get an unrecognized key type here, that's a bug, right?

cody-littley (approved):

sei-db/state_db/ss/composite/store.go
yzang2019: What shall we store in SS in this case? Just account and nonce=0? Agreed it's a little bit tricky here

sei-db/state_db/sc/flatkv/importer.go
yzang2019: That's true, good point, let me either add a todo or just optimize in this PR

sei-db/state_db/sc/flatkv/store_iterator.go
yzang2019: Make sense

sei-db/state_db/ss/composite/store.go
blindchaser: i'm thinking we should use `vtype/account_data.go/IsDelete()` to distinguish existing but nonce==0 vs non-existing/deleted, which decide if we put it in ss.

blindchaser (approved):

sei-db/state_db/ss/composite/store.go
yzang2019: make sense, will make a change

sei-db/state_db/sc/flatkv/store_iterator.go
yzang2019: Fixed

Kbhat1 (approved):

sei-db/state_db/ss/composite/store.go
Kbhat1: Yea +1 think we can raise a new err @yzang2019

yzang2019: Agreed, returning an error now

PR #3253, opened 2026-04-15
Migrate evmrpc telemetry from utils/metrics to direct Prometheus client

amir-deris:
## Summary

- Replaces `utils/metrics` calls in `evmrpc` (`filter.go`, `utils.go`, `websockets.go`) with a new local `rpc_telemetry.go` layer that registers metrics directly on `prometheus.DefaultRegisterer`
- Preserves the existing metric series names (`sei_rpc_request_counter`, `sei_rpc_request_latency_ms`, `sei_websocket_connect`) and their label shapes so production dashboards are unaffected
- Avoids the global OTel Prometheus exporter for this path — the exporter applies a `sei_chain_` namespace prefix and counter suffix rules that would break the legacy series names.

Benefits: 

1. Decoupling — evmrpc no longer needs to import utils/metrics, which is a shared package backed by armon/go-metrics. Changes to that package can't accidentally break evmrpc telemetry.

2. Explicitness — The metric definitions (names, labels, types, quantile objectives) are now visible directly in     rpc_telemetry.go instead of buried in a generic helper. It's easier to audit what's being emitted.                    
3. Forward compatibility — As the codebase migrates toward OpenTelemetry, evmrpc's metrics are now isolated from the OTel Prometheus bridge's name-mangling rules. You can evolve the rest of the telemetry stack without risking dashboard breakage for these specific series.
## Note
Since a new `success` label is added for `sei_rpc_request_latency_ms`, and there are existing dashboards in sei-infra [sei_node_monitoring](https://github.com/sei-protocol/sei-infra/blob/main/monitoring/deploy/configs/grafana/dashboards/sei_node_monitoring.json) and [launch_warroom](https://github.com/sei-protocol/sei-infra/blob/main/monitoring/deploy/configs/grafana/dashboards/launch_warroom.json), we might need to update those dashboards.


evmrpc/rpc_telemetry.go
```diff
@@ -0,0 +1,71 @@
...
+		)
+		// Match armon/go-metrics prometheus sink objectives for MeasureSince samples.
+		rpcLatencySummary = prometheus.NewSummaryVec(
+			prometheus.SummaryOpts{
+				Name:       "sei_rpc_request_latency_ms",
+				Help:       "RPC request latency in milliseconds",
+				Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+				MaxAge:     10 * time.Second,
+			},
+			[]string{"endpoint", "connection"},
```
bdchatham: I think we'd also what a success dimension here since a request that fails is going to have notably different performance characteristics than one that succeeded

evmrpc/rpc_telemetry.go
```diff
@@ -0,0 +1,71 @@
...
+
+func recordRPCRequest(endpoint, connection string, success bool) {
+	initRPCTelemetryMetrics()
+	rpcRequestCounter.WithLabelValues(endpoint, connection, strconv.FormatBool(success)).Inc()
+}
+
+func recordRPCLatency(endpoint, connection string, start time.Time) {
+	initRPCTelemetryMetrics()
+	// Match armon go-metrics MeasureSince: timer value = elapsed / time.Millisecond.
+	ms := float64(time.Since(start).Nanoseconds()) / float64(time.Millisecond)
```
bdchatham, line 64: Typically I have always measured latency in nanos and transformed in whatever observability tool we're using. Not critical now but something to keep in-mind when instrumenting components like this.

evmrpc/filter.go
```diff
@@ -794,7 +793,7 @@ func (f *LogFetcher) GetLogsByFilters(ctx context.Context, crit filters.FilterCr
 	processBatch := func(batch []*coretypes.ResultBlock) {
 		defer func() {
 			// Add metrics for log processing
-			metrics.IncrementRpcRequestCounter("num_blocks_fetched", "logs", true)
+			recordRPCRequest("num_blocks_fetched", "logs", true)
```
bdchatham, line 796: nit: I know you are just getting parity but these parameters seem not idiomatic to what the method is written to do

amir-deris: If I understand correctly, the issue is for `recordRPCRequest`, the endpoint is meant to be a JSON-RPC method like `eth_getLogs`, etc and `connectionType` meant to be `http/web socket`. But here we are using adhoc method name and connection type. Is that correct?
If so, it seems the goal is an internal metric (blocks touched per batch, log path vs block path), we can add a separate Prometheus metric if that's ok.

evmrpc/rpc_telemetry.go
amir-deris: Very good point. I will add that.

evmrpc/rpc_telemetry.go
amir-deris: Good idea. It seems that is part of a larger change as there are some other areas that are also working with millisecond.

evmrpc/filter.go
bdchatham: Rephrasing for clarity - basically whoever wrote this before overloaded the concept of "endpoint", at least that's my understanding from looking specifically at this code and not being able to look at callers or correlated pieces. Regardless, not a big issue for you now, just a callout really.

bdchatham (approved):

masih (changes requested):
@amir-deris we want to standardise on OTEL and use runtime bindings to report to promethus instead of direct metrics rporting via prometheus client. 

Please take a look at [this](https://github.com/sei-protocol/sei-chain/blob/dcdde1bd550a923063067fb91a3f439e72253381/sei-db/state_db/sc/memiavl/metrics.go#L4) as an example.

amir-deris:
I will close this PR and create a separate PR implementing @masih  suggestion.

evmrpc/rpc_telemetry.go
Pray4Web3: func getCachedReceipt(globalBlockCache BlockCache, blockHeight int64, txHash common.Hash) (*evmtypes.Receipt, bool) {
	if entry, found := globalBlockCache.Get(blockHeight); found {
		entry.RLock()
		defer entry.RUnlock()
		if receipt, hasReceipt := entry.Receipts[txHash]; hasReceipt {
			return receipt, true
		}
	}
	return nil, false
}

evmrpc/filter.go
Pray4Web3: defer func() {
		for _, batch := range sortedBatches {
			f.globalLogSlicePool.Put(batch)
		}
	}()

PR #3257, opened 2026-04-15
perf(app): reuse decoded transactions across ProcessProposalHandler hot path

amir-deris:
## Problem

In `ProcessProposalHandler`, transactions were decoded twice per block:
1. Inside `checkTotalBlockGas` (decoded from raw bytes, then discarded)
2. Again inside `ProcessBlock` → `DecodeTransactionsConcurrently`

## Changes

**Decode once, reuse everywhere:**
- `ProcessProposalHandler` now calls `DecodeTxBytesConcurrently` upfront and passes the result through to both `checkTotalBlockGas` and `ProcessBlock`.

**Refactored decode pipeline:**
- `DecodeTxBytesConcurrently` — new function, decodes raw bytes to `sdk.Tx` in parallel (no EVM preprocessing).
- `FinalizeDecodedTransactionsConcurrently` — new function, runs EVM preprocessing on already-decoded txs in parallel (no-op when Giga executor is enabled).
- `DecodeTransactionsConcurrently` — now delegates to the two above; existing callers are unaffected.

**Signature changes:**
- `ProcessBlock` gains a `preDecoded []sdk.Tx` parameter. When non-nil and aligned with `txs`, the pre-decoded slice is used directly (EVM preprocessing still runs); otherwise falls back to a full decode.
- `checkTotalBlockGas` now accepts `[]sdk.Tx` instead of `[][]byte`, eliminating its internal decode loop.

## Benchmark results
BenchmarkTxDecodePipeline_LegacyDuplicateDecodePlusFinalize-10              1622            711444 ns/op          839021 B/op      17959 allocs/op
BenchmarkTxDecodePipeline_OptimizedDecodeOncePlusFinalize-10                2734            435730 ns/op          499572 B/op      11140 allocs/op

It shows about 38.8% faster operation, 40.5% less memory, and 38.0% fewer allocations.


app/benchmark_tx_decode_test.go
```diff
@@ -0,0 +1,117 @@
...
+			txs = append(txs, c)
+		}
+	}
+	return txs
+}
+
+// BenchmarkTxDecodePipeline_LegacyDuplicateDecodePlusFinalize approximates the pre-reuse hot path:
+// a full byte decode (e.g. ProcessProposal gas accounting) plus ProcessBlock’s DecodeTransactionsConcurrently,
+// which decodes the same bytes again and runs EVM preprocessing.
+func BenchmarkTxDecodePipeline_LegacyDuplicateDecodePlusFinalize(b *testing.B) {
```
bdchatham, line 89: Curious about the long-term value of this benchmark pair — particularly `Legacy`. It's a reconstruction of the old shape composed from two current functions, so once this lands nothing in prod calls that sequence and the bench can't catch a regression. `Optimized` is more useful but doesn't guard the thing this PR actually introduces which is the reuse plumbing between `ProcessProposalHandler` and `ProcessBlock`.

Suggestion: drop both, put the 711k→436k numbers in the PR description as one-shot evidence, and add a follow-up for a `ProcessBlock`-level bench (with vs without `preDecoded`) that would actually catch reuse-path regressions. 

Ultimately, unless we had a standard convention for benchmarks already, I'd say a benchmark harness is something entirely separate from these optimizations we eventually want to have and weave your OTel work into.

WDYT?

app/app.go
```diff
@@ -1208,13 +1208,12 @@ func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcess
...
 
-	// TODO: this check decodes transactions which is redone in subsequent processing. We might be able to optimize performance
-	// by recording the decoding results and avoid decoding again later on.
+	typedTxs := app.DecodeTxBytesConcurrently(req.Txs)
 
 	// Use the clean context for gas validation only. We cannot reassign
 	// ctx because ProcessBlock writes to ctx's store downstream.
 	checkCtx := app.GetProcessProposalCleanContext()
-	if !app.checkTotalBlockGas(checkCtx, req.Txs) {
+	if !app.checkTotalBlockGas(checkCtx, typedTxs) {
```
bdchatham, line 1216: Worth a short comment here documenting the ordering invariant this relies on: at this point `nil` entries in `typedTxs` mean **proto decode failures only** (matching the old `err != nil → skip` branch that used to live inside `checkTotalBlockGas`). EVM preprocessing happens later, inside `ProcessBlock` → `FinalizeDecodedTransactionsConcurrently`.

If someone later reorders Finalize to run before this gas check, preprocess-failure nils would silently be excluded from gas accounting and the semantics change without an obvious signal. A one-liner here would save a future debugging session:

```go
// Invariant: at this point nil entries in typedTxs are proto decode failures only.
// EVM preprocessing runs later inside ProcessBlock; do not reorder that before this check.
```

app/app.go
```diff
@@ -1755,7 +1757,13 @@ func (app *App) ProcessBlock(ctx sdk.Context, txs [][]byte, req *BlockProcessReq
 	events = append(events, beginBlockResp.Events...)
 
 	evmTxs := make([]*evmtypes.MsgEVMTransaction, len(txs)) // nil for non-EVM txs
-	typedTxs := app.DecodeTransactionsConcurrently(ctx, txs)
+	var typedTxs []sdk.Tx
+	if len(preDecoded) == len(txs) && preDecoded != nil {
```
bdchatham, line 1761: Nit: `preDecoded != nil` is redundant here. `len(nil) == 0`, so:
- If `len(txs) > 0`: the length equality already implies non-nil.
- If `len(txs) == 0`: both branches produce an empty `typedTxs` with no work done — equivalent.

can simplify to:

```suggestion
	if len(preDecoded) == len(txs) {
```

app/benchmark_tx_decode_test.go
amir-deris: I totally agree, I think committing these benchmarks in the codebase for these small changes doesn't provide a long term value. I can just report the numbers in the pr description, as you said. 
We can come up with better ways to measure the overall impact of these changes, for example by using metrics.

app/app.go
```diff
@@ -1755,7 +1760,13 @@ func (app *App) ProcessBlock(ctx sdk.Context, txs [][]byte, req *BlockProcessReq
 	events = append(events, beginBlockResp.Events...)
 
 	evmTxs := make([]*evmtypes.MsgEVMTransaction, len(txs)) // nil for non-EVM txs
-	typedTxs := app.DecodeTransactionsConcurrently(ctx, txs)
+	var typedTxs []sdk.Tx
+	if len(preDecoded) == len(txs) {
+		typedTxs = preDecoded
+		app.FinalizeDecodedTransactionsConcurrently(ctx, typedTxs)
+	} else {
+		typedTxs = app.DecodeTransactionsConcurrently(ctx, txs)
```
masih, line 1768: The issue here is that this path 2Xs the number of goroutines per process block. Can we do better?

app/app.go
```diff
@@ -2128,6 +2132,47 @@ func (app *App) DecodeTransactionsConcurrently(ctx sdk.Context, txs [][]byte) []
...
+	for i, typedTx := range typedTxs {
+		if typedTx == nil {
+			continue
+		}
+		isEVM, _ := evmante.IsEVMMessage(typedTx)
+		if !isEVM {
+			continue
+		}
+		wg.Add(1)
+		go func(idx int, tx sdk.Tx) {
```
masih, line 2167: The (pre-existing) run-away concurrency pattern without any sort of cap on volume of work is unfortunate.

I am not too concerned about the number of goroutines being spawned; the concerning part os the decoding overhead. This type of work processing could be the root cause of unpredictable block time fluctuations for example.

Addressing this goes beyond the scope of this PR though. So no actionable here. I thought it is worth pointing out.

app/app.go
```diff
@@ -2128,6 +2132,47 @@ func (app *App) DecodeTransactionsConcurrently(ctx sdk.Context, txs [][]byte) []
...
+			defer wg.Done()
+			defer func() {
+				if err := recover(); err != nil {
```
masih: Another pre-existing error/panic handing which is very complex for what it is doing; we have 3 layers of this that I can count in process block. 

Again, no actioanbles just now. Pointing out for future improvements.

app/app.go
```diff
@@ -2353,9 +2398,8 @@ func (app *App) checkTotalBlockGas(ctx sdk.Context, txs [][]byte) (result bool)
 
 	totalGas, totalGasWanted := uint64(0), uint64(0)
 	nonzeroTxsCnt := 0
-	for _, tx := range txs {
-		decodedTx, err := app.txDecoder(tx)
-		if err != nil {
+	for _, decodedTx := range typedTxs {
+		if decodedTx == nil {
```
masih, line 2452: Nice

masih (commented):
No obvious blockers that I can see; feeling excited about this one! Great work @amir-deris !

Question: can we do better than 2Xing the goroutines for the path where `preDecoded` is empty?

app/app.go
amir-deris: Very good point. I will implement a path that combines two set of go routines into one set.

app/app.go
```diff
@@ -1714,7 +1716,10 @@ func (app *App) ProcessTXsWithOCCGiga(ctx sdk.Context, txs [][]byte, typedTxs []
 	return execResults, ctx
 }
 
-func (app *App) ProcessBlock(ctx sdk.Context, txs [][]byte, req *BlockProcessRequest, lastCommit abci.CommitInfo, simulate bool) (events []abci.Event, txResults []*abci.ExecTxResult, endBlockResp abci.ResponseEndBlock, err error) {
+// ProcessBlock executes block transactions. If preDecoded is non-nil and len(preDecoded)==len(txs),
+// those decoded transactions are reused (bytes are not decoded again); EVM preprocessing still runs
+// on the block context.
+func (app *App) ProcessBlock(ctx sdk.Context, txs [][]byte, req *BlockProcessRequest, lastCommit abci.CommitInfo, simulate bool, preDecoded []sdk.Tx) (events []abci.Event, txResults []*abci.ExecTxResult, endBlockResp abci.ResponseEndBlock, err error) {
```
masih, line 1722: Would love to see a testcaase with non-nil preDecode

masih (approved):

bdchatham (approved):

PR #3262, opened 2026-04-16
test(flatkv): add flatkv integration testings

blindchaser:
## Summary

Adds a rootmulti-level integration test suite for FlatKV that exercises the
real composite-store wiring end-to-end (memiavl + FlatKV behind `rootmulti.Store`),
plus a small production helper (`flatkv.VerifyLtHash`) and an invariant-pinning
doc comment on `lthash.ComputeLtHash`. The suite covers every FlatKV WriteMode
(DualWrite / SplitWrite / CosmosOnly / Shadow), all four recovery directions
(FlatKV-behind-cosmos, cosmos-behind-FlatKV, rollback, reopen cycles), the full
state-sync `Snapshot` / `Restore` pipeline — including the production race
between Snapshot export and concurrent block commits — and the SS-enabled
Query path (fast, proof, missing-key). Equivalence against memiavl is asserted
byte-for-byte in both directions at the latest version and at every historical
version.

## Test plan

All 30 new integration tests pass (`go test ./sei-cosmos/storev2/rootmulti/
-run TestFlatKV -count=1`, ~32s) and pass under the race detector
(`-race`, ~39s, no data races reported). Coverage grouped by file:

- `flatkv_equivalence_test.go`
  - `TestFlatKVDualWriteExhaustiveEquivalence`: rich multi-block EVM
    workload → memiavl full iteration and FlatKV `Exporter` full drain
    at latest must be equal in both directions.
  - `TestFlatKVDualWriteEquivalenceAtHistoricalVersions`: same invariant
    asserted at every committed version, not only the latest.

- `flatkv_hash_test.go`
  - `TestFlatKVDualWriteHashConsistency`: 10-block DualWrite lattice
    hash changes block-to-block and is always 32 bytes.
  - `TestFlatKVSplitWriteHashConsistency`: SplitWrite keeps the memiavl
    `evm` IAVL hash stable while the lattice hash evolves.
  - `TestFlatKVLatticeHashDeterminism`: two independent stores on the
    same input produce identical app and lattice hashes.
  - `TestFlatKVLatticeHashSensitivity`: single-byte EVM divergence flips
    both lattice and app hash.
  - `TestFlatKVDualWriteDoubleFlush`: simulated `FinalizeBlock` +
    `Commit` double flush produces the same app hash as a single flush
    and `evm_lattice` survives.
  - `TestFlatKVMixedCosmosAndEVMBlocks`: lattice hash changes only on
    EVM-write blocks, is stable on cosmos-only blocks.

- `flatkv_workload_test.go`
  - `TestFlatKVFullScanLtHashVerification`: 10-block run, after close
    `VerifyLtHash` passes and FlatKV `CommittedRootHash` equals the
    `evm_lattice` in the last commit info.
  - `TestFlatKVDeleteAndOverwriteWorkload`: delete-then-recreate across
    blocks and same-block set-then-delete of a non-existent key leaves
    LtHash identical to the previous block.
  - `TestFlatKVMultiAccountWorkload`: 20 accounts × 10 blocks
    randomized (60% set / 20% delete / 20% skip) with
    `verifyHistoricalHashes` and `verifyFlatKVSelfConsistent` after
    close.
  - `TestFlatKVSplitWriteReadRouting`: EVM data absent from memiavl,
    present in FlatKV; memiavl `Has()` returns false.
  - `TestFlatKVLargeChangesetDeterminism`: 200-key single-block payload
    triggers FlatKV's parallel LtHash path; two independent runs
    produce identical app hashes.

- `flatkv_recovery_test.go`
  - `TestFlatKVRollbackWithLatticeHash`: rollback to mid-chain reproduces
    the original v3 app and lattice hash; chain continues.
  - `TestFlatKVCrashRecoveryThroughRootMulti`: FlatKV rolled back to v3
    while cosmos stays at v5 → on reopen reconciles to v3.
  - `TestFlatKVSplitWriteCrashRecovery`: same reconciliation under
    SplitWrite where EVM data lives only in FlatKV; FlatKV must still
    hold the v3 EVM value after reconcile.
  - `TestFlatKVReverseCrashRecoveryFlatKVAhead`: cosmos rolled back,
    FlatKV ahead → reconciles to the lower version.
  - `TestFlatKVSplitWriteReverseCrashRecovery`: SplitWrite variant of
    the reverse-crash case; asserts the memiavl `evm` subtree hash stays
    stable across post-reconcile blocks.
  - `TestFlatKVMultiCloseReopenCycles`: three sequential open / write /
    close cycles; `verifyHistoricalHashes` passes at the end.
  - `TestFlatKVRollbackToVersionOne`: rollback to the earliest boundary
    (v1), replayed blocks produce byte-identical hashes to the original
    chain.
  - `TestFlatKVSequentialRollbacks`: two rollbacks (to v7 then v5)
    without reopen in between.
  - `TestFlatKVRollbackThenCrashRecovery`: explicit rollback followed
    by a simulated crash where cosmos advanced past FlatKV.

- `flatkv_snapshot_test.go`
  - `TestFlatKVSnapshotRestoreWithLatticeHash`: DualWrite snapshot/restore
    round-trip; source and restored destination produce identical app
    hashes as the chain advances on both.
  - `TestFlatKVSplitWriteSnapshotRestore`: same round-trip under
    SplitWrite, asserting the memiavl `evm` subtree stays empty on both
    source and destination, FlatKV carries the v5 EVM value through the
    snapshot, and post-restore app hashes match byte-for-byte.
  - `TestFlatKVConcurrentSnapshotAndCommit`: `Snapshot` on one goroutine,
    block commits 6..15 on another; snapshot must complete, restore into
    a fresh node, and that restored node must match a cleanly-built
    reference byte-for-byte as the chain advances. Passes under
    `-race`.
  - `TestFlatKVPruneBoundaryQueries`: `SnapshotKeepRecent=1` on both
    backends; queries within the retained window still reconstruct the
    original app and lattice hashes, FlatKV stays self-consistent on
    close.

- `flatkv_modes_test.go`
  - `TestFlatKVShadowModeNoLatticeInAppHash`: lattice-disabled shadow
    run produces identical FlatKV `CommittedRootHash` to the
    lattice-enabled reference run on the same input; `evm_lattice` is
    absent from `CommitInfo` in shadow mode and present otherwise;
    `VerifyLtHash` passes on both.
  - `TestFlatKVQueryWithSSAndReadMode`: cosmos and EVM queries via SS
    return committed values; SC proof path returns a valid ics23 proof
    that verifies against the committed app hash and rejects a tampered
    value.
  - `TestFlatKVQueryNonExistentKey`: missing acc key (cosmos module,
    no proof), missing evm key (SS path, no proof), and missing evm key
    with `Prove=true` (SC absence proof) all return `Code=0` with nil
    `Value`; the proof path additionally returns non-nil `ProofOps`.
  - `TestFlatKVSSAsyncRestartConsistency`: SS with
    `AsyncWriteBuffer=100` (production default) + clean close/reopen;
    historical cosmos and EVM queries after reopen match originally
    committed values, validating the `RecoverCompositeStateStore` WAL
    replay path wired through rootmulti for DualWrite.


yzang2019 (approved):

yzang2019:
Minor: 
No explicit non-EVM module prefix test — The collectFlatKVEVM helper asserts moduleName == keys.EVMStoreKey for every exported node, but there's no test that writes non-EVM data through the FlatKV legacy path and verifies it round-trips correctly through export/import. Given PR #3229's module-prefix work, this would be a valuable addition.

cody-littley (approved):

PR #3263, opened 2026-04-16
Avoid runtime.AddCleanup issye

masih:
fix to runtime cleaup issue


masih:
This is fixed on main; ignoring for now

PR #3264, opened 2026-04-16
fix: add retry to apt-get update in Docker CI

wen-coding:
## Summary
- Add retry loop (3 attempts, 5s sleep) to `apt-get update` in `docker/localnode/Dockerfile`
- Ubuntu mirrors intermittently return hash mismatches or stale package indices, causing all integration test jobs to fail at the Docker build step

### DIFF
--- a/docker/localnode/Dockerfile
+++ b/docker/localnode/Dockerfile
@@ -2,7 +2,8 @@ FROM golang:1.25.6-bookworm AS go-dist
 
 FROM ubuntu:latest
 ENV HOME="/root" PATH="/root/go/bin:/sei-protocol/sei-chain/integration_test/upgrade_module/scripts/:$PATH"
-RUN apt-get update && apt-get install -y make build-essential git jq python3 curl vim uuid-runtime
+RUN for i in 1 2 3; do apt-get update && break || sleep 5; done && \
+    apt-get install -y make build-essential git jq python3 curl vim uuid-runtime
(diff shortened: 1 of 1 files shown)


masih (approved):

docker/localnode/Dockerfile
```diff
@@ -2,7 +2,8 @@ FROM golang:1.25.6-bookworm AS go-dist
 
 FROM ubuntu:latest
 ENV HOME="/root" PATH="/root/go/bin:/sei-protocol/sei-chain/integration_test/upgrade_module/scripts/:$PATH"
-RUN apt-get update && apt-get install -y make build-essential git jq python3 curl vim uuid-runtime
+RUN for i in 1 2 3; do apt-get update && break || sleep 5; done && \
```
amir-deris: If all 3 retries fail, the for loop exits 0 (success from the last sleep 5). This means a persistent failure would silently proceed to apt-get install
   and fail there with a less obvious error.                                                                                                              
                                                                                                                                                          
  Suggestion: Use a pattern that preserves failure exit code:
```
  RUN for i in 1 2 3; do apt-get update && break || { [ $i -lt 3 ] && sleep 5; }; done && \                                                               
      apt-get install -y make build-essential git jq python3 curl vim uuid-runtime  
```

amir-deris (approved):

docker/localnode/Dockerfile
wen-coding: good catch, done

PR #3265, opened 2026-04-16
feat(evmrpc): migrate RPC telemetry to OpenTelemetry Meter API

amir-deris:
## Summary

Follow-up to #3253, rebased to use the standardized OpenTelemetry Meter API per reviewer feedback (@masih: \"standardise on OTEL and use runtime bindings to report to prometheus instead of direct metrics reporting via prometheus client\").

- Replaces `utils/metrics` calls in `evmrpc` (`filter.go`, `utils.go`, `websockets.go`) with a new `metrics.go` layer using `otel.Meter("evmrpc")` (the global MeterProvider convenience wrapper)
- Keeps `IncrementRpcRequestCounter`, `MeasureRpcRequestLatency`, and `IncWebsocketConnects` in `utils/metrics` but annotates them `TODO(PLT-326)` for removal once dashboards are migrated to the new `evmrpc_*` OTel series
- Removes `num_blocks_fetched` batch counters from internal `processBatch` call sites — these were unused and not wired to any dashboard or alert
- Threads `ctx` through all RPC handler signatures (14+ files: `association.go`, `block.go`, `info.go`, `subscribe.go`, `tracers.go`, `tx.go`, `txpool.go`, and others) to support OTel's context-based propagation and adopt the new defer-with-`recover()` pattern
- Adds `success`, `error_class`, and `jsonrpc_code` labels to the latency histogram so failed requests and error types can be tracked with low cardinality
- Adds a `ctx context.Context` parameter to `NewPendingTransactionFilter` (previously missing)
- Adds a test for the new telemetry layer

New metric names (OTel naming convention, exported via the process-wide MeterProvider):
- `evmrpc_request_latency_seconds` — request latency histogram (labels: `endpoint`, `connection`, `success`, `error_class`, `jsonrpc_code`)
- `evmrpc_websocket_connects_total` — websocket connection count

### Metric name migration

Legacy metrics (`sei_rpc_request_counter` / `sei_rpc_request_latency_ms` / `sei_websocket_connect`) are still dual-emitted during the migration window so dashboards continue to receive data. They carry TODO markers and will be removed once dashboards are migrated to the new `evmrpc_*` OTel series.

### `success` label semantic correction

The old `recordMetricsWithError` had a logic bug — `success := panicValue == nil || err != nil` — which caused requests that returned an error without panicking to be recorded as `success=true`. This PR fixes that to `success := panicValue == nil && err == nil`.

The correction applies to both the new OTel metric and the dual-emitted legacy `sei_rpc_request_counter` series.

**Deploy-time impact:** A search of the platform repo found no Prometheus alert rules referencing these metrics, so no alerts will fire. Two Grafana dashboard panels filter on `success="false"` and will show a spike at deploy time:

- `clusters/prod/monitoring/grafana-dashboards-protocol.yaml` — EVM RPC failed requests panel
- `clusters/dev/monitoring/dashboards/protocol/sei_node_monitoring.json` — EVM RPC failed requests panel

This spike is a visual artifact of the correction (previously miscounted errors now classified correctly), not a real increase in failures. No action is required, but heads-up for anyone watching these panels during the rollout.


evmrpc/utils.go
```diff
@@ -293,15 +293,15 @@ func recordMetrics(apiMethod string, connectionType ConnectionType, startTime ti
 func recordMetricsWithError(apiMethod string, connectionType ConnectionType, startTime time.Time, err error) {
 	// Automatically detect success/failure based on panic state
 	panicValue := recover()
-	success := panicValue == nil || err != nil
+	success := panicValue == nil && err == nil
```
amir-deris, line 296: The previous success value would become true as long as there was no panic (due to short circuit for OR), even in cases that err was not nil. Updated here to account for both panic and err.

evmrpc/rpc_telemetry.go
```diff
@@ -0,0 +1,80 @@
...
+		filterLogFetchBatches: must(rpcTelemetryMeter.Int64Counter(
+			"evmrpc_filter_log_fetch_batches_total",
+			metric.WithDescription("Internal filter/getLogs block batches completed (per pipeline path, not per RPC)"),
+			metric.WithUnit("{batch}"),
+		)),
+	}
+)
+
+func recordRPCRequest(endpoint, connection string, success bool) {
+	ctx := context.Background()
```
bdchatham, line 49: nit: we should either 1) take a context when the struct is built or 2) take a context through the api. 

1. is not suitable because we are creating a singleton object effectively statically.
2. is the right way to go, although will expand the scope of the refactor. 

Once you pass this in, you can provide it to the delegated methods directly instead of generating anew.

evmrpc/rpc_telemetry.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func recordWebsocketConnect() {
+	rpcTelemetryMetrics.websocketConnects.Add(context.Background(), 1)
+}
+
+func recordFilterLogFetchBatchComplete(pipeline string) {
+	ctx := context.Background()
+	rpcTelemetryMetrics.filterLogFetchBatches.Add(ctx, 1,
+		metric.WithAttributes(attribute.String("pipeline", pipeline)),
```
bdchatham: Pipeline is a bit of an ambiguous attribute name. Can you fill me in on your thinking?

amir-deris: I agree with you, the pipeline doesn't seem like a sensible name here. It seems the label distinguishes block-fetch worker batches vs log-extraction worker batches in the filter path,. How about the following options, do you prefer any of them?
`stage`
`phase`
`batch_kind` or `kind`
`operation`

evmrpc/rpc_telemetry.go
amir-deris: Sounds good. I will work on receiving the context as an argument for option 2.

evmrpc/info.go
```diff
@@ -59,19 +59,19 @@ func (i *InfoAPI) BlockNumber() hexutil.Uint64 {
...
 	startTime := time.Now()
-	defer recordMetricsWithError("eth_Coinbase", i.connectionType, startTime, err)
+	defer recordMetricsWithError(context.Background(), "eth_Coinbase", i.connectionType, startTime, err)
 	return i.keeper.GetFeeCollectorAddress(i.ctxProvider(LatestCtxHeight))
 }
 
 func (i *InfoAPI) Accounts() (result []common.Address, returnErr error) {
 	startTime := time.Now()
-	defer recordMetricsWithError("eth_Accounts", i.connectionType, startTime, returnErr)
+	defer recordMetricsWithError(context.Background(), "eth_Accounts", i.connectionType, startTime, returnErr)
```
bdchatham: We should sort these out to use real request-scoped contexts. Any code path that creates a background context (e.g. context.Background() or context.TODO()) deep in the stack is a smell. It means we've lost the request context that should have been passed from the caller.

The way this should work is every incoming request gets its own context at the API layer, and that context gets threaded down through the stack so each layer can use it. OTel wires this up automatically. Once the SDK and your API framework are wired up, spans and trace data get attached to that request context as it flows through, without you having to do anything extra per call site. This is how OTel makes our existing setup more powerful out of the box.

So the two things we need to close the loop on our telemetry strategy are: 1) make sure the OTel SDK and API framework integration are configured correctly so trace propagation is set up at the boundary, and 2) make sure every handler is passing its request context down the call stack rather than creating new detached contexts along the way.

evmrpc/rpc_telemetry.go
```diff
@@ -0,0 +1,77 @@
...
+		)),
+		websocketConnects: must(rpcTelemetryMeter.Int64Counter(
+			"evmrpc_websocket_connects_total",
+			metric.WithDescription("Number of new websocket connections"),
+			metric.WithUnit("{count}"),
+		)),
+		filterLogFetchBatches: must(rpcTelemetryMeter.Int64Counter(
+			"evmrpc_filter_log_fetch_batches_total",
+			metric.WithDescription("Internal filter/getLogs block batches completed (per pipeline path, not per RPC)"),
+			metric.WithUnit("{batch}"),
```
bdchatham: nit: I'd use unit `count`.

evmrpc/rpc_telemetry.go
bdchatham: Let's take a step back to reframe this. 

What are we trying to measure with this metric: 

```
filterLogFetchBatches: must(rpcTelemetryMeter.Int64Counter(
			"evmrpc_filter_log_fetch_batches_total",
			metric.WithDescription("Internal filter/getLogs block batches completed (per pipeline path, not per RPC)"),
			metric.WithUnit("{batch}")
```

For instance, as a service owner, when I am looking into performance or operational behavior, what does the count tell me about the operation of the service? And for the questions I am trying to use it to answer, does it give me the whole answer? Are other metrics needed to supplement the reasoning?

This is a meta question so don't feel like this should all be deduced from the line of code we're looking at 😄

evmrpc/info.go
amir-deris: Thanks for feedback. Sounds good, I will work on adding the context correctly and not creating context.Background on the fly.

evmrpc/rpc_telemetry.go
amir-deris: That is a great point. I believe the intention initially was to keep the existing metrics and only do migration to OTEL. But now that we are doing this work, it makes sense to step back and think about the big picture and the value this initiative is going to bring. Along the way, we can decide which signals bring value and we should keep. 
In this particular instance, it doesn't seem this metric is bringing much value and is not part of any dashboards I can find. So I can remove it if that is ok.

evmrpc/filter.go
```diff
@@ -1231,10 +1227,6 @@ func (f *LogFetcher) fetchBlocksByCrit(ctx context.Context, crit filters.FilterC
 
 // Batch processing function for blocks
 func (f *LogFetcher) processBatch(ctx context.Context, start, end int64, crit filters.FilterCriteria, bloomIndexes [][]BloomIndexes, res chan *coretypes.ResultBlock, errChan chan error) {
-	defer func() {
-		metrics.IncrementRpcRequestCounter("num_blocks_fetched", "blocks", true)
-	}()
-
```
bdchatham, line 1237: I'm aligned with removing this. The naming is not representative of what it's actually measuring. 

Only argument I could see for keeping this is if any of our own telemetry or external infra providers would use this. 

@masih - technically this could break external party's observability using this. The value of this metric seems limited at best. I'm thinking we remove it and replace it with a better one in a follow-up,

wdyt?

bdchatham (approved):

evmrpc/rpc_telemetry.go
```diff
@@ -0,0 +1,65 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+)
+
+// JSON-RPC and websocket connection metrics use the OpenTelemetry Meter API.
+// The process-wide MeterProvider (e.g. Prometheus exporter with namespace) is
+// configured by the application.
+
+var (
+	rpcTelemetryMeter = otel.Meter("evmrpc_rpc")
```
masih, line 17: * There would be a single `meter` this entire package. I would simply name this `meter`.
* I recommend removing the repetitive `_rpc` suffix.

evmrpc/rpc_telemetry.go
```diff
@@ -0,0 +1,65 @@
...
+
+	rpcTelemetryMetrics = struct {
```
masih, line 19: Similarly, I would call this `metrics`.

evmrpc/rpc_telemetry.go
```diff
@@ -0,0 +1,65 @@
...
+		requests          metric.Int64Counter
```
masih, line 20: * General note on naming: this reads too abstract. I would add a `Count` at the end of it.
* The counter itself is redundant, if we always measure latency in a histogram. See my other comment.

evmrpc/rpc_telemetry.go
```diff
@@ -0,0 +1,65 @@
...
+		requestLatencyMs  metric.Float64Histogram
```
masih, line 21: _Always_ standardise on `seconds` for latency/duration measurements. Duration.Seconds() conveniently returns a float64 without the loss of precision.

evmrpc/rpc_telemetry.go
```diff
@@ -0,0 +1,65 @@
...
+		websocketConnects metric.Int64Counter
```
masih, line 22: Similarly I would add a Count at the end, and perhaps pick a simpler name: `wsConnectionCount` or similar.

evmrpc/rpc_telemetry_test.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"context"
+	"testing"
+	"time"
+)
+
+func TestRecordRPCMetricsNoPanic(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
```
masih, line 11: Use `t.Context()`.

evmrpc/tx_test.go
```diff
@@ -122,15 +123,15 @@ func TestSign(t *testing.T) {
 	require.Nil(t, err)
 	_, err = kb.NewAccount("test", mnemonic, "", hd.CreateHDPath(sdk.GetConfig().GetCoinType(), 0, 0).String(), algo)
 	require.Nil(t, err)
-	accounts, _ := infoApi.Accounts()
+	accounts, _ := infoApi.Accounts(context.Background())
```
masih, line 126: Use `t.Context()` across tests. Ditto in other places.

evmrpc/utils.go
```diff
@@ -286,22 +286,22 @@ func filterTransactions(
...
 
 	// these are only metrics that are specifically typed errors for tracking.
 	if err != nil {
 		metrics.IncrementErrorMetrics(apiMethod, err)
 	}
 
-	metrics.IncrementRpcRequestCounter(apiMethod, string(connectionType), success)
-	metrics.MeasureRpcRequestLatency(apiMethod, string(connectionType), startTime)
+	recordRPCRequest(ctx, apiMethod, string(connectionType), success)
```
masih, line 303: This is the only place we call record count and latency.

Because latency is a histogram, I can use it to also get the "total". This makes a separate counter for total number of requests redundant.

Try running the code and browsing the exposed metrics on prometheus endpoint

evmrpc/websockets.go
```diff
@@ -2,16 +2,14 @@ package evmrpc
...
-	"github.com/sei-protocol/sei-chain/utils/metrics"
 )
 
 type wsConnectionHandler struct {
 	underlying http.Handler
 }
 
 func (h *wsConnectionHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	metrics.IncWebsocketConnects()
+	recordWebsocketConnect(r.Context())
```
masih, line 12: Is count of connection the only thing we want to measure for web sockets? That would tell me the rate of new connections but will not tell me the number of _currently connected clients_ ? Feel free to iterate on this in a follw up PR but from operational perspective, I envision an operator to need more data.

evmrpc/rpc_telemetry.go
```diff
@@ -0,0 +1,65 @@
...
+			metric.WithDescription("Number of new websocket connections"),
+			metric.WithUnit("{count}"),
+		)),
+	}
+)
+
+func recordRPCRequest(ctx context.Context, endpoint, connection string, success bool) {
+	rpcTelemetryMetrics.requests.Add(ctx, 1,
+		metric.WithAttributes(
+			attribute.String("endpoint", endpoint),
```
masih, line 45: Define attrKeys for repeated tags as vars?

evmrpc/rpc_telemetry.go
```diff
@@ -0,0 +1,65 @@
...
+			attribute.String("connection", connection),
+			attribute.Bool("success", success),
```
masih, line 47: A better approach is to take error and tag by error type for a fixed set of well known errors. 

Also do we measure JSON RPC response code somewhere? that would be super useful.

evmrpc/filter.go
masih: I am OK with removing this. I recommend looking into git history to see why this was added. I suspect it was added to debug a point of potential contention.

evmrpc/association.go
```diff
@@ -59,7 +59,7 @@ type AssociateRequest struct {
 
 func (t *AssociationAPI) Associate(ctx context.Context, req *AssociateRequest) (returnErr error) {
 	startTime := time.Now()
-	defer recordMetricsWithError("sei_associate", t.connectionType, startTime, returnErr)
+	defer recordMetricsWithError(ctx, "sei_associate", t.connectionType, startTime, returnErr)
```
masih, line 62: Why are we keeping `recordMetricsWithError` calls at all? 

Is the idea to not break the existing metrics, roll out both then remove the old stuff?

evmrpc/rpc_telemetry.go
```diff
@@ -0,0 +1,65 @@
+package evmrpc
```
masih, line 1: Conventionally I would simply call this file `metrics.go`. Each package would have one as needed. that go file would initialise once on startup and that's it; all package level vars.

masih (commented):
Yay glad to see OTEL integration picking up 🚀 

Blockers:
* standardise on `seconds` to measure time/duration/latency
* Remove redundant counters

Left a bunch of naming suggestions and a few questions 

Thanks @amir-deris 🙌

evmrpc/association.go
amir-deris: I thought we need to keep these metrics to be backward compatible with any dashboards we might have. Should I remove these metrics?

evmrpc/filter.go
amir-deris: It seems it was added in this PR as part of regular implementation (not debugging or hotfixing):
https://github.com/sei-protocol/sei-chain/pull/2195/

evmrpc/rpc_telemetry.go
amir-deris: Actually I just noticed there is another meter definition in `evmrpc/worker_pool_metrics.go` and its meter definition clashes with this one if we rename this `meter`:
`meter = otel.Meter("evmrpc_workerpool")`
Should we combine all meters into one file (`metrics.go`)?

evmrpc/utils.go
amir-deris: Yes you are right. It seems otel automatically sends 3 data points to prometheus backend:
`evmrpc_request_latency_seconds_bucket{le="..."}`
`evmrpc_request_latency_seconds_count`
`evmrpc_request_latency_seconds_sum`
The `_count` one already captures the number of requests, so we don't need another separate metric for that. I will remove it.

evmrpc/rpc_telemetry.go
amir-deris: I will remove this counter, as discussed in histogram comment.

evmrpc/rpc_telemetry.go
amir-deris: I added new method `classifyRPCMetricError` to record error type and response code with latency.

evmrpc/websockets.go
amir-deris: Captured this new issue: https://linear.app/seilabs/issue/PLT-277/adding-otel-metrics-for-currently-connected-clients-of-evmrpc-web

masih (commented):
THis is a stacked PR. 

I advise against stacked PRs in general and the work is done from dev perspective once it lands on main.

Approving and will re-review the full work on PR to main.
Please make sure the final PR doesn't get too big otherwise the punishment is to buy ice cream for the team 🍨 😜

masih (approved):

evmrpc/utils.go
```diff
@@ -286,22 +286,24 @@ func filterTransactions(
...
-func recordMetrics(apiMethod string, connectionType ConnectionType, startTime time.Time) {
-	recordMetricsWithError(apiMethod, connectionType, startTime, nil)
+func recordMetrics(ctx context.Context, apiMethod string, connectionType ConnectionType, startTime time.Time) {
+	recordMetricsWithError(ctx, apiMethod, connectionType, startTime, nil)
 }
 
-func recordMetricsWithError(apiMethod string, connectionType ConnectionType, startTime time.Time, err error) {
+func recordMetricsWithError(ctx context.Context, apiMethod string, connectionType ConnectionType, startTime time.Time, err error) {
 	// Automatically detect success/failure based on panic state
 	panicValue := recover()
```
bdchatham: **This `recover()` no longer catches anything.**

The pre-PR pattern `defer recordMetricsWithError(...)` made `recordMetricsWithError` itself the deferred function, so `recover()` worked. After this PR, every caller wraps as `defer func() { recordMetricsWithError(...) }()` (e.g. `association.go:60`, `info.go:88`, `send.go:73`) to capture the named return. Per the [Go spec](https://go.dev/ref/spec#Handling_panics): _"recover returns nil if recover was not called directly by a deferred function."_ `recordMetricsWithError` is now invoked _by_ the closure, not deferred itself — so `panicValue` is always nil.

Consequences:
- `errorClassPanic` is dead — that label value is now unreachable.
- `success` evaluates `true` on a panicked handler that returned a zero-value `err`.
- The closing `if panicValue != nil { panic(panicValue) }` re-raise never fires (panics still propagate up to whichever recover does run, but this site's metric labeling is wrong).

Fix: have the closure call `recover()` and pass the result in:
```go
defer func() {
    recordMetricsWithError(ctx, "sei_associate", t.connectionType, startTime, returnErr, recover())
}()
```
…and accept `panicValue any` here instead of calling `recover()` internally.

evmrpc/utils.go
```diff
@@ -286,22 +286,24 @@ func filterTransactions(
...
-	success := panicValue == nil || err != nil
+	success := panicValue == nil && err == nil
```
bdchatham, line 294: Correctness fix — but this silently changes the meaning of the **legacy** dual-emitted `sei_rpc_request_counter{success="true"}` series too.

The old `||` was buggy: non-panicking errors were marked `success=true`. After this lands, those flip to `success=false` on the legacy series. Existing alerts / SLO queries shaped like `rate(sei_rpc_request_counter{success="true"}[5m]) / rate(sei_rpc_request_counter[5m])` will step-down at deploy.

The PR description says "existing dashboards continue to work" — true at the metric-name level, but not at the semantic level. Suggest:
- Call this out explicitly in the rollout note.
- Sweep `sei-protocol/sei-infra/monitoring/` and the alerting config for queries filtering on `success` and adjust in the same release window.

evmrpc/utils.go
```diff
@@ -286,22 +286,24 @@ func filterTransactions(
...
 	// these are only metrics that are specifically typed errors for tracking.
 	if err != nil {
-		metrics.IncrementErrorMetrics(apiMethod, err)
+		utilmetrics.IncrementErrorMetrics(apiMethod, err)
 	}
 
-	metrics.IncrementRpcRequestCounter(apiMethod, string(connectionType), success)
-	metrics.MeasureRpcRequestLatency(apiMethod, string(connectionType), startTime)
+	recordRPCLatency(ctx, apiMethod, string(connectionType), success, err, panicValue != nil, startTime)
+	// TODO: remove legacy dual-emit once dashboards are migrated to evmrpc_* OTEL metrics. Use metrics.requestLatencySeconds histogram instead.
```
bdchatham: Bare `TODO`s for "remove later" reliably rot. Suggest:

1. File a tracking issue with explicit cleanup criteria — dashboards / alerts / recording rules migrated to `evmrpc_*`, ≥2 weeks of new-metric retention, then drop the dual emission.
2. Replace the TODO bodies here, in `evmrpc/websockets.go`, and on the three legacy helpers in `utils/metrics/metrics_util.go` with `// TODO(PLT-XXX): ...` linking the issue.

Otherwise the dual emission + paired `sei_*` helpers will outlive everyone's memory of why they're here.

evmrpc/metrics.go
```diff
@@ -0,0 +1,106 @@
...
+
+	metrics = struct {
+		requestLatencySeconds metric.Float64Histogram
+		wsConnectionCount     metric.Int64Counter
+	}{
+		requestLatencySeconds: must(rpcTelemetryMeter.Float64Histogram(
+			"evmrpc_request_latency_seconds",
+			metric.WithDescription("RPC request latency in seconds (labeled by success, bounded error_class, and jsonrpc_code when known)"),
+			metric.WithUnit("s"),
+		)),
```
bdchatham, line 54: Two histogram-shape concerns worth addressing before this scales out:

**Cardinality.** `jsonrpc_code` is an Int64 attribute; for `error_class=jsonrpc_error` it falls through to whatever `coder.ErrorCode()` returns. JSON-RPC 2.0 reserves -32000..-32099 for server-defined codes, plus go-ethereum surfaces additional ones. Worst-case bound (~30 endpoints × 2 connections × 7 error_class × ~30 codes seen × ~12 default buckets) is ~150k series per pod for this one histogram. Either:
- Bucket the code on the histogram (e.g. `"spec"` / `"server"` / `"other"`), or
- Drop `jsonrpc_code` from the histogram and put it on a low-cost counter only.

Also, `success` and `error_class` are partially redundant — drop `error_class` (or set to empty) on the success path so the high-volume happy path doesn't carry both.

**Bucket boundaries.** Default OTel buckets start at 5ms; cached `eth_chainId` / `net_version` collapse to bucket 0, and `debug_traceBlockByNumber` runs past the 10s ceiling. Consider explicit boundaries:

```go
metric.WithExplicitBucketBoundaries(
    0.0005, 0.001, 0.0025, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25,
    0.5, 1, 2.5, 5, 10, 30,
)
```

evmrpc/metrics.go
```diff
@@ -0,0 +1,106 @@
...
+		wsConnectionCount: must(rpcTelemetryMeter.Int64Counter(
+			"evmrpc_websocket_connects_total",
```
bdchatham, line 56: **`_total` is auto-appended on counters** by the OTel Prometheus exporter (default behavior in `go.opentelemetry.io/otel/exporters/prometheus`). Naming the instrument `evmrpc_websocket_connects_total` likely emits `sei_chain_evmrpc_websocket_connects_total_total` after the namespace prefix and counter-suffix pass.

Worth verifying on a running pod's `/metrics` before the dashboard PR ships. Fix is to rename the instrument to `evmrpc_websocket_connects` and let the exporter add the suffix.

(The dashboard PR doesn't query this counter directly so it should be unaffected, but anyone writing future alerts against it would hit the double suffix.)

bdchatham (commented):
Fresh expert pass on top of my approval — five callouts worth considering. (1) is a real bug from the named-return wrapping pattern; (2) is a deploy-coordination concern that contradicts the "existing dashboards continue to work" framing; (3)–(5) are good-to-fix.

evmrpc/utils.go
amir-deris: Thanks for feedback. That is a good point, it seems we don't have any alerts based on the success attributes. We only have two existing dashboards which filter for success=false and can potentially show a spike after the rollout of this PR's changes, due to how we now correctly assign error status:
| File | Query | Impact |
|------|-------|--------|
| `clusters/prod/monitoring/grafana-dashboards-protocol.yaml` (line 19189) | `sei_rpc_request_counter{success="false"}` | Will spike up at deploy |
| `clusters/dev/monitoring/dashboards/protocol/sei_node_monitoring.json` (line 4251) | `sei_rpc_request_counter{success="false"}` | Will spike up at deploy |
  
 
Updated PR description accordingly.

evmrpc/utils.go
amir-deris: Done. Created this ticket https://linear.app/seilabs/issue/PLT-326
Also updated the TODO comments with ticket number.l

evmrpc/metrics.go
amir-deris: Noted. Fixed the histogram bucket boundaries and also added specific buckets for the error codes.

evmrpc/metrics.go
amir-deris: I verified this with the local cluster: `make docker-cluster-start`, then making websocket request:
`printf '%s\n' '{"jsonrpc":"2.0","method":"eth_chainId","params":[],"id":1}' | websocat -n1 ws://127.0.0.1:8546/`
Then checking the metrics in the container:
`curl -s http://localhost:26660/metrics | grep evmrpc_websocket_connects_total`
It shows:
`sei_chain_evmrpc_websocket_connects_total{otel_scope_name="evmrpc",otel_scope_schema_url="",otel_scope_version=""} 5`
so it seems the `_total` is **not** doubled. 
I also reviewed the otel package `go.opentelemetry.io/otel/exporters/prometheus` with Claude and it seems there is some logic in there to not duplicate the _total:
`The naming is handled entirely by otlptranslator — your instrument name evmrpc_websocket_connects_total gets tokenized, "total" is stripped and re-appended at the end, so you always get 
  exactly one _total. No code change needed on your side.`

PR #3267, opened 2026-04-16
Add debug_traceTransactionProfile endpoint

Kbhat1:
- Adds `debug_traceTransactionProfile`, a trace endpoint that returns the normal tx trace plus a granular timing/profile breakdown
- Captures store-level access data during tracing, including module stats, iterator activity, and low-level read events when supported
- Cleans up trace-scoped resources after each request to avoid leaking snapshots or iterators.

- Verified end to end on node


sei-cosmos/store/gaskv/store.go
```diff
@@ -73,19 +77,21 @@ func (gs *Store) Set(key []byte, value []byte) {
...
 	gs.parent.Set(key, value)
 	if gs.tracer != nil {
-		gs.tracer.Set(key, value, gs.moduleName)
+		gs.tracer.Set(key, value, gs.moduleName, time.Since(start))
 	}
 }
 
 // Implements KVStore.
 func (gs *Store) Has(key []byte) bool {
 	defer telemetry.MeasureSince(time.Now(), "store", "gaskv", "has")
```
yzang2019, line 98: Suggest remove this metric as perf optimization, and for all operations including Get/Set/Has/Delete/Iterator.

sei-cosmos/store/gaskv/store.go
```diff
@@ -73,19 +77,21 @@ func (gs *Store) Set(key []byte, value []byte) {
...
 	gs.gasMeter.ConsumeGas(gs.gasConfig.HasCost, types.GasHasDesc)
+	start := time.Now()
```
yzang2019: Suggest moving this time.Now() inside: 
if gs.tracer != nil {
  // move here
}
 to save the unnecessary time.Now() call when tracer is off. Same for all other functions

sei-cosmos/store/gaskv/store.go
```diff
@@ -73,19 +77,21 @@ func (gs *Store) Set(key []byte, value []byte) {
...
 	res := gs.parent.Has(key)
 	if gs.tracer != nil && res {
```
yzang2019: I think this needs a fix, even if Has() doesn't find a key, it still tried to access the database and did all the lookup, so we should still trace and record it

sei-cosmos/store/gaskv/store.go
Kbhat1: Yea, this is from a older PR, I'll create a new one to remove

yzang2019 (approved):

jewei1997 (approved):

PR #3268, opened 2026-04-17
Giga storage integration test

yzang2019:
This PR enable giga storage for some integration tests. We started by doing SC dual writes + split read, SS split write + split read.


cody-littley (approved):
LGTM, but the way we override config values seems super fragile. Perhaps a discussion for a later time is how we can structure config in a way that makes this sort of override easier and simpler.

Kbhat1 (approved):

yzang2019:
> LGTM, but the way we override config values seems super fragile. Perhaps a discussion for a later time is how we can structure config in a way that makes this sort of override easier and simpler.

Yeah, that worth a discussion and follow up PR to refactor

PR #3270, opened 2026-04-17
new short term election algorithm

pompon0:
The new algorithm is choosing leader of (height,round), based on hash(height,round) and weights of validators (public lottery). The new algorithm is enabled via config flag: Consensus.StatelessLeaderElection (false by default). It will be the only option in 6.5. Additional changes:
* updated crypto package to include zeroize fix (otherwise it panics under a new compiler)
* made consensus tests run against both stateful and stateless leader election. The caveat is that many of the consensus tests are very poor quality and they hardcode the expected leader schedule. To minimize the amount of work, these tests have been adjusted in one of the following ways: a) the test harness looks for a starting round which produces the leader schedule that the test expects; b) the test harness adjusts the weights so that the leader schedule matches what the test expects.
* I have manually run the tests outside of the consensus module with StatelessLeaderElection = true, however these are not included in CI.

### DIFF
--- a/sei-tendermint/config/config.go
+++ b/sei-tendermint/config/config.go
@@ -1117,6 +1117,8 @@ type ConsensusConfig struct {
 
 	DoubleSignCheckHeight int64 `mapstructure:"double-sign-check-height"`
 
+	StatelessLeaderElection bool `mapstructure:"stateless-leader-election"`
+
 	// TODO: The following fields are all temporary overrides that should exist only
 	// for the duration of the v0.36 release. The below fields should be completely
 	// removed in the v0.37 release of Tendermint.
@@ -1176,6 +1178,7 @@ func DefaultConsensusConfig() *ConsensusConfig {
 		DoubleSignCheckHeight:       int64(0),
 		// Sei Configurations
 		GossipTransactionKeyOnly: true,
+		StatelessLeaderElection: false,
 	}
 }
 
--- a/sei-tendermint/config/config.go
+++ b/sei-tendermint/config/config.go
@@ -1118,6 +1118,9 @@ type ConsensusConfig struct {
 	DoubleSignCheckHeight int64 `mapstructure:"double-sign-check-height"`
 
 	// Whether the new stateless leader election should be used.
+	// THIS IS A TEMPORARY DISASTER RECOVERY MECHANISM IN CASE OF A CHAIN STALL.
+	// REQUIRES COORDINATION OF MAJORITY OF VALIDATORS TO SET TO TRUE.
+	// IF YOU SET IT TO TRUE JUST FOR YOUR NODE, IT WILL BE UNABLE TO PARTICIPATE IN THE CONSENSUS.
 	StatelessLeaderElection bool `mapstructure:"stateless-leader-election"`
 
 	// TODO: The following fields are all temporary overrides that should exist only
--- a/sei-tendermint/config/config.go
+++ b/sei-tendermint/config/config.go
@@ -1179,7 +1179,7 @@ func DefaultConsensusConfig() *ConsensusConfig {
 		DoubleSignCheckHeight:       int64(0),
 		// Sei Configurations
 		GossipTransactionKeyOnly: true,
-		StatelessLeaderElection:  false,
+		StatelessLeaderElection:  true,
 	}
 }
 
--- a/sei-tendermint/config/config.go
+++ b/sei-tendermint/config/config.go
@@ -1179,7 +1179,7 @@ func DefaultConsensusConfig() *ConsensusConfig {
 		DoubleSignCheckHeight:       int64(0),
 		// Sei Configurations
 		GossipTransactionKeyOnly: true,
-		StatelessLeaderElection:  true,
+		StatelessLeaderElection:  false,
 	}
 }
 
--- a/sei-tendermint/config/config.go
+++ b/sei-tendermint/config/config.go
@@ -1117,6 +1117,7 @@ type ConsensusConfig struct {
 
 	DoubleSignCheckHeight int64 `mapstructure:"double-sign-check-height"`
 
+	// Whether the new stateless leader election should be used.
 	StatelessLeaderElection bool `mapstructure:"stateless-leader-election"`
 
 	// TODO: The following fields are all temporary overrides that should exist only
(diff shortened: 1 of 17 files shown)


sei-will (approved):

sei-tendermint/config/config.go
```diff
@@ -1176,6 +1179,7 @@ func DefaultConsensusConfig() *ConsensusConfig {
 		DoubleSignCheckHeight:       int64(0),
 		// Sei Configurations
 		GossipTransactionKeyOnly: true,
+		StatelessLeaderElection:  false,
```
wen-coding, line 1185: nit: we should add some warning it's dangerous to change this value not in coordination with other validators.

sei-tendermint/internal/consensus/types/round_state.go
```diff
@@ -388,6 +410,39 @@ type RoundState struct {
 	TriggeredTimeoutPrecommit bool
 }
 
+// 32 bytes crypto hash seed, generated via random.org.
+var leaderElectionSeed = [32]byte(utils.OrPanic1(hex.DecodeString(
+	"3793f16d412703e5805755e5282f681c70e771f151c8864c656c6c259243f85f",
```
wen-coding, line 416: nit: I understand why you are doing this, but I'm a bit worried that if someone changes one byte in the middle, no one will ever notice, maybe it's better to put a meaningful string there?

sei-tendermint/internal/consensus/types/round_state.go
```diff
@@ -388,6 +410,39 @@ type RoundState struct {
...
+)))
+
+// Leader for each round is drawn at random from the validator set with
+// probabilities proportional to the voting powers.
+//
+// The following pseudorandom function is used to select the leader:
+// pos(height,round) := hash(height ++ round)
+// Validators are assigned subintervals of [0,TotalVotingPower) of length
+// equal to their voting poser.
+// Validator i is the leader of (height,round) <=> pos(height,round)%TotalVotingPower \in validator_interval[i]
```
wen-coding, line 426: Are we planning to do per-height calculation later?

sei-tendermint/types/validator_set.go
```diff
@@ -62,6 +64,10 @@ type ValidatorSet struct {
 	totalVotingPower int64
 }
 
+func (vals *ValidatorSet) All() iter.Seq[*Validator] {
```
wen-coding: nit: add comment why the returned set is deterministic? We might need to warn future AI agent that do not mutate Validators without sorting them by ValidatorsByVotingPower?

sei-tendermint/internal/consensus/common_test.go
```diff
@@ -555,18 +557,163 @@ func makeState(ctx context.Context, t *testing.T, args makeStateArgs) (*State, [
 	})
 
 	vss := make([]*validatorStub, validators)
+	localIndex := 0
```
wen-coding, line 560: Had a very quick look over the tests, I think after you merge PR#3273 I'll check again.

sei-tendermint/internal/consensus/state.go
```diff
@@ -1150,7 +1151,7 @@ func (cs *State) enterPropose(ctx context.Context, height int64, round int32, en
 
 	// If this validator is the proposer of this round, and the previous block time is later than
 	// our local clock time, wait to propose until our local clock time has passed the block time.
-	if key, ok := cs.privValidatorPubKey.Get(); ok && cs.isProposer(key.Address()) {
+	if key, ok := cs.privValidatorPubKey.Get(); ok && cs.roundState.Leader() == key {
```
wen-coding, line 1154: nit: Would bytes.Equal(leader.Address(), key.Address()) be more future proof?

pompon0: why would it? Address is derived from public key. Honestly I don't understand why the concept of Address exists - to save space maybe?

sei-tendermint/internal/consensus/types/round_state.go
pompon0: That's something I'm opinionated on - why do you consider a string to be a better option? String can be changed as any other value. String is more prone to accidental collision because it has less entropy. I can add a change detector test to be verbose about the fact that it is unchangeable (no matter what the type is)

sei-tendermint/internal/consensus/types/round_state.go
pompon0: yup, for the long term solution. And I would like first to compute what is the actual impact on block latency if we intriduce it.

sei-tendermint/config/config.go
pompon0: done

sei-tendermint/internal/consensus/types/round_state.go
pompon0: added a test

wen-coding: sure, that works.

sei-tendermint/internal/consensus/common_test.go
pompon0: merged

sei-tendermint/types/validator_set.go
pompon0: Renamed to Ordered and documented the function. If the ordering changes, it should be caught by the change detector test that I've added.

sei-tendermint/internal/consensus/types/round_state_test.go
```diff
@@ -0,0 +1,57 @@
...
+		"round-state-leader-2",
+		"round-state-leader-3",
+		"round-state-leader-4",
+		"round-state-leader-5",
+		"round-state-leader-6",
+		"round-state-leader-7",
+		"round-state-leader-8",
+		"round-state-leader-9",
+	}
+	weights := [10]int64{5, 6, 7, 8, 9, 10, 5, 6, 7, 8}
```
wen-coding, line 31: nit: do you think it helps to add a test which runs 10k selections and verify the distribution looks roughly correct?

wen-coding (approved):

sei-tendermint/internal/consensus/types/round_state_test.go
pompon0: added

PR #3272, opened 2026-04-20
Replace bump-version.sh with pure-Go go:generate tool

masih:
Introduce scripts/bump_version, a combined archival + code-generation tool that replaces the three shell scripts (bump-version.sh, generate-precompile-setup.sh, generate-all-precompiles-setup.sh).

New developer workflow:
  1. Commit precompile code changes
  2. Append a new upgrade name (vMajor.Minor) to app/tags
  3. Run go generate ./...

What the tool does on each run:
- Validates the latest entry in app/tags (must be vMajor.Minor; patch is dropped with a warning; duplicate or regressed versions are errors)
- Validates each module's versions file (valid semver, no duplicates)
- Uses go-git (pure Go, no os/exec) to find the commit that introduced the previous upgrade version, then diffs per-module tree hashes against HEAD to determine which precompile modules actually changed
- Archives changed modules into precompiles/<module>/legacy/<folder>/ with package/import rewrites; archives common/ when it changed (triggers all)
- Fails hard on git errors (matching old bump-version.sh set -e behavior) rather than silently falling back to archiving everything
- Regenerates precompiles/<module>/setup.go from a text/template (.go.tmpl) for every module with a versions file, formatted via go/format
- Is idempotent: skips archival if precompiles/common/legacy/<folder>/ exists

Naming conventions enforced:
  app/tags / versions keys: vMajor.Minor  (e.g. v7.0) legacy folder / package:  vMajorMinor  (e.g. v70) -- Idropped the zero at the end given v6.5 was generated with that pattern before I could land this PR.

All generated .go files carry a "Code generated by scripts/bump_version; DO NOT EDIT." header. Added go-git/v5 as the sole new dependency.


arajasek (approved):

philipsu522 (approved):
awesome, so much cleaner. a minor follow up question: does it make sense to run this in ci somehow? i.e. all we need to do is `echo "v7.0" >> app/tags` and the artifacts are autogenerated?

masih:
> does it make sense to run this in ci somehow? i.e. all we need to do is `echo "v7.0" >> app/tags` and the artifacts are autogenerated?

As in doing the pre-complie generation itself, right? Totally; We can have a workflow like this:
* When it is "Release Time", open a PR and add the new upgrade name line to `tags`
* There would then be a CI job that gets triggered by changes to `tags` file, does the `go generate ./...` and pushes the changes to your PR for you. @seidroid can be the agent.

~Will~ [Captured an issue](https://linear.app/seilabs/issue/PLT-320) for this; would make a nice cherry on top. 🍰

codchen (approved):

PR #3274, opened 2026-04-20
evmrpc: refactor getHeader to always require tmBlock, fix CurrentHeader block fetch

amir-deris:
## Summary

- `getHeader` previously accepted an optional `blockNumber *big.Int` and a nullable `tmBlock`, doing its own block fetch when `tmBlock` was nil. This PR removes that dual path: `getHeader` now always receives a non-nil `tmBlock` and derives the height from it directly.
- `CurrentHeader` is updated to accept a `context.Context`, fetch the Tendermint block itself, and call `getHeader` with it. A new `fallbackEthereumHeaderWithoutTendermintBlock` helper handles the error path (block RPC unavailable).
- `header.Time` now uses the actual block timestamp (`tmBlock.Block.Time`) instead of `time.Now()`, and `header.ParentHash` is always populated since `tmBlock` is never nil.


evmrpc/simulate.go
```diff
@@ -554,7 +553,14 @@ func (b *Backend) GetEVM(_ context.Context, msg *core.Message, stateDB vm.StateD
 }
 
 func (b *Backend) CurrentHeader() *ethtypes.Header {
-	header := b.getHeader(context.Background(), big.NewInt(b.ctxProvider(LatestCtxHeight).BlockHeight()), nil)
+	height := b.ctxProvider(LatestCtxHeight).BlockHeight()
+	ctx := context.Background()
```
bdchatham, line 557: Out of scope for this PR, but a first glance this seems off but something to look into later.

bdchatham (approved):

evmrpc/simulate.go
amir-deris: That is a good point. I actually tried passing in context as an argument for `CurrentHeader`, but the issue is the `Backend` is supposed to implement the interface in go ethereum:
https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/backend.go#L64
and passing an extra argument breaks that contract.

evmrpc/simulate.go
```diff
@@ -599,49 +605,44 @@ func (b *Backend) getBlockByNumberOrHash(ctx context.Context, blockNrOrHash rpc.
 	return block, isLatestBlock, nil
 }
 
-func (b *Backend) getHeader(ctx context.Context, blockNumber *big.Int, tmBlock *coretypes.ResultBlock) *ethtypes.Header {
+// fallbackEthereumHeaderWithoutTendermintBlock builds a minimal header when the block cannot be loaded
+// (e.g. CurrentHeader when Block RPC fails). BaseFee is overwritten by CurrentHeader afterward.
+func (b *Backend) fallbackEthereumHeaderWithoutTendermintBlock(height int64) *ethtypes.Header {
```
masih: `fallbackToEthHeaderOnly` or something less wordy?

evmrpc/simulate.go
masih: Yeah; sad face. We need to talk about "context" at some point.

masih (approved):
🍻

PR #3275, opened 2026-04-20
Utility for migrating from memiavl to flatKV

cody-littley:

This PR introduces a utility for migrating data from database A to database B. It's generic so that we can encapsulate migration logic. Additionally, we can reuse this utility in the future if we ever have a future migration of similar nature.

Lots of unit tests.


sei-db/state_db/sc/migration/migration_manager.go
```diff
@@ -0,0 +1,426 @@
...
+	for !oldDone || !newDone {
+		select {
+		case e := <-oldDBErr:
+			oldErr = e
+			oldDone = true
+		case e := <-newDBErr:
+			newErr = e
+			newDone = true
+		case <-ctx.Done():
+			return ctx.Err()
```
yzang2019, line 378: Do we need to drain both channels before returning here?

sei-db/state_db/sc/migration/memiavl_migration_iterator.go
```diff
@@ -0,0 +1,169 @@
...
+// computeStartTreeIndex returns the index of the first tree that may contain
+// unmigrated keys according to the given boundary.
+func computeStartTreeIndex(treeNames []string, boundary MigrationBoundary) int {
+	switch boundary.Status() {
+	case MigrationNotStarted:
+		return 0
+	case MigrationComplete:
+		return len(treeNames)
+	}
+	return sort.Search(len(treeNames), func(i int) bool {
```
yzang2019, line 156: I don't think tree name is guaranteed to be sorted?

sei-db/state_db/sc/migration/migration_manager.go
```diff
@@ -0,0 +1,426 @@
...
+	// Pairs destined for each DB, grouped by store name and keyed by KVPair.Key.
+	oldDBPairsByStore := make(map[string]map[string]*proto.KVPair)
+	newDBPairsByStore := make(map[string]map[string]*proto.KVPair)
+
+	// Create change sets that move the values to migrate from the old DB to the new DB.
+	for _, value := range valuesToMigrate {
+		// Write the value to the new DB.
+		putPair(newDBPairsByStore, value.ModuleName, &proto.KVPair{Key: value.Key, Value: value.Value})
+		// Delete the value from the old DB.
+		putPair(oldDBPairsByStore, value.ModuleName, &proto.KVPair{Key: value.Key, Delete: true})
```
yzang2019, line 308: Would recommend add a metrics how how many keys are migrated successfully

yzang2019 (approved):

sei-db/state_db/sc/migration/memiavl_migration_iterator.go
cody-littley: They are [currently sorted](https://github.com/sei-protocol/sei-chain/blob/63f25b14a7b9011d2a263284a39dc2ce125a5aa7/sei-db/state_db/sc/memiavl/multitree.go#L270). But you make a good point, it's fragile for us to assume they will always be sorted. I added a defensive sort just in case.

sei-db/state_db/sc/migration/migration_manager.go
cody-littley: If the context gets cancelled, this is basically a rugpull, and my goal is to tear down everything to avoid deadlocked threads. IMO, we should treat an unexpected context cancel equivalent to a "we're crashing now, free up everything in preparation for that".

The most useful time for this sort of teardown behavior is in unit tests. If we always use `t.Context()`, then we ensure that we don't leave behind deadlocked threads when a test ends. The goal is to make it so that each thread has a fast teardown mechanism for when we want to tear it down.

sei-db/state_db/sc/migration/migration_manager.go
cody-littley: Metrics added. The count metric will get reset after a restart or a state sync, but that should be ok as long as we are aware that it works that way.

sei-db/ledger_db/block/blocksim/bin/blocksim
blindchaser, line 1: do we want to commit the binaries here?

sei-db/ledger_db/block/blocksim/bin/configure-logger
blindchaser, line 1: same here

sei-db/state_db/sc/migration/mock_migration_iterator.go
```diff
@@ -0,0 +1,134 @@
...
+	boundary    MigrationBoundary
+}
+
+var _ MigrationIterator = (*MapMigrationIterator)(nil)
+
+// NewMapMigrationIterator creates a MapMigrationIterator from the given data,
+// positioned at the start (boundary defaults to MigrationBoundaryNotStarted).
+// If autoRebuild is true, the iterator re-reads from Data before every
+// NextBatch call, so external mutations are picked up automatically.
+func NewMapMigrationIterator(data map[string]map[string][]byte, autoRebuild bool) *MapMigrationIterator {
```
blindchaser: looks this only being used in the testing files? how about rename it to mock_migration_iterator_test.go

sei-db/state_db/sc/migration/migration_metrics.go
```diff
@@ -0,0 +1,222 @@
...
+//
+// Cardinality rationale: at a 10-minute interval a month-long migration
+// tops out at ~4k unique boundary_hex label values — well within
+// Prometheus' comfort zone — and the OTel exporter's staleness markers
+// keep only the most recent label active in the scrape set.
+func (m *MigrationMetrics) startBoundarySnapshotLoop(interval time.Duration) {
+	if m == nil || m.boundarySnapshot == nil {
+		return
+	}
+	go func() {
```
blindchaser, line 200: the goroutine is fire-and-forget. do we want to add some done chan struct{} or sync.WaitGroup to expose()a Close() method or something?

sei-db/state_db/sc/migration/migration_manager.go
```diff
@@ -0,0 +1,454 @@
...
+		sort.Slice(pairs, func(i, j int) bool {
+			return string(pairs[i].Key) < string(pairs[j].Key)
+		})
+		changeSets = append(changeSets, &proto.NamedChangeSet{
+			Name:      name,
+			Changeset: proto.ChangeSet{Pairs: pairs},
+		})
+	}
+	return changeSets
+}
```
blindchaser, line 470: during block processing, could queries arrive while ApplyChangeSets runs? do we need a RWMutex here

sei-db/state_db/sc/migration/migration_boundary.go
```diff
@@ -0,0 +1,171 @@
...
+func (mb *MigrationBoundary) Serialize() []byte {
+	switch mb.status {
+	case MigrationNotStarted, MigrationComplete:
+		return []byte{byte(mb.status)}
+	case MigrationInProgress:
+		nameBytes := []byte(mb.moduleName)
+		buf := make([]byte, 1+4+len(nameBytes)+len(mb.key))
+		buf[0] = byte(mb.status)
+		binary.BigEndian.PutUint32(buf[1:5], uint32(len(nameBytes))) //nolint:gosec // module names are short strings
+		copy(buf[5:5+len(nameBytes)], nameBytes)
```
blindchaser: the number 5 (1-byte status + 4-byte length prefix) appears ~8 times across Serialize and DeserializeMigrationBoundary. consider add single constant (like const headerSize = 5)

sei-db/ledger_db/block/blocksim/bin/blocksim
cody-littley: No, we don't! Removed.

This happened when I switched branches. It's was in the `.gitignore` in that branch, but not this one.

sei-db/ledger_db/block/blocksim/bin/configure-logger
cody-littley: removed

sei-db/state_db/sc/migration/mock_migration_iterator.go
cody-littley: renamed

sei-db/state_db/sc/migration/migration_metrics.go
cody-littley: The goroutine dies when the context is cancelled. But it's not tricky to add an explicit Close() method. Added.

sei-db/state_db/sc/migration/migration_boundary.go
cody-littley: done

sei-db/state_db/sc/migration/migration_manager.go
cody-littley: Requires discussion.

Although it is currently thread safe to read values concurrently with a write, we don't have API for atomic batch reads. Even without this code, a batch read might return an inconsistant slice of data (i.e. some data from block N, other data from block N+1).

cody-littley: Added mutex to make this safe.

blindchaser (approved):

PR #3277, opened 2026-04-20
Export ProposerPriorityHash metric for divergence monitoring

wen-coding:
## Summary
- Add two Prometheus gauges exported every 1024 heights to let operators detect ProposerPriority divergence between validators
- `tendermint_state_proposer_priority_hash`: first 6 bytes of the hash packed into a float64 (48 bits fits in the float64 mantissa; constant cardinality of one series per node)
- `tendermint_state_proposer_priority_hash_height`: paired height so operators correlate samples taken at the same checkpoint
- Log line at each checkpoint carries the full 32-byte hex hash plus the packed value for grep-based debugging

## Test plan
- [x] `go vet ./sei-tendermint/internal/state/...` passes
- [x] `go test ./sei-tendermint/internal/state` passes
- [x] New test `TestApplyBlockProposerPriorityHash` applies 1024 real blocks and verifies exactly one emission at the interval boundary, with the correct packed hash and height values

### DIFF
--- a/sei-tendermint/internal/state/execution.go
+++ b/sei-tendermint/internal/state/execution.go
@@ -3,6 +3,7 @@ package state
 import (
 	"bytes"
 	"context"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"time"
@@ -23,7 +24,6 @@ var logger = seilog.NewLogger("tendermint", "internal", "state")
 // proposerPriorityHashInterval is how often (in heights) the
 // ProposerPriorityHash metric is exported. Used so operators can compare
 // hashes across validators and detect ProposerPriority divergence.
-// Must be a power of 2 so the check uses bitwise AND instead of modulo.
 const proposerPriorityHashInterval = 1024
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


philipsu522 (approved):

sei-tendermint/internal/state/execution.go
```diff
@@ -303,6 +309,56 @@ func (blockExec *BlockExecutor) ApplyBlock(ctx context.Context, state State, blo
...
+	// min of stale/zero data at Sei's block times. Acceptable for a
+	// monitoring signal that is only checked in response to incidents.
+	//
+	// proposerPriorityHashInterval is a power of 2 so we use bitwise AND
+	// instead of modulo.
+	if block.Height&(proposerPriorityHashInterval-1) == 0 {
+		if full := state.Validators.ProposerPriorityHash(); len(full) >= 8 {
+			// Pack the first 6 bytes big-endian into a uint64, then cast to
+			// float64 — safe because 2^48 < 2^53 (float64 mantissa).
+			var packed uint64
```
pompon0: binary.LittleEndian.Uint64

sei-tendermint/internal/state/execution.go
```diff
@@ -303,6 +309,56 @@ func (blockExec *BlockExecutor) ApplyBlock(ctx context.Context, state State, blo
...
+	//
+	// Note on restart staleness: Prometheus Gauges live in memory. After a
+	// process restart the gauges reset to zero until the next emission at
+	// the following multiple of proposerPriorityHashInterval — up to ~8.5
+	// min of stale/zero data at Sei's block times. Acceptable for a
+	// monitoring signal that is only checked in response to incidents.
+	//
+	// proposerPriorityHashInterval is a power of 2 so we use bitwise AND
+	// instead of modulo.
+	if block.Height&(proposerPriorityHashInterval-1) == 0 {
```
pompon0: how about just blockHeight%proposerPriorityHashInter? This seems like a fragile premature optimization.

pompon0 (approved):

sei-tendermint/internal/state/execution.go
wen-coding: done

sei-tendermint/internal/state/execution.go
wen-coding: Sure, was trying to make it a tiny bit faster, that's fair, changed.

PR #3281, opened 2026-04-21
fixes to mempool handling

pompon0:
* made handlePeerUpdates in mempool reactor run even if Broadcast = false, so that CheckTx failures are accounted properly.
* make autobahn producer wait on a AtomicRecv instead of TxsAvailable channel so that block production is not blocked by execution.


sei-will (approved):

sei-tendermint/internal/mempool/reactor/reactor.go
```diff
@@ -201,9 +201,6 @@ func (r *Reactor) processMempoolCh(ctx context.Context) error {
 // handle PeerUpdate messages. When the reactor is stopped, we will catch the
 // signal and close the p2p PeerUpdatesCh gracefully.
 func (r *Reactor) processPeerUpdates(ctx context.Context) error {
-	if !r.cfg.Broadcast {
-		return nil
-	}
```
wen-coding, line 206: Do we have test coverage for the new behavior?

sei-tendermint/internal/autobahn/producer/state.go
```diff
@@ -49,16 +49,16 @@ func NewState(cfg *Config, txMempool *mempool.TxMempool, consensus *consensus.St
...
-		select {
-		case <-ctx.Done():
-		case <-s.txMempool.TxsAvailable():
-		}
+// It waits for any transactions OR until `cfg.BlockInterval` passes.
+func (s *State) makePayload(ctx context.Context) (*types.Payload, error) {
+	// Wait for transactions. We give up and produce an empty block if mempool is empty for
+	// cfg.BlockInterval.
+	_ = utils.WithTimeout(ctx, s.cfg.BlockInterval, func(ctx context.Context) error {
+		return s.txMempool.TxStore().WaitForTxs(ctx)
```
wen-coding, line 57: just curious: it should not matter if at this exact moment txs are arriving right? We are doing memory read which should be really fast.

sei-tendermint/internal/mempool/reactor/reactor.go
```diff
@@ -222,10 +219,14 @@ func (r *Reactor) processPeerUpdates(ctx context.Context) error {
 				pctx, pcancel := context.WithCancel(ctx)
 				peerRoutines[update.NodeID] = pcancel
 				r.ids.ReserveForPeer(update.NodeID)
-				s.Spawn(func() error {
-					r.broadcastTxRoutine(pctx, update.NodeID)
-					return nil
-				})
+				// We keep peer management even then broadcasting is disabled,
```
wen-coding, line 222: nit: do you mean even when?

sei-tendermint/internal/mempool/reactor/reactor.go
```diff
@@ -222,10 +219,14 @@ func (r *Reactor) processPeerUpdates(ctx context.Context) error {
...
+				// so that failedCheckTxCounts WAI.
+				if r.cfg.Broadcast {
```
wen-coding, line 224: nit: since we previously broke error accounting, do we have test coverage for this?

sei-tendermint/internal/mempool/reactor/reactor.go
wen-coding, line 127: nit: do we need to add any counter at least? Just in case some validator goes Haywire and send us the same transaction tons of times and waste all bandwidth for nothing.

sei-tendermint/internal/mempool/mempool.go
wen-coding, line 442: So what happens if some spammer keeps sending us an invalid tx which will fail at app `CheckTx`? (let's say a bad signature.)

The tx will be added to the cache, fail `CheckTx`, and then removed from the cache here? Should we also add a separate counter for certain app-level failures which should not happen to prevent against spamming?

sei-tendermint/internal/autobahn/producer/state.go
pompon0: you mean some edge case between timeout and transactions arriving?

sei-tendermint/internal/mempool/reactor/reactor.go
pompon0: There are more serious holes in the tendermint p2p protocol than that. I don't thing we should spend much time on fixing them. If you feel the need to add more metrics for better visibility feel free to do so.

sei-tendermint/internal/mempool/mempool.go
pompon0: I would really love to avoid dealing with this code and just jump to rewriting it for autobahn. Your concern is valid though, of course.

sei-tendermint/internal/mempool/reactor/reactor.go
pompon0: added a test

sei-tendermint/internal/mempool/reactor/reactor.go
pompon0: added a test

sei-tendermint/internal/autobahn/producer/state.go
wen-coding: We might pack fewer transactions than keep grabbing in a loop, but I don't think we care that much.

wen-coding (approved):

PR #3290, opened 2026-04-21
hardcoding stateless leader election

pompon0:
After 6.5 upgrade everyone will be using stateless leader election.

### DIFF
--- a/sei-tendermint/config/config.go
+++ b/sei-tendermint/config/config.go
@@ -1117,6 +1117,8 @@ type ConsensusConfig struct {
 
 	DoubleSignCheckHeight int64 `mapstructure:"double-sign-check-height"`
 
+	StatelessLeaderElection bool `mapstructure:"stateless-leader-election"`
+
 	// TODO: The following fields are all temporary overrides that should exist only
 	// for the duration of the v0.36 release. The below fields should be completely
 	// removed in the v0.37 release of Tendermint.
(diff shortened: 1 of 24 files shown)


sei-tendermint/config/config.go
```diff
@@ -1117,10 +1117,9 @@ type ConsensusConfig struct {
 
 	DoubleSignCheckHeight int64 `mapstructure:"double-sign-check-height"`
 
-	// Whether the new stateless leader election should be used.
-	// THIS IS A TEMPORARY DISASTER RECOVERY MECHANISM IN CASE OF A CHAIN STALL.
-	// REQUIRES COORDINATION OF MAJORITY OF VALIDATORS TO SET TO TRUE.
-	// IF YOU SET IT TO TRUE JUST FOR YOUR NODE, IT WILL BE UNABLE TO PARTICIPATE IN THE CONSENSUS.
+	// Deprecated: stateless leader election is always enabled when constructing
+	// the consensus RoundState. This field is retained only for config parsing
+	// compatibility and is ignored.
```
wen-coding, line 1122: hmm, I think what you are doing is fine after a bit, but maybe we can start with a tiny PR which just sets StatelessLeaderElection to true in config and deprecate later?

pompon0: ok

PR #3291, opened 2026-04-21
StatelessLeaderElection defaults to true

pompon0:
Switching the default to true, will disable stateful leader election completely in 6.6


wen-coding (approved):

pompon0:
technically not app-hash-breaking, but not compatible with 6.4

philipsu522 (approved):

PR #3294, opened 2026-04-21
Parquet disable full range scan

jewei1997:
If the pebbledb tx index for the new parquet-based receiptdb is disabled, we will currently fallback to range scans over all the parquet files. This is a problem because it is a very expensive operation. This PR disallows the full range scan if the tx index is disabled.

unit tests.

### DIFF
--- a/sei-db/ledger_db/receipt/receipt_store.go
+++ b/sei-db/ledger_db/receipt/receipt_store.go
@@ -33,8 +33,11 @@ var (
 	// ErrTxIndexDisabled indicates that a receipt-by-tx-hash lookup missed the
 	// in-memory cache and cannot be served because the parquet backend's pebble
 	// tx hash index is disabled. A full parquet scan would require reading every
-	// file on disk and is intentionally not attempted.
-	ErrTxIndexDisabled = errors.New("receipt tx hash index is disabled; parquet fallback scan is not allowed")
+	// file on disk and is intentionally not attempted. The error wraps
+	// ErrNotFound so callers that treat "not found" as a null/nil result (e.g.
+	// eth_getTransactionReceipt) continue to behave correctly; the wrapping
+	// preserves the underlying reason for operators and tests via errors.Is.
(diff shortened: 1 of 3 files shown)


sei-db/ledger_db/receipt/receipt_store.go
```diff
@@ -30,6 +30,11 @@ var (
 	ErrNotFound               = errors.New("receipt not found")
 	ErrNotConfigured          = errors.New("receipt store not configured")
 	ErrRangeQueryNotSupported = errors.New("range query not supported by this backend")
+	// ErrTxIndexDisabled indicates that a receipt-by-tx-hash lookup missed the
+	// in-memory cache and cannot be served because the parquet backend's pebble
+	// tx hash index is disabled. A full parquet scan would require reading every
+	// file on disk and is intentionally not attempted.
+	ErrTxIndexDisabled = errors.New("receipt tx hash index is disabled; parquet fallback scan is not allowed")
```
cody-littley: Will the end user get a "not found" error, or will they get this message? We should just be telling the user the data is not found, the fact that we aren't doing a scan is probably not helpful to them.

cody-littley (approved):

sei-db/ledger_db/receipt/receipt_store.go
jewei1997: Fixed

What the user saw before the UX fix (parquet backend + tx-index-backend = "", query for unknown hash):
```
  {"jsonrpc":"2.0","id":1,"error":{"code":-32000,
   "message":"receipt tx hash index is disabled; parquet fallback scan is not allowed"}}
```
Non-standard, leaks internal details, and breaks clients that expect the Ethereum null convention.

What the user sees after the fix:
```
  {"jsonrpc":"2.0","id":1,"result":null}
```
Standard Ethereum "unknown tx" behavior.

yzang2019 (approved):

PR #3295, opened 2026-04-22
Use journal instead of CacheKV for giga snapshot/rollback

codchen:
## Problem                                                                                                                              
giga/deps/xevm/state implemented Snapshot()/RevertToSnapshot() by stacking CacheMultiStore layers: each Snapshot() call wrapped the current context in a new cache, and RevertToSnapshot() restored an older context (dropping the outer cache) to undo KV-store writes. This meant N nested EVM snapshots within a single transaction produced N nested CacheMultiStore layers, with all the associated allocation and copy-on-write overhead. The transient/in-memory state (access lists, logs, refunds, transient storage, account status) was already journal-driven; only the persisted KV state relied on the CacheMultiStore trick.                                                                                                                                  
## Solution                                                  
Adopt the go-ethereum journal pattern for all state—both KV-stored and in-memory:                                                    
- One CacheMultiStore per stateDB, created at NewDBImpl. It isolates the transaction's writes from the underlying store and is flushed once at Finalize(). No additional layers are created per snapshot.
- Snapshot() = record journal length + revision ID. No context replacement, no cache allocation.                                     
- RevertToSnapshot() = binary-search for the revision, replay journal entries backward. Every KV mutation records the previous value in a new journal entry type; reverting calls the inverse keeper operation.                                                            
### Event-manager isolation                                                                                                              
Snapshot() pushes the current EventManager onto a stack and attaches a fresh one to the context. RevertToSnapshot() pops back to the snapshot's EM, discarding any events emitted inside the reverted range. Finalize() drains all surviving EMs (in order) into committedCtx.EventManager().                                                                                                         
### GetCommittedState
Previously read from snapshottedCtxs[0] (the context before the first stateDB-level CMS layer). Now reads from committedCtx, the original context saved at NewDBImpl time, which sits below the single stateDB CMS and therefore always reflects the pre-transaction committed state.                                                                                                                     
### Other changes
- EVMKeeper interface gains SetAddressMapping so deleteMappingChange can restore address associations on revert.                     
- clearAccountState (used by Finalize) is unchanged and un-journaled; a new clearAccountStateJournaled wrapper (called from CreateAccount) captures prev state before clearing.                                                                                  
- DeepCopy now uses maps.Copy instead of manual loops.

### DIFF
--- a/giga/deps/xevm/state/journal.go
+++ b/giga/deps/xevm/state/journal.go
@@ -4,12 +4,19 @@ import (
 	"encoding/binary"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/sei-protocol/sei-chain/giga/deps/xevm/types"
 	sdk "github.com/sei-protocol/sei-chain/sei-cosmos/types"
 )
 
 type journalEntry interface {
 	// revert undoes the changes introduced by this journal entry.
 	revert(*DBImpl)
+}
+
+// revision marks a snapshot point in the journal.
+type revision struct {
+	id           int
+	journalIndex int
 }
 
 type (
@@ -43,8 +50,46 @@ type (
 		delta sdk.Int
 	}
 
-	watermark struct {
-		version int
+	// storageChange records a KV storage mutation so it can be reverted.
+	storageChange struct {
+		addr common.Address
+		key  common.Hash
+		prev common.Hash
+	}
+
+	// codeChange records a code mutation so it can be reverted.
+	codeChange struct {
+		addr     common.Address
+		prevCode []byte
+	}
+
+	// nonceChange records a nonce mutation so it can be reverted.
+	nonceChange struct {
+		addr common.Address
+		prev uint64
+	}
+
+	// balanceChange records an Add or Sub balance so it can be reverted.
+	balanceChange struct {
+		evmAddr common.Address
+		seiAddr sdk.AccAddress
+		usei    sdk.Int
+		wei     sdk.Int
+		isAdd   bool // true if AddBalance was called
+	}
+
+	// createAccountChange records the previous state cleared by clearAccountStateJournaled.
+	createAccountChange struct {
+		addr      common.Address
+		prevCode  []byte
+		prevNonce uint64
+		prevSlots map[common.Hash]common.Hash
+	}
+
+	// deleteMappingChange records a DeleteAddressMapping so it can be reverted.
+	deleteMappingChange struct {
+		evmAddr common.Address
+		seiAddr sdk.AccAddress
 	}
 )
 
@@ -112,8 +157,6 @@ func (e *transientStorageChange) revert(s *DBImpl) {
 	}
 }
 
-func (e *watermark) revert(s *DBImpl) {}
-
 func (e *accountStatusChange) revert(s *DBImpl) {
 	accts := s.tempState.transientAccounts
 	if e.prev == nil {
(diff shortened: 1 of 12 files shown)


codeinputmachine:
Hey @codchen, Code Input detected this PR has a merge conflict. Some of the conflicts of this PR can be resolved with a semantic merge driver. Code Input can do that automatically: https://codeinput.com/r/EESR7EMMZd3 Let me know if you need more help with this conflict or how Code Input works.

giga/deps/xevm/state/journal.go
```diff
@@ -122,3 +165,50 @@ func (e *accountStatusChange) revert(s *DBImpl) {
...
+	s.k.SetNonce(s.ctx, e.addr, e.prev)
+}
+
+func (e *balanceChange) revert(s *DBImpl) {
+	// Suppress events on revert
+	ctx := s.ctx.WithEventManager(sdk.NewEventManager())
+	denom := s.k.GetBaseDenom(s.ctx)
+	if e.isAdd {
+		// Was AddBalance: reverse by subtracting
+		_ = s.k.BankKeeper().SubUnlockedCoins(ctx, e.seiAddr, sdk.NewCoins(sdk.NewCoin(denom, e.usei)), true)
```
arajasek: do we want to be propagating the errors here?

arajasek (approved):

giga/deps/xevm/state/journal.go
codchen: i think it isn't expected to err here. Maybe we should just panic out?

arajasek: I think panicking might be the most reasonable thing, honestly.

giga/deps/xevm/state/statedb.go
```diff
@@ -126,39 +144,20 @@ func (s *DBImpl) Finalize() (surplus sdk.Int, err error) {
...
 
-	s.flushCtxs()
-	// write all events in order
-	for i := 1; i < len(s.snapshottedCtxs); i++ {
-		s.flushEvents(s.snapshottedCtxs[i])
+	// Write the single CMS layer to the underlying store.
+	s.ctx.MultiStore().(sdk.CacheMultiStore).Write()
+	s.ctx.GigaMultiStore().WriteGiga()
+
+	// Emit all surviving events (from snapshots + current) to the committed ctx's EventManager.
```
philipsu522, line 152: do we have tests before/after (even existing one) to ensure that the same events are emitted before & after?

giga/deps/xevm/state/journal.go
philipsu522: ➕

giga/deps/xevm/state/statedb.go
```diff
@@ -85,20 +100,23 @@ func (s *DBImpl) AddPreimage(_ common.Hash, _ []byte) {}
...
-		s.ctx = s.snapshottedCtxs[0]
-	}
+	// Reset back to the committed (pre-stateDB) state by discarding the CMS layer.
+	s.ctx = s.committedCtx
 	feeCollector, _ := s.k.GetFeeCollectorAddress(s.Ctx())
 	s.coinbaseEvmAddress = feeCollector
 	s.tempState = NewTemporaryState()
 	s.journal = []journalEntry{}
-	s.snapshottedCtxs = []sdk.Context{}
-	s.Snapshot()
```
philipsu522, line 101: we used to create a new Snapshot() after cleanup, do we want to enforce that invariant in the new journaled state?

if we don't want it, maybe we add a test enforcing that this isn't required after CleanupForTracer()

giga/deps/xevm/state/statedb.go
codchen: added one

giga/deps/xevm/state/journal.go
codchen: made it panic

giga/deps/xevm/state/statedb.go
codchen: added

philipsu522 (approved):

PR #3300, opened 2026-04-22
Support autobahn node restart by skipping CometBFT handshaker (CON-252)

wen-coding:
## Summary

Enables an autobahn validator to be stopped and restarted. The CometBFT handshaker used to abort because autobahn never writes to the CometBFT block/state stores, so it observed `storeHeight=0 < appHeight=N` → `ErrAppBlockHeightTooHigh`.

- `node.go`: `shouldHandshake = !stateSync && !gigaEnabled` — skip the handshaker in giga mode.
- `giga_router.go` (`runExecute`): now owns `InitChain`. Calls it on fresh start (`last == 0`); skips it on restart (`last > 0`), where `BaseApp.FinalizeBlock`'s existing nil-check rebuilds `deliverState` from the committed CMS.
- `giga_router_test.go`: drop the simulated-handshaker `InitChain` from `TestGigaRouter_FinalizeBlocks`; tighten the contract-test doc on `TestInitChainCommitThenFinalize`.
- `integration_test/autobahn/autobahn_test.go`: replace the `t.Skip` in the Recovery subtest with a real restart test. Adds `restartNode` helper.

## Startup paths

|  | shouldHandshake | InitChain by | FinalizeBlock's deliverState |
|---|---|---|---|
| A (fresh, no giga) | true | handshaker | InitChain |
| B (fresh, giga) | **false** | **runExecute** | InitChain |
| C (restart, no giga) | true | — (appHeight>0) | handshaker replay if needed, else existing |
| D (restart, giga) | **false** | — (last>0) | `BaseApp` nil-check fallback from committed CMS |

Cases B and D are what change. In B, `runExecute` takes over the `InitChain` duty the handshaker used to perform. In D, skipping the handshaker is the whole fix: the CometBFT state/block stores are never populated by autobahn, so reconciling against them is vacuous; the app's committed CMS is the source of truth, and `BaseApp.FinalizeBlock`'s existing `if deliverState == nil { setDeliverState(header) }` fallback picks it up on the first block post-restart.

## Scope / TODO

Covers **restart with local disk intact** (`app.Info().LastBlockHeight = N`, autobahn WAL has block ≥ N). Does **not** cover a new validator joining or a disk-wiped node — that needs app state sync + autobahn WAL sync. Tracked via `TODO(autobahn-recovery)` in `node.go`.

## Test plan

- [x] `go build ./...` clean; `gofmt -s -l .` clean
- [x] `go test ./internal/p2p/... ./node/... ./internal/consensus/... ./internal/autobahn/...` — green
- [x] `make autobahn-integration-test` — all 5 subtests pass. Recovery: halted @ 482 → resumed @ 537 in 12.4s after restarting a killed node.

### DIFF
--- a/integration_test/autobahn/autobahn_test.go
+++ b/integration_test/autobahn/autobahn_test.go
@@ -283,13 +283,55 @@ func TestAutobahn(t *testing.T) {
 	t.Run("BankTransfer", testBankTransfer)
 	t.Run("LivenessUnderMaxFaults", testLivenessUnderMaxFaults)
 	t.Run("HaltsBeyondMaxFaults", testHaltsBeyondMaxFaults)
-	// TODO: Re-enable once autobahn supports node restart. Currently, a restarted seid
-	// fails because autobahn writes to the app state but not to the CometBFT block/state
-	// store. On restart, the CometBFT handshaker sees appHeight >> storeHeight and
-	// cannot reconcile.
-	t.Run("Recovery", func(t *testing.T) {
-		t.Skip("autobahn node restart not yet supported")
-	})
+	t.Run("Recovery", testRecovery)
+}
+
+// restartNode re-invokes the container's seid-start script inside sei-node-<i>.
+// The script backgrounds seid and exits, so `docker exec -d` is the right mode:
+// it returns immediately while seid keeps running.
+func restartNode(t *testing.T, i int) {
+	t.Helper()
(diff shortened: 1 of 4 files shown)


pompon0 (approved):
nice, I really like the fact that we depend on the tendermint flow less and less (here we get rid of the handshake)

arajasek (approved):

PR #3304, opened 2026-04-23
feat: wire autobahn state sync with naive provider (CON-252)

wen-coding:
> Stacks on top of #3300. Base will change to `main` once #3300 merges.

> **Status: draft — initial feedback round.** No integration test yet; see "Scope / Follow-ups" below.

## Summary

Extends #3300 to cover the **new-validator joining** and **disk-wiped node** recovery paths via CometBFT state sync. A joiner state-syncs the app to some height `M`, then `runExecute` resumes at `M+1` pulling subsequent blocks from giga peers.

## Changes

- **`node.go`** — stop force-disabling `stateSync` in giga mode. Gate on a direct `app.Info()` check since CometBFT's `state.LastBlockHeight` never advances under autobahn. `postSyncHook` is a no-op in giga mode (no block-sync reactor to hand control to). `ssReactor` wired only when `stateSync` is actually enabled.
- **`statesync/reactor.go`** — new optional `stateProviderFactory` param to `NewReactor`, used in place of the RPC/P2P selection when set. Non-giga callers pass `nil`; existing behaviour preserved.
- **`statesync/giga_stateprovider.go`** *(new)* — naive provider: empty `AppHash` (opts out of pre-verification), minimal `Commit`, `sm.State` from `GenesisDoc` + static committee.
- **`statesync/syncer.go`** — skip the post-restore `AppHash` comparison when `trustedAppHash` is empty. Vanilla providers always return non-empty, so their behaviour is unchanged.
- **`autobahn/data/state.go`** — expose public `SkipTo` wrapping the existing internal `skipTo`. Used post-state-sync to align data cursors so peer-streamed blocks from `M+1` onward insert correctly.
- **`p2p/giga_router.go`** — new state-sync branch in `runExecute` (`last > 0 && NextBlock() <= last`): `SkipTo(last+1)`, no `PushAppHash`. Fresh / plain-restart branches unchanged from #3300.

## New startup path (case E, added to #3300's A–D)

|  | `shouldHandshake` | `stateSync` | `InitChain` by | first `FinalizeBlock` `deliverState` |
|---|---|---|---|---|
| **E join/recovery, giga (no local WAL)** | **false** | **true** | — (after sync, `last>0` skips it) | CMS rebuilt from snapshot chunks |

## Trust model — naive for v1, AppQC-bundled for v2

This PR **trusts the snapshot producer optimistically**: `stateProvider.AppHash` returns empty, `syncer.verifyApp` skips the comparison, and a corrupt snapshot wedges the joiner until external restart-with-wipe. The vanilla `SyncAny` retry loop only helps against honest-but-unavailable peers, not malicious ones — a malicious peer can loop-trap a joiner.

**The planned v2 mechanism (tracked by `TODO(autobahn-snapshot-proof)` in `giga_stateprovider.go`):**

> Peers serving a giga snapshot MUST include an `AppQC@snapshot_height` in `snapshot.Metadata`. The giga state provider decodes the AppQC, verifies `appQC.Verify(committee)` (2f+1 committee signatures — cryptographically self-verifiable from a single peer), compares `AppQC.AppHash` to the snapshot's claimed `AppHash`, and returns that as the authoritative `trustedAppHash`. Any corrupt snapshot fails `stateProvider.AppHash` and triggers vanilla `SyncAny`'s `errRejectSnapshot` → next snapshot → repeat. No loop-trap, no node restart required.

**Why not in this PR:**
- Autobahn only retains the **latest** AppQC in memory (`avail.latestAppQC` is a single `utils.Option`, not a queue). Historical AppQCs aren't persisted.
- `AppQC@M` forms *after* the app commits block `M` (committee votes are asynchronous), so the snapshot taken at Commit(M) doesn't yet have its anchor.
- v2 needs either (a) a "seal snapshot on AppQC formation" hook between the snapshot manager and avail state, or (b) coordinated delayed snapshot creation. Both are real scope, landing in a follow-up.

Until v2 lands: operators should only point joiners at known-honest peers; a bad snapshot is an ops issue, not a safety issue (the cluster is unaffected, only the joiner is stuck).

## Test plan

- [x] `go build ./...` clean; `gofmt -s -l .` clean
- [x] `go test ./internal/statesync/... ./internal/p2p/... ./internal/autobahn/... ./node/... ./internal/consensus/...` — all green
- [x] New unit tests: 5 × `TestGigaStateProvider_*`, 2 × `TestState_SkipTo_*`, 2 × `TestSyncer_verifyApp_EmptyTrustedAppHashSkipsCheck`

**Not yet:**
- [ ] `JoinFromStateSync` integration subtest — wipe a node's data dir, `statesync.enable = true`, restart, verify catchup. `snapshot-interval=100` already configured in the docker cluster. Deferred to land together with v2 (snapshot-proof), since integration-testing the naive-trust path without proper verification is of limited value.

## Scope / Follow-ups

- **v2 snapshot proof** (`TODO(autobahn-snapshot-proof)`) — bundle `AppQC` in `snapshot.Metadata`, verify cryptographically, enable in-loop retry.
- **Integration test** (`JoinFromStateSync`) — lands with v2.
- **`TODO(epoch)`** in `giga_stateprovider.go` — committee / validator set derivation moves to epoch lookup once autobahn supports dynamic committees.

### DIFF
--- a/sei-tendermint/internal/autobahn/data/state.go
+++ b/sei-tendermint/internal/autobahn/data/state.go
@@ -434,6 +434,21 @@ func (s *State) NextBlock() types.GlobalBlockNumber {
 	panic("unreachable")
 }
 
+// SkipTo advances all internal cursors to n, discarding everything before it.
+// It is safe to call only on a State that has no data past n (e.g. a fresh
+// State just after NewState, before any Push*). The caller is responsible
+// for ensuring no concurrent Push* races with this call — giga state sync
+// uses it between data.NewState and GigaRouter.Run.
+func (s *State) SkipTo(n types.GlobalBlockNumber) {
+	for inner, ctrl := range s.inner.Lock() {
+		if n <= inner.first {
+			return
+		}
+		inner.skipTo(n)
+		ctrl.Updated()
+	}
+}
+
 // Block returns the block with the given global number.
 // This function is used for syncing - GlobalBlock can be derived from Block and FullCommitQC,
 // which have to be fetched upfront anyway.
--- a/sei-tendermint/internal/autobahn/data/skipto_test.go
+++ b/sei-tendermint/internal/autobahn/data/skipto_test.go
@@ -1,0 +1,56 @@
+package data
+
+import (
+	"testing"
+
+	"github.com/sei-protocol/sei-chain/sei-tendermint/internal/autobahn/types"
+	"github.com/sei-protocol/sei-chain/sei-tendermint/libs/utils"
+	"github.com/sei-protocol/sei-chain/sei-tendermint/libs/utils/require"
+)
+
+// TestState_SkipTo_FreshAdvance: on a freshly constructed State, SkipTo(n)
+// should advance all cursors to n. This is the giga state-sync use case:
+// after the app is at height M via snapshot, call SkipTo(M+1) on the local
+// fresh data state so subsequent peer-streamed blocks insert starting at
+// M+1 instead of genesis.
+func TestState_SkipTo_FreshAdvance(t *testing.T) {
+	rng := utils.TestRng()
+	committee, _ := types.GenCommittee(rng, 3)
+	state := utils.OrPanic1(NewState(&Config{Committee: committee},
+		utils.OrPanic1(NewDataWAL(utils.None[string](), committee))))
+
+	// Sanity: fresh state starts at committee.FirstBlock().
+	require.Equal(t, committee.FirstBlock(), state.NextBlock())
+
+	target := committee.FirstBlock() + 1000
+	state.SkipTo(target)
+
+	// Every cursor visible through the public API should now reflect
+	// target. We only check NextBlock() (public); the other cursors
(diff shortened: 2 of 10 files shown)


sei-tendermint/internal/autobahn/data/state.go
```diff
@@ -434,6 +434,21 @@ func (s *State) NextBlock() types.GlobalBlockNumber {
 	panic("unreachable")
 }
 
+// SkipTo advances all internal cursors to n, discarding everything before it.
+// It is safe to call only on a State that has no data past n (e.g. a fresh
+// State just after NewState, before any Push*). The caller is responsible
+// for ensuring no concurrent Push* races with this call — giga state sync
+// uses it between data.NewState and GigaRouter.Run.
+func (s *State) SkipTo(n types.GlobalBlockNumber) {
```
pompon0, line 442: IMO this method should do a proper pruning instead, instead of imposing on the caller that State is empty.

wen-coding:
We plan to implement state sync outside the protocol for now.

PR #3306, opened 2026-04-23
create tooling for routing based on module name

cody-littley:

Create a utility that can route traffic based on module name. We need this, since we will be doing incremental migrations, where some modules stay in memiavl and others migrate to flatKV.

### During evm migration

image

### After evm migration

image

unit tests


sei-db/state_db/sc/migration/migration_types.go
```diff
@@ -32,7 +33,19 @@ func (s MigrationStatus) String() string {
...
-type DBWriter func(changesets []*proto.NamedChangeSet) error
+//
+// The ctx is forwarded unchanged so a DBWriter may be another Router's
+// ApplyChangeSets, allowing routers to be composed without adapter glue.
+type DBWriter func(ctx context.Context, changesets []*proto.NamedChangeSet) error
 
 // Read a value from the database.
 type DBReader func(store string, key []byte) ([]byte, bool, error)
+
+// An object capable of routing/splitting reands and writes between databases.
```
blindchaser: typo: reads

sei-db/state_db/sc/migration/migration_types.go
```diff
@@ -32,7 +33,19 @@ func (s MigrationStatus) String() string {
 }
 
 // Write a batch of values to the database. Assumed to be atomic.
```
blindchaser: it looks ModuleRouter.ApplyChangeSets is not atomic. can we double check the atomicity in the doc

sei-db/state_db/sc/migration/module_router.go
```diff
@@ -0,0 +1,180 @@
...
+	aDone, bDone := false, false
+	for !aDone || !bDone {
+		select {
+		case e := <-aErrCh:
+			aErr = e
+			aDone = true
+		case e := <-bErrCh:
+			bErr = e
+			bDone = true
+		case <-ctx.Done():
```
blindchaser: once writer goroutines are launched, returning ctx.Err() does not mean the writes stopped. A non-cooperative or slow writer can still mutate the DB after the caller observes failure and may retry, creating concurrent writes or partial state changes.

sei-db/state_db/sc/migration/module_router.go
```diff
@@ -0,0 +1,180 @@
...
+type routeTarget int
+
+const (
+	routeUnknown routeTarget = iota
+	routeA
+	routeB
+)
+
+func (m *ModuleRouter) routeFor(store string) routeTarget {
+	if _, ok := m.aModules[store]; ok {
```
blindchaser: consider clone aModules and bModules before assign value in the construction to avoid data race in routeFor

sei-db/state_db/sc/migration/migration_types.go
cody-littley: fixed

sei-db/state_db/sc/migration/migration_types.go
cody-littley: Good catch. Updated godoc to point this out.

```
// May not be atomic. If not atomic, then the caller must provide crash safe atomicity.
```

sei-db/state_db/sc/migration/module_router.go
cody-littley: This is now refactored in a way that makes this sort of thing impossible. The new `Route` struct can't be modified by anything outside.

sei-db/state_db/sc/migration/module_router.go
cody-littley: Removed early return on context cancellation. Also, added the following to the godoc:

```
	// If this method returns an error, it is not safe to attempt to retry. An error should be considered
	// fatal, and should result in any managed databases being shut down and crash recovered.
```

Kbhat1 (approved):

sei-db/state_db/sc/migration/migration_types.go
```diff
@@ -31,8 +32,24 @@ func (s MigrationStatus) String() string {
...
 type DBReader func(store string, key []byte) ([]byte, bool, error)
+
+// An object capable of routing/splitting reads and writes between databases.
+type Router interface {
+	// Read a value from the database.
+	Read(store string, key []byte) ([]byte, bool, error)
+
+	// Apply a batch of change sets to the database.
+	//
+	// Not atomic. Caller is respsible for providing crash safe atomicity.
```
blindchaser: typo: responsible

sei-db/state_db/sc/migration/module_router.go
```diff
@@ -0,0 +1,162 @@
...
+
+// NewModuleRouter creates a new ModuleRouter from one or more Routes.
+//
+// At least one Route must be provided. No module name may appear in more
+// than one Route. Data targeting a module that is not registered with any
+// Route returns an error.
+//
+// This is intentionally fragile to misconfiguration: it is important to
+// very specifically specify which modules belong to which database.
+func NewModuleRouter(routes ...*Route) (*ModuleRouter, error) {
```
blindchaser, line 85: when a caller invokes NewModuleRouter(slice...), Go shares the underlying slice — mutating slice[i] afterwards changes m.routes[i], which ApplyChangeSets iterates without locks

consider add defensive slice copy for routes

blindchaser (approved):

sei-db/state_db/sc/migration/migration_types.go
cody-littley: fixed

sei-db/state_db/sc/migration/module_router.go
cody-littley: added defensive copy

PR #3309, opened 2026-04-24
rpc: populate /status under Autobahn

wen-coding:
## Summary

- `/status` now reports real `SyncInfo.LatestBlockHeight`, `LatestAppHash`, and `ValidatorInfo` under Autobahn instead of zero/panicking. Integration tests and tooling that poll `seid status | .SyncInfo.latest_block_height` (7+ files in sei-chain CI) start working.
- New `SyncInfo.LastCommittedBlockHeight` field exposes the highest consensus-finalized global block via a cached, lock-free CommitQC watch on `GigaRouter`. Populated only under Autobahn — the field is `,omitempty` and absent from CometBFT `/status` responses, so the wire format stays byte-identical for existing CometBFT clients.
- Gated via a new `env.GigaEnabled` flag derived from `cfg.AutobahnConfigFile != ""` — zero behavior change under CometBFT.

## Design notes

- `LatestBlockHeight`/`LatestAppHash` are sourced from `ABCIInfo` (same int64/[]byte the app records in `FinalizeBlock`). Under Autobahn, `BlockStore.Height()` stays 0, so the existing CometBFT path produces nothing. `ABCIInfo` is fetched once per `/status` and reused for `ApplicationInfo.Version` to avoid a double ABCI round-trip.
- `LastCommittedBlockHeight` reads `CommitQC.GlobalRange(committee).Next - 1`. `lastCommitQCRecv` is subscribed once in `NewGigaRouter`; per-RPC it's a bare `atomic.Pointer.Load`. The staleness reasoning is documented on the struct field. Under CometBFT the value is left at zero and elided by `,omitempty`; clients that want it should fall back to `LatestBlockHeight` (commit == app-apply in one step there).
- `validatorAtHeight` skips the `ConsensusState.GetValidators()` fast path under `GigaEnabled` (would nil-deref on an unpopulated CometBFT validator set) and uses the state-store lookup, which works under both engines.

## Left to follow-up (documented in `TODO(autobahn):` on `Status()`)

- Other `SyncInfo` fields: `LatestBlockHash`, `LatestBlockTime`, `Earliest*`, `CatchingUp`, `MaxPeerBlockHeight`. Each needs a new Autobahn accessor (block metadata / pruning boundary / sync-state signal). No integration-test reader today; external tooling will want them.
- `ValidatorInfo.VotingPower`: `validatorAtHeight` is called with `latestUncommittedHeight()`, which evaluates to `BlockStore.Height()+1 == 1` under Autobahn, so `StateStore.LoadValidators(1)` always returns the genesis set. Reporting correct voting power once the validator set changes requires passing the current Autobahn height instead.
- `env.BlockStore.Height() == 0` under Autobahn still breaks `/block`, `/block_results`, `/commit`, and the evmrpc endpoints that walk through them (`eth_getBlockByNumber`, `eth_gasPrice`, etc.). That's a broader follow-up — a decision on where block metadata lives under Autobahn.

## Test plan

- [x] `go test ./sei-tendermint/internal/{rpc/core,consensus,p2p,p2p/giga,node}/... -count=1` — all pass.
- [x] `TestGigaRouter_FinalizeBlocks` extended to assert `Router.Giga()` + `GigaRouter.LastCommittedBlockNumber()` return the expected non-zero values after real Autobahn consensus drives blocks end-to-end.
- [x] Live-cluster spike on a 4-node Autobahn docker cluster:
  - `curl /status` returns advancing `latest_block_height` / `last_committed_block_height` (delta=0 steady state, non-zero on load).
  - `seid status | .SyncInfo.latest_block_height` (integration-test query) returns real number.
  - `seid status | .NodeInfo.network` returns chain id.
  - EVM state reads (`eth_chainId`, `eth_blockNumber`, `eth_getBalance`, `eth_getTransactionCount`) work. End-to-end `seid tx evm send` → balance poll works.
  - EVM block-metadata reads still fail with `blockchain height: 0` — expected, covered by follow-up.
- [x] CometBFT path unchanged: all existing test assertions preserved; `/status` JSON response byte-identical to pre-PR (the new `last_committed_block_height` field is elided via `,omitempty`).

### DIFF
--- a/sei-tendermint/internal/p2p/giga_router.go
+++ b/sei-tendermint/internal/p2p/giga_router.go
@@ -50,6 +50,18 @@ type GigaRouter struct {
 	service   *giga.Service
 	poolIn    *giga.Pool[NodePublicKey, rpc.Server[giga.API]]
 	poolOut   *giga.Pool[NodePublicKey, rpc.Client[giga.API]]
+
+	// lastCommitQCRecv is subscribed once at construction and reused for the
+	// lifetime of the GigaRouter. Load() is lock-free (a single
+	// atomic.Pointer.Load).
+	//
+	// Staleness-safety: the receiver points at the same atomicWatch held inside
+	// avail.inner.latestCommitQC — a value field on a heap-allocated *inner
+	// that is never replaced for the lifetime of the State, only Store()d
+	// into. Every Load therefore observes the most recent Store. A
+	// reconstructed avail.State (only on process restart) would also
+	// reconstruct this GigaRouter, so the receiver can't outlive its watch.
+	lastCommitQCRecv utils.AtomicRecv[utils.Option[*atypes.CommitQC]]
 }
 
 func NewGigaRouter(cfg *GigaRouterConfig, key NodeSecretKey) (*GigaRouter, error) {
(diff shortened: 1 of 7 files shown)


codchen (approved):

arajasek (approved):

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -88,9 +100,31 @@ func NewGigaRouter(cfg *GigaRouterConfig, key NodeSecretKey) (*GigaRouter, error
...
+// has been recorded yet. Safe for high-frequency callers — uses a cached
+// lock-free receiver; no locks taken on this path.
+func (r *GigaRouter) LastCommittedBlockNumber() int64 {
+	qc, ok := r.lastCommitQCRecv.Load().Get()
+	if !ok {
+		return 0
+	}
+	// GlobalRange is a half-open [First, Next) interval; the highest
+	// committed block number is Next-1. If the range is empty, return 0.
+	gr := qc.GlobalRange(r.data.Committee())
```
pompon0: nit: autobahn/types.GlobalRangeOpt should do

pompon0 (approved):

sei-tendermint/internal/p2p/giga_router.go

PR #3310, opened 2026-04-24
rpc: route /block + /block_results + /block_by_hash + /validators through Autobahn under GigaEnabled (CON-257)

wen-coding:
## Summary

Adds Autobahn handlers for `/block`, `/block_results`, `/block_by_hash`, and `/validators`, gated on `env.gigaRouter()` returning Some.

## Verified (live 4-node Autobahn cluster)

| Endpoint | Before | After |
|---|---|---|
| `eth_chainId`, `net_version`, `eth_blockNumber` | ✓ | ✓ |
| `eth_gasPrice` | crash | ✓ |
| `eth_getBlockByNumber("latest")` | error / nulls | ✓ |
| `eth_getBlockByNumber("0xFUTURE")` | error | ✓ `result: null` (spec) |
| `eth_getBlockByHash(<latest>)` | `result: null` | ✓ full block |
| `eth_getBlockByHash(<unknown>)` | `result: null` | ✓ `result: null` |
| `eth_call` / `eth_estimateGas` / `eth_getLogs` / `eth_feeHistory` | crash / error | ✓ |
| `eth_getTransactionReceipt` / `eth_getTransactionByHash` | null fields | ✓ |
| `seid tx evm send` → receipt poll | — | ✓ end-to-end |
| `/validators` | stuck at `block_height: 1` | ✓ matches actual chain head |
| `/validators?per_page=2&page={1,2,3}` | — | ✓ proper paging + range error |
| `/validators?height={200,999999}` | wrong head reported | ✓ historical query / proper sentinel |

## Non-state-breaking

All changes in the RPC serving path. Block execution (`app.FinalizeBlock` / `app.Commit`) and consensus data persistence untouched. App hashes match across versions — safe to deploy asymmetrically.

## Out of scope (TODOs in code)

- **Lower-bound height check under Autobahn**: currently `env.BlockStore.Base()=0`; will source from `BlockDB.GetLowestBlockHeight` once BlockDB is wired.
- **`env.BlockResults` per-tx detail**: empty under Autobahn (FinalizeBlock responses not persisted). evmrpc doesn't read them; revisit if downstream needs them.
- **`env.Commit` / `env.Header` / `env.HeaderByHash` / `/tx` / `/tx_search`**: not on evmrpc's critical path; separate workstream.
- **`data.State.inner.blockHashes`**: temporary in-memory index; replaced by `sei-db/ledger_db/block.BlockDB` once a writer is wired.

### DIFF
--- a/sei-tendermint/internal/autobahn/data/state.go
+++ b/sei-tendermint/internal/autobahn/data/state.go
@@ -458,8 +458,13 @@ func (s *State) NextBlock() types.GlobalBlockNumber {
 // the hash index is maintained in lockstep by insertBlock / pruneFirst.
 //
 // Returns an error in the signature for forward-compat with the eventual
-// switch to sei-db/ledger_db/block.BlockDB.GetBlockByHash, which can fail
-// with real I/O errors. The current in-memory implementation never errors.
+// switch to sei-db/ledger_db/block.BlockDB.GetBlockByHash. Today's
+// in-memory implementation never errors.
+//
+// TODO(autobahn): when BlockDB is wired, take a ctx parameter and narrow
+// the error contract — db-internal errors should surface by shutting down
+// the persistence background task (matching how persistence handles errors
+// today), so the query path's error stays bounded to context.Canceled.
 func (s *State) GlobalBlockByHash(hash types.BlockHeaderHash) (utils.Option[*types.GlobalBlock], error) {
 	for inner := range s.inner.Lock() {
 		n, ok := inner.blockHashes[hash]
--- a/sei-tendermint/internal/autobahn/data/state.go
+++ b/sei-tendermint/internal/autobahn/data/state.go
@@ -149,6 +149,17 @@ type inner struct {
 	blocks       map[types.GlobalBlockNumber]*types.Block             // [first,nextBlock) + subset of [nextBlock,nextQC)
 	appProposals map[types.GlobalBlockNumber]appProposalWithTimestamp // [first,nextAppProposal)
 
+	// blockHashes is a hash → height index mirroring blocks. Maintained
+	// in lockstep with blocks via insertBlock / pruneFirst, so it covers
+	// exactly the same retain window without a separate prune cursor or
+	// startup warmup. Powers BlockByHash.
+	//
+	// TODO(autobahn): remove once a writer is wired into block execution
+	// that populates sei-db/ledger_db/block.BlockDB. BlockDB has a built-in
+	// hash index that survives process restart and lives outside Autobahn's
+	// RetainHeight pruning, making this in-memory index obsolete.
+	blockHashes map[types.BlockHeaderHash]types.GlobalBlockNumber
+
 	// first <= nextAppProposal <= nextBlockToPersist <= nextBlock <= nextQC
 	//
 	// This invariant guarantees no race between pruning and persisting:
@@ -168,6 +179,7 @@ func newInner(committee *types.Committee) *inner {
 		qcs:                map[types.GlobalBlockNumber]*types.FullCommitQC{},
 		blocks:             map[types.GlobalBlockNumber]*types.Block{},
 		appProposals:       map[types.GlobalBlockNumber]appProposalWithTimestamp{},
+		blockHashes:        map[types.BlockHeaderHash]types.GlobalBlockNumber{},
 		first:              first,
 		nextAppProposal:    first,
 		nextBlockToPersist: first,
@@ -226,10 +238,12 @@ func (i *inner) insertBlock(committee *types.Committee, n types.GlobalBlockNumbe
 	qc := i.qcs[n]
 	storedGR := qc.QC().GlobalRange(committee)
 	want := qc.Headers()[n-storedGR.First].Hash()
-	if got := block.Header().Hash(); want != got {
+	got := block.Header().Hash()
+	if want != got {
 		return fmt.Errorf("block %d header hash mismatch: want %v, got %v", n, want, got)
 	}
 	i.blocks[n] = block
+	i.blockHashes[got] = n
 	return nil
 }
 
@@ -255,6 +269,7 @@ func (i *inner) pruneFirst(now time.Time, m *dataMetrics) {
 	delete(i.appProposals, i.first)
 	delete(i.blocks, i.first)
 	delete(i.qcs, i.first)
+	delete(i.blockHashes, b.Header().Hash())
 	i.first += 1
 }
 
@@ -434,6 +449,28 @@ func (s *State) NextBlock() types.GlobalBlockNumber {
 	panic("unreachable")
 }
 
+// GlobalBlockByHash returns the finalized GlobalBlock whose stored header
+// hashes to the given value, or None if no such block is currently in the
+// retained range. The lookup-and-construct happens under a single lock so
+// the returned block matches the looked-up hash atomically — pruning can't
+// change which height a hash maps to between the index check and the block
+// construction. Tracks the same retain window as Block / GlobalBlock since
+// the hash index is maintained in lockstep by insertBlock / pruneFirst.
+//
+// Returns an error in the signature for forward-compat with the eventual
+// switch to sei-db/ledger_db/block.BlockDB.GetBlockByHash, which can fail
+// with real I/O errors. The current in-memory implementation never errors.
+func (s *State) GlobalBlockByHash(hash types.BlockHeaderHash) (utils.Option[*types.GlobalBlock], error) {
+	for inner := range s.inner.Lock() {
+		n, ok := inner.blockHashes[hash]
+		if !ok {
+			return utils.None[*types.GlobalBlock](), nil
+		}
+		return utils.Some(inner.globalBlockAt(s.Committee(), n)), nil
+	}
+	panic("unreachable")
+}
+
 // Block returns the block with the given global number.
 // This function is used for syncing - GlobalBlock can be derived from Block and FullCommitQC,
 // which have to be fetched upfront anyway.
@@ -469,6 +506,21 @@ func (s *State) TryBlock(n types.GlobalBlockNumber) (*types.Block, error) {
 	panic("unreachable")
 }
 
+// globalBlockAt assembles the GlobalBlock at height n from inner state.
+// Caller must already hold the inner lock and have verified n is in
+// [inner.first, inner.nextBlock).
+func (i *inner) globalBlockAt(c *types.Committee, n types.GlobalBlockNumber) *types.GlobalBlock {
+	b := i.blocks[n]
+	qc := i.qcs[n].QC()
+	return &types.GlobalBlock{
+		GlobalNumber:  n,
+		Timestamp:     qc.Proposal().BlockTimestamp(c, n).OrPanic("global block not in QC"),
+		Header:        b.Header(),
+		Payload:       b.Payload(),
+		FinalAppState: qc.Proposal().App(),
+	}
+}
+
 // GlobalBlock returns the block with the given global number.
 // Returns ErrPruned if the block has already been pruned.
 func (s *State) GlobalBlock(ctx context.Context, n types.GlobalBlockNumber) (*types.GlobalBlock, error) {
@@ -481,15 +533,7 @@ func (s *State) GlobalBlock(ctx context.Context, n types.GlobalBlockNumber) (*ty
 		if n < inner.first {
 			return nil, ErrPruned
 		}
-		b := inner.blocks[n]
-		qc := inner.qcs[n].QC()
-		return &types.GlobalBlock{
-			GlobalNumber:  n,
-			Timestamp:     qc.Proposal().BlockTimestamp(s.Committee(), n).OrPanic("global block not in QC"),
-			Header:        b.Header(),
-			Payload:       b.Payload(),
-			FinalAppState: qc.Proposal().App(),
-		}, nil
+		return inner.globalBlockAt(s.Committee(), n), nil
 	}
 	panic("unreachable")
 }
--- a/sei-tendermint/internal/autobahn/data/state.go
+++ b/sei-tendermint/internal/autobahn/data/state.go
@@ -507,8 +507,8 @@ func (s *State) TryBlock(n types.GlobalBlockNumber) (*types.Block, error) {
 }
 
 // globalBlockAt assembles the GlobalBlock at height n from inner state.
-// Caller must already hold the inner lock and have verified n is in
-// [inner.first, inner.nextBlock).
+// Caller must have verified n is in [inner.first, inner.nextBlock); n
+// outside that range nil-derefs on inner.blocks[n] / inner.qcs[n].
 func (i *inner) globalBlockAt(c *types.Committee, n types.GlobalBlockNumber) *types.GlobalBlock {
 	b := i.blocks[n]
 	qc := i.qcs[n].QC()
--- a/sei-tendermint/internal/p2p/giga_router.go
+++ b/sei-tendermint/internal/p2p/giga_router.go
@@ -163,11 +163,11 @@ func (r *GigaRouter) BlockByNumber(ctx context.Context, n int64) (*coretypes.Res
 	if err != nil {
 		// Map Autobahn's pruning sentinel to CometBFT's, so callers
 		// (env.Block, evmrpc, ops tooling) get the same error type they
-		// already handle on the CometBFT path. env.getHeight returns
-		// ErrHeightNotAvailable for the same case (height < base).
+		// already handle on the CometBFT path. base is None because the
+		// active lower bound (data.State.inner.first) is internal to
+		// data.State; both call sites format through the same helper.
 		if errors.Is(err, data.ErrPruned) {
-			return nil, fmt.Errorf("%w (requested height: %d)",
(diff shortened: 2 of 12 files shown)


sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -120,6 +123,131 @@ func (r *GigaRouter) LastCommittedBlockNumber() int64 {
...
+		for i, tx := range srcTxs {
+			tmTxs[i] = tx
+		}
+	}
+	return &coretypes.ResultBlock{
+		BlockID: types.BlockID{Hash: blockHash},
+		Block: &types.Block{
+			Header: types.Header{
+				ChainID: r.cfg.GenDoc.ChainID,
+				Height:  utils.Clamp[int64](uint64(gb.GlobalNumber)),
```
pompon0: why do we need to cast twice here?

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -120,6 +123,131 @@ func (r *GigaRouter) LastCommittedBlockNumber() int64 {
...
+// tools (block explorers, ethers/web3 clients, monitoring) don't see
+// different shapes when Sei runs under Autobahn vs CometBFT. Under CometBFT
+// a missing block returns ResultBlock{Block: nil}; partial state populates
+// what's present and leaves the rest at zero values — the call never fails
+// on missing data. We do the same: populate what we have, leave the rest
+// at zero, never nil-deref. A malformed-block condition (nil header or
+// payload despite a successful lookup) shows up as zero-valued fields in
+// the response — observable without crashing the handler.
+func (r *GigaRouter) translateGlobalBlock(gb *atypes.GlobalBlock) *coretypes.ResultBlock {
+	if gb == nil {
```
pompon0: either assume non-nil, or use Option.

pompon0: same in other places where you check for nil - this is a type's property whether its fields are allowed to be nil/non-nil. You should not switch the expectations of the same type in different contexts, otherwise passing a value from one context to another would be insecure.

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -120,6 +123,131 @@ func (r *GigaRouter) LastCommittedBlockNumber() int64 {
...
+//
+// Lookup-and-construct happens under a single data.State lock acquire, so
+// the returned block matches the requested hash atomically. Hashes below
+// the pruning watermark are not indexed and read as "unknown".
+//
+// TODO(autobahn): replace this with a direct read from
+// sei-db/ledger_db/block.BlockDB.GetBlockByHash once a writer is wired into
+// block execution. The data.State-side index can also go away at that point.
+func (r *GigaRouter) BlockByHash(_ context.Context, hash []byte) (*coretypes.ResultBlock, error) {
+	if len(hash) != len(atypes.BlockHeaderHash{}) {
```
pompon0: from the perspective of type-safety you should convert to atypes.BlockHeaderHash at the call site, rather than parse stuff inside. If you want to keep it weakly typed, consider moving the weakly typed functions to another file, preferrably as methods of a wrapper of GigaRouter, sth like:
```
type GigaRouterRPC *GigaRouter

func (r GigaRouterRPC) BlockByHash(...) {...}
...
```

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -120,6 +123,131 @@ func (r *GigaRouter) LastCommittedBlockNumber() int64 {
...
+		return nil, fmt.Errorf("invalid height %d", n)
+	}
+	gb, err := r.data.GlobalBlock(ctx, gbn)
+	if err != nil {
+		// Map Autobahn's pruning sentinel to CometBFT's, so callers
+		// (env.Block, evmrpc, ops tooling) get the same error type they
+		// already handle on the CometBFT path. env.getHeight returns
+		// ErrHeightNotAvailable for the same case (height < base).
+		if errors.Is(err, data.ErrPruned) {
+			return nil, fmt.Errorf("%w (requested height: %d)",
```
pompon0: can we get this error formatting deduplicated with CometBFT path?

sei-tendermint/internal/rpc/core/blocks.go
```diff
@@ -102,9 +125,65 @@ func (env *Environment) Block(ctx context.Context, req *coretypes.RequestBlockIn
...
 }
 
+// gigaRouter returns the GigaRouter when Autobahn is enabled. Callers should
+// only invoke this when env.GigaEnabled is true; if either env.Router or
+// env.Router.Giga() is unset under that condition the wiring in node.go is
+// broken (GigaEnabled is set alongside the router) — surface that loudly
+// rather than silently falling through to a CometBFT path that would also
+// fail under Autobahn.
+func (env *Environment) gigaRouter() (*p2p.GigaRouter, error) {
+	if env.Router == nil {
```
pompon0: make it Option

pompon0: is nil even a valid value here though?

sei-tendermint/internal/rpc/core/blocks.go
```diff
@@ -102,9 +125,65 @@ func (env *Environment) Block(ctx context.Context, req *coretypes.RequestBlockIn
 	return &coretypes.ResultBlock{BlockID: blockMeta.BlockID, Block: block}, nil
 }
 
+// gigaRouter returns the GigaRouter when Autobahn is enabled. Callers should
+// only invoke this when env.GigaEnabled is true; if either env.Router or
```
pompon0: At this point GigaEnables seems redundant with just calling gigaRouter(), which should return either `(Option[*p2p.GigaRouter],error)` or `Option[*p2p.GigaRouter]` or `(*p2p.GigaRouter, bool)`

sei-tendermint/internal/autobahn/data/state.go
```diff
@@ -434,6 +449,28 @@ func (s *State) NextBlock() types.GlobalBlockNumber {
...
+// GlobalBlockByHash returns the finalized GlobalBlock whose stored header
+// hashes to the given value, or None if no such block is currently in the
+// retained range. The lookup-and-construct happens under a single lock so
+// the returned block matches the looked-up hash atomically — pruning can't
+// change which height a hash maps to between the index check and the block
+// construction. Tracks the same retain window as Block / GlobalBlock since
+// the hash index is maintained in lockstep by insertBlock / pruneFirst.
+//
+// Returns an error in the signature for forward-compat with the eventual
+// switch to sei-db/ledger_db/block.BlockDB.GetBlockByHash, which can fail
```
pompon0: nit: I wouldn't care about signature compatibility that much, but up to you.

pompon0: nit: also, it would be great if we could integrate with the db in a way that db internal errors are not reported as query result, but rather shutdown the db background tasks, similarly to how we currently do persistence. Then query would need to take ctx as an argument, and it would still return (Option[Block],error), but the only possible error would be context.Canceled. Not an action item for now, just thinking out loud.

sei-tendermint/internal/autobahn/data/state.go
```diff
@@ -469,6 +506,21 @@ func (s *State) TryBlock(n types.GlobalBlockNumber) (*types.Block, error) {
 	panic("unreachable")
 }
 
+// globalBlockAt assembles the GlobalBlock at height n from inner state.
+// Caller must already hold the inner lock and have verified n is in
```
pompon0: nit: if they didn't hold the inner lock, then inner would not be accessible

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -120,6 +123,131 @@ func (r *GigaRouter) LastCommittedBlockNumber() int64 {
...
 
+// MaxGasPerBlock returns the producer's configured max gas per block,
+// clamped to int64 range. Config validation only enforces > 0
+// (sei-tendermint/config/autobahn.go), so we Clamp here to match the
+// producer's own internal use at producer/state.go and avoid a silent
+// overflow on a misconfigured chain. Exposed so the RPC layer can populate
+// ResultBlockResults.ConsensusParamUpdates under Autobahn (where
+// FinalizeBlock responses are not stored on disk).
+func (r *GigaRouter) MaxGasPerBlock() int64 {
+	return utils.Clamp[int64](r.cfg.Producer.MaxGasPerBlock)
```
pompon0: Do not duplicate the logic: make the clamping a method of producer config, or change the field type to int64, if it is always used as int64 anyway (although then it would need to be validated to be non-negative somewhere).

sei-tendermint/internal/p2p/giga_router_test.go
```diff
@@ -365,9 +366,92 @@ func TestGigaRouter_FinalizeBlocks(t *testing.T) {
...
+			if err != nil {
+				return fmt.Errorf("router[%v].BlockByHash(unknown): %w", i, err)
+			}
+			if rbu == nil {
+				return fmt.Errorf("router[%v].BlockByHash(unknown): nil result", i)
+			}
+			if rbu.Block != nil {
+				return fmt.Errorf("router[%v].BlockByHash(unknown): got non-nil block, want nil", i)
+			}
+			// Wrong-size hashes are also "unknown" in the CometBFT contract.
```
pompon0: it is a weird place to test behavior on storage misses. How about a separate unit test, perhaps with a smaller setup?

sei-tendermint/internal/rpc/core/blocks.go
wen-coding: That's a good point, changed.

sei-tendermint/internal/p2p/giga_router.go
wen-coding: Keeping the field as uint64 and clamps to int64 when accessed.

sei-tendermint/internal/p2p/giga_router_test.go
wen-coding: split into a separate unittest.

sei-tendermint/internal/p2p/giga_router.go
wen-coding: done, the check is at call site now.

sei-tendermint/internal/p2p/giga_router.go

sei-tendermint/internal/autobahn/data/state.go
wen-coding: Changed.

sei-tendermint/internal/p2p/giga_router.go
wen-coding: I think we can assume non-nil here to make it simpler, changed.

sei-tendermint/internal/rpc/core/blocks.go
wen-coding: Changed to Option. The env.Router can be nil in inspect mode, it should keep working for now.

sei-tendermint/internal/autobahn/data/state.go
wen-coding: That's reasonable, added TODO

sei-tendermint/internal/p2p/giga_router.go

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -120,6 +123,119 @@ func (r *GigaRouter) LastCommittedBlockNumber() int64 {
...
+//
+// TODO(autobahn): switch this to read from sei-db/ledger_db/block.BlockDB
+// once a writer is wired (e.g. from app.FinalizeBlocker or executeBlock).
+// Today no production code calls BlockDB.WriteBlock, so Autobahn's in-memory
+// data.State is the only place a full block lives — but it's pruned per
+// Sei's RetainHeight and exposes only a height index (no GetBlockByHash).
+// BlockDB has the right shape (height + hash indexes, async pruning) and
+// is the long-term home for this read path.
+func (r *GigaRouter) BlockByNumber(ctx context.Context, n int64) (*coretypes.ResultBlock, error) {
+	gbn, ok := utils.SafeCast[atypes.GlobalBlockNumber](n)
```
pompon0: same as with BlockByHash - cast outside

sei-tendermint/internal/p2p/giga_router_test.go
```diff
@@ -365,9 +366,72 @@ func TestGigaRouter_FinalizeBlocks(t *testing.T) {
...
+			}
+			// Round-trip the just-fetched block hash back through
+			// BlockByHash and assert we land on the same height + bytes.
+			var hashKey atypes.BlockHeaderHash
+			copy(hashKey[:], rb.BlockID.Hash)
+			rbh, err := giga.BlockByHash(ctx, hashKey)
+			if err != nil {
+				return fmt.Errorf("router[%v].BlockByHash(%x): %w", i, rb.BlockID.Hash, err)
+			}
+			if rbh == nil || rbh.Block == nil {
```
pompon0: shouldn't we just compare rb and rbh (via either require.Equal or utils.TestDiff(), whichever works better here)

sei-tendermint/internal/p2p/giga_router_test.go
```diff
@@ -365,9 +366,72 @@ func TestGigaRouter_FinalizeBlocks(t *testing.T) {
...
+			if rbh.Block.Height != committed {
+				return fmt.Errorf("router[%v].BlockByHash(%x): got height %v, want %v",
+					i, rb.BlockID.Hash, rbh.Block.Height, committed)
+			}
+			if !bytes.Equal(rbh.BlockID.Hash, rb.BlockID.Hash) {
+				return fmt.Errorf("router[%v].BlockByHash(%x): got hash %x, want round-trip",
+					i, rb.BlockID.Hash, rbh.BlockID.Hash)
+			}
+		}
+		// At least one of the global blocks we just finalized must carry txs
```
pompon0: this invariant seems rather obfuscated to me. We already check that all the transactions sent to the chain have been included. You can do the same exact check but using BlockByNumber() for fetching instead. In fact you can even assert that GlobalBlock(n).Payload.Txs is exactly the same as txs in BlockByNumber(n).

sei-tendermint/internal/rpc/core/consensus.go
```diff
@@ -17,7 +17,45 @@ import (
 // for the validators in the set as used in computing their Merkle root.
 //
 // More: https://docs.tendermint.com/master/rpc/#/Info/validators
+//
+// Under Autobahn the CometBFT StateStore is not populated, but the committee
+// is fixed at genesis (no validator-updates path under Autobahn), so any
+// retained height returns the genesis committee. Pagination + error shape
+// mirror the CometBFT path so external tools see consistent responses.
 func (env *Environment) Validators(ctx context.Context, req *coretypes.RequestValidators) (*coretypes.ResultValidators, error) {
+	if _, ok := env.gigaRouter().Get(); ok {
```
pompon0: IsPresent()

sei-tendermint/internal/rpc/core/env.go
```diff
@@ -127,6 +120,26 @@ func validatePage(pagePtr *int, perPage, totalCount int) (int, error) {
...
+// is the definition of "Autobahn is active for this Environment"). Returns
+// None when running under CometBFT or when the Router itself isn't set.
+// Handlers that produce different shapes under Autobahn just branch on the
+// returned Option:
+//
+//	if r, ok := env.gigaRouter().Get(); ok {
+//	    // Autobahn path, r is the router
+//	}
+//
+// This both replaces a previous bool gate (GigaEnabled) and gives every
```
pompon0: this comment explains code history, imo not very helpful.

sei-tendermint/internal/rpc/core/status.go
```diff
@@ -179,7 +175,7 @@ func (env *Environment) validatorAtHeight(h int64) utils.Option[*types.Validator
 	// validator set), so we skip that fast path and fall straight through to
 	// the state-store lookup below, which is kept in sync regardless of
 	// consensus engine.
-	if !env.GigaEnabled {
+	if _, autobahn := env.gigaRouter().Get(); !autobahn {
```
pompon0: ditto

pompon0 (approved):

sei-tendermint/internal/p2p/giga_router.go

sei-tendermint/internal/p2p/giga_router_test.go
wen-coding: utils.TestDiff uses go-cmp which panics on unexported sync.Mutex inside *types.Block, using require.Equal instead

sei-tendermint/internal/rpc/core/consensus.go
wen-coding: done

sei-tendermint/internal/rpc/core/env.go
wen-coding: removed

sei-tendermint/internal/p2p/giga_router_test.go
wen-coding: Removed

sei-tendermint/internal/rpc/core/status.go
wen-coding: changed

arajasek (approved):

PR #3312, opened 2026-04-27
feat(flatkv): add seidb tooling for  flatkv dump state and state analysis

blindchaser:
## Summary

This PR adds FlatKV-aware seidb tooling for read-only analysis and dump workflows. The new tooling clones immutable FlatKV snapshots plus changelog state into a temporary store before scanning, so live nodes keep their writer lock while CLI commands operate on a stable view.

- `sei-db/tools/cmd/seidb/main.go`: registers `dump-flatkv` alongside the existing seidb operations.
- `sei-db/tools/cmd/seidb/operations/flatkv_open.go`: selects the latest `current` snapshot for height `0` or the highest snapshot at or below an explicit historical height. The clone path hardlinks snapshot and changelog files, skips live `LOCK` files, falls back to byte-copy only for cross-device links, retries ENOENT races from concurrent snapshot pruning, and removes the temporary clone on close.
- `sei-db/tools/cmd/seidb/operations/dump_flatkv.go`: dumps FlatKV physical keys into ordered bucket files for `account`, `code`, `storage`, and `legacy`. It preserves physical key bytes, filters optional single-bucket output without changing iterator order, checks iterator errors, and returns final flush/close failures so truncated buffered dumps do not report success.
- `sei-db/tools/cmd/seidb/operations/flatkv_state_size.go`: adds FlatKV state-size aggregation with per-DB totals and top EVM contract storage accounting. It classifies physical keys by FlatKV type prefix, emits a dedicated `flatkv` DynamoDB row shape, and rejects partial scan results by surfacing iterator errors.
- `sei-db/tools/cmd/seidb/operations/state_size.go`: auto-detects sibling FlatKV data or accepts `--flatkv-dir`, scans FlatKV only for compatible module filters, prints FlatKV totals next to memIAVL totals, and skips FlatKV export when open or scan errors would make the result incomplete.

## Test plan

- `sei-db/tools/cmd/seidb/operations/dump_flatkv_test.go`: covers all-bucket dumps, single-bucket filtering, dump file headers, physical key/value formatting, and bucket validation.
- `sei-db/tools/cmd/seidb/operations/flatkv_state_size_test.go`: covers physical key classification, contract address extraction, per-DB size accounting, top-contract ordering, and DynamoDB analysis row shape.
- `sei-db/tools/cmd/seidb/operations/flatkv_open_test.go`: covers latest snapshot selection from `current`, explicit historical snapshot selection, missing current and missing snapshot failures, changelog cloning, lock-file exclusion, retry-on-ENOENT behavior, and real read-only opens for latest and historical heights.
- Verified with `gofmt -s -l` and `go test ./sei-db/tools/cmd/seidb/operations`.


sei-db/tools/cmd/seidb/operations/flatkv_state_size.go
```diff
@@ -0,0 +1,234 @@
...
+// row in the shared DynamoDB state-size table. Consumers key off this name
+// to distinguish FlatKV from memIAVL module rows.
+const flatkvAnalysisModuleName = "flatkv"
+
+// FlatKVStateSizeResult holds the complete analysis of a FlatKV store.
+type FlatKVStateSizeResult struct {
+	TotalNumKeys   uint64
+	TotalKeySize   uint64
+	TotalValueSize uint64
+	TotalSize      uint64
```
cody-littley: Could these be collapsed down into a FlatKVDBSize struct?

sei-db/tools/cmd/seidb/operations/flatkv_state_size.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	if len(innerKey) == 0 {
+		return flatkvBucketLegacy
+	}
+	switch innerKey[0] {
+	case 0x0a:
+		return flatkvBucketAccount
+	case 0x07:
+		return flatkvBucketCode
+	case 0x03:
```
cody-littley: We have constants defined for these, right?

cody-littley (approved):

Kbhat1 (approved):

sei-db/tools/cmd/seidb/operations/flatkv_state_size.go
blindchaser: good point, addressed

PR #3314, opened 2026-04-27
optimizations of Tx Hash handling

pompon0:
* Merged Tx.Hash and Tx.Key into one method to avoid confusion
* Added hashedTx wrapper type which keeps precomputed TxHash


sei-will (approved):

wen-coding (approved):

sei-tendermint/internal/mempool/priority_queue_test.go
```diff
@@ -385,45 +386,45 @@ func TestTxPriorityQueue_TryReplacement(t *testing.T) {
...
+				{hashedTx: newHashedTx(types.Tx("def")), isEVM: true, evmAddress: "addr1", evmNonce: 0, priority: 100},
 			}, false, false, []*WrappedTx{
-				{isEVM: true, evmAddress: "addr1", evmNonce: 0, priority: 100, tx: []byte("def")},
-				{isEVM: true, evmAddress: "addr1", evmNonce: 1, priority: 100, tx: []byte("abc")},
+				{hashedTx: newHashedTx(types.Tx("def")), isEVM: true, evmAddress: "addr1", evmNonce: 0, priority: 100},
+				{hashedTx: newHashedTx(types.Tx("abc")), isEVM: true, evmAddress: "addr1", evmNonce: 1, priority: 100},
 			}, []*WrappedTx{
-				{isEVM: true, evmAddress: "addr1", evmNonce: 0, priority: 100, tx: []byte("def")},
-				{isEVM: true, evmAddress: "addr1", evmNonce: 1, priority: 100, tx: []byte("abc")},
+				{hashedTx: newHashedTx(types.Tx("def")), isEVM: true, evmAddress: "addr1", evmNonce: 0, priority: 100},
```
wen-coding, line 395: nit: I like what you are doing in this PR. But I'm a bit worried that we need to change so many tests. Would it make our lives easier to have a constructor
`NewWrappedTx(tx types.Tx, opts...)` instead so we don't give the tests a chance to know about newHashedTx?

pompon0: Although I do agree that strengthening WrappedTx type and making tests internals agnostic would benefit the correctness, I decided to scope down this pr to just hash field protection to limit the diff size. Afaict all the test changes are mechanical and personally I feel comfortable with making them.

PR #3321, opened 2026-04-27
fix(evmrpc): include projected child base fee in eth_feeHistory response

amir-deris:
## Summary

Fixes a spec mismatch in `eth_feeHistory` ([execution-apis spec](https://github.com/ethereum/execution-apis/blob/main/src/eth/fee_market.yaml)): `baseFeePerGas` must contain one more entry than `gasUsedRatio`, where the extra entry is the projected base fee for the child of the newest block in the response range.

- Appends the child block's projected base fee to `result.BaseFee` after the main loop, gated on the current chain version check.
- Renames `safeGetBaseFee` → `safeGetHeaderBaseFee` and corrects the lookup to use `GetNextBaseFeePerGas` at the **parent** committed height (matching the semantics used in block headers and `encodeRPCTransaction`).
- Adds `safeGetChildBaseFeeAfter` to retrieve the projected base fee for the block immediately following a given height.
- Adds `TestFeeHistoryBaseFeePerGasIncludesChild` to assert `len(baseFeePerGas) == len(gasUsedRatio) + 1`.


evmrpc/info.go
```diff
@@ -279,13 +298,33 @@ func (i *InfoAPI) Syncing() (result any, returnErr error) {
 	return nil, &ErrEVMNotSupported{Msg: "eth_syncing is not supported on Sei EVM RPC"}
 }
 
-func (i *InfoAPI) safeGetBaseFee(targetHeight int64) (res *big.Int) {
+// safeGetHeaderBaseFee returns the base fee per gas for txs in block blockNum (same as eth block header
+// and encodeRPCTransaction: GetNextBaseFee at parent committed height).
+func (i *InfoAPI) safeGetHeaderBaseFee(blockNum int64) (res *big.Int) {
+	defer func() {
+		if err := recover(); err != nil {
+			res = nil
```
masih, line 307: Log the error?

evmrpc/info_test.go
```diff
@@ -381,6 +381,17 @@ func TestFeeHistoryGasUsedRatioCalculation(t *testing.T) {
 	require.Equal(t, 1, len(gasUsedRatios2), "Should have exactly one gas used ratio for single block")
 }
 
+// execution-apis eth_feeHistory: len(baseFeePerGas) == len(gasUsedRatio) + 1 (extra child base fee).
+func TestFeeHistoryBaseFeePerGasIncludesChild(t *testing.T) {
```
masih, line 385: Could you also assert the values in test please?

evmrpc/info.go
```diff
@@ -234,6 +242,17 @@ func (i *InfoAPI) FeeHistory(ctx context.Context, blockCount gmath.HexOrDecimal6
 		}
 		result.Reward = append(result.Reward, rewards)
 	}
+
+	// execution-apis eth_feeHistory / go-ethereum: baseFeePerGas has one more element than gasUsedRatio,
+	// the projected base fee for the child of the newest block in the range.
+	// Note: len(baseFeePerGas) may still differ from len(gasUsedRatio)+1 when some heights skip header
+	// base fees (pruned / partial data) while gasUsedRatio rows exist — same class of partial history as before.
```
masih, line 249: Thank you for documenting this; do we plan to fix this in a follow up PR?

masih (approved):
Thank you for fixing this! nice 🙌

bdchatham (approved):

evmrpc/info.go
amir-deris: Created this ticket to follow up: https://linear.app/seilabs/issue/PLT-293/align-eth-feehistory-response-shape-with-ethereum-geth-client

PR #3323, opened 2026-04-27
fix: tolerate optional package failures in localnode Dockerfile

amir-deris:
## Summary

- Installs apt packages individually in the localnode Dockerfile instead of in a single `apt-get install` invocation
- A failure to install any one package no longer aborts the entire build
- Fixes intermittent CI build failures caused by transient or environment-specific package unavailability

## Changes

`docker/localnode/Dockerfile`: replaced single bulk `apt-get install` with a loop that installs each package independently and prints a warning on failure instead of exiting.

### DIFF
--- a/docker/localnode/Dockerfile
+++ b/docker/localnode/Dockerfile
@@ -3,7 +3,9 @@ FROM golang:1.25.6-bookworm AS go-dist
 FROM ubuntu:latest
 ENV HOME="/root" PATH="/root/go/bin:/sei-protocol/sei-chain/integration_test/upgrade_module/scripts/:$PATH"
 RUN for i in 1 2 3; do apt-get update && break || { [ $i -lt 3 ] && sleep 5; }; done && \
-    apt-get install -y make build-essential git jq python3 curl vim uuid-runtime
+    for pkg in make build-essential git jq python3 curl vim uuid-runtime; do \
+      apt-get install -y "$pkg" || echo "Skipping optional package: $pkg"; \
+    done
 COPY --from=go-dist /usr/local/go /usr/local/go
 RUN set -eux; \
     curl -fsSL https://foundry.paradigm.xyz | bash; \
--- a/docker/localnode/Dockerfile
+++ b/docker/localnode/Dockerfile
@@ -3,9 +3,8 @@ FROM golang:1.25.6-bookworm AS go-dist
 FROM ubuntu:latest
 ENV HOME="/root" PATH="/root/go/bin:/sei-protocol/sei-chain/integration_test/upgrade_module/scripts/:$PATH"
 RUN for i in 1 2 3; do apt-get update && break || { [ $i -lt 3 ] && sleep 5; }; done && \
-    for pkg in make build-essential git jq python3 curl vim uuid-runtime; do \
-      apt-get install -y "$pkg" || echo "Skipping optional package: $pkg"; \
-    done
+    for i in 1 2 3; do apt-get install -y make build-essential git jq python3 curl uuid-runtime && break || { [ $i -lt 3 ] && sleep 5; }; done && \
+    for pkg in vim; do apt-get install -y "$pkg" || echo "Skipping optional package: $pkg"; done
 COPY --from=go-dist /usr/local/go /usr/local/go
 RUN set -eux; \
     curl -fsSL https://foundry.paradigm.xyz | bash; \


docker/localnode/Dockerfile
```diff
@@ -3,7 +3,9 @@ FROM golang:1.25.6-bookworm AS go-dist
 FROM ubuntu:latest
 ENV HOME="/root" PATH="/root/go/bin:/sei-protocol/sei-chain/integration_test/upgrade_module/scripts/:$PATH"
 RUN for i in 1 2 3; do apt-get update && break || { [ $i -lt 3 ] && sleep 5; }; done && \
-    apt-get install -y make build-essential git jq python3 curl vim uuid-runtime
+    for pkg in make build-essential git jq python3 curl vim uuid-runtime; do \
```
wen-coding: Are you sure all packages here are optional? If we don't have all packages here could we continue to build? I think at least we need to have make build-essential git jq python3 curl

My original change is to retry one-line apt-get install 3 times so it fails less. Do we have other ways to make apt-get install succeed more? I saw 404s. Can we cache these packages locally?

bdchatham: +1 to concern around optionality. 

Retry policy strikes me as the simplest reliable fix

wen-coding: Can we move the apt-installed packages into some Docker base image? I don't suppose we need to always install the most up-to-date version.

amir-deris: Some of these packages seem to be optional, like `vim`. The build and all the tests succeeded today, even though `vim` couldn't be installed. 
I can split them into required and optional packages. Required ones will fail the build, and optional ones can continue. 
I can also add retry logic similar to apt-get update. 
```
About caching, there is a catch for CI: this cache lives on the build host. In ephemeral GitHub Actions runners, the host is fresh every run, so the cache is always cold unless you explicitly save/restore 
  it via actions/cache. That adds setup complexity and the cache can still be busted (e.g. new runner image). So in practice, for CI this only helps if you're using self-hosted runners or a persistent build machine.
```
Another approach is to specify the ubuntu version explicitly instead of going with latest. 
``` When that tag rolls to a new major version (e.g. 22.04 → 24.04), apt mirrors take time to propagate and      
  package availability can change. The retry logic is a band-aid on top of this. Pinning to a specific version like ubuntu:24.04 would eliminate the whole class of flakiness — the retry
   and optional/required split become much less important when the base image is stable and predictable.```

Comment (wen-coding) [reply docker/localnode/Dockerfile]: I asked AI and this is the list it suggested:

1. Use a language base image that already has the tools. golang:bookworm (which this Dockerfile already pulls as go-dist) ships with make, build-essential, git, curl preinstalled. Half this apt list is redundant if you base the runtime stage on golang:…-bookworm instead of bare ubuntu:latest.
2. Maintain a custom org base image (org/ci-base:tag) with everything baked in, rebuilt on a schedule or when the package list changes. apt-get runs once a week, not 200×/day.
3. Devcontainer / Nix / Bazel toolchains for repos that want full hermeticity.

Does any of the options sound reasonable to you?

File: docker/localnode/Dockerfile [hunk 2]
```diff
@@ -3,9 +3,8 @@ FROM golang:1.25.6-bookworm AS go-dist
 FROM ubuntu:latest
 ENV HOME="/root" PATH="/root/go/bin:/sei-protocol/sei-chain/integration_test/upgrade_module/scripts/:$PATH"
 RUN for i in 1 2 3; do apt-get update && break || { [ $i -lt 3 ] && sleep 5; }; done && \
-    for pkg in make build-essential git jq python3 curl vim uuid-runtime; do \
-      apt-get install -y "$pkg" || echo "Skipping optional package: $pkg"; \
-    done
+    for i in 1 2 3; do apt-get install -y make build-essential git jq python3 curl uuid-runtime && break || { [ $i -lt 3 ] && sleep 5; }; done && \
+    for pkg in vim; do apt-get install -y "$pkg" || echo "Skipping optional package: $pkg"; done
```
wen-coding, line 7: Does any of our tests seriously use vim?

wen-coding (approved):

docker/localnode/Dockerfile
amir-deris: Thanks for great feedback. I think option 1 and 2 seem great. I have to look into option 3. 
I will work on implementing a long term fix based on these options in a different ticket.
Follow up: https://linear.app/seilabs/issue/PLT-290/issue-with-our-dockerfiles

docker/localnode/Dockerfile
amir-deris: I don't think so. But I am not sure if this Dockerfile is used only for tests or has other usages.

amir-deris:
Since the immediate issue is resolved and main branch has build success again, I will close this temporary fix PR and work on long term fix.

PR #3326, opened 2026-04-28
replace eigen logger

cody-littley:

LittDB originally used an in-house logger from Eigen. We obviously don't want to use the Eigen logger. This PR swaps out the logger for a generic `slog.Logger`.

Note that LittDB is still not compiling automatically with the remainder of the toolchain. To build these changes, `cd` to the `litt` directory and run `make build`. There are additional third party dependencies to be removed prior to full integration.


sei-db/db_engine/litt/benchmark/benchmark_engine.go
```diff
@@ -5,13 +5,13 @@ package benchmark
 import (
 	"context"
 	"fmt"
+	"log/slog"
```
bdchatham, line 8: Thinking it would make sense to use seilog instead

https://github.com/sei-protocol/seilog

cody-littley: seilog is an implementation of slog. When this is fully integrated with our stack, it will be using a seilog instance. What benefit do we gain by using the more specific type at the call sites?

cody-littley: Another consideration. We may eventually want to spin this out into its own repo, potentially for use by outside parties. If we do that, being more generic for the required log framework would be a good thing to do.

bdchatham: By "implementation of slog" do you mean that it uses slog? 

Our existing sei-chain stack uses seilog throughout, if I'm not mistaken. I can appreciate the goal to make this generally useful but I think it'd be good to focus on making it homogenous in our own stack for now. In the worst case, let a user choose their logger and our own stack will just pick seilog. 

Seilog would add continuity with our existing patterns. Two key wins I see over slog are hot-reload for log level and module-scoped log messages.

cody-littley: seilog implements the slog interface, meaning that you can still use a slogger at runtime even if the interfaces use type slog.

I'm open to moving to seilog in the future, but I'd like to keep the scope of this PR focused. The focus for this PR was to incrementally move LittDB towards being clean by removing a third party dependency.

Seilog is currently missing a critical feature that will require nontrivial refactoring to support. Specifically, seilog requires me to set environment variables in order to configure it, which means that I can't configure it at boot time like I do a normal slogger. I can work around this, but it requires a complex multi-binary startup sequence (see cryptosim pattern).

bdchatham: Makes sense and I'm aligned with using the interface being objectively the right decision. 

What ultimately lead me to write that comment was this: 

`if cfg.LittConfig.Logger == nil {
		cfg.LittConfig.Logger = slog.Default()`
		
I left the comment on the slog import to broadly discuss how we are injecting our actual log implementation and match existing sei-chain patterns. I now realize I should have been more specific. 

If there are feature limitations, can you create a git issue on the seilog package, or wherever it belongs? As long as we're continuously keeping up with a homogenous logging layer over time then I'm happy.

blindchaser (approved):

Kbhat1 (approved):

PR #3332, opened 2026-04-29
fix(evmrpc): route receipt lookups through block cache for consistency

jewei1997:

GetEvmTxIndex, encodeRPCTransaction, and countBlockTxsLikeEncodeTmBlock were calling k.GetReceipt directly while sibling code paths (e.g. filterTransactions) went through getOrSetCachedReceipt. That mismatch let the per-block receipt cache and the backing store disagree within a single request. Switch the remaining callers to getOrSetCachedReceipt so all receipt reads in a block share the same cache view.

existing tests


cody-littley (approved):

evmrpc/tx.go
```diff
@@ -334,9 +334,9 @@ func (t *TransactionAPI) getTransactionWithBlock(block *coretypes.ResultBlock, t
 }
 
 func (t *TransactionAPI) encodeRPCTransaction(ethtx *ethtypes.Transaction, block *coretypes.ResultBlock, txIndex uint32) (*export.RPCTransaction, error) {
-	receipt, err := t.keeper.GetReceipt(t.ctxProvider(LatestCtxHeight), ethtx.Hash())
-	if err != nil {
-		return nil, err
+	receipt, found := getOrSetCachedReceipt(t.cacheCreationMutex, t.globalBlockCache, t.ctxProvider(LatestCtxHeight), t.keeper, block, ethtx.Hash())
+	if !found {
+		return nil, errors.New("receipt not found for transaction")
```
masih: An unwrapped custom error could bite back when it comes to UX or any sort of error type checking up the stack; I recommend wrapping this with `receipt.ErrNotFound` since the type exists already?

masih (approved):
No blockers, one recommendation; thanks for getting this sorted @jewei1997 🙌

PR #3334, opened 2026-04-29
made all Application.CheckTx calls recover panics (CON-259)

pompon0:
This has been achieved by:
* making tendermint logic operate on Proxy rather than Application
* adding panic -> error conversion to Proxy.CheckTx and renamed it to Proxy.CheckTxSafe, so that Proxy no longer implements Application.

Additionally removed some remains of stateful leader election tests, unused PersistentKVStore.


sei-will (approved):

app/abci.go
```diff
@@ -76,16 +77,14 @@ func (app *App) CheckTx(ctx context.Context, req *abci.RequestCheckTxV2) (*abci.
 	sdkCtx := app.GetCheckTxContext(req.Tx, req.Type == abci.CheckTxTypeV2Recheck)
 	tx, err := app.txDecoder(req.Tx)
 	if err != nil {
-		res := sdkerrors.ResponseCheckTx(err, 0, 0, false)
-		return &abci.ResponseCheckTxV2{ResponseCheckTx: &res}, err
+		return nil, err
```
wen-coding, line 80: Why are we returning a nil result here now? Would this break any downstream tools?

sei-tendermint/internal/mempool/mempool.go
```diff
@@ -419,87 +416,74 @@ func (txmp *TxMempool) CheckTx(
...
-		}
-		if res.ExpireTxHandler != nil {
-			res.ExpireTxHandler()
-		}
-	}
-
+	res, err := txmp.app.CheckTxSafe(ctx, &abci.RequestCheckTxV2{Tx: tx})
 	if err != nil {
-		removeHandler(true)
-		res.Log = txmp.AppendCheckTxErr(res.Log, err.Error())
```
wen-coding, line 447: Do we not need the error on log any more?

sei-tendermint/internal/proxy/proxy.go
```diff
@@ -2,87 +2,90 @@ package proxy
...
 	defer addTimeSample(app.metrics.MethodTiming.With("method", "commit", "type", "sync"))()
 	return app.app.Commit(ctx)
 }
 
-func (app *proxyClient) CheckTx(ctx context.Context, req *types.RequestCheckTxV2) (*types.ResponseCheckTxV2, error) {
+func (app *Proxy) CheckTxSafe(ctx context.Context, req *types.RequestCheckTxV2) (res *types.ResponseCheckTxV2, err error) {
 	defer addTimeSample(app.metrics.MethodTiming.With("method", "check_tx", "type", "sync"))()
+	defer func() {
+		if r := recover(); r != nil {
+			err = fmt.Errorf("panic recovered in CheckTxSafe: %v", r)
```
wen-coding: Should we attach debug.Stack() here for debugging?

Maybe we should log this as well?

sei-tendermint/internal/consensus/common_test.go
```diff
@@ -1062,7 +1059,8 @@ func randConsensusNetWithPeers(
 		app.SetValidators(vals)
 		// sm.SaveState(stateDB,state)	//height 1's validatorsInfo already saved in LoadStateFromDBOrGenesisDoc above
 
-		css[i] = newStateWithConfig(t, thisConfig, state, privVal, app)
+		proxyApp := proxy.New(app, proxy.NopMetrics())
```
wen-coding, line 1062: nit: if it's always proxy.NopMetrics() maybe we can use a constructor to hide this.

sei-tendermint/internal/mempool/mempool.go
```diff
@@ -419,87 +416,74 @@ func (txmp *TxMempool) CheckTx(
...
-	removeHandler := func(removeFromCache bool) {
-		if removeFromCache {
-			txmp.cache.Remove(txHash)
-		}
-		if res.ExpireTxHandler != nil {
-			res.ExpireTxHandler()
-		}
-	}
-
+	res, err := txmp.app.CheckTxSafe(ctx, &abci.RequestCheckTxV2{Tx: tx})
```
wen-coding, line 422: Do we have a test which shows we recover after ChekTX panics?

wen-coding (approved):

sei-tendermint/internal/mempool/mempool.go
pompon0: there is no res any more, so there is no Log to add it to.

sei-tendermint/internal/proxy/proxy.go
pompon0: it will be logged by the caller. I will add the stack.

sei-tendermint/internal/consensus/common_test.go
pompon0: in prod we use actual metrics. In tests we usually use kvstore.NewProxy() to hide this, but in those cases where app needs to be adjusted before wrapping we call proxy.New() directly

app/abci.go
pompon0: CheckTx is only called by tendermint

sei-tendermint/internal/mempool/mempool.go
pompon0: done

PR #3337, opened 2026-04-29
fix(seidb-tool): addresses two issues against the FlatKV seidb tooling

blindchaser:
## Summary

This PR addresses two auditor findings against the FlatKV seidb tooling: snapshot clones could silently degrade to multi-GB byte-copies on tmpfs, and a snapshot/WAL race let `catchup` advance over missing versions without complaint. The fixes pin the tool clone to the source filesystem, validate WAL coverage after the clone, and make `catchup` enforce strict continuity from `committedVersion + 1`.

- `sei-db/tools/cmd/seidb/operations/flatkv_open.go`: places the temp clone as a sibling of `dbDir` so `os.Link` keeps working when `$TMPDIR` is tmpfs; replaces the EXDEV byte-copy fallback for snapshot files with a fatal error so a misconfigured deployment cannot silently RAM-copy multi-GB snapshots; introduces `errSourceChurning` and `verifyClonedWALCovers` to detect the race where a live writer rolls a new snapshot and front-truncates the WAL between our snapshot and changelog clone steps; routes that race through the existing retry loop alongside `os.ErrNotExist`. Changelog files keep their byte-copy semantics (WAL recovery may rewrite a corrupted tail).
- `sei-db/state_db/sc/flatkv/store_catchup.go`: makes `catchup` reject WAL gaps loudly. When `committedVersion > 0`, `walFirstVer` must be `<= committedVersion + 1` or catchup returns an error instead of silently jumping forward and corrupting `committedLtHash`. Adds an inner `expectedNext` check inside the replay callback so an internal hole in the WAL (corruption or external surgery) is also rejected. Returns clean no-op when the WAL is entirely behind `committedVersion`.

## Test plan

- `sei-db/tools/cmd/seidb/operations/flatkv_open_test.go`: covers same-filesystem placement of the tooling clone (`TestPrepareFlatKVToolingClonePlacesTempDirOnSameFilesystem`) and the snapshot/WAL truncation race (`TestPrepareFlatKVToolingCloneDetectsWALTruncationRace`), asserting that the latter surfaces `errSourceChurning` so the retry loop can re-select the snapshot. Existing retry-on-ENOENT, current-symlink-missing, and historical-height tests continue to pass.
- `sei-db/state_db/sc/flatkv/store_catchup_test.go`: `TestCatchupRejectsWALGap` builds a v1..v5 store, front-truncates the WAL to start at v4, rewinds `committedVersion` to v2, and asserts catchup returns a `WAL gap` error instead of silently advancing. `TestCatchupNoOpWhenWALBehindCommittedVersion` guards the steady-state case where the WAL is entirely behind `committedVersion`.
- Verified with `gofmt -s -l` (clean) and `go test ./sei-db/state_db/... ./sei-db/tools/cmd/seidb/... ./sei-cosmos/storev2/rootmulti/... -run "Flatkv|FlatKV"` (all green).


sei-db/tools/cmd/seidb/operations/flatkv_open.go
```diff
@@ -12,19 +12,29 @@ import (
...
+	// live writer truncates the WAL past our snapshot between snapshot and
+	// changelog clone steps.
 	maxCloneRetries = 3
 )
 
+// errSourceChurning marks transient races where the source FlatKV directory
+// mutates (snapshot pruned, WAL truncated) between our reads. It is the
+// sentinel that prepareFlatKVToolingCloneWith uses to decide whether to
+// retry instead of bailing out.
+var errSourceChurning = errors.New("flatkv source kept churning during clone")
```
cody-littley, line 36: Not a blocker for this PR. As a future task, we should evaluate and update the flatKV snapshot threading model so that such races are not possible. Even if we're safe now, having to think about potential races makes safety a lot harder to reason about.

cody-littley (approved):

Kbhat1 (approved):

PR #3341, opened 2026-04-29
Producer MaxGas from genesis (autobahn); ConsensusParams on RPC ctx (CON-257)

wen-coding:
Two independent fixes on the path from `genesis.consensus_params.block.max_gas` to its consumers.

## Fix 1 — Autobahn producer sources `MaxGasPerBlock` from genesis

Autobahn-only. `Producer.MaxGasPerBlock` was populated from `autobahn.json`'s `max_gas_per_block`, while the EVM runtime and CometBFT's reaping read from `genDoc.ConsensusParams.Block.MaxGas` — two configurable sources for the same chain rule. `buildGigaConfig` now sources from genesis (validated > 0), matching CometBFT. The autobahn.json field is removed.

## Fix 2 — `RPCContextProvider` populates ConsensusParams

Both engines. `app.GetCheckCtx()` and `CreateQueryContext` build ctxs without `WithConsensusParams` — only the tx-execution path (`getContextForTx`) does. So every RPC ctx had `ctx.ConsensusParams() == nil`, masked under the old `block.go` (which read `blockRes.ConsensusParamUpdates`) and worked around in `info.go` with a fallback cascade. After `2593ceb28` switched `EncodeTmBlock` to `ctx.ConsensusParams()`, `eth_getBlockByNumber.gasLimit` returns 0.

`RPCContextProvider` now calls `WithConsensusParams(app.GetConsensusParams(ctx))` on both branches. `eth_getBlockByNumber.gasLimit` matches the GASLIMIT opcode bit-exactly. The `info.go` cascade can be cleaned up in a follow-up.

## Test plan

- [x] `go test ./sei-tendermint/node/` — incl. new `TestBuildGigaConfig_GenesisMaxGas` (nil / zero / negative)
- [x] `go test ./app/` — incl. new `TestRPCContextProviderPopulatesConsensusParams` (verified to fail without Fix 2)
- [x] `go test ./evmrpc/...` — all 4 packages
- [x] `make autobahn-integration-test` — 5/5 subtests, full cluster boot/teardown
- [x] `gofmt -s -l .` clean, `go vet` clean


sei-tendermint/node/setup_test.go
```diff
@@ -75,7 +77,13 @@ func makeTestGigaDeps() (*mempool.TxMempool, *types.GenesisDoc) {
 		mempool.NopMetrics(),
 		mempool.NopTxConstraintsFetcher,
 	)
-	genDoc := &types.GenesisDoc{ChainID: "test-chain", InitialHeight: 1}
+	genDoc := &types.GenesisDoc{
+		ChainID:       "test-chain",
+		InitialHeight: 1,
```
pompon0: while at it, can we make InitialHeight nontrivial? Or will it break tests?

sei-tendermint/node/setup_test.go
```diff
@@ -193,15 +200,38 @@ func TestBuildGigaConfig_InvalidConfigFile(t *testing.T) {
...
+	valKey := makeTestValidatorKey([]byte("val-seed"))
+	v := makeValidator([]byte("val-seed"), []byte("node-seed"), "localhost:26660")
+	cfgFile := writeAutobahnConfig(t, defaultFileConfig([]config.AutobahnValidator{v}))
+
+	t.Run("nil ConsensusParams", func(t *testing.T) {
+		txMempool, genDoc := makeTestGigaDeps()
+		genDoc.ConsensusParams = nil
+		_, err := buildGigaConfig(cfgFile, nodeKey, valKey, txMempool, genDoc)
+		assert.Error(t, err)
+		assert.Contains(t, err.Error(), "max_gas")
```
pompon0: pls, no error parsing, use errors.Is instead.

pompon0 (approved):

sei-tendermint/node/setup_test.go
wen-coding: done

sei-tendermint/node/setup_test.go
wen-coding: Changed, I don't think it matters here

arajasek (approved):

PR #3344, opened 2026-04-29
fix(evmrpc): return null for out-of-range tx index and fix deferred error capture

amir-deris:
## Problem

Two bugs in `evmrpc/tx.go`:

1. **Deferred error capture was always nil** — `defer recordMetricsWithError(...)` was called directly rather than inside a closure, so `returnErr` was captured at defer-registration time (always `nil`) instead of at function-return time. This caused metrics to never record actual errors.

2. **uint32 overflow returned an error instead of null** — `eth_getTransactionByBlockNumberAndIndex` and `eth_getTransactionByBlockHashAndIndex` returned an RPC error when the transaction index exceeded `math.MaxUint32`. The Ethereum JSON-RPC spec requires these methods to return `null` for an out-of-range index, not an error.

## Fix

- Wrapped all `defer recordMetricsWithError(...)` calls in closures so `returnErr` is evaluated at return time.
- Introduced a typed `txUint32OverflowError` so the overflow case can be distinguished from other errors.
- After recording metrics, suppresses the overflow error (sets `returnErr = nil`) in `GetTransactionByBlockNumberAndIndex` and `GetTransactionByBlockHashAndIndex` to comply with the Ethereum JSON-RPC spec.

## Test plan

- [x] `eth_getTransactionByBlockNumberAndIndex` with a tx index > `0xFFFFFFFF` returns `null`, not an error
- [x] `eth_getTransactionByBlockHashAndIndex` with a tx index > `0xFFFFFFFF` returns `null`, not an error


evmrpc/tx.go
```diff
@@ -110,7 +110,9 @@ func getTransactionReceipt(
 	includeSynthetic bool,
 ) (result map[string]interface{}, returnErr error) {
 	startTime := time.Now()
-	defer recordMetricsWithError("eth_getTransactionReceipt", t.connectionType, startTime, returnErr)
+	defer func() {
+		recordMetricsWithError("eth_getTransactionReceipt", t.connectionType, startTime, returnErr)
```
amir-deris: The previous format `defer functionCall(...)` captures arguments at the call site which results in `returnErr` be nil for `recordMetricsWithError`.

masih: A go convention that's worth following is to use `_` prefix for named return variables to indicate that "this variable must never be directly set". I would stick with simple names too. e.g. `_err`.

masih (approved):

bdchatham (approved):

PR #3346, opened 2026-04-30
set checktx flag on ctx

codchen:
set recheck tx flag properly

unit test in ante handlers

### DIFF
--- a/app/cosmos_checktx_test.go
+++ b/app/cosmos_checktx_test.go
@@ -16,6 +16,118 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// TestReCheckTxSkipsSignatureVerification verifies that actual cryptographic
+// signature verification is skipped when IsRecheckTx is true, so a tx signed
+// with the wrong private key is accepted on recheck but rejected on DeliverTx.
+func TestReCheckTxSkipsSignatureVerification(t *testing.T) {
+	testApp := app.Setup(t, false, false, false)
+	ctx := testApp.NewContext(false, tmproto.Header{Height: 1, ChainID: "sei-test", Time: time.Now().UTC()})
+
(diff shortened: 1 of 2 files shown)


app/cosmos_checktx_test.go
```diff
@@ -16,6 +16,118 @@ import (
...
+	sigV2, err := tx.SignWithPrivKey(txConfig.SignModeHandler().DefaultMode(), signerData, txBuilder, wrongPrivKey, txConfig, 0)
+	require.NoError(t, err)
+	require.NoError(t, txBuilder.SetSignatures(sigV2))
+
+	signedTx := txBuilder.GetTx()
+
+	gasCtx := ctx.WithGasMeter(sdk.NewInfiniteGasMeter(1, 1))
+
+	// ReCheckTx: VerifySignature is skipped entirely, so the wrong-key sig passes.
+	recheckCtx := gasCtx.WithIsCheckTx(false).WithIsReCheckTx(true)
```
pompon0, line 65: WithIsReCheckTx sets checkTx to true anyway, so WithIsCheckTx here is not effective.

app/cosmos_checktx_test.go
```diff
@@ -16,6 +16,118 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// TestReCheckTxSkipsSignatureVerification verifies that actual cryptographic
+// signature verification is skipped when IsRecheckTx is true, so a tx signed
+// with the wrong private key is accepted on recheck but rejected on DeliverTx.
+func TestReCheckTxSkipsSignatureVerification(t *testing.T) {
```
pompon0, line 22: can we have such test with EVM txs as well? Because this test will become obsolete after sei solo.

pompon0 (approved):

arajasek (approved):

PR #3349, opened 2026-04-30
rpc: emit non-nil LastCommit on Autobahn-routed blocks (CON-256)

wen-coding:
`translateGlobalBlock` (added by #3310) returned a `ResultBlock` with `Block.LastCommit == nil`. Trace replay walks `StateAtTransaction → initializeBlock → ToReqBeginBlock`, which unconditionally dereferences `b.LastCommit.Signatures[i]` when building BeginBlock votes — so any debug-trace path hit a nil deref and the RPC handler crashed with a generic `method handler crashed`. `debug_traceCall` was the most visible victim.

## Things done

Emit a non-nil `LastCommit` with `Signatures` sized to `len(genDoc.Validators)`, all entries `BlockIDFlagAbsent`. That's the only requirement `ToReqBeginBlock` actually has — it iterates by index against the validator set, so the slice must be the right size, but the *values* don't matter on the trace path. BeginBlock handlers that read `LastCommitInfo` (distribution, slashing) are no-ops during replay anyway: distribution wouldn't pay rewards a second time on an already-committed block, and slashing's missed-block counter update lands in a snapshot/copy state that's discarded. Synthesizing per-validator sigs from the prior block's `CommitQC` would add storage IO without buying anything observable — and the values we'd populate (raw ed25519 sigs, single uniform timestamp) aren't tendermint-compatible anyway.

## Verification

`make autobahn-integration-test` green; `debug_traceCall` and `debug_traceTransaction` (callTracer + prestateTracer/diffMode) return proper traces on a 4-node Autobahn cluster; 3/3 EVMCompatabilityTest trace tests pass.


sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -209,7 +209,27 @@ func (r *GigaRouter) BlockByHash(_ context.Context, hash atypes.BlockHeaderHash)
...
+// without panicking. Genuine QC fetch errors (ctx cancel, etc.)
+// propagate.
+func (r *GigaRouter) translateGlobalBlock(ctx context.Context, gb *atypes.GlobalBlock) (*coretypes.ResultBlock, error) {
+	// LastCommit of block n carries the proof that n-1 was committed.
+	// gb.GlobalNumber-1 doesn't underflow: NewGigaRouter rejects
+	// InitialHeight<1, Committee.FirstBlock()=InitialHeight>=1, and
+	// data.State only hands out blocks at >=FirstBlock. fetchCommitQC
+	// maps the sub-genesis read at n-1 (i.e. when n==FirstBlock) to
+	// ErrPruned → None, which commitSigsFromQC turns into all-absent.
+	lastQC, err := r.fetchCommitQC(ctx, gb.GlobalNumber-1)
```
pompon0: I think you should check that globalnumber>0 here

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -209,7 +209,27 @@ func (r *GigaRouter) BlockByHash(_ context.Context, hash atypes.BlockHeaderHash)
...
+// LastCommit is built from the CommitQC that finalized
+// gb.GlobalNumber-1 — the CometBFT-shape semantic for block N's
+// LastCommit (proof of N-1's commit, included in N's data). When the
+// prior QC is unavailable (genesis block or pruned out of data.State's
+// retain window) the resulting LastCommit carries an all-absent
+// signature slice sized to genDoc.Validators, which is enough for
+// downstream consumers including trace replay's BeginBlock to run
+// without panicking. Genuine QC fetch errors (ctx cancel, etc.)
+// propagate.
+func (r *GigaRouter) translateGlobalBlock(ctx context.Context, gb *atypes.GlobalBlock) (*coretypes.ResultBlock, error) {
```
pompon0: it seems weird that translation function needs to access storage to do the translation, no? Either make it load the block within, or provide all the necessary data as arguments

pompon0:
the commitqc of previous block is with high probability the same as commitqc of the current block. Do we care? Can we just return empty LastCommit instead? It is not like it provides tendermint-compatible values anyway.

sei-tendermint/internal/p2p/giga_router.go

sei-tendermint/internal/p2p/giga_router.go

wen-coding:
> the commitqc of previous block is with high probability the same as commitqc of the current block. Do we care? Can we just return empty LastCommit instead? It is not like it provides tendermint-compatible values anyway.

Hmm, actually that's a good point. Turns out this would actually work, changed the PR.

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -203,18 +203,31 @@ func (r *GigaRouter) BlockByHash(_ context.Context, hash atypes.BlockHeaderHash)
...
+// anything observable.
 func (r *GigaRouter) translateGlobalBlock(gb *atypes.GlobalBlock) *coretypes.ResultBlock {
 	srcTxs := gb.Payload.Txs()
 	tmTxs := make(types.Txs, len(srcTxs))
 	for i, tx := range srcTxs {
 		tmTxs[i] = tx
 	}
+	absentSigs := make([]types.CommitSig, len(r.cfg.GenDoc.Validators))
+	for i := range absentSigs {
+		absentSigs[i] = types.CommitSig{BlockIDFlag: types.BlockIDFlagAbsent}
```
pompon0: when we call FinalizeBlock, LastCommit is literally empty (not even "absent sigs" entries are present). From the pov of the application (which thinks that it is in control of the committee, while in autobahn it is actually hardcoded rn) this prevents jailing validators. Can we do the same here for consistency?

pompon0 (approved):

sei-tendermint/internal/p2p/giga_router.go
wen-coding: Yeah I don't think we need absentSigs here either, removed.

arajasek (approved):

PR #3350, opened 2026-04-30
evmrpc: use BlockID.Hash in GetTransaction (CON-257)

wen-coding:
## Summary

`Backend.GetTransaction` returned the eth `blockHash` by recomputing `block.Block.Header.Hash()`. That's value-equivalent to `BlockID.Hash` under CometBFT (full Header populated by the BlockStore), but **wrong under Autobahn** — `GigaRouter.translateGlobalBlock` returns a ResultBlock with only `ChainID/Height/Time` on the Header, so `Header.Hash()` recomputes a Merkle root that doesn't match any stored value.

## Downstream symptom

go-ethereum's `tracers.API.TraceTransaction` calls `backend.GetTransaction(hash)`, then runs `blockByNumberAndHash(num, blockHash)`. When that helper sees `block.Hash() != blockHash`, it falls back to `BlockByHash(blockHash)` — which under Autobahn returns `ErrBlockNotFoundByHash` because the recomputed sparse-header hash isn't indexed anywhere. `debug_traceTransaction` surfaces `"block not found by hash"` instead of a real trace.

## Fix

Read `blockHash` from `block.BlockID.Hash`. That's the actual block hash the EVM receipt store recorded during `FinalizeBlock`:
- **CometBFT**: equal to `Header.Hash()` — value-equivalent substitution
- **Autobahn**: the real Autobahn block hash that downstream consumers (BlockByHash, EVM receipts) already use

## Verification

- `go test ./evmrpc/...` — 4 packages green (incl. new test, fails without fix)
- Live `debug_traceTransaction` on a 4-node Autobahn cluster now returns proper traces with `callTracer` / `prestateTracer` / `prestateTracer-diffMode` (previously: `block not found by hash`)

### DIFF
--- a/evmrpc/simulate.go
+++ b/evmrpc/simulate.go
@@ -300,7 +300,17 @@ func (b *Backend) GetTransaction(ctx context.Context, txHash common.Hash) (found
 	txIndex := hexutil.Uint(receipt.TransactionIndex)
 	tmTx := block.Block.Txs[txIndex]
 	tx = getEthTxForTxBz(tmTx, b.txConfigProvider(block.Block.Height).TxDecoder())
-	blockHash = common.BytesToHash(block.Block.Header.Hash().Bytes())
+	// Use BlockID.Hash rather than Header.Hash(): under CometBFT they
+	// are equal, but under Autobahn the Block.Header returned by /block
+	// is sparse (the GigaRouter's translateGlobalBlock only populates
+	// ChainID/Height/Time), so Header.Hash() recomputes a Merkle root
+	// that doesn't match any stored value — and downstream
+	// debug_traceTransaction fails with "block not found by hash" when
+	// it tries to round-trip this value through BlockByHash.
+	// BlockID.Hash carries the actual block hash that the EVM receipt
+	// store recorded during FinalizeBlock: same on both engines,
+	// correct under both.
+	blockHash = common.BytesToHash(block.BlockID.Hash)
 	return true, tx, blockHash, uint64(txHeight), uint64(txIndex), nil //nolint:gosec
 }
 
--- a/evmrpc/simulate_test.go
+++ b/evmrpc/simulate_test.go
@@ -22,10 +22,13 @@ import (
 	"github.com/sei-protocol/sei-chain/sei-cosmos/client"
 	sdk "github.com/sei-protocol/sei-chain/sei-cosmos/types"
 	receipt "github.com/sei-protocol/sei-chain/sei-db/ledger_db/receipt"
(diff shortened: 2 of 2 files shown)


evmrpc/simulate.go
```diff
@@ -300,7 +300,17 @@ func (b *Backend) GetTransaction(ctx context.Context, txHash common.Hash) (found
 	txIndex := hexutil.Uint(receipt.TransactionIndex)
 	tmTx := block.Block.Txs[txIndex]
 	tx = getEthTxForTxBz(tmTx, b.txConfigProvider(block.Block.Height).TxDecoder())
-	blockHash = common.BytesToHash(block.Block.Header.Hash().Bytes())
```
bdchatham, line 303: Is this translation only required in this code path or potentially other places too?

wen-coding: I think this is the only place that matters for now.
We also have:
- `evmrpc/subscribe.go` using `header.Header.Hash().String()` on `EventDataNewBlockHeader`
- `evmrpc/filter.go` using `block.Block.Hash()` on `EventDataNewBlock`
These two are both triggered on events, and Autobahn currently don't invoke events. I don't know whether Autobahn will ever need these events. If we do then we can fix them there.

bdchatham (approved):
LGTM

evmrpc/simulate.go
```diff
@@ -300,7 +300,17 @@ func (b *Backend) GetTransaction(ctx context.Context, txHash common.Hash) (found
...
+	// are equal, but under Autobahn the Block.Header returned by /block
+	// is sparse (the GigaRouter's translateGlobalBlock only populates
+	// ChainID/Height/Time), so Header.Hash() recomputes a Merkle root
+	// that doesn't match any stored value — and downstream
+	// debug_traceTransaction fails with "block not found by hash" when
+	// it tries to round-trip this value through BlockByHash.
+	// BlockID.Hash carries the actual block hash that the EVM receipt
+	// store recorded during FinalizeBlock: same on both engines,
+	// correct under both.
+	blockHash = common.BytesToHash(block.BlockID.Hash)
```
pompon0, line 313: are the  block metadata guaranteed to be correct here? We find holes in Tendermint verification every now and then.

pompon0 (approved):

evmrpc/simulate.go
wen-coding: Are you asking about the timing issue? It's correct on our end:
GetTransaction(txHash) does Receipt(txHash) first, so it will only return the hash here if a receipt exists, so the block must have finished execution and the hash should be good here.

jewei1997 (approved):

PR #3352, opened 2026-05-01
config(state-sync): hide tendermint tuning knobs from template

blindchaser:
## Summary

Trim the rendered `[statesync]` section in `sei-tendermint/config/toml.go` so newly generated `config.toml` files show only the operator-facing state sync settings. Hidden settings are still parsed from existing config files, so upgraded nodes keep their current behavior.

- Hidden from newly generated `config.toml`: `backfill-blocks`, `backfill-duration`, `discovery-time`, `temp-dir`, `chunk-request-timeout`, `fetchers`, `verify-light-block-timeout`, and `blacklist-ttl`.
- Kept in newly generated `config.toml`: `enable`, `use-p2p`, `rpc-servers`, `trust-height`, `trust-hash`, `trust-period`, and `use-local-snapshot`.
- Kept in generated `app.toml`: `snapshot-interval`, `snapshot-keep-recent`, and `snapshot-directory`. `snapshot-keep-recent` remains visible because operators commonly tune it for snapshot retention and disk usage.

- `sei-tendermint/config/toml.go`: removes the eight tuning keys above from the rendered template and replaces their old per-key comments with one grouped comment that still names every hidden key.
- `sei-tendermint/config/toml_test.go`: verifies the rendered template no longer contains assignments for the hidden keys.
- `sei-tendermint/config/statesync_compat_test.go`: verifies hidden keys still parse when present in an existing `config.toml`, and verifies a fresh config without those keys uses `DefaultStateSyncConfig()` values and passes `ValidateBasic()`.

## Test plan

- `sei-tendermint/config/toml_test.go`: `checkConfig` rejects rendered assignments for `backfill-blocks`, `backfill-duration`, `discovery-time`, `temp-dir`, `chunk-request-timeout`, `fetchers`, `verify-light-block-timeout`, and `blacklist-ttl`. `configContainsKey` avoids prefix false positives such as `fetchers` matching a longer key.
- `sei-tendermint/config/statesync_compat_test.go`: `TestHiddenStateSyncKnobsStillParseFromExistingConfig` writes all hidden keys into a config file and asserts `commands.ParseConfig` populates the corresponding `StateSyncConfig` fields.
- `sei-tendermint/config/statesync_compat_test.go`: `TestFreshStateSyncConfigValidatesWithoutHiddenKnobs` writes a fresh-style config without hidden keys, asserts omitted fields come from `DefaultStateSyncConfig()`, and asserts `cfg.StateSync.ValidateBasic()` succeeds.
- Validation run locally: `go test ./config/... -run 'TestConfig|TestHiddenStateSync|TestFreshStateSync' -count=1 -timeout 180s` from `sei-tendermint`.


Kbhat1 (approved):

sei-tendermint/config/toml.go
```diff
@@ -458,38 +458,18 @@ trust-hash = "{{ .StateSync.TrustHash }}"
...
-verify-light-block-timeout = "{{ .StateSync.VerifyLightBlockTimeout }}"
-
-blacklist-ttl = "{{ .StateSync.BlacklistTTL }}"
+# Advanced state-sync tuning knobs are intentionally omitted from this template:
+#   - backfill-blocks, backfill-duration: post-sync historical light-block backfill
+#   - discovery-time, temp-dir: snapshot discovery and chunk staging
+#   - chunk-request-timeout, fetchers: chunk transfer behavior
+#   - verify-light-block-timeout, blacklist-ttl: light client verification
+# They are still parsed if set explicitly here; see the Sei docs for the full list
+# and recommended values.
```
masih, line 472: This is good but how the code on the latest version behave _when_ these configs are still present?
This is likely the case for most nodes, where they upgrade the binary but rarely change the config.

masih (approved):
Thank @blindchaser and sorry it took me so long to review this one. 


Left a question that we might need to address before cutting the next release; not a blocker and totally fine to approach in a separate PR if needed 🍻

sei-tendermint/config/toml.go
blindchaser: Backward compatible: only the rendered template hides these keys. The StateSyncConfig struct fields and their mapstructure tags (backfill-blocks, backfill-duration, discovery-time, temp-dir, chunk-request-timeout, fetchers, verify-light-block-timeout, blacklist-ttl) are unchanged, so existing config.toml files continue to parse and honor those values exactly as before. 

Covered by TestHiddenStateSyncKnobsStillParseFromExistingConfig in sei-tendermint/config/statesync_compat_test.go (existing config path) and TestFreshStateSyncConfigValidatesWithoutHiddenKnobs (fresh config falls back to DefaultStateSyncConfig() defaults).

PR #3359, opened 2026-05-01
TraceDB: Add async debug_trace caching

Kbhat1:
- Adds an opt-in TraceDB-backed cache for bakeable `debug_trace*` calls, stored under `<home>/data/trace_db`.
- Runs a background TraceBaker off the consensus path to precompute configured tracers for committed blocks; cache misses still fall back to live tracing.
- Adds bounded workers/queueing, startup catch-up, and optional rolling pruning via `[evm]` config.

- Ran fully on local node
- Verifying on mainnet node
- Unit tests

Screen Shot 2026-05-07 at 12 59 07 PM


jewei1997:
Ran through LLM review and got this

## Major issues

**1. Baker is never stopped — goroutine leak + use-after-close on pebble.** `evmrpc/server.go:95` calls `StartTraceBakerForDebugAPI(...)` and discards the return value. `App.HandleClose` (`app/app.go:1073`) calls `tc.Close()` on the TraceDB but the baker's `Stop()` is never invoked. Workers will continue to call `b.cache.Put`/`PutBlock`/`SetLastBakedHeight` against a closed pebble DB during shutdown — pebble returns errors rather than panicking on closed-DB writes, but they'll be silent (logged at Debug). Worse, `catchUpLoop` and `pruneLoop` continue to run.

  Fix: capture the baker, register it for shutdown, and Stop it *before* closing the TraceDB so workers drain first.

**2. `Enqueue` will panic on send-to-closed-channel if `Stop` is ever wired up.** `Stop()` does `close(b.queue)`, but `Enqueue` selects on `case b.queue <- height:` with no done-channel guard:

```go
select {
case b.queue <- height:
default:
    ...
}
```

If a block commit fires after `Stop`, the send panics. Add a `case <-b.done: return` in Enqueue (or guard with an atomic-bool/RWMutex). Currently latent because Stop is never called (issue 1), but the moment that's fixed this surfaces.

**3. `bakeableTracerName` ignores config fields that change semantics.** It rejects only when `TracerConfig` is non-empty:

```go
if len(config.TracerConfig) > 0 { return "" }
```

But `tracers.TraceConfig` also carries `Reexec`, `Timeout`, `LogConfig` (struct-logger fields), and `EnableMemory`/`EnableReturnData`/`DisableStorage`/`DisableStack` from the embedded `logger.Config`. A caller passing `Reexec: 100` (force re-execution) silently gets a stale cache hit. Tighten the check to also require `Reexec == 0`, `Timeout == nil`, and zero-value LogConfig — or define a canonical tuple that goes into the cache key.

**4. `WindowBlocks: 0` (the default) means pruning is disabled — disk grows unbounded.** Operator footgun: enabling `trace_bake_enabled` without setting a window quietly fills the disk. Either pick a sane non-zero default (e.g. 100k blocks ≈ a few days) or refuse to start when enabled with window=0.
```

**Comment by Kbhat1** [comment]
@jewei1997

Will respond in more detail but don't think any of these are real triggerable bugs. Can clean up but can you take a look at the logic independently and lmk if any thoughts?

**Comment by Kbhat1** [comment]
1. Tendermint is stopped before pebble closes, so the workers can't finish a trace and never reach the Put 
2. Stop() has no callers — the panic path is unreachable. If we ever wire it up we'd add the done-guard at the same time
3. Sei's StateAtBlock ignores Reexec entirely, and logger.Config only applies when Tracer is nil. Same output either way
4. that's the documented behavior, "0 disables prune" is right there in the config comment

Will clean up w/some suggestions but don't seem like blockers

**Review by jewei1997 — APPROVED** [review]

File: evmrpc/trace_baker.go [hunk 1]
```diff
@@ -0,0 +1,374 @@
...
+		return
+	}
+	select {
+	case <-b.done:
+		return
+	case b.queue <- height:
+	default:
+		d := atomic.AddUint64(&b.dropped, 1)
+		if d == 1 || d%256 == 0 {
+			bakerLogger.Info("trace baker queue full; dropping height",
```
codchen, line 147: do we have a recover plan for dropped heights, or it'd just fallback to live tracing on the rpc side?

codchen (approved):

evmrpc/trace_baker.go
Kbhat1: Just fallback to live tracing

PR #3360, opened 2026-05-01
TraceDB: Snapshot-backed state for the trace baker

Kbhat1:
- Adds optional snapshot-backed trace baking so the baker can replay from in-memory memiavl state instead of SS-pebble.
- Refcounts memiavl snapshots and releases trace leases through geth's existing `StateReleaseFunc`, avoiding GC finalizers.
- Opt-in via `trace_bake_use_snapshot`; falls back when the backend cannot provide a snapshot.

- Verified on node
- Unit tests
- Long-running node on mainnet
- `go test ./sei-db/state_db/sc/memiavl -run 'TreeCopy|Snapshot' -count=1`

Screen Shot 2026-05-07 at 12 59 13 PM

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes how `debug_trace*` replays acquire/release historical state and adds new snapshot lifecycle management; while gated behind config and with fallbacks, bugs could cause leaks, stale reads, or trace failures under load.
> 
> **Overview**
> Adds **optional snapshot-backed trace baking**: when `trace_bake_use_snapshot` is enabled and the node is using the `storev2` root multi-store, `EndBlock` captures an in-memory SC snapshot and the trace baker/debug endpoints replay against that snapshot instead of SS-pebble.
> 
> Plumbs a new `TraceContextProvider` through `evmrpc` so debug tracing can obtain a context plus a *release function*, and implements `SnapshotAwareRPCContextProvider` to build contexts from leased snapshots (with consensus params populated) and fall back to the existing RPC context on misses/unsupported backends.
> 
> Extends the state-commit/memiavl layer to support `Committer.Copy()` snapshots and explicit ref-release, including refcounted memiavl snapshot mmaps and new tests to prevent use-after-unmap and verify eviction/lease semantics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4e744cf5ff9e7e619c01112348c00e2a4935bc9a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/evmrpc/server.go
+++ b/evmrpc/server.go
@@ -45,6 +45,7 @@ func NewEVMHTTPServer(
 	homeDir string,
 	stateStore types.StateStore,
 	isPanicOrSyntheticTxFunc func(ctx context.Context, hash common.Hash) (bool, error), // used in *ExcludeTraceFail endpoints
+	traceCtxProviders ...TraceContextProvider,
 ) (EVMServer, error) {
 
 	// Initialize global worker pool with configuration (metrics are embedded in pool)
@@ -89,8 +90,13 @@ func NewEVMHTTPServer(
 	sendAPI := NewSendAPI(tmClient, txConfigProvider, &SendConfig{slow: config.Slow}, k, beginBlockKeepers, ctxProvider, homeDir, simulateConfig, app, antehandler, ConnectionTypeHTTP, globalBlockCache, cacheCreationMutex, watermarks)
 
 	ctx := ctxProvider(LatestCtxHeight)
+	traceCtxProvider := defaultTraceContextProvider(ctxProvider)
+	if len(traceCtxProviders) > 0 && traceCtxProviders[0] != nil {
+		traceCtxProvider = traceCtxProviders[0]
+	}
 	txAPI := NewTransactionAPI(tmClient, k, ctxProvider, txConfigProvider, homeDir, ConnectionTypeHTTP, watermarks, globalBlockCache, cacheCreationMutex)
 	debugAPI := NewDebugAPI(tmClient, k, beginBlockKeepers, ctxProvider, txConfigProvider, simulateConfig, app, antehandler, ConnectionTypeHTTP, config, globalBlockCache, cacheCreationMutex, watermarks)
+	debugAPI.backend.SetTraceContextProvider(traceCtxProvider)
 	if config.TraceBakeEnabled {
 		StartTraceBakerForDebugAPI(debugAPI, TraceBakerConfig{
 			Workers:      config.TraceBakeWorkers,
@@ -109,6 +115,7 @@ func NewEVMHTTPServer(
 
 	seiTxAPI := NewSeiTransactionAPI(tmClient, k, ctxProvider, txConfigProvider, homeDir, ConnectionTypeHTTP, isPanicOrSyntheticTxFunc, watermarks, globalBlockCache, cacheCreationMutex)
 	seiDebugAPI := NewSeiDebugAPI(tmClient, k, beginBlockKeepers, ctxProvider, txConfigProvider, simulateConfig, app, antehandler, ConnectionTypeHTTP, config, globalBlockCache, cacheCreationMutex, watermarks)
+	seiDebugAPI.backend.SetTraceContextProvider(traceCtxProvider)
 
 	// DB semaphore aligned with worker count
 	dbReadSemaphore := make(chan struct{}, workerCount)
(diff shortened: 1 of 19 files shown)


codchen (approved):

evmrpc/server.go
```diff
@@ -45,6 +45,7 @@ func NewEVMHTTPServer(
 	homeDir string,
 	stateStore types.StateStore,
 	isPanicOrSyntheticTxFunc func(ctx context.Context, hash common.Hash) (bool, error), // used in *ExcludeTraceFail endpoints
+	traceCtxProviders ...TraceContextProvider,
```
blindchaser, line 48: only the first element is ever read. A non-variadic *TraceContextProvider parameter (or a small options struct) avoids the "what if someone passes two" ambiguity and keeps the signature self-documenting.

x/evm/keeper/trace_snapshot.go
```diff
@@ -0,0 +1,123 @@
...
+	if s == nil {
+		return nil
+	}
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.snapshots[height]
+}
+
+// Close releases all retained snapshots.
+func (s *TraceSnapshotStore) Close() {
```
blindchaser, line 99: TraceSnapshotStore.Close() returns nothing, but inside it ignores per-snapshot release errors:

x/evm/keeper/trace_snapshot.go:97-112

_ = releaser.ReleaseSnapshotRefs()

Same as the WARN above: refcount mismatches are real bugs. Either return error or log at WARN level on close to keep ops visibility.

x/evm/keeper/abci.go
```diff
@@ -60,9 +60,16 @@ func (k *Keeper) BeginBlock(ctx sdk.Context) {
 
 func (k *Keeper) EndBlock(ctx sdk.Context, height int64, blockGasUsed int64) {
 	defer telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), telemetry.MetricKeyEndBlocker)
-	// Bake height-1: at EndBlock(N) the indexer's safe latest is N-1, so
-	// N-1 is the most recent block guaranteed to be queryable.
+	// Bake height-1: at EndBlock(N) the indexer's safe latest is N-1. When
```
blindchaser, line 63: EndBlock snapshot semantics are subtle — comment is dense, easy to mis-read

The off-by-one here is correct but non-obvious: storev2/rootmulti.flush() doesn't run until Commit(), so at EndBlock(N) the SC tree state is state_after_commit_of_(N-1) and snap.Version() == N-1. The baker then traces H=N-1, whose initializeBlock calls ctxProvider(H-1) = ctxProvider(N-2), which finds snap[N-2] Put at the previous EndBlock(N-1). Worth a one-line "lined up because rs.flush is called from Commit, not from EndBlock" in the comment to save the next reader a half-hour.

Also: initializeBlock calls the provider twice — once for prevBlockHeight (H-1) and once for blockNumber (H) (for WithNextMs). That means a single trace leases both snap[H-1] and snap[H]. As long as TraceBakeSnapshotWindow >= 2 this is fine, but if an operator misconfigures window=1 the second lease will miss and silently fall through to SS-pebble for oracle_mem/WithNextMs. Consider clamping window to >= 2 (or whatever the documented minimum is) at config-load time.

sei-db/state_db/sc/memiavl/db.go
```diff
@@ -921,6 +930,9 @@ func (db *DB) rewriteSnapshotBackground() error {
 	cloned := db.copy()
 	go func() {
 		defer close(ch)
+		// Release per-tree snapshot refs; don't call cloned.Close() which
+		// would also tear down the live db's writer pool and stream handler.
+		defer func() { _ = cloned.MultiTree.Close() }()
```
blindchaser: _ = will silently hide the new "Snapshot over-close" path. Since over-close is a real bug (it means refcounts are unbalanced), at least:


defer func() {
    if err := cloned.MultiTree.Close(); err != nil {
        logger.Error("failed to release cloned snapshot refs after rewrite", "err", err)
    }
}()

blindchaser (approved):

evmrpc/server.go
Kbhat1: keeping the variadic override for minimal PR scope since only one provider is intentionally supported

x/evm/keeper/trace_snapshot.go
Kbhat1: Updated, release/close errors are now logged at WARN instead of being swallowed

x/evm/keeper/abci.go
Kbhat1: Fixed the root cause by making snapshot window >= 2. Leaving the comment wording as is for now

sei-db/state_db/sc/memiavl/db.go
Kbhat1: Updated

PR #3375, opened 2026-05-04
autobahn: add data_prune_after to bound data.State memory (CON-256)

wen-coding:
data.State.runPruning is a background goroutine that drops in-memory blocks/QCs/AppProposals older than a configurable duration, but the config knob (data.Config.PruneAfter) was never wired up — giga_router constructed data.NewState with only Committee set, so the pruner never spawned. data.State.PruneBefore (the giga_router-driven path based on cosmos-sdk RetainHeight) is also a no-op when the chain is configured with pruning="nothing" (the localnode default, common in test setups), so in-memory data.State grew with the chain under sustained load and eventually OOM-killed nodes.

Plumb DataPruneAfter through:
\`AutobahnFileConfig.data_prune_after\` (json) → \`GigaRouterConfig.DataPruneAfter\` → \`data.Config.PruneAfter\` → \`data.State.runPruning\`.

Production default (gen-autobahn-config): 30m, gives operators plenty of recent history for /block, /tx, /trace_*, etc. while bounding memory under load. Localnode/test override (step4_config_override.sh): 1m, keeps data.State small under sustained-throughput tests where cosmos pruning is "nothing".

## Things done

- [x] AutobahnFileConfig.DataPruneAfter
- [x] GigaRouterConfig.DataPruneAfter + thread into data.Config
- [x] node/setup.go pass-through from autobahn.json
- [x] gen-autobahn-config production default (30m)
- [x] step4_config_override.sh localnode override (1m)
- [x] gofmt, vet clean


pompon0:
fyi, pruneAfter in data was used in sei-v3, but in sei-chain it is application which is responsible for pruning, via retainHeight field in ResponseCommit. I currently don't know if we want to change ownership of pruning from application to consensus. IMO it would make sense, given that application should rather be solely concerned with the latest state at all times. However perhaps sei-chain app makes some assumptions wrt which blocks are available (I can imagine that it does, but I haven't looked into that yet).

wen-coding:
> fyi, pruneAfter in data was used in sei-v3, but in sei-chain it is application which is responsible for pruning, via retainHeight field in ResponseCommit. I currently don't know if we want to change ownership of pruning from application to consensus. IMO it would make sense, given that application should rather be solely concerned with the latest state at all times. However perhaps sei-chain app makes some assumptions wrt which blocks are available (I can imagine that it does, but I haven't looked into that yet).

Sorry I'm confused. Not planning to change prune ownership in this PR (although we can discuss whether that should be done, I'm generally of the opinion this is consensus cleanup which should probably be controlled via consensus), I just want to set a smaller prune period in tests, so that in less powerful machines (my Mac) we can still run long throughput tests without the validators getting OOM. The 30m default in gen-autobahn-config is a defensive cap (still opt-out — operators can drop the field), not a replacement for app-driven pruning.

pompon0:
Currently pruning is driven by retainHeight computed via https://github.com/sei-protocol/sei-chain/blob/52d368fb0f9ae8a14c17cfbd9ce621666d2fe52e/sei-cosmos/baseapp/abci.go#L680 . If pruning in tests does not work, we should first check whether this function actually advances retainHeight and fix it if it doesn't (or if it doesn't keep up with the block rate).

PR #3380, opened 2026-05-04
perf(ProcessProposal): skip IsTxGasless in checkTotalBlockGas for non-gasless txs

amir-deris:
## Summary

`checkTotalBlockGas` previously called `antedecorators.IsTxGasless` unconditionally for every tx before doing EVM/Cosmos gas accounting. For typical `MsgEVMTransaction` txs, `IsTxGasless` always returned `(false, nil)` immediately via the default branch — but still paid the cost of a `defer`/`recover` frame and a message-type walk on every tx.

This PR mirrors the `couldBeGaslessTransaction` heuristic already used in `DeliverTxWithResult` and OCC paths:

1. **Fast-path guard**: `IsEVMMessage` is hoisted to the top of the loop (it was already called later for gas accounting). For confirmed single-message EVM txs (`isEVM && evmErr == nil`), `IsTxGasless` is skipped entirely — `MsgEVMTransaction` can never be gasless under the current rules.
2. **Heuristic guard**: for all other txs, `couldBeGaslessTransaction` is checked first. Only oracle vote and `MsgAssociate` txs proceed to the full keeper-backed `IsTxGasless` call.

Semantics are unchanged: txs for which `IsTxGasless` would have returned `(false, nil)` are now classified as non-gasless without the extra overhead.


arajasek (approved):

app/app.go
codchen, line 2459: another optimization we can make is for this loop to run concurrently and add to total block gas as an atomic int. The only advantage of sequential is if the loop early returns, which is rare.

codchen (approved):

app/app.go
amir-deris: Thanks for feedback. Created a follow up ticket: https://linear.app/seilabs/issue/PLT-324/

masih (approved):

PR #3382, opened 2026-05-04
implement CommitKVStore using a migration router

cody-littley:

Implement the `CommitKVStore` interface using a `migration.Router`. This will allow us to implement the composite DB's `GetChildStoreByName()` method.

Note that this is not a permanent solution. The long term solution will be to remove the `GetChildStoreByName()` api entirely. But that can be an incremental cleanup, and does not block MVP deployment.

unit tests

### DIFF
--- a/sei-db/state_db/sc/migration/router_kvstore.go
+++ b/sei-db/state_db/sc/migration/router_kvstore.go
@@ -45,7 +45,7 @@ func NewRouterCommitKVStore(
 // Close is illegal during the standard CommitKVStore lifecycle for this type:
 // the wrapped Router is owned by the caller and must outlive this view.
 func (r *RouterCommitKVStore) Close() error {
-	panic("RouterCommitKVStore.Close: illegal during standard lifecycle")
+	return fmt.Errorf("RouterCommitKVStore.Close: illegal during standard lifecycle")
 }
 
 func (r *RouterCommitKVStore) Get(key []byte) []byte {
--- a/sei-db/state_db/sc/migration/router_kvstore.go
+++ b/sei-db/state_db/sc/migration/router_kvstore.go
@@ -1,0 +1,115 @@
+package migration
+
+import (
+	"context"
+	"fmt"
+
+	ics23 "github.com/confio/ics23/go"
+	"github.com/sei-protocol/sei-chain/sei-db/proto"
+	"github.com/sei-protocol/sei-chain/sei-db/state_db/sc/types"
+	db "github.com/tendermint/tm-db"
+)
+
+// rootHashSize matches the digest length used by the other CommitKVStore
+// implementations in this codebase: memiavl returns sha256 (32 B) and flatkv
+// returns Blake3-256 (32 B).
+const rootHashSize = 32
+
+var _ types.CommitKVStore = (*RouterCommitKVStore)(nil)
+
+// RouterCommitKVStore adapts a [Router] (which is keyed by store name on every
+// call) to the store-name-less [types.CommitKVStore] interface by binding the
+// view to a single module store name.
+//
+// The CommitKVStore interface does not return errors. Any error returned by the
+// underlying router is therefore surfaced as a panic. This is a short-term
+// limitation; the long-term plan is to plumb errors through the interface.
+type RouterCommitKVStore struct {
+	router          Router
+	storeName       string
+	versionProvider func() int64
+}
+
+func NewRouterCommitKVStore(
+	router Router,
+	storeName string,
+	versionProvider func() int64,
+) *RouterCommitKVStore {
+	return &RouterCommitKVStore{
+		router:          router,
+		storeName:       storeName,
+		versionProvider: versionProvider,
+	}
+}
+
+// Close is illegal during the standard CommitKVStore lifecycle for this type:
+// the wrapped Router is owned by the caller and must outlive this view.
+func (r *RouterCommitKVStore) Close() error {
+	panic("RouterCommitKVStore.Close: illegal during standard lifecycle")
+}
+
+func (r *RouterCommitKVStore) Get(key []byte) []byte {
+	value, _, err := r.router.Read(r.storeName, key)
+	if err != nil {
+		panic(fmt.Errorf("RouterCommitKVStore.Get(store=%q): %w", r.storeName, err))
+	}
+	return value
+}
+
+func (r *RouterCommitKVStore) Has(key []byte) bool {
+	_, found, err := r.router.Read(r.storeName, key)
+	if err != nil {
+		panic(fmt.Errorf("RouterCommitKVStore.Has(store=%q): %w", r.storeName, err))
+	}
+	return found
+}
+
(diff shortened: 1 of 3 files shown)


sei-db/state_db/sc/migration/router_kvstore.go
```diff
@@ -0,0 +1,115 @@
...
+		router:          router,
+		storeName:       storeName,
+		versionProvider: versionProvider,
+	}
+}
+
+// Close is illegal during the standard CommitKVStore lifecycle for this type:
+// the wrapped Router is owned by the caller and must outlive this view.
+func (r *RouterCommitKVStore) Close() error {
+	return fmt.Errorf("RouterCommitKVStore.Close: illegal during standard lifecycle")
```
blindchaser, line 48: should we panic instead of error here to match testing?

sei-db/state_db/sc/migration/router_kvstore.go
```diff
@@ -0,0 +1,115 @@
...
+	router          Router
+	storeName       string
+	versionProvider func() int64
+}
+
+func NewRouterCommitKVStore(
+	router Router,
+	storeName string,
+	versionProvider func() int64,
+) *RouterCommitKVStore {
```
blindchaser, line 37: context.Background() is hardcoded, do we need any outer ctx for per-Set cancellation? not sure

sei-db/state_db/sc/migration/router_kvstore.go
```diff
@@ -0,0 +1,115 @@
...
+func (r *RouterCommitKVStore) Has(key []byte) bool {
+	_, found, err := r.router.Read(r.storeName, key)
+	if err != nil {
+		panic(fmt.Errorf("RouterCommitKVStore.Has(store=%q): %w", r.storeName, err))
+	}
+	return found
+}
+
+func (r *RouterCommitKVStore) Set(key []byte, value []byte) {
+	r.applyOne(&proto.KVPair{Key: key, Value: value})
```
blindchaser, line 68: Set/Remove issues one ApplyChangeSets per call, In the prev PR, MigrationManager.ApplyChangeSets advances exactly one migration batch per call. A single cosmos-sdk transaction can issue dozens of Set calls; that becomes dozens of migration batches per commit

sei-db/state_db/sc/migration/router_kvstore.go
```diff
@@ -0,0 +1,115 @@
...
+}
```
blindchaser, line 69: Sei KVStore impls (storev2/commitment/Store.Set, cachekv.Set) only append to an in-memory changeset; the real write happens at Commit. 

how about RouterCommitKVStore buffers writes and flushes once at commit time (needs a Commit/Flush hook, mirroring storev2/commitment/Store).

sei-db/state_db/sc/migration/router_kvstore.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (r *RouterCommitKVStore) Has(key []byte) bool {
+	_, found, err := r.router.Read(r.storeName, key)
+	if err != nil {
+		panic(fmt.Errorf("RouterCommitKVStore.Has(store=%q): %w", r.storeName, err))
+	}
+	return found
+}
+
+func (r *RouterCommitKVStore) Set(key []byte, value []byte) {
```
blindchaser, line 67: how do we ensure this actually works instead of baseapp wraps runTx in recover() and turns the panic into a per-tx error, then keeps processing

`// If this method returns an error, it is not safe to attempt to retry. An error should be considered
	// fatal, and should result in any managed databases being shut down and crash recovered.`

sei-db/state_db/sc/migration/router_kvstore.go
cody-littley: Test was incorrect, and actually failing. Fixed the test. IMO, better to return error than to panic, especially when the interface already has an error return value.

sei-db/state_db/sc/migration/router_kvstore.go
cody-littley: In the main feature branch where I've got all of this code integrated, I actually removed the goroutines from the routers (race conditions + premature optimization 😅). Would you be ok with me addressing that issue in the feature branch instead of here? I can make the next follow up PR be the one that contains those changes.

sei-db/state_db/sc/migration/router_kvstore.go
cody-littley: As far as I can tell, there are no callers to `Set()` or `Delete()` anywhere in production code, only in unit tests.

The goal of `router_kvstore.go` was to implement the current interface in a way that didn't require a huge refactor. My plan after that is to incrementally remove methods (like these ones) one at a time, until the entire CommitKVStore interface can be fully deleted.

I added the following documentation to telegraph our intentions:

```
// Planed for future deprecation but not yet deprecated.
type CommitKVStore interface {
	// Planed for future deprecation but not yet deprecated.
	// This is the proper method to call to get a value from the store.
	Get(key []byte) []byte

	// Planed for future deprecation but not yet deprecated. This is the proper method to call to check
	// if a key exists in the store.
	Has(key []byte) bool

	// Deprected: do not call in production code, use CommitKVStore.ApplyChangeSets() instead.
	Set(key, value []byte)

	// Deprected: do not call in production code, use CommitKVStore.ApplyChangeSets() instead.
	Remove(key []byte)

	// Deprected but safe to call
	Version() int64

	// Deprected: do not call in production code, may panic on stores backed by flatKV
	RootHash() []byte

	// Partially deprecated: may panic if called on a store that does not support iteration (e.g. evm/ after migration)
	Iterator(start, end []byte, ascending bool) dbm.Iterator

	// Partially deprecated: may panic if called on a store that does not support proofs (e.g. evm/ after migration)
	GetProof(key []byte) *ics23.CommitmentProof

	// deprecated: some implementations always return errors, and ones that don't mean you are closing something
	// that shouldn't be closed directly
	io.Closer
}

```

sei-db/state_db/sc/migration/router_kvstore.go
cody-littley: I think this topic is probably worth a discussion. Currently FlatKV also only writes to an in-memory buffer.

This `Set()` method isn't used in production, only unit tests. And the plan is to delete it. My current goal is to make unit tests pass, and I'm less concerned with performance implications.

sei-db/state_db/sc/migration/router_kvstore.go
cody-littley: We should discuss this with execution folk. They should NEVER treat a DB error/panic as a transaction failure, as continuing after an error probably means data corruption and a later app hash break.

blindchaser (approved):

sei-db/state_db/sc/migration/router_kvstore.go
Kbhat1: Could you add me to this discussion?

Kbhat1 (approved):

PR #3383, opened 2026-05-04
fix(giga): write receipt for state-transition errors that bump the nonce (CON-256)

wen-coding:
## Invariant

A receipt is written for an EVM tx **iff that tx bumped the sender's nonce.** Equivalently: a tx is observable via `eth_getTransactionByHash` / `eth_getTransactionReceipt` exactly when the chain treats it as having advanced the sender's account state. Txs that the chain rejected before the nonce bump leave no receipt and are invisible to standard EVM clients.

This PR fixes a place where the Giga executor violated the invariant: a state-transition error inside `Execute()` (notably EIP-7623 floor-data-gas underflow, post-Pectra) bumped the sender's nonce but produced no receipt, so `eth_getTransactionReceipt` returned null forever and clients that polled for it hung. The invariant is now enforced symmetrically across V2 and Giga.

## How the invariant is enforced

**V2 path (no change in this PR — already correct):**
1. `BasicDecorator.AnteHandle` registers a `WithDeliverTxCallback` (`x/evm/ante/basic.go:33-39`) that fires after `DeliverTx` regardless of success/failure (and regardless of whether `msgServer.EVMTransaction` returns err). The callback bumps the sender's nonce and calls `SetNonceBumped`.
2. After `msgServer.EVMTransaction` returns err, the SDK populates `ExecTxResult.Code != 0` and `ExecTxResult.Log = err.Error()`.
3. At EndBlock, `GetAllEVMTxDeferredInfo` (`x/evm/keeper/deferred.go:24`) synthesizes a `DeferredInfo{TxHash, Error: txRes.Log}` for each tx that didn't append its own.
4. The EndBlock loop (`x/evm/keeper/abci.go:100-113`) writes a synthetic receipt for entries with `Error != ""` — gated on `GetNonceBumped`, which is exactly the invariant. Existing coverage: `TestEndBlock_NoReceiptForNonceMismatch` and `TestEndBlock_ReceiptCreatedWhenNonceBumped` in `x/evm/keeper/abci_test.go`.

**Giga path (fixed in this PR):**
1. `executeEVMTxWithGigaExecutor` already explicitly bumps the sender's nonce in its `execErr != nil` branch (`app/app.go`).
2. It also calls `AppendToEvmTxDeferredInfo` for surplus tracking — but with `Error == ""`, so the EndBlock synthetic-receipt loop above skips it.
3. **Fix:** call `WriteReceipt` directly in the err-branch, gated on the explicit nonce bump it already does. EIP-1559 effective gas price is used (not `GasFeeCap`) so the receipt's `EffectiveGasPrice` matches go-ethereum semantics for dynamic-fee txs.
4. **Edge case:** if `stateDB.Finalize` fails (practically impossible — Finalize early-returns without flushing on internal err), the prior nonce bump may not have been persisted. Per the invariant we cannot claim the tx happened, so the err-branch early-returns without writing a receipt or appending deferred info.

## Tests

- `TestGiga_FailedExecution_ProducesReceipt` (`giga/tests/giga_test.go`) drives the EIP-7623 floor-data-gas-underflow case through the Giga executor, asserts the explicit nonce bump (`nonceBefore+1 == nonceAfter`) to lock in the invariant, and asserts the receipt was written with `status=0`, `gasUsed=gasLimit`, and the floor-data-gas error in `VmError`.

## Things done

- [x] Giga path (`app/app.go`): write receipt in `execErr != nil` branch using EIP-1559 effective gas price; skip on `Finalize` failure
- [x] Giga test: nonce-bump assertion + receipt assertion
- [x] `gofmt -s` clean


giga/tests/giga_test.go
```diff
@@ -1646,6 +1646,86 @@ func TestGiga_FeeValidationOrder_WrongNonce_NoNonceBump(t *testing.T) {
...
+// TestGiga_FailedExecution_ProducesReceipt asserts that an EVM tx executed via the
+// Giga path which fails with a state-transition error inside go-ethereum's Execute()
+// — Execute() returns err before any opcode runs (notably EIP-7623 floor-data-gas
+// insufficient, which post-Pectra fires in normal operation) — still has a status=0
+// receipt written to the transient receipt store with gasUsed=gasLimit.
+//
+// Without the fix in app.go's executeEVMTxWithGigaExecutor, the receipt was dropped
+// for these "should not happen" cases, so eth_getTransactionReceipt returned null
+// forever for an included tx, hanging any client that polls. This is the Giga-path
+// counterpart to TestEVMTransactionStateTransitionErrorProducesReceipt in
```
arajasek: nit: where is `TestEVMTransactionStateTransitionErrorProducesReceipt`? i don't see it in the file mentioned.

arajasek (approved):

giga/tests/giga_test.go
wen-coding: Good point, changed.

app/app.go
codchen, line 1897: I don't think we want to create a receipt under this scenario, although it is practically impossible (or at least not expected) to have an error here.

codchen (approved):

app/app.go
wen-coding: That's a good point, I'm adding an early return here so it doesn't flow through

PR #3384, opened 2026-05-05
giga: report EIP-1559 effective gas price on receipt (CON-256)

wen-coding:
GIGA's `executeEVMTxWithGigaExecutor` was passing `ethTx.GasPrice()` into the `core.Message` used to construct the receipt. For dynamic-fee (EIP-1559) txs `Transaction.GasPrice()` returns `GasFeeCap`, so the receipt wrongly reported `maxFee` for any tx that actually paid `baseFee + tip < maxFee`.

This breaks EIP-1559 RPC semantics for clients (ethers, hardhat-ethers, etc.) that read `receipt.effectiveGasPrice` to compute spend / display per-tx cost.

## What's actually wrong

V2's `x/evm/keeper/msg_server.go` constructs the message via `server.GetEVMMessage(ctx, tx, sender)`, which has the EIP-1559 adjustment built in:

```go
baseFee := k.GetBaseFee(ctx)
if baseFee != nil {
    msg.GasPrice = msg.GasPrice.Add(msg.GasTipCap, baseFee)
    if msg.GasPrice.Cmp(msg.GasFeeCap) > 0 {
        msg.GasPrice = msg.GasFeeCap
    }
}
```

GIGA's path hand-rolls a `core.Message` literal in `executeEVMTxWithGigaExecutor` and missed this step.

## The fix

Use the already-computed `effectiveGasPrice` (line 1866 of `app/app.go`, which already does `min(baseFee + tipCap, feeCap)` and is what the chain actually charges via `stateDB.SubBalance` at line 1871) when constructing the receipt's `core.Message`.

The chain's actual fee charge has always been correct — only the value reported on the receipt was wrong.

## Things done

- [x] Single-line code change in `app/app.go executeEVMTxWithGigaExecutor`: `GasPrice: ethTx.GasPrice()` → `GasPrice: effectiveGasPrice`
- [x] Unit test in `giga/deps/xevm/keeper/receipt_test.go` documents the `WriteReceipt` contract: `receipt.EffectiveGasPrice == msg.GasPrice` (caller is responsible for passing the EIP-1559 effective price)
- [x] `gofmt -s` clean

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1923,12 +1923,16 @@ func (app *App) executeEVMTxWithGigaExecutor(ctx sdk.Context, msg *evmtypes.MsgE
 		vmError = execResult.Err.Error()
 	}
 
-	// Create core.Message from ethTx for WriteReceipt
-	// WriteReceipt needs msg for GasPrice, To, From, Data, Nonce fields
+	// Create core.Message from ethTx for WriteReceipt.
+	// GasPrice must be the EIP-1559 effective gas price (min(baseFee+tip,
+	// maxFee)) — that's what the chain actually charges (see line 1866)
+	// and what the receipt's EffectiveGasPrice field needs to report.
+	// ethTx.GasPrice() returns GasFeeCap for dynamic-fee txs, which puts
+	// the wrong value on the receipt and breaks EIP-1559 clients.
 	evmMsg := &core.Message{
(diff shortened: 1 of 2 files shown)


masih (approved):

arajasek (approved):
Thanks!

PR #3395, opened 2026-05-05
litt sharding simple sharding

cody-littley:

This PR simplifies how littDB assignes values to shards. The net result of this change is that we don't need the siphash library any more. Previously, LittDB needed this library for shard assignment.

Benchmarked this particular schema when evaluating LittDB performance for block storage.

---

> [!NOTE]
> **High Risk**
> High risk because it changes LittDB’s on-disk formats (segment metadata, key files, and keymap Address serialization) and the shard selection logic, which can affect data compatibility and read/write correctness across restarts.
> 
> **Overview**
> This PR **replaces key-hash-based sharding (with per-segment salt)** with **round-robin shard assignment at write time**, and makes reads fully **address-driven** by embedding `shardID` (and `valueSize`) into the on-disk `types.Address`.
> 
> It updates segment/key-file/metadata serialization to a new `LatestSegmentVersion` (dropping legacy versions and salt fields), tightens sharding-factor validation (now capped at `MaxShardingFactor = 256`), removes the siphash dependency, and adjusts/extends tests and docs to match the new wire formats and shard-selection behavior (including out-of-range shardID handling).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9d2dd25e37f888a03c97814de01a5550e334c6e9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Kbhat1 (approved):

sei-db/db_engine/litt/README.md
```diff
@@ -233,18 +233,13 @@ the [value](#value) associated with a [key](#key) can be retrieved from disk.
 An address is encoded in a 64-bit integer. It contains two pieces of information:
```
blindchaser: from ai: The implementation now serializes a 13-byte address, not a 64-bit integer, and the offset points at the value length prefix. Update this before merging.

sei-db/db_engine/litt/types/address.go
```diff
@@ -7,41 +7,83 @@ import (
...
+	return Address{
+		index:     index,
+		offset:    offset,
+		shardID:   shardID,
+		valueSize: valueSize,
+	}
+}
+
+// DeserializeAddress converts a byte slice to an Address. The slice must be exactly AddressSerializedSize bytes.
 func DeserializeAddress(bytes []byte) (Address, error) {
```
blindchaser, line 44: Issue: DeserializeAddress accepts any byte as ShardID, but these paths index s.shards without validating it is < len(s.shards). A bad keymap entry or damaged key file can turn a read/restart into a runtime panic.

blindchaser (approved):

sei-db/db_engine/litt/README.md
cody-littley: Fixed wording:

```
## Address

An address partially describes the location on disk where a [value](#value) is stored. Together with a [key](#key),
the [value](#value) associated with a [key](#key) can be retrieved from disk.

An address contains the following information:

- the [segment](#segment) [index](#segment-index) where the [value](#value) is stored
- the [shard](#shard) within that segment that holds the [value](#value)
- the offset within the [value file](#segment-value-files) where the first byte of
  the [value](#value) is stored
- the length of the [value](#value) in bytes

Retrieving a [value](#value) starting from an Address is a self-contained
operation that does not need to consult any segment-level metadata or recompute anything from the [key](#key).
```

sei-db/db_engine/litt/types/address.go
cody-littley: added defensive checks

PR #3396, opened 2026-05-05
feat(app): migrate app and ante telemetry to OpenTelemetry Meter API

amir-deris:
## Summary

Migrates telemetry in the `app` and `app/ante` packages from the legacy `telemetry`/`utilmetrics` helpers to the standardized OpenTelemetry Meter API, following the same pattern established in #3265 for `evmrpc`.

- Adds `app/metrics.go` with a struct-based OTel instrument set registered via `otel.Meter("app")`, initialized once in `NewApp` after `SetupOtelMetricsProvider()`
- Adds `app/ante/metrics.go` with a lazily-initialized `anteMetrics` struct via `sync.OnceValue` (fires on first CheckTx, after the global MeterProvider is set)
- Replaces `telemetry.MeasureSince` / `utilmetrics` calls in `abci.go` and `invariance.go` with dual-emitted OTel instruments; legacy calls remain with `TODO(PLT-327)` markers until dashboards are migrated
- Threads `ctx`/`ctx.Context()` through all `.Record()` and `.Add()` call sites to support OTel's context-based propagation
- Replaces the per-BeginBlock `GaugeSeidVersionAndCommit` call with an `app_build_info` observable gauge that fires on Prometheus scrape — no per-block overhead

## New metrics (OTel naming convention, exported via the process-wide MeterProvider)

**ABCI phase durations** — histograms bucketed at SLO thresholds (p50 ≤ 500ms, p95 ≤ 1.5s, p99 ≤ 2.5s):
- `app_abci_begin_block_duration_seconds`
- `app_abci_end_block_duration_seconds`
- `app_abci_module_end_block_duration_seconds`
- `app_abci_check_tx_duration_seconds`
- `app_abci_deliver_tx_duration_seconds`
- `app_abci_deliver_batch_tx_duration_seconds`
- `app_block_process_duration_seconds` — block tx processing duration by execution type

**Transaction counters and gas:**
- `app_tx_count_total`
- `app_tx_process_type_total` — by execution type label
- `app_tx_gas_total` — cumulative gas by type label
- `app_tx_gas_used` / `app_tx_gas_wanted` — per-tx gauges

**App-level flow counters:**
- `app_optimistic_processing_total` — cache hit (`enabled=true`) vs miss
- `app_failed_total_gas_wanted_check_total`
- `app_giga_fallback_to_v2_total`

**Light invariance:**
- `app_lightinvariance_supply_duration_seconds`
- `app_lightinvariance_supply_invalid_key_total`
- `app_lightinvariance_supply_unmarshal_failure_total`

**Build info:**
- `app_build_info` — observable gauge, always 1, labels: `seid_version`, `commit`

**Ante:**
- `app_pending_nonce_total` — pending nonce events by type (added, expired, rejected, accepted)

## Migration note

Legacy metrics (`telemetry.MeasureSince`, `utilmetrics.MeasureDeliverTxDuration`, etc.) are dual-emitted during the migration window so existing dashboards continue to receive data. Each legacy call site is annotated `TODO(PLT-327)` and will be removed once dashboards are verified against the new OTel series.

---

> [!NOTE]
> **Medium Risk**
> Touches core ABCI/block-processing and invariance paths to add dual-emitted OTel metric recording; main risk is performance overhead or incorrect metric/context wiring in consensus-critical code paths, though logic is largely additive.
> 
> **Overview**
> Adds a new OTel `meter` instrument set in `app/metrics.go` and `app/ante/metrics.go` and wires it into runtime so the app emits standardized OpenTelemetry counters/histograms/gauges.
> 
> Updates ABCI/block lifecycle code (`BeginBlock`/`EndBlock`/`CheckTx`/`DeliverTx`/`DeliverTxBatch`/`Commit`) plus tx execution flows (sync/OCC/giga, optimistic processing, gas-wanted rejection) and light invariance checks to **record OTel metrics alongside existing legacy `telemetry`/`utilmetrics` calls**, with `TODO(PLT-327)` markers for later removal.
> 
> Refactors `LightInvarianceChecks` to accept a `context.Context` for OTel recording, updates tests accordingly, and renames execution-type label constants in `utils/metrics/labels.go` (e.g., adds `SynchronousGiga`, `OccGiga`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 873f9ec18e36fce99507cf00bd45b8cba27de98e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


app/metrics.go
```diff
@@ -0,0 +1,199 @@
...
+
+	appMetrics.blockProcessDuration = must(meter.Float64Histogram(
+		"app_block_process_duration_seconds",
+		metric.WithDescription("Duration of block tx processing by execution type"),
+		metric.WithUnit("s"),
+		histogramBuckets,
+	))
+
+	appMetrics.txCount = must(meter.Int64Counter(
+		"app_tx_count_total",
```
bdchatham: **[blocker]** The Prometheus exporter appends `_total` to OTel counters automatically; this instrument will export as `app_tx_count_total_total`. Same hits every `_total`-suffixed counter here (`app_tx_process_type_total`, `app_tx_gas_total`, `app_optimistic_processing_total`, `app_failed_total_gas_wanted_check_total`, `app_giga_fallback_to_v2_total`, `app_lightinvariance_supply_*_total`) and the histograms — `_seconds` + `WithUnit("s")` → `*_seconds_seconds`.

Drop the `_total` suffix from counter names and `_seconds` from histogram names; the exporter adds them. Verify against `/metrics` from a node running this build before any dashboard work.

app/metrics.go
```diff
@@ -0,0 +1,199 @@
...
+
+	"github.com/sei-protocol/sei-chain/sei-cosmos/version"
+)
+
+const appMeterName = "app"
+
+// histogramBuckets aligns with block-time SLO thresholds
+// (p50 ≤ 500ms, p95 ≤ 1.5s, p99 ≤ 2.5s) expressed in seconds.
+var histogramBuckets = metric.WithExplicitBucketBoundaries(
+	0.025, 0.05, 0.1, 0.25, 0.5, 0.75, 1.0, 1.5, 2.0, 2.5, 5.0, 10.0,
```
bdchatham: **[should-fix]** Buckets jump 2.5 → 5.0; with the SLO at p99=2.5s, any p99 between 2.5s and 5s reports inside `[2.5, 5.0]` with up to ~2.5s of `histogram_quantile` interpolation error. Add `3.0` (ideally `4.0`) to tighten resolution at the threshold.

Separately: this set is reused for `commitDuration` and `invarianceDuration`. Unless they all really share the same proper bucket sizing, consider separating these.

app/metrics.go
```diff
@@ -0,0 +1,199 @@
...
+		"app_tx_process_type_total",
+		metric.WithDescription("Transactions processed by execution type"),
+	))
+
+	appMetrics.txGas = must(meter.Int64Counter(
+		"app_tx_gas_total",
+		metric.WithDescription("Cumulative transaction gas by type (gas_used, gas_wanted)"),
+	))
+
+	appMetrics.txGasUsed = must(meter.Int64Gauge(
```
bdchatham: **[blocker]** Synchronous `Int64Gauge` is last-write-wins per scrape. Under OCC concurrent execution (`ProcessTXsWithOCCV2`), thousands of `DeliverTx` per 15-30s scrape window means the published value is "the gas of one arbitrary tx" — not aggregable, not meaningful. `txGasWanted` (below) has the same issue.

`app_tx_gas_total` (lines 126-129) already covers cumulative gas, and the `_count` series of `app_abci_deliver_tx_duration_seconds` covers tx rate. If you want a per-tx gas distribution, convert these to `Float64Histogram`. Otherwise drop both — they're carrying forward the `telemetry.SetGauge` anti-pattern.

app/abci.go
```diff
@@ -121,22 +140,30 @@ func (app *App) CheckTx(ctx context.Context, req *abci.RequestCheckTxV2) *abci.R
...
-		telemetry.IncrCounter(1, "tx", "count")
-		telemetry.IncrCounter(1, "tx", resultStr)
-		telemetry.SetGauge(float32(gInfo.GasUsed), "tx", "gas", "used")
-		telemetry.SetGauge(float32(gInfo.GasWanted), "tx", "gas", "wanted")
+		telemetry.IncrCounter(1, "tx", "count")                             // TODO(PLT-327): remove once app_tx_count_total verified
+		telemetry.IncrCounter(1, "tx", resultStr)                           // TODO(PLT-327): remove once app_tx_count_total verified
+		telemetry.SetGauge(float32(gInfo.GasUsed), "tx", "gas", "used")     // TODO(PLT-327): remove once app_tx_gas_used verified
+		telemetry.SetGauge(float32(gInfo.GasWanted), "tx", "gas", "wanted") // TODO(PLT-327): remove once app_tx_gas_wanted verified
+		appMetrics.txCount.Add(ctx.Context(), 1)
+		appMetrics.txCount.Add(ctx.Context(), 1, otelmetrics.WithAttributes(attribute.String("result", resultStr)))
```
bdchatham: **[blocker]** Double-counts every tx: line 163 is an unlabeled `Add(1)` and line 164 is a labeled `Add(1, result=...)`. `sum(app_tx_count_total)` returns 2× actual tx count; `sum by (result)(app_tx_count_total)` is off by N (the unlabeled adds land in a series with no `result` label).

Legacy emitted two distinct metric *names* (`tx_count` and `tx_<result>`); 1:1 port onto a labeled counter doesn't compose. Drop line 163, or split into two metrics (`app_tx_count_total` + `app_tx_result_total`).

app/ante/metrics.go
```diff
@@ -0,0 +1,27 @@
...
+	"go.opentelemetry.io/otel/metric"
+)
+
+type anteMetrics struct {
+	pendingNonce metric.Int64Counter
+}
+
+// getAnteMetrics returns the package-level OTel instruments, initializing them
+// lazily on first call (after the global MeterProvider is set in NewApp).
+var getAnteMetrics = sync.OnceValue(func() *anteMetrics {
```
bdchatham: **[should-fix]** Adds an atomic load to every `CheckTx` (hot path), and — more importantly — if anything calls `getAnteMetrics()` before `SetupOtelMetricsProvider()` runs, the noop provider binds permanently for the process lifetime.

Better: eager init at the same point `initAppMetrics()` runs in `NewApp` (`app/app.go:491`). Mirror the `app` pattern — package-level `var anteMetrics anteMetrics` + `func InitAnteMetrics()` called explicitly after `SetupOtelMetricsProvider`.

app/app.go
```diff
@@ -1217,7 +1220,9 @@ func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcess
 	// Invariant: at this point nil entries in typedTxs are proto decode failures only.
 	// EVM preprocessing runs later inside ProcessBlock; do not reorder that before this check.
 	if !app.checkTotalBlockGas(checkCtx, typedTxs) {
-		metrics.IncrFailedTotalGasWantedCheck(string(req.Header.ProposerAddress))
+		utilmetrics.IncrFailedTotalGasWantedCheck(string(req.Header.ProposerAddress)) // TODO(PLT-327): remove once app_failed_total_gas_wanted_check_total verified
+		appMetrics.failedGasWantedCheck.Add(ctx.Context(), 1,
+			otelmetric.WithAttributes(attribute.String("proposer", string(req.Header.ProposerAddress))))
```
bdchatham: **[should-fix]** `string(req.Header.ProposerAddress)` is a raw 20-byte consensus address; non-UTF8 bytes make the Prom label value untrustworthy and unreadable in dashboards. Cardinality is bounded (~50 validators on pacific-1), so not a blocker — but the value isn't useful as-is.

Either drop the label (proposer attribution belongs in a span event or log, not a metric label), or hex-encode: `attribute.String("proposer", hex.EncodeToString(req.Header.ProposerAddress))`.

bdchatham (commented):
Migration shape is right and dual-emit is sound. Three blockers worth fixing before merge — all silent (no runtime error, just wrong data):

1. Counter and histogram names will double-suffix under the Prometheus exporter (`_total_total`, `_seconds_seconds`).
2. `app_tx_count_total` double-counts every tx via the unlabeled + labeled `Add` pair.
3. `txGasUsed` / `txGasWanted` as `Int64Gauge` is last-write-wins under OCC concurrency.

Plus should-fixes around bucket density at the p99=2.5s SLO threshold, the lazy `sync.OnceValue` init in `app/ante/`, and the raw `proposer` label encoding. Inline below.

Dashboard rewrites land alongside the PLT-327 removal step; existing legacy refs in `clusters/prod/monitoring/grafana-dashboards-protocol.yaml` are summary-typed (`{quantile="0.5"}`), so query rewrites are `histogram_quantile`-based, not 1:1. No alert/recording rules touch these names — the dual-emit window is sufficient buffer.

app/ante/metrics.go
```diff
@@ -0,0 +1,36 @@
...
+
+	pendingNonce metric.Int64Counter
+}
+
+var appAnteMetrics anteMetrics
+
+// InitAnteMetrics registers all OTel instruments for the ante package.
+// Safe to call concurrently; instruments are registered exactly once.
+func InitAnteMetrics() {
+	appAnteMetrics.mu.Lock()
```
masih: Use sync.Once instead?

app/metrics.go
amir-deris: I believe this is not correct. I ran a local setup with 4 node cluster and prometheus and Grafana and I can see the metric names don't show duplicate `_total` or `_seconds`:

Image

I believe similar comment was addressed in this pr: https://github.com/sei-protocol/sei-chain/pull/3265/changes#r3190780389

app/metrics.go
amir-deris: Great point. I will remove these two metrics.

app/metrics.go
```diff
@@ -13,12 +14,20 @@ import (
...
 var histogramBuckets = metric.WithExplicitBucketBoundaries(
-	0.025, 0.05, 0.1, 0.25, 0.5, 0.75, 1.0, 1.5, 2.0, 2.5, 5.0, 10.0,
+	0.025, 0.05, 0.1, 0.25, 0.5, 0.75, 1.0, 1.5, 2.0, 2.5, 3.0, 4.0, 5.0, 10.0,
+)
+
+// millisecondBuckets is for metrics that typically complete in under 100ms, expressed in seconds.
+// Covers µs-range fast paths (25µs–1ms) and occasional slower outliers up to 100ms.
+var millisecondBuckets = metric.WithExplicitBucketBoundaries(
+	0.000025, 0.000050, 0.0001, 0.0005, 0.001, 0.0025, 0.005, 0.010, 0.020, 0.050, 0.075, 0.1,
 )
```
bdchatham: nit: I'd add `0.25, 0.5, 1.0` — `commitDuration` is on this set, and pebble compaction stalls during heavy write load can push commit past 100ms (verified by walking through `BaseApp.Commit` → `cms.Commit(true)` — the foreground write blocks during L0→L1 stalls). 100ms ceiling buckets that exactly when you most want tail visibility.

bdchatham (approved):

app/app.go
```diff
@@ -1523,7 +1550,9 @@ func (app *App) ProcessTxsSynchronousGiga(ctx sdk.Context, txs [][]byte, typedTx
 
 		txResults[i] = result
 		ctx.GigaMultiStore().WriteGiga()
-		metrics.IncrTxProcessTypeCounter(metrics.SYNCHRONOUS)
+		utilmetrics.IncrTxProcessTypeCounter(utilmetrics.SYNCHRONOUS) // TODO(PLT-327): remove once app_tx_process_type_total verified
+		appMetrics.txProcessType.Add(ctx.Context(), 1,
+			otelmetric.WithAttributes(attribute.String("type", utilmetrics.SYNCHRONOUS)))
```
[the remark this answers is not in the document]
masih: @amir-deris is this one worth addressinng?

app/metrics_test.go
```diff
@@ -0,0 +1,46 @@
...
+	appMetrics.deliverTxDuration.Record(ctx, 0.01)
+	appMetrics.deliverBatchTxDuration.Record(ctx, 0.1)
+	appMetrics.commitDuration.Record(ctx, 0.3)
+	appMetrics.blockProcessDuration.Record(ctx, 0.5)
+
+	appMetrics.txCount.Add(ctx, 1)
+	appMetrics.txProcessType.Add(ctx, 1, otelmetric.WithAttributes(attribute.String("type", "sequential")))
+	appMetrics.txGas.Add(ctx, 100, otelmetric.WithAttributes(attribute.String("type", "gas_used")))
+
+	appMetrics.optimisticProcessing.Add(ctx, 1, otelmetric.WithAttributes(attribute.Bool("enabled", true)))
```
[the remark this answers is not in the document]
masih: yup, we want this one fixed too

app/abci.go
```diff
@@ -59,10 +66,18 @@ func (app *App) EndBlock(ctx sdk.Context, height int64, blockGasUsed int64) (res
...
+		appMetrics.endBlockDuration.Record(ctx.Context(), time.Since(endBlockStart).Seconds())
+	}()
 	ctx = ctx.WithEventManager(sdk.NewEventManager())
-	defer telemetry.MeasureSince(time.Now(), "module", "total_end_block")
+	moduleEndBlockStart := time.Now()
+	defer func() {
+		telemetry.MeasureSince(moduleEndBlockStart, "module", "total_end_block") // TODO(PLT-327): remove once app_abci_module_end_block_duration_seconds verified
+	}()
 	res.ValidatorUpdates = legacyabci.EndBlock(ctx, height, blockGasUsed, app.EndBlockKeepers)
+	appMetrics.moduleEndBlockDuration.Record(ctx.Context(), time.Since(moduleEndBlockStart).Seconds())
```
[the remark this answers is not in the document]
amir-deris: Looking into this.

app/app.go
```diff
@@ -506,9 +509,11 @@ func New(
 	bApp.SetInterfaceRegistry(interfaceRegistry)
 
 	// Bind OTEL metrics provider once at application construction
-	if err := metrics.SetupOtelMetricsProvider(); err != nil {
+	if err := utilmetrics.SetupOtelMetricsProvider(); err != nil {
 		logger.Error(err.Error())
 	}
+	initAppMetrics()
+	appante.InitAnteMetrics()
```
masih: These can be linitialised at package level once without any extra code. That would remove the need for sync once. 

Let's have a 1:1 today to go over this

app/metrics.go
```diff
@@ -0,0 +1,198 @@
...
+
+// histogramBuckets aligns with block-processing
+// latency SLO thresholds (p50 ≤ 500ms, p95 ≤ 1.5s, p99 ≤ 2.5s) in seconds; 3s, 4s refine quantiles just above the p99 line.
+var histogramBuckets = metric.WithExplicitBucketBoundaries(
+	0.025, 0.05, 0.1, 0.25, 0.5, 0.75, 1.0, 1.5, 2.0, 2.5, 3.0, 4.0, 5.0, 10.0,
+)
+
+// millisecondBuckets keeps fine-grained µs–ms buckets for fast ABCI paths and adds a tail (0.5s–10s)
+// so batch deliver and commit do not collapse into +Inf. Values are in seconds.
+var millisecondBuckets = metric.WithExplicitBucketBoundaries(
```
masih: Why millisecond buckets? we are standardising time duration on seconds right?

amir-deris: Actually the values in this bucket are in seconds, they are just small values to represent fast operations. Perhaps the name of `millisecondBuckets` is misleading and it should be something else?

app/ante/metrics.go
```diff
@@ -0,0 +1,24 @@
...
+			"app_pending_nonce_total",
+			metric.WithDescription("Pending nonce events by type (added, expired, rejected, accepted)"),
+			metric.WithUnit("{count}"))),
+}
+
+func must[V any](v V, err error) V {
+	if err != nil {
+		panic(err)
+	}
+	return v
```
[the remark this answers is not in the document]
amir-deris: It seems for go, the emphasis is more on keeping packages independent at a cost of code duplication.

app/ante/evm_checktx.go
```diff
@@ -323,7 +328,8 @@ func CheckNonce(ctx sdk.Context, latestCtxGetter func() sdk.Context, ek *evmkeep
 
 			if txNonce < nextNonceToBeMined {
 				// this nonce has already been mined, we cannot accept it again
-				metrics.IncrementPendingNonce("rejected")
+				metrics.IncrementPendingNonce("rejected") // TODO(PLT-327): remove once app_pending_nonce_total verified
+				anteMetrics.pendingNonce.Add(ctx.Context(), 1, otelmetric.WithAttributes(attribute.String("event", "rejected")))
```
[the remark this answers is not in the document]
amir-deris: It seems the concern is mostly for traces and capturing context, and metrics should work fine for now. When we get to traces, will deal with these issues separately. Keeping it simple in this PR for now.

app/metrics.go
```diff
@@ -0,0 +1,194 @@
...
+	"context"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+
+	"github.com/sei-protocol/sei-chain/sei-cosmos/version"
+)
+
+var (
```
amir-deris, line 13: Using package level variables without need of `sync.Once` and calling `init` functions.

masih (approved):
Less is more ✊

PR #3397, opened 2026-05-06
Add replacement APIs for methods we intend to deprecate.

cody-littley:

Adds new methods to the top level `Committer` interface. The goal is to eventually deprecate the child `CommitKVStore` interface, and these new methods are a step in that direction.

```
type Committer interface {
    // ...
    Get(store string, key []byte) (value []byte, ok bool, err error)
    Has(store string, key []byte) (bool, error)
    Iterator(store string, start []byte, end []byte, ascending bool) (dbm.Iterator, error)
    GetProof(store string, key []byte) (*ics23.CommitmentProof, error)
    // ...
}
```

Unit tests. These methods are not yet invoked in any production pathways.


blindchaser (approved):

Kbhat1 (approved):
AI Review:

```
Verdict: no real bugs or blockers. The PR is additive, the methods are not wired into production paths yet, the tests pin the intended behaviors, and locking on the underlying Tree (RWMutex + immutable nodes) is fine. Both Committer implementers in the repo (memiavl.CommitStore, composite.CompositeCommitStore) are updated; no other implementer exists.

A few things worth flagging — none are blockers but matter for the follow-on work that intends to actually use these methods:

1. EVM data is invisible through these methods in SplitWrite mode. Because composite.GetChildStoreByName only consults cosmosCommitter, and SplitWrite routes EVM changesets exclusively to FlatKV, cs.Get(EVMStoreKey, ...) returns (nil, false, nil) for data that exists in FlatKV. The author pins this in TestCompositeSplitWriteEVMReadsAreInvisible with a TODO. Fine for now ("not invoked in production"), but the deprecation goal for CommitKVStore is blocked on fixing this routing. Any caller migrating off the deprecated path for EVM keys will silently see empty results.

Nothing here should block landing. The forward-looking issue is #1 — SplitWrite + EVM reads must be solved before these methods can replace GetChildStoreByName for the EVM store.
  ```

cody-littley:
@Kbhat1 
> 1. EVM data is invisible through these methods in SplitWrite mode. Because composite.GetChildStoreByName only consults cosmosCommitter, and SplitWrite routes EVM changesets exclusively to FlatKV, cs.Get(EVMStoreKey, ...) returns (nil, false, nil) for data that exists in FlatKV. The author pins this in TestCompositeSplitWriteEVMReadsAreInvisible with a TODO. Fine for now ("not invoked in production"), but the deprecation goal for CommitKVStore is blocked on fixing this routing. Any caller migrating off the deprecated path for EVM keys will silently see empty results.

As the code currently sits, this is a valid observation. The fix for this issue requires additional work to land, i.e. we need to integrate the `migration.Router` logic into `composite.Store`. I've got that work waiting in a branch, and I'm intentionally not merging it here since that would result in a very large PR.

PR #3398, opened 2026-05-06
Constants for migration code

cody-littley:

Defines a bunch of constants that are needed for the migration implementation. These constant's aren't used yet, but are used in my feature branch where I've implemented migration.

unit tests


Kbhat1 (approved):

blindchaser (commented):
the new `store_keys.go` pulls many module packages from x/* / sei-cosmos/x/* / sei-ibc-go/modules/* / sei-wasmd/x/wasm/types into sei-db. This transitively drags cosmos-sdk + go-ethereum into the storage layer's dependency closure and we have commentsin sei-db/common/keys/evm.go:22-25 says:

// ... avoids pulling in the heavy x/evm/types dependency // (which transitively imports cosmos-sdk).

how about we moving MemIAVLStoreKeys (and the per-module StoreKey constants it references) into sei-db/common/keys as string literals, mirroring the existing EVMStoreKey = "evm". 

If we want to preserve the "upstream rename caught at compile time" property of the current code, we can add a _test.go in sei-db/common/keys that imports the x/* packages and asserts keys.AuthStoreKey == authtypes.StoreKey etc.

blindchaser (commented):
var EVMStoreKey = evmtypes.StoreKey
var BankStoreKey = banktypes.StoreKey

use const  as right-hand sides are const?

cody-littley:
@blindchaser 
> how about we moving MemIAVLStoreKeys (and the per-module StoreKey constants it references) into sei-db/common/keys as string literals, mirroring the existing EVMStoreKey = "evm".

Done.

cody-littley:
@blindchaser 
> var EVMStoreKey = evmtypes.StoreKey
> var BankStoreKey = banktypes.StoreKey
> use const as right-hand sides are const?

All now keys are now defined as constants.

blindchaser (approved):

blindchaser:
one comment about the testing: suggest to add a `store_keys_upstream_test.go`  asserts the literals equal the upstream constants.

PR #3399, opened 2026-05-06
perf(app): parallelize per-tx gas classification in checkTotalBlockGas

amir-deris:
## Summary

`checkTotalBlockGas` iterates every tx in a proposed block to compute total gas. Each tx's classification (EVM check, gasless check, gas extraction) is independent, making the loop a natural fan-out target.

### Changes

- **`checkTotalBlockGas`** — replaced the serial loop with a two-phase design:
  - **Phase 1 (parallel):** up to 32 concurrent goroutines, one per tx, each calling `classifyTxForGas`. Results are written into a pre-allocated `[]txGasResult` slice indexed by position (no lock needed on the slice).
  - **Phase 2 (serial):** single pass over results for malicious-tx detection, overflow-safe accumulation, and `MaxGasWanted`/`MaxGas` limit enforcement.
- **`classifyTxForGas`** — extracted helper that owns all per-tx logic. EVM txs (the vast majority) run fully unlocked. The non-EVM `IsTxGasless` path takes a `ctxMu` mutex because `sdk.Context` wraps a `CacheMultiStore` that is not goroutine-safe.
- **`txGasResult`** — small struct (`gasWanted`, `gasContrib`, `skip`, `malicious`) that carries each tx's outcome between phases.
- Dropped `nonzeroTxsCnt` — it was declared and incremented but never read.

### Safety

- No data races: each goroutine owns its `results[i]` slot; `ctxMu` serializes the only shared-state reader (`IsTxGasless`).
- On typical EVM-heavy blocks `ctxMu` is uncontested — oracle-vote txs are the only non-EVM path that acquires it.
- Semantics of the overflow check and gas-limit enforcement are unchanged; they moved to the serial accumulation phase where the running total is available.

### Improvements
TBC

---

> [!NOTE]
> **Medium Risk**
> Touches consensus-critical proposal validation logic and introduces concurrency (goroutines, shared context locking), which could cause subtle race/ordering issues if incorrect despite added guards and tests.
> 
> **Overview**
> **Parallelizes `checkTotalBlockGas` per-tx classification** by fanning out tx inspection to a bounded worker pool (max 32) and then performing a serial accumulation/enforcement pass against `MaxGasWanted`/`MaxGas`.
> 
> Adds `txGasResult`/`txGasClassifyJob` and extracts `classifyTxForGas` to encapsulate per-tx gasless/EVM/associate handling, including per-worker panic recovery and an atomic "malicious" short-circuit; uses a mutex to serialize `IsTxGasless` due to non-goroutine-safe `sdk.Context`.
> 
> Expands tests to cover gas-estimate selection/fallback, `MaxGasWanted` enforcement independent of `MaxGas`, behavior with many txs (>32), and nil/empty tx slices.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit edf945e15cae1e690b2f2e9ec77b987fc820efe2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


app/app.go
```diff
@@ -2493,102 +2493,139 @@ func RegisterSwaggerAPI(rtr *mux.Router) {
...
-			// such tx will not be processed and thus won't consume gas. Skipping
+	results := make([]txGasResult, len(typedTxs))
+
+	// ctx (CacheMultiStore) is not goroutine-safe; IsTxGasless is the only caller that
+	// reads from it. EVM txs skip that path entirely, so ctxMu is uncontested on
+	// typical EVM-heavy blocks.
+	var ctxMu sync.Mutex
+	var wg sync.WaitGroup
+
+	const maxConcurrent = 32
```
masih, line 2544: Document the rationale for 32; if there is none please either paremeterise it in config or make it a factor of CPU cores with capped min and max.

app/app.go
```diff
@@ -2493,102 +2493,139 @@ func RegisterSwaggerAPI(rtr *mux.Router) {
...
-				// Other business logic errors (like duplicate votes) - continue processing but tx is not gasless
-				logger.Info("transaction failed gasless check but not malicious", "err", err)
-				continue
-			}
-			if isGasless {
-				continue
-			}
+		sem <- struct{}{}
+		wg.Add(1)
+		go func() {
```
masih: What you want here is a worker pool, right?

This implementation works, but: it creates a goroutine _per tx_. The idea is to use a pool of workers across which work is distributed. Specially since the max degree of concurrency is fixed.

Using the worker pool pattern here means we will have less allocations and the chances are if the work completes fast we might need to instantiate less than 32 concurrent goroutines (ignoring the obvious gain from goroutine reuse at max concurency)

app/app.go
```diff
@@ -2493,102 +2493,139 @@ func RegisterSwaggerAPI(rtr *mux.Router) {
...
+			results[i] = app.classifyTxForGas(ctx, tx, &ctxMu)
+		}()
+	}
+	wg.Wait()
+
+	// Serial phase: malicious check + overflow-safe accumulation + limit enforcement.
+	totalGas, totalGasWanted := uint64(0), uint64(0)
+	for _, r := range results {
+		if r.malicious {
+			return false
```
masih, line 2591: Why proceed with the remaining concurrent processing if a malicious result is found?

Have a look at err group. that primitive SDK type allows you to then return an error and as a result cancel all remaining concurrent work, which would make a more effective optimisation given that the point of this PR is exactly that.

masih (commented):
Nice; thank you @amir-deris for picking this work up so quickly. this is impactful!

The current implementation works so technically no hard blocker.
However, I have left a bunch of comments that are worth addressing which should result in a more optimal solution.

🍻

amir-deris:
Closing this PR as it is not showing significant improvement in various load testing scenarios that I did, also because it adds complexity that is not justified.

PR #3401, opened 2026-05-06
feat: gate block-validation bypasses behind //go:build mock_block_validation

bdchatham:
## Why

A node running an alternate execution engine (e.g., Giga V2) needs to bypass certain header-hash checks to validate against canonical-chain blocks. The previous shape exposed `--skip-app-hash-validation` as a runtime CLI flag on every seid binary, which made the production validator one config edit away from accepting bad blocks.

## Approach

The build tag `mock_block_validation` is the toggle — there is no runtime knob.

```go
// types/consensus_policy_default.go (//go:build !mock_block_validation)
type ConsensusPolicy struct{}
func (ConsensusPolicy) SkipAppHashValidation() bool  { return false }
func (ConsensusPolicy) SkipDataHashValidation() bool { return false }

// types/consensus_policy_mock_block_validation.go (//go:build mock_block_validation)
type ConsensusPolicy struct{}
func (ConsensusPolicy) SkipAppHashValidation() bool  { return true }
func (ConsensusPolicy) SkipDataHashValidation() bool { return true }
```

`ConsensusPolicy` is an empty struct in both builds; methods return different constants per build and the compiler folds them. Bypass branches DCE in production. There is no Go API in either build to construct a policy with the opposite behavior.

`Block.ValidateBasic(policy)` wraps the gated hash checks in `if !policy.Skip*() { ... }` in-place — no structural changes to the function body. `validateBlock(state, block, policy)` carries the policy from `BlockExecutor` down. Production `start.go` passes `DefaultConsensusPolicy()` unconditionally — no cobra, no AppOptions, no viper.

## Naming

`mock_block_validation` matches the repo's existing `mock_balances` precedent — both are mock build tags for engineering / debug binaries. Underscore form is required for valid Go build tags. Image tag uses hyphen for readability: `sei-chain:mock-block-validation-<sha>`.

## Per-check flags, narrow scope

The original `ftr-skip-apphash` branch gated 6 hash checks under one umbrella flag; this PR ships 2:

| Hash check | Gated | Reason |
|---|---|---|
| AppHash | Yes | State root from execution — guaranteed to diverge for an alternate execution engine. |
| DataHash | Yes | Defensive against tx-encoding format skew across versions. |
| LastCommitHash | **No** | Validator-authenticity territory (merkle root of validator signatures). Should match against a same-version canonical chain. |
| EvidenceHash | **No** | Validator-authenticity territory (merkle root of misbehavior evidence). Same reasoning. |
| ValidatorsHash, NextValidatorsHash | **No** | Validator-set updates flow through cosmos-sdk x/staking, not through the EVM execution engine. Should match. |
| ConsensusHash, LastBlockID, VerifyCommit, ProposerAddress | No (unconditional) | Network-config or canonical-chain-derived; identical regardless of local execution. |
| LastResultsHash | (separate) | Gated by the existing `SkipLastResultsHashValidation` global, set from `gigaExecutorConfig.Enabled` — a first-class production runtime feature with intentional gas-accounting divergence, different security model. |

If a non-gated check fails on a candidate node, that's actionable signal — either a real bug in the setup or a wire-format incompatibility worth investigating, not silently bypassed.

## CI

`.github/workflows/ecr.yml` adds a single new build step alongside the existing `mock_balances` and production builds, mirroring the same pattern. Every push produces:

- `sei-chain:<sha>` — production
- `sei-chain:mock-<sha>` — mock_balances (unchanged)
- `sei-chain:mock-block-validation-<sha>` — new

## Verification

- `strings seid-default | grep skip-app-hash` → 0 hits (no flag, no string in either binary)
- `go build ./...` and `go build -tags mock_block_validation ./...` both clean

## Out of scope

- The two Dockerfile commits from `ftr-skip-apphash` (libevmone runtime + ubuntu 24.04 base for GLIBCXX) are giga-runtime support and ship as a separate PR.
- A `BlockValidator` interface that hides per-step validation behind a strategy-pattern abstraction was discussed and deferred — defensible refactor when there's a second consumer.

### DIFF
--- a/.github/workflows/ecr.yml
+++ b/.github/workflows/ecr.yml
@@ -50,6 +50,18 @@ jobs:
           build-args: |
             SEI_CHAIN_REF=${{ inputs.ref || github.sha }}
             GO_BUILD_TAGS=mock_balances
+      - name: Build and push seid with mock block validation
+        uses: docker/build-push-action@v6
+        with:
+          context: '.'
+          cache-from: type=registry,ref=${{ steps.login-ecr.outputs.registry }}/sei/build-cache:shared
+          cache-to: type=registry,ref=${{ steps.login-ecr.outputs.registry }}/sei/build-cache:shared,mode=max
+          platforms: linux/amd64
+          push: true
+          tags: ${{ steps.login-ecr.outputs.registry }}/sei/sei-chain:mock_block_validation-${{ inputs.tag || inputs.ref || github.sha }}
+          build-args: |
+            SEI_CHAIN_REF=${{ inputs.ref || github.sha }}
+            GO_BUILD_TAGS=mock_block_validation
       - name: Build and push seid
         uses: docker/build-push-action@v6
         with:
--- a/sei-tendermint/internal/state/execution.go
+++ b/sei-tendermint/internal/state/execution.go
@@ -53,6 +53,12 @@ type BlockExecutor struct {
 
 	metrics *Metrics
 
+	// consensusPolicy is a compile-time validation bypass that only takes
+	// effect in mock_block_validation builds; production binaries always see
+	// the zero-value (no bypass). Distinct from types.SkipLastResultsHashValidation
+	// below, which is a runtime atomic.Bool flipped on for the Giga executor.
+	consensusPolicy types.ConsensusPolicy
+
 	// cache the verification results over a single height
 	cache map[string]struct{}
 }
@@ -66,16 +72,18 @@ func NewBlockExecutor(
 	blockStore BlockStore,
 	eventBus *eventbus.EventBus,
 	metrics *Metrics,
+	consensusPolicy types.ConsensusPolicy,
 ) *BlockExecutor {
 	return &BlockExecutor{
-		eventBus:   eventBus,
-		store:      stateStore,
-		app:        app,
-		mempool:    pool,
-		evpool:     evpool,
-		metrics:    metrics,
-		cache:      make(map[string]struct{}),
-		blockStore: blockStore,
+		eventBus:        eventBus,
+		store:           stateStore,
+		app:             app,
+		mempool:         pool,
+		evpool:          evpool,
+		metrics:         metrics,
+		cache:           make(map[string]struct{}),
+		blockStore:      blockStore,
+		consensusPolicy: consensusPolicy,
 	}
 }
 
@@ -156,7 +164,7 @@ func (blockExec *BlockExecutor) ValidateBlock(ctx context.Context, state State, 
 		return nil
 	}
 
-	err := validateBlock(state, block)
+	err := validateBlock(state, block, blockExec.consensusPolicy)
 	if err != nil {
 		// Check if this is a LastResultsHash mismatch and log detailed info
 		if !types.SkipLastResultsHashValidation.Load() && !bytes.Equal(block.LastResultsHash, state.LastResultsHash) {
--- a/sei-tendermint/types/consensus_policy.go
+++ b/sei-tendermint/types/consensus_policy.go
@@ -1,0 +1,7 @@
+package types
+
+// DefaultConsensusPolicy returns the zero-value policy. Behavior is
+// build-tag-dependent: production builds enforce every check; mock_block_validation
+// builds bypass every gated check (see consensus_policy_default.go and
+// consensus_policy_mock_block_validation.go).
+func DefaultConsensusPolicy() ConsensusPolicy { return ConsensusPolicy{} }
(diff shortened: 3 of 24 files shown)


.github/workflows/ecr-execution-shadow.yml
```diff
@@ -0,0 +1,48 @@
+name: ECR Execution Shadow
```
masih: Why an entire new job instead of parameterising the existing ECR CI workflow?

sei-tendermint/types/consensus_policy_execution_shadow.go
```diff
@@ -0,0 +1,10 @@
+//go:build execution_shadow
```
masih: The word "shadow" means something to some of us at Sei, and has an evolving meaning. 

I recommend using a dry-er naming scheme instead, something along the lines of the CLI flag: `mock_apphash_validation`. That would: 1) keep to the existing naming pattern (mock balances), and 2) relieve us from having to explain what "shadow" means, which itself has different meanings even between us engineers.

.github/workflows/ecr-execution-shadow.yml
bdchatham: Mock execution images are fairly specialized scenarios. Not every binary needs to be packaged into one is the motivation here

.github/workflows/ecr.yml
```diff
@@ -50,6 +50,18 @@ jobs:
...
             GO_BUILD_TAGS=mock_balances
+      - name: Build and push seid with mock block validation
+        uses: docker/build-push-action@v6
+        with:
+          context: '.'
+          cache-from: type=registry,ref=${{ steps.login-ecr.outputs.registry }}/sei/build-cache:shared
+          cache-to: type=registry,ref=${{ steps.login-ecr.outputs.registry }}/sei/build-cache:shared,mode=max
+          platforms: linux/amd64
+          push: true
+          tags: ${{ steps.login-ecr.outputs.registry }}/sei/sei-chain:mock-block-validation-${{ inputs.tag || inputs.ref || github.sha }}
```
masih: I think we use underscore delimitation for other image tags. Could we keep to a consistent pattern?

masih (approved):
🙌

sei-tendermint/types/consensus_policy.go
```diff
@@ -0,0 +1,4 @@
+package types
+
+// DefaultConsensusPolicy enforces every check.
```
blindchaser: nit: consider reword as under the mock_block_validation build tag it skips all gated checks, right?

sei-tendermint/types/consensus_policy_default.go
```diff
@@ -0,0 +1,10 @@
+//go:build !mock_block_validation
+
+package types
+
+// ConsensusPolicy is empty in production builds. Its methods return
+// constant false so the compiler DCEs the bypass branches.
+type ConsensusPolicy struct{}
```
blindchaser, line 7: i feel Type name ConsensusPolicy is broad for what it gates, would something like ValidationBypassPolicy be more accurate?

sei-tendermint/internal/state/execution.go
```diff
@@ -156,7 +160,7 @@ func (blockExec *BlockExecutor) ValidateBlock(ctx context.Context, state State,
 		return nil
 	}
 
-	err := validateBlock(state, block)
+	err := validateBlock(state, block, blockExec.consensusPolicy)
 	if err != nil {
 		// Check if this is a LastResultsHash mismatch and log detailed info
 		if !types.SkipLastResultsHashValidation.Load() && !bytes.Equal(block.LastResultsHash, state.LastResultsHash) {
```
blindchaser, line 170: [Nit] the codebase now has two "skip a hash check" mechanisms with very different lifecycles: types.SkipLastResultsHashValidation (runtime atomic, production Giga feature) and types.ConsensusPolicy (build-tag, mock binary only). They intentionally solve different problems, but having both unlabelled near each other in validation.go / execution.go may confuse future readers. Consider a short comment near ConsensusPolicy clarifying it is a compile-time bypass for the mock binary, distinct from the runtime Giga flag.

sei-tendermint/types/consensus_policy_default.go
bdchatham: Good callout. I intentionally left this broad since there are other dimensions we may want to bypass in the future for testing/shadowing purposes

blindchaser: it totally makes sense if we plan to add more to bypass in the future

blindchaser (approved):

arajasek (approved):

PR #3406, opened 2026-05-07
test(evm): migrate response-reading lib.js helpers off -b block (CON-256)

wen-coding:
## Summary

Picks up where #3363 left off. Migrates the lib.js helpers that read post-execution events from `-b block` to `-b sync` + per-helper side-effect waits.

Pattern (introduced in #3363, extended here): submit with `-b sync`, then poll the actual on-chain side effect rather than `seid q tx` (which doesn't return under Autobahn). When the side effect is observable, return the post-state value; when it isn't, fall back to admin-sequence-advance + block-walk to recover the inclusion height.

**Migrated:**

- `storeWasm` — side-effect: max wasm code_id grows (new `getMaxWasmCodeId`)
- `instantiateWasm` — side-effect: new contract appears under codeId (new `listContractsByCode`)
- `registerPointerForERC20/721/1155` — side-effect: `seid query evm pointer` returns non-empty (factored into shared `registerPointerForCw`)
- `proposeCW20toERC20Upgrade`, `proposeParamChange` — side-effect: new gov proposal with matching title (factored into `findProposalByTitle`)
- `executeWasm`, `associateWasm` — fallback path: `waitForAdminTxCommit` waits on admin sequence advance, then block-walks to recover the actual inclusion height (no easy single-query side effect for arbitrary wasm execute / contract association)

Wait-timeout cases are wrapped with the same error prefixes (`"storeWasm failed"`, `"instantiateWasm failed"`, `"contract deployment failed"`) that callers already match on, so existing assertions in `DisableWasmTest` and similar continue to work.

**Other:**

- Removed `associateSigner` + hardcoded `seilocalSignerPrivateKeys` — `setupSigners` now uniformly funds the EVM address and reads back the sei address. The conditional branch became redundant after #3431.
- Dropped 3 `res.code` assertions in `CW1155toERC1155PointerTest.js` — `res.code` is now CheckTx (always 0 for mempool admission), not DeliverTx. Outcome verification stays via the existing post-state balance checks.

### DIFF
--- a/contracts/test/lib.js
+++ b/contracts/test/lib.js
@@ -422,9 +422,18 @@ async function storeWasm(path, from=adminKeyName) {
     const command = `seid tx wasm store ${path} --from ${from} --gas=5000000 --fees=1000000usei -y -b sync -o json`
     const response = JSON.parse(await execute(command))
     if (response.code !== 0) throw new Error(`storeWasm failed: ${response.raw_log}`)
+    // Capture the new code_id inside the wait: a second re-query after the
+    // wait would race with any other concurrent store landing in the same
+    // window. Serial test execution makes that race dormant today, but
+    // capture-during-wait closes it deterministically.
+    let newCodeId = 0
     try {
         await waitForCondition(
-            async () => (await getMaxWasmCodeId()) > codeIdBefore,
+            async () => {
+                const cur = await getMaxWasmCodeId()
+                if (cur > codeIdBefore) { newCodeId = cur; return true }
+                return false
+            },
             `new wasm code_id > ${codeIdBefore}`,
         )
     } catch (e) {
@@ -433,7 +442,7 @@ async function storeWasm(path, from=adminKeyName) {
         // with the "storeWasm failed" prefix that callers match on.
         throw new Error(`storeWasm failed: ${e.message}`)
     }
-    return String(await getMaxWasmCodeId())
+    return String(newCodeId)
 }
 
 async function getPointerForCw20(cw20Address) {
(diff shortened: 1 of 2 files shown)


wen-coding:
This doesn't work until we link in block db which supports tx lookup.

contracts/test/lib.js
```diff
@@ -799,15 +843,78 @@ async function queryWasm(contractAddress, operation, args={}){
 
 async function executeWasm(contractAddress, msg, coins = "0usei") {
     const jsonString = JSON.stringify(msg).replace(/"/g, '\\"'); // Properly escape JSON string
-    const command = `seid tx wasm execute ${contractAddress} "${jsonString}" --amount ${coins} --from ${adminKeyName} --gas=5000000 --fees=1000000usei -y --broadcast-mode block -o json`;
-    const output = await execute(command);
-    return JSON.parse(output);
+    const command = `seid tx wasm execute ${contractAddress} "${jsonString}" --amount ${coins} --from ${adminKeyName} --gas=5000000 --fees=1000000usei -y -b sync -o json`;
+    return await waitForAdminTxCommit(command)
```
philipsu522, line 847: nice

contracts/test/lib.js
```diff
@@ -799,15 +843,78 @@ async function queryWasm(contractAddress, operation, args={}){
...
+// Note: the returned `code` is the CheckTx code (0 for accepted into
+// mempool), not the DeliverTx code; callers can't assert tx success/
+// failure from the response alone — they must inspect post-state.
+async function waitForAdminTxCommit(command) {
+    const senderAddr = await getKeySeiAddress(adminKeyName)
+    const heightBefore = await getCurrentBlockHeight()
+    const seqBefore = await getAccountSequence(senderAddr)
+    const response = JSON.parse(await execute(command))
+    if (response.code !== 0) return response  // CheckTx rejection — surface immediately
+    await waitForCondition(
```
philipsu522, line 901: do we timeout in case there's some sort of bug here?

philipsu522 (approved):

contracts/test/lib.js
wen-coding: Yes, waitForCondition has a default 30s timeout, adding a note.

codchen (approved):

PR #3410, opened 2026-05-08
Removed callbacks from TxMempool

pompon0:
EVM keeper used to mirror mempool state to be able to track highest nonce per account in mempool, which required passing callbacks from EVM keeper to TxMempool. Additionally it passed a IsPending callback, so that TxMempool could verify if evm transaction is ready to be executed on demand. This pr is simplifying the execution flow as follows:
* made Application take a LocalClient instead of generic Client, so that new Application -> Tendermint calls do not require exposing new network rpcs
* added EVMNonce and EVMBalance methods to LocalClient so that Tendermint can evaluate evm tx eligibility without callbacks needed
* removed callbacks from ResponseCheckTxV2
* made GetTransactionCount EVM RPC call Tendermint to evaluate the high nonce in mempool per account instead of mirroring mempool state.

Additionally
* marked some config fields as deprecated, given that only local tendermint node is supported
* changed type of evm address fields from hex encoded string to common.Address
* removed expirationIndex, since it was just mirroring txStore.

---

> [!NOTE]
> **High Risk**
> High risk because it changes core `CheckTx`/mempool interactions, EVM nonce/balance eligibility logic, and refactors client/RPC service wiring across the node and gRPC/REST paths.
> 
> **Overview**
> Simplifies EVM `CheckTx` by **removing mempool callback plumbing** (`PendingTxChecker`, `CheckTxCallback`, `ExpireTxHandler`) and instead attaching `EVMRequiredBalance` plus explicit sender metadata (`EVMSenderAddress` as `common.Address`, new `SeiSenderAddress`) to `ResponseCheckTxV2`.
> 
> Adds `EvmNonce`/`EvmBalance` accessors on the application/local client side so Tendermint/RPC can evaluate nonce and balance without mirroring mempool state; `eth_getTransactionCount(pending)` now returns `tmClient.EvmNextPendingNonce` and non-pending paths use `GetNonce` directly.
> 
> Refactors service wiring to use a `client.LocalClient` throughout (`RegisterLocalServices` replacing separate tx/tendermint service registration), updates broadcast/query helpers to accept an explicit node + `context.Context`, and deprecates/removes out-of-process ABCI config/flags (socket/grpc proxy-app).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d4d790292c8929b5e34c86c373618869e4e7cdb2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sei-will (approved):

sei-tendermint/internal/mempool/mempool.go
```diff
@@ -272,6 +277,46 @@ func (txmp *TxMempool) Config() *Config { return txmp.config }
 
 func (txmp *TxMempool) App() *proxy.Proxy { return txmp.app }
 
+func (txmp *TxMempool) EvmNextPendingNonce(addr common.Address) uint64 {
+	an := evmAddrNonce{addr, txmp.app.EvmNonce(addr)}
+	for byAddrNonce := range txmp.byAddrNonce.Lock() {
```
wen-coding, line 282: How often is this function called? We only need a read lock here right?

sei-tendermint/internal/mempool/mempool.go
```diff
@@ -995,10 +1017,14 @@ func (txmp *TxMempool) handleRecheckResult(tx types.Tx, res *abci.ResponseCheckT
 	// if an existing transaction is evicted during CheckTx and while this
 	// callback is being executed for the same evicted transaction.
 	if !txmp.txStore.IsTxRemoved(wtx) {
-		err := txmp.checkResponseState(res.ResponseCheckTx)
+		err := txmp.checkResponseState(wtx)
+		if evm, ok := wtx.evm.Get(); ok {
+			evm.requiredBalance = new(big.Int).Set(res.EVMRequiredBalance)
+			wtx.evm = utils.Some(evm)
```
wen-coding, line 1009: Could some other thread be reading this wtx.evm at the same time?

sei-tendermint/internal/mempool/mempool.go
```diff
@@ -434,60 +478,41 @@ func (txmp *TxMempool) CheckTx(ctx context.Context, tx types.Tx, txInfo TxInfo)
...
-		// throw off the correct priority percentiles otherwise.
-		//
-		// We do not use the priority hint here as it may be misleading and
-		// inaccurate. The true priority as determined by the application is the
-		// most accurate.
-		txmp.priorityReservoir.Add(res.Priority)
-		err = txmp.addNewTransaction(wtx, res.ResponseCheckTx, txInfo)
-		if err != nil {
+	if !txmp.isPending(wtx) {
+		if err := txmp.addNewTransaction(wtx); err != nil {
```
codchen, line 500: `addNewTransaction` may return nil when the tx isn't really inserted because of capacity. This wasn't an issue before but becomes one now because we don't want to `txmp.addNonce(wtx)` below in such cases. I think we can just make `addNewTransaction` return error https://github.com/sei-protocol/sei-chain/pull/3410/changes#diff-91c10e3fa6ed77b7ad77bae9a7cfbc8b102e837fdf521fa96817ef134b65f05aR934

sei-tendermint/internal/mempool/mempool.go
pompon0: It is only called by GetTransactionCount EVM RPC, which I expect is called less often than checkTx.

sei-tendermint/internal/mempool/mempool.go
pompon0: nope, handleRecheckResult is called under TxMempool.mtx write lock and all other accesses of evm field are under TxMempool.mtx read/write lock. There are many preexisting race conditions on other fields of WrappedTx but I am not fixing them in this pr.

sei-tendermint/internal/mempool/mempool.go
```diff
@@ -434,84 +478,51 @@ func (txmp *TxMempool) CheckTx(ctx context.Context, tx types.Tx, txInfo TxInfo)
...
+		if err := txmp.pendingTxs.Insert(wtx); err != nil {
+			txmp.cache.Remove(txHash)
 			return nil, err
 		}
 	}
-
-	if res.CheckTxCallback != nil {
-		res.CheckTxCallback(res.Priority)
-	}
+	txmp.addNonce(wtx)
```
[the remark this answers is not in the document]
pompon0: that's a preexisting behavior

sei-tendermint/internal/mempool/mempool.go
pompon0: that's preexisting behavior, will fix in another pr

wen-coding (approved):

PR #3417, opened 2026-05-11
test(flatkv): Expands docker-level FlatKV coverage 

blindchaser:
## Summary

Adds an offline `memiavl -> FlatKV` import path for the `evm` module and the docker-level coverage that locks in correctness across import, state-sync, cross-validator agreement, and crash recovery. The two new abstractions are:

- `flatkv.ImportTranslator` -- a single-shot translator that converts raw EVM/non-EVM changesets into physically-encoded `PhysicalKVPair`s using the same `classifyAndPrefix` / `processStorage|Code|Legacy` / `mergeAccountUpdates` logic as the live write path, but with no read-modify-write against the import target. Account fragments (nonce, codehash) are buffered across `Translate` calls and flushed exactly once per address by `Finalize`.
- `flatkv.KVImporter.Abort` -- a tear-down entrypoint that records an external failure as the first pipeline error so the worker pipeline drains without calling `FinalizeImport` / `WriteSnapshot`, leaving the on-disk store at its pre-import committed version. Paired with an idempotent `Close` so the offline-import CLI can defer cleanup on both success and error paths.

Implementation changes grouped by file:

- `sei-db/state_db/sc/flatkv/import_translator.go`: new file. `ImportTranslator` enforces single-shot semantics: `Translate` after `Finalize` returns `ErrImportTranslatorFinalized` instead of panicking on a nil map. Deletes are dropped up front so `mergeAccountUpdates` cannot interpret nil values as "set field to zero" on an empty target. The shared `appendNonDeletes` generic hoists "drop tombstone, serialize" into one place across storage/code/legacy/account.
- `sei-db/state_db/sc/flatkv/importer.go`: `KVImporter` gains `Err()`, `Abort(reason)`, and a `finishOnce`-gated `Close`. Close is idempotent: the first call drains workers + finalizes + snapshots, subsequent calls return the cached `finishErr`. If a pipeline error is recorded before Close (by a worker, dispatcher, or `Abort`), Close skips `FinalizeImport` / `WriteSnapshot` so the store stays at its pre-import version. A package-level `flushHookForTest` (atomic.Pointer) gates worker flushes for whitebox backpressure tests.
- `sei-db/tools/cmd/seidb/operations/import_flatkv_from_memiavl.go`: new `ImportFlatKVFromMemiavlCmd` + `MemiavlLatestVersionCmd`. The import flow refuses any height other than `memiavl-latest`: a stale `H` would let the next GIGA_STORAGE startup silently roll memiavl back via `CompositeCommitStore.reconcileVersions`, truncating cosmos blocks in `(H, memiavlLatest]`. Refuses to run over an existing committed FlatKV without `--force`. The deferred cleanup routes every non-nil error through `KVImporter.Abort` so a half-imported snapshot is never committed.
- `sei-db/tools/cmd/seidb/main.go`: registers the two new cobra commands.
- `sei-db/state_db/sc/migration/OPERATIONS.md`: new operations runbook for the `V0 -> V1 MigrateEVM` transition: failure-mode catalog, detection signals, repair playbook, and tooling gap analysis. Scoped to MigrateEVM only.

CI / integration coverage:

- `.github/workflows/integration-test.yml`: appends FlatKV physical-consistency checks to the existing state-sync row and adds a dedicated `FlatKV Integration` matrix row that exercises the EVM fixture, offline import, crash recovery, state-sync crash recovery, total-loss recovery, and partial-loss-fails-loudly scenarios in order.
- `integration_test/contracts/deploy_flatkv_evm_fixture.sh`: deploys a deterministic EVM fixture (native transfer + storage contract) and writes expected balances / storage / code at known heights into `flatkv_evm_*.txt` files for downstream verifiers.
- `integration_test/contracts/import_flatkv_evm_cluster.sh`: stops all four validators with `pkill`, aligns every node to a uniform import height via `seid rollback` + consensus-WAL drop, runs the offline import, flips `evm-ss-split=false` and `sc-enable-lattice-hash=false` to preserve the AppHash trajectory across the import boundary, then restarts and waits for both Tendermint and the EVM RPC port.
- `integration_test/contracts/verify_*_flatkv_digest.sh` (`cross_validator`, `statesync`): compare FlatKV `dump-flatkv --height H` output (sha256 over account|code|storage|legacy) across validators or donor/receiver at a real committed chain height, sidestepping the `SnapshotInterval=10000` floor that makes flatkv snapshot-version intersection degenerate to `{0}`.
- `integration_test/contracts/verify_flatkv_crash_recovery.sh`: SIGKILLs one validator mid-block-production, waits for catch-up, then digest-compares FlatKV across all four validators at a shared committed height.
- `integration_test/contracts/verify_flatkv_statesync_crash_recovery.sh`: wipes one validator's local state, starts state-sync, SIGKILLs mid-restore, restarts, and requires logical FlatKV equivalence with the donor.
- `integration_test/contracts/verify_flatkv_total_loss_recovery.sh`: full local-state-loss + state-sync recovery with FlatKV EVM equivalence assertions.
- `integration_test/contracts/verify_flatkv_partial_loss_fails_loudly.sh`: deletes only the FlatKV directory and requires the node to either fail loudly on restart or fully self-heal and re-match the cluster digest.
- `integration_test/contracts/verify_flatkv_evm_store.sh`: smoke-checks the dumped `storage` bucket contains the fixture contract address and a row whose serialized value length matches the on-disk `StorageData` layout (1B tag + 8B big-endian height + 32B slot).
- `integration_test/seidb/flatkv_evm_test.yaml`: runner-driven verification that historical balance / storage / code queries at the fixture heights match the captured expectations on the post-import cluster, and that non-EVM modules remain queryable.

## Test plan

- `sei-db/state_db/sc/flatkv/import_translator_test.go`:
  - Empty / nil changeset is a no-op and `Finalize` on an unused translator returns nothing.
  - Storage, code, legacy (EVM-prefixed and non-EVM), nonce-only, and codehash-only entries serialize through the correct `vtype.Deserialize*` paths.
  - Nonce + codehash in one call and across calls both merge into a single `AccountData` for the same address.
  - Deletes are dropped (storage / code / nonce / codehash) and produce zero pairs.
  - Empty key and malformed nonce payload are rejected with a clear error.
  - `Finalize` is idempotent on an exhausted translator; `Translate` after `Finalize` returns `ErrImportTranslatorFinalized` instead of panicking.
- `sei-db/state_db/sc/flatkv/importer_test.go`:
  - Close idempotency on both the happy path and the post-error path; the second/third Close never re-closes `ingestCh`.
  - `Err()` returns the dispatcher's first pipeline error before any explicit `Close`.
  - `setErr` CAS is atomic: only the first error is recorded; `done` is closed exactly once.
  - `AddNode` after `done` is closed does not block even when `ingestCh` is saturated past its buffer.
  - Large import (`importBatchSize * 3 + 100` pairs) triggers at least three mid-pipeline flushes.
  - `Abort` skips `FinalizeImport` / `WriteSnapshot`, leaves `s.Version()` at its pre-import value, and surfaces the supplied reason through both the Abort return and `Err()`; subsequent `Close` returns the cached error.
  - `Abort(nil)` substitutes a non-nil reason; `Abort` after a successful `Close` is a no-op (does not rewind the committed version).
  - Backpressure: `flushHookForTest` gates every worker flush; the producer is observably blocked once `ingestCh + worker.ch + in-flight batch` are full, releases cleanly when the gate opens, and every pair is persisted.
- `sei-db/state_db/sc/flatkv/wal_torn_write_test.go`:
  - Mid-record WAL truncation: reopened store stays at version 4 with the pre-torn root hash, lthash consistent.
  - Partial length-prefix tail: same recovery as mid-record truncation.
  - Clean WAL tail: store replays version 5 with the new root, lthash consistent.
- `sei-db/tools/cmd/seidb/operations/import_flatkv_from_memiavl_test.go`:
  - Round-trip a heterogeneous EVM changeset (storage, code, nonce, codehash, EOA-only nonce, contract-only codehash, legacy EVM key) through `importMemiavlModulesToFlatKV` and read every entry back through the flatkv `CommitStore`.
  - Refuses to run over an existing committed FlatKV without `--force`; `--force` correctly overwrites the prior store.
  - Refuses out-of-range, stale (memiavl-ahead), and future (height-ahead-of-memiavl-latest) heights with operator-facing error messages.
  - Large-dataset (~70k pairs) round-trip crosses `importBatchSize` multiple times, exercises dispatcher backpressure, and exercises the translator's cross-batch account merge buffer; spot-checked addresses across the dataset return the expected nonce / codehash / code / storage, and a synthetic out-of-range address misses.
- Docker integration (run from `.github/workflows/integration-test.yml`):
  - State-sync FlatKV digest equality between donor and receiver at a shared committed chain height.
  - Cross-validator FlatKV digest equality at a shared committed chain height under the `Snapshot Restore` job.
  - End-to-end EVM fixture probe (historical balance, storage, code, missing-account, latest-still-sees-fixture, non-EVM module still queryable) both before and after the offline import.
  - FlatKV physical-layout smoke check on the dumped storage bucket.
  - SIGKILL-mid-block crash recovery with post-restart digest equality across all four validators.
  - State-sync interrupted by SIGKILL, then resumed; final FlatKV content matches the donor.
  - Total-loss + state-sync recovery: FlatKV EVM content is logically equivalent across the cluster.
  - Partial-loss (FlatKV-only deletion): node either fails loudly with a flatkv/version/LoadVersion/reconcile log line, or stays healthy and re-matches the cluster digest.

---

> [!NOTE]
> **Medium Risk**
> Adds substantial new integration coverage and introduces/changes FlatKV import tooling and importer lifecycle semantics; failures could affect state migration/recovery workflows and test stability rather than runtime consensus logic.
> 
> **Overview**
> **Expands docker CI coverage for FlatKV** by adding a consolidated `FlatKV Integration` matrix job plus new scripts that (1) deploy an EVM fixture, (2) run an offline memiavl→FlatKV import across validators, and (3) validate post-import correctness via EVM historical/latest RPC checks, on-disk `dump-flatkv` assertions, cross-node digest comparisons, and destructive recovery scenarios (SIGKILL crash recovery, state-sync crash recovery, total state loss recovery, and FlatKV-only loss behavior). `Chain Operation Test` also now verifies state-sync correctness by diffing FlatKV digests.
> 
> **Adds and hardens FlatKV import tooling**: introduces `seidb import-flatkv-from-memiavl` (evm-only, height safety checks, `--force` overwrite guard, fail-fast error polling) and a companion `memiavl-latest-version` helper for orchestrated multi-node imports. Under the hood it adds a new `ImportTranslator` to translate memiavl changesets into FlatKV physical rows (buffering and finalizing account writes), and updates `KVImporter` to support `Err()` reporting, idempotent `Close()`, and `Abort()` to prevent partial imports from finalizing/writing snapshots.
> 
> **Improves test coverage** with new unit tests for the translator, importer concurrency/lifecycle/backpressure/abort semantics, and FlatKV WAL torn-write recovery, plus an operational doc (`sei-db/state_db/sc/migration/OPERATIONS.md`) describing migration failure modes and recovery tooling roadmap.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 867afb8cde1dfe2bc0bf2fd600ec56b46eb6affc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/sei-db/state_db/sc/migration/OPERATIONS.md
+++ b/sei-db/state_db/sc/migration/OPERATIONS.md
@@ -1,0 +1,383 @@
+# MigrateEVM Operations Roadmap
+
+This document is the operational companion to [README.md](README.md). README explains *what* the migration is and *how* the data path is wired; this document explains *what can go wrong*, *how to detect it*, *how to repair it*, and *what tooling needs to exist* to make MigrateEVM (V0 -> V1) safe to operate in production and reproducible in CI.
+
+It is deliberately scoped to **MigrateEVM only** (the V0 -> V1 transition that moves the `evm/` module from memiavl to flatkv). The same framework will apply to MigrateAllButBank (V1 -> V2) and MigrateBank (V2 -> V3) but those are out of scope for the first iteration.
+
+---
+
+## 1. Scope
+
+MigrateEVM is the first of three migrations defined in [migration_versions.go](migration_versions.go):
+
+- `Version0_MemiavlOnly = 0` -- pre-migration. All modules in memiavl.
+- `Version1_MigrateEVM  = 1` -- post-migration. `evm/` lives in flatkv; everything else still in memiavl.
+
+The transition between them is driven by `MigrationManager.ApplyChangeSets`, configured via `WriteMode = MigrateEVM` (see [write_mode.go](write_mode.go) and `buildMigrateEVMRouter` in [router_builder.go](router_builder.go)).
+
+This roadmap covers operations that touch the `evm/` module's storage during and around that transition. It does **not** cover EVM execution semantics, JSON-RPC behavior, or non-EVM modules.
+
+---
+
+## 2. Lifecycle Reference
+
+```mermaid
+flowchart LR
+    v0["V0 Memiavl-only<br/>evm in memiavl<br/>MigrationVersionKey=0"]
+    mid["V0 to V1 transition<br/>evm split by boundary cursor<br/>MigrationBoundaryKey set"]
+    v1["V1 EVMMigrated<br/>evm in flatkv only<br/>MigrationVersionKey=1<br/>MigrationBoundaryKey deleted"]
+    v0 -->|MigrateEVM begins| mid
+    mid -->|cursor reaches end| v1
+```
+
+| Stage | evm in memiavl | evm in flatkv | `MigrationVersionKey` | `MigrationBoundaryKey` | Data path |
+|---|---|---|---|---|---|
+| **V0** | full | empty | `0` (or absent) in memiavl `MigrationStore` | absent | direct memiavl; `BuildRouter` not called (see README sec "Version 0") |
+| **transition** | keys with logical_k > boundary | keys with logical_k <= boundary | `0` in memiavl | serialized `MigrationBoundary` in flatkv `MigrationStore` | `MigrationManager` routes by boundary |
+| **V1** | empty (modulo cleanup) | full | `1` in flatkv | absent (deleted on final block) | `buildEVMMigratedRouter`: evm direct to flatkv |
+
+Authoritative source for the boundary key handling: [migration_manager.go:371-394](migration_manager.go).
+
+---
+
+## 3. Atomicity & Crash-Safety Statement
+
+The single most important fact for understanding the failure modes below:
+
+> `MigrationManager.ApplyChangeSets` writes to memiavl and flatkv in **two parallel goroutines** ([migration_manager.go:396-426](migration_manager.go)). The interface contract on `DBWriter` explicitly states *"May not be atomic. If not atomic, then the caller must provide crash safe atomicity."* ([migration_types.go:39-40](migration_types.go)).
+
+Per-block guarantees:
+- **memiavl writer** applies `oldDBChangeSet`: deletions of the migrated batch + any incoming user writes routed to memiavl.
+- **flatkv writer** applies `newDBChangeSets`: inserts for the migrated batch + any incoming user writes routed to flatkv + the new `MigrationBoundaryKey` (or, on the final block, `MigrationVersionKey=1` and a delete of the boundary).
+- The two writers can finish in either order. If the process dies between them, on-disk state is split.
+
+The `MigrationManager` does not implement a write-ahead log of its own; it relies on memiavl's existing changelog WAL for one side and flatkv's commit-snapshot mechanism for the other. There is no two-phase commit across them.
+
+This is the physical root cause of `A1`, `A2`, and `A3` below.
+
+---
+
+## 4. Failure Mode Catalog
+
+Each entry uses this format:
+
+- **Trigger** -- what physical event causes the failure
+- **Disk symptom** -- what an operator would observe on disk
+- **User symptom** -- what an end-user (RPC caller) would observe
+- **Self-recoverable?** -- whether normal restart + continued block production fixes it
+- **Detection signal** -- how to tell from outside that this happened
+- **Recovery path** -- the sequence of operator/tool steps to fix
+
+### A1 -- Lost-batch crash (memiavl wrote, flatkv didn't)
+
+- **Trigger**: process killed after memiavl writer commits the batch deletion but before flatkv writer commits the corresponding insert + boundary advance.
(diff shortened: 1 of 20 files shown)


sei-db/state_db/sc/migration/OPERATIONS.md
cody-littley, line 1: This is really nice!

cody-littley (approved):

Kbhat1 (approved):

PR #3419, opened 2026-05-11
Wire eth_subscribe(newHeads) for Autobahn (CON-257)

wen-coding:
## Summary

- Under Autobahn, `eth_subscribe("newHeads")` is currently broken: the legacy fan-out subscribes to the Tendermint event bus on `tm.event = 'NewBlockHeader'`, but Autobahn drives `app.FinalizeBlock` from `giga_router` rather than CometBFT consensus, so `PublishEventNewBlockHeader` never fires.
- This PR adds a direct in-process notifier (`BlockHeaderNotifier`) consumed by `evmrpc.SubscriptionAPI`'s fan-out goroutine. The notifier is fed from `App.Commit` itself — `FinalizeBlocker` stashes `(hash, header, response)` and `App.Commit` publishes after the underlying `BaseApp.Commit` succeeds. The legacy CometBFT path is untouched (notifier is only constructed when `tmConfig.AutobahnConfigFile != ""`).
- Notifier semantics: single-producer, single-consumer, capacity 1, overwrite-on-full — `newHeads` subscribers care about the latest head, not a backlog, so stale heads are dropped in favor of new ones.
- Zero sei-tendermint API additions. The notification is driven entirely by ABCI methods that already exist (`FinalizeBlock` + `Commit`); `giga_router` stays a pure ABCI consumer.

## Test plan

- [x] `gofmt -s -l` clean on all modified files; `go build ./...` clean.
- [x] `go test ./evmrpc/ ./sei-tendermint/internal/p2p/ ./sei-tendermint/internal/proxy/ ./sei-cosmos/baseapp/ ./app/` — all PASS.
  - existing `TestSubscribeNewHeads` (legacy event-bus path) — unchanged behaviour.
  - `TestSubscribeNewHeadsAutobahn` (notifier path, in-process WS server end-to-end).
  - unit tests for `BlockHeaderNotifier` (deliver, overwrite-on-full, nil-safe) and `encodeCommittedBlock` (happy path + zero gasLimit).
  - `pickHeadBaseFee` off-by-one coverage.
- [x] `evm_rpc_tests.sh` (HTTP-method fixtures) against a real non-Autobahn cluster — **160/160 pass**, no regressions on the request/response RPC surface.
- [x] `integration_test/evm_module/ws_test/` — consensus-mode-agnostic, gated on `SEI_EVM_WS_RUN_INTEGRATION=1`, wired into `evm_rpc_tests.sh`. End-to-end against both cluster modes:
  - Non-Autobahn cluster: PASS (`number=0x2b`, hash `0xe0dd089b...`)
  - Autobahn cluster: PASS (`number=0xf5`, hash `0xc4d1bc59...`, `GigaRouter initialized` confirmed in all 4 node logs)
- [x] Reviewer: confirm `non-app-hash-breaking` label is correct — this PR only adds a notification side-effect after commit; no state machine or app-hash-relevant logic changes.

---

> [!NOTE]
> **Medium Risk**
> Touches the block execution lifecycle by stashing FinalizeBlock data and publishing from `App.Commit`, which could affect head notification correctness/timing if mis-ordered or if Commit error paths differ. WS subscription behavior now has two code paths (event bus vs notifier) and relies on correct base-fee/gas-limit derivation for the new notifier stream.
> 
> **Overview**
> Fixes `eth_subscribe("newHeads")` under Autobahn by adding an in-process `BlockHeaderNotifier` that bypasses the Tendermint event bus.
> 
> `FinalizeBlocker` now stashes the (hash, header, finalize response) tuple and `App.Commit` publishes it *only after* a successful commit; stale stashes are defensively cleared to prevent announcing the wrong block. The WS server/`SubscriptionAPI` is extended to consume either the legacy event-bus stream or the new notifier stream (Autobahn), including a dedicated `encodeCommittedBlock` payload builder and a base-fee selection fix using the parent height.
> 
> Adds unit + integration coverage for notifier overwrite semantics, committed-head encoding, stale-stash clearing, and an end-to-end WS `newHeads` test for the notifier-backed path; integration scripts now run WS subscription tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fab0c5b142fa266b311c9b29101664b8f8d955de. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


evmrpc/subscribe.go
```diff
@@ -281,6 +303,57 @@ func (s *SubscriptionManager) Unsubscribe(ctx context.Context, id SubscriberID)
...
+		"stateRoot":             appHash,
+		"timestamp":             hexutil.Uint64(evt.header.Time.Unix()), //nolint:gosec
+		"transactionsRoot":      common.Hash{},                          // see function doc
+		"mixHash":               common.Hash{},                          // inapplicable to Sei
+		"excessBlobGas":         hexutil.Uint64(0),                      // inapplicable to Sei
+		"parentBeaconBlockRoot": common.Hash{},                          // inapplicable to Sei
+		"hash":                  blockHash,
+		"withdrawlsRoot":        common.Hash{}, // inapplicable to Sei
+		"baseFeePerGas":         (*hexutil.Big)(baseFee),
+		"withdrawalsRoot":       common.Hash{},     // inapplicable to Sei
```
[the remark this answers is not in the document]
wen-coding: done

codchen (approved):

evmrpc/subscribe.go
```diff
@@ -281,6 +313,56 @@ func (s *SubscriptionManager) Unsubscribe(ctx context.Context, id SubscriberID)
...
+	miner := common.BytesToAddress(evt.header.ProposerAddress)
+	appHash := common.BytesToHash(evt.header.AppHash)
+	var gasWanted int64
+	for _, txRes := range evt.response.TxResults {
+		gasWanted += txRes.GasUsed
+	}
+	var gasLimit uint64
+	if cp := evt.response.ConsensusParamUpdates; cp != nil && cp.Block != nil {
+		gasLimit = uint64(cp.Block.MaxGas) //nolint:gosec
+	}
```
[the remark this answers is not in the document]
wen-coding: Good catch, fixed

evmrpc/subscribe.go
```diff
@@ -281,6 +321,56 @@ func (s *SubscriptionManager) Unsubscribe(ctx context.Context, id SubscriberID)
...
+// would be nil for the vast majority of blocks.
+func encodeCommittedBlock(evt blockHeaderEvent, baseFee *big.Int, gasLimit int64) map[string]interface{} {
+	blockHash := common.BytesToHash(evt.hash)
+	number := big.NewInt(evt.header.Height)
+	miner := common.BytesToAddress(evt.header.ProposerAddress)
+	appHash := common.BytesToHash(evt.header.AppHash)
+	var gasWanted int64
+	for _, txRes := range evt.response.TxResults {
+		gasWanted += txRes.GasUsed
+	}
```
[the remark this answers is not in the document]
wen-coding: done

integration_test/evm_module/ws_test/ws_test.go
```diff
@@ -0,0 +1,109 @@
...
+		"id":      1,
+		"method":  "eth_subscribe",
+		"params":  []string{"newHeads"},
+	}); err != nil {
+		t.Fatalf("write subscribe: %v", err)
+	}
+
+	// First message must be the subscription confirmation. Bound the wait
+	// so a broken handshake fails fast.
+	if err := conn.SetReadDeadline(time.Now().Add(10 * time.Second)); err != nil {
```
amir-deris: This err shadows definition of err on line 42. Can just reuse it (similarly in other lines):
`if err = conn.SetReadDeadline....`

amir-deris (approved):

integration_test/evm_module/ws_test/ws_test.go
wen-coding: done

arajasek (approved):

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -42,6 +42,11 @@ type GigaRouterConfig struct {
 	Producer       *producer.Config
 	TxMempool      *mempool.TxMempool
 	GenDoc         *types.GenesisDoc
+
+	// BlockHeaderListener, if non-nil, is invoked after each block is
+	// committed. Used to feed evmrpc's eth_subscribe("newHeads") without
+	// going through the legacy Tendermint event bus.
+	BlockHeaderListener types.BlockHeaderListener
```
pompon0: Option[...]

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -287,6 +292,21 @@ func (r *GigaRouter) executeBlock(ctx context.Context, b *atypes.GlobalBlock) (*
 	if err != nil {
 		return nil, fmt.Errorf("r.cfg.App.Commit(): %w", err)
 	}
+	// Block-header listener fires here — after a successful Commit (so
+	// subscribers only see committed state) and before the mempool
+	// reconciliation below (which is bookkeeping, not part of the block
+	// itself). On a Commit error we return early and the listener does
+	// not fire — there is no committed block to announce.
```
pompon0: this looks suspicious: afaik Header.AppHash should be the output of the PREVIOUS block

pompon0: also header construction is duplicated

sei-tendermint/node/public.go
```diff
@@ -32,6 +32,13 @@ func New(
 	nodeMetrics *NodeMetrics,
 	consensusPolicy tmtypes.ConsensusPolicy,
 ) (service.Service, error) {
+	// Capture any optional BlockHeaderListener implementation before
+	// wrapping the app with the ABCI proxy, which only forwards ABCI
+	// methods.
```
pompon0: this looks bad - either add a method returning BlockHeaderListener to Application interface, or add a separate argument to New

sei-tendermint/node/seed.go
```diff
@@ -87,6 +87,7 @@ func makeSeedNode(
 		utils.None[*mempool.TxMempool](),
 		genDoc,
 		dbProvider,
+		nil,
```
pompon0: utils.None

sei-tendermint/node/public.go
pompon0: I feel like adding OnBlockCommitted to Application should be good enough

sei-tendermint/internal/p2p/giga_router.go
wen-coding: good catch, changed

sei-tendermint/node/seed.go
wen-coding: done

sei-tendermint/node/public.go
wen-coding: Adding OnBlockCommitted would require us to touch every App impl.

But I got your point. Thought about it more and decided to override Commit in App to push, see if this looks better? We don't need the duct tape any more.

evmrpc/subscribe.go
```diff
@@ -281,6 +337,61 @@ func (s *SubscriptionManager) Unsubscribe(ctx context.Context, id SubscriberID)
...
+	var totalGasUsed int64
+	for _, txRes := range evt.response.TxResults {
+		totalGasUsed += txRes.GasUsed
+	}
+	return map[string]interface{}{
+		"difficulty":            (*hexutil.Big)(utils.Big0),   // inapplicable to Sei
+		"extraData":             hexutil.Bytes{},              // inapplicable to Sei
+		"gasLimit":              hexutil.Uint64(gasLimit),     //nolint:gosec
+		"gasUsed":               hexutil.Uint64(totalGasUsed), //nolint:gosec
+		"logsBloom":             ethtypes.Bloom{},             // inapplicable to Sei
```
masih: We are planning to fix this in the future right?

evmrpc/subscribe.go
```diff
@@ -281,6 +337,61 @@ func (s *SubscriptionManager) Unsubscribe(ctx context.Context, id SubscriberID)
...
+// (see runNewHeadsFromNotifier); ConsensusParamUpdates on the response
+// would be nil for the vast majority of blocks.
+func encodeCommittedBlock(evt blockHeaderEvent, baseFee *big.Int, gasLimit int64) map[string]interface{} {
+	blockHash := common.BytesToHash(evt.hash)
+	number := big.NewInt(evt.header.Height)
+	miner := common.BytesToAddress(evt.header.ProposerAddress)
+	appHash := common.BytesToHash(evt.response.AppHash)
+	var totalGasUsed int64
+	for _, txRes := range evt.response.TxResults {
+		totalGasUsed += txRes.GasUsed
```
masih, line 372: Is gas used from tx result always right? Thinking of specific edgecases [here](https://github.com/sei-protocol/sei-chain/blob/13e55134f19a2d58092e98787eb0a7ac88a7cafc/evmrpc/block.go#L417)

masih (approved):

evmrpc/subscribe.go
wen-coding: changed to TODO

evmrpc/subscribe.go
wen-coding: hmm, good catch, but this is too large to fix in this PR, added a TODO as well

app/app.go
```diff
@@ -717,6 +739,9 @@ func New(
 	if err != nil {
 		panic(fmt.Sprintf("error reading EVM config due to %s", err))
 	}
+	if tmConfig != nil && tmConfig.AutobahnConfigFile != "" {
```
pompon0, line 749: this is rather ugly that App needs to understand that it will interact with autobahn vs tendermint. Can we just make it always work instead?

wen-coding: You mean have non-Autobahn also ditch event bus and move here? It's possible but it might bring in semantics change, which is why I'm a bit hesitant. In this current code, there are a few header fields we haven't figured out how to properly populate yet.
Can I add a TODO for now, if we figure out it's semantically equivalent, then switch non-Autobahn?

pompon0: yes, it would be great do ditch event bus, and it would be great to use shared logic with sei v2, so that the compatibility is more obvious. TODO is fine with me.

app/app.go
```diff
@@ -448,9 +456,23 @@ type App struct {
...
+	// going through the Tendermint event bus.
+	blockHeaderNotifier *evmrpc.BlockHeaderNotifier
+	// pendingHeadEvent holds the FinalizeBlock outputs awaiting the next
+	// Commit. Access is unsynchronized: callers MUST serialize each
+	// FinalizeBlock+Commit pair (GigaRouter.executeBlock does this under
+	// TxMempool.Lock; CometBFT consensus is naturally serial). Cleared
+	// defensively at FinalizeBlocker entry so a stale tuple from a prior
+	// non-stashing return path (EthReplay/EthBlockTest) or a failed Commit
+	// cannot be republished on a later block.
+	pendingHeadEvent      *pendingHeadEvent
```
pompon0: how about we hide the pending event within BlockHeaderNotifier and actually add a mutex to protect the access? Imo it would make the semantics more clear.

wen-coding: done

PR #3425, opened 2026-05-13
Add GoReleaser release pipeline for static seid binaries

monty-sei:
## What
Publishes a pre-built, statically-linked `linux/amd64` `seid` (plus `checksums.txt`) on every GitHub Release, attached automatically alongside the existing Docker images.

## How
- **Static build:** `scripts/build-static.sh` builds `seid` in Alpine (musl-native) as a GoReleaser `before:` hook. (Ubuntu's musl-gcc can't fully static-link this; Alpine's GNU ld + musl can.)
- **OSS packaging:** GoReleaser packages that binary via a `tool:` shim (`scripts/goreleaser-shim.sh`) instead of recompiling, an OSS-friendly stand-in for the Pro-only `prebuilt` builder, so the released binary is byte-identical to `make build`.
- **Auto-attach:** chained off the existing `version.json` / `release-publish` flow (uci `v0.0.11` exposes `version` + `release_created`). The goreleaser job runs with the exact published tag, and only when a release is actually cut. This avoids the tag-push trigger (suppressed for bot/`GITHUB_TOKEN` releases) and `checkout-latest-tag` (wrong across our `release/**` branches).
- **Manual hatch:** `goreleaser-release.yml` is a `workflow_dispatch` to (re)build and attach to an existing tag.
- **CI guard:** `cross-arch-build.yml` builds the static binary on release work and asserts it's actually statically linked.

## Notes for reviewers
- Depends on uci `v0.0.11` (merged).
- The static binary omits hardware-Ledger (`LEDGER_ENABLED=false`), contained to this artifact. `make build`/source is unchanged (Ledger-enabled), and a server `seid` doesn't need it.
- `linux/amd64` only; `linux/arm64` is a fast-follow (PLT-757). Docker images already cover arm64.
- Validated: the static build boots a local node (height 7,400+); local `goreleaser release --snapshot` produces a byte-identical static binary plus checksums; the full auto-attach was rehearsed in `sandbox` (the chain works end to end; the only hiccup was sandbox's immutable-releases setting, which sei-chain doesn't have: confirmed `immutable=false`). Final confirmation will be an RC.

Refs PLT-41, PLT-757.


.github/workflows/goreleaser-release.yml
```diff
@@ -0,0 +1,24 @@
...
+permissions:
+  contents: write
+
+jobs:
+  goreleaser:
+    uses: sei-protocol/uci/.github/workflows/goreleaser-release.yml@v0.0.10
+    with:
+      go-version: '1.25.6'
+      submodules: 'true'
+      prebuild-script: |
```
masih: Why prebuild script at all? I think you can encapsulate this entirely in the gorelease config instead of making it a parameter to the action?

masih: Discussed in 1:1; i see no blocker here. Approving and we can sort this out if needed later

masih (approved):

.github/workflows/uci-release-publish.yml
```diff
@@ -14,6 +14,16 @@ concurrency:
...
+
+  goreleaser:
+    name: GoReleaser
+    needs: publish
+    if: needs.publish.outputs.release_created == 'true'
+    uses: sei-protocol/uci/.github/workflows/goreleaser-release.yml@v0.0.11
+    with:
+      ref: ${{ needs.publish.outputs.version }}
+      go-version: '1.25.6'
+      submodules: 'true'
```
masih: non-blocker: I would read this as "git submodules"? if yes, do we have any submodules?

if no, I would rename this to avoid confusion with git terminology.

.github/workflows/cross-arch-build.yml
```diff
@@ -53,6 +53,53 @@ jobs:
...
+      - name: Checkout code
+        uses: actions/checkout@v5
+
+      # Build inside Alpine (musl-native): Ubuntu's musl-gcc can't fully static-link
+      # on 24.04 (glibc libgcc -> _dl_find_object, absent in musl), and zig cc rejects
+      # the -z muldefs flag needed to link the libwasmvm v152/v155 archives. Alpine's
+      # gcc is GNU ld + musl, so it links cleanly. Same script the goreleaser release
+      # build runs (scripts/build-static.sh); goreleaser then packages it via the tool shim.
+      - name: Build statically (Alpine / musl)
+        run: bash scripts/build-static.sh
```
masih, line 76: Wondering how we can further simplify the chain of bash+ shims etc. but a total non-blocker for first pass.

masih (approved):

monty-sei:
The `/backport` bot could not open this one: #3425 changes files under `.github/workflows/`, and GitHub blocks the backport App from pushing workflow changes without the `workflows` permission. Cherry-picked and pushed the backport manually instead: #3694 (base `release/v6.6`, clean cherry-pick of `b8f05ce`).

PR #3435, opened 2026-05-14
Cjl/litt integration

cody-littley:

- Make it so that LittDB is included in the regular build system. 
- Fixed all of the lint errors exposed by integrating the build.
- Causes CI to run LittDB tests. 
- sabled some tests that were pulling in a docker dependency. The tests were slow, mildly flaky, and the features they covered are not important to our use case.

unit test coverage, not yet used in prod


sei-db/db_engine/litt/util/ssh.go
```diff
@@ -42,7 +40,7 @@ func NewSSHSession(
 
 	var err error
 
-	hostKeyCallback := ssh.InsecureIgnoreHostKey()
+	hostKeyCallback := ssh.InsecureIgnoreHostKey() //nolint:gosec // overridden when knownHosts provided
```
blindchaser, line 43: [SUGGESTION] nolint:gosec for InsecureIgnoreHostKey() understates the risk

The comment says "overridden when knownHosts provided", but if a caller passes knownHosts == "" the insecure callback is what's actually used. This is pre-existing behavior, but if you're going to suppress the lint here, the nolint comment should reflect that fact (e.g. "intentional fallback; callers must supply knownHosts in production"), or — better — refuse to construct the session without knownHosts. Not introduced by this PR, but it's now being silenced rather than fixed.

sei-db/db_engine/litt/littdb_config.go
```diff
@@ -190,7 +188,7 @@ func DefaultConfigNoPaths() *Config {
 		ControlChannelSize:       64,
 		TargetSegmentFileSize:    math.MaxUint32,
 		MaxSegmentKeyCount:       50_000,
-		TargetSegmentKeyFileSize: 2 * units.MiB,
+		TargetSegmentKeyFileSize: 2 * unit.MB,
```
blindchaser, line 193: unit.MB is defined as 1024 * 1024 in sei-db/common/unit/data_units.go, which matches docker/go-units.MiB, so all sizes/throughputs are preserved. However, naming a 2^20 constant MB is the opposite of the SI convention and will eventually trip someone up. Not blocking — just worth a doc comment on the constant if you can't rename.

sei-db/db_engine/litt/disktable/segment/segment.go
```diff
@@ -446,7 +444,7 @@ func (s *Segment) Write(data *types.KVPair) (keyCount uint32, keyFileSize uint64
 	// Forward the value to the key and its address file control loop, which asynchronously writes it to the key file.
 	keyRequest := &types.ScopedKey{
 		Key:     data.Key,
-		Address: types.NewAddress(s.index, firstByteIndex, uint8(shard), uint32(len(data.Value))),
+		Address: types.NewAddress(s.index, firstByteIndex, uint8(shard), uint32(len(data.Value))), //nolint:gosec // shard <= 255, value len fits uint32
```
blindchaser: uint8(shard) is justified only because shard < ShardingFactor ≤ MaxShardingFactor = 256. The comment says "shard <= 255" which is right, but if a corrupted metadata file ever deserializes a shardingFactor > 256, this will silently truncate rather than fail loudly. Consider asserting shardingFactor <= MaxShardingFactor in loadMetadataFile/deserialize.

sei-db/db_engine/litt/disktable/table_metadata.go
```diff
@@ -139,7 +137,7 @@ func (t *tableMetadata) serialize() []byte {
 
 	// Write the TTL
 	ttlNanoseconds := t.GetTTL().Nanoseconds()
-	binary.BigEndian.PutUint64(data[4:12], uint64(ttlNanoseconds))
+	binary.BigEndian.PutUint64(data[4:12], uint64(ttlNanoseconds)) //nolint:gosec // serialized as time.Duration
```
blindchaser, line 142: uint64(ttlNanoseconds) and time.Duration(uint64) round-trip is fine, but TTL is signed; a negative TTL would deserialize to a huge positive duration. Probably impossible in practice, but worth validating TTL > 0 on load

blindchaser (approved):

sei-db/db_engine/litt/littdb_config.go
cody-littley: I'm willing to rename to `unit.MiB`. Ten second refactor in the IDE, but will yield hundreds of diffs scattered across the codebase. Would you be ok if this happens in a separate PR?

cody-littley: https://linear.app/seilabs/issue/STO-544/rename-unitsmb-to-unitsmib

sei-db/db_engine/litt/util/ssh.go
cody-littley: Created a follow up task for this. The ssh feature is not something we will be using. It was useful at Eigen since people wanted to use SSH to perform periodic DB backups, but we don't have that need here. If not actively disabled, this code won't even execute, so there is no risk in the short term.

https://linear.app/seilabs/issue/STO-545/safer-ssh-configuration

sei-db/db_engine/litt/disktable/segment/segment.go
cody-littley: The shard for each key is serialized/deserialized to/from a single byte, so it's impossible for data corruption to cause the value to exceed the max.

```
shardID:   bytes[8], // shardID is type uint8
```

You are correct that the metadata file could be corrupted in a way that leads to this violation getting broken. In order to avoid that, I've modified serialization/deserialization for the metadata to also store data in a single byte, making this sort of bounds overflow impossible.

sei-db/db_engine/litt/disktable/table_metadata.go
cody-littley: Added a check for negative values:

	intTTL := int64(binary.BigEndian.Uint64(data[4:12])) //nolint:gosec // serialized as time.Duration
	if intTTL < 0 {
		return nil, fmt.Errorf("TTL is negative: %d", intTTL)
	}
	ttl := time.Duration(intTTL)

sei-db/db_engine/litt/disktable/segment/segment.go
```diff
@@ -150,7 +148,7 @@ func CreateSegment(
 	shardSizes := make([]uint64, metadata.shardingFactor)
 
 	shardChannels := make([]chan any, metadata.shardingFactor)
-	for shard := uint32(0); shard < metadata.shardingFactor; shard++ {
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/db_engine/litt/disktable/segment/metadata_file.go
```diff
@@ -27,11 +25,11 @@ const (
 	// V3MetadataSize is the size of the metadata file at LatestSegmentVersion (ShardedAddressSegmentVersion).
 	// Layout:
 	//   - 4 bytes for version
-	//   - 4 bytes for the sharding factor
+	//   - 1 byte for the sharding factor
 	//   - 8 bytes for lastValueTimestamp
 	//   - 4 bytes for keyCount
 	//   - 1 byte for sealed
-	V3MetadataSize = 21
+	V3MetadataSize = 18
```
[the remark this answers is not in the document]
cody-littley: Not a problem. The current version `3` is already a bump compared the prior deployed version (i.e. in the EigenDA repo). Since this isn't deployed to production yet, it's safe to make as many serialization breaks as we need without a version bump. That will change after we have our first production deployment, though.

sei-db/db_engine/litt/disktable/control_loop.go
```diff
@@ -322,7 +320,7 @@ func (c *controlLoop) expandSegments() error {
 		c.highestSegmentIndex+1,
 		c.segmentPaths,
 		c.snapshottingEnabled,
-		c.metadata.GetShardingFactor(),
+		uint8(c.metadata.GetShardingFactor()), //nolint:gosec // bounded by litt.MaxShardingFactor (255)
```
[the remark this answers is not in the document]
cody-littley: Changed metadata file's data format to a `uint8`, making this sort of validation unnecessary.

Kbhat1 (approved):

PR #3437, opened 2026-05-14
Part 2: Bigtable node read fallback

Kbhat1:
## Description

- Wraps the composite state store: Get/Has below the local prune horizon route to Bigtable; recent state, writes, pruning, and iterators stay local.
- Coverage-gated for safety: an operator-declared floor opens RPC height gates, and reads ahead of ingestion error instead of returning empty state.
- LRU read cache + per-outcome metrics (cache hit / backend hit / miss / behind) to keep latency and Bigtable read costs measurable.

## Testing

- Verified on node


sei-db/config/toml.go
```diff
@@ -154,6 +154,23 @@ evm-ss-split = {{ .StateStore.EVMSplit }}
...
+historical-offload-bigtable-table = "{{ .StateStore.HistoricalOffloadBigtableTable }}"
+historical-offload-bigtable-family = "{{ .StateStore.HistoricalOffloadBigtableFamily }}"
+historical-offload-bigtable-app-profile = "{{ .StateStore.HistoricalOffloadBigtableAppProfile }}"
+historical-offload-bigtable-shards = {{ .StateStore.HistoricalOffloadBigtableShards }}
+
+# Earliest version (inclusive) fully ingested into the historical backend.
+# When > 0 it becomes the node's advertised earliest version, so RPC height
+# gates admit pruned heights the fallback can serve; point reads below it stay
+# local. Leave 0 until backfill/ingestion coverage reaches the desired height.
+historical-offload-earliest-version = {{ .StateStore.HistoricalOffloadEarliestVersion }}
```
yzang2019: I don't think we want to expose these config yet since it's not for mainnet. We should remove these from the config toml file to avoid being automatically generated

sei-db/state_db/ss/offload/historical/store.go
```diff
@@ -0,0 +1,332 @@
...
+		perKeyEarliest: perKeyEarliest,
+		reader:         reader,
+		cache:          cache,
+		metrics:        newFallbackMetrics(),
+		coverageFloor:  coverageFloor,
+	}
+}
+
+func (s *FallbackStateStore) shouldFallback(storeKey string, version int64) bool {
+	if version < 0 || (s.coverageFloor > 0 && version < s.coverageFloor) {
```
yzang2019, line 105: shouldFallback silently returns empty state for pruned keys below coverageFloor.

When version < coverageFloor, shouldFallback returns false, routing the read to the primary store. But the primary already pruned that version — that's why fallback exists. The caller gets back an empty []byte, nil with no indication the data was silently dropped. This seems to be a correctness bug: the node reports "key doesn't exist" for data that does exist in Bigtable.

sei-db/state_db/ss/offload/historical/store.go
```diff
@@ -0,0 +1,332 @@
...
+}
+
+func (s *FallbackStateStore) earliestVersionFor(storeKey string) int64 {
+	if s.perKeyEarliest != nil {
+		return s.perKeyEarliest.GetEarliestVersionForKey(storeKey)
+	}
+	return s.primary.GetEarliestVersion()
+}
+
+func (s *FallbackStateStore) readContext() (context.Context, context.CancelFunc) {
```
yzang2019, line 160: StateStore.Get/Has don't accept a context.Context. The internal readTimeout of 10s is hardcoded in newReader. If a block takes 200ms to finalize but a Bigtable read stalls, the node's state-query path blocks for 10s with no way for the caller to cancel. This also means graceful shutdown can't drain in-flight reads.

Consider a context-accepting internal path even if the StateStore interface can't change yet.

sei-db/state_db/ss/offload/historical/store.go
```diff
@@ -0,0 +1,332 @@
...
+func (s *FallbackStateStore) missExpired(key readCacheKey, value readCacheValue) bool {
+	if value.missExpiresAt.IsZero() || time.Now().Before(value.missExpiresAt) {
+		return false
+	}
+	s.cache.Remove(key)
+	return true
+}
+
+func (s *FallbackStateStore) cacheValue(key readCacheKey, value []byte) {
+	if value == nil || len(value) > maxReadCacheValueBytes {
```
yzang2019, line 236: cacheValue drops any value exceeding maxCacheableValueSize (8 KB). There's no metric or log for this. If a frequently-read key (e.g., a large EVM contract state) exceeds 8 KB, every read bypasses the cache and hits Bigtable. Under load this becomes the dominant source of backend traffic, and the operator has no visibility into it.

We can emit a metric (fallback_cache_skip_oversize) counting skipped entries. Consider a separate size-aware cache (e.g., groupcache or a byte-budget LRU) for large values instead of silently dropping them.

sei-db/state_db/ss/offload/historical/store.go
```diff
@@ -0,0 +1,332 @@
...
+		s.cacheHas(cacheKey)
+	} else {
+		s.metrics.recordRead("has", fallbackOutcomeBackendMiss)
+		s.cacheMiss(cacheKey)
+	}
+	return found, nil
+}
+
+func (s *FallbackStateStore) Iterator(storeKey string, version int64, start, end []byte) (dbm.Iterator, error) {
+	return s.primary.Iterator(storeKey, version, start, end)
```
yzang2019, line 281: GetEarliestVersion() returns coverageFloor when the primary's earliest is higher, implying the store can serve data from coverageFloor onward. But Iterator() and ReverseIterator() delegate straight to primary and never query Bigtable. A range query between coverageFloor and the primary's actual earliest version returns zero rows, violating the contract implied by GetEarliestVersion.

sei-db/state_db/ss/offload/historical/store.go
```diff
@@ -0,0 +1,332 @@
...
+var errBackendBehind = errors.New("historical backend behind requested version")
+
+// ensureBackendCoverage refuses fallback reads above the backend's last
+// ingested version so consumer lag surfaces as an error instead of silently
+// empty state (which would also poison the miss cache).
+func (s *FallbackStateStore) ensureBackendCoverage(ctx context.Context, version int64) error {
+	if version <= s.backendLast.Load() {
+		return nil
+	}
+	s.backendMu.Lock()
```
yzang2019, line 123: Every shouldFallback call hits ensureBackendCoverage, which grabs backendMu and does a synchronous reader.LastVersion() RPC when the check interval expires. During that RPC (up to 10s timeout), every goroutine doing a fallback read blocks on the mutex. Under high read concurrency this might become a bottleneck

sei-db/config/ss_config.go
```diff
@@ -80,6 +80,21 @@ type StateStoreConfig struct {
...
 	SeparateEVMSubDBs bool `mapstructure:"evm-separate-dbs"`
+
+	// HistoricalOffloadBigtableProjectID enables Bigtable fallback reads when
+	// project, instance, and table are set.
+	HistoricalOffloadBigtableProjectID  string `mapstructure:"historical-offload-bigtable-project-id"`
+	HistoricalOffloadBigtableInstance   string `mapstructure:"historical-offload-bigtable-instance"`
+	HistoricalOffloadBigtableTable      string `mapstructure:"historical-offload-bigtable-table"`
+	HistoricalOffloadBigtableFamily     string `mapstructure:"historical-offload-bigtable-family"`
+	HistoricalOffloadBigtableAppProfile string `mapstructure:"historical-offload-bigtable-app-profile"`
+	HistoricalOffloadBigtableShards     int    `mapstructure:"historical-offload-bigtable-shards"`
```
yzang2019, line 96: Suggestion:
1. Move these configs to one struct
2. Add a FallbackSSBackend = "BigTable" config, and default to empty (meaning disabled)

sei-db/state_db/ss/offload/historical/store.go
```diff
@@ -0,0 +1,332 @@
...
+		return false, err
+	}
+	found, err := s.reader.Has(ctx, storeKey, key, version)
+	if err != nil {
+		s.metrics.recordRead("has", fallbackOutcomeError)
+		return false, err
+	}
+	if found {
+		s.metrics.recordRead("has", fallbackOutcomeBackendHit)
+		s.cacheHas(cacheKey)
```
yzang2019, line 272: The common access pattern would be Has-then-Get, but then the issue here is that it would end up hitting BitTable twice.
 
Fix: either make Has() also fetch and cache the value, or let Get() recognize a found=true, valueKnown=false entry and upgrade it in-place rather than treating it as a full miss.

PR #3438, opened 2026-05-14
rpc sharding by tx sender for autobahn

pompon0:
For giga testnet we need every account to send transaction only to a single lane, so that they are included in nonce order. In this pr we are adding proxying of transactions to the correct lane (validator). How it works:
* only evmrpc endpoint is handled
* it is using http evmrpc connections for proxying (ws connections are also supported, but with the current impl they would be inefficient - see the comments in code)
* [evm account -> lane] mapping is deterministic and defined by autobahn committee
* evmrpc urls of validators of each lane are specified in autobahn config
* whenever user calls sendRawTransaction or getTransactionCount rpcs to a node, the lane of the evm account is computed. If the lane is produced by the local node, it is added to the mempool. Otherwise url of the lane owner is fetched and the same evmrpc is called on that url.
* If autobahn is not enabled, we fallback to the previous semantics.

---

> [!NOTE]
> **Medium Risk**
> Introduces request redirection for `eth_sendRawTransaction` and pending `eth_getTransactionCount`, which changes transaction submission/nonce semantics and adds cross-validator HTTP RPC dependencies driven by config.
> 
> **Overview**
> Adds **sender-based RPC sharding** for Autobahn by introducing `LocalClient.EvmProxy(sender)` and wiring it through the Tendermint local client/environment and `GigaRouter` (deterministic shard selection via `Committee.EvmShard`).
> 
> `evmrpc` now **redirects** `eth_sendRawTransaction` and pending `eth_getTransactionCount` to the shard owner’s EVM RPC HTTP endpoint when the sender maps to a different validator, with a fallback to existing local behavior when no proxy is configured or Autobahn isn’t active. Adds a new OpenTelemetry counter `evmrpc_redirected_requests_total`.
> 
> Extends Autobahn config generation/consumption to include per-validator `evmrpc` URLs (new `evmrpc_url.txt` in localnode init, new `evmrpc` field in config with URL marshal/unmarshal) and updates/expands tests to cover proxy behavior and the shard→URL mapping.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 33794ff7c8e634d33bccb95ce29ba12914ff270b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


evmrpc/send.go
```diff
@@ -77,6 +78,28 @@ func (s *SendAPI) SendRawTransaction(ctx context.Context, input hexutil.Bytes) (
 		return
 	}
 	hash = tx.Hash()
+	sender, err := getSender(tx, s.keeper.ChainID(s.ctxProvider(LatestCtxHeight)))
+	if err != nil {
+		return hash, err
+	}
```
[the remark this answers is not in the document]
pompon0: fixed

docker/localnode/scripts/step1_configure_init.sh
```diff
@@ -33,12 +33,14 @@ cp docker/localnode/config/config.toml "$TENDERMINT_CONFIG_FILE"
 SEI_NODE_ID=$(seid tendermint show-node-id)
 NODE_IP=$(hostname -i | awk '{print $1}')
 P2P_PORT=26656  # Must match [p2p] laddr in config.toml
+EVMRPC_PORT=8545  # Must match the EVM RPC HTTP port (evmrpc DefaultConfig HTTPPort).
```
wen-coding, line 36: nit: would be nice if both ports come from a common config

evmrpc/metrics.go
```diff
@@ -57,6 +58,11 @@ var (
 			metric.WithDescription("Number of new websocket connections"),
 			metric.WithUnit("{count}"),
 		)),
+		redirectedRequestCount: must(rpcTelemetryMeter.Int64Counter(
+			"evmrpc_redirected_requests_total",
+			metric.WithDescription("Number of EVM RPC requests redirected to another validator"),
```
wen-coding: nit: can you clarify whether this is request you forwarded to someone else or it is request you received from redirection?

evmrpc/send.go
```diff
@@ -77,25 +78,46 @@ func (s *SendAPI) SendRawTransaction(ctx context.Context, input hexutil.Bytes) (
 		return
 	}
 	hash = tx.Hash()
-	var txData ethtx.TxData
-	txData, err = ethtx.NewTxDataFromTx(tx)
-	if err != nil {
-		return
+	// getSender fails for AccessListTx, in which case we are not able to proxy or simulate,
+	// but we still need to handle it.
+	sender, senderErr := getSender(tx, s.keeper.ChainID(s.ctxProvider(LatestCtxHeight)))
```
wen-coding: Hmm, if LatestCtxHeight changes the sharding will suddenly change?

evmrpc/send.go
```diff
@@ -77,25 +78,46 @@ func (s *SendAPI) SendRawTransaction(ctx context.Context, input hexutil.Bytes) (
...
+	if senderErr == nil {
+		if url, ok := s.tmClient.EvmProxy(sender); ok {
+			recordRedirectedRequest(ctx, "eth_sendRawTransaction", string(s.connectionType))
+			// HTTP transport pooling already happens globally underneath net/http, so
+			// creating a fresh RPC client per proxied request is fine here. If we
+			// start proxying over WebSocket, we'll need explicit custom pooling since
+			// the underlying TCP connection lifecycle is strictly bound to Dial -> Close calls.
+			client, err := rpc.DialContext(ctx, url.String())
+			if err != nil {
+				return hash, fmt.Errorf("rpc.DialContext(%q): %w", url.String(), err)
```
wen-coding, line 93: Would we see one error log per tx when one validator is down? That would be too spammy

evmrpc/send.go
```diff
@@ -77,25 +78,46 @@ func (s *SendAPI) SendRawTransaction(ctx context.Context, input hexutil.Bytes) (
...
 	hash = tx.Hash()
-	var txData ethtx.TxData
-	txData, err = ethtx.NewTxDataFromTx(tx)
-	if err != nil {
-		return
+	// getSender fails for AccessListTx, in which case we are not able to proxy or simulate,
+	// but we still need to handle it.
+	sender, senderErr := getSender(tx, s.keeper.ChainID(s.ctxProvider(LatestCtxHeight)))
+	if senderErr == nil {
+		if url, ok := s.tmClient.EvmProxy(sender); ok {
```
wen-coding, line 85: What's our plan for handling the case where one validator is down?

evmrpc/send.go
```diff
@@ -125,30 +147,30 @@ func (s *SendAPI) SendRawTransaction(ctx context.Context, input hexutil.Bytes) (
...
-			err = fmt.Errorf("failed to get sender for dynamic fee tx: %w", err)
-			return
+			return common.Address{}, fmt.Errorf("failed to get sender for dynamic fee tx: %w", err)
 		}
-	} else if tx.Protected() {
-		signer := ethtypes.NewEIP155Signer(s.keeper.ChainID(s.ctxProvider(LatestCtxHeight)))
-		from, err = signer.Sender(tx)
+		return from, nil
+	case tx.Protected():
+		from, err := ethtypes.NewEIP155Signer(chainID).Sender(tx)
```
wen-coding: Would we always have NewEIP155Signer in tx.Protected()? Would `ethtypes.LatestSignerForChainID(chainID))` be better? So we accept non-1559 transactions?

evmrpc/send.go
```diff
@@ -125,30 +147,30 @@ func (s *SendAPI) SendRawTransaction(ctx context.Context, input hexutil.Bytes) (
 	return
 }
 
-func (s *SendAPI) simulateTx(ctx context.Context, tx *ethtypes.Transaction) (estimate uint64, err error) {
-	var from common.Address
-	if tx.Type() == ethtypes.DynamicFeeTxType {
-		signer := ethtypes.NewLondonSigner(s.keeper.ChainID(s.ctxProvider(LatestCtxHeight)))
-		from, err = signer.Sender(tx)
+func getSender(tx *ethtypes.Transaction, chainID *big.Int) (common.Address, error) {
+	switch {
```
wen-coding: Can we just use `LatestSignerForChainID(chainID).Sender(tx)` so we don't need to do the switch ourselves?

sei-tendermint/internal/autobahn/types/committee.go
```diff
@@ -87,6 +89,13 @@ func (c *Committee) FirstBlock() GlobalBlockNumber { return c.firstBlock }
 // GenesisTimestamp is the timestamp at genesis.
 func (c *Committee) GenesisTimestamp() time.Time { return c.genesisTimestamp }
 
+func (c *Committee) EvmShard(addr common.Address) PublicKey {
+	h := sha256.Sum256(addr[:])
+	x := new(big.Int).SetBytes(h[:])
+	i := int(x.Mod(x, big.NewInt(int64(c.replicas.Len()))).Int64())
+	return c.replicas.At(i)
```
wen-coding, line 96: We don't have that many validators so maybe we can use a cheaper algorithm?

i := int(binary.BigEndian.Uint64(addr[:8]) % uint64(c.replicas.Len()))

I suppose we will have a different algorithm when Autobahn hits mainnet, so we don't need to worry about attacker reverse-engineering this algorithm for now?

sei-will (approved):

docker/localnode/scripts/step1_configure_init.sh
pompon0: these are preexisting configs. Merging them would be backward incompatible.

evmrpc/metrics.go
pompon0: reworded to "forwarded to another validator"

evmrpc/send.go
pompon0: I'm not sure if we log all evmrpc RPC errors, but this will be consistent with the current behavior - we return error if tx does not land in the correct mempool

evmrpc/send.go
pompon0: for testnet - some accounts will be censored. We need to implement a proper solution afterwards.

evmrpc/send.go
pompon0: I have just copied over what simulateTx did. Is this not correct?

evmrpc/send.go
pompon0: done

sei-tendermint/internal/autobahn/types/committee.go
pompon0: good point, although imo it will not be a bottleneck and eventually it will be even more expensive.

evmrpc/send.go
wen-coding: I think LatestSignerForChainID(chainID) is a better approach guaranteed in go-ethereum library to handle all cases. I think the current code has a gap, but it only affected simulateTx's gas estimate, now we are skipping the proxy for it.

wen-coding (approved):

evmrpc/send.go
pompon0: make sense, switched to LatestSignerForChainID

PR #3440, opened 2026-05-14
blocksync: fix sendError in block sync (CON-276)

wen-coding:

`BlockPool.AddBlock`, `BlockPool.removeTimedoutPeers`, and `bpPeer.onTimeout` previously called `sendError` (a blocking send on `errorsCh`) while still holding `pool.mtx`. This change releases the mutex before the send.

In `AddBlock` and `removeTimedoutPeers` the send is registered as a `defer` *before* the `defer pool.mtx.Unlock()`, so LIFO ordering runs `Unlock` first and then performs the send. This keeps panic safety from the deferred unlock and avoids tracking an explicit `Unlock` on every return path. `onTimeout`'s critical section is a single boolean write, so it uses an explicit `Unlock` instead.

`AddBlock`'s return values and error wrapping are unchanged.

- `go test ./sei-tendermint/internal/blocksync/... -race -count=1` — full package green.
- New regression test `TestBlockPoolAddBlockReleasesLockBeforeSend` asserts the invariant for `AddBlock`. Verified discriminating by temporarily reverting the `AddBlock` edit — the test then blocks until the package timeout.
- `gofmt -s -l` clean.

### DIFF
--- a/sei-tendermint/internal/blocksync/pool_test.go
+++ b/sei-tendermint/internal/blocksync/pool_test.go
@@ -3,6 +3,7 @@ package blocksync
 import (
 	"crypto/rand"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"math"
 	"strings"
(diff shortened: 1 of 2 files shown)


sei-tendermint/internal/blocksync/pool_test.go
```diff
@@ -318,3 +319,58 @@ func testBlockPoolRejectsWrongPeerWithoutDiscardingGoodBlock(t *testing.T, goodF
...
+	requestsCh := make(chan BlockRequest, 1000)
+	pool := NewBlockPool(1, requestsCh, errorsCh, makeRouter(peers))
+	require.NoError(t, pool.Start(ctx))
+	t.Cleanup(func() { pool.Wait() })
+
+	pool.SetPeerRange(peerID, 1, 100)
+
+	errorsCh <- peerError{errors.New("filler"), peerID}
+
+	farHeight := int64(1 + maxDiffBetweenCurrentAndReceivedBlockHeight + 1000)
```
pompon0, line 350: please document what 1 + ... + 1000 is about

sei-tendermint/internal/blocksync/pool_test.go
```diff
@@ -318,3 +319,58 @@ func testBlockPoolRejectsWrongPeerWithoutDiscardingGoodBlock(t *testing.T, goodF
...
+	farBlock := &types.Block{Header: types.Header{Height: farHeight}}
+
+	addBlockDone := make(chan struct{})
+	go func() {
```
pompon0, line 354: spawn -> wait looks time sensitive, can you make it more robust?

sei-tendermint/internal/blocksync/pool_test.go
```diff
@@ -318,3 +319,58 @@ func testBlockPoolRejectsWrongPeerWithoutDiscardingGoodBlock(t *testing.T, goodF
...
+
+	<-errorsCh
+
+	select {
+	case <-addBlockDone:
+	case <-time.After(2 * time.Second):
+		t.Fatal("AddBlock did not complete after errorsCh was drained")
+	}
+
+	select {
```
pompon0: what's the point of this clause?

sei-tendermint/internal/blocksync/pool_test.go
```diff
@@ -318,3 +319,58 @@ func testBlockPoolRejectsWrongPeerWithoutDiscardingGoodBlock(t *testing.T, goodF
...
+
+	time.Sleep(200 * time.Millisecond)
+
+	probeDone := make(chan struct{})
+	go func() {
+		_ = pool.MaxPeerHeight()
+		close(probeDone)
+	}()
+
+	select {
```
pompon0: can we skip the individual timeouts? I'm pretty sure it will be flaky otherwise.

sei-tendermint/internal/blocksync/pool_test.go
pompon0: if not, then perhaps we shouldn't really have this test.

sei-tendermint/internal/blocksync/pool_test.go
wen-coding: done

sei-tendermint/internal/blocksync/pool_test.go
wen-coding: Changed to check runtime.Stack to detect sendError happened, how does this look?

sei-tendermint/internal/blocksync/pool_test.go
wen-coding: removed

sei-tendermint/internal/blocksync/pool_test.go
wen-coding: done

sei-will (approved):

codchen (approved):

PR #3443, opened 2026-05-15
feat(giga): don't fail if evmone can't load

arajasek:

This allows nodes on Ubuntu 22.04 or earlier to use Giga without having to install new deps.

Tests pass.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -811,6 +811,8 @@ func New(
 		// evmone is loaded best-effort
 		if evmoneVM, err := gigalib.InitEvmoneVM(); err == nil {
 			app.GigaEvmKeeper.EvmoneVM = evmoneVM
+		} else {
+			logger.Debug("failed to load evmone VM", "error", err)
 		}
 		if gigaExecutorConfig.OCCEnabled {
 			logger.Info("benchmark: Giga Executor with OCC is ENABLED - using new EVM execution path with parallel execution")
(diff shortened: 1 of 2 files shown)


app/app.go
```diff
@@ -776,11 +776,10 @@ func New(
 	app.GigaOCCEnabled = gigaExecutorConfig.OCCEnabled
 	tmtypes.SkipLastResultsHashValidation.Store(gigaExecutorConfig.Enabled)
 	if gigaExecutorConfig.Enabled {
-		evmoneVM, err := gigalib.InitEvmoneVM()
-		if err != nil {
-			panic(fmt.Sprintf("failed to load evmone: %s", err))
+		// evmone is loaded best-effort
+		if evmoneVM, err := gigalib.InitEvmoneVM(); err == nil {
```
wen-coding, line 812: Is this only invoked once on startup? Can we add a log here showing the error and say we are switching to geth executor?

BTW, is there any chance that we can't load either the evmone or the geth executor?

arajasek: Yes, it's only invoked on startup. Even if it successfully loads, we never use it (because it hasn't been tested yet). We only ever use the Geth executor (see [here](https://github.com/sei-protocol/sei-chain/blob/a4864c522263d67131bf72367f919ae81aad1765/app/app.go#L2045)).

I'm happy to add that as a debug log. I just don't want users to be confused and think something is wrong.

> BTW, is there any chance that we can't load either the evmone or the geth executor?

No, there's no "loading" of the geth executor, since it's just native golang.

arajasek: Added debug log

codchen (approved):

wen-coding (approved):

philipsu522 (approved):

PR #3451, opened 2026-05-16
evmrpc: drop dead isPanicOrSynthetic param from EncodeTmBlock (CON-296)

wen-coding:
## Summary

`EncodeTmBlock` has carried an `isPanicOrSynthetic func(...)` parameter since [PR #2058 (Jan 2025)](https://github.com/sei-protocol/sei-chain/pull/2058). [PR #2388 (Sep 2025)](https://github.com/sei-protocol/sei-chain/pull/2388)'s cherry-pick refactored the loop to source msgs from `filterTransactions` and dropped the in-loop call to `isPanicOrSynthetic`, but kept the parameter and the entire wiring chain that feeds it (`SeiBlockAPI.isPanicTx`, `NewSeiBlockAPI`/`NewSei2BlockAPI` constructor params, `getBlockByHash`/`getBlockByNumber` signatures, the closure in `server.go`).

This PR removes the dead parameter and its plumbing. No replacement check is added.

## Why no replacement check

The relevant intent statement is in [`evmrpc/README.md`](https://github.com/sei-protocol/sei-chain/blob/main/evmrpc/README.md)'s "Tracing Failure Management Endpoints" section:

> Due to Sei's unique mempool implementation and the absence of transaction simulation, some transactions may fail **pre-state checks**. These failures can occur due to:
> - Nonce mismatches ("nonce too low" or "nonce too high")
> - Insufficient funds
> - Other panic conditions
>
> These transactions are included in blocks but **not executed**. The following endpoints help filter out these failed transactions.

So `*ExcludeTraceFail` should exclude:

| Class | Already filtered by |
|---|---|
| Chain-generated synthetic (`TxType == ShellEVMTxType`) | `filterTransactions(includeSyntheticTxs=false)` — explicit check at [utils.go:262](https://github.com/sei-protocol/sei-chain/blob/main/evmrpc/utils.go#L262) |
| Ante / pre-state failures (`EffectiveGasPrice==0` + nonce VmError) | `filterTransactions` via `isReceiptFromAnteError` at [utils.go:249](https://github.com/sei-protocol/sei-chain/blob/main/evmrpc/utils.go#L249) |

What's *not* in scope per the README: in-VM failures like reverts. Those have valid traces (the REVERT path is rendered inside the trace output, `trace.Error` stays empty), and the legacy trace-based implementation included them. So `EncodeTmBlock` doesn't need any additional per-tx panic filter — the two classes the README cares about are both already dropped before they reach the loop body.

PR #2058's original logic (re-trace the block, filter on `trace.Error`) was consistent with this — `trace.Error` is set for ante failures (no execution to trace) but empty for reverts. PR #2388's silent regression (dead param) didn't change visible behavior on the block side, because `filterTransactions`'s `isReceiptFromAnteError` already covered the same set. So this PR is a no-op for behavior: it just removes the orphaned plumbing.

## Scope

- Removes `EncodeTmBlock`'s `isPanicOrSynthetic` parameter
- Removes `getBlockByHash` / `getBlockByNumber`'s `isPanicTx` parameter
- Removes `SeiBlockAPI.isPanicTx` field
- Removes `isPanicTx` from `NewSeiBlockAPI` / `NewSei2BlockAPI` constructor signatures
- Drops the `isPanicOrSyntheticTxFunc` arguments at `server.go`'s `NewSeiBlockAPI` / `NewSei2BlockAPI` call sites
- Drops the inline stub at `evmrpc/height_availability_test.go`
- Updates three test call sites of `EncodeTmBlock` to drop the `nil` argument

The tx-side `SeiTransactionAPI.isPanicTx` (used by `sei_getTransactionReceiptExcludeTraceFail`) is **kept**. That endpoint checks before fetching the receipt so it can return `ErrPanicTx` to the client — a different shape from the block path. The tx-side discriminator itself is being corrected separately in #3450.

## Test plan

- [x] `gofmt -s -l` clean on all modified files
- [x] `go build ./...` clean
- [x] `go test ./evmrpc/ -count=1` passes (22s)
- [x] `go test ./evmrpc/tests/ -count=1` passes (27s)

### DIFF
--- a/evmrpc/block.go
+++ b/evmrpc/block.go
@@ -83,7 +83,6 @@ type BlockAPI struct {
 
 type SeiBlockAPI struct {
 	*BlockAPI
-	isPanicTx func(ctx context.Context, hash common.Hash) (bool, error)
 }
 
 func NewBlockAPI(tmClient client.LocalClient, k *keeper.Keeper, ctxProvider func(int64) sdk.Context, txConfigProvider func(int64) client.TxConfig, connectionType ConnectionType, watermarks *WatermarkManager, globalBlockCache BlockCache, cacheCreationMutex *sync.Mutex) *BlockAPI {
@@ -108,7 +107,6 @@ func NewSeiBlockAPI(
 	ctxProvider func(int64) sdk.Context,
 	txConfigProvider func(int64) client.TxConfig,
 	connectionType ConnectionType,
-	isPanicTx func(ctx context.Context, hash common.Hash) (bool, error),
 	watermarks *WatermarkManager,
 	globalBlockCache BlockCache,
 	cacheCreationMutex *sync.Mutex,
@@ -127,8 +125,7 @@ func NewSeiBlockAPI(
 		cacheCreationMutex:   cacheCreationMutex,
 	}
 	return &SeiBlockAPI{
-		BlockAPI:  blockAPI,
-		isPanicTx: isPanicTx,
+		BlockAPI: blockAPI,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


wen-coding:
I mean, we don't need this, we need #3459

PR #3452, opened 2026-05-16
evmrpc: remove /block_results dependencies from block trace, sei_getCosmosTx, getHeader (CON-256)

wen-coding:
## Summary

Three eth-side block-scoped RPC paths leaned on `/block_results` data that isn't populated under Autobahn — they returned empty (or wrong) results regardless of block contents.

- Block trace backend (`debug_traceBlockBy*`, `sei_traceBlockBy*ExcludeTraceFail`) now iterates the block's own tx list.
- `sei_getCosmosTx` looks up the cosmos hash directly via `receipt.TransactionIndex`, no iteration needed.
- `getHeader` (trace path's eth-block construction) sources gasLimit from `sdkCtx.ConsensusParams()`, matching the [block.go:489-491](https://github.com/sei-protocol/sei-chain/blob/main/evmrpc/block.go#L489-L491) pattern already used by `eth_getBlockByNumber`. Under legacy this also fixes a latent silent fallback to `DefaultBlockGasLimit`.

Existing `.iox` fixtures only asserted response kind, so empty results / early-exit error paths looked indistinguishable from the fix; the trace fixtures now bind to the seeded deploy block and assert non-empty output, and a positive-case `getCosmosTx-deploy.iox` was added.

## Test plan

- [x] `go test ./evmrpc/` (22.3s)
- [x] `go test ./evmrpc/tests/` (27.7s)
- [x] Full RPC IO suite, docker localnet, Autobahn cluster: **161/161**
- [x] Full RPC IO suite, docker localnet, legacy cluster: **160/160** (run on the earlier 5-fixture state; new fixture additive)
- [x] Buggy binary (block-trace fix reverted) on Autobahn: **156/160** — exactly the four updated trace fixtures fail
- [x] Buggy binary (`sei_getCosmosTx` fix reverted) on Autobahn: **160/161** — exactly the new `getCosmosTx-deploy.iox` fixture fails
- [x] `gofmt -s -l` clean


evmrpc/simulate_test.go
```diff
@@ -24,6 +24,7 @@ import (
 	sdk "github.com/sei-protocol/sei-chain/sei-cosmos/types"
 	receipt "github.com/sei-protocol/sei-chain/sei-db/ledger_db/receipt"
```
amir-deris: While we are here, `receipt` seems to be redundant alias and can be removed.

evmrpc/simulate_test.go
```diff
@@ -24,6 +24,7 @@ import (
...
 	"github.com/sei-protocol/sei-chain/sei-tendermint/libs/bytes"
+	types2 "github.com/sei-protocol/sei-chain/sei-tendermint/proto/tendermint/types"
```
amir-deris: Would `tenderminttypes` be a better alias name?

amir-deris (approved):

evmrpc/simulate_test.go
wen-coding: removed

evmrpc/simulate_test.go
wen-coding: changed

pompon0 (approved):

PR #3432, opened 2026-05-13
blocksync: pre-decode CommitSig caps on inbound Block (CON-277)

wen-coding:
Adds a pre-decode validation hook on the blocksync channel that caps the
CommitSig count from each Commit source within an inbound Block at
`types.MaxVotesCount`, matching what `Commit.ValidateBasic` enforces
downstream. The check runs on the raw wire bytes before
`gogoproto.Unmarshal`. `Commit.ValidateBasic` also gains a direct count
check as a backstop.

Two Commit sources are covered:

- `Block.last_commit`
- `Block.evidence -> EvidenceList -> Evidence -> LightClientAttackEvidence
  -> LightBlock -> SignedHeader -> Commit`

Each source gets its own independent budget — the wireguard scanner keys
its counters by `(*Schema, field number)`, and the two paths use distinct
`Schema` values for the leaf Commit. Multiple
`LightClientAttackEvidence` entries within one Block share the
evidence-path budget.

The hook is implemented as a generic wire-format scanner
(`internal/protoutils/wireguard`) driven by a declarative `Schema`. Each
`Schema` describes the validation for a single proto type — which fields
to descend into and which repeated fields to cap — and Schemas are
composed by reference so a rule defined once applies wherever it is
referenced.

`MaxCount` is accumulated globally across the whole scan, keyed by
`(Schema, field number)`. That means if a peer encodes a parent message
multiple times (which proto decoders merge into one value by appending),
the inner field's counter still sees the cumulative total rather than
resetting per instance.

Proto field numbers are resolved at init via the generic
`wireguard.MustFieldNum[T any](name) Number`, which reflects on the
generated `protobuf` struct tags, so they track `.proto` regenerations.
Renaming or removing a referenced field panics at startup rather than
silently disabling the check.

The new `PreDecode utils.Option[func([]byte) error]` field on
`conn.ChannelDescriptorT` is the integration point; the blocksync channel
sets it to a `wireguard.Scan` over the blocksync schema. Other channels
can adopt the same pattern later.

### Tests

- Unit tests for the generic scanner: schema composition, `MaxCount`
  enforcement, cross-instance accumulation, duplicate-parent handling,
  distinct-schemas-share-no-counter, nested descent, malformed-wire
  rejection, `MustFieldNum` happy and panic paths.
- Unit tests for the blocksync schema, covering both Commit sources:
  at-cap accepted, over-cap rejected, separate budgets between
  `last_commit` and the evidence path, shared budget across multiple
[...]

PR #3458, opened 2026-05-18
wireguard: protoc extension and cap remaining inbound paths (CON-298)

wen-coding:
Extends the pre-decode wire scanner introduced in #3432 to the remaining
inbound paths that carry a `tmproto.Commit`:

- **Consensus DataChannel** — `tmcons.Proposal.last_commit` and
  `Proposal.evidence -> EvidenceList -> Evidence -> LightClientAttackEvidence
  -> LightBlock -> SignedHeader -> Commit`.
- **Consensus reassembled Block** (`state.getBlockFromBlockParts`) — the
  `tmproto.Block` reassembled from `BlockPart` bytes is scanned before
  `proto.Unmarshal` (covers `Block.LastCommit` and the evidence path).
- **Evidence channel** — an inbound `tmproto.Evidence` with the
  `LightClientAttackEvidence` variant is scanned to cap signatures inside
  the conflicting `LightBlock.SignedHeader.Commit`.
- **Statesync LightBlock channel** —
  `LightBlockResponse.light_block.signed_header.commit`.

Each cap is `types.MaxVotesCount`, mirroring `Commit.ValidateBasic`.

### Two approaches, side-by-side for review

The branch has two commits so reviewers can compare:

1. **`cap commit signatures on consensus, evidence, and statesync channels`** —
   hand-written `tmschemas/channels.go` composing the schemas explicitly per
   channel.
2. **`wireguard: emit channel Schemas from proto annotations`** —
   replaces the hand-written builders with a protoc plugin (modeled on
   `internal/hashable/plugin`). A single annotation on
   `Commit.signatures` (`(wireguard.max_count) = 10000`) drives schema
   generation for every type that transitively reaches a Commit.
   Descents are auto-discovered. `tmschemas/channels.go`'s validators
   become one-liners pointing at the generated `SchemaFor*` roots.

The branch tip is approach (2). Approach (1) is preserved as the
intermediate commit for comparison; either can ship.

### Notes on approach (2)

- One field option: `(wireguard.max_count) = N` on a repeated field.
  Descent is automatic — a message-typed field whose target has a schema
  gets a Nested rule.
- The plugin lives at `sei-tendermint/internal/protoutils/wireguard/plugin/`,
  invoked from `sei-tendermint/internal/wireguard.buf.gen.yaml`. Output
  is `*.wireguard.go` next to each `.pb.go`, same convention as the
  existing `hashable` plugin.
- A `--strict` flag (default off in this PR) makes the plugin error at
  codegen on any reachable repeated field that lacks `max_count`. With
  strict on, adding a new uncapped repeated field becomes a build break
  — exactly the "don't shoot ourselves later" property. Flipping it on
  requires auditing existing unannotated repeated fields across the
  proto tree (Data.txs, ValidatorSet.validators, ABCI events, etc.) and
  is tracked as a follow-up.
- One behavioral change vs approach (1): signature budgets are shared
  per `(*Schema, field)` rather than separate per source. The cumulative
  decoded-CommitSig count across `last_commit` and the evidence path
  inside one message is bounded by `MaxVotesCount`. Tighter ceiling than
  separate budgets; well above any legitimate validator-set size.

### Tests

`internal/protoutils/wireguard/tmschemas/channels_test.go` covers each
validator: at-cap acceptance, over-cap rejection on each Commit source,
shared-budget across `last_commit` + evidence path, shared budget across
multiple `LightClientAttackEvidence` entries, duplicate-non-repeated-field
rejection, and that off-path messages (`BlockPart`, `LightBlockRequest`,
`DuplicateVoteEvidence`) are walked past without rule matches.

`gofmt -s`, `go vet`, `golangci-lint` clean.


sei-tendermint/proto/tendermint/types/block.wireguard.go
```diff
@@ -0,0 +1,15 @@
...
+	utils "github.com/sei-protocol/sei-chain/sei-tendermint/libs/utils"
+)
+
+// SchemaForBlock is the wireguard.Schema generated for tendermint.types.Block.
+var SchemaForBlock = &wireguard.Schema{
+	Rules: map[wireguard.Number]wireguard.Rule{
+		wireguard.MustFieldNum[Block]("evidence"):    {Nested: utils.Some(SchemaForEvidenceList)},
+		wireguard.MustFieldNum[Block]("last_commit"): {Nested: utils.Some(SchemaForCommit)},
+	},
+}
```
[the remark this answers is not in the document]
wen-coding: This is intentional.

After thinking for a bit I was worried that:
proto X {
  repeated Y
}
proto Y {
  repeated Signaure
}
Could lead to huge memory usage if we allow 10k * number of Y.

The limit is intentionally set high so we should never need that much. And the global limit gives us a nice cap on memory upper limit.

sei-tendermint/proto/tendermint/consensus/wal.wireguard.go
```diff
@@ -0,0 +1,28 @@
...
+		wireguard.MustFieldNum[WALMessage_MsgInfo]("msg_info"): {Nested: utils.Some(SchemaForMsgInfo)},
+	},
+}
+
+// SchemaForTimedWALMessage is the wireguard.Schema generated for tendermint.consensus.TimedWALMessage.
+var SchemaForTimedWALMessage = &wireguard.Schema{
+	Rules: map[wireguard.Number]wireguard.Rule{
+		wireguard.MustFieldNum[TimedWALMessage]("msg"): {Nested: utils.Some(SchemaForWALMessage)},
+	},
+}
```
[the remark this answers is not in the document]
wen-coding: Intentional. Auto-descend emits a SchemaFor* for every message that transitively reaches a (wireguard.max_count) annotation; it doesn't try to figure out which channels actually wire them up. That's the design — annotations alone drive codegen, no hand-curated allowlist. The cost is ~6 extra MustFieldNum reflection calls at process startup and ~30 lines of generated Go. The benefit: if/when a new channel (e.g. privval RPC, WAL replay) wires a PreDecode, the schema is already there with no extra work.

sei-tendermint/internal/protoutils/wireguard/tmschemas/channels.go
```diff
@@ -0,0 +1,37 @@
...
+// ValidateEvidenceMessage is the PreDecode hook for the evidence channel.
+func ValidateEvidenceMessage(bz []byte) error {
+	return wireguard.Scan(bz, tmproto.SchemaForEvidence)
+}
+
+// ValidateStatesyncLightBlockChannel is the PreDecode hook for the
+// statesync LightBlock channel.
+func ValidateStatesyncLightBlockChannel(bz []byte) error {
+	return wireguard.Scan(bz, ssproto.SchemaForMessage)
+}
```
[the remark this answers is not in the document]
wen-coding: It's actually connected, adding a wiring test to prove.

sei-will (approved):

sei-tendermint/internal/protoutils/wireguard/plugin/main.go
```diff
@@ -0,0 +1,374 @@
...
+			return err
+		}
+	}
+
+	return emit(p, inSchema, byName, maxCountExt)
+}
+
+// validateMaxCountValues rejects (wireguard.max_count) = 0, which would
+// silently mean "no cap" at runtime (the wireguard.Scan check is
+// `if rule.MaxCount > 0`). If someone wants to forbid a repeated field
```
pompon0: nit: I don't think it is intuitive that MaxCount = 0 means unbounded for wireguard.Scan (but I suppose that's subjective). However omitting field from the proto means that the field will be ignored, which is not the same as returning error if it occurs.

sei-tendermint/internal/protoutils/wireguard/plugin/main.go
```diff
@@ -0,0 +1,374 @@
...
+			}
+			if opts.Get(ext.TypeDescriptor()).Uint() == 0 {
+				return fmt.Errorf("%s.%s: (wireguard.max_count) must be > 0", d.FullName(), f.Name())
+			}
+		}
+	}
+	return nil
+}
+
+func allMDs(files *protoregistry.Files) iter.Seq[protoreflect.MessageDescriptor] {
```
pompon0, line 180: there is a significant amount of duplicate code from hashable.hashable. Perhaps add a todo to dedup.

sei-tendermint/internal/protoutils/wireguard/plugin/main.go
```diff
@@ -0,0 +1,374 @@
...
+		g.P("package ", file.GoPackageName)
+		g.P()
+		wgIdent := g.QualifiedGoIdent(protogen.GoIdent{GoName: "Schema", GoImportPath: protogen.GoImportPath(wireguardRuntime)})
+		ruleIdent := g.QualifiedGoIdent(protogen.GoIdent{GoName: "Rule", GoImportPath: protogen.GoImportPath(wireguardRuntime)})
+		numIdent := g.QualifiedGoIdent(protogen.GoIdent{GoName: "Number", GoImportPath: protogen.GoImportPath(wireguardRuntime)})
+		mustFieldIdent := g.QualifiedGoIdent(protogen.GoIdent{GoName: "MustFieldNum", GoImportPath: protogen.GoImportPath(wireguardRuntime)})
+		utilsSomeIdent := g.QualifiedGoIdent(protogen.GoIdent{GoName: "Some", GoImportPath: protogen.GoImportPath(utilsPkg)})
+
+		for _, m := range targets {
+			emitSchema(g, m, byName, inSchema, ext, wgIdent, ruleIdent, numIdent, mustFieldIdent, utilsSomeIdent)
```
pompon0: these args seem to deserve a struct

pompon0: at least the identifier constants

sei-tendermint/internal/protoutils/wireguard/plugin/main.go
```diff
@@ -0,0 +1,374 @@
...
+	inSchema map[protoreflect.FullName]bool,
+	ext protoreflect.ExtensionType,
+	wgIdent, ruleIdent, numIdent, mustFieldIdent, utilsSomeIdent string,
+) {
+	// Use the descriptor from byName (which has dynamic extension options
+	// resolved) rather than m.Desc (protogen's view, which doesn't).
+	d := byName[m.Desc.FullName()]
+	g.P("// SchemaFor", m.GoIdent.GoName, " is the wireguard.Schema generated for ", d.FullName(), ".")
+	g.P("var SchemaFor", m.GoIdent.GoName, " = &", wgIdent, "{")
+	g.P("\tRules: map[", numIdent, "]", ruleIdent, "{")
```
pompon0: nit: aren't the output files gofmt'ed automatically anyway? You shouldn't need explicit indentation

sei-tendermint/internal/protoutils/wireguard/plugin/main.go
```diff
@@ -0,0 +1,374 @@
...
+		GoName:       "SchemaFor" + goIdent.GoName,
+		GoImportPath: goIdent.GoImportPath,
+	}
+	return g.QualifiedGoIdent(schemaIdent)
+}
+
+// descriptorGoIdent reverse-engineers the Go identifier for a proto message
+// from its descriptor: go_package option for the import path, and a
+// dot-joined CamelCase nesting for the name.
+func descriptorGoIdent(d protoreflect.MessageDescriptor) (protogen.GoIdent, bool) {
```
pompon0: unused?

sei-tendermint/internal/protoutils/wireguard/plugin/main.go
```diff
@@ -0,0 +1,374 @@
...
+		if !hasMax && nestedTarget == nil {
+			continue
+		}
+
+		// For a oneof variant, the wire tag is on the wrapper struct
+		// (e.g. Message_BlockResponse), not the parent message.
+		ownerType := g.QualifiedGoIdent(m.GoIdent)
+		if f.ContainingOneof() != nil {
+			ownerType = g.QualifiedGoIdent(protogen.GoIdent{
+				GoName:       m.GoIdent.GoName + "_" + pf.GoName,
```
pompon0, line 340: curious: do we really need to construct this identifier by hand? Is this because we generate against gogoproto artifacts?

sei-tendermint/internal/protoutils/wireguard/tmschemas/channels.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/sei-protocol/sei-chain/sei-tendermint/internal/protoutils/wireguard"
+	bcproto "github.com/sei-protocol/sei-chain/sei-tendermint/proto/tendermint/blocksync"
+	tmcons "github.com/sei-protocol/sei-chain/sei-tendermint/proto/tendermint/consensus"
+	ssproto "github.com/sei-protocol/sei-chain/sei-tendermint/proto/tendermint/statesync"
+	tmproto "github.com/sei-protocol/sei-chain/sei-tendermint/proto/tendermint/types"
+)
+
+// ValidateBlocksyncMessage is the PreDecode hook for the blocksync channel.
+func ValidateBlocksyncMessage(bz []byte) error {
+	return wireguard.Scan(bz, bcproto.SchemaForMessage)
```
pompon0: imo Schema should have just Scan method at this point, so that we don't need these boilerplate wrappers

sei-tendermint/internal/protoutils/wireguard/tmschemas/channels_test.go
```diff
@@ -0,0 +1,264 @@
...
+	return &bcproto.Message{Sum: &bcproto.Message_BlockResponse{
+		BlockResponse: &bcproto.BlockResponse{
+			Block: &tmproto.Block{
+				LastCommit: lastCommit,
+				Evidence:   tmproto.EvidenceList{Evidence: evidenceList(evidenceCommits...)},
+			},
+		},
+	}}
+}
+
```
pompon0: can we have these tests in the packages under test? I suppose mostly in sei-tendermint/types .

sei-tendermint/internal/protoutils/wireguard/tmschemas/wiring_test.go
```diff
@@ -0,0 +1,116 @@
...
+// grep — it catches the regression where someone removes or renames the
+// call, which is the most likely way to break the wiring.
+func TestWiring_ConsensusAssembledBlock(t *testing.T) {
+	bz, err := os.ReadFile("../../../consensus/state.go")
+	require.NoError(t, err, "could not read consensus/state.go to verify wiring")
+	require.Contains(t, string(bz), "tmschemas.ValidateConsensusAssembledBlock",
+		"consensus state.go does not reference tmschemas.ValidateConsensusAssembledBlock; the block-parts reassembly site lost its wireguard check")
+}
+
+// TestWiring_OtherChannelsHaveNoPreDecode documents which channels
```
pompon0: suggestion: if we are really afraid of that, we can do the following:
* add a proto message extension to explicitly declare which message types should have schema generated
* actually generate a `Wireguard() Schema` method (or equivalent, for example "Scan([]byte) error") for the message type
* make ChannelDescriptor require `interface { proto.Message; Wireguard() Schema }`, so that type system actually requires the guard.

This would restrict the interface a bit, but imo we do not expect anyone to write a wireguard method by hand.

Additionally we can make the "strict" mode more granular, by annotating particular proto messages as "strict" and apply this check only to them.

All of these are optional ideas, not strong suggestions.

pompon0 (approved):

sei-tendermint/internal/protoutils/wireguard/tmschemas/wiring_test.go
wen-coding: That's doable, but let's do it in a later PR when we cover the Autobahn messages.

sei-tendermint/internal/protoutils/wireguard/plugin/main.go
wen-coding: That's fair, changed codegen to reject MaxCount=0. We still maintain the semantics unannotated means ignore, we can raise an error when we enforce strict mode.

sei-tendermint/internal/protoutils/wireguard/plugin/main.go
wen-coding: Added a TODO.

sei-tendermint/internal/protoutils/wireguard/plugin/main.go
wen-coding: Done, added emitCtx and emitIdents.

sei-tendermint/internal/protoutils/wireguard/plugin/main.go
wen-coding: removed

sei-tendermint/internal/protoutils/wireguard/plugin/main.go
wen-coding: removed

sei-tendermint/internal/protoutils/wireguard/plugin/main.go
wen-coding: Good point, now we index *protogen.Message by Fullname at the top of run() and reuse m.GoIdent to qualify cross-file targets.

sei-tendermint/internal/protoutils/wireguard/tmschemas/channels.go
wen-coding: done

sei-tendermint/internal/protoutils/wireguard/tmschemas/channels_test.go
wen-coding: done, moved to sei-tendermint/proto/tendermint/types/

sei-tendermint/internal/protoutils/wireguard/wireguard.go
```diff
@@ -54,6 +54,13 @@ func Scan(bz []byte, schema *Schema) error {
 	return scan(bz, schema, map[counterKey]int{})
 }
 
+// Scan is the method form of the package-level Scan. It's the shape a
+// ChannelDescriptor's PreDecode hook expects, so the generated SchemaForX
+// values can be wired in directly without a wrapping closure.
+func (s *Schema) Scan(bz []byte) error {
+	return Scan(bz, s)
+}
```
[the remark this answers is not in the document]
wen-coding: This is intentional change to enforce a global limit. Yes the limit in PreDecode is stricter than Commit.ValidateBasic because it needs to put a cap on memory used. Since PreDecode already kicks out messages with excessive signatures, the check in Commit.ValidateBasic is basically a more relaxed defense in depth.

PR #3459, opened 2026-05-18
evmrpc: filter ante stubs from *ExcludeTraceFail (CON-296)

wen-coding:
## Summary

All three `*ExcludeTraceFail` endpoints (`sei_getTransactionReceiptExcludeTraceFail`, `sei_getBlockBy*ExcludeTraceFail`, `sei_traceBlockBy*ExcludeTraceFail`) leak non-nonce ante failures. The chain writes deferred-info stub receipts for them (`EffectiveGasPrice=0`, `GasUsed=0`); none of the existing filters catch this shape.

We can't widen `isReceiptFromAnteError` — [PR #2343](https://github.com/sei-protocol/sei-chain/pull/2343) narrowed it deliberately so regular `eth_getBlockBy*` keeps surfacing those receipts (real state changes from the nonce bump).

## Fix

One discriminator, three call sites:

```go
// utils.go
func isReceiptUntraceable(receipt *types.Receipt) bool {
    return receipt.TxType == types.ShellEVMTxType ||
        (receipt.EffectiveGasPrice == 0 && receipt.GasUsed == 0)
}
```

Applied at `isPanicOrSyntheticTx` (tx), `EncodeTmBlock` via new `excludeUntraceable bool` (block, `false` for regular endpoints), and `stripUntraceableTraces` (trace, both fresh and cached paths).

## Test coverage matrix

| Receipt class | Tx side | Block direct | Block HTTP | Trace HTTP |
|---|---|---|---|---|
| Synthetic (`TxType=Shell`) | ✓ | filterTransactions | — | filterTransactions |
| Ante stub (`EffGP=0, GasUsed=0`) | ✓ | ✓ | ✓ | ✓ |
| Revert / OOG (`EffGP>0, GasUsed>0`) | ✓ | ✓ | sanity baseline | — |
| Success (`Status=1`) | logical-identical to revert in discriminator | logical-identical | logical-identical | ✓ (kept alongside stub) |
| No receipt | ✓ (late-receipt test) | filterTransactions | — | filterTransactions; helper-internal parity |

Tests:
- `TestGetTransactionReceiptExcludeTraceFail` (tx-level table)
- `TestEncodeTmBlock_ExcludeUntraceable` (direct, both `true`/`false`)
- `TestGetBlockByHashExcludeTraceFail_AnteStub` (HTTP, asserts PR #2343 baseline + filter divergence in one test)
- `TestTraceBlockByNumberExcludeTraceFail_AnteStub` (HTTP, stub dropped + success kept)

Each fails when its discriminator is disabled. Full evmrpc suite passes.

### DIFF
--- a/evmrpc/block.go
+++ b/evmrpc/block.go
@@ -83,7 +83,6 @@ type BlockAPI struct {
 
 type SeiBlockAPI struct {
 	*BlockAPI
-	isPanicTx func(ctx context.Context, hash common.Hash) (bool, error)
 }
 
 func NewBlockAPI(tmClient client.LocalClient, k *keeper.Keeper, ctxProvider func(int64) sdk.Context, txConfigProvider func(int64) client.TxConfig, connectionType ConnectionType, watermarks *WatermarkManager, globalBlockCache BlockCache, cacheCreationMutex *sync.Mutex) *BlockAPI {
@@ -108,7 +107,6 @@ func NewSeiBlockAPI(
 	ctxProvider func(int64) sdk.Context,
 	txConfigProvider func(int64) client.TxConfig,
 	connectionType ConnectionType,
-	isPanicTx func(ctx context.Context, hash common.Hash) (bool, error),
 	watermarks *WatermarkManager,
 	globalBlockCache BlockCache,
 	cacheCreationMutex *sync.Mutex,
@@ -127,8 +125,7 @@ func NewSeiBlockAPI(
 		cacheCreationMutex:   cacheCreationMutex,
 	}
 	return &SeiBlockAPI{
-		BlockAPI:  blockAPI,
-		isPanicTx: isPanicTx,
+		BlockAPI: blockAPI,
 	}
 }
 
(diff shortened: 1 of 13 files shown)


sei-will (approved):

evmrpc/tracers.go
```diff
@@ -324,14 +323,37 @@ func filterExcludeFailFromBlockCache(cache *keeper.TraceDB, height int64, tracer
...
-		out = append(out, t)
+		if k != nil {
+			if receipt, err := k.GetReceipt(sdkctx, trace.TxHash); err == nil && isReceiptUntraceable(receipt) {
+				continue
+			}
+		}
+		out = append(out, trace)
 	}
-	return out, true
+	return out
```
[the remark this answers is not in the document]
wen-coding: done

sei-will (approved):

evmrpc/block.go
```diff
@@ -83,7 +83,6 @@ type BlockAPI struct {
 
 type SeiBlockAPI struct {
```
pompon0, line 84: nit: seems like BlockAPI can be used directly now

evmrpc/tracers.go
```diff
@@ -324,14 +331,46 @@ func filterExcludeFailFromBlockCache(cache *keeper.TraceDB, height int64, tracer
 	if err := json.Unmarshal(bz, &traces); err != nil {
 		return nil, false
 	}
+	return stripUntraceableTraces(traces, k, sdkctx), true
+}
+
+// stripUntraceableTraces filters out trace results that shouldn't surface from
```
pompon0: nit: the name looks confusing to me, but I have no idea what it should be called though.

pompon0 (approved):

evmrpc/tracers.go
wen-coding: renamed to dropUntraceableTraces

evmrpc/block.go
wen-coding: From data point of view you are correct. But README said everything with Sei* are Sei-specific interface, so removing this can be a bit confusing.

Anyways, I suggest we leave this cleanup for later when we are done with Autobahn integration tests, we will need to touch RPC next, and at that time we may be able to clean up a few Cosmos only RPC endpoints.

PR #3467, opened 2026-05-19
feat(metrics): migrate sei-cosmos to OpenTelemetry (PLT-353)

amir-deris:
Adds OTel instrumentation to `sei-cosmos` following the same pattern as PLT-329, PLT-330, PLT-336, PLT-339, and PLT-343.

## New instruments

**baseapp** (meter `seicosmos_baseapp`)
- `mid_block_duration` — histogram, seconds
- `end_block_duration` — histogram, seconds
- `deliver_tx_duration` — histogram, seconds
- `tx` — counter, total delivered transactions
- `tx_result` — counter, `result` label (successful/failed)
- `tx_gas_used` — gauge
- `tx_gas_wanted` — gauge
- `commit_duration` — histogram, seconds
- `abci_query_duration` — histogram, seconds, `path` label
- `process_proposal_duration` — histogram, seconds
- `finalize_block_duration` — histogram, seconds
- `get_tx_priority_hint_duration` — histogram, seconds
- `run_tx_duration` — histogram, seconds, `mode` label (replaces `MeasureThroughputSinceWithLabels` for `TxCount`)
- `run_msgs_duration` — histogram, seconds (replaces `MeasureThroughputSinceWithLabels` for `MessageCount`)
- `run_msg_latency` — histogram, seconds, `type` label (replaces both `sei.cosmos.run.msg.latency` and `cosmos.run.msg.latency`)

**storev2/rootmulti** (meter `seicosmos_storev2_rootmulti`)
- `sc_commit_latency` — histogram, seconds
- `ss_version` — gauge
- `historical_abci_query` — counter, `success` + `proof` labels
- `iavl_total_key_bytes` — gauge, `store_name` label
- `iavl_total_value_bytes` — gauge, `store_name` label
- `iavl_total_num_keys` — gauge, `store_name` label
- `state_sync_keys_exported` — counter

**tasks** (meter `seicosmos_tasks`)
- `scheduler_retries` — counter
- `scheduler_incarnations` — counter

**store/types** (meter `seicosmos_store_types`)
- `gas_exceeded` — counter, `error` + `descriptor` labels
- `bounded_cache` — gauge, `type` label

**x/upgrade** (meter `seicosmos_x_upgrade`)
- `begin_blocker_duration` — histogram, seconds
- `plan_height` — gauge, `name` label

**x/upgrade/keeper** (meter `seicosmos_x_upgrade_keeper`)
- `plan_height` — gauge, `name` label

**x/auth/vesting** (meter `seicosmos_x_auth_vesting`)
- `new_account` — counter
- `account_amount` — gauge, `denom` label

**x/bank/keeper** (meter `seicosmos_x_bank_keeper`)
- `send_amount` — gauge, `denom` label

**x/distribution/keeper** (meter `seicosmos_x_distribution_keeper`)
- `withdraw_reward_amount` — gauge, `denom` label
- `withdraw_commission_amount` — gauge, `denom` label

**x/staking/keeper** (meter `seicosmos_x_staking_keeper`)
- `delegate` — counter
- `delegate_amount` — gauge, `denom` label
- `redelegate` — counter
- `redelegate_amount` — gauge, `denom` label
- `undelegate` — counter
- `undelegate_amount` — gauge, `denom` label

**x/gov/keeper** (meter `seicosmos_x_gov_keeper`)
- `proposal` — counter
- `vote` — counter, `proposal_id` label
- `deposit` — counter, `proposal_id` label

## Notes
- All packages use dual-emit with `TODO(PLT-353)` comments pending dashboard verification.


sei-cosmos/baseapp/abci.go
```diff
@@ -388,7 +414,12 @@ func (app *BaseApp) Snapshot(height int64) {
 // Query implements the ABCI interface. It delegates to CommitMultiStore if it
 // implements Queryable.
 func (app *BaseApp) Query(ctx context.Context, req *abci.RequestQuery) (res *abci.ResponseQuery, err error) {
-	defer telemetry.MeasureSinceWithLabels([]string{"abci", "query"}, time.Now(), []metrics.Label{{Name: "path", Value: req.Path}})
+	queryStart := time.Now()
+	defer func() {
+		baseappMetrics.abciQueryDuration.Record(ctx, time.Since(queryStart).Seconds(), otelmetric.WithAttributes(attribute.String("path", req.Path)))
```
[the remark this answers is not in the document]
masih: This one is a blocker I'm afraid 

Cc @amir-deris

sei-cosmos/x/upgrade/keeper/metrics.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+var (
+	meter = otel.Meter("seicosmos_x_upgrade_keeper")
+
+	upgradeKeeperMetrics = struct {
+		planHeight metric.Int64Gauge
+	}{
+		planHeight: must(meter.Int64Gauge(
+			"plan_height",
```
bdchatham: **Instrument-name collision with `sei-cosmos/x/upgrade/metrics.go`.**

Both files declare a `Int64Gauge` named `plan_height` with identical `name` + `info` attribute schema. The OTel-Prometheus exporter does not propagate the meter scope (`seicosmos_x_upgrade` vs `seicosmos_x_upgrade_keeper`) into the Prometheus series name — both collapse to `sei_chain_plan_height`. Result is either a duplicate-registration warning with one side silently dropped, or two coexisting series differentiated only by `otel_scope_name` where naive PromQL `sum(sei_chain_plan_height)` double-counts.

The `x/upgrade/abci.go` BeginBlocker site is the load-bearing one (fires on every block while a plan exists). The keeper-side recording at `ScheduleUpgrade` is one-shot and redundant — drop it, or rename to something like `upgrade_pending_plan_height`.

Separately: the `info` label is a free-form proposal string. Cardinality-unfriendly and not useful as a metric axis — keep `name` (bounded by released upgrade handlers, ~20 lifetime values), drop `info`. Move it to a span attribute / event at upgrade-schedule time if you need it.

sei-cosmos/x/gov/keeper/msg_server.go
```diff
@@ -71,13 +77,17 @@ func (k msgServer) Vote(goCtx context.Context, msg *types.MsgVote) (*types.MsgVo
...
 
-	defer telemetry.IncrCounterWithLabels(
-		[]string{types.ModuleName, "vote"},
-		1,
-		[]metrics.Label{
-			telemetry.NewLabel("proposal_id", strconv.FormatUint(msg.ProposalId, 10)),
-		},
-	)
+	defer func() {
+		govMetrics.voteTotal.Add(goCtx, 1, otelmetric.WithAttributes(attribute.String("proposal_id", strconv.FormatUint(msg.ProposalId, 10))))
```
bdchatham: **Unbounded `proposal_id` label on `voteTotal` (and `depositTotal` below, and `VoteWeighted`).**

Every governance proposal ever cast a vote on lives in the TSDB forever; multiplied by per-validator series, this grows monotonically without bound. Sei mainnet is at ~200 historical proposals today and continues linear growth.

Drop the label. A plain `vote_total` / `deposit_total` counter answers "how many votes per block?" which is the only thing the metric is actually useful for. Per-proposal vote/deposit breakdown is a query-time concern against on-chain events — not a metric axis.

sei-cosmos/x/bank/keeper/msg_server.go
```diff
@@ -56,6 +58,8 @@ func (k msgServer) Send(goCtx context.Context, msg *types.MsgSend) (*types.MsgSe
 	defer func() {
 		for _, a := range msg.Amount {
 			if a.Amount.IsInt64() {
+				bankMetrics.sendAmount.Record(goCtx, a.Amount.Int64(), otelmetric.WithAttributes(attribute.String("denom", a.Denom)))
```
bdchatham: **Unbounded user-controlled `denom` label — most severe cardinality issue in this PR.**

Sei supports tokenfactory denoms (`factory/<creator>/<subdenom>` — any address can mint) plus IBC denoms with ~100-char hash traces. Sei mainnet already has 10k+ factory denoms, growing unbounded. This same pattern appears on `delegateAmount`/`redelegateAmount`/`undelegateAmount` (staking), `withdrawRewardAmount`/`withdrawCommissionAmount` (distribution), and `accountAmount` (vesting) — every `*Amount` gauge across this PR.

Options, in order of preference:
1. **Drop `denom` entirely** — record a per-module amount metric without the denom axis. If you need a breakdown, that's a log/span attribute.
2. **Bucket into `denom_class`** — small helper that returns `"usei"` / `"ibc"` / `"factory"` / `"other"` based on prefix. Preserves the useful axis (native vs synthetic flow) without exposing the registry to thousands of long-string labels.
3. If a specific dashboard needs aggregate-by-denom for native denoms, allowlist a fixed set via Prometheus `metric_relabel_configs` rather than emitting them as labels.

This would be a hard block on a single-metric PR; flagging now so it doesn't go out as part of a larger migration.

sei-cosmos/store/types/cache.go
```diff
@@ -65,6 +68,8 @@ func NewBoundedCache(backend CacheBackend, limit int) *BoundedCache {
 }
 
 func (c *BoundedCache) emitKeysEvictedMetrics(keysToEvict int) {
+	storeMetrics.boundedCache.Record(context.Background(), int64(keysToEvict), otelmetric.WithAttributes(attribute.String("type", "keys_evicted")))
```
bdchatham, line 71: nit: can we take a follow-up to wire the context through here or is there a blocker?

sei-cosmos/store/types/gas.go
```diff
@@ -113,6 +116,11 @@ func (g *basicGasMeter) ConsumeGas(amount Gas, descriptor string) {
 
 // cosmos_tx_gas_exceeded
 func (g *basicGasMeter) incrGasExceededCounter(errorType string, descriptor string) {
+	storeMetrics.gasExceeded.Add(context.Background(), 1, otelmetric.WithAttributes(
```
bdchatham, line 119: Same callout for context wiring.

sei-cosmos/storev2/rootmulti/store.go
```diff
@@ -132,7 +134,11 @@ func (rs *Store) Commit(bumpVersion bool) types.CommitID {
 		panic("Commit should always bump version in root multistore")
 	}
 	commitStartTime := time.Now()
-	defer telemetry.MeasureSince(commitStartTime, "storeV2", "sc", "commit", "latency")
+	defer func() {
+		storev2Metrics.scCommitLatency.Record(context.Background(), time.Since(commitStartTime).Seconds())
```
bdchatham, line 144: Same context wiring callout

sei-cosmos/baseapp/metrics.go
```diff
@@ -0,0 +1,127 @@
...
+		0.000025, 0.000050, 0.0001, 0.0005, 0.001, 0.0025, 0.005, 0.010, 0.020, 0.050, 0.075, 0.1, 0.25, 0.5, 1, 10,
+	)
+
+	baseappMetrics = struct {
+		midBlockDuration          metric.Float64Histogram
+		endBlockDuration          metric.Float64Histogram
+		deliverTxDuration         metric.Float64Histogram
+		txCount                   metric.Int64Counter
+		txResult                  metric.Int64Counter
+		txGasUsed                 metric.Int64Gauge
```
bdchatham, line 27: Walking this back partway — the description on `txGasUsed` ("Gas used by the **last** delivered transaction") and the staking `*Amount` gauges ("in the last delegation transaction") makes it clear the last-value semantic is intentional, matching the legacy `telemetry.SetGauge` paths. Not a bug.

Two things still worth considering:

**1. Description / implementation mismatch on the other amount gauges.** These say "by denomination" with no "last" — they read like flow metrics, but the impl is last-value:
- `bank/keeper/metrics.go` `sendAmount` — "Amount sent in MsgSend transactions by denomination"
- `distribution/keeper/metrics.go` `withdrawRewardAmount` — "Amount withdrawn as delegation rewards by denomination"
- `distribution/keeper/metrics.go` `withdrawCommissionAmount` — "Amount withdrawn as validator commission by denomination"
- `auth/vesting/metrics.go` `accountAmount` — "Amount funded into new vesting accounts by denomination"
- `store/types/metrics.go` `boundedCache` — "Bounded cache eviction events by type" (reads like an event counter)

Either tighten the descriptions to "...of the last X..." to match the gauge semantic, or switch these specific instruments to `Int64Counter` if cumulative flow is the actual intent. Right now they advertise one thing and emit another.

**2. Utility of last-value at scrape time, even when intentional.** "Last tx's gas" at a 15s scrape interval is effectively a random sample from the tail of each window — not bad for sanity-checking that gas-heavy txs exist, but not useful for trends or alerting. A `Float64Histogram` on the same record sites preserves the maximum-observed-in-window via `_max` aggregation AND gives you `_sum`, `_count`, distribution. Strictly more informative, ~30% more storage. Worth considering as an upgrade (not a fix).

Not blocking on either — but (1) is worth reconciling before merge.

sei-cosmos/store/types/cache.go
amir-deris: I also looked into not using `context.Background()`, but this Cache type is currently only used in tests, and threading context meaningfully from ABCI → KV store → cache backend is a non-trivial SDK-wide change with marginal metric value. I would prefer to leave that for now if that's ok?

sei-cosmos/storev2/rootmulti/store.go
amir-deris: I can address this one since the `historicalAbciQuery` counter would get exemplar linkage — we could correlate "this rate-limited query" directly to a specific gRPC trace and that's the most actionable OTel win here.

bdchatham:
Thanks for the quick turnaround — all four blockers look resolved. A few non-blocking callouts to consider for this PR or a followup:

- **Staking dropped the denom label entirely** while bank/distribution/vesting went with `denom_class`. Worth adding `denom_class` on staking too for forward-compat with LSM / non-usei staking — cheaper than a schema break later.
- **`DenomClass` lives in `sei-cosmos/telemetry/wrapper.go`** (the legacy go-metrics bridge) but it's OTel-only — probably belongs in its own file in the telemetry package.
- **Grafana panels don't surface OTel descriptions.** The tightened "...in the last X" wording is great for the data sheet but invisible in panel legends. Either rename to `*_amount_last` or document on the dashboards when we author them.
- **`abci_query_duration` bucket ceiling tops at 10s.** Archive store queries can run 10–60s — worth adding 30s/60s/120s buckets for this histogram now that `path` is bounded.
- **Recording rule + alert opportunity.** With `path` bounded, `route:sei_chain_abci_query_duration:p99_5m` is cheap to author, and an `AbciQueryRouteUnknownRatio` alert would catch new paths leaking into `*/unknown` before they go invisible.

LGTM otherwise. Happy to file these as separate issues if useful.

sei-cosmos/store/types/cache.go
bdchatham: Yeah, definitely okay to take as a follow-up for getting tracing wired

bdchatham (approved):

sei-cosmos/x/staking/keeper/msg_server.go
```diff
@@ -225,7 +225,11 @@ func (k msgServer) Delegate(goCtx context.Context, msg *types.MsgDelegate) (*typ
 
 	if msg.Amount.Amount.IsInt64() {
 		defer func() {
+			stakingMetrics.delegateTotal.Add(goCtx, 1)
+			// TODO(PLT-353): remove once staking_delegate_total verified
 			telemetry.IncrCounter(1, types.ModuleName, "delegate")
+			stakingMetrics.delegateAmount.Record(goCtx, msg.Amount.Amount.Int64())
```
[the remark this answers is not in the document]
amir-deris: The reason is the `delegateAmount`, `redelegateAmount` and `undelegateAmount` are in base denom (usei) and no need to include the attribute for this case.

amir-deris:
> Staking dropped the denom label entirely while bank/distribution/vesting went with denom_class. Worth adding denom_class on staking too for forward-compat with LSM / non-usei staking — cheaper than a schema break later.

For denom class, Delegate/Redelegate/Undelegate all operate on msg.Amount, which is validated against k.BondDenom(ctx) — always usei on sei-chain. Also the metric unit `{usei}` reflects that assumption so I didn't add denom class attribute for those metrics. Also updated the metric unit to `{utoken}` for cases where denom is specified.

> DenomClass lives in sei-cosmos/telemetry/wrapper.go (the legacy go-metrics bridge) but it's OTel-only — probably belongs in its own file in the telemetry package.

Good point! Moved it to its own file.

> Grafana panels don't surface OTel descriptions. The tightened "...in the last X" wording is great for the data sheet but invisible in panel legends. Either rename to *_amount_last or document on the dashboards when we author them.

Sounds good! Added `last` to metric names.

> abci_query_duration bucket ceiling tops at 10s. Archive store queries can run 10–60s — worth adding 30s/60s/120s buckets for this histogram now that path is bounded.

Added extra buckets for `abci_query_duration`.

> Recording rule + alert opportunity. With path bounded, route:sei_chain_abci_query_duration:p99_5m is cheap to author, and an AbciQueryRouteUnknownRatio alert would catch new paths leaking into */unknown before they go invisible.

Good point. Once we get to new dashboards/alerts, we can revisit this topic.

masih (approved):

PR #3469, opened 2026-05-20
test: enable Autobahn Upgrade and SeiDB CI, migrate cosmos bash off -b block (CON-256)

wen-coding:
## Summary

Enable three new Autobahn CI jobs:

- **Autobahn Upgrade Module (Major)**
- **Autobahn Upgrade Module (Minor)**
- **Autobahn SeiDB State Store**

Each was blocked by code or test-fixture issues that didn't surface under CometBFT. This PR fixes those issues so the jobs can join the matrix.

## What was blocking them

1. **`-b block` hangs under Autobahn** — the deploy and upgrade-proposal bash scripts these jobs depend on used `seid tx ... -b block`. Under Autobahn the KV indexer that `BroadcastTxCommit` polls isn't populated, so each call hung to its 60s timeout. Migrated to `-b sync` + poll on-chain side effects; patterns mirror lib.js's `storeWasm` / `instantiateWasm` / `findProposalByTitle` from #3406. New helpers landed in `integration_test/utils/_tx_helpers.sh` (the canonical home introduced by #3480).

2. **`proposal_target_height.sh` assumed 300ms blocks** — Autobahn runs ~400ms steady-state (and ~100ms during catchup bursts). The hardcoded estimate landed the proposal's target height too close to now, so `ScheduleUpgrade` rejected it as "scheduled in the past" once the voting period elapsed. Replaced with a self-calibrating sample (two heights 3s apart).

3. **Restart-bootstrap window was too tight** — under Autobahn the early-upgraded node needs longer than 30s to bootstrap before the cluster reaches `TARGET_HEIGHT`. Widened the upgrade-test yaml lookahead from 30s to 60s.

## Test plan

- [x] Locally on 4-node Autobahn docker cluster: deploy scripts exit=0 (wasm contracts, tokenfactory denoms, timelocked-token)
- [x] CI: CometBFT Upgrade Major/Minor green (regression check)
- [x] CI: Autobahn Upgrade Major/Minor + Autobahn SeiDB State Store green
- [x] No regression on existing CometBFT integration jobs


integration_test/utils/_tx_helpers.sh
```diff
@@ -133,3 +133,127 @@ submit_tx_and_wait() {
...
+            local id
+            for ((id=max_id_before+1; id<=cur; id++)); do
+                local observed; observed=$($seidbin q gov proposal "$id" -o json 2>/dev/null \
+                    | jq -r '.content.title // .title // ""')
+                if [ "$observed" = "$title" ]; then echo "$id"; return 0; fi
+            done
+            # Use max so a transient query failure can't shrink the
+            # window and let a prior-run proposal with the same title
+            # re-match on the next iteration.
+            max_id_before=$cur
```
amir-deris: `find_proposal_by_title` — silent miss on transient query failure 
 ` max_id_before=$cur`   # advances past the range even if a query inside the loop failed
  If `seid q gov proposal "$id`" returns empty on a transient error, observed="" doesn't match, and `max_id_before` gets bumped to `cur`. On the next iteration, that proposal ID is outside the scan window and permanently skipped. Fix: only update `max_id_before=$cur` after scanning all IDs in the range successfully.

amir-deris (approved):

masih (approved):

integration_test/utils/_tx_helpers.sh
wen-coding: Good catch, changed to only advance max_id_before when every id in the range was successfully scanned, transient empty responses leave the window unchanged so the missed id gets retried

PR #3473, opened 2026-05-20
testing(flatkv): evm migration int testings

blindchaser:
## Summary

This branch adds the FlatKV EVM Migrate path and the first reusable Docker migration test flow, using EVM state as the initial scenario. Validators start with EVM state committed through memIAVL, restart in `migrate_evm`, drain EVM account/code/storage keys into FlatKV incrementally as the restarted chain continues producing blocks, and verify that the post-migration FlatKV state is deterministic across validators. The core abstractions are a migration-aware composite store, explicit write-mode routing, a per-block migration manager, FlatKV initial-version seeding, and Docker-level migration orchestration.

Migration hardening added on top of `main`:
- Preserves pre-migration AppHash replay by keeping FlatKV's seeded lattice hash out of `LastCommitInfo` until FlatKV participates in committed state.
- Ensures migration progresses on empty/live blocks and keeps the migration tail bounded by routing live writes with boundary plus old-DB existence: existing not-yet-migrated keys stay in memIAVL, while brand-new keys go to FlatKV.
- Keeps legacy EVM prefix scans available during active migration by serving iterators from memIAVL, while exact reads remain migration-aware across both backends.
- Makes FlatKV `SetInitialVersion` durable for read-only tooling by writing seeded metadata and a seeded snapshot.
- Stabilizes Docker migration runs by requiring validators to stop from one committed height before flipping `sc-write-mode`, with retry instead of mixed-height restart.
- Reduces CI fixture OOG risk by splitting bulk EVM storage deployment into smaller batches.
- Hardens unrelated upgrade-log CI checks around height/log-format races so migration failures stay visible.

- `sei-db/config/toml.go`, `app/seidb.go`, `docker/localnode/config/app.toml`: add `sc-keys-to-migrate-per-block` so operators and tests can control per-block migration batch size. The config path enforces a single source of truth from TOML through `StateCommitConfig`.
- `sei-db/state_db/sc/composite/store.go`: builds routers after both backends are loaded, seeds a fresh FlatKV to the next memIAVL version on latest-version migration entry, and keeps targeted historical loads strict. It guarantees AppHash neutrality at the already-committed pre-migration height by only appending the FlatKV lattice hash after FlatKV participates in committed state.
- `sei-db/state_db/sc/composite/store.go`: routes child-store reads, writes, iteration, and proofs through `RouterCommitKVStore`. Empty blocks still call the router in active migration modes so migration progress is not tied to user writes.
- `sei-db/state_db/sc/migration/router_builder.go`: defines write-mode routing for `memiavl_only`, `migrate_evm`, `evm_migrated`, later migration modes, `flatkv_only`, and dual-write testing. The routing table makes EVM migration explicit and rejects unsupported routes loudly. Each router builder now passes the memIAVL iterator builder straight into `NewMigrationManager` instead of mutating the manager after construction.
- `sei-db/state_db/sc/migration/migration_manager.go`: manages the migration boundary, copies bounded batches from old DB to new DB, writes migrated keys to FlatKV, deletes copied old keys, and atomically writes `migration-version` while clearing the boundary on completion. `NewMigrationManager` now takes `oldDBIteratorBuilder` as a constructor parameter so the manager is fully wired before any caller can use it.
- `sei-db/state_db/sc/migration/migration_manager.go`: preserves live-block correctness by routing writes with both migration boundary and old-DB existence. Existing not-yet-migrated keys stay in old DB so their latest value is copied when the iterator reaches them; brand-new keys go to new DB so continuously-created EVM keys do not extend the migration tail forever. Exact reads fall back from old DB to new DB for not-yet-migrated ranges.
- `sei-db/state_db/sc/migration/migration_manager.go`: keeps legacy EVM prefix-scan paths available during the migration window via the old DB iterator, while rejecting iteration after completion and rejecting access to the reserved migration metadata store.
- `sei-db/state_db/sc/migration/migration_metrics.go`: encapsulates the per-run aggregator inside `MigrationMetrics`. `RecordBatch` now drives both OTel emission and the in-process completion summary, and exposes new counters for batches, original-pair routing (old/new DB), and per-DB pairs written. A nil metrics argument to `NewMigrationManager` is replaced with a local-only metrics so the completion log keeps working without an OTel exporter.
- `sei-db/state_db/sc/flatkv/store_meta.go`: adds durable `SetInitialVersion` seeding for FlatKV global and per-DB metadata, plus a seeded snapshot. This makes the next FlatKV commit land at the intended version and gives read-only tooling a stable seeded baseline.
- `sei-db/state_db/sc/flatkv/store_meta.go`: adds FlatKV latest-version helpers for startup sanity checks and recovery paths without requiring a full writable open first.
- `sei-db/tools/cmd/seidb/main.go`, `sei-db/tools/cmd/seidb/operations/migrate_evm_status.go`: add `seidb migrate-evm-status`, which opens FlatKV read-only through a snapshot/WAL clone and emits JSON status for `version_at`, `migration_version`, completion, and boundary presence.
- `Makefile`, `docker/docker-compose.yml`, `docker/localnode/scripts/step4_config_override.sh`: propagate `GIGA_MIGRATE_FROM_MEMIAVL` into validator containers and boot nodes in `memiavl_only` for migration CI. TOML edits are anchored to real keys to avoid matching comments.
- `.github/workflows/integration-test.yml`: adds the `FlatKV EVM Migrate` integration row. It boots a 4-validator cluster in `memiavl_only`, deploys an EVM fixture, runs the migration script, and verifies post-migration EVM state. This establishes the CI shape future FlatKV migration scenarios can reuse: prepare fixture, run coordinated mode flip, poll status, and verify post-migration state.
- `integration_test/contracts/deploy_flatkv_evm_fixture.sh`: creates a deterministic EVM fixture before migration, including account balance, contract code/storage, and a configurable bulk-storage workload split into small deploy batches to stay below EVM gas limits.
- `integration_test/contracts/verify_flatkv_evm_migrate.sh`: enforces the operator sequence for the Docker migration test: verify all nodes are in `memiavl_only`, wait for fixture blocks to settle, freeze validators at a common committed height, stop them cleanly, flip `sc-write-mode` to `migrate_evm`, restart, poll `seidb migrate-evm-status`, and compare FlatKV bucket digests at a shared height.
- `integration_test/contracts/verify_flatkv_evm_migrate.sh`: rejects unsafe stop conditions before flipping config. If validators commit an extra block during shutdown, the script restarts them in `memiavl_only`, waits for convergence, and retries rather than allowing mixed-height migration.
- `integration_test/contracts/verify_flatkv_evm_store.sh`: verifies that the pre-migration EVM fixture remains readable after migration and improves diagnostics around FlatKV bucket dumps.
- `scripts/evm_stress/main.go`: adds `contract-storage` workload mode, which deploys one-storage-slot contracts from many funded accounts to generate account, code, and storage rows for migration stress.
- `integration_test/upgrade_module/scripts/verify_upgrade_needed_log.sh`: hardens the upgrade log check against observed height/log-format races so unrelated upgrade CI does not hide migration signal.

## Test plan

- `sei-db/state_db/sc/composite/store_migration_test.go`: covers happy-path MemiavlOnly to MigrateEVM lifecycle, crash-and-resume, deterministic roots across independent stores, and post-completion `evm_migrated` mode flip. These tests enforce read transparency, persisted boundary recovery, and FlatKV root invariants.
- `sei-db/state_db/sc/composite/store_test.go`: covers AppHash-neutral migration entry, idempotent FlatKV seeding across restarts, child-store routing, write-mode behavior, and commit-info/lattice invariants.
- `sei-cosmos/storev2/rootmulti/flatkv_migration_test.go`: covers rootmulti-level migration lifecycle, empty-block progress, legacy EVM iteration during migration, AppHash determinism across runs, migration-version persistence, and post-completion mode flip.
- `sei-cosmos/storev2/rootmulti/flatkv_helpers_test.go`: adds shared rootmulti migration helpers for configs, restart, rollback, historical hash verification, and FlatKV offline inspection.
- `sei-db/state_db/sc/migration/migration_manager_test.go`: covers persisted boundary resume, per-block batching, completion atomicity, writer failure paths, randomized migration equivalence, new-key tail handling, existing-key old-DB routing, reserved migration-store protection, target-version acceptance after completion, and the constructor-supplied `oldDBIteratorBuilder` path used by post-completion iteration assertions. Run-stat assertions now read through `MigrationMetrics.RunStats()`.
- `sei-db/state_db/sc/migration/migration_transitions_test.go`: updates the direct migration transition test for the FlatKV EVM Migrate (0 -> 1) path.
- `sei-db/state_db/sc/flatkv/store_meta_test.go`: covers `SetInitialVersion` happy path, seeded snapshot creation, invalid version rejection, read-only rejection, post-commit rejection, reopen persistence, and rollback around seeded metadata.
- `sei-db/tools/cmd/seidb/operations/flatkv_open_test.go`: verifies read-only FlatKV tooling can open a store immediately after `SetInitialVersion` creates the seeded snapshot.
- `integration_test/contracts/verify_flatkv_evm_migrate.sh`: exercises the full Docker migration sequence, including common-height freeze, clean shutdown, restart in `migrate_evm`, status polling, completion summaries, and cross-validator FlatKV digest agreement.
- `integration_test/contracts/deploy_flatkv_evm_fixture.sh` and `integration_test/contracts/verify_flatkv_evm_store.sh`: validate fixture generation and post-migration EVM read transparency for balances, code, storage, and missing-account cases.
- Manual/local checks run during development: `bash -n integration_test/contracts/verify_flatkv_evm_migrate.sh`, focused Go tests for `./sei-db/state_db/sc/migration`, `./sei-db/state_db/sc/composite`, `./sei-cosmos/storev2/rootmulti`, `./sei-db/tools/cmd/seidb/operations`, and `./scripts/evm_stress`.


sei-db/state_db/sc/composite/store_migration_test.go
cody-littley, line 1: These tests duplicate tests I'm working on in a branch. But I think it's worth merging both. It's nice validation to have multiple test suites written independently, as it's less likely that we've both made a bad assumption as to how something should work.

sei-db/state_db/sc/migration/migration_manager.go
```diff
@@ -54,6 +58,33 @@ type MigrationManager struct {
...
+
+type migrationBatchStats struct {
+	keysMigrated             int64
+	keyBytesMigrated         int64
+	valueBytesMigrated       int64
+	originalPairsRoutedOldDB int64
+	originalPairsRoutedNewDB int64
+	oldDBPairsWritten        int64
+	newDBPairsWritten        int64
 }
```
cody-littley: Would it make sense to encapsulate this stuff inside `MigrationMetrics`? We could then expose metrics from these values, which might end up being useful if we encounter problems when we deploy this code. The unit tests could continue to access these values after the move.

sei-db/state_db/sc/migration/migration_transitions_test.go
```diff
@@ -12,7 +12,7 @@ import (
 // Test the MigrateEVM data migration. At the start of this migration, all data lives in memIAVL.
 // At the end of this migration, all evm/ data lives in flatkv, and all other data remains in memIAVL.
 //
-// This test evaluates the 0->1 migration path.
+// This test evaluates the FlatKV EVM migrate path.
```
cody-littley: Nit, could we add `(0 -> 1)` to this godoc? The migration versions are referenced in a bunch of other places, and I think it's a useful way of thinking about things sometimes.

sei-db/state_db/sc/migration/router_builder.go
```diff
@@ -131,7 +131,7 @@ func buildMemiavlOnlyRouter(
 	return router, nil
 }
 
-/* Data flow: MigrateEVM (0 -> 1)
+/* Data flow: FlatKV EVM migrate
```
cody-littley: Can we retain the `(0 -> 1)` part?

sei-db/state_db/sc/migration/router_builder.go
```diff
@@ -179,6 +179,7 @@ func buildMigrateEVMRouter(
 	if err != nil {
 		return nil, fmt.Errorf("NewMigrationManager: %w", err)
 	}
+	migrationManager.oldDBIteratorBuilder = buildMemIAVLIteratorBuilder(memIAVL)
```
cody-littley: Ok to add this, but we should probably make it a constructor parameter in the `NewMigrationManager()` method instead of directly setting it.

sei-db/state_db/sc/migration/migration_transitions_test.go
blindchaser: addressed

sei-db/state_db/sc/migration/router_builder.go
blindchaser: sure

sei-db/state_db/sc/migration/router_builder.go
blindchaser: good point

sei-db/state_db/sc/migration/migration_manager.go
blindchaser: updated

Kbhat1:
AI Review:

```migration_manager.go:453 rejects EVM iteration after migration completes, but EVM EndBlock still unconditionally iterates EVM prefixes. RouterCommitKVStore.Iterator panics on that error at .worktree/pr-3473/sei-db/state_db/sc/migration/router_kvstore.go:86. The next block after migrate_evm completes will hit paths like .worktree/pr-3473/x/evm/keeper/tx.go:11 and .worktree/pr-3473/x/evm/keeper/storage_cleanup.go:37, called from .worktree/pr-3473/x/evm/keeper/abci.go:84, and panic. Flipping to evm_migrated does not fix it because routeToFlatKV installs no iterator at .worktree/pr-3473/sei-db/state_db/sc/migration/router_builder.go:634. This blocks live post-migration operation.```

```sei-db/state_db/sc/composite/store.go:319 makes empty changesets mutate state in migration modes. Normal FinalizeBlocker returns an AppHash after app.GetWorkingHash() at .worktree/pr-3473/app/app.go:1438, but BaseApp.Commit calls app.Get before committing](pr-3473/sei-cosmos/baseapp/abci.go:268). That second call has no app writes left, but rootmulti.flush()still callsApplyChangeSets(nil)at [store.go](pr-3473/sei-cosmos/storev2/rootmulti/store.go:211), andMigrationManager.ApplyChangeSets` advances the migration batch at /state_db/sc/migration/migration_manager.go:283. So committed state can diverge from the AppHash already returnedto Tendermint for the block.```

File: sei-db/state_db/sc/migration/migration_manager.go [hunk 5]
```diff
@@ -382,9 +462,21 @@ func (m *MigrationManager) GetProof(store string, key []byte) (*ics23.Commitment
...
+	}
+	if m.oldDBIteratorBuilder == nil {
+		return nil, fmt.Errorf("iteration not supported for store %q", store)
+	}
+
+	// Preserve the legacy memiavl iterator path during the migration window.
+	// This keeps existing EndBlock cleanup code that still does prefix scans
+	// from panicking before the first migration batch commits, without adding a
+	// per-block full FlatKV scan on large stores.
+	return m.oldDBIteratorBuilder(store, start, end, ascending)
```
cody-littley, line 510: We should make this check stronger. It should only return the oldDB iterator if the migration boundary equals `MigrationBoundaryNotStarted`, as an iterator constructed at any time after migration actually starts will return invalid data.

sei-db/state_db/sc/composite/store.go
```diff
@@ -316,8 +337,22 @@ func (cs *CompositeCommitStore) buildRouter() error {
...
+//     Non-empty changesets always forward (caller writes must reach the
+//     backends regardless of which flush they arrive on).
 func (cs *CompositeCommitStore) ApplyChangeSets(changesets []*proto.NamedChangeSet) error {
-	if len(changesets) == 0 {
+	if cs.config.WriteMode.IsMigrationMode() {
+		if len(changesets) == 0 && cs.migrationForwardedThisCommit {
+			return nil
+		}
+		cs.migrationForwardedThisCommit = true
+	} else if len(changesets) == 0 {
```
cody-littley, line 354: If it's important to prevent us from migrating multiple batches per block, then we probably need a deeper fix. This only prevents migrating multiple batches from empty calls to ApplyChangeSets(), but doesn't protect us from scenarios where we are passed multiple non-empty change sets.

That being said, I'd like to understand the motivation for this change. Is doing multiple migration batches per block a problem with some workflow? Are the number of times `ApplyChangeSets()` is called not deterministic across nodes?

cody-littley: If we do need to enforce the `one migration batch per block` invariant, we probably need to adjust the router API to look like this:

```
	ApplyChangeSets(changesets []*proto.NamedChangeSet, firstBatchInBlock bool) error
```

Then, we only do the migration work the first time we call `ApplyChangeSets()` in each block, regardless of the number of changes provided.

sei-db/state_db/sc/migration/migration_manager.go
```diff
@@ -332,9 +375,44 @@ func (m *MigrationManager) ApplyChangeSets(changesets []*proto.NamedChangeSet) e
...
+		"keyBytesMigrated", stats.keyBytesMigrated,
+		"valueBytesMigrated", stats.valueBytesMigrated,
+		"originalPairsRoutedOldDB", stats.originalPairsRoutedOldDB,
+		"originalPairsRoutedNewDB", stats.originalPairsRoutedNewDB,
+		"oldDBPairsWritten", stats.oldDBPairsWritten,
+		"newDBPairsWritten", stats.newDBPairsWritten,
+		"elapsed", m.metrics.Elapsed())
+}
+
+func (m *MigrationManager) shouldWriteOriginalPairToNewDB(store string, key []byte) (bool, error) {
```
cody-littley: I suggest renaming this to `shouldForwardWriteToNewDB()`. I was initially confused by this since I misinterpreted the meaning of "original pair". Also, a short godoc would be nice.

sei-db/state_db/sc/migration/migration_manager.go
```diff
@@ -332,9 +375,44 @@ func (m *MigrationManager) ApplyChangeSets(changesets []*proto.NamedChangeSet) e
...
+	if m.boundary.IsMigrated(store, key) {
+		return true, nil
+	}
```
cody-littley, line 427:
```suggestion
	if m.boundary.IsMigrated(store, key) {
	    // Always forward writes to migrated keys to the new store.
		return true, nil
	}
```

sei-db/state_db/sc/migration/migration_manager.go
blindchaser: Good catch, agreed, anything past `NotStarted` would silently return incomplete results since left-of-boundary keys have already been deleted from the old DB. Tightened the check to `!m.boundary.Equals(MigrationBoundaryNotStarted)` and updated the godoc to spell out why.

sei-db/state_db/sc/migration/migration_manager.go
blindchaser: Done. renamed to `shouldForwardWriteToNewDB`

sei-db/state_db/sc/migration/migration_manager.go
blindchaser: updated

Kbhat1 (approved):

sei-db/state_db/sc/composite/store.go
blindchaser: addressed

sei-db/state_db/sc/migration/migration_manager.go
blindchaser: I tried tightening it to your suggestion and CI panicked at height 432:

panic: RouterCommitKVStore.Iterator(store="evm"):
       iteration not supported once migration has started
  at x/evm/keeper.RemoveFirstNTxHashes
  at EndBlock → ProcessBlock → consensus halt

EVM EndBlock does a per-block prefix scan to GC old tx hashes. The error becomes a panic because CommitKVStore.Iterator() has no error return.

The proper fix is a merged iterator (old + new DB, with tombstone masking), which we don't have yet. The only known caller (RemoveFirstNTxHashes) is best-effort GC, incomplete results just leave a few stale hashes, no consensus impact — but panicking out of EndBlock does kill consensus.

Reverting for now and keeping the original Complete-only check. Updated the godoc to flag the caveat so future callers don't assume completeness during InProgress. Merged iterator tracked as follow-up

cody-littley (approved):

PR #3476, opened 2026-05-20
TxMempool rewrite (CON-305)

pompon0:
Addressed a bunch of issues:
* diverging state between various indices in TxMempool
* \Omega(n) complexity of inserting a transaction above capacity
* delayed promotion of transactions from pending to ready (happens only on Update)
* removed mechanism which was skipping 1 high priority tx in case it did not fit into the current limit - this mechanism was ad hoc and it did not respect nonces

Not addressed:
* UnconfirmedTxs rpc call is still inefficient and hogs mempool
* poor behavior under autobahn: it uses ReapTxs(remove=true) so that producers do not include the same transactions in multiple blocks, but it also makes non-reaped higher nonce transactions pending (lower nonce txs have been reaped, but are not executed yet, so the account nonce did not move). For autobahn we will implement a simpler but more tighlty integrated mempool (because it needs no gossip and all proposals are roughly guaranteed to be sequenced, since every producer has their own lane)

Consistency has been achieved mostly by moving all the data under a single mutex in txStore. Amortized O(log n) updates were achieved by delaying pruning until mempool size reaches 2*capacity (more precisely capacity is counted both in number of txs and in bytes, so 1 enormous transaction can also trigger recomputation of the mempool, in which case we have amortized O(log n) per tx byte instead)

Additional changes:
* PendingSize and Size capacity limits have been merged into 1 limit = PendingSize + Size, same for byte size capacity limits
* removed tracking of peers that a given transaction was received from in favor of just gossiping to everyone. It is expected to increase the actual network traffic just by ~1/p where p is the number of peers.
* fixed node hanging introduced in https://github.com/sei-protocol/sei-chain/pull/3410


sei-tendermint/internal/mempool/tx.go
```diff
@@ -99,320 +83,493 @@ func (wtx *WrappedTx) EVMNonce() uint64 {
...
+		if remove {
+			for _, wtx := range wtxs {
+				delete(inner.byHash, wtx.Hash())
+				s.metrics.RemovedTxs.Add(1)
+				if el, ok := wtx.readyEl.Get(); ok {
+					s.readyTxs.Remove(el)
 				}
-				cb(ptx)
-				p.sizeBytes.Add(int64(-ptx.Size()))
+				s.compact(inner, false)
```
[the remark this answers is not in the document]
pompon0: fixed

sei-tendermint/internal/mempool/mempool.go
```diff
@@ -182,50 +166,18 @@ type TxMempool struct {
...
 
-	// cache defines a fixed-size cache of already seen transactions as this
-	// reduces pressure on the proxyApp.
-	cache TxCache
-
 	// blockFailedTxs tracks tx hashes that have previously failed during
 	// block execution. Used to prevent infinite re-entry of txs that
 	// consistently fail before fee charging in DeliverTx.
-	blockFailedTxs TxCache
+	blockFailedTxs *LRUTxCache
```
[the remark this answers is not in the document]
pompon0: done

sei-tendermint/internal/mempool/tx.go
```diff
@@ -99,320 +83,493 @@ func (wtx *WrappedTx) EVMNonce() uint64 {
...
+			delete(inner.byHash, old.Hash())
+			s.metrics.RemovedTxs.Add(1)
+			state.total.Dec(old.Size())
+			if oldReady {
+				state.ready.Dec(old.Size())
+				state.ready.Inc(wtx.Size())
+				if !wtx.readyEl.IsPresent() {
+					wtx.readyEl = utils.Some(s.readyTxs.PushBack(wtx.Tx()))
+				}
+			}
```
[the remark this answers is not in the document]
pompon0: done

sei-tendermint/internal/mempool/tx.go
```diff
@@ -99,320 +83,494 @@ func (wtx *WrappedTx) EVMNonce() uint64 {
...
+		total := inner.state.Load().total
+		total.Inc(wtx.Size())
+		if !total.LessEqual(&inner.softLimit) || s.insert(inner, wtx) != nil {
+			s.cache.Remove(wtx.Hash())
+			s.metrics.CacheSize.Set(float64(s.cache.Size()))
+			s.metrics.RemovedTxs.Add(1)
+			s.metrics.EvictedTxs.Add(1)
+			if el, ok := wtx.readyEl.Get(); ok {
+				s.readyTxs.Remove(el)
+			}
```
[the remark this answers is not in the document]
pompon0: this WAI, readyTxs list accumulates transactions of mempool that were ready at any point in the past. Reinserting them to the list would cause regossip.

sei-tendermint/internal/mempool/reactor/reactor_test.go
```diff
@@ -376,10 +372,10 @@ func TestReactorConcurrency(t *testing.T) {
 				deliverTxResponses[i] = &abci.ExecTxResult{Code: 0}
 			}
 
-			require.NoError(t, txmp.Update(ctx, 1, convertTex(txs), deliverTxResponses, mempool.NopTxConstraintsFetcher, true))
+			require.NoError(t, txmp.Update(ctx, 1, convertTex(txs), deliverTxResponses, mempool.NopTxConstraints(), true))
```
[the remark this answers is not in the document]
pompon0: fixed

sei-tendermint/internal/mempool/tx.go
```diff
@@ -99,320 +83,500 @@ func (wtx *WrappedTx) EVMNonce() uint64 {
...
+	}
+	for _, account := range inner.accounts {
+		account.nextNonce = account.firstNonce
+	}
+	for _, wtx := range wtxs {
+		total := inner.state.Load().total
+		total.Inc(wtx.Size())
+		limitOk := total.LessEqual(&inner.softLimit)
+		// NOTE: insertion is lazily evaluated here.
+		if !limitOk || s.insert(inner, wtx) != nil {
```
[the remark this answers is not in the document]
pompon0: this is not a real scenario, since reinserted txs during compaction never encounter nonce collisions

sei-tendermint/internal/mempool/mempool.go
```diff
@@ -237,88 +184,47 @@ type TxMempool struct {
...
-	if !ok {
-		return
-	}
-	an := evmAddrNonce{evm.address, evm.nonce}
-	for byAddrNonce := range txmp.byAddrNonce.Lock() {
-		if byAddrNonce[an] == wtx {
-			delete(byAddrNonce, an)
-		}
-	}
-}
```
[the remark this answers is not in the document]
pompon0: addressed

sei-tendermint/internal/mempool/tx.go
```diff
@@ -99,320 +83,510 @@ func (wtx *WrappedTx) EVMNonce() uint64 {
...
+			// NOTE: evicted txs are not cached unconditionally
+			if !limitOk || !s.config.KeepInvalidTxsInCache {
+				inner.cache.Remove(wtx.Hash())
+			}
+			s.metrics.RemovedTxs.Add(1)
+			s.metrics.EvictedTxs.Add(1)
+			if el, ok := wtx.readyEl.Get(); ok {
+				s.readyTxs.Remove(el)
+			}
```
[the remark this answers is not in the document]
pompon0: that's incorrect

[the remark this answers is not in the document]
pompon0: that's incorrect

sei-tendermint/internal/mempool/tx.go
```diff
@@ -99,320 +83,510 @@ func (wtx *WrappedTx) EVMNonce() uint64 {
...
 		}
-		inner.txs = append(inner.txs, tx)
-		p.sizeBytes.Add(int64(tx.Size()))
-		return nil
 	}
-	panic("unreachable")
+	s.metrics.CacheSize.Set(float64(inner.cache.Size()))
```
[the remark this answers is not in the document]
pompon0: they don't disappear because they are not removed from the list

sei-tendermint/internal/mempool/mempool.go
```diff
@@ -494,84 +388,56 @@ func (txmp *TxMempool) CheckTx(ctx context.Context, tx types.Tx, txInfo TxInfo)
...
+	//
+	// We do not use the priority hint here as it may be misleading and
+	// inaccurate. The true priority as determined by the application is the
+	// most accurate.
+	txmp.priorityReservoir.Add(wtx.priority)
 
-	// only add new transaction if checkTx passes and is not pending
-	if !txmp.isPending(wtx) {
-		if err := txmp.addNewTransaction(wtx); err != nil {
-			return nil, err
```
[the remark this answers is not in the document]
pompon0: that's a backward compatible behavior.

sei-tendermint/internal/libs/clist/clist.go
```diff
@@ -224,6 +194,21 @@ func (l *CList[T]) Back() *CElement[T] {
...
+	for el := l.head; el != nil; {
+		next := el.Next()
+		el.setRemoved()
+		el = next
+	}
+	l.waitCh = make(chan struct{})
+	l.head = nil
+	l.tail = nil
+	l.len = 0
+}
```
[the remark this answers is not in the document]
pompon0: that's plain bullshit

sei-tendermint/internal/mempool/mempool.go
```diff
@@ -430,43 +329,34 @@ func (txmp *TxMempool) CheckTx(ctx context.Context, tx types.Tx, txInfo TxInfo)
...
-		txmp.metrics.NumberOfLocalCheckTx.Add(1)
+		c.Increment(hTx.Hash())
 	}
 	res, err := txmp.app.CheckTxSafe(ctx, &abci.RequestCheckTxV2{Tx: tx})
 	if err != nil || !res.IsOK() {
+		txmp.txStore.CachePush(hTx.Hash())
 		txmp.metrics.NumberOfFailedCheckTxs.Add(1)
-		txmp.metrics.observeCheckTxPriorityDistribution(0, false, txInfo.SenderNodeID, true)
-		txmp.cache.Remove(txHash)
+		txmp.metrics.observeCheckTxPriorityDistribution(0, false, "", true)
```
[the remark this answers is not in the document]
pompon0: fixed

sei-tendermint/internal/evidence/pool.go
```diff
@@ -534,7 +534,6 @@ func (evpool *Pool) removeEvidenceFromList(
 		ev := e.Value()
 		if _, ok := blockEvidenceMap[evMapKey(ev)]; ok {
 			evpool.evidenceList.Remove(e)
-			e.DetachPrev()
```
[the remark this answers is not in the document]
pompon0: removed stale comment

sei-tendermint/internal/mempool/tx.go
```diff
@@ -99,320 +83,510 @@ func (wtx *WrappedTx) EVMNonce() uint64 {
...
+			wtx, ok := inner.byNonce[an]
+			if !ok || account.balance.Cmp(wtx.evm.OrPanic("non-evm tx").requiredBalance) < 0 {
+				break
+			}
+			account.nextNonce += 1
+			state.ready.Inc(wtx.Size())
+			if !wtx.readyEl.IsPresent() {
+				wtx.readyEl = utils.Some(s.readyTxs.PushBack(wtx.Tx()))
+			}
+		}
```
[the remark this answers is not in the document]
pompon0: wtf

sei-tendermint/internal/mempool/mempool.go
```diff
@@ -182,88 +188,45 @@ type TxMempool struct {
...
 		txsAvailable:         make(chan struct{}, 1),
+		txLocks:              newLockMap[types.TxHash](),
 		height:               -1,
-		cache:                NopTxCache{},
-		blockFailedTxs:       NopTxCache{},
 		metrics:              metrics,
-		txStore:              NewTxStore(),
-		gossipIndex:          clist.New[*WrappedTx](),
-		priorityIndex:        NewTxPriorityQueue(),
-		pendingTxs:           NewPendingTxs(cfg),
```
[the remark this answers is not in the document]
pompon0: it kind of make sense, but for a different reason - we want to start gossip dampening as soon as we are start evicting transactions

sei-tendermint/internal/mempool/mempool.go
```diff
@@ -495,83 +379,40 @@ func (txmp *TxMempool) CheckTx(ctx context.Context, tx types.Tx, txInfo TxInfo)
...
+	txmp.metrics.TotalTxsSizeBytes.Set(float64(txmp.TotalTxsBytesSize()))
 
-	txs := make([]types.Tx, 0, len(txHashes))
-	for _, txHash := range txHashes {
-		wtx := txmp.txStore.GetTxByHash(txHash)
-		txs = append(txs, wtx.Tx())
-	}
-	return txs
+	txmp.notifyTxsAvailable()
+	return res.ResponseCheckTx, nil
```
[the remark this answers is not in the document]
pompon0: wai

sei-will (approved):

sei-tendermint/internal/mempool/tx.go
```diff
@@ -99,320 +84,519 @@ func (wtx *WrappedTx) EVMNonce() uint64 {
...
+		maxGasEstimated = utils.Max[int64]()
+	}
+	totalGasWanted := int64(0)
+	totalGasEstimated := int64(0)
+	totalSize := int64(0)
+
+	var wtxs []*WrappedTx
+	for inner := range s.inner.Lock() {
+		if uint64(inner.state.Load().ready.count) >= s.config.TxNotifyThreshold { //nolint:gosec // count is non-negative
+			for _, wtx := range inner.inInclusionOrder() {
```
wen-coding, line 564: Is the reap cost still O(nlog(n))? Do we care?

sei-tendermint/internal/mempool/tx.go
```diff
@@ -99,320 +84,519 @@ func (wtx *WrappedTx) EVMNonce() uint64 {
...
+	// because it would cause them to be regossiped.
+	readyTxs *clist.CList[types.Tx]
+	// Sampler of priorites of all inserted READY txs.
+	// Used by TxMempool to damp re-gossiping of transactions.
+	priorityReservoir *reservoir.Sampler[int64]
 }
 
-func NewTxStore() *TxStore {
+func NewTxStore(cfg *Config, app *proxy.Proxy, metrics *Metrics) *txStore {
+	softLimit := txCounter{count: cfg.Size + cfg.PendingSize, bytes: utils.Clamp[uint64](cfg.MaxTxsBytes + cfg.MaxPendingTxsBytes)}
```
wen-coding, line 173: after we merge size+pendingsize, can a sender flood with high-nonce pending txs to evict ready txs?

sei-tendermint/internal/mempool/mempool.go
```diff
@@ -182,88 +188,45 @@ type TxMempool struct {
...
 	// however, a caller must explicitly grab a write-lock via Lock when updating
 	// the mempool via Update().
 	mtx                  sync.RWMutex
 	txConstraintsFetcher TxConstraintsFetcher
-
-	priorityReservoir *reservoir.Sampler[int64]
 }
 
+func (txmp *TxMempool) Size() int                 { return txmp.txStore.State().total.count }
+func (txmp *TxMempool) SizeBytes() uint64         { return txmp.txStore.State().total.bytes }
```
wen-coding: SizeBytes() include pending now? Would this break any outside tools? (metrics/dashboards)

sei-tendermint/internal/mempool/tx.go
```diff
@@ -99,320 +84,519 @@ func (wtx *WrappedTx) EVMNonce() uint64 {
...
+func (s *txStore) insert(inner *txStoreInner, wtx *WrappedTx) error {
+	if _, ok := inner.byHash[wtx.Hash()]; ok {
+		return errDuplicateTx
+	}
+	state := inner.state.Load()
+	if evm, ok := wtx.evm.Get(); ok {
+		// Fetch the evm account state.
+		account, ok := inner.accounts[evm.address]
+		if !ok {
+			// TODO(gprusak): consider whether we should move these queries out of the mutex.
```
wen-coding, line 302: Account state (balance, firstNonce) is only refreshed in compact(_, true) from Update. Can a tx get stuck pending if mempool's cached nextNonce drifts from on-chain reality?

sei-tendermint/internal/mempool/tx.go
```diff
@@ -99,320 +84,519 @@ func (wtx *WrappedTx) EVMNonce() uint64 {
...
 		}
-
-		if p.config.TTLDuration > 0 {
-			idxFirstNotExpiredTx := len(inner.txs)
-			for i, ptx := range inner.txs {
-				// once found, we can break because these are ordered
-				if now.Sub(ptx.timestamp) <= p.config.TTLDuration {
-					idxFirstNotExpiredTx = i
-					break
+		if remove {
```
wen-coding, line 584: This is only true for Autobahn right? I think you said we will have a totally different(?) implementation for Autobahn?

sei-tendermint/internal/consensus/replay_test.go
```diff
@@ -655,11 +655,12 @@ func testHandshakeReplay(
 	store.commits = commits
 
 	state := genesisState.Copy()
+	replayMempool := newReplayTxMempool(kvstore.NewProxy())
 	// run the chain through state.ApplyBlock to build up the tendermint state
 	state = buildTMStateFromChain(
 		ctx,
 		t,
-		sim.Mempool,
+		replayMempool,
```
wen-coding, line 659: I like this, should we just remove Mempool field from sim now?

sei-tendermint/internal/mempool/cache.go
```diff
@@ -56,26 +32,29 @@ type cacheKey = string
...
 	c.cacheMap = make(map[cacheKey]*list.Element, c.size)
 	c.list.Init()
 }
 
-func (c *LRUTxCache) Push(txHash types.TxHash) bool {
-	c.mtx.Lock()
-	defer c.mtx.Unlock()
+func (c *lruTxCache) Push(txHash types.TxHash) bool {
+	if c.size <= 0 {
+		return true
```
wen-coding, line 56: Is it reasonable to return true here? We didn't actually push anything?

sei-tendermint/internal/mempool/tx.go
pompon0: it is, it is done once per block proposal. I don't think we do.

sei-tendermint/internal/mempool/tx.go
pompon0: no, inInclusionOrder() contains ready txs first (ready always take precedence during compaction).

sei-tendermint/internal/mempool/mempool.go
pompon0: good point, I didn't notice that Size and SizeBytes took different sets of txs. Reverting.

sei-tendermint/internal/mempool/tx.go
pompon0: tendermint impl tightly binds mempool state and application state - given that Commit and Update are executed under mempool lock there is no race condition here. So the cached nonce can drift from reality iff node is behind

sei-tendermint/internal/mempool/tx.go
pompon0: Yes, I did. I'll get rid of this mode as soon as autobahn gets its own mempool. This "remove" option is consistent with what we did so far, but it is hardly robust.

sei-tendermint/internal/mempool/cache.go
pompon0: Here I'm replacing NopCache with LRU cache with capacity 0, so I had to extend the LRU cache semantics. Afaict "true" here is the correct value: it is equivalent to what NopCache did, and result of Push is interpreted in mempool as "whether the value wasn't already present in cache", and NOT "whether the value wasn't already present but now is". I.e. what Push does with capacity 0 is: it pushes element successfully, then immediately evicts it (although implementation do evicting first, then pushing, so that's why we need a special case here).

sei-tendermint/internal/consensus/replay_test.go
pompon0: good point, done

sei-tendermint/internal/mempool/mempool.go
```diff
@@ -174,6 +179,7 @@ type TxMempool struct {
 	metrics *Metrics
 	config  *Config
 	app     *proxy.Proxy
+	txLocks *lockMap[types.TxHash]
```
wen-coding, line 182: hmm okay, this map has no capacity limit. I guess it's fine because CheckTx should normally be fast.

sei-tendermint/internal/mempool/mempool.go
```diff
@@ -273,119 +236,36 @@ func NewTxMempool(
...
-	return float64(txmp.NumTxsNotPending()) / float64(txmp.config.Size)
-}
-
-func (txmp *TxMempool) NumTxsNotPending() int {
-	return txmp.txStore.Size()
-}
-
-func (txmp *TxMempool) BytesNotPending() int64 {
-	return txmp.txStore.AllTxsBytes()
+	return float64(txmp.Size()) / float64(txmp.config.Size+txmp.config.PendingSize)
```
wen-coding: Is it guaranteed that PendingSize >= 0 or Size + PendingSize > 0? Should we add some checks?

sei-tendermint/internal/mempool/tx.go
```diff
@@ -1,92 +1,77 @@
...
+	Address common.Address
+	Nonce   uint64
 }
 
 type hashedTx struct {
-	tx   types.Tx
-	hash types.TxHash
+	tx        types.Tx
+	hash      types.TxHash
+	protoSize int64
```
wen-coding, line 35: I assume this is added because then we don't need to recompute protoSize several times. Maybe worth adding a comment here in case future AI agents try to remove it.

sei-tendermint/internal/mempool/tx.go
```diff
@@ -99,320 +84,519 @@ func (wtx *WrappedTx) EVMNonce() uint64 {
...
+	hardLimit txCounter
+	state     utils.AtomicSend[txStoreState]
+
+	// Cache of already seen txs, reducess pressure on app.
+	// It is a superset of transactions in txStore.
+	// * successfully inserted transactions are automatically added to cache.
+	// * txs which fail Insert() are NOT added to cache and can be reattempted later.
+	// * invalid transactions can be recorded via CachePush.
+	// * txs dropped due to pruning are removed from cache.
+	// * txs successfully executed are kept in cache to avoid reinsert.
```
wen-coding: nit: explain for how long?

sei-tendermint/internal/mempool/tx.go
```diff
@@ -99,320 +84,519 @@ func (wtx *WrappedTx) EVMNonce() uint64 {
...
+	// * txs failed execution are eligible to be reexecuted once (iff config.KeepInvalidTxsInCache).
```
wen-coding: I think they will be reexecuted once if config.KeepInvalidTxsInCache is false?
```
if !s.config.KeepInvalidTxsInCache {
   // retry logic
}
```

sei-tendermint/internal/mempool/tx.go
```diff
@@ -99,320 +84,519 @@ func (wtx *WrappedTx) EVMNonce() uint64 {
...
-			result := evaluate(inner.txs[i])
-			switch result {
-			case abci.Accepted:
-				acceptedTxs = append(acceptedTxs, inner.txs[i])
-				poppedIndices = append(poppedIndices, i)
-			case abci.Rejected:
-				rejectedTxs = append(rejectedTxs, inner.txs[i])
-				poppedIndices = append(poppedIndices, i)
+	// Cap priority to obtain a linear order of txs per account by nonce.
+	// NOTE: this precisely emulates the heap behavior described in this functions docstring.
```
wen-coding: nit: I almost misread it as reap behavior, maybe just say what behavior we are guaranteeing here.

sei-tendermint/internal/mempool/tx.go
```diff
@@ -99,320 +84,519 @@ func (wtx *WrappedTx) EVMNonce() uint64 {
...
+	accPrio := make(map[common.Address]int64, len(inner.accounts))
+	for _, txs := range utils.Slice(ready, pending) {
+		// Sort by nonce.
+		slices.SortFunc(txs, func(a, b *WrappedTx) int { return cmp.Compare(a.EVMNonce(), b.EVMNonce()) })
```
wen-coding, line 394: nit: maybe worth adding some comment why we sort nonce across different accounts here. Not obvious unless you read the following code.

wen-coding (approved):

sei-tendermint/internal/mempool/mempool.go
pompon0: it is rate limited on the rpc level.

sei-tendermint/internal/mempool/tx.go
```diff
@@ -99,320 +84,519 @@ func (wtx *WrappedTx) EVMNonce() uint64 {
...
-			inner.txs = inner.txs[idxFirstNotExpiredTx:]
-		}
-
-		if len(inner.txs) == 0 {
-			return
 		}
-
-		if p.config.TTLDuration > 0 {
-			idxFirstNotExpiredTx := len(inner.txs)
-			for i, ptx := range inner.txs {
```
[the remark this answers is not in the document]
pompon0: WAI

sei-tendermint/internal/mempool/tx.go
```diff
@@ -99,320 +84,519 @@ func (wtx *WrappedTx) EVMNonce() uint64 {
...
+			wtx, ok := inner.byNonce[an]
+			if !ok || account.balance.Cmp(wtx.evm.OrPanic("non-evm tx").requiredBalance) < 0 {
+				break
+			}
+			account.nextNonce += 1
+			state.ready.Inc(wtx.Size())
+			if !wtx.readyEl.IsPresent() {
+				wtx.readyEl = utils.Some(s.readyTxs.PushBack(wtx.Tx()))
+			}
+		}
```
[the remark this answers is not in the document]
pompon0: that's incorrect

sei-tendermint/internal/mempool/mempool.go
pompon0: added max(1,...).

sei-tendermint/internal/mempool/tx.go
pompon0: done

sei-tendermint/internal/mempool/tx.go
pompon0: I mean, it is a LRU cache, so if someone tries to aggressively reinsert it, then it will stay there forever.

sei-tendermint/internal/mempool/tx.go
pompon0: true

sei-tendermint/internal/mempool/tx.go
pompon0: I don't think duplicating the docstring makes sense. I rephrased this statement. Perhaps it is better now.

sei-tendermint/internal/mempool/tx.go
pompon0: done

sei-tendermint/internal/mempool/mempool.go
```diff
@@ -182,88 +188,45 @@ type TxMempool struct {
...
 	// however, a caller must explicitly grab a write-lock via Lock when updating
 	// the mempool via Update().
 	mtx                  sync.RWMutex
 	txConstraintsFetcher TxConstraintsFetcher
-
-	priorityReservoir *reservoir.Sampler[int64]
 }
 
+func (txmp *TxMempool) Size() int                 { return txmp.txStore.State().total.count }
+func (txmp *TxMempool) SizeBytes() uint64         { return txmp.txStore.State().ready.bytes }
```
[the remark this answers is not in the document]
pompon0: that's backward compatible behavior

PR #3479, opened 2026-05-20
metrics(app): add app_block_gas_wanted and app_block_gas_wanted_ratio histograms (PLT-354)

amir-deris:
## Summary

Adds two OTel histogram instruments to make per-block gas utilisation observable — answering "how close are blocks getting to the `MaxGasWanted` cap in practice" without offline block parsing.

- `app_block_gas_wanted` (`Int64Histogram`, unit `{gas}`) — absolute `totalGasWanted` summed across all transactions in an accepted proposal
- `app_block_gas_wanted_ratio` (`Float64Histogram`, unit `1`) — `totalGasWanted / MaxGasWanted`, stays meaningful if the cap is raised via governance

Both are emitted in `checkTotalBlockGas` (`app/app.go`) on the **success path only** — the function already short-circuits with `return false` when the sum exceeds the cap, so only complete, accepted totals are recorded. The existing `app_failed_total_gas_wanted_check` counter covers rejection events.

Bucket boundaries:
- Gas: 1 M–10 M in 1 M steps, then 5 M steps to 50 M (current cap)
- Ratio: 0.1 steps from 0.1 → 0.9, plus 0.95 and 1.0

Closes PLT-354.

## Test plan

- [x] Deploy to a test node and verify `app_block_gas_wanted_bucket` and `app_block_gas_wanted_ratio_bucket` appear in the Prometheus scrape. Can use this p50 query in Grafana by running 
`histogram_quantile(0.5, rate({__name__="sei-chain_app_block_gas_wanted_ratio_bucket"}[10m]))`
- [x] Confirm values are non-zero and ratio stays ≤ 1.0 under normal load

### DIFF
--- a/app/metrics.go
+++ b/app/metrics.go
@@ -65,7 +65,7 @@ var (
 		gigaFallback         metric.Int64Counter
 
 		// Per-block gas utilisation
-		blockGasWanted      metric.Float64Histogram
+		blockGasWanted      metric.Int64Histogram
 		blockGasWantedRatio metric.Float64Histogram
 
 		// Light invariance check
@@ -167,7 +167,7 @@ var (
 			metric.WithUnit("{count}"),
 		)),
 
-		blockGasWanted: must(meter.Float64Histogram(
+		blockGasWanted: must(meter.Int64Histogram(
 			"app_block_gas_wanted",
 			metric.WithDescription("Per-block total gas wanted across all transactions"),
 			metric.WithUnit("{gas}"),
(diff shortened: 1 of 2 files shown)


app/metrics.go
```diff
@@ -149,6 +167,20 @@ var (
 			metric.WithUnit("{count}"),
 		)),
 
+		blockGasWanted: must(meter.Float64Histogram(
```
bdchatham: nit: does this need to be a float?

bdchatham (commented):
Seems like the PR description is slightly stale

app/metrics.go
amir-deris: Good point. I will convert to `Int64Histogram` for better semantics.

bdchatham (approved):

masih (approved):

PR #3484, opened 2026-05-21
Refactor mempool helpers and tests

masih:
Refactor mempool helpers for better code readability and refactor tests to remove redundant code assertions, use higher-level primitives for eventual checks.

While here, fold the scattered metric writes in insertTx into a single block at the end of the success path.


pompon0 (approved):

masih:
Closing in favour of rewrite

PR #3487, opened 2026-05-21
add new storage test panel

cody-littley:

In order to speed up test runtime, split storage unit tests into a stand alone panel.

Screenshot 2026-05-21 at 12 49 40 PM

I downloaded the logs from the new panels and from the old panels, and used an LLM to verify that the expected tests are running in each panel.


masih (approved):
I understand this the pre-step in more tests to come since today it won't make the slowest mandatory CI job any faster. Which means total wait time is unchanged.

No blockers regardless and glad to see experimentation here 🙌

.github/workflows/state_db-tests.yml
```diff
@@ -0,0 +1,194 @@
+name: DB
```
blindchaser: how about name it `State DB Tests`

blindchaser (approved):

PR #3491, opened 2026-05-21
LittDB: secondary keys

cody-littley:

LittDB currently only provides a key->value mapping. This PR widens that concept, allowing a particular value to have multiple keys. This is needed for doing things like storing each block so you can look it up either by block height or block hash. It will also potentially be useful if we decide we want to index transactions.

Specifying secondary keys does not increase disk footprint by much. i.e. the DB isn't storing the same data multiple times. It instead allows multiple keys to reference the same on-disk data.

unit tests


sei-db/db_engine/litt/disktable/disk_table.go
```diff
@@ -724,43 +724,86 @@ func (d *DiskTable) CacheAwareGet(
...
+			return fmt.Errorf("key is too large, length must not exceed 2^16 bytes: %d bytes", len(kv.Key))
+		}
+		if len(kv.Value) > math.MaxUint32 {
+			return fmt.Errorf("value is too large, length must not exceed 2^32 bytes: %d bytes", len(kv.Value))
+		}
 
+		// Validate every secondary key in this request, and detect duplicate keys (primary vs
+		// secondary, secondary vs secondary) within the request. Cross-request collisions remain
+		// the caller's responsibility, matching existing semantics for primary keys.
+		seen := make(map[string]struct{}, 1+len(kv.SecondaryKeys))
```
blindchaser, line 758: the implementation is the entire batch is rejected if any request fails validation while doc contradicts

sei-db/db_engine/litt/memtable/mem_table.go
```diff
@@ -92,29 +92,64 @@ func (m *memTable) KeyCount() uint64 {
...
+	}
 
 	return nil
 }
 
-func (m *memTable) PutBatch(batch []*types.KVPair) error {
-	for _, kv := range batch {
-		err := m.Put(kv.Key, kv.Value)
+func (m *memTable) PutBatch(batch []*types.PutRequest) error {
+	for _, req := range batch {
```
blindchaser, line 154: DiskTable.PutBatch is atomic while memtable.PutBatch is not. should we update memtable.PutBatch ?

sei-db/db_engine/litt/README.md
```diff
@@ -119,8 +119,8 @@ Source: [table.go](table.go)
 ```go
 type Table interface {
 Name() string
-Put(key []byte, value []byte) error
-PutBatch(batch []*types.KVPair) error
+Put(key []byte, value []byte, secondaryKeys ...*types.SecondaryKey) error
```
blindchaser: Nit: some go code in README.md has no identation

sei-db/db_engine/litt/types/key_kind.go
```diff
@@ -0,0 +1,42 @@
...
+func (k KeyKind) IsSecondary() bool {
+	return k == KeyKindSecondary || k == KeyKindFinalSecondary
+}
+
+// TerminatesGroup reports whether this kind closes its group. Recovery commits a buffered group
+// when it encounters a record whose kind terminates the group (KeyKindStandalone or
+// KeyKindFinalSecondary).
+func (k KeyKind) TerminatesGroup() bool {
+	return k == KeyKindStandalone || k == KeyKindFinalSecondary
+}
```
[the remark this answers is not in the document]
cody-littley: removed these methods

sei-db/db_engine/litt/README.md
cody-littley: fixed indentation

sei-db/db_engine/litt/disktable/disk_table.go
cody-littley: Fixed the documentation.

sei-db/db_engine/litt/memtable/mem_table.go
cody-littley: Note that `memtable` is a test-only implementation that I intend to delete at some point in time, since it just makes everything more complex. 😅 

But until we delete it, it makes sense to maintain similar behavior for the sake of consistency. I've updated memtable so it has similar semantics to disktable.

blindchaser (approved):

Kbhat1 (approved):
Two questions, non-blocking

(1) Where do we check that a secondary key isn't already used? doc says secondaries must be globally unique. Is that enforced just in the current write, or against all prior segments too?
(2) Should we add a test to make sure that when a segment expires the secondaries go w it? Looking again to see if this is covered

cody-littley:
> Where do we check that a secondary key isn't already used? doc says secondaries must be globally unique. Is that enforced just in the current write, or against all prior segments too?

This is handled inside the keymap. There's an optional setting doubleWriteProtection that checks all keys, including secondary ones. It's turned off by default though, since this is a very expensive check at high throughput (requires a random read for every write).

```
func (p *PebbleDBKeymap) Put(keys []*types.ScopedKey) error {

	if p.doubleWriteProtection {
		for _, k := range keys {
			_, ok, err := p.Get(k.Key)
			if err != nil {
				return fmt.Errorf("failed to get key: %w", err)
			}
			if ok {
				return fmt.Errorf("key %s already exists", k.Key)
			}
		}
	}
```

> Should we add a test to make sure that when a segment expires the secondaries go w it? Looking again to see if this is covered

```
// TestSecondaryKeyTTLGroupExpiration verifies that a primary and all of its secondaries expire
// together: once the TTL window passes for the primary, every secondary becomes unreachable on
// the same GC pass. The buildMemKeyDiskTableSingleShard test config uses TargetSegmentFileSize=100,
// so writing a few hundred bytes of cycler data is enough to rotate past the segment that holds
// our group; once the old segment is sealed and its lastValueTimestamp is older than the TTL, GC
// reaps it.
func TestSecondaryKeyTTLGroupExpiration(t *testing.T) {
```

PR #3494, opened 2026-05-21
fix(sei-cosmos): harden paginated RPC queries against DoS via limit, offset, and count_total caps (PLT-361)

amir-deris:
## Problem

Three separate vectors allow a single RPC call to trigger unbounded KV store iteration:

1. **Limit too large** — `MaxLimit` was `math.MaxUint64`; callers could request billions of items in one call.
2. **`count_total=true` unbounded scan** — after serving the requested page, the paginator continued iterating the entire remaining store just to populate `pagination.total`. Implicit `limit=0` also silently enabled this behaviour.
3. **Offset too large** — no cap on `pagination.offset`; a caller with `offset=1_000_000_000` forces the iterator to skip a billion entries before serving a single result.
4. **`GetBlockWithTxs` allocation** — user-supplied `limit` was passed directly into `make([]*txtypes.Tx, 0, limit)` before any validation.

### `sei-cosmos/types/query/pagination.go`
- Lowers `MaxLimit` to `1_000`
- Adds `MaxOffset = 10_000` and `VerifyPaginationOffset()`; enforced in `ParsePagination` and `paginate()`
- Adds `MaxScanLimit = 10_000` — fires when `count_total=true` and the iterator travels more than `MaxScanLimit` entries *past the end of the requested page* (`count > end + MaxScanLimit`), preventing full-store counts while still allowing `count_total` on reasonably-sized stores
- Removes the implicit `countTotal = true` side-effect when `limit == 0`; callers must opt in explicitly

### `sei-cosmos/types/query/filtered_pagination.go`
- Same `MaxOffset` and `MaxScanLimit` guards applied to `FilteredPaginate` and `GenericFilteredPaginate`
- Fixes a bug in the original scan cap where `totalIter` (raw store iterations) was compared against `end` (a filtered-hit count), causing the limit to fire mid-page for selective filters — a query with a 1% pass rate and `limit=100` could be rejected before accumulating its first result
- Replaces the single mixed-space guard with a two-phase approach:
  - **Phase 1** (`numHits < end`): caps raw iterations at `end + MaxScanLimit` — prevents full-store walks when the filter produces too few hits to fill the page (no-hits DoS path); cannot fire once the page starts completing
  - **Phase 2** (`numHits >= end`): tracks iterations after page completion via `pageCompleteIter` and caps at `MaxScanLimit` — limits post-page `count_total` scanning without any risk of mid-page interference

### `sei-cosmos/x/auth/tx/service.go`
- Calls `pagination.VerifyPaginationLimit(limit)` before `make([]*txtypes.Tx, 0, limit)` in `GetBlockWithTxs`

## Behaviour summary

| Request | Before | After |
|---|---|---|
| `limit > 1,000` | accepted, full scan | `InvalidArgument` |
| `limit = 0` | default 100 items + implicit `count_total=true` | default 100 items, no implicit count |
| `count_total=true`, store has > `end + 10,000` entries | full store scan | `InvalidArgument` |
| `count_total=true`, selective filter, `numHits < end` | rejected mid-page assembly | completes page; errors only if filter is too sparse to fill page within `end + MaxScanLimit` raw iterations |
| `offset > 10,000` | accepted, skips N entries | `InvalidArgument` |
| `GetBlockWithTxs` with `limit=1_000_000_000` | allocates ~1 GB slice | `InvalidArgument` |

## Regression risk

Breaking change for clients sending `limit > 1,000`, `offset > 10,000`, or relying on `limit=0` to return a total count. Clients should use `limit ≤ 1,000`, follow `next_key` for subsequent pages, and set `count_total=true` explicitly only when the store is known to be small.

`ExportGenesis` and the `TotalSupply` invariant were migrated to `collectAllTotalSupply`, which internally paginates at `MaxLimit` per page to collect all denominations without being blocked by the new limit.


sei-cosmos/types/query/pagination.go
```diff
@@ -15,8 +14,7 @@ import (
 const DefaultLimit = 100
 
 // MaxLimit is the maximum limit the paginate function can handle
-// which equals the maximum value that can be stored in uint64
-const MaxLimit = math.MaxUint64
+const MaxLimit = uint64(10_000)
```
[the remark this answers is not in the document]
amir-deris: This comment is related to previous version of code and no longer valid.

sei-cosmos/x/bank/keeper/keeper.go
```diff
@@ -106,6 +106,28 @@ func (k BaseKeeper) GetPaginatedTotalSupply(ctx sdk.Context, pagination *query.P
...
+
+		if pageRes == nil || len(pageRes.NextKey) == 0 {
+			return totalSupply, nil
+		}
+		pageReq = &query.PageRequest{
+			Key:   pageRes.NextKey,
+			Limit: query.MaxLimit,
+		}
+	}
+}
```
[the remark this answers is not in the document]
amir-deris: The `IterateTotalSupply` panics for invalid values, and using it in `sei-cosmos/x/bank/keeper/invariants.go` would cause panic which would be a different behaviour compared to today. Alternatively, we could make `IterateTotalSupply` to not panic and return error, but that would require a bigger refactor.

sei-cosmos/types/query/filtered_pagination.go
```diff
@@ -83,11 +90,18 @@ func FilteredPaginate(
...
+		nextKey   []byte
+		totalIter uint64
 	)
 
 	for ; iterator.Valid(); iterator.Next() {
+		totalIter++
+		if countTotal && totalIter > end+MaxScanLimit {
+			return nil, status.Errorf(codes.InvalidArgument,
+				"count_total scan exceeds maximum of %d items past the page; use key-based pagination instead", MaxScanLimit)
+		}
```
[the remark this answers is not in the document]
amir-deris: Will work on this to fix it!

sei-cosmos/types/query/filtered_pagination.go
```diff
@@ -83,11 +97,26 @@ func FilteredPaginate(
...
+		nextKey          []byte
+		totalIter        uint64
+		pageCompleteIter uint64
 	)
 
 	for ; iterator.Valid(); iterator.Next() {
+		totalIter++
+		// Phase 1: page not yet complete — cap raw iterations to prevent full-store
+		// walks when the filter produces too few hits to fill the page.
+		if numHits < end && totalIter > offset+MaxScanLimit {
```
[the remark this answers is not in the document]
amir-deris: That is intentional. After the offset, we allow `MaxScanLimit` to fill a page. If it is not filled yet, the caller needs to use key based pagination.

sei-cosmos/types/query/filtered_pagination_test.go
```diff
@@ -170,6 +170,68 @@ func (s *paginationTestSuite) TestReverseFilteredPaginations() {
...
+	numItems := int(query.MaxScanLimit) + 2
+	for i := 0; i < numItems; i++ {
+		kvStore.Set([]byte(fmt.Sprintf("%08d", i)), []byte("v"))
+	}
+
+	_, err := query.FilteredPaginate(kvStore, &query.PageRequest{Limit: 1, CountTotal: true}, func(_ []byte, _ []byte, _ bool) (bool, error) {
+		return true, nil
+	})
+	s.Require().Error(err)
+	s.Require().Contains(err.Error(), "scanned more than")
```
bdchatham, line 213: Could we add a selective-filter case here? The scan-limit tests are all all-pass / all-fail, but the bug this PR actually fixes is the mid-page one — a ~1-in-N filter that needs a few thousand raw scans to fill its page should still succeed, not get rejected. That's the regression I'd want pinned so we don't quietly bring back the old `totalIter`-vs-`end` comparison.

sei-cosmos/x/bank/keeper/keeper_test.go
```diff
@@ -157,6 +157,12 @@ func (suite *IntegrationTestSuite) TestSendCoinsAndWei() {
 	require.Equal(sdk.NewInt(53), keeper.GetBalance(ctx, addr3, sdk.DefaultBondDenom).Amount)
 }
 
+func (suite *IntegrationTestSuite) TestGetPaginatedTotalSupplyMaxLimitExceeded() {
+	_, _, err := suite.app.BankKeeper.GetPaginatedTotalSupply(suite.ctx, &query.PageRequest{Limit: query.MaxLimit + 1})
+	suite.Require().Error(err)
```
bdchatham, line 162: Can we add a multi-page case here for 1,000 denoms asserting `collectAllTotalSupply` returns the full set? Everything's single-page today, so the cross-page merge is untested, and that's the path feeding the in-consensus `TotalSupply` invariant. Given the fork stakes I'd want it covered.

sei-cosmos/types/query/pagination.go
```diff
@@ -14,9 +13,15 @@ import (
...
-const MaxLimit = math.MaxUint64
+// MaxLimit is the maximum limit per page the paginate function can handle
+const MaxLimit = uint64(1_000)
+
+// MaxScanLimit is the maximum number of store entries the paginate function
+// will iterate past the page end when count_total is requested.
+const MaxScanLimit = uint64(10_000)
+
+// MaxOffset is the maximum offset allowed in a PageRequest.
+const MaxOffset = uint64(10_000)
```
bdchatham, line 24: Shouldn't need to be exported

sei-cosmos/types/query/filtered_pagination.go
```diff
@@ -15,6 +17,15 @@ import (
...
 // to the client.
+//
+// Scan limits: to prevent unbounded store walks, this function caps iteration at MaxScanLimit
+// entries both before the page is filled (Phase 1) and after (Phase 2). If the cap is hit:
+//   - Phase 1 (page not yet full): returns an error — the page could not be assembled.
+//   - Phase 2, CountTotal=true: returns an error — the total count could not be determined.
+//   - Phase 2, CountTotal=false: returns the assembled page with NextKey=nil. This does NOT
+//     guarantee there are no further results; it means no next filtered hit was found within
+//     MaxScanLimit entries past the page boundary. Callers that require reliable full traversal
+//     of sparse datasets should use key-based pagination instead.
```
bdchatham: Slight concern with the growing size of these comments. A bit of a smell to me suggesting we could make the code more expressive so that it effectively explains itself. It does already include comments in the implementation

masih:
THere is some overlap between this PR and #3524 that's worth distilling before review

sei-cosmos/types/query/pagination.go
amir-deris: Actually the problem is they are also used in tests which are in `query_test` package, while this const is defined in `query` package and therefore needs to be exported to be available in there.

bdchatham (approved):

PR #3495, opened 2026-05-22
Pipeline giga static block processing

codchen:
## Summary
- split giga static transaction preparation from stateful execution
- add an async per-block static processing pipeline so later block stateless work can overlap earlier block stateful execution
- wire giga synchronous and OCC execution paths to consume prepared EVM metadata, keeping Cosmos transactions as legacy fallbacks

## Tests
- `go test ./app`
- `make giga-integration-test`
- `make giga-mixed-integration-test`
- `make autobahn-integration-test`
- `make parquet-integration-test`


app/app.go
```diff
@@ -1515,7 +1549,202 @@ func (app *App) ProcessTxsSynchronousV2(ctx sdk.Context, txs [][]byte, typedTxs
...
+
+	typedTxs := preDecoded
+	if len(typedTxs) != len(txs) {
+		typedTxs = app.DecodeTxBytesConcurrently(txs)
+	}
+	job.result = &gigaStaticBlockResult{
+		typedTxs:  typedTxs,
+		staticTxs: app.prepareGigaStaticTxsConcurrently(ctx, typedTxs, chainID),
+	}
+}
```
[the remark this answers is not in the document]
codchen: Fixed in 391db048 by adding a top-level recover around runGigaStaticBlockJob. If the async static block processor panics, it now logs the panic, clears the static result so ProcessBlock falls back to normal decoding/static prep, and still closes job.done so waiters do not block.

app/app.go
```diff
@@ -1524,7 +1558,213 @@ func (app *App) ProcessTxsSynchronousV2(ctx sdk.Context, txs [][]byte, typedTxs
...
+		_, _ = fmt.Fprintf(h, "%d:", len(tx))
+		_, _ = h.Write(tx)
+	}
+	return gigaStaticBlockKey{height: height, id: hex.EncodeToString(h.Sum(nil))}
+}
+
+func (app *App) pruneGigaStaticPipelineLocked(height int64) {
+	for key := range app.gigaStaticPipeline {
+		if key.height+gigaStaticPipelineRetentionBlocks < height {
+			delete(app.gigaStaticPipeline, key)
```
masih, line 1676: This can still grow unbounded for the same height. Since the churn here is driven by proposers, we should cap this to some maximum. Otherwise OOM is a concern.

app/app.go
```diff
@@ -1524,7 +1558,213 @@ func (app *App) ProcessTxsSynchronousV2(ctx sdk.Context, txs [][]byte, typedTxs
...
+
+	ethTx, _ := evmMsg.AsTransaction()
+	if ethTx == nil {
+		return &gigaStaticTxResult{err: fmt.Errorf("failed to convert to eth transaction")}
+	}
+
+	sender, seiAddr, _, recoverErr := evmante.RecoverSenderFromEthTx(ctx, ethTx, chainID)
+	if recoverErr != nil {
+		return &gigaStaticTxResult{errResult: &abci.ExecTxResult{
+			Code: 1,
```
philipsu522: nit - would it be better to use some const var instead of an int

philipsu522 (approved):

app/app.go
codchen: Fixed in 64a147d by replacing the raw signature-recovery failure code with gigaStaticSignatureRecoveryFailureCode.

app/app.go
codchen: Fixed in 64a147d by capping the static pipeline both per height and globally. When the cap is reached, we skip the opportunistic background job and let ProcessBlock fall back to normal decoding/static prep, so proposer churn can no longer grow this map unbounded.

PR #3499, opened 2026-05-22
Refine address mapping for delegations and rebinds

masih:
Simplify the address-mapping and code-storage helpers in x/evm/keeper and the parallel giga/deps/xevm/keeper package. Comments document the non-obvious invariants (the three-way GetCodeHash semantics, the bidirectional index contract, the cast-address receive rule), magic constants are named, and a few stale branches in SetCode and SetAddressMapping are simplified along the way.

Expand the keeper test suites on both sides to cover behaviours that were previously implicit: SetCode with delegation-shaped input, SetAddressMapping when either side is rebound to a new partner, and GetCodeHash for accounts with a nonce but no balance. The two keeper packages share underlying storage via GigaEvmKeeper.UseRegularStore=true, so coverage is mirrored to keep them from drifting.

Mirror both fixes into giga/deps/xevm/keeper.


masih:
@codex review this

PR #3504, opened 2026-05-25
ci(PLT-368): build Docker images and seid once, reuse across integration test matrix

amir-deris:
## Problem

Each of the ~25 integration test matrix jobs independently rebuilt the localnode and rpcnode Docker images and recompiled the seid binary from source. This caused ~25× redundant builds per CI run, inflating total compute by 50–70% and adding several minutes of startup overhead to every job.

## Solution

Introduce a `prepare-cluster` job that runs once before the matrix, builds all artifacts, and uploads them. Matrix jobs download and reuse the prebuilt artifacts instead of rebuilding.

### New workflow structure
- **`prepare-cluster`**: builds `sei-chain/localnode` and `sei-chain/rpcnode` Docker images and compiles `seid` inside the localnode container; packages as `integration-build.tar.gz` + `integration-docker-images.tar.zst`; uploads as a 1-day artifact
- **`integration-tests` matrix**: downloads artifacts, loads images via `docker load`, starts cluster with `SKIP_BUILD=true` — no image builds or Go compilation per job

### New Makefile targets
- `build-seid-in-localnode` — compiles `seid` inside the localnode container with correct Linux target and Go PATH
- `build-integration-ci-artifacts` — orchestrates image + binary build for CI
- `docker-cluster-start-ci` — starts the 4-node cluster from prebuilt artifacts
- `run-rpc-node-integration-ci` — starts the RPC node from prebuilt image
- `ensure-integration-ci-images` — guard that fails fast if images weren't loaded

### Other changes
- Simplified `integration-test-check` job: replaced fragile GitHub API polling with direct `needs` context checks
- Used `zstd -T0 --fast` instead of `gzip -1` for Docker image archiving — parallel compression, faster upload/download across all 25 jobs
- Fixed `build-seid-in-localnode`: added `PATH=/usr/local/go/bin:$PATH` (Dockerfile PATH omits the Go install dir) and `make clean` before build (prevents Make from skipping the `build` target when a `build/` directory already exists)
- Fixed `run-rpc-node-integration-ci`: poll until the cluster reaches block 100 (first snapshot interval) before starting the RPC node — without this, `SKIP_BUILD=true` causes the RPC node to configure state sync before any snapshot exists, crashing `seid` silently and hanging the Chain Operation Test indefinitely

## Results

Measured against the latest successful main branch run vs this PR (both on `ubuntu-large`, 4-core · $0.016/min):

main:    [===== 25 matrix jobs in parallel =====] 
15 min wall clock
190 runner-min total

This  PR:      
[== prepare-cluster ==] 6 min (sequential)
[== 25 matrix jobs ==] 11 min (parallel) 
18 min wall clock, 150 runner-min total
                         

| Metric | main | PR | Δ |
|---|---|---|---|
| Total wall clock | ~15 min | ~18 min | +3 min |
| Matrix phase only | ~15 min | ~11 min | -4 min |
| Per-job image build overhead | ~3–5 min × 25 jobs | 0 | eliminated |
| **Compute saved per run** | — | — | **~40 runner-min** |
| **Cost saved per run** | — | — | **~$0.65** |
| **Cost saved per month** (10 runs/day) | — | — | **~$194** |
| **Cost saved per year** | — | — | **~$2,358** |

The +3 min wall clock is entirely the sequential `prepare-cluster` step; the matrix phase itself is 4 min faster. The real saving is compute: main burned ~3–5 min of image builds in every parallel job — eliminated here. At 25 jobs per run this is ~40 compute-minutes saved, or ~$0.65/run at the `ubuntu-large` rate of $0.016/min.

Each job previously triggered base image pulls as part of docker build. With prepare-cluster doing one build and matrix jobs just doing docker load from the artifact, we're hitting Docker Hub once per run instead of up to 25 times. This matters because Docker Hub has rate limits on unauthenticated/free pulls — hitting them mid-CI could be a flaky failure mode.


masih (approved):
Thanks for documenting the wall clock time increase.

Given most of intermittent failures in integration tests come from docker build stages (e.g. ubuntu mirror fetch failures) this PR is an improvement despite the increase in total run time 🙌

bdchatham (approved):

PR #3506, opened 2026-05-26
Hash Vault Primitive

cody-littley:

Introduces a new storage micro-utility: the `HashVault`. Goal of this utility is to prevent any chain of events from causing a validator to change its mind about the app hash. No matter the circumstances, once a validator commits to an app hash, it should require direct human intervention to change that app hash.

This PR does not integrate the HashVault. In a follow up PR, the HashVault will be added to the `composite.Store`, which will not return app hashes until ensuring they are durable.

unit tests

### DIFF
--- a/sei-db/state_db/sc/hashvault/pebble_hashvault_rollback.go
+++ b/sei-db/state_db/sc/hashvault/pebble_hashvault_rollback.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"math"
 	"os"
 
 	"github.com/cockroachdb/pebble/v2"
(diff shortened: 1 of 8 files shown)


sei-db/state_db/sc/hashvault/pebble_hashvault_rollback.go
```diff
@@ -0,0 +1,104 @@
...
+
+	return hardRollbackAbove(db, config.DataDir, blockHeight)
+}
+
+// hardRollbackAbove deletes hashes strictly above blockHeight and clears the prune boundary in one
+// atomic batch.
+func hardRollbackAbove(db *pebble.DB, dataDir string, blockHeight uint64) error {
+	batch := db.NewBatch()
+	defer func() { _ = batch.Close() }()
+	if err := batch.DeleteRange(hashKey(blockHeight+1), hashKeyUpperBound(), nil); err != nil {
```
blindchaser, line 58: when blockHeight == math.MaxUint64, blockHeight+1 overflows to 0

Fix: special-case max height before DeleteRange, and add a regression test.

blindchaser (approved):

sei-db/state_db/sc/hashvault/pebble_hashvault_rollback.go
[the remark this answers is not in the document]
cody-littley: fixed

cody-littley: fixed

Kbhat1 (approved):

PR #3510, opened 2026-05-27
Remove crisis module wiring and code

codchen:
## Summary

- remove x/crisis from the Sei app wiring and delete the sei-cosmos/x/crisis module plus its crisis protobuf definitions
- remove stale crisis wiring from wasmd/simapp test harnesses, local genesis setup, state-dump tooling, and module lists
- confirmed crisis does not have a dedicated rootmulti KVStore: its former state was params-subspace data under the params store, so this PR does not add a StoreUpgrades deletion for crisis
- skip stale/unknown module entries during streaming genesis import and validation, matching the non-streaming behavior for existing genesis files that still include crisis app_state

## Test Plan

- go test ./sei-cosmos/types/module
- go test ./sei-cosmos/x/genutil/client/cli
- go test ./app
- go test ./app ./app/legacyabci ./cmd/seid/cmd ./sei-wasmd/app ./sei-wasmd/x/wasm/keeper ./sei-ibc-go/testing/simapp/... ./sei-db/tools/cmd/seidb/operations
- 4-node Docker upgrade test: started the pre-removal parent binary 5ac709dc4, let the chain run, passed the software upgrade at height 306, restarted all validators on the crisis-deletion binary, and verified all nodes reached height 366 with catching_up=false and matching app hash 60C93CD1D72AD425D3FD862D115D722B098FEC48A28FD795C466807403F4648A


sei-ibc-go/testing/simapp/simd/cmd/root.go
```diff
@@ -170,8 +169,7 @@ func initRootCmd(rootCmd *cobra.Command, encodingConfig params.EncodingConfig) {
 	rootCmd.AddCommand(server.RosettaCommand(encodingConfig.InterfaceRegistry, encodingConfig.Marshaler))
 }
 
-func addModuleInitFlags(startCmd *cobra.Command) {
-	crisis.AddModuleInitFlags(startCmd)
```
masih, line 174: This would break the startup for anyone setting the flag. 

Perhaps a more graceful approach is to print a WARNING saying this is noop, and flag will be removed in the next release. 

Alternatively a more meaningful fatal error to point them to some TBD comms (e.g. release notes) that would explain the module removal).

app/app.go
```diff
@@ -1236,6 +1214,15 @@ func (app *App) SetStoreUpgradeHandlers() {
 		// configure store loader that checks if version == upgradeHeight and applies store upgrades
 		app.SetStoreLoader(upgradetypes.UpgradeStoreLoader(upgradeInfo.Height, &storeUpgrades))
 	}
+
+	if (upgradeInfo.Name == "v6.6.0") && !app.UpgradeKeeper.IsSkipHeight(upgradeInfo.Height) {
```
masih: Going forward since v6.5, upgrade names will follow `v<major>.<minor>` format. No more repetitive `.0` suffix.

```suggestion
	if (upgradeInfo.Name == "v6.6") && !app.UpgradeKeeper.IsSkipHeight(upgradeInfo.Height) {
```

app/app.go
masih, line 1243: Genesis JSON import [skips over](https://github.com/sei-protocol/sei-chain/blob/4972f18133874cdfb8c11b804c8ac2c8a36e8705/sei-cosmos/types/module/module.go#L399-L405) unknown module names. _but_ streaming genesis would i think panic because it seem to [directly index modules by name and proceed to initialise](https://github.com/sei-protocol/sei-chain/blob/4972f18133874cdfb8c11b804c8ac2c8a36e8705/sei-cosmos/types/module/module.go#L385).

If I haven't missed anything, we would then need to skip over removed modules [here](https://github.com/sei-protocol/sei-chain/blob/4972f18133874cdfb8c11b804c8ac2c8a36e8705/sei-cosmos/types/module/module.go#L385).

app/app.go
```diff
@@ -1236,6 +1214,15 @@ func (app *App) SetStoreUpgradeHandlers() {
...
+		storeUpgrades := storetypes.StoreUpgrades{
+			Deleted: []string{crisisStoreKeyName},
```
masih: But there is [no such key name](https://github.com/sei-protocol/sei-chain/blob/4972f18133874cdfb8c11b804c8ac2c8a36e8705/app/app.go#L263-L278) and this would result in an error i believe.

philipsu522 (approved):

philipsu522 (commented):
In the past, when updating some fields for modules I believe we saw some incompatibilities. For this particular PR, in the genesis file we see:
https://github.com/sei-protocol/chain-configs/blob/main/pacific-1/genesis.json#L715

We should probably test workflows like state sync'ing with the new binary and the genesis file

sei-ibc-go/testing/simapp/simd/cmd/root.go
codchen: This simapp is only used in unit tests so I think it should be okay, and once ibc is removed we will remove this simapp altogether too

app/app.go
codchen: Good catch, this was a valid issue. Fixed in 667afb9 by checking whether the streamed module exists before calling InitGenesis and skipping unknown modules, matching the non-streaming path. I also added a regression test with a streamed `crisis` genesis payload followed by a registered module.

While checking the adjacent paths, I found the same unknown-module hazard in streaming genesis validation and guarded that too.

app/app.go
```diff
@@ -920,7 +901,6 @@ func New(
 		EvmKeeper:        &app.EvmKeeper,
 	}
 	app.EndBlockKeepers = legacyabci.EndBlockKeepers{
-		CrisisKeeper:  &app.CrisisKeeper,
```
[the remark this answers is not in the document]
codchen: Fixed in 667afb9. Streaming genesis import now skips unknown module entries before calling InitGenesis, so existing chain genesis files that still include `crisis` app_state follow the same behavior as the non-streaming import path. Added a regression test for this exact `crisis` payload shape.

codchen:
Checked the genesis/state-sync concern.

The concern is valid for paths that stream or validate app_state from existing genesis files, because pacific-1/arctic-1/atlantic-2 genesis still include a `crisis` entry. Normal JSON genesis import already iterates registered modules and skips unknown module keys, so the removed `crisis` app_state is not used there. Standard state sync should rely on the genesis doc plus snapshots rather than running removed module InitGenesis from app_state, so I do not see a separate failure in that path from the `crisis` entry itself.

I did find and fix the risky streaming paths in 667afb9:
- streaming InitGenesis skips unknown modules before calling InitGenesis
- streaming ValidateGenesis drains/skips unknown module chunks instead of indexing a missing module
- added regression coverage using a streamed `crisis` genesis payload

Validated with:
- `go test ./sei-cosmos/types/module`
- `go test ./sei-cosmos/x/genutil/client/cli`
- `go test ./app`

PR #3512, opened 2026-05-27
feat(metrics): migrate sei-cosmos module packages to OpenTelemetry (PLT-414)

amir-deris:
## Summary

Migrates the following `sei-cosmos` packages from legacy `telemetry`/`go-metrics` to OpenTelemetry, using the dual-emit pattern (OTel instrument recorded first, old call retained with `TODO(PLT-414)` for removal once verified):

- **`types/module`** — `module_total_mid_block_duration`, `module_mid_block_duration` (per-module label)
- **`x/evidence`** — `begin_blocker_duration`
- **`x/capability`** — `begin_blocker_duration`
- **`x/crisis`** — `end_blocker_duration`, `init_genesis_unmarshal_duration`
- **`x/distribution`** — `begin_blocker_duration`
- **`x/gov`** — `end_blocker_duration`
- **`x/slashing`** — `begin_blocker_duration`; keeper: `validator_slashed` (with `type` + `validator` attributes)
- **`x/staking`** — `begin_blocker_duration`, `end_blocker_duration`

Each package gets a new `metrics.go` defining its OTel meter and instruments, following the pattern established in `x/upgrade` (PLT-353).


sei-cosmos/x/slashing/keeper/metrics.go
```diff
@@ -0,0 +1,27 @@
...
+
+var (
+	meter = otel.Meter("seicosmos_x_slashing_keeper")
+
+	slashingKeeperMetrics = struct {
+		validatorSlashed metric.Int64Counter
+	}{
+		validatorSlashed: must(meter.Int64Counter(
+			"validator_slashed",
+			metric.WithDescription("Number of validator slash events by type and validator address"),
```
bdchatham, line 16: **Drop `validator` from this counter; keep `type`.**

`type` (`double_sign` / `missing_signature`) is low-cardinality and the high-value dimension. `validator` is a consensus address — bounded by the active set at any instant, but it accretes monotonically over the chain's lifetime: every validator that's ever jailed/rotated/churned leaves a permanent series in a cumulative counter and never ages out.

One accuracy note for the description ("by type and validator address"): dropping `validator` does **not** preserve parity with the legacy series. The legacy `IncrValidatorSlashedCounter` (`sei-cosmos/telemetry/wrapper.go:84`) *does* carry `validator` as a label, so `sei_cosmos_validator_slashed` is per-validator today. Dropping it on the OTel side is a deliberate cardinality-hygiene choice that reduces granularity vs. legacy — worth framing as that rather than a no-op.

If per-validator slashing attribution is operationally needed, it belongs on a structured log / chain event — slashing is rare enough that a Loki query is the right tool, not a permanent metric label.

sei-cosmos/types/module/module.go
```diff
@@ -580,14 +582,21 @@ func (m Manager) RunMigrations(ctx sdk.Context, cfg Configurator, fromVM Version
...
+		telemetry.MeasureSince(midBlockStart, "module", "total_mid_block")
+	}()
 	for _, moduleName := range m.OrderMidBlockers {
 		module, ok := m.Modules[moduleName].(MidBlockAppModule)
 		if !ok {
 			continue
 		}
 		moduleStartTime := time.Now()
 		module.MidBlock(ctx, height)
+		moduleMetrics.midBlockDuration.Record(ctx.Context(), time.Since(moduleStartTime).Seconds(), otelmetric.WithAttributes(attribute.String("module", moduleName)))
```
bdchatham: **Pre-compute the per-module attribute set once at init rather than per block.**

This `WithAttributes(attribute.String("module", moduleName))` runs inline, per-module, every block — each call allocates a `[]attribute.KeyValue` and the SDK then builds an `attribute.Set` (sort + dedup + hash) to find the aggregator. Module names are static, so this is avoidable steady-state GC pressure on the consensus hot path.

Build the option once per module at startup and reuse it:
```go
// precomputed once, keyed by module name:
midBlockOpt := metric.WithAttributeSet(attribute.NewSet(attribute.String("module", moduleName)))
// in the loop:
moduleMetrics.midBlockDuration.Record(ctx.Context(), time.Since(moduleStartTime).Seconds(), midBlockOpt)
```

Begin/end blockers (no attributes) and the slashing counter (rare) don't need this — only the per-module mid-block loop.

bdchatham (commented):
Two focused follow-ups from a cross-review pass. (Both non-blocking on their own — the larger naming-collision discussion on the per-package `begin_blocker_duration` / `end_blocker_duration` instruments is separate.)

sei-cosmos/x/slashing/keeper/metrics.go
amir-deris: Thanks for feedback. I believe the validator address could be very useful for the slashing metrics, and it can potentially justify the cardinality risk, given there is limited number of validators and their address is fixed. I think also that's why the original metric had it:

```
 If you want a Grafana alert or dashboard panel for "how many times has validator X been double-signed?", a counter label is exactly the right tool. Loki is
  not a substitute for that — it's slower, harder to alert on, and requires log shipping to be reliable. For a slashing event that directly affects stake security, having it in Prometheus
  is the right call.
```

If ok, we can keep it and revisit if it becomes a problem?

sei-cosmos/types/module/module.go
amir-deris: Awesome feedback. Will use that!

bdchatham (approved):

masih (approved):

PR #3515, opened 2026-05-28
evmrpc: gate historical debug trace calls

codchen:
## Summary
- Reuse the existing `max_trace_lookback_blocks` config for all `debug_trace*` lookback checks.
  - `-1` keeps the existing unlimited-lookback behavior.
  - `0` means only the latest block is within lookback.
  - `>0` allows that many blocks behind latest.
- Count `debug_trace*` attempts that exceed the configured lookback with an endpoint/connection-labelled OTel counter.
- Preserve the existing block-number trace error semantics while extending the same guard to transaction, hash, traceCall, state access, and profile trace endpoints.

## Tests
- `go test ./evmrpc/config`
- `go test ./evmrpc -run 'Test(IsHistoricalDebugTraceBlock|GuardHistoricalDebugTraceHeight)$'`
- `go test ./evmrpc -run 'TestTrace(Transaction|Call|BlockByNumberLookbackLimit|BlockByNumberUnlimitedLookback)$'`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/evmrpc/tracers.go
+++ b/evmrpc/tracers.go
@@ -92,6 +92,80 @@ func (api *DebugAPI) prepareTraceContext(ctx context.Context) (context.Context, 
 	}, nil
 }
 
+func (api *DebugAPI) guardHistoricalDebugTraceByTxHash(ctx context.Context, endpoint string, hash common.Hash) error {
+	if api.keeper == nil {
+		return nil
+	}
+	receipt, err := api.keeper.GetReceipt(api.ctxProvider(LatestCtxHeight), hash)
+	if err != nil || receipt == nil {
+		return nil
+	}
+	return api.guardHistoricalDebugTraceHeight(ctx, endpoint, int64(receipt.BlockNumber)) //nolint:gosec
+}
+
+func (api *DebugAPI) guardHistoricalDebugTraceByNumber(ctx context.Context, endpoint string, number rpc.BlockNumber) error {
+	height, err := api.resolveDebugTraceBlockNumber(ctx, number)
+	if err != nil {
+		return err
+	}
+	return api.guardHistoricalDebugTraceHeight(ctx, endpoint, height)
+}
+
+func (api *DebugAPI) guardHistoricalDebugTraceByHash(ctx context.Context, endpoint string, hash common.Hash) error {
+	if api.backend == nil {
+		return nil
+	}
+	block, _, err := api.backend.BlockByHash(ctx, hash)
+	if err != nil || block == nil {
+		return nil
+	}
+	return api.guardHistoricalDebugTraceHeight(ctx, endpoint, int64(block.NumberU64())) //nolint:gosec
+}
+
+func (api *DebugAPI) guardHistoricalDebugTraceByNumberOrHash(ctx context.Context, endpoint string, blockNrOrHash rpc.BlockNumberOrHash) error {
+	if number, ok := blockNrOrHash.Number(); ok {
+		return api.guardHistoricalDebugTraceByNumber(ctx, endpoint, number)
+	}
+	if hash, ok := blockNrOrHash.Hash(); ok {
+		return api.guardHistoricalDebugTraceByHash(ctx, endpoint, hash)
+	}
+	return api.guardHistoricalDebugTraceHeight(ctx, endpoint, api.ctxProvider(LatestCtxHeight).BlockHeight())
+}
+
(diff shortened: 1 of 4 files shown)


evmrpc/config/config.go
```diff
@@ -121,6 +121,14 @@ type Config struct {
 	// Max number of blocks allowed to look back for tracing
 	MaxTraceLookbackBlocks int64 `mapstructure:"max_trace_lookback_blocks"`
 
+	// DisableHistoricalDebugTrace rejects debug_trace* calls whose target block is older than
+	// HistoricalDebugTraceBlockAge blocks behind the latest block.
+	DisableHistoricalDebugTrace bool `mapstructure:"disable_historical_debug_trace"`
+
+	// HistoricalDebugTraceBlockAge defines how many blocks behind latest a debug_trace*
+	// target can be before it is considered historical. Set to -1 to disable classification.
+	HistoricalDebugTraceBlockAge int64 `mapstructure:"historical_debug_trace_block_age"`
```
masih: Instead of "Age" would it be clearer to use the term "lookback"?

Because the former usually caries a notion of time, whereas the config here is simply a counter of relative offset from latest height.

evmrpc/config/config.go
```diff
@@ -121,6 +121,14 @@ type Config struct {
 	// Max number of blocks allowed to look back for tracing
 	MaxTraceLookbackBlocks int64 `mapstructure:"max_trace_lookback_blocks"`
 
+	// DisableHistoricalDebugTrace rejects debug_trace* calls whose target block is older than
+	// HistoricalDebugTraceBlockAge blocks behind the latest block.
+	DisableHistoricalDebugTrace bool `mapstructure:"disable_historical_debug_trace"`
```
masih: When we use the term "historical" do we refer to an explicit system component with the same name? I think not. I think that "historical" is the lingo we used internally to talk about this subject; right?

If that's right, then my suggestion is to remove the term "historical" entirely from config and variable names. What we really mean here is a lookback limit. We can then get rid of the dual flag of one boolean and one int "lookback amount", in favour of a single config, named something like `max_debug_trace_lookback`. Zero means ignored. Positive value means limit is applied to whatever is configured.

That simplifies things operationally speaking.

masih (commented):
Thinking out loud: Do we want to apply the limit to `sei_traceBlockByHashExcludeTraceFail` and `sei_traceBlockByNumberExcludeTraceFail` too? I would just to have consistent behaviour even though I think we disabled those endpoints by default. The edge case I am thinking of is incase someone has them enabled.  If we do decide to do this then I would drop the `debug_` from config names and call it something like `max_trace_lookback` with docs explaining where it applies etc. etc.

philipsu522 (approved):

evmrpc/tracers.go
```diff
@@ -92,6 +92,80 @@ func (api *DebugAPI) prepareTraceContext(ctx context.Context) (context.Context,
 	}, nil
 }
 
+func (api *DebugAPI) guardHistoricalDebugTraceByTxHash(ctx context.Context, endpoint string, hash common.Hash) error {
+	if api.keeper == nil {
+		return nil
+	}
+	receipt, err := api.keeper.GetReceipt(api.ctxProvider(LatestCtxHeight), hash)
+	if err != nil || receipt == nil {
+		return nil
```
masih, line 101: * Did we mean to `return err` if error is non-nil?
* if not, log the error in DEBUG level or something similar?

evmrpc/tracers.go
```diff
@@ -92,6 +92,80 @@ func (api *DebugAPI) prepareTraceContext(ctx context.Context) (context.Context,
...
+	return api.guardHistoricalDebugTraceHeight(ctx, endpoint, height)
+}
+
+func (api *DebugAPI) guardHistoricalDebugTraceByHash(ctx context.Context, endpoint string, hash common.Hash) error {
+	if api.backend == nil {
+		return nil
+	}
+	block, _, err := api.backend.BlockByHash(ctx, hash)
+	if err != nil || block == nil {
+		return nil
```
masih, line 120: ditto re returning error

masih (approved):
Much nicer thanks Tony 🙌 

Left couple of comments on error handing.

evmrpc/tracers.go
codchen: the geth behavior is to return null when receipt is not found, so we are aligning to that here

PR #3517, opened 2026-05-28
fix(storev2): hold read lock in GetCommitKVStore during concurrent map access

amir-deris:
## Summary

- Adds `RLock`/`RUnlock` to `GetCommitKVStore` in `rootmulti/store.go` so reads of the `ckvStores` map are properly synchronized against concurrent writes.
- Adds `TestGetCommitKVStoreNoDataRace` to exercise the concurrent read/write path; run with `go test -race ./sei-cosmos/storev2/rootmulti/...`.

## Test plan

- [x] `go test -race ./sei-cosmos/storev2/rootmulti/...` passes with no data race reports


bdchatham:
Could we tighten the test? Something like:

```go
func TestGetCommitKVStore_ReaderRespectsWriteLock(t *testing.T) {
    store := &Store{
        storeKeys: map[string]types.StoreKey{},
        ckvStores: map[types.StoreKey]types.CommitKVStore{},
    }
    key := types.NewKVStoreKey("bank")
    store.storeKeys[key.Name()] = key
    store.ckvStores[key] = mem.NewStore()

    store.mtx.Lock()

    readDone := make(chan types.CommitKVStore, 1)
    go func() {
        readDone <- store.GetCommitKVStore(key)
    }()

    select {
    case <-readDone:
        t.Fatal("GetCommitKVStore returned while write lock held — RLock missing")
    case <-time.After(50 * time.Millisecond):
    }

    newVal := mem.NewStore()
    store.ckvStores = map[types.StoreKey]types.CommitKVStore{key: newVal}
    store.mtx.Unlock()

    require.Same(t, newVal, <-readDone)
}
```

This deterministically verifies the `RLock` is doing its job — the read blocks while the write lock is held, then observes the post-write value — instead of relying on the scheduler to overlap two unsynchronized goroutines and the race detector to notice.

bdchatham (approved):

masih:
@codex review this pr

PR #3519, opened 2026-05-28
feat(mempool): observability for the TxMempool rewrite invariants

bdchatham:
Adds two metrics that expose the TxMempool rewrite's compaction behavior without changing logic: `tendermint_mempool_compact_total{trigger}` and `tendermint_mempool_compact_duration_seconds`. The `trigger` label distinguishes the three call sites of `compact()` (`insert_overflow`, `update`, `reap`); rate-of-`insert_overflow` is the capacity-pressure signal. Emitted via OpenTelemetry (`otel.Meter("tendermint_mempool")`) following the sei-db convention.

Companion to platform sei-protocol/platform#743 (recording rules, dashboard panels, alerts). The reactor-side gossip-bytes counter and any pending→ready promotion signal are separate followups — the latter is already directly tested by `mempool_pending_promotion_test.js` in the release-test suite, so a metric for it isn't yet pulling weight.

## Metric rationale

These metrics let us answer four questions about the rewrite that we couldn't ask cleanly before.

### What each metric is for, when stress-testing

**`compact_total{trigger="insert_overflow"}` is the capacity-pressure signal.** When this counter ticks, the mempool was momentarily over `hardLimit` (= 2× softLimit). A non-zero rate means ingress is outpacing block consumption — the only operational question is whether admission control is supposed to be catching it. If both `check_tx_met_drop_utilisation_threshold` *and* `insert_overflow` are rising, pressure escaped the gate.

**`compact_total{trigger="update"}` is the block heartbeat.** It fires once per Update, so its rate is a block-rate proxy. If it flatlines while seiload is still pushing, the consensus loop is stuck — different signal from `block_height_delta` because it's measured at the mempool, not the indexer.

**`compact_total{trigger="reap"}` is this node's proposer share.** Validators that propose blocks Reap; followers don't. Useful for telling "is this node doing block-proposal work?" without consulting Tendermint internals.

**`compact_duration_seconds` is where we'll catch the rewrite's real failure mode.** `compact()` is O(m log m) over the full mempool. As load grows, this latency grows. The number we'll watch is `rate(compact_total[5m]) × avg(compact_duration_seconds)` — wall-time-fraction spent inside compact. Once that approaches ~50%, compact is now blocking block production. The histogram lets us see this coming long before p99 actually hits block-interval (~200ms).

### The composite questions only these together can answer

1. **Does the rewrite's amortization actually hold under load?** `rate(insert_overflow) / rate(inserted_txs)` should stay sub-linear as throughput rises. If the ratio goes to 1, the 2× hardLimit amortization broke and we're back to per-insert prune.
2. **Is compact CPU-bound or block-bound?** Compare per-call duration to per-block compact rate. `compact_duration_p99` rising independently of `insert_overflow` rate means individual compacts got expensive (likely GC pressure on 100k+ entries); both rising means we're pruning more often AND each prune is slower — the failure mode.
3. **Is admission control doing its job or just delaying the problem?** `rate(check_tx_met_drop_utilisation_threshold) / rate(insert_overflow)` — if it's high, the gate is absorbing the spike. If it's near zero while overflow is high, the gate's threshold is mistuned.

### What we're still blind to

These metrics don't tell us *why* a compact was slow (sort vs. map-rebuild vs. GC) — that needs pprof or per-phase timing inside `compact()`. Per-peer gossip bandwidth is the next PR. The pending→ready promotion behavior is directly tested by `mempool_pending_promotion_test.js`; a continuous prod signal for it can be added later if operational need surfaces.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./internal/mempool/ -count=1` passes
- [ ] Validate emissions in a nightly chaos run
- [ ] Wire up platform-side recording rules + dashboard panels per #743

References: design doc sei-protocol/platform#743, TxMempool rewrite #3476.

### DIFF
--- a/sei-tendermint/internal/mempool/metrics.go
+++ b/sei-tendermint/internal/mempool/metrics.go
@@ -5,6 +5,8 @@ import (
 	"strconv"
 
 	"github.com/go-kit/kit/metrics"
+	stdprometheus "github.com/prometheus/client_golang/prometheus"
+	"github.com/sei-protocol/sei-chain/sei-tendermint/libs/utils"
 	"github.com/sei-protocol/sei-chain/sei-tendermint/types"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
@@ -24,15 +26,15 @@ var (
 		compactTotal           metric.Int64Counter
 		compactDurationSeconds metric.Float64Histogram
 	}{
-		compactTotal: must(mempoolMeter.Int64Counter(
+		compactTotal: utils.OrPanic1(mempoolMeter.Int64Counter(
 			"tendermint_mempool_compact_total",
 			metric.WithDescription("Number of compact() invocations, labeled by call site (insert_overflow, update, reap)."),
 		)),
-		compactDurationSeconds: must(mempoolMeter.Float64Histogram(
+		compactDurationSeconds: utils.OrPanic1(mempoolMeter.Float64Histogram(
 			"tendermint_mempool_compact_duration_seconds",
 			metric.WithDescription("Wall-clock duration of compact(), which re-sorts and rebuilds indices over the full mempool (O(m log m))."),
 			metric.WithUnit("s"),
-			metric.WithExplicitBucketBoundaries(0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 20, 30),
+			metric.WithExplicitBucketBoundaries(stdprometheus.ExponentialBucketsRange(0.001, 30, 14)...),
 		)),
 	}
 
@@ -40,13 +42,6 @@ var (
 	triggerUpdateAttr         = metric.WithAttributes(attribute.String("trigger", "update"))
 	triggerReapAttr           = metric.WithAttributes(attribute.String("trigger", "reap"))
 )
-
-func must[V any](v V, err error) V {
-	if err != nil {
-		panic(err)
-	}
-	return v
-}
 
 //go:generate go run ../../scripts/metricsgen -struct=Metrics
 
--- a/sei-tendermint/internal/mempool/metrics.go
+++ b/sei-tendermint/internal/mempool/metrics.go
@@ -113,25 +113,16 @@ type Metrics struct {
 	// utilisation was above the configured threshold. Note that not all such transactions are dropped, only those that also have a low priority.
 	CheckTxMetDropUtilisationThreshold metrics.Counter
 
-	// CompactTotal counts invocations of the txStore compact path, labeled by
-	// the triggering call site. trigger=insert_overflow fires when Insert pushes
-	// total past hardLimit; trigger=update fires on every Update recompute
-	// pass; trigger=reap fires when Reap is called with remove=true. Rate of
-	// trigger=insert_overflow is the capacity-pressure signal.
+	// Number of compact() invocations, labeled by call site
+	// (insert_overflow, update, reap).
 	CompactTotal metrics.Counter `metrics_labels:"trigger"`
 
-	// CompactDurationSeconds observes wall-clock duration of compact() — which
-	// re-sorts the mempool in priority order and rebuilds the byHash/byNonce
-	// indices. Complexity is O(m log m) over the full mempool, so the upper
-	// buckets accommodate large mempools (100k entries → 1-3s typical, 5-10s
-	// under GC pressure; +20s and +30s preserve quantile signal past that).
+	// Wall-clock duration of compact(), which re-sorts and rebuilds indices
+	// over the full mempool (O(m log m)).
 	CompactDurationSeconds metrics.Histogram `metrics_bucketsizes:"0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 20, 30"`
 
-	// PromotionTotal counts pending-to-ready transitions, labeled by the
-	// transaction lane that promoted (tx_type=evm — counted once per EVM nonce
-	// advance in the inline promotion loop in txStore.insert). Cosmos txs are
-	// auto-ready on insert and not counted here; future cosmos-side counts
-	// would add tx_type=cosmos without renaming.
+	// Number of pending-to-ready transitions, labeled by tx_type.
+	// Cosmos txs are auto-ready on insert and not counted.
 	PromotionTotal metrics.Counter `metrics_labels:"tx_type"`
 }
 
--- a/sei-tendermint/internal/mempool/metrics.go
+++ b/sei-tendermint/internal/mempool/metrics.go
@@ -6,22 +6,58 @@ import (
 
 	"github.com/go-kit/kit/metrics"
 	"github.com/sei-protocol/sei-chain/sei-tendermint/types"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
 )
 
 const (
 	// MetricsSubsystem is a subsystem shared by all metrics exposed by this
 	// package.
 	MetricsSubsystem = "mempool"
-
-	labelTrigger = "trigger"
-	labelTxType  = "tx_type"
-
-	triggerInsertOverflow = "insert_overflow"
-	triggerUpdate         = "update"
-	triggerReap           = "reap"
-
-	txTypeEVM = "evm"
 )
+
+// New metrics for the TxMempool rewrite are emitted via OpenTelemetry rather
+// than go-kit; go-kit is unmaintained and OTel is the strategic direction
+// across the sei-protocol monorepo (sei-db has been on OTel for some time).
+// The existing 22 Metrics fields above remain go-kit until a follow-up PR
+// migrates the whole package.
+var (
+	mempoolMeter = otel.Meter("tendermint_mempool")
+
+	otelMetrics = struct {
+		compactTotal           metric.Int64Counter
+		compactDurationSeconds metric.Float64Histogram
+		promotionTotal         metric.Int64Counter
+	}{
+		compactTotal: must(mempoolMeter.Int64Counter(
+			"compact_total",
+			metric.WithDescription("Number of compact() invocations, labeled by call site (insert_overflow, update, reap)."),
+		)),
+		compactDurationSeconds: must(mempoolMeter.Float64Histogram(
+			"compact_duration_seconds",
+			metric.WithDescription("Wall-clock duration of compact(), which re-sorts and rebuilds indices over the full mempool (O(m log m))."),
+			metric.WithUnit("s"),
+			metric.WithExplicitBucketBoundaries(0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 20, 30),
+		)),
+		promotionTotal: must(mempoolMeter.Int64Counter(
+			"promotion_total",
+			metric.WithDescription("Number of pending-to-ready transitions, labeled by tx_type. Cosmos txs are auto-ready on insert and not counted."),
+		)),
+	}
+
+	triggerInsertOverflowAttr = metric.WithAttributes(attribute.String("trigger", "insert_overflow"))
+	triggerUpdateAttr         = metric.WithAttributes(attribute.String("trigger", "update"))
+	triggerReapAttr           = metric.WithAttributes(attribute.String("trigger", "reap"))
+	txTypeEVMAttr             = metric.WithAttributes(attribute.String("tx_type", "evm"))
+)
+
+func must[V any](v V, err error) V {
+	if err != nil {
+		panic(err)
+	}
+	return v
+}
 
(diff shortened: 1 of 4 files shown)


sei-tendermint/internal/mempool/tx.go
```diff
@@ -432,7 +433,14 @@ func (s *txStore) Insert(wtx *WrappedTx) error {
...
 // O(m log m), prunes transactions above softLimit and recomputes all the indices.
-func (s *txStore) compact(inner *txStoreInner, clearAccounts bool) {
+// reason labels the triggering call site for the compact_total counter
+// (insert_overflow | update | reap).
+func (s *txStore) compact(inner *txStoreInner, clearAccounts bool, reason string) {
+	start := time.Now()
+	defer func() {
+		s.metrics.CompactTotal.With("reason", reason).Add(1)
+		s.metrics.CompactDurationSeconds.Observe(time.Since(start).Seconds())
+	}()
```
bdchatham: Can we just move this outside of the method call so you don't need to pass a string reason in?

bdchatham:
## Coral cross-review applied (commit `c92945f`)

Both platform-engineer and observability-platform-engineer reviewed the initial commit. All must-fix findings landed in this revision:

**Platform-engineer findings (Utilisation gauge staleness):**
- `Flush()` cleared the store but left gauges stale → set Size/PendingSize/TotalTxsSizeBytes/Utilisation to 0 after `Clear()`.
- `ReapTxs(remove=true)` evicts via `compact("reap")` but never re-emits the gauges → set them at end of ReapTxs when `remove=true`. (Same bug class existed for the pre-existing `Size` gauge — fixed together.)

**Observability findings:**
- `compact_total{reason}` → `compact_total{trigger}` — Prometheus convention bias; `reason` reads as failure cause, `trigger` reads as caller dimension. Free pre-scrape; expensive after.
- `promotion_total` → `promotion_total{tx_type="evm"}` — locks the EVM scope into a queryable dimension instead of metric-name lore; preserves the door for `tx_type="cosmos"` later without renaming.
- `compact_duration_seconds` buckets extended to `…, 5, 10, 20, 30` — the docstring set expectation of 5-10s under GC pressure, but the upper bucket of 10s meant p99 would saturate at +Inf exactly when needed most.

**Non-blocking findings (deferred):**
- One-line panic-defer comment (cosmetic)
- Hoisting `.With()` outside the defer (micro-opt; compact is per-block, not nanosecond-hot)

**Bug found in the consumer-side platform PR #743**: several PromQL queries reference Go-case names (`tendermint_mempool_PendingSize`, `tendermint_mempool_Size`) but the actual scrape names are snake_case (`pending_size`, `size`). Filing a fix-up to platform#743 separately.

Ready for human review when convenient.

sei-tendermint/internal/mempool/tx.go
```diff
@@ -346,6 +346,7 @@ func (s *txStore) insert(inner *txStoreInner, wtx *WrappedTx) error {
 			}
 			account.nextNonce += 1
 			state.ready.Inc(wtx.Size())
+			s.metrics.PromotionTotal.With("tx_type", "evm").Add(1)
```
bdchatham: Can you make both these strings constants so we don't allocate them every time this code path runs?

sei-tendermint/internal/mempool/metrics.go
```diff
@@ -103,6 +103,32 @@ type Metrics struct {
...
+	// transaction lane that promoted (tx_type=evm — counted once per EVM nonce
+	// advance in the inline promotion loop in txStore.insert). Cosmos txs are
+	// auto-ready on insert and not counted here; future cosmos-side counts
+	// would add tx_type=cosmos without renaming.
+	PromotionTotal metrics.Counter `metrics_labels:"tx_type"`
+
+	// Utilisation mirrors the same ratio the CheckTx drop gate evaluates:
+	// total count / (cfg.Size + cfg.PendingSize). Exposed as a gauge so
+	// recording rules don't need to re-derive it from Size+PendingSize.
+	Utilisation metrics.Gauge
```
bdchatham: I'm fine with dropping the utilisation gauge as long as it's pretty trivial to calculate with the other metrics

sei-tendermint/internal/mempool/metrics.go
```diff
@@ -103,6 +112,18 @@ type Metrics struct {
...
+	// (insert_overflow, update, reap).
+	CompactTotal metrics.Counter `metrics_labels:"trigger"`
+
+	// Wall-clock duration of compact(), which re-sorts and rebuilds indices
+	// over the full mempool (O(m log m)).
+	CompactDurationSeconds metrics.Histogram `metrics_bucketsizes:"0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 20, 30"`
+
+	// Number of pending-to-ready transitions, labeled by tx_type.
+	// Cosmos txs are auto-ready on insert and not counted.
+	PromotionTotal metrics.Counter `metrics_labels:"tx_type"`
```
amir-deris: I was wondering should these new metrics be OTel native from the start? it seems the `go-kit/metrics` is the old format that is tied to Prometheus. Also this repo doesn't seem to get updated in a while:
https://github.com/go-kit/kit

bdchatham:
Both reviewer comments addressed:

**@cursor[bot]** — fixed the `PromotionTotal` double-counting in `compact()` (commit \`d5d4a1a\`). The metric was incrementing for every re-inserted EVM tx during compact's `account.nextNonce = firstNonce` reset, so the rate would have been dominated by mempool size × block rate instead of new external promotions. `insert()` now takes a `countPromotion bool` flag that the public `Insert()` sets to true and `compact()`'s re-insertion path sets to false. The PR-description invariant `rate(promotion_total) ≈ rate(inserted_txs)` now holds.

**@amir-deris** — converted the 3 new metrics to OTel native (commit \`105bfd2\`). Following the sei-db convention (`sei-db/db_engine/pebbledb/mvcc/metrics.go`):

```go
var (
    mempoolMeter = otel.Meter("tendermint_mempool")
    otelMetrics = struct {
        compactTotal           metric.Int64Counter
        compactDurationSeconds metric.Float64Histogram
        promotionTotal         metric.Int64Counter
    }{ /* ... */ }
    triggerInsertOverflowAttr = metric.WithAttributes(attribute.String("trigger", "insert_overflow"))
    /* ... */
)

// Emit site:
otelMetrics.compactTotal.Add(context.Background(), 1, triggerInsertOverflowAttr)
otelMetrics.compactDurationSeconds.Record(context.Background(), time.Since(start).Seconds())
```

Attribute sets are package-level vars so emission is allocation-free on the hot path. The existing 22 go-kit metrics in this package remain untouched — they should move to OTel in a follow-up that handles the whole package consistently rather than mixing per-metric.

sei-tendermint/internal/mempool/tx.go
```diff
@@ -525,7 +528,10 @@ func (s *txStore) Update(spec updateSpec) {
 				wtx.priority = newPriority
 			}
 		}
+		start := time.Now()
 		s.compact(inner, true)
+		otelMetrics.compactTotal.Add(context.Background(), 1, triggerUpdateAttr)
```
bdchatham, line 533: Taking a follow-up to properly pass context in

bdchatham:
## Final Coral review round — one must-fix landed

Both experts re-reviewed at \`04eee9659\`. Platform-engineer returned \"ready\"; observability caught a real blocker.

**Naming mismatch** — fixed in \`a36521af\`. The OTel→Prom exporter in sei-chain registers with \`namespace=\"sei-chain\"\` (\`utils/metrics/metrics_util.go\`), so the meter name (\"tendermint_mempool\") becomes the \`otel_scope_name\` label, not a series prefix. With the short names \`compact_total\` / \`compact_duration_seconds\`, the exported series would have been \`sei_chain_compact_total\` — the platform PR #743's recording rules expecting \`tendermint_mempool_compact_total\` would have silently returned no data.

Fix: prefixed the metric names with \`tendermint_mempool_\` so the exported series is \`sei_chain_tendermint_mempool_compact_total\`. Matches the existing precedent (\`sei_chain_app_tx_count_total\` from \`meter=\"app\"\`, metric=\"tx_count\"\`) and sits alongside the existing go-kit \`tendermint_mempool_*\` metrics in queries.

Platform appendix in sei-protocol/platform#743 updated in parallel (\`f39756b\`) to reference \`sei_chain_tendermint_mempool_*\` in PromQL.

**Other findings — all green:**
- Histogram bucket spec preserved (14 boundaries + \`+Inf\` + \`_count\` + \`_sum\` = 17 series/instance)
- Cardinality recount: ~5,000 series net-new
- Mixed go-kit + OTel coexists cleanly on the same \`/metrics\` endpoint
- Attribute-set pre-baking is safe (no aliasing)
- OTel hot-path emit cost ~50-200ns/call, comparable to go-kit
- Bugbot's pre-fix finding was a real bug we should have caught in round 1

Ready for human review.

sei-tendermint/internal/mempool/metrics.go
```diff
@@ -14,6 +17,37 @@ const (
...
+			metric.WithExplicitBucketBoundaries(0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 20, 30),
+		)),
+	}
+
+	triggerInsertOverflowAttr = metric.WithAttributes(attribute.String("trigger", "insert_overflow"))
+	triggerUpdateAttr         = metric.WithAttributes(attribute.String("trigger", "update"))
+	triggerReapAttr           = metric.WithAttributes(attribute.String("trigger", "reap"))
+)
+
+func must[V any](v V, err error) V {
```
pompon0: nit: use sei-tendermint/libs/utils.OrPanic1

sei-tendermint/internal/mempool/metrics.go
```diff
@@ -14,6 +17,37 @@ const (
...
+	}{
+		compactTotal: must(mempoolMeter.Int64Counter(
+			"tendermint_mempool_compact_total",
+			metric.WithDescription("Number of compact() invocations, labeled by call site (insert_overflow, update, reap)."),
+		)),
+		compactDurationSeconds: must(mempoolMeter.Float64Histogram(
+			"tendermint_mempool_compact_duration_seconds",
+			metric.WithDescription("Wall-clock duration of compact(), which re-sorts and rebuilds indices over the full mempool (O(m log m))."),
+			metric.WithUnit("s"),
+			metric.WithExplicitBucketBoundaries(0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 20, 30),
```
pompon0: is this some reusable standard bucket list? If not can we use exponential buckets instead of harcoding a list of values?

sei-tendermint/internal/mempool/tx.go
```diff
@@ -589,7 +595,10 @@ func (s *txStore) Reap(l ReapLimits, remove bool) (types.Txs, int64) {
 					s.readyTxs.Remove(el)
 				}
 			}
+			start := time.Now()
```
pompon0, line 598: time measurement should be done within s.compact to avoid code duplication.

pompon0 (approved):

masih (approved):

sei-tendermint/internal/mempool/tx.go
bdchatham: The dimension of the metric depends on where this is called and instead of specifying that as an argument just having the caller own this.

bdchatham:
@pompon0 — addressed in \`bd7f50392\`:

1. **\`utils.OrPanic1\`** — swapped the local \`must[V any]\` helper for the shared one.
2. **Exponential buckets** — switched to \`stdprometheus.ExponentialBucketsRange(0.001, 30, 14)\` (the same helper used in \`internal/consensus/metrics.gen.go\`). 14 boundaries spanning 1ms to 30s with a ~2.32× factor per bucket.
3. **Timing inside \`compact()\`** — keeping it at the call sites. \`compact()\` stays pure (no OTel emission inside) and the per-\`trigger\` counter increment is already at the caller, so the timing sits naturally next to it. The duplication is 2 lines × 3 sites and visually clear; moving timing inside would either require a label parameter (which the prior review pass cut) or split the histogram/counter emission across two locations.

sei-tendermint/internal/mempool/metrics.go
```diff
@@ -14,6 +19,30 @@ const (
...
+	otelMetrics = struct {
+		compactTotal           metric.Int64Counter
+		compactDurationSeconds metric.Float64Histogram
+	}{
+		compactTotal: utils.OrPanic1(mempoolMeter.Int64Counter(
+			"tendermint_mempool_compact_total",
+			metric.WithDescription("Number of compact() invocations, labeled by call site (insert_overflow, update, reap)."),
+		)),
+		compactDurationSeconds: utils.OrPanic1(mempoolMeter.Float64Histogram(
+			"tendermint_mempool_compact_duration_seconds",
```
amir-deris, line 34: It probably doesn't need `_seconds` in the name. That will be added by default.

amir-deris (approved):

PR #3522, opened 2026-05-29
autobahn mempool

pompon0:
Mempool constructed specifically for autobahn tx lifecycle:
* mempool rejects pending (non-ready) rpcs 
* mempool insert rpcs are blocking if mempool is full (except for BroadcastTxAsync)
* mempool capacity is counted in local lane blocks - it can contain up to types.BlocksPerLane blocks since the last locally executed lane block
* in particular mempool tracks EVM nonces until the local block is locally executed, at which point it can confirm whether the account nonces have been bumped as expected.
* blocks are pushed from mempool to lane in avail.State as soon as there is capacity in avail.State (which is different than mempool capacity, because mempool counts blocks wrt to local execution, while avail.State counts blocks wrt to global execution)
* full local blocks are constructed as soon as there is enough txs in the mempool
* non-full local blocks are constructed after a BlockInterval timeout. Empty blocks are supported for now, because integration tests require that.


sei-tendermint/internal/autobahn/producer/mempool.go
```diff
@@ -0,0 +1,167 @@
...
+						delete(x.evmNonces, addr)
+					}
+				}
+			}
+		}
+		// n > m.next shouldn't really happen,
+		// because local mempool is the only source of local lane blocks,
+		// but we handle it gracefully anyway.
+		m.next = max(m.next, n)
+	}
```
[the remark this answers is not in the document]
pompon0: WAI

sei-tendermint/internal/autobahn/producer/mempool.go
```diff
@@ -0,0 +1,171 @@
...
+		b := m.nextBlock
+		b.gasEstimated += utils.Clamp[uint64](gasEstimated)
+		b.gasWanted += utils.Clamp[uint64](resp.GasWanted)
+		b.sizeBytes += uint64(len(tx))
+		b.txs = append(b.txs, tx)
+		if resp.IsEVM {
+			addr := resp.EVMSenderAddress
+			b.evmNonces[addr] = m.evmNonces[addr]
+		}
+	}
```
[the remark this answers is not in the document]
pompon0: fixed

sei-tendermint/internal/rpc/core/mempool.go
```diff
@@ -36,7 +37,15 @@ func (env *Environment) EvmProxy(sender common.Address) (*url.URL, bool) {
 // https://docs.tendermint.com/master/rpc/#/Tx/broadcast_tx_async
 // Deprecated and should be removed in 0.37
 func (env *Environment) BroadcastTxAsync(ctx context.Context, req *coretypes.RequestBroadcastTx) (*coretypes.ResultBroadcastTx, error) {
-	go func() { _, _ = env.Mempool.CheckTx(ctx, req.Tx) }()
+	if giga, ok := env.gigaRouter().Get(); ok {
+		go func() { _, _ = giga.Mempool().InsertTx(ctx, req.Tx) }()
+		return &coretypes.ResultBroadcastTx{Hash: req.Tx.Hash().Bytes()}, nil
```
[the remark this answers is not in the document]
pompon0: added TryInsertTx which specifically does not block when mempool is full, avoiding interaction with expired context AND preventing accumulation of background routines.

sei-tendermint/internal/autobahn/producer/state.go
```diff
@@ -5,135 +5,149 @@ import (
...
+						// Wait for non-empty payload.
+						if err := ctrl.WaitUntil(ctx, func() bool {
+							return toProduce < m.next || (toProduce == m.next && m.CanPushBlock())
+						}); err != nil {
+							return err
+						}
+						// Seal the payload if needed.
+						if toProduce == m.next {
+							m.PushBlock()
+						}
```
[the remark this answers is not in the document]
pompon0: done

sei-tendermint/internal/autobahn/producer/mempool.go
```diff
@@ -0,0 +1,174 @@
...
+					// Happy path: all account's txs got executed.
+					delete(m.evmNonces, addr)
+				} else if gotNonce := s.cfg.App.EvmNonce(addr); gotNonce < wantNonce {
+					// Some txs have not been executed - reset account tracking.
+					// NOTE: app execution is not synchronized with mempool, so nonce could have already
+					// proceeded past wantNonce and that is expected.
+					delete(m.evmNonces, addr)
+					for _, x := range m.blocks {
+						delete(x.evmNonces, addr)
+					}
```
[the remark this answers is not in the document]
pompon0: true, fixed

sei-will (approved):

sei-tendermint/internal/autobahn/producer/mempool.go
```diff
@@ -0,0 +1,192 @@
...
+	if gasWanted > s.cfg.MaxGasPerBlock {
+		return nil, errTooLarge
+	}
+
+	for m, ctrl := range s.mempool.Lock() {
+		if waitIfFull {
+			// mempool is constructed as a FIFO - we do not delay insertions of large txs (going over cap)
+			// in favor of waiting for smaller txs. This simple algorithm allows us to cap
+			// pending txs to size of a single block. We can refine this rule later if needed.
+			if err := ctrl.WaitUntil(ctx, func() bool { return !m.IsFull() }); err != nil {
```
wen-coding, line 175: This might be fine for the loadtest but I'd already worry a bit for the public testnet. My main concern is that if we shut the door to everyone when the pool is full, we can reject high-value user txs when a spammer fills the pool with valid txs with valid nonces.
Two suggestions:
1. per-sender limit:
```
if err := ctrl.WaitUntil(ctx, func() bool {
    return !m.IsFull() && m.pending[addr] < m.perSenderCap
}); err != nil { ... }
```
2. replace FIFO with "replacing with more expensive tx", let the economics do its work.

sei-tendermint/internal/autobahn/producer/mempool.go
```diff
@@ -0,0 +1,197 @@
...
+	evmNonces map[common.Address]uint64
+}
+
+func (m *mempool) IsFull() bool {
+	return uint64(m.next-m.first) >= m.capacity && len(m.nextBlock.txs) > 0
+}
+
+func (m *mempool) CanSealBlock(allowEmpty bool) bool {
+	return uint64(m.next-m.first) < m.capacity && (allowEmpty || len(m.nextBlock.txs) > 0)
+}
```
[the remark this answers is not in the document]
pompon0: WAI

sei-tendermint/node/setup.go
```diff
@@ -253,15 +251,15 @@ func buildGigaConfig(
...
 		Producer: &producer.Config{
-			MaxGasPerBlock:   maxGasPerBlock,
-			MaxTxsPerBlock:   fc.MaxTxsPerBlock,
-			MaxTxsPerSecond:  fc.MaxTxsPerSecond,
-			MempoolSize:      fc.MempoolSize,
-			BlockInterval:    time.Duration(fc.BlockInterval),
-			AllowEmptyBlocks: fc.AllowEmptyBlocks,
+			App:                     app,
+			MaxGasWantedPerBlock:    genDoc.ConsensusParams.Block.MaxGasWantedUint64(),
+			MaxGasEstimatedPerBlock: genDoc.ConsensusParams.Block.MaxGasUint64(),
```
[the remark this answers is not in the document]
pompon0: backward compatible with tendermint

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -521,11 +525,11 @@ func (s *State) fullCommitQC(ctx context.Context, n types.RoadIndex) (*types.Ful
 }
 
 // WaitForCapacity waits until the given lane has enough capacity for a new block.
-func (s *State) WaitForCapacity(ctx context.Context, lane types.LaneID) error {
+func (s *State) WaitForCapacity(ctx context.Context, toProduce types.BlockNumber) error {
```
wen-coding: nit: actually this is only waiting for your own capacity since we don't provide lane? Do we need to hint on that on function name?

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -521,11 +525,11 @@ func (s *State) fullCommitQC(ctx context.Context, n types.RoadIndex) (*types.Ful
...
+	lane := s.key.Public()
 	for inner, ctrl := range s.inner.Lock() {
-		q := inner.blocks[lane]
 		if err := ctrl.WaitUntil(ctx, func() bool {
-			return q.next < inner.persistedBlockStart[lane]+BlocksPerLane
+			return toProduce < inner.persistedBlockStart[lane]+BlocksPerLane
```
wen-coding, line 532: I know this is pre-existing, we probably need to change inner.persistedBlockStart[lane] to appropriate value when validators can join in the middle

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -564,23 +568,24 @@ func (s *State) WaitForLaneQCs(
 
 // ProduceBlock appends a new block to the producers lane.
 // Blocks until the lane has enough capacity for the new block.
```
wen-coding: hmm, I don't you do blocking here any more.

sei-tendermint/internal/autobahn/data/state.go
```diff
@@ -594,6 +598,43 @@ func (s *State) AppProposal(ctx context.Context, n types.GlobalBlockNumber) (*ty
...
+	// TODO(gprusak): decide whether 0 is a good result in this case in general.
+	if i.first == i.nextAppProposal {
+		return 0
+	}
+	n := i.nextAppProposal - 1
+	r := i.qcs[n].QC().LaneRange(lane)
+	// TODO: this header can be actually extracted from FullCommitQC, so consider moving all this logic there.
+	h := i.blocks[n].Header()
+	x := lane.Compare(h.Lane())
+	// NOTE: here we assume the specific ordering of lane blocks in the CommitQC:
```
wen-coding, line 611: Do we actually check the ordering now in the CommitQC?

sei-tendermint/internal/autobahn/producer/mempool.go
```diff
@@ -0,0 +1,197 @@
...
+			return nil, errMempoolFull
+		}
+		if resp.IsEVM {
+			addr := resp.EVMSenderAddress
+			nonce, ok := m.evmNonces[addr]
+			if !ok {
+				nonce = s.cfg.App.EvmNonce(addr)
+			}
+			if nonce != resp.EVMNonce {
+				return nil, fmt.Errorf("%w: got %v, want %v", errBadNonce, resp.EVMNonce, nonce)
```
wen-coding, line 188: So right now we only accept the next Nonce? Say the user sent tx with nonce 100 and 101, the packet containing 100 got dropped, 101 will rejected?

And also we don't allow higher-priority tx to replace old ones any more?

I guess it might be okay for the loadtest, but for public testnet this doesn't work. Maybe add a TODO?

For loadtest we also need to remember to retry in sequence if the loadtest client died.

sei-tendermint/internal/autobahn/producer/state.go
```diff
@@ -5,135 +5,153 @@ import (
...
-				return fmt.Errorf("s.Data().WaitForCapacity(): %w", err)
-			}
-			payload, err := s.nextPayload(ctx)
-			if err != nil {
-				return fmt.Errorf("s.nextPayload(): %w", err)
+		availState := s.consensus.Avail()
+		lane := availState.PublicKey()
+		firstBlock := s.mempoolFirst()
+		scope.Spawn(func() error {
+			// Task pruning executed lane blocks from the mempool
```
wen-coding, line 80: Not saying right or wrong, just an observation.

In non-Autobahn we used to have success/fail results from TxResults, so there is a feedback about individual tx. Now it's per-account. I think originally we allowed second chance for failed tx, do we care about allowing that here? (I imagine if you are landing a transfer tx and sending enough gas at the same time, but your "send enough gas" tx got delayed, your transfer tx can maybe get a second chance?)

Maybe document the behavior?

sei-tendermint/internal/autobahn/producer/mempool.go
pompon0: This mempool implementation is minimal, in a sense that it does not really collect transactions to build blocks out of, but rather immediately sequences the received txs. It has the following consequences:
* as long as the txs insertion rate is below throughput limit (200k/40 = 5k tps), it does not matter
* it is end user responsibility to insert per-account transactions in order and retry in case of problems (I think this is a reasonable UX, but TBD)
* InsertTx(), and therefore eth_sendRawTransaction calls are blocking rather than rejecting in case mempool is full. I think this is a desirable semantics, given that the alternative would be for the client to spam retries, which is a waste of traffic for everyone. Whether it is good enough for standard tooling is TBD, I suppose.

The actual "mempool" here (the unsequenced transactions) are the blocked rpcs which will be sequenced/rejected once there is available capacity. This part indeed might need some structure before public testnet. For private testnet and loadtest imo it would be just extra computational burden.

sei-tendermint/internal/autobahn/data/state.go
pompon0: I meant FullCommitQC, which contains information about block ordering within a tipcut

sei-tendermint/internal/autobahn/producer/mempool.go
pompon0: the packet won't get dropped, because we support HTTP and WS which are both built on top of TCP. We don't allow replacement, because the lower priority already gets sequenced on the lane when the higher priority tx arrives, it does not make sense. Replacement doesn't make also sense, because there is no gossip so there is no need of reconciliation of txs conflicting on nonces. Mempool simply sequences a tx within a lane and then forgets about it - user is free to send a different tx with the same nonce and arbitrary priority as soon as the previous tx fails to get executed. Loadtest client is responsible for retrying txs. It has enough information to detect problem and resend what is needed.

sei-tendermint/internal/autobahn/producer/state.go
pompon0: it is compatible with the TxMempool of tendermint: failed transactions are removed from the mempool and are allowed to be reinserted. The "second chance" limitation is on the cache level. There is no cache of bad transaction for the autobahn mempool currently so in particular there is no restriction on 2 attempts per transaction.

sei-tendermint/internal/autobahn/producer/mempool.go
pompon0: both suggestions are reasonable afaic, but let's implement them after private testnet.

sei-tendermint/internal/autobahn/avail/state.go
pompon0: wdym?

sei-tendermint/internal/autobahn/producer/mempool.go
wen-coding: Sure, just add a TODO maybe?

sei-tendermint/internal/autobahn/producer/mempool.go
wen-coding: TCP guarantees byte order, if there are two HTTP connections sending tx N and tx N+1, and they happen to be in wrong order, I don't think TCP can fix that? If there is some tool which fires concurrent txs without awaiting sequence, they could be rejected even if user didn't do anything wrong?

Depending on how our loadtest tool is written, if it always awaits tx N before sending tx N+1, then this won't be a problem. But I'd be a bit worried for public testnet.

sei-tendermint/internal/autobahn/avail/state.go
wen-coding: I'm just panicking we are always setting inner.persistedBlockStart[lane] to 0 when there is no WAL on the disk. Not related to this PR, I think we will fix that in the state sync work?

sei-tendermint/internal/autobahn/data/state.go
wen-coding: Okay, I vaguely recall we do check ordering on ingesting CommitQC, so we should be good here.

sei-tendermint/internal/autobahn/producer/state.go
wen-coding: I think your description is accurate for failed-tx cache. But for under-funded txs at admission, I think those are held in pending queue and re-assessed when execution results come in. It's okay if we force user to retry in this case, dunno how much it matters.

wen-coding (approved):

sei-tendermint/internal/autobahn/producer/mempool.go
pompon0: > Depending on how our loadtest tool is written, if it always awaits tx N before sending tx N+1, then this won't be a problem

I think it is a reasonable expectation.

sei-tendermint/internal/autobahn/avail/state.go
pompon0: renamed to WaitForLocalCapacity

sei-tendermint/internal/autobahn/avail/state.go
pompon0: I mean as soon as anchor is fetched from the network, persistedBlockStart is also moved forward. avail state is supposed to be independent from statesync.

PR #3524, opened 2026-05-29
fix(tokenfactory): add pagination to DenomsFromCreator query (PLT-410)

amir-deris:
## Summary

- `DenomsFromCreator` performed an unbounded full prefix scan with no limit, making it unsafe for external callers.
- Added `pagination` field to `QueryDenomsFromCreatorRequest` and `QueryDenomsFromCreatorResponse`, routing the **gRPC handler** through `query.Paginate` which enforces the MaxLimit/MaxOffset bounds added in PLT-361 (#3494).
- The **wasm query path** deliberately bypasses pagination and continues to return all denoms unbounded. Wasm contract execution is gas-metered, so unbounded iteration does not pose a DoS risk, and changing this behavior would be app-hash breaking for any deployed contract that makes state-changing decisions based on the full denom set.

## Changes

- `proto/tokenfactory/query.proto` — added `PageRequest`/`PageResponse` pagination fields to `DenomsFromCreator` request and response
- `x/tokenfactory/types/query.pb.go` — regenerated via `scripts/protoc.sh`
- `x/tokenfactory/types/query.pb.gw.go` — regenerated; REST gateway now reads `?pagination.*` query parameters
- `x/tokenfactory/keeper/creators.go` — added paginated `getDenomsFromCreator` for the gRPC path; added `GetAllDenomsFromCreator` (unbounded iterator) for the wasm path
- `x/tokenfactory/keeper/grpc_query.go` — threads pagination through the gRPC handler
- `x/tokenfactory/client/cli/query.go` — exposes `--page-*` flags on the `denoms-from-creator` command
- `x/tokenfactory/client/wasm/query.go` — wasm handler calls `GetAllDenomsFromCreator` directly, preserving pre-pagination behavior
- `x/tokenfactory/keeper/grpc_query_test.go` — added pagination tests (no pagination, limit, key-based cursor, offset)
- `wasmbinding/test/query_test.go` — updated assertions to compare only `Denoms` field

## Notes

- gRPC callers that omit pagination get `DefaultLimit = 100` results; hard caps (MaxLimit = 1000, MaxOffset = 10000) are enforced by `query.Paginate` once PLT-361 lands
- Wasm path is intentionally unbounded — changing it would risk app-hash divergence on deployed contracts that iterate the full denom set during tx execution


x/tokenfactory/keeper/creators.go
```diff
@@ -2,24 +2,25 @@ package keeper
...
+	var denoms []string
+	pageRes, err := query.Paginate(store, pagination, func(key []byte, _ []byte) error {
+		denoms = append(denoms, string(key))
+		return nil
+	})
+	if err != nil {
+		return nil, nil, err
 	}
-	return denoms
+	return denoms, pageRes, nil
```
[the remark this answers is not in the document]
amir-deris: Increased the default limit to 2000 instead of 100 as a sensible default for backward compatiblity.

masih:
@claude review this pr

x/tokenfactory/client/wasm/query.go
```diff
@@ -2,10 +2,16 @@ package wasm
...
 	sdk "github.com/sei-protocol/sei-chain/sei-cosmos/types"
+	"github.com/sei-protocol/sei-chain/sei-cosmos/types/query"
 	tokenfactorykeeper "github.com/sei-protocol/sei-chain/x/tokenfactory/keeper"
 	"github.com/sei-protocol/sei-chain/x/tokenfactory/types"
 )
 
+// defaultDenomsFromCreatorLimit is the wasm query default, higher than the gRPC DefaultLimit of 100
+// for backwards compatibility with contracts that expect all denoms in one response, but still
+// bounded to limit DoS risk since denom creation has no fee.
+const defaultDenomsFromCreatorLimit = 2000
```
masih: Does this value make sense given https://github.com/sei-protocol/sei-chain/pull/3494 is open?

2K will be rejected once that PR lands. What are your thoughts in terms of coordinating the two PRs?

proto/tokenfactory/query.proto
```diff
@@ -69,12 +70,14 @@ message QueryDenomAuthorityMetadataResponse {
...
   string creator = 1 [(gogoproto.moretags) = "yaml:\"creator\""];
+  cosmos.base.query.v1beta1.PageRequest pagination = 2;
 }
 
-// QueryDenomsFromCreatorRequest defines the response structure for the
+// QueryDenomsFromCreatorResponse defines the response structure for the
 // DenomsFromCreator gRPC query.
 message QueryDenomsFromCreatorResponse {
   repeated string denoms = 1 [(gogoproto.moretags) = "yaml:\"denoms\""];
+  cosmos.base.query.v1beta1.PageResponse pagination = 2;
```
masih, line 80: I think this would be an apphash breaking change but the PR is labeled as non-apphash breaking.

masih (commented):
I think we still have a pagination issue [here](https://github.com/sei-protocol/sei-chain/blob/8cf29d8b9f4619a279549c909330eef26d8de6d7/sei-cosmos/types/query/pagination.go#L55)? totally fine if that's going to be resolved in a follow up PR

x/tokenfactory/client/wasm/query.go
amir-deris: Thanks for great feedback, that is a great point. I will modify this to reference the `query.MaxLimit` instead which should still work and not conflict with PR [3494](https://github.com/sei-protocol/sei-chain/pull/3494)

proto/tokenfactory/query.proto
amir-deris: Thanks for feedback. It seems that is not the case regarding app hash breaking change, since this is a protocol buffer query and results are not part of the state. Do we have any test that would fail if this was an app hash breaking change and incorrectly labeled `non-app-hash-breaking`?
```
The app hash is the Merkle root of committed state — it only changes when something alters what gets written to the KV store during block execution (DeliverTx, BeginBlock, EndBlock, InitChain). The changes here
  are entirely in the query plane:

  - QueryDenomsFromCreatorRequest and QueryDenomsFromCreatorResponse are only used by the gRPC/REST query handler — they never touch state, never go through consensus, and are never committed.
  - query.pb.go / query.pb.gw.go regeneration only affects how query messages are serialized over the wire; it has no effect on the state machine.
  - x/tokenfactory/keeper/creators.go now calls query.Paginate instead of a raw iterator, but getDenomsFromCreator is a read-only keeper method — it doesn't write anything.

  The only thing that could make a query-proto change app-hash-breaking is if the query result were hashed into state somewhere (e.g. stored as a module param or emitted as a deterministic event used in consensus),
  which isn't the case here.```

**Comment by amir-deris** [comment]
> I think we still have a pagination issue [here](https://github.com/sei-protocol/sei-chain/blob/8cf29d8b9f4619a279549c909330eef26d8de6d7/sei-cosmos/types/query/pagination.go#L55)? totally fine if that's going to be resolved in a follow up PR

@masih I believe that issue is going to be resolved in this pr: https://github.com/sei-protocol/sei-chain/pull/3494/ so there shouldn't be any pagination issue 🤔 Please correct me if I am wrong.

File: x/tokenfactory/client/wasm/query.go [hunk 4]
```diff
@@ -23,5 +28,10 @@ func (handler TokenFactoryWasmQueryHandler) GetDenomAuthorityMetadata(ctx sdk.Co
 
 func (handler TokenFactoryWasmQueryHandler) GetDenomsFromCreator(ctx sdk.Context, req *types.QueryDenomsFromCreatorRequest) (*types.QueryDenomsFromCreatorResponse, error) {
 	c := sdk.WrapSDKContext(ctx)
+	if req.Pagination == nil {
+		req.Pagination = &query.PageRequest{Limit: defaultDenomsFromCreatorLimit}
+	} else if req.Pagination.Limit == 0 {
+		req.Pagination.Limit = defaultDenomsFromCreatorLimit
+	}
```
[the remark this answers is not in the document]
amir-deris: The issue goes away once PR #3494 lands. Here's why:

  PR #3494 changes MaxLimit from math.MaxUint64 to 1_000. With that value, the overflow in Paginate:

  end := offset + limit  // offset + 1_000

  ...can't realistically overflow a uint64 — you'd need an offset of ~18.4 * 10^15 to wrap around. No real query would have that.

  PR #3494 also adds VerifyPaginationLimit which rejects any limit > MaxLimit, but limit == MaxLimit (1000) is accepted, so setting defaultDenomsFromCreatorLimit = query.MaxLimit still works.

  The net effect after both PRs land: The wasm handler defaults limit=0 to query.MaxLimit (now 1000), and end = offset + 1000 is always safe. The overflow bug from the Bugbot feedback only exists in
  the window between this PR and PR #3494 — and only for the extremely niche input of {pagination: {offset: N, limit: 0}}.

  So the feedback was technically valid against the current code, but isn't worth addressing separately since #3494 is close to merging.

bdchatham (approved):

PR #3525, opened 2026-05-30
Add Autobahn fullnode (CON-309)

wen-coding:
## Summary

Adds a non-validator **fullnode** role to Autobahn. A fullnode loads the committee from `autobahn.json` as a routing table, dials a single committee member at a time over giga to pull finalized blocks (`StreamFullCommitQCs` + `GetBlock`), executes them locally via `runExecute` (FinalizeBlock + Commit + PushAppHash), and accepts inbound block-sync from other peers. Read endpoints (`eth_call`, `eth_getBalance`, `eth_getTransactionReceipt`, `/block`, `/status`) are served from the locally-executed state; `eth_sendRawTransaction` is the only thing forwarded — via `EvmProxy` to the shard-owning validator, because submission needs the producer's mempool.

**Role follows `cfg.Mode`**: `mode = "validator"` runs the validator path (requires the local key to be in the committee, and rejects `priv-validator.laddr` since autobahn signs in-process); any other mode runs as a fullnode. Mode is the operator's explicit role declaration, kept separate from committee membership so a newly-joined committee member can finish catch-up as a fullnode before the operator flips to `mode = "validator"`. A WARN is logged at startup if mode and committee membership disagree, so operator misconfiguration is visible.

**Type split.** `GigaRouter` is now built by two separate constructors:

- `NewGigaValidatorRouter(*GigaValidatorConfig, ...)` — committee members. Embeds `GigaRouterCommonConfig` and adds `ValidatorKey`, `ViewTimeout`, `Producer`. Runs consensus + producer + the full giga service.
- `NewGigaFullnodeRouter(*GigaRouterCommonConfig, ...)` — non-validators. No fullnode-specific config fields — the common config is sufficient. No consensus, no producer, no mempool. Single-active-subscriber dial loop: shuffled-once committee walk, `cfg.DialInterval` between attempts.

`GigaRouter` interface exposes the read path (`Run`, `RunInboundConn`, `LastCommittedBlockNumber`, `MaxGasEstimatedPerBlock`, `BlockByNumber`, `BlockByHash`, `EvmProxy`, `Mempool`). `Mempool() utils.Option[*producer.State]` returns `Some(r.producer)` on validators and `None` on fullnodes — the producer-backed mempool that `internal/rpc/core/mempool.go` and `rpc/client/local/local.go` reach through. `dialAndRunConn` takes `utils.Option[NodePublicKey]` for the expected peer key: validators pass `Some(addr.Key)` (consensus voting requires peer identity), fullnodes pass `None` (block-sync data is QC-signed, so peer identity doesn't need a dial-layer check).

**Inbound block-sync on both roles.** Validators and fullnodes both expose `RunInboundConn`. Committee peers get the full giga RPC on validators (`RunServer`); non-committee peers get the block-sync subset (`RunBlockSyncServer`: `StreamFullCommitQCs` + `GetBlock`), capped per node by `max_inbound_fullnode_peers` in autobahn.json (default `10`; `Option[uint64]` so absent and explicit-zero are distinguishable). Cap is an `atomic.Int64` counter on the hot path. Enables tree topologies — operators can deploy a relay fullnode that pulls from committee members and serves a fleet of downstream fullnodes.

**EVM tx writes.** `eth_sendRawTransaction` is sender-shard-mapped via `Committee.EvmShard` and forwarded over HTTP to the shard owner. Every committee member must expose an `evmrpc` URL (required field on `AutobahnFileConfig`, validated at parse time). Validators short-circuit on their own shard to local mempool; fullnodes always forward.

**`LastCommittedBlockNumber` reads `app.LastBlockHeight()` directly.** New method on the ABCI `Application` interface for a fast in-memory read of the app's executed height (`Info()` is too heavy for the hot path). Both validator and fullnode return the same value, matching CometBFT `/status` — clients querying receipts at the reported height never see a height the app hasn't reached. Single source of truth (the app), no parallel cache.

**Run() lifecycle.** GigaRouter is constructed in `node/setup.go` (`buildGigaRouter`) and spawned by `node.go`'s `OnStart` via `SpawnCritical`, alongside the transport. `Router.Run` does not spawn `giga.Run` — the rule is that whoever constructs an object owns calling its `Run`.

**Block-sync-only Service.** `giga.Service.state utils.Option[*consensus.State]`. `NewService` constructs the full validator service; `NewBlockSyncService` constructs the fullnode-side (None). Consensus / avail handlers reach state through a `validatorState()` helper that `OrPanic`s on the (structurally impossible) None branch.

**Config.** `autobahn-config-file` is a top-level TOML key (placed above any `[section]` header so viper sees it at root scope where mapstructure expects). All other autobahn knobs live in the JSON file it points to. `PersistentStateDir` is rootified against the node's `--home` dir, matching how cometbft handles relative paths elsewhere.

**Integration test.** `make autobahn-integration-test` boots a 4-validator autobahn docker cluster + an RPC node sidecar (`mode = "full"`) and routes every RPC read through the fullnode. Subtests cover block production, bank transfers, liveness under f faults, halt beyond f faults, and recovery; polling helpers (`waitForStableHeight`, `waitForHeightAdvance`) absorb fullnode drain + failover delays.

**Supporting changes.** `step1_configure_init.sh` learns an `AUTOBAHN=true` path (polls validator dirs, runs `gen-autobahn-config`, injects `autobahn-config-file` into `config.toml`; `seid init --overwrite`; explicit `if [ ! -f ]; exit 1` instead of `set -e` so transient curl probes don't kill init). `Makefile` adopts `$(DOCKER_PLATFORM)` (unblocks native-ARM CI) and propagates `AUTOBAHN` + `CLUSTER_SIZE`; `integration-test.yml` adds the `autobahn-integration-tests` job and switches the matrix to GHCR-pulled images.

Note: parallel multi-upstream block-sync + peer discovery (full mesh) is deferred to a follow-up PR — fullnodes here are single-active outbound + open inbound. That's enough for tree topologies (operator-configured relay fullnodes).

## Test plan

- [x] `gofmt -s` + `go vet` clean
- [x] `golangci-lint run` clean
- [x] `go test ./sei-tendermint/internal/p2p/... ./sei-tendermint/config/... ./sei-tendermint/node/... ./sei-tendermint/internal/proxy/...` pass
- [x] `make autobahn-integration-test`: the whole client-facing EVM flow (balance / chainId / nonce / send / receipt) runs through a fullnode sidecar that catches up via giga and forwards writes; halt/liveness/permanent-fault scenarios verified by polling helpers against the fullnode's CometBFT RPC.
- [x] CI green


codchen (approved):

app/app.go
```diff
@@ -480,10 +480,21 @@ type App struct {
 
 	forkInitializer func(sdk.Context)
 
-	httpServerStartSignal     chan struct{}
-	wsServerStartSignal       chan struct{}
-	httpServerStartSignalSent bool
-	wsServerStartSignalSent   bool
+	httpServerStartSignal chan struct{}
+	wsServerStartSignal   chan struct{}
```
pompon0: I wonder if these really are intended to be independent, or we can have just serverStartSignal channel that is closed via evmRPCReadyOnce. Optionally, note that you can wrap evmRPCReadyOnce and serverStartSignal into one sei-tendermint/libs/utils.Once object.

sei-tendermint/config/config.go
```diff
@@ -78,6 +78,27 @@ type Config struct {
...
+
+	// AutobahnRole selects how the node participates in Autobahn. Valid values:
+	//   "" (default) / "validator" — full participant; the node's validator key
+	//                                must be present in the committee.
+	//   "rpc-only"                 — non-validator RPC node; loads the
+	//                                committee and forwards
+	//                                eth_sendRawTransaction to the shard
+	//                                owner. No consensus participation.
+	// Ignored when AutobahnConfigFile is empty.
+	AutobahnRole string `mapstructure:"autobahn-role"`
```
pompon0: nit: shouldn't this be conditioned on the presence of the validator key instead?

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -82,6 +91,24 @@ func NewGigaRouter(cfg *GigaRouterConfig, key NodeSecretKey) (*GigaRouter, error
...
+		// side lands (eth_getTransactionReceipt / eth_call / log queries on
+		// the rpc node) we'll likely need to bring back the data layer here
+		// so the rpc node can subscribe to finalized blocks and serve the
+		// CometBFT block/state read APIs — but still without consensus.Key,
+		// producer.State, or the giga.Service inbound peer-side. Revisit
+		// this branch then and pull in only the pieces that are actually
+		// needed; leaving them all out for now keeps the write-only milestone
+		// minimal and the read-path requirements explicit.
+		logger.Info("GigaRouter initialized (rpc-only)", "validators", len(cfg.ValidatorAddrs))
+		return &GigaRouter{
```
pompon0: if GigaRouter is supposed to support both modes, all the validator-only fields should be put into an utils.Option section of the GigaRouter struct.

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -143,6 +219,12 @@ func (r *GigaRouter) LastCommittedBlockNumber() int64 {
 // FinalizeBlock responses are not stored on disk) without reaching into
 // the unexported router.cfg.
 func (r *GigaRouter) MaxGasPerBlock() int64 {
+	if r.cfg.RPCOnly {
```
pompon0: the pattern of returning dummy values depending on the mode is not type-safe. Neither is returning errors in case the mode is incompatible. For example, instead, you can  create a wrapper type sth like GigaValidatorRouter{}, and a cast method: GigaRouter.AsValidator() Option[GigaValidatorRouter], and move all validator-only methods there. This is just an example, there are probably other safe patters available as well. The general idea is to conditionally provide a wider role-specific interface, instead of custom mismatch handling in each method.

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -387,6 +475,16 @@ func (r *GigaRouter) runExecute(ctx context.Context) error {
...
 func (r *GigaRouter) Run(ctx context.Context) error {
+	if r.cfg.RPCOnly {
+		// Write forwarding happens over fresh HTTP from evmrpc per request
+		// (see evmrpc.SendRawTransaction → env.EvmProxy), so no spawn loops
+		// are needed yet — block until shutdown. App init (InitChain) is
+		// handled by InitRPCOnly, called synchronously from node startup
+		// before RPC begins serving. See the TODO(autobahn-read-path) in
+		// NewGigaRouter for the loops the read side will pull back in.
+		<-ctx.Done()
+		return ctx.Err()
```
pompon0: "return nil" would be totally fine. There is no obligation for Run to block if it has nothing to do.

pompon0:
I might be misunderstanding the goal here, but if the point is to support just RPC forwarding endpoint, then it should not create GigaRouter (and App?) at all. If this is a stop gap measure until we support full-nodes - nodes actually processing blocks, then I'd like to understand why we cannot implement it rn (i.e. skip the stop gap).

app/app.go
wen-coding: We don't need to touch app.go now we are doing read side.

sei-tendermint/config/config.go
wen-coding: done

sei-tendermint/internal/p2p/giga_router.go

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -448,30 +640,51 @@ func (r *GigaRouter) RunInboundConn(ctx context.Context, hConn *handshakedConn)
...
 	}
-	if !ok {
-		return fmt.Errorf("peer not whitelisted")
+	if !isCommittee {
+		select {
+		case v.inboundRPCOnlyPermits <- struct{}{}:
+			defer func() { <-v.inboundRPCOnlyPermits }()
+		default:
+			return fmt.Errorf("inbound rpc-only peer limit (%d) reached", v.inboundRPCOnlyCap)
+		}
```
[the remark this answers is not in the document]
wen-coding: Maintaining whitelist manually is too much pain, added TODO to add a few whitelist later.

sei-tendermint/internal/p2p/giga/service.go
```diff
@@ -4,17 +4,45 @@ import (
...
+//
+// All block-sync methods read/write s.data directly; consensus/avail methods
+// dereference s.state, which is nil in block-sync-only mode (callers must not
+// reach those methods on an rpc-only service).
 type Service struct {
 	getBlockReqs chan req
-	state        *consensus.State
+	data         *data.State
+	// state is the validator-mode consensus state. nil in block-sync-only mode.
+	state *consensus.State
```
pompon0: Option

sei-tendermint/internal/p2p/giga_router_rpc_only.go
```diff
@@ -0,0 +1,165 @@
...
+	"github.com/sei-protocol/sei-chain/sei-tendermint/libs/utils"
+	"github.com/sei-protocol/sei-chain/sei-tendermint/libs/utils/scope"
+)
+
+// gigaRPCOnlyRouter is the GigaRouter impl for non-validator nodes. It
+// pulls finalized blocks from committee members via the giga block-sync
+// subscriber and executes them locally; it does not run consensus or
+// producer, does not accept inbound giga connections, and sources its
+// "last committed block" from data.State directly.
+type gigaRPCOnlyRouter struct {
```
pompon0: can we stick to fullnode name instead of RPConly? This is the naming in tendermint and afaik also the industry standard (I might be wrong)

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -42,20 +42,90 @@ type GigaRouterConfig struct {
 	DialInterval   time.Duration
 	ValidatorAddrs map[atypes.PublicKey]GigaNodeAddr
 	Consensus      *consensus.Config
-	Producer       *producer.Config
-	TxMempool      *mempool.TxMempool
-	GenDoc         *types.GenesisDoc
+	// Producer is only set on validator paths; rpc-only nodes don't
+	// produce blocks and source MaxGasPerBlock from genesis instead.
+	Producer  utils.Option[*producer.Config]
+	TxMempool *mempool.TxMempool
```
pompon0: fullnodes are not supposed to have mempool

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -42,20 +42,90 @@ type GigaRouterConfig struct {
 	DialInterval   time.Duration
 	ValidatorAddrs map[atypes.PublicKey]GigaNodeAddr
 	Consensus      *consensus.Config
```
pompon0: fullnodes should not need consensus config (they need committee, but that's currently derived from ValidatorAddrs anyway).

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -101,35 +179,74 @@ func NewGigaRouter(cfg *GigaRouterConfig, key NodeSecretKey) (*GigaRouter, error
 	if err != nil {
 		return nil, fmt.Errorf("data.NewState(): %w", err)
 	}
+	if cfg.RPCOnly {
+		// Every committee member must expose an EVMRPC URL: rpc-only nodes
```
pompon0: I think we should require all urls in both validator and fullnode modes by now. I have added a fallback to local mempool just for the sake of easier tests, but with fullnode in the picture it won't help with anything anymore.

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -101,35 +179,74 @@ func NewGigaRouter(cfg *GigaRouterConfig, key NodeSecretKey) (*GigaRouter, error
...
-		data:      dataState,
-		consensus: consensusState,
-		producer:  producerState,
-		service:   giga.NewService(consensusState),
-		poolIn:    giga.NewPool[NodePublicKey, rpc.Server[giga.API]](),
-		poolOut:   giga.NewPool[NodePublicKey, rpc.Client[giga.API]](),
-
-		// Subscribe once here (takes avail's internal lock once); subsequent
-		// Load() calls from RPC handlers are lock-free atomic pointer reads.
+	producerConfig := cfg.Producer.OrPanic("validator-mode requires GigaRouterConfig.Producer")
```
pompon0: return error?

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -42,20 +42,90 @@ type GigaRouterConfig struct {
...
+// embed *gigaRouterCommon and inherit the read-path / execute-loop logic
+// from it; mode-specific behaviour (Run, RunInboundConn, EvmProxy,
+// LastCommittedBlockNumber, MaxGasPerBlock, dialAndRunConn) is implemented
+// on the embedding types.
+type gigaRouterCommon struct {
+	cfg     *GigaRouterConfig
+	key     NodeSecretKey
+	data    *data.State
+	service *giga.Service
+	poolOut *giga.Pool[NodePublicKey, rpc.Client[giga.API]]
```
pompon0: non-validator connections are supposed to create a mesh therefore:
* non-validator nodes should also have both inbound and outbound connections
* blocksyncing should be allowed in both directions by default (that's why I didn't want to restrict sending blocks from non-validators -> validators for now, we can add this as a separate feature).

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -42,20 +42,90 @@ type GigaRouterConfig struct {
...
+// runs consensus + producer, accepts inbound giga connections (full
+// service for committee peers, block-sync subset for rpc-only peers), and
+// dials every committee member in parallel for vote fan-out.
+type gigaValidatorRouter struct {
+	*gigaRouterCommon
+
+	consensus      *consensus.State
+	producer       *producer.State
+	producerConfig *producer.Config
+	poolIn         *giga.Pool[NodePublicKey, rpc.Server[giga.API]]
```
pompon0: fyi, the giga.Pool concept was supposed to support attaching custom tasks within the lifetime of each connection, but it does complicate stuff and the fancy features are unused by now, so soon I think I'll simplify it, unless a use case arises somewhere.

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -42,20 +42,90 @@ type GigaRouterConfig struct {
...
+
+	// inboundRPCOnlyPermits is a buffered-channel semaphore capping
+	// concurrent non-committee inbound block-sync connections.
+	// Non-blocking acquire (select + default) so excess peers get a clean
+	// rejection at handshake time rather than queuing.
+	inboundRPCOnlyPermits chan struct{}
```
pompon0: https://pkg.go.dev/golang.org/x/sync/semaphore ?

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -101,35 +179,74 @@ func NewGigaRouter(cfg *GigaRouterConfig, key NodeSecretKey) (*GigaRouter, error
...
-		producer:  producerState,
-		service:   giga.NewService(consensusState),
-		poolIn:    giga.NewPool[NodePublicKey, rpc.Server[giga.API]](),
-		poolOut:   giga.NewPool[NodePublicKey, rpc.Client[giga.API]](),
-
-		// Subscribe once here (takes avail's internal lock once); subsequent
-		// Load() calls from RPC handlers are lock-free atomic pointer reads.
+	producerConfig := cfg.Producer.OrPanic("validator-mode requires GigaRouterConfig.Producer")
+	producerState := producer.NewState(producerConfig, cfg.TxMempool, consensusState)
+	// None → built-in default; Some(0) → reject all (channel cap 0); Some(n>0) → operator override.
```
pompon0: nit: let's push this default outside of p2p to config/config.go and require a concrete number in giga_router.go

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -70,10 +140,17 @@ type GigaRouter struct {
 	lastCommitQCRecv utils.AtomicRecv[utils.Option[*atypes.CommitQC]]
 }
 
-func NewGigaRouter(cfg *GigaRouterConfig, key NodeSecretKey) (*GigaRouter, error) {
+func NewGigaRouter(cfg *GigaRouterConfig, key NodeSecretKey) (GigaRouter, error) {
```
pompon0: how about we have separate constructors and separate config types? I suspect it may make the code less cluttered.

pompon0: gigaFullNodeConfig may be embedded in gigaValidatorConfig, for example.

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -386,30 +504,52 @@ func (r *GigaRouter) runExecute(ctx context.Context) error {
...
 	})
 }
 
-func (r *GigaRouter) dialAndRunConn(ctx context.Context, key NodePublicKey, hp tcp.HostPort) error {
+// spawnReadPath spawns the three goroutines that both validator and
+// rpc-only routers run: the data layer, the executeBlock loop, and the
+// giga service (block fetcher). Mode-specific spawns (dial loop +
+// consensus/producer on the validator path; the single-active subscriber
+// on the rpc-only path) live at each Run call site.
+func (r *gigaRouterCommon) spawnReadPath(ctx context.Context, s scope.Scope) {
```
pompon0: nit: scope by design is not supposed to be passed to subroutines if avoidable, because ctx is strictly bound to scope, while it is not possible to enforce this relation in the function signature. In this case it should be perfectly fine to implement runReadPath() which uses scope internally.

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -438,39 +578,52 @@ func (r *GigaRouter) dialAndRunConn(ctx context.Context, key NodePublicKey, hp t
...
 	})
 }
 
-func (r *GigaRouter) RunInboundConn(ctx context.Context, hConn *handshakedConn) error {
+func (r *gigaValidatorRouter) RunInboundConn(ctx context.Context, hConn *handshakedConn) error {
 	if !hConn.msg.SeiGigaConnection {
 		return fmt.Errorf("not a SeiGiga connection")
 	}
-	// Filter unwanded connections.
 	key := hConn.msg.NodeAuth.Key()
```
pompon0: make Key optional in handshake.

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -438,39 +578,52 @@ func (r *GigaRouter) dialAndRunConn(ctx context.Context, key NodePublicKey, hp t
...
-			ok = true
+			isCommittee = true
 			break
 		}
 	}
-	if !ok {
-		return fmt.Errorf("peer not whitelisted")
+	if !isCommittee {
+		select {
+		case r.inboundRPCOnlyPermits <- struct{}{}:
```
pompon0: nit: here you don't even need a blocking semaphore, just atomic update on pool size would be good enough.

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -438,39 +578,52 @@ func (r *GigaRouter) dialAndRunConn(ctx context.Context, key NodePublicKey, hp t
...
+			if isCommittee {
+				return r.service.RunServer(ctx, server)
+			}
+			return r.service.RunBlockSyncServer(ctx, server)
 		})
 	})
 }
 
-func (r *GigaRouter) EvmProxy(sender common.Address) (*url.URL, bool) {
+func (r *gigaValidatorRouter) EvmProxy(sender common.Address) (*url.URL, bool) {
```
pompon0: shouldn't this be common?

sei-tendermint/internal/p2p/giga_router_rpc_only_test.go
```diff
@@ -0,0 +1,121 @@
...
+		GenDoc:         genDoc,
+		RPCOnly:        true,
+	}, makeKey(rng))
+	require.NoError(t, err)
+
+	// Shape: rpc-only routers run data + a block-sync service (outbound only)
+	// to pull finalized blocks from committee members. NewGigaRouter returns
+	// the rpc-only concrete impl through the interface; assert on the
+	// concrete type so we can inspect the shared common fields.
+	require.True(t, routerIface.IsRPCOnly())
```
pompon0: if the only purpose of IsRPCOnly is to guard type cast, then it is not useful.

sei-tendermint/internal/p2p/giga_router_fullnode_test.go
```diff
@@ -0,0 +1,121 @@
...
+	require.NoError(t, err)
+
+	// Shape: rpc-only routers run data + a block-sync service (outbound only)
+	// to pull finalized blocks from committee members. NewGigaRouter returns
+	// the rpc-only concrete impl through the interface; assert on the
+	// concrete type so we can inspect the shared common fields.
+	require.True(t, routerIface.IsRPCOnly())
+	router, ok := routerIface.(*gigaRPCOnlyRouter)
+	require.True(t, ok, "rpc-only NewGigaRouter should return *gigaRPCOnlyRouter")
+	require.NotNil(t, router.data)
```
pompon0: do we really need those non-nil checks?

sei-tendermint/internal/p2p/giga_router.go
wen-coding: That's right, it would be more like a tree, but I agree that they do need both inbound and outbound connections, I just didn't want to do that in this one PR, because this PR would be too big. But maybe we should keep poolIn and poolOut.

sei-tendermint/internal/p2p/giga/service.go
wen-coding: Already did that, but then everywhere I need to do OrPanic(), it's a bit ugly, so I reverted it: https://github.com/sei-protocol/sei-chain/pull/3525/changes/542a89ca47b0b47a766dedd79682a6582a1f6cb0

We won't spawn s if state is nil, changing it to Option still doesn't catch the nil case at compile time.

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -367,30 +457,44 @@ func (r *GigaRouter) runExecute(ctx context.Context) error {
...
-					err := r.dialAndRunConn(ctx, addr.Key, addr.HostPort)
+					err := r.dialAndRunConn(ctx, addr.Key, addr.HostPort, r.service.RunClient)
 					logger.Info("giga connection failed", "addr", addr, "err", err)
 					if err := utils.Sleep(ctx, r.cfg.DialInterval); err != nil {
 						return err
 					}
 				}
 			})
 		}
-		s.SpawnNamed("data", func() error { return r.data.Run(ctx) })
```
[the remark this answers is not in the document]
wen-coding: PushAppHash should only fail on shutdown, so this is fine

sei-tendermint/internal/p2p/giga/service.go
```diff
@@ -4,17 +4,37 @@ import (
...
+		state:        state,
+	}
+}
+
+// NewBlockSyncService constructs a Service that only serves and consumes
+// block-sync streams (no consensus / avail).
+func NewBlockSyncService(d *data.State) *Service {
+	return &Service{
+		getBlockReqs: make(chan req),
+		data:         d,
```
pompon0, line 38: this leaves state = nil, either split Service types, or make state Option[*consensus.State]

sei-tendermint/internal/p2p/giga_router_fullnode.go
```diff
@@ -0,0 +1,162 @@
...
+func (r *gigaFullnodeRouter) AsValidator() utils.Option[GigaValidatorRouter] {
+	return utils.None[GigaValidatorRouter]()
+}
+
+// EvmProxy on a fullnode always forwards — no validator key, no local
+// mempool, no self-shard short-circuit. buildDataState
+// rejects configs missing any URL, so .Get() never silent-drops in production.
+func (r *gigaFullnodeRouter) EvmProxy(sender common.Address) (*url.URL, bool) {
+	shardValidator := r.data.Committee().EvmShard(sender)
+	return r.cfg.ValidatorAddrs[shardValidator].EVMRPC.Get()
```
pompon0: as I mentioned before, can we make EVMRPC required by now?

sei-tendermint/internal/p2p/giga_router_fullnode.go
```diff
@@ -0,0 +1,162 @@
...
+const fullnodeMaxBackoff = 5 * time.Minute
+
+// fullnodeStallTimeout: close the connection if data.NextBlock has not
+// advanced for this long. Bounds the time a single peer can hold the loop
+// open without delivering blocks.
+const fullnodeStallTimeout = 60 * time.Second
+
+// fullnodeProgressPollInterval: how often watchProgress samples
+// data.NextBlock.
+const fullnodeProgressPollInterval = 5 * time.Second
```
pompon0: you can just await the next block instead. No need for polling.

sei-tendermint/internal/p2p/giga_router_fullnode.go
```diff
@@ -0,0 +1,162 @@
...
+		s.Spawn(func() error { return r.runFullnodeSubscriber(ctx) })
+		s.SpawnNamed("data", func() error { return r.data.Run(ctx) })
+		s.SpawnNamed("execute", func() error { return r.runExecute(ctx) })
+		s.SpawnNamed("service", func() error { return r.service.Run(ctx) })
+		return nil
+	})
+}
+
+// fullnodeHealthyMinAge: minimum time-since-last-healthy required before
+// the inter-cycle backoff escalates. Guards against a brief post-reconnect
```
pompon0: nit: move these constants inside functions that are actually using them, since they are not used anywhere else anyway.

sei-tendermint/internal/p2p/giga_router_fullnode.go
```diff
@@ -0,0 +1,162 @@
...
+// failure burst doubling the backoff before the next progress arrives.
+const fullnodeHealthyMinAge = 30 * time.Second
+
+// fullnodeMaxBackoff caps the inter-cycle backoff so a persistently-
```
pompon0: fullnode not even trying to connect for 5m is not very useful imo. Can we make it sth like 10s instead? It will round robin over nodes anyway which will limit per-validator impact.

pompon0: Or just stick with DialInterval without backoff. DialInterval is expected to be fine when reconnecting to the same peer, so it should be fine with a round robin over multiple peers as well

sei-tendermint/internal/p2p/giga_router_test.go
```diff
@@ -194,9 +193,16 @@ type testNodeCfg struct {
 }
 
 func (c *testNodeCfg) GigaNodeAddr() GigaNodeAddr {
+	// EVMRPC must be present for NewGigaRouter to accept the config on
+	// either path; the URL value is unused by the tests in this file.
+	u, err := url.Parse(fmt.Sprintf("http://%s:8545", c.addr.Addr().String()))
```
pompon0: nit: utils.OrPanic1

sei-tendermint/internal/p2p/giga_router_test.go
```diff
@@ -332,9 +343,10 @@ func TestGigaRouter_FinalizeBlocks(t *testing.T) {
 			}
 			s.SpawnNamed(fmt.Sprintf("producer[%v]", i), func() error {
 				giga := router.Giga().OrPanic("non-giga router")
+				v := giga.AsValidator().OrPanic("validator giga must satisfy GigaValidatorRouter")
```
pompon0: nit: why not just expose GigaValidatorRouter and GigaFullNodeRouter and use type cast from GigaRouter directly?

pompon0: OR just have `Mempool() Option[*Mempool]` instead of 'AsValidator`?

sei-tendermint/internal/p2p/giga_router.go
pompon0: I mean, yes, we don't want to merge poolIn and poolOut. We just want them both for validators and fullnodes

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -37,41 +39,110 @@ func (a GigaNodeAddr) String() string {
...
 }
 
-func NewGigaRouter(cfg *GigaRouterConfig, key NodeSecretKey) (*GigaRouter, error) {
+// buildDataState validates the common config and constructs the data
+// layer (committee, WAL, State) shared by both giga constructors.
+//
+// TODO(autobahn): once sei-db/ledger_db/block.BlockDB has a writer wired
+// (see BlockByNumber's TODO), the data WAL is redundant.
+func buildDataState(cfg *GigaRouterCommonConfig) (*data.State, error) {
+	if cfg.GenDoc == nil {
```
pompon0: nit: can we just skip the nil checks as usual and panic on nil deref?

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -37,41 +39,110 @@ func (a GigaNodeAddr) String() string {
...
+	if cfg.App == nil {
+		return nil, fmt.Errorf("GigaRouterCommonConfig.App must be set")
+	}
+	// Guards runFullnodeSubscriber's `(i + 1) % len(addrs)` against
+	// direct construction that bypasses AutobahnFileConfig.Validate.
+	if len(cfg.ValidatorAddrs) == 0 {
+		return nil, fmt.Errorf("GigaRouterCommonConfig.ValidatorAddrs is empty; need at least one committee member")
+	}
+	for vk, addr := range cfg.ValidatorAddrs {
+		if !addr.EVMRPC.IsPresent() {
```
pompon0: ok, if we already have this check, then just make EVMRPC field not optional

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -80,73 +151,102 @@ func NewGigaRouter(cfg *GigaRouterConfig, key NodeSecretKey) (*GigaRouter, error
...
+
+// NewGigaValidatorRouter returns the concrete validator type (not the
+// interface) so router.go can reach RunInboundConn in-package without a
+// runtime downcast.
+func NewGigaValidatorRouter(cfg *GigaValidatorConfig, key NodeSecretKey) (*gigaValidatorRouter, error) {
+	dataState, err := buildDataState(&cfg.GigaRouterCommonConfig)
+	if err != nil {
+		return nil, err
+	}
+	if cfg.Producer == nil {
```
pompon0: ditto

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -291,20 +391,38 @@ func (r *GigaRouter) executeBlock(ctx context.Context, b *atypes.GlobalBlock) (*
...
-		return nil, fmt.Errorf("r.cfg.App.FinalizeBlock(): %w", err)
+		return nil, fmt.Errorf("app.FinalizeBlock(): %w", err)
 	}
 	commitResp, err := app.Commit(ctx)
 	if err != nil {
-		return nil, fmt.Errorf("r.cfg.App.Commit(): %w", err)
+		return nil, fmt.Errorf("app.Commit(): %w", err)
 	}
+	// Publish right after Commit (not in the caller loop) so readers of
+	// LastCommittedBlockNumber don't briefly see app.LastBlockHeight ahead
```
pompon0: We don't have such a guarantee, because Commit and lastExecutedBlock.Store is not atomic. Do we really need such a guaratee? If so, can we just remove lastExecutedBlock and access app.LastBlockHeight instead?

pompon0: we can add a dedicated function to Application interface to make it efficient, if needed.

sei-tendermint/internal/p2p/giga_router.go
pompon0: ping

sei-tendermint/internal/p2p/giga_router_fullnode.go
pompon0: also this should be a method of Common

sei-tendermint/internal/p2p/router.go
```diff
@@ -235,7 +232,16 @@ func (r *Router) acceptPeersRoutine(ctx context.Context) error {
 					}
 					if giga, ok := r.giga.Get(); ok && hConn.msg.SeiGigaConnection {
 						release()
-						return giga.RunInboundConn(ctx, hConn)
+						// Only validators accept inbound giga connections; fullnodes
```
pompon0: nit: make RunInboundConn a method of GigaRouter interface, so that Router does not have to care.

sei-tendermint/node/setup.go
```diff
@@ -233,33 +247,141 @@ func buildGigaConfig(
...
+}
+
+// buildAndStartGigaRouter picks validator-vs-fullnode by cfg.Mode:
+// "validator" runs the validator path, any other mode runs as a fullnode.
+// Mode is the operator's explicit role declaration, kept separate from
+// committee membership so a newly-joined committee member can finish
+// catch-up as a fullnode before the operator flips to mode = "validator".
+// A warning is logged if mode and committee membership disagree so an
+// operator misconfiguration is visible at startup.
+func buildAndStartGigaRouter(
```
pompon0: it does not start the router though?

sei-tendermint/internal/p2p/router.go
```diff
@@ -102,13 +102,10 @@ func NewRouter(
...
-		gr, err := NewGigaRouter(gigaCfg, privKey)
-		if err != nil {
-			return nil, fmt.Errorf("NewGigaRouter(): %w", err)
-		}
-		router.giga = utils.Some(gr)
-	}
+	// The GigaRouter is constructed by setup-side code (which picks the
+	// validator vs fullnode constructor based on whether a local validator
+	// key is present) and passed in via options.Giga. Just attach.
+	router.giga = options.Giga
```
pompon0, line 108: the rule is that one who constructed an object is responsible for calling Run() of that object, so that it can control its whole lifecycle (since we cannot rely on the non-existent golang object ownership model). Therefore if we are passing a constructed GigaRouter to NewRouter, then Router.Run is NOT responsible for spawning GigaRouter.Run.

Perhaps we should codify this rule somewhere or ALTERNATIVELY we can make it more explicit as follows (I worked with this before, it is more verbose but harder to ignore):
* instead of having type A having Run() method, we make the constructor NewA() return (*A, run func(ctx) error). This way the Run task is separate from A, and whoever holds it is responsible for running it (instead of implicit responsibility)
* this will make things ugly at places, but still better than forgetting to spawn Run/spawn it in multiple times

What would be a bad idea (based on my experiences) would be to do sth like NewA(s scope.Scope) (*A,error), which would internally s.Spawn(func(){ a.Run() }), because it makes it infeasible to handle the Run() errors gracefully.

sei-tendermint/internal/p2p/giga_router.go
wen-coding: changed

sei-tendermint/internal/p2p/giga_router.go

sei-tendermint/internal/p2p/giga/service.go
wen-coding: done

sei-tendermint/internal/p2p/giga_router_fullnode.go
wen-coding: done

sei-tendermint/internal/p2p/giga_router_fullnode.go
wen-coding: done

sei-tendermint/internal/p2p/giga_router_fullnode.go
wen-coding: done

sei-tendermint/internal/p2p/giga_router_fullnode.go
wen-coding: backoff removed

sei-tendermint/internal/p2p/giga_router_test.go
wen-coding: done

sei-tendermint/internal/p2p/giga_router_test.go
wen-coding: using Mempool() now

sei-tendermint/internal/p2p/giga_router.go

sei-tendermint/internal/p2p/giga_router.go

sei-tendermint/internal/p2p/giga_router.go

sei-tendermint/internal/p2p/giga_router.go
wen-coding: added LastBlockHeight(), dropped lastExecutedBlock and seedLastExecuted

sei-tendermint/internal/p2p/router.go
wen-coding: done

sei-tendermint/node/setup.go
wen-coding: renamed to buildGigaRouter

sei-tendermint/internal/p2p/router.go
wen-coding: done

sei-tendermint/internal/p2p/giga_router.go

sei-tendermint/internal/p2p/giga_router.go
wen-coding: done, now fullnode can relay blocks. The actual peer discovery can be implemented after the design doc has been reviewed.

sei-tendermint/internal/p2p/giga_router_fullnode.go
```diff
@@ -0,0 +1,67 @@
...
+	addrs := slices.Collect(maps.Values(r.cfg.ValidatorAddrs))
+	rand.Shuffle(len(addrs), func(i, j int) { addrs[i], addrs[j] = addrs[j], addrs[i] })
+	for i := 0; ; i = (i + 1) % len(addrs) {
+		addr := addrs[i]
+		err := r.dialAndRunConn(ctx, utils.None[NodePublicKey](), addr.HostPort, r.service.RunBlockSyncClient)
+		logger.Info("fullnode giga connection ended; failing over", "addr", addr, "err", err)
+		if err := utils.Sleep(ctx, r.cfg.DialInterval); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
wen-coding: This is deliberate choice

sei-tendermint/internal/p2p/giga_router.go
```diff
@@ -80,73 +127,94 @@ func NewGigaRouter(cfg *GigaRouterConfig, key NodeSecretKey) (*GigaRouter, error
...
+		validatorKey:   cfg.ValidatorKey.Public(),
+	}, nil
 }
 
-// Mempool exposes Autobahn's producer-backed mempool surface to callers that
-// need features not shared with CometBFT's TxMempool.
-func (r *GigaRouter) Mempool() *producer.State {
-	return r.producer
+func (r *gigaRouterCommon) LastCommittedBlockNumber() int64 {
+	return r.app.LastBlockHeight()
```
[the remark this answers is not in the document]
wen-coding: This is intentional

sei-tendermint/config/config.go
```diff
@@ -76,9 +76,32 @@ type Config struct {
...
-}
+
+	// AutobahnMaxInboundFullnodePeers caps concurrent inbound block-sync
+	// connections from non-committee peers. Applied on both validators and
+	// fullnodes (relay fullnodes serving downstream block-sync are subject
+	// to the same cap). Pointer so we can tell absent ("use
+	// DefaultAutobahnMaxInboundFullnodePeers") from explicit 0 ("reject
+	// all inbound block-sync from non-committee peers"). Positive values
+	// override the default.
+	AutobahnMaxInboundFullnodePeers *int `mapstructure:"autobahn-max-inbound-fullnode-peers"`
```
pompon0: nit: shouldn't it be a field in AutobahnConfigFile?

sei-tendermint/internal/p2p/router_test.go
```diff
@@ -315,43 +315,53 @@ func TestRouter_GigaSetWhenConfigured(t *testing.T) {
...
+			GenDoc: &types.GenesisDoc{
+				ChainID:       "giga-e2e-test",
+				InitialHeight: 42,
+				GenesisTime:   time.Now(),
+			},
 		},
+		ValidatorKey: valKey,
+		ViewTimeout:  func(atypes.View) time.Duration { return 3 * time.Second },
 		Producer: &producer.Config{
-			App:                     proxyApp,
```
pompon0, line 337: nit: how about you remove App from producer.Config already in this pr?

sei-tendermint/internal/p2p/router_test.go
```diff
@@ -315,43 +315,53 @@ func TestRouter_GigaSetWhenConfigured(t *testing.T) {
...
-	})
+	}
+	gigaRouter, err := NewGigaValidatorRouter(gigaCfg, nodeKey)
+	require.NoError(t, err)
+	opts.Giga = utils.Some[GigaRouter](gigaRouter)
 
 	router := makeRouterWithOptionsAndKey(opts, nodeKey)
 	require.True(t, router.giga.IsPresent(), "GigaRouter should be set when Giga config is provided")
 
-	giga, _ := router.giga.Get()
```
pompon0, line 354: nit: these are assertions about config of GigaRouter, which doesn't seem useful, because it is constructed by hand a few lines above

pompon0: this casting of router.Giga() back to retrieve the gigaRouter we have constructed just above doesn't seem useful either.

sei-tendermint/node/node.go
```diff
@@ -521,6 +524,12 @@ func (n *nodeImpl) OnStart(ctx context.Context) error {
 	if m, ok := n.mempool.Get(); ok {
 		n.SpawnCritical("mempool", m.Run)
 	}
+	// Run the GigaRouter alongside the transport. Constructed in createRouter
+	// and reached through n.router.Giga(); its lifecycle is owned here, not
+	// by the Router.
+	if giga, ok := n.router.Giga().Get(); ok {
```
pompon0: nit: I'd recommend to keep a reference to the gigaRouter in the nodeImpl directly, so that ownership is more explicit.

sei-tendermint/internal/p2p/giga_router_fullnode.go
```diff
@@ -0,0 +1,86 @@
...
+			poolIn:             giga.NewPool[NodePublicKey, rpc.Server[giga.API]](),
+			poolOut:            giga.NewPool[NodePublicKey, rpc.Client[giga.API]](),
+			app:                cfg.App,
+			inboundFullnodeCap: int64(cfg.MaxInboundFullnodePeers),
+		},
+	}, nil
+}
+
+func (r *gigaFullnodeRouter) MaxGasEstimatedPerBlock() uint64 {
+	return r.cfg.GenDoc.ConsensusParams.Block.MaxGasUint64()
```
pompon0: nit: I think we can move this method to common, since both fullnode and validator return the same value, just stored in a different place. Later it will be a property of the Epoch, which also will be available to both Validators and FullNodes.

pompon0 (approved):

sei-tendermint/config/config.go
wen-coding: changed

sei-tendermint/internal/p2p/router_test.go
wen-coding: done

sei-tendermint/internal/p2p/router_test.go
wen-coding: removed

sei-tendermint/node/node.go
wen-coding: done

sei-tendermint/internal/p2p/giga_router_fullnode.go
wen-coding: done

sei-tendermint/internal/p2p/giga/data.go
```diff
@@ -86,7 +86,7 @@ func (s *Service) clientGetBlock(ctx context.Context, client rpc.Client[API]) er
 				if !ok {
 					return nil
```
philipsu522, line 96: should we be erroring if we don't get a block? what is the behavior if we don't error out here?

sei-tendermint/internal/p2p/giga/data.go
```diff
@@ -111,7 +111,7 @@ func (x *Service) runBlockFetcher(ctx context.Context) error {
 			scope.Spawn(func() error {
 				defer release()
 				for {
-					if _, err := x.state.Data().TryBlock(n); !errors.Is(err, data.ErrNotFound) {
+					if _, err := x.data.TryBlock(n); !errors.Is(err, data.ErrNotFound) {
```
philipsu522, line 123: nit - should we have some sort of back off retry here?

sei-tendermint/node/setup.go
```diff
@@ -233,33 +242,141 @@ func buildGigaConfig(
...
+) (p2p.GigaRouter, error) {
+	_, validatorAddrs, err := loadAutobahnCommittee(cfg.AutobahnConfigFile)
+	if err != nil {
+		return nil, err
+	}
+	if valKey, ok := validatorKey.Get(); ok {
+		_, inCommittee := validatorAddrs[valKey.Public()]
+		switch {
+		case cfg.Mode == config.ModeValidator && !inCommittee:
+			logger.Error("Autobahn: mode is \"validator\" but local validator key is not in the committee", "valKey", valKey.Public())
```
philipsu522: perhaps this should be a warn?  if i set mode as validator but i happen to not be active, i believe it should basically run as a full node. for example, if delegation has not been provided to get into active set yet. i think we also error out in L237 above

sei-tendermint/internal/p2p/giga/data.go
wen-coding: It could be the upstream peer is also catching up so it doesn't have block N yet, erroring here would cancel the whole block-sync.

sei-tendermint/internal/p2p/giga/data.go
wen-coding: done, added 1 second wait between retries

sei-tendermint/node/setup.go
wen-coding: I'll just change it to warn for now. Because currently we don't have a path going from full node to validator node yet, so running a full node is not a good mode to be in here.

We can revisit the decision after Epoch is implemented.

sei-tendermint/internal/p2p/giga_router_common.go
```diff
@@ -0,0 +1,398 @@
...
+	dataState, err := data.NewState(&data.Config{Committee: committee}, dataWAL)
+	if err != nil {
+		return nil, fmt.Errorf("data.NewState(): %w", err)
+	}
+	return dataState, nil
+}
+
+func (r *gigaRouterCommon) LastCommittedBlockNumber() int64 {
+	return r.app.LastBlockHeight()
+}
```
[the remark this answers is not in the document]
wen-coding: This is in a previous discussion. We don't have any guarantee the two are in sync here.

philipsu522 (approved):

PR #3534, opened 2026-06-02
docs: remove empty parens from testnet description

x4v13r64:
Tiny fix: empty trailing parens '()' on the testnet description line.


x4v13r64:
Whoops, dropping this — opened by accident while exploring.

PR #3537, opened 2026-06-03
feat(integration): Go-native runner for YAML integration tests (POC)

amir-deris:
## Summary

- Adds `integration_test/runner/` — a Go-native driver for the existing YAML integration test suite that replaces the Python `runner.py` for covered modules
- Keeps all existing YAML test files unchanged; the runner parses the same schema
- Wires one CI matrix job ("Mint & Staking & Bank Module") to the Go runner as a proof-of-concept, leaving the parallel Autobahn job on Python for side-by-side comparison

## Motivation

The Python runner has a few pain points:
- Failures stop the entire file (`exit(1)`) with no subtest isolation
- No `-run` flag to target a single test case without editing files
- Output is flat `print()` statements with no structured test reporting
- Python dependency in a Go project's CI

The Go runner addresses all of these through standard `go test` integration.

## How it works

Each YAML file maps to a `t.Run` subtest, so failures are isolated and the full `-v` output shows every command and its captured output. The eval verifier uses `big.Int → float64 → string` comparison (in that order) to handle the large token supply integers in the existing YAML assertions without float64 precision loss.

**Run a single module locally** (requires a running cluster):
```
go test -tags yaml_integration -v ./integration_test/runner/... -run TestBankModule
```

**Target a single test case:**
```
go test -tags yaml_integration -v ./integration_test/runner/... -run TestBankModule/Test_sending_funds
```


.github/workflows/integration-test.yml
```diff
@@ -85,9 +85,7 @@ jobs:
           {
             name: "Mint & Staking & Bank Module",
             scripts: [
-              "python3 integration_test/scripts/runner.py integration_test/staking_module/staking_test.yaml",
-              "python3 integration_test/scripts/runner.py integration_test/bank_module/send_funds_test.yaml",
-              "python3 integration_test/scripts/runner.py integration_test/mint_module/mint_test.yaml"
+              "go test -tags yaml_integration -v -timeout 10m ./integration_test/runner/... -run \"TestMintModule|TestStakingModule|TestBankModule/Test_sending_funds\""
```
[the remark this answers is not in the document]
amir-deris: The feedback is not valid. Go's testing package explicitly replaces spaces with underscores in subtest names for the purpose of -run
  matching.

  The chain here is:
  - YAML registers the subtest as "Test sending funds" (from send_funds_test.yaml:1)
  - runner.go:73 calls t.Run(tc.Name, ...) which creates subtest TestBankModule/Test_sending_funds for -run purposes
  - Go testing docs: "Spaces are replaced with underscores in output and the -run and -bench flags"
  
  So -run TestBankModule/Test_sending_funds correctly matches t.Run("Test sending funds", ...). The filter works as intended — the
  feedback is based on a misunderstanding of how Go handles spaces in subtest names.

integration_test/runner/runner.go
```diff
@@ -0,0 +1,334 @@
...
+	}
+
+	out, err := c.Output()
+	stdout := strings.TrimSpace(string(out))
+	if err != nil {
+		if exit, ok := err.(*exec.ExitError); ok {
+			t.Logf("    (exit %d) stderr: %s", exit.ExitCode(), strings.TrimSpace(string(exit.Stderr)))
+			return stdout, nil
+		}
+		return stdout, fmt.Errorf("exec: %w", err)
```
bdchatham: Tiny one: the `%w` here is swallowed. The only caller — `runCase` — logs this with `%v` (`t.Fatalf("input[%d] failed: %v", i, err)`, L91), so the wrapped chain is never unwrapped and the wrap adds only the literal `exec:` prefix. Either drop the wrap (`return stdout, err`) or switch the caller to `%w`/`errors.As` if you actually intend to inspect the chain. Non-blocking.

integration_test/runner/runner.go
```diff
@@ -0,0 +1,334 @@
...
+		c = exec.Command("docker", args...) //nolint:gosec
+	} else {
+		c = exec.Command("/bin/bash", "-c", cmd)
+		c.Env = append(os.Environ(), envMapSlice(envMap)...)
+	}
+
+	out, err := c.Output()
+	stdout := strings.TrimSpace(string(out))
+	if err != nil {
+		if exit, ok := err.(*exec.ExitError); ok {
```
bdchatham: Idiom nit: prefer `errors.As` over a bare type assertion here —
```go
var exit *exec.ExitError
if errors.As(err, &exit) {
```
`c.Output()` returns the `*exec.ExitError` directly today, so the assertion works, but `errors.As` sees through any future `%w` wrapping and is the conventional way to recover a typed error from a chain. Non-blocking.

bdchatham:
**Not a blocker, and explicitly not a regression in this PR** — flagging a pre-existing coverage gap that this work happened to surface, so it's tracked rather than lost.

While reviewing I compared the full suite against what CI actually invokes: **27 YAML files exist, but only 23 are referenced by any CI workflow.** Four are orphaned (run by no matrix row):

- `bank_module/multi_sig_send_test.yaml`
- `bank_module/simulation_tx.yaml`
- `tokenfactory_module/create_tokenfactory_test.yaml`
- `template/template_test.yaml` ← intentional (it's the authoring example, not a real test)

The first three look accidental, not deliberate:
- The two bank files were added in #3522, in the *same* change that heavily edited `integration-test.yml` — only `send_funds_test.yaml` got wired into the matrix; its two siblings didn't.
- Unlike the genuinely environment-gated tests (the upgrade / statesync cases carry explicit "requires a specially prepared cluster" notes), these carry no skip marker and need no special setup — they're just unreferenced.

This PR **faithfully preserves the existing coverage** — the Go `-run` scope reproduces exactly what the Python matrix ran, so nothing is dropped here. If anything the Go runner *improves* visibility: `TestBankModule` already wires all three bank files and `TestTokenFactoryModule` wires the tokenfactory one, so `go test ./...` makes these orphans discoverable in a way the per-row Python CLI strings never did.

**Suggested follow-up (independent of this PR, and applies to both harnesses):** wire these three back into CI — in the Python matrix today, and in the Go `-run` scope as each module migrates — or add an explicit skip marker if any are intentionally held out. Happy to file a tracking issue so it doesn't ride on this POC.

integration_test/runner/runner_test.go
```diff
@@ -0,0 +1,97 @@
...
+	// the CI RPC-node cluster (make run-rpc-node-integration-ci).
+	runner.RunFile(t, "../chain_operation/statesync_operation.yaml", opts)
+}
+
+// TestUpgradeModule covers major and minor upgrade scenarios.
+// These require a specially prepared cluster where node binaries can be
+// swapped mid-test; they are not suitable for the standard docker-cluster-start setup.
+func TestUpgradeModule(t *testing.T) {
+	runner.RunFile(t, "../upgrade_module/major_upgrade_test.yaml", opts)
+	runner.RunFile(t, "../upgrade_module/minor_upgrade_test.yaml", opts)
```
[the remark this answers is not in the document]
amir-deris: Will remove this test from this PR as it requires special handling with `.sh` scripts and can be refactored in future PRs.

amir-deris:
> **Not a blocker, and explicitly not a regression in this PR** — flagging a pre-existing coverage gap that this work happened to surface, so it's tracked rather than lost.
> 
> While reviewing I compared the full suite against what CI actually invokes: **27 YAML files exist, but only 23 are referenced by any CI workflow.** Four are orphaned (run by no matrix row):
> 
> * `bank_module/multi_sig_send_test.yaml`
> * `bank_module/simulation_tx.yaml`
> * `tokenfactory_module/create_tokenfactory_test.yaml`
> * `template/template_test.yaml` ← intentional (it's the authoring example, not a real test)
> 
> The first three look accidental, not deliberate:
> 
> * The two bank files were added in [autobahn mempool #3522](https://github.com/sei-protocol/sei-chain/pull/3522), in the _same_ change that heavily edited `integration-test.yml` — only `send_funds_test.yaml` got wired into the matrix; its two siblings didn't.
> * Unlike the genuinely environment-gated tests (the upgrade / statesync cases carry explicit "requires a specially prepared cluster" notes), these carry no skip marker and need no special setup — they're just unreferenced.
> 
> This PR **faithfully preserves the existing coverage** — the Go `-run` scope reproduces exactly what the Python matrix ran, so nothing is dropped here. If anything the Go runner _improves_ visibility: `TestBankModule` already wires all three bank files and `TestTokenFactoryModule` wires the tokenfactory one, so `go test ./...` makes these orphans discoverable in a way the per-row Python CLI strings never did.
> 
> **Suggested follow-up (independent of this PR, and applies to both harnesses):** wire these three back into CI — in the Python matrix today, and in the Go `-run` scope as each module migrates — or add an explicit skip marker if any are intentionally held out. Happy to file a tracking issue so it doesn't ride on this POC.

@bdchatham Thanks for great feedback, I didn't notice that some of these yaml files are unused actually. It seems these yaml files that are unused were added in this old pr: https://github.com/sei-protocol/sei-chain/pull/949 and not a new thing. I will enable them in the go runner to eliminate this gap.

.github/workflows/integration-test.yml
```diff
@@ -128,9 +128,7 @@ jobs:
           {
             name: "Mint & Staking & Bank Module",
             scripts: [
-              "python3 integration_test/scripts/runner.py integration_test/staking_module/staking_test.yaml",
-              "python3 integration_test/scripts/runner.py integration_test/bank_module/send_funds_test.yaml",
-              "python3 integration_test/scripts/runner.py integration_test/mint_module/mint_test.yaml"
+              "go test -tags yaml_integration -v -timeout 10m ./integration_test/runner/... -run \"TestMintModule|TestStakingModule|TestBankModule\""
```
masih: Please use top level CI timeout (defaults to 30m) or extrapolate via env var.

integration_test/runner/runner.go
```diff
@@ -0,0 +1,336 @@
...
+	// Defaults to /root/go/bin:/root/.foundry/bin (matching runner.py).
+	ExtraPath string
+}
+
+// RunFile reads path as a YAML list of TestCases and runs each as a subtest of t.
+func RunFile(t *testing.T, path string, opts Options) {
+	t.Helper()
+	data, err := os.ReadFile(path) //nolint:gosec
+	if err != nil {
+		t.Fatalf("read %s: %v", path, err)
```
masih: Use testify? Ditto for the rest of the file.

integration_test/runner/runner.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+func runCase(t *testing.T, tc TestCase, opts Options) {
+	t.Helper()
+	envMap := make(map[string]string)
+
+	for i, inp := range tc.Inputs {
+		container := inp.Node
+		if container == "" {
+			container = opts.DefaultContainer
```
masih, line 116: non-blocker: I recommend using functional options pattern. That way all defaults are in one place and a lot clearer to reason about gaps in config. Otherwise we end up with zero-value checks across multiple functions which works but easy to miss gaps or manage what the default options should be.

integration_test/runner/runner.go
```diff
@@ -0,0 +1,336 @@
...
+		args := make([]string, 1, 1+2*len(envMap)+4)
+		args[0] = "exec"
+		for k, v := range envMap {
+			args = append(args, "-e", k+"="+v)
+		}
+		args = append(args, container, "/bin/bash", "-c",
+			"export PATH=$PATH:"+extraPath+" && "+cmd)
+		c = exec.Command("docker", args...) //nolint:gosec
+	} else {
+		c = exec.Command("/bin/bash", "-c", cmd)
```
masih: We don't know what the shell is going to be. ideally we should parameterise this, or interpret it from runtime PATH, or stick with `sh` and make sure commands are compatible with it.

This also completely breaks on non-unix based systems but that's OK for us today.

masih (commented):
Nice; glad to see python replacement! 

This reminds me: why this PR doesn't remove the python runner?

amir-deris:
> Nice; glad to see python replacement!
> 
> This reminds me: why this PR doesn't remove the python runner?

@masih I thought I keep the scope small and do it in future PRs. But if you prefer, I can go ahead and migrate all the tests and remove the python runner?

masih:
> I thought I keep the scope small and do it in future PRs.

ah good call; let's do that

.github/workflows/integration-test.yml
```diff
@@ -128,9 +129,7 @@ jobs:
           {
             name: "Mint & Staking & Bank Module",
             scripts: [
-              "python3 integration_test/scripts/runner.py integration_test/staking_module/staking_test.yaml",
-              "python3 integration_test/scripts/runner.py integration_test/bank_module/send_funds_test.yaml",
-              "python3 integration_test/scripts/runner.py integration_test/mint_module/mint_test.yaml"
+              "go test -tags yaml_integration -v -timeout $JOB_TIMEOUT ./integration_test/runner/... -run \"TestMintModule|TestStakingModule|TestBankModule\""
```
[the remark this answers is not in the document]
masih: We want to keep the order the same @amir-deris just to avoid any related failures

integration_test/runner/runner.go
```diff
@@ -0,0 +1,359 @@
...
+
+	if container != "" {
+		// capacity: 1 ("exec") + 2*len(envMap) ("-e" + "k=v" per entry) + 4 (container, "/bin/bash", "-c", cmd)
+		args := make([]string, 1, 1+2*len(envMap)+4)
+		args[0] = "exec"
+		for k, v := range envMap {
+			args = append(args, "-e", k+"="+v)
+		}
+		args = append(args, container, opts.Shell, "-c",
+			"export PATH=$PATH:"+opts.ExtraPath+" && "+cmd)
```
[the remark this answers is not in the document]
masih: This one tracks for me, @amir-deris could you take a look please?

bdchatham (approved):

PR #3556, opened 2026-06-08
fix(evmrpc): bound and validate storageKeys in eth_getProof

amir-deris:
## Summary

`eth_getProof` had two vulnerabilities in its `storageKeys` handling:

- **No length cap**: There was no cap on the number of keys that could be sent, and it could cause issue with RPC server processing under concurrency.
- **No key validation**: keys were accepted as raw byte strings via `common.BytesToHash([]byte(key))` instead of being decoded as hex-encoded storage slots. This differs from `GetStorageAt`, which already uses `decodeHash` to reject malformed input.

### Changes

- `MaxStorageKeysPerProof = 1000` — requests exceeding this limit are rejected before any store I/O (matches geth's EIP-1186 implementation)
- Replace `common.BytesToHash([]byte(key))` with `decodeHash(key)` — rejects non-hex strings and correctly interprets hex slot addresses, consistent with `GetStorageAt`
- Update `TestGetProof` to use properly hex-encoded slot keys; add assertions that malformed keys and oversized key arrays are both rejected


bdchatham (approved):
LGTM

sei-will (approved):

evmrpc/state.go
```diff
@@ -28,6 +28,8 @@ import (
 
 var errNoProofCapableQueryableKVStore = errors.New("cannot find a proof-capable queryable KV store")
 
+const MaxStorageKeysPerProof = 1024
```
masih, line 31: There is no need to export this constant, in that it is only used by the evmrpc package. Right?

If so, then I recommend unexporting it.

masih (approved):

evmrpc/state.go
masih: Ah I see this is used in _test package too. never mind

PR #3560, opened 2026-06-09
fix(giga): robust handling of unsupported iterators

arajasek:

Since the giga cachekv doesn't support iterations, https://github.com/sei-protocol/sei-chain/pull/3103 changed the giga executor to fall back to v2 for associations and migrations, which are the primary triggers for iteration. There are (at least) 2 others, however:
- self-destructs (triggered by stateDB.Finalize)
- CreateAccount-over-existing-code (panics mid-execution).

This PR adds a lightweight check for self-destructs, falling back to v2 in that case. It also robust-ifies unsupported iteration handling by adding a special error for that case, which upon being detected triggers a v2 fallback.

In theory _all_ we need is the unsupported iteration detection, but it is more optimal to special-case associations and migrations (as in #3103) and self-destructs (as in this PR).

New tests from Cursor pass. Will confirm this fixes the pacific issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -479,7 +479,8 @@ autobahn-integration-test:
 	@GOWORK=off go test -tags autobahn_integration -v -count=1 -timeout 30m ./integration_test/autobahn/...
 .PHONY: autobahn-integration-test
 
-# Run a mixed-mode cluster: node 0 uses GIGA_EXECUTOR, nodes 1-3 use standard V2.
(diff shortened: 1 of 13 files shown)


masih:
@claude ultra review this PR

masih (approved):

stevenlanders (approved):

philipsu522 (approved):

PR #3561, opened 2026-06-09
test(giga): self-destruct storage cleanup coverage under OCC giga

stevenlanders:

Adds integration coverage for self-destructing an EVM contract that carries storage under the **concurrent giga executor** (`ProcessTXsWithOCCGiga`).

- New `contracts/src/SelfDestructTester.sol` — seeds storage on init (so there's real account state to clean up), plus a factory that creates + destroys a child in one tx.
- New `EVMGigaTest.js` scenario covering both EIP-6780 paths: later-tx selfdestruct (balance sweep only; code/storage retained) and same-tx create+destroy (full account cleanup).
- Flips node 0 of the mixed-mode determinism cluster to `GIGA_OCC=true` (`docker-compose.giga-mixed.yml`). The previous `GIGA_OCC=false` setting ran node 0 on the synchronous giga path over a plain (iterable) cachekv, so the self-destruct storage-purge path was never exercised. With OCC enabled the purge iterates the giga cachekv store — the path that hit `panic: unexpected iterator call on cachekv store` in production.

This is the regression/repro coverage for the v2-fallback fix in #3560. Without that fix node 0 diverges from V2 (giga returns `ErrPanic`) and halts on an AppHash mismatch; with it, the giga node falls back to V2 for self-destruct txs and the cluster stays consistent.

Test changes only — no production code paths modified.

- `hardhat compile` succeeds (evm target: prague); `node --check` passes on the test file.
- Verified the merged compose config pins node 0 to `GIGA_EXECUTOR=true GIGA_OCC=true` and nodes 1-3 to V2, regardless of host env.
- Runs via the existing `EVM GIGA Mixed (Determinism)` CI job and `make giga-mixed-integration-test`. Expected to fail on `main` (reproduces the panic) and pass once #3560 lands.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -479,7 +479,8 @@ autobahn-integration-test:
 	@GOWORK=off go test -tags autobahn_integration -v -count=1 -timeout 30m ./integration_test/autobahn/...
 .PHONY: autobahn-integration-test
 
-# Run a mixed-mode cluster: node 0 uses GIGA_EXECUTOR, nodes 1-3 use standard V2.
+# Run a mixed-mode cluster: node 0 uses GIGA_EXECUTOR with OCC, nodes 1-3 use standard V2.
+# (node-level GIGA_EXECUTOR/GIGA_OCC values are pinned in docker-compose.giga-mixed.yml)
 # Any determinism divergence between giga and V2 will cause the giga node to halt.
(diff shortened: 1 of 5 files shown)


philipsu522 (approved):

stevenlanders:
closing in favor of #3560 which includes this test

PR #3565, opened 2026-06-09
refactor(consensus-policy): idiomatic fmt.Errorf %w wrapping (follow-up to #3429)

bdchatham:
## What

Follow-up to #3429 addressing @masih's post-merge review: replace the bespoke `ConsensusPolicyError` API (`With` / `cause` / `Unwrap` / custom `Is` / `Label`) with native Go `fmt.Errorf("%w", sentinel)` wrapping of the same exported sentinel pointers, and consolidate the type to a single `Kind` field.

## Changes

- **`ConsensusPolicyError` shrinks** to:
  ```go
  type ConsensusPolicyError struct { Kind string }
  func (e *ConsensusPolicyError) Error() string { return e.Kind }
  ```
  Deleted: the `msg`/`cause` fields and the `With`, `Unwrap`, `Label`, and custom label-based `Is` methods.
- **Call sites** use `fmt.Errorf("…: %w", sentinel)`. The custom `Is` is gone because sites now wrap the **same exported sentinel pointer**, so `errors.Is` matches by identity (strictly more correct). The 6 sites that must also keep an inner typed error reachable use **multi-`%w`**: `replay.go` AppHash ×4, and `validation.go` `LastCommitVerify` (inner `ErrInvalidCommitHeight`/`Signatures`) + `EvidenceOverflow` (inner `ErrEvidenceOverflow`).
- **Metric** (`recordUnsafeValidationSkipped`) reads the exported `Kind` via `errors.As`. The 13 `sei_unsafe_validation_skipped_total{kind=…}` label strings are **byte-identical** (a MUST-NOT-change contract).
- Doc comments rewritten off `.With`/`cause` onto the `fmt.Errorf`-wrap + `errors.As`-by-`Kind` mechanism.

## Behavior-preserving (cross-reviewed)

Verified by two independent lenses (idiom + correctness) and an adversarial diff cross-review:
- **3-tag swallow semantics unchanged:** default halts all; `mock_block_validation` swallows AppHash+DataHash; `mock_chain_validation` swallows all swallow-eligible except `ErrLastCommitVerify`.
- **`errors.As` reachability preserved:** `TestValidateBlockCommit` and `TestValidateBlockEvidence` pass on the default tag (both `errors.Is(sentinel)` and `errors.As(innerType)` hold under multi-`%w`).
- **No new failures:** stash-compare under both mock tags shows FAIL sets byte-identical to baseline (the known-environmental swallowed-validation tests, not introduced here). Default tag fully green.
- `gofmt -s` clean; builds green under default / `mock_block_validation` / `mock_chain_validation`.

## Relates to

- #3429 — the original typed-error change this follows up
- [`.With` redundancy thread](https://github.com/sei-protocol/sei-chain/pull/3429#discussion_r3379810072)
- [`label`/`msg`/`cause` over-engineering thread](https://github.com/sei-protocol/sei-chain/pull/3429#discussion_r3379816352)


sei-tendermint/internal/state/validation.go
```diff
@@ -42,50 +42,50 @@ func validateBlock(state State, block *types.Block, policy types.ConsensusPolicy
 	}
 	// Validate prev block info.
 	if !block.LastBlockID.Equals(state.LastBlockID) {
-		if err := policy.HandleError(types.ErrLastBlockID.With(
-			"wrong Block.Header.LastBlockID.  Expected %v, got %v",
-			state.LastBlockID, block.LastBlockID)); err != nil {
+		if err := policy.HandleError(fmt.Errorf(
+			"wrong Block.Header.LastBlockID.  Expected %v, got %v: %w",
```
masih: Small conventional styling:
```suggestion
			"wrong Block.Header.LastBlockID: expected %v, got %v: %w",
```

Ditto in other places.

sei-tendermint/internal/state/validation.go
```diff
@@ -42,50 +42,50 @@ func validateBlock(state State, block *types.Block, policy types.ConsensusPolicy
...
 		}
 	}
 
 	// Validate app info.
 	if !bytes.Equal(block.AppHash, state.AppHash) {
-		if err := policy.HandleError(types.ErrAppHash.With(
-			"wrong Block.Header.AppHash.  Expected %X, got %v",
-			state.AppHash, block.AppHash)); err != nil {
+		if err := policy.HandleError(fmt.Errorf(
+			"wrong Block.Header.AppHash.  Expected %X, got %v: %w",
```
masih: Why use two different encodings for state and block apphash (i.e. `%X` vs `%v`? I recommend using consistent formatting. 

Better yet, if apphash is a stringer then you can just use `%s`.

Ditto in other places.

sei-tendermint/internal/state/validation.go
```diff
@@ -99,8 +99,8 @@ func validateBlock(state State, block *types.Block, policy types.ConsensusPolicy
 		// LastCommit.Signatures length is checked in VerifyCommit.
 		if err := state.LastValidators.VerifyCommit(
 			state.ChainID, state.LastBlockID, block.Height-1, block.LastCommit); err != nil {
-			if swErr := policy.HandleError(types.ErrLastCommitVerify.With(
-				"VerifyCommit(): %w", err)); swErr != nil {
+			if swErr := policy.HandleError(fmt.Errorf(
+				"VerifyCommit() %w: %w", types.ErrLastCommitVerify, err)); swErr != nil {
```
masih: Always start error messages with lower case.

sei-tendermint/types/consensus_policy.go
```diff
@@ -13,83 +13,44 @@
 //	                         avoid a downstream buildLastCommitInfo panic
 //
 // Validation failures are modeled as *ConsensusPolicyError sentinels. Call
-// sites wrap a contextual message via Err*.With(format, args...); the policy
-// matches sentinels with errors.Is, so identity survives %w wrapping.
+// sites attach context with idiomatic fmt.Errorf("...: %w", ErrX): wrapping
+// the sentinel pointer keeps errors.Is(err, ErrX) true by identity and lets
+// recordUnsafeValidationSkipped recover the sentinel — and its Kind label —
+// via errors.As. Sites that must keep an inner typed error reachable too use
+// Go 1.20+ multi-%w (fmt.Errorf("%w: %w", ErrX, inner)).
```
masih: Go 1.20+ is implied by the module version this repository enforces.

sei-tendermint/types/consensus_policy.go
```diff
@@ -13,83 +13,44 @@
...
-// captured in cause and exposed via Unwrap, so errors.Is/As can still reach
-// the underlying typed error (e.g. ErrInvalidCommitHeight, ErrEvidenceOverflow).
-type ConsensusPolicyError struct {
-	label, msg string
-	cause      error
-}
+// ConsensusPolicyError is a swallow-eligible validation failure. Kind is the
+// stable metric label on sei_unsafe_validation_skipped_total{kind=...} and
+// MUST NOT change. Matched via errors.Is by identity; recovered via errors.As.
+type ConsensusPolicyError struct{ Kind, msg string }
```
masih: I personally use `Code` instead of `Kind` and use constant `iota` numerical values instead of string _when code/kind is needed_.

Here, i would ask: why do we need Kind at all when there are sentinel error types?

amir-deris (approved):

sei-tendermint/types/consensus_policy.go
masih: @bdchatham can you explain why you decided to keep the sting kind?

bdchatham: On Code+iota: kept it as a string `Kind` deliberately, because this value's whole job is to be the Prometheus label on `sei_unsafe_validation_skipped_total{kind=...}` — a byte-identical external contract (dashboards/alerts/recording rules key on the exact strings). An iota `Code` would still need a parallel `map[Code]string` to produce that label, so it adds a second source of truth plus positional-renumbering fragility (insert/reorder a const and labels silently shift) — for no gain, where the string field *is* the label. iota's a great default for internal enums; this one is deliberately externalized, and an exported field recovered via `errors.As` also matches the package's existing typed-error idiom (`errors.go`'s `ErrInvalidCommitHeight`). Happy to revisit if you'd still prefer it. The styling notes (1–4) are addressed in b468088.

masih: Having a kind, because metrics need label doesn't quite make sense for me. That's letting metrics affect the API design. 

I think in this case, we could get rid of the Kind completeley (Code would also be redundant) if we simply either reflect on the type name or hardcode the naming in a simple switch-case statement _in the metric reporting subsystem_. But the concern of "metrics want label" should not leak into a core error type for ConsensusPolicy. The design critique in me would say that: It is the business of metric reporting to decide how to map errors to strings, not the error type.

bdchatham: It's a fair point, I'll tweak this

sei-tendermint/internal/consensus/replay.go
```diff
@@ -261,7 +261,7 @@ func (h *Handshaker) ReplayBlocks(
 	switch {
 	case storeBlockHeight == 0:
 		if err := checkAppHashEqualsOneFromState(appHash, state); err != nil {
-			if err = h.consensusPolicy.HandleError(types.ErrAppHash.With("%w", err)); err != nil {
```
pompon0, line 264: this will not work, only one %w is supported by fmt.Errorf. Non-sentinel error can be inlined with %v.

pompon0: nvm, that's not true since go 1.20+

sei-tendermint/types/consensus_policy.go
masih: Left a comment on Slack; let's chat about [this](https://github.com/sei-protocol/sei-chain/pull/3565/changes/7497c83e423b25678a0e1a58875ff028d85fb20c) commit synchronously. Looks like the core error type is removed which would make it inconvenient to infer if a given error is in the consensus policy category or not. We ant both sentinel _and_ root error type for errors is and group errors as inference.

Glad to see the metric labels are isolated in telemetry; ty for reflecting on my comment. 
Having the root error type means that switch statement can return fast if a given error is not a consensus policy error to begin with.

masih (approved):

PR #3570, opened 2026-06-09
ci: split slow integration test jobs to reduce CI wall-clock by ~17%

amir-deris:
## Summary
Split 3 slow integration test jobs into parallel matrix entries, each targeting independent test files. No test logic changed — only how tests are distributed across runners.

**EVM Interoperability** (2 → 4 jobs):
- `EVM Interoperability (Pointer Tests)` — 7 pointer-type Hardhat tests
- `EVM Interoperability (Misc Tests)` — SeiSolo, SetCodeTx, TransientStorage
- Same split applied to Autobahn variants

**EVM Module** (2 → 4 jobs):
- `EVM Module (Compat)` — EVMCompatabilityTest.js (identified as 70–77% of runtime via timing instrumentation)
- `EVM Module (Precompile & Endpoints)` — EVMPrecompileTest.js, SeiEndpointsTest.js, AssociateTest.js + FlatKV steps
- Same split applied to Autobahn variants

**dApp Tests** (1 → 2 jobs):
- `dApp Tests (Uniswap)` — uniswapTest.js (219s, 62% of runtime)
- `dApp Tests (NFT & Steak)` — nftMarketplaceTests.js + SteakTests.js (133s combined)

## Results (measured over multiple CI runs)

| Milestone | Critical path job | Wall-clock |
|---|---|---|
| Baseline (pre-PR) | Autobahn EVM Interoperability (667s avg) | ~19.3 min |
| After EVM Interop split | Autobahn EVM Module (622s) | 17.6 min |
| After EVM Module split | dApp Tests (600s) | 17.1 min |
| After dApp split | Autobahn EVM Interop (Pointer Tests) (559s) | **16.0 min** |

**Total: ~3.3 min saved (~17%) per CI run.**

## Approach
Used timing instrumentation (`[timing]` log lines) on `evm_tests.sh` and `dapp_tests.sh` to identify the dominant test file in each job before splitting. Instrumentation removed after data was collected.

## Bug fix: AssociateTest flake on Autobahn
Fixed a pre-existing flake in `AssociateTest.js` where `verifyAssociation` read `afterEvm` immediately after `associateKeyStrict` returned, relying on `waitForBlocks(2)` as a proxy for the EVM balance merge landing. On Autobahn this was insufficient. Replaced with a `waitForCondition` poll that waits until the EVM balance actually reaches the expected value before asserting. Same failure reproduced on PR #3566 (pre-split, original `evm_tests.sh`), confirming it is Autobahn-specific and unrelated to the job split.


bdchatham:
Verified the split preserves coverage — the six new scripts run exactly the same test files as the three they replace (EVM Module 4≡4, Interop 10≡10, dApp 3≡3), nothing dropped or duplicated. Wall-clock improvement looks real too.

One thing worth confirming before merge: the old scripts ran each group sequentially against a single chain, whereas the split now runs them on separate fresh clusters. So if any test depended on on-chain state left behind by a sibling that ran earlier in the same script — e.g. AssociateTest/SeiEndpoints after EVMCompatability, or the steak/nft reorder — the split would break it silently. If each `npx hardhat test` stands up its own state (the usual case), this is a non-issue.

If test isolation isn't a concern here, LGTM.

bdchatham (approved):

amir-deris:
> Verified the split preserves coverage — the six new scripts run exactly the same test files as the three they replace (EVM Module 4≡4, Interop 10≡10, dApp 3≡3), nothing dropped or duplicated. Wall-clock improvement looks real too.
> 
> One thing worth confirming before merge: the old scripts ran each group sequentially against a single chain, whereas the split now runs them on separate fresh clusters. So if any test depended on on-chain state left behind by a sibling that ran earlier in the same script — e.g. AssociateTest/SeiEndpoints after EVMCompatability, or the steak/nft reorder — the split would break it silently. If each `npx hardhat test` stands up its own state (the usual case), this is a non-issue.
> 
> If test isolation isn't a concern here, LGTM.

@bdchatham thanks for feedback. That is a valid concern but it seems that is not causing issue here. For dApp split: `SteakTests.js` and `nftMarketplaceTests.js` both set up fresh accounts and deploy contracts in their own before() hooks. 

For `EVMCompatabilityTest.js`, it is isolated into
  `evm_compat_tests.sh`, while `EVMPrecompileTest.js` + `SeiEndpointsTest.js` + `AssociateTest.js` stayed together in `evm_precompile_tests.sh` and no cross-cluster dependency introduced.

masih (approved):

PR #3572, opened 2026-06-10
LittDB Garbage Collection Fine Control

cody-littley:

Previously, LittDB garbage collection was controlled entirely by a TTL. That is, data was deleted when it got old, and there was no other criteria. 

This PR adds a new mechanism that allows additional constraints to be put on garbage collection. This will be necessary in order to precisely control the timing of deletions in the block store. We need to keep around block data until a logical block height which is only loosely coupled to wall clock time.

Note: in order to add this feature, I had to refactor how we configure LittDB tables. That refactor is responsible for the majority of the line changes in this PR.


Kbhat1 (approved):
Thanks @cody-littley

PR #3576, opened 2026-06-11
Composite Store unit tests

cody-littley:

New suite of unit tests for `composite.Store`

### DIFF
--- a/sei-db/state_db/sc/composite/random_test_framework_test.go
+++ b/sei-db/state_db/sc/composite/random_test_framework_test.go
@@ -14,7 +14,9 @@ package composite
 // coupling.
 
 import (
+	"bytes"
 	"encoding/binary"
+	"fmt"
 	"sort"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


sei-db/state_db/sc/composite/random_test_framework_test.go
```diff
@@ -0,0 +1,1514 @@
...
+	return out
+}
+
+// liveKeySetFromOracle rebuilds a liveKeySet that exactly covers the keys in
+// the given oracle. Used after a rollback / reconcile to resume the workload
+// from a restored state, since the in-flight liveKeySet no longer matches the
+// rolled-back keyspace.
+func liveKeySetFromOracle(o *storeOracle) *liveKeySet {
+	s := newLiveKeySet()
+	for store, m := range o.stores {
```
blindchaser, line 190: Go map iteration order is randomized per run, so the resulting keys slice order is nondeterministic. Every later keysInUse.Sample(...) indexes into that slice, meaning which keys get read/updated/deleted/sampled depends on map order, not the RNG seed. This contradicts the harness's explicit contract ("a given seed yields a byte-identical apply/commit sequence", "fully reproducible from r's seed", framework header lines 158–160 and 439–440). The guarantee holds for Phase 1 only; it's violated everywhere liveKeySetFromOracle is used — steady-state Phases 2+ (random_test.go:82,124,157) and every migration resume (:368,392,410).

Practical impact: when CI fails and prints Random seed: N, re-running with seed N will not reproduce the failing sequence past the first rollback — the printed seed is misleading exactly when you need it. Fix is small: collect the pairs into a slice, sort by (store, key), then Add in sorted order.

sei-db/state_db/sc/composite/random_test_framework_test.go
```diff
@@ -0,0 +1,1514 @@
...
+		}
+	}
+}
+
+// assertMigrationInFlight verifies that, across migratingStores, at least one
+// tracked key is still in memiavl (un-migrated) AND at least one is already in
+// flatkv (migrated). Use immediately before a mid-migration restart to confirm
+// the test actually exercises the in-flight resume path. Ported from the
+// migration framework's AssertMigrationInFlight.
+func assertMigrationInFlight(t *testing.T, cs *CompositeCommitStore, oracle *storeOracle, migratingStores ...string) {
```
blindchaser, line 1407: flakiness: assertMigrationInFlight is seed-dependent and can spuriously fail

blindchaser (approved):

sei-db/state_db/sc/composite/random_test_framework_test.go
cody-littley: change made

sei-db/state_db/sc/composite/random_test_framework_test.go
cody-littley: fixed

PR #3583, opened 2026-06-15
[codex] add evm-only giga executor path

codchen:
## Summary

- add `giga/evmonly` as the final-form EVM-only giga execution path boundary
- define raw Ethereum RLP block input and commit-neutral outputs as an EVM-native state changeset plus receipts
- port the core sei-v3-style execution shape directly into `giga/evmonly`:
  - raw RLP tx parsing and sender recovery
  - go-ethereum execution against an SDK-free native `vm.StateDB`
  - key-addressable EVM-native balance, nonce, code, and storage state reads
  - deterministic post-block `StateChangeSet` output
  - Ethereum receipt and per-tx metadata construction
- add a map-backed `MemoryState` backend for tests and early integration
- leave custom precompiles as fail-closed placeholders behind an SDK-free registry/context
- refresh the README to describe the current implementation, input/output contract, and known limitations

## Notes

Custom precompile behavior is intentionally still open. Registered custom precompile addresses return `ErrCustomPrecompilesOpen`, including through geth's precompile map, so calls fail closed instead of silently executing as empty accounts.

The current port is sequential. The state boundary and changeset shape are intended to be replaceable with the sei-v3 OCC scheduler/store once that layer is brought over.

Historical `BLOCKHASH` lookups beyond the parent block are not wired yet; `BlockHash` is currently used for receipt/log metadata.

## Validation

```sh
go test ./giga/evmonly/...
go test ./giga/...
```


giga/evmonly/README.md
```diff
@@ -0,0 +1,134 @@
...
+them behind an EVM-native context that is visible to the executor's conflict
+tracking without reintroducing Cosmos keeper dependencies.
+
+## Current implementation
+
+The `evmonly` package currently provides:
+
+- sequential execution of the ordered block transaction list
+- RLP decoding and sender recovery through go-ethereum signers
+- go-ethereum `core.ApplyMessage` execution against an SDK-free `vm.StateDB`
```
philipsu522, line 28: does "SDK" mean Cosmos?

giga/evmonly/state_db.go
```diff
@@ -0,0 +1,653 @@
...
+
+func (s *nativeStateDB) AccessEvents() *vm.AccessEvents {
+	return nil
+}
+
+func (s *nativeStateDB) Finalise(bool) {
+	for _, acct := range s.accounts {
+		if acct.SelfDestructed {
+			acct.Code = nil
+			acct.Storage = map[common.Hash]common.Hash{}
```
philipsu522: do we need to set acct.Created = false?

giga/evmonly/executor.go
```diff
@@ -0,0 +1,260 @@
...
+
+	stateDB := newNativeStateDB(e.state)
+	blockCtx := buildBlockContext(req.Context)
+	evm := vm.NewEVM(blockCtx, stateDB, chainConfig, vm.Config{}, customPrecompileMap(e.cfg.CustomPrecompiles))
+	stateDB.SetEVM(evm)
+
+	gasLimit := req.Context.GasLimit
+	if gasLimit == 0 {
+		gasLimit = math.MaxUint64
+	}
```
philipsu522: is this dangerous? could someone maliciously set gasLimit = 0 to bypass any limits we may have?

giga/evmonly/state_db.go
```diff
@@ -0,0 +1,653 @@
+package evmonly
```
philipsu522, line 1: this is a lot of critical state changing code. i think we need a lot of tests here. Could we use AI to generate a ton of unit cases, interleaving ordering of contract creation/deletion, invalid txs (out of gas, invalid state transition, invalid nonce, verify receipt outputs...etc).

giga/evmonly/executor.go
codchen: gasLimit here is the block gas limit, so it'd be decided by consensus and not set by tx senders

giga/evmonly/README.md
codchen: yep

giga/evmonly/state_db.go
codchen: Created is used to decided whether `SelfDestruct6780` should early return, and I think the behavior of `SelfDestruct6780` is to not early return if selfdestruct is called multiple times, so I think we shouldn't set Created to false upon selfdestruct (consistent with geth)

giga/evmonly/state_db.go
```diff
@@ -0,0 +1,660 @@
...
+			})
+		}
+		if !bytes.Equal(acct.Code, base.Code) {
+			changes.Code = append(changes.Code, CodeChange{
+				Address: addr,
+				Code:    cloneBytes(acct.Code),
+				Delete:  len(acct.Code) == 0,
+			})
+		}
+		storageKeys := storageKeyUnion(base.Storage, acct.Storage)
```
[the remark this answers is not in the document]
codchen: Fixed in 145f08917. `StateChangeSet` now carries `StorageClears`, and `MemoryState.ApplyChangeSet` applies those account-level storage wipes before per-slot updates. Covered by selfdestruct storage-clear regression tests.

giga/evmonly/state_db.go
```diff
@@ -0,0 +1,918 @@
...
+
+func (s *nativeStateDB) Prepare(_ params.Rules, sender, coinbase common.Address, dest *common.Address, precompiles []common.Address, txAccesses ethtypes.AccessList) {
+	s.accessList.reset()
+	clearNestedHashMaps(s.transientStates)
+	s.AddAddressToAccessList(sender)
+	s.AddAddressToAccessList(coinbase)
+	if dest != nil {
+		s.AddAddressToAccessList(*dest)
+	}
+	for _, addr := range precompiles {
```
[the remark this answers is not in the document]
codchen: Shanghai will always be on

giga/evmonly/occ.go
```diff
@@ -0,0 +1,527 @@
...
+		result.OCCStats = OCCStats{
+			Attempted:      true,
+			Fallback:       true,
+			FallbackReason: occFallbackReasonSpeculativeError,
+		}
+		return result, nil
+	}
+	validation := validateOCCResults(results, gasLimit)
+	if !validation.valid {
+		result, err := e.executeBlockSequential(ctx, req)
```
masih: With whole-block OCC, this fallback would become the common case. We would hit it with high certainty at 80K+ tps throughput on mainnet/real traffic?

This then means the computation cost is ~2x of plain sequential execution where we really need to maximise throughput. From security perspective an attacker can induce the majority common case with few cheap transactions IIUC.


is there an alternative world where we can make the validation more granular instead? How does the the "v3" scheduler affect the common case?

It would be worth to write a benchmark for `executeBlockOCC` to confirm table-stake theoretical throughput.

giga/evmonly/parser.go
```diff
@@ -0,0 +1,41 @@
...
+
+func parseBlockTxs(ctx context.Context, txs [][]byte, signer ethtypes.Signer) ([]PreparedTx, error) {
+	parsed := make([]PreparedTx, len(txs))
+	for i, raw := range txs {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+		}
+		tx, sender, err := parseTx(raw, signer)
```
masih: `ethtypes.Sender` is the slow part, and this is single threaded. That would make hitting the 3 digit tps we want theoretically unlikely? A go benchmark should be able to confirm this.

An alternative that comes to mind is to fan sender recovery out across the worker pool. This should make a big difference.

giga/evmonly/state_db.go
```diff
@@ -0,0 +1,660 @@
...
+			s.AddSlotToAccessList(tuple.Address, key)
+		}
+	}
+}
+
+func (s *nativeStateDB) PointCache() *ethutils.PointCache {
+	return nil
+}
+
+func (s *nativeStateDB) Snapshot() int {
```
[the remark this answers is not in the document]
masih: This needs a review given the context of work here is to maximise throughput. `recordAccount` additionally clones whole accounts  which makes snapshot and revert to snapshot expensive.

Can we implement a snapshot-by-mutation-kind mechanism instead? That makes both snapshot and revert much more efficient.

I would write the benchmark first.

giga/evmonly/occ.go
```diff
@@ -0,0 +1,527 @@
...
+	req PreparedBlock,
+	txIndex int,
+	txIndexUint uint,
+	chainConfig *params.ChainConfig,
+	blockCtx vm.BlockContext,
+	baseFee *big.Int,
+	gasLimit uint64,
+) (occTxExecution, error) {
+	p := req.Txs[txIndex]
+	stateDB := newNativeStateDB(e.state)
```
masih: The stateDB makes a lot of allocations which at high tps scale adds up quickly.

Can we make a reusable stateDB  and reuse it in the OCC path? We could build one mutable snapshot per block that has the parent state with a thin per tx writer overlay. You can imagine all sorts of pooling per worker to reduce the allocations.

giga/evmonly/state.go
```diff
@@ -0,0 +1,182 @@
...
+import (
+	"math/big"
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// StateReader supplies EVM-native state to an executor.
+type StateReader interface {
+	GetBalance(common.Address) *big.Int
```
masih, line 17: Thinking out loud, I am not sure if returning *big.Int is wise at interface level. This would force allocation for all implementors of this interface.

Given this interface has a heavy footfall and it might be cumbersome to refactor these interfaces later, would be worth to give this more thought to see what makes a more memory efficient API boundary. For example, can we use `uint256.Int` passed by value.

giga/evmonly/occ.go
```diff
@@ -0,0 +1,527 @@
...
+	results := make([]occTxExecution, txCount)
+	chunkSize := occChunkSize(txCount, workers)
+	pool := e.occPool
+	if pool == nil {
+		pool = newOCCWorkerPool(workers)
+		defer pool.Close()
+	}
+	if err := pool.Run(ctx, occRanges(txCount, chunkSize), func(workerCtx context.Context, txRange occTxRange) error {
+		for idx, idxUint := txRange.start, txRange.startUint; idx < txRange.end; idx, idxUint = idx+1, idxUint+1 {
+			result, err := e.executeTxSpeculative(workerCtx, req, idx, idxUint, chainConfig, blockCtx, baseFee, gasLimit)
```
masih: We go straight to speculative execution then fall back to sequential; I recommend doing cheap validations first and fallback quickly to sequential in cases where speculative execution is not possible.

 For example, sum of declared gas limit validation.

giga/evmonly/occ.go
```diff
@@ -0,0 +1,527 @@
...
+		workers = txCount
+	}
+	results := make([]occTxExecution, txCount)
+	chunkSize := occChunkSize(txCount, workers)
+	pool := e.occPool
+	if pool == nil {
+		pool = newOCCWorkerPool(workers)
+		defer pool.Close()
+	}
+	if err := pool.Run(ctx, occRanges(txCount, chunkSize), func(workerCtx context.Context, txRange occTxRange) error {
```
masih: Non-blocker for now _but_: We would need to improve worker load balancing here. The pool is initialised with fixed number of workers regardless of the number of transactions. An example of this is evident in the tests: a 24 tx block, 4 worker config, chunked to 2 would leave half the workers idle. It would be nice to make this adaptive relative to block itself. It also needs more data: what is the common case / how much fluctuation do we expect in terms of txs per block on mainnet.

giga/evmonly/occ_pool.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+	}
+	done.Wait()
+	if err != nil {
+		return err
+	}
+	if ctxErr := ctx.Err(); ctxErr != nil {
+		return ctxErr
+	}
+	return nil
```
masih: You can remove the if statement above entirely and simply return `ctx.Err()`.

giga/evmonly/occ_pool.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+		select {
+		case p.jobs <- job:
+		case <-jobCtx.Done():
+			done.Done()
+			break dispatch
+		case <-p.stop:
+			done.Done()
+			return fmt.Errorf("OCC worker pool is closed")
+		}
```
masih: This select statement seems unsafe and can deadlock. The select should check if the channels are closed via `ok, _ = <- etc.`, and the pool should drain the jobs channel on close. An alternative is to protect run/close order with a mutex so they can't overlap.

The specific case I can see happening, keeping in mind that:
* jobs is a buffered channel.
* `select` order is random
is:
* workers `Close`
* there is still buffered jobs 
* those jobs never run because workers are closed.
* `done.Done()` is never called.
* `done.Wait()` blocks forever.

giga/evmonly/occ_pool.go
```diff
@@ -0,0 +1,119 @@
...
+	for _, txRange := range ranges {
+		done.Add(1)
+		job := occPoolJob{
+			ctx:     jobCtx,
+			txRange: txRange,
+			run:     run,
+			done:    &done,
+			cancel:  cancel,
+			errOnce: &errOnce,
+			err:     &err,
```
masih: This logic could be significantly simplified by Err Group instead of wait group. 

That removes the erronce and err reference passing, and simplifies the loop. The could should be much easier to read too not to mention the battle tested native Go SDK capabilities.

giga/evmonly/types.go
```diff
@@ -0,0 +1,158 @@
...
+
+// Release returns a pooled BlockResult to its executor-owned pool. It is a
+// no-op for results that were not allocated from a pool.
+func (r *BlockResult) Release() {
+	if r == nil || r.lease == nil {
+		return
+	}
+	lease := r.lease
+	r.lease = nil
+	lease.release()
```
masih, line 100: This is not safe for concurrent use, which would mean `BlockResult.Release()` is not idempotent. 

That means  we can end up double decrementing the refs if more than one goroutine observe the same lease.

giga/evmonly/executor.go
```diff
@@ -0,0 +1,414 @@
...
+	release := result.retain()
+	if err := e.resultSink.StoreBlockResult(ctx, height, result, release); err != nil {
+		release()
+		return fmt.Errorf("store block result for block %d: %w", height, err)
+	}
+	return nil
+}
+
+func (e *Executor) useOCC(txCount int) bool {
```
masih, line 155: should this still return true after executor is closed? That would mean a whole redundant pool re-instantiation.

giga/evmonly/parser.go
philipsu522: +1:
- should probably add metrics / traces to all these calls 
- this for loop seems relatively straightfwd to parallelize unless..
- is it true that if 1 tx is malformed (fails parse) this entire function should fail?

giga/evmonly/occ.go
philipsu522: @DakaiKang and @muursh probably have some ideas here soon - they have been discussing inferring potential dependencies and tagging txs with that metadata. my 2c is this is probalby ok for a testnet, but definitely worth leaving a major todo here (or even the cheap gas limit load balancing is a good idea). 

would also suggest metrics here, occ traditionally has had pretty high performance swings

giga/evmonly/occ.go
```diff
@@ -0,0 +1,527 @@
...
+				return err
+			}
+			results[idx] = result
+		}
+		return nil
+	}); err != nil {
+		if ctxErr := ctx.Err(); ctxErr != nil {
+			return nil, ctxErr
+		}
+		result, seqErr := e.executeBlockSequential(ctx, req)
```
philipsu522: do we go straight to sequential if there is a single failure? seems like it diverges from our v2 occ impl? if it's intentional just for simplicity for now, let's add a comment

giga/evmonly/occ.go
```diff
@@ -0,0 +1,527 @@
...
+	var totalGasLimit uint64
+	var totalGasUsed uint64
+	validation := occValidationResult{valid: true}
+	for _, result := range results {
+		if len(result.changeSet.StorageClears) > 0 {
+			validation.valid = false
+			validation.fallbackReason = occFallbackReasonConflict
+			return validation
+		}
+		if result.gasLimit > math.MaxUint64-totalGasLimit || result.gasUsed > math.MaxUint64-totalGasUsed {
```
philipsu522: nit - should we check gas limit earlier before execution? seems cheap enough to check and fail early

giga/evmonly/state_db.go
```diff
@@ -0,0 +1,1072 @@
...
+		accounts:                 map[common.Address]*nativeAccount{},
+		base:                     map[common.Address]*nativeAccount{},
+		preimages:                map[common.Hash][]byte{},
+		accessList:               newAccessList(),
+		transientStates:          map[common.Address]map[common.Hash]common.Hash{},
+		finaliseAddrs:            map[common.Address]struct{}{},
+		committedStorage:         map[common.Address]map[common.Hash]storageValue{},
+		txStorageWrites:          map[common.Address]map[common.Hash]struct{}{},
+		txStorageClears:          map[common.Address]struct{}{},
+		commutativeBalanceDeltas: map[common.Address]*uint256.Int{},
```
philipsu522: looks like all of these are reinitialized every time statedB is created - do you think that would be an issue? i wonder if there is anything we can do here

giga/evmonly/occ.go
```diff
@@ -0,0 +1,527 @@
...
+	blockCtx vm.BlockContext,
+	baseFee *big.Int,
+	gasLimit uint64,
+) (occTxExecution, error) {
+	p := req.Txs[txIndex]
+	stateDB := newNativeStateDB(e.state)
+	stateDB.enableAccessTracking()
+	evm := vm.NewEVM(blockCtx, stateDB, chainConfig, vm.Config{}, nil)
+	stateDB.SetEVM(evm)
+	gasPool := new(core.GasPool).AddGas(gasLimit)
```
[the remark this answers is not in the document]
philipsu522: @codchen this looks like a correctness issue

giga/evmonly/executor.go
codchen: good point - updated to return false after close

giga/evmonly/occ.go
```diff
@@ -0,0 +1,934 @@
...
+				delete(s.storage, key)
+			}
+		}
+	}
+	for _, change := range result.changeSet.Storage {
+		s.storage[storageChangeKey{address: change.Address, key: change.Key}] = change.Value
+	}
+}
+
+func (s *blockSTMState) ChangeSet() StateChangeSet {
```
philipsu522, line 675: i wonder if this will be a bottleneck at load. you mentioned 200k tests still work - does that load test invoke overhead here?

codchen: this function is called on tx-level https://github.com/sei-protocol/sei-chain/pull/3583/changes#diff-69d9f1cd6d79e313457c7eb8a3df7380373776e00935771d8dd3ae80a3825cfcR376 so it should be intrinsically parallelized already (since tx-level tasks are parallelized)

giga/evmonly/occ.go
```diff
@@ -0,0 +1,934 @@
...
+				blockedGasLimit,
+			)
+			if err != nil {
+				return StateChangeSet{}, nil, err
+			}
+			reruns = append(reruns, rerun)
+			continue
+		}
+		if blockedPrefix >= 0 {
+			continue
```
masih: This doesn't matter on a conflict free traffic. _But_ it seems fatal on _any_ traffic with conflicts.

Because IIUC:
* the validation walks the txs, applies prefix only until it hits the first tx that needsSTMRerun. That index becomes blockedPrefix, and after that point nothing else is applied.
* The conflict check looks at the stateAccessIndex, which only contains applied txs. So if two txs conflict with each other somewhere after blockedPrefix, the walk can't see it yet. That conflict is only discovered in a later round, after the frontier has moved past it.
* Every rerun task in a round gets the same blockedSource. This basically means the prefix state at the first blocked tx. So a tx that depends on something at position blockedPrefix + 5 re-executes against state that still doesn't contain it, fails validation again, and burns another incarnation.

Put all of this together and i think we get: a block with k conflict spots needs ~k rounds, and each round is a full O(n) walk. And occMaxTxIncarnations = 10 means any dependency chain longer than ~10 gives up and runs the whole block through executeBlockSequential  _after ~10 rounds of wasted parallel work_. 

The thing is, i can imagine this kind of tx in a very typical traffic; for example: same sender many txs (e.g. gaming traffic we have seen before), same recipient many payers (e.g. any kind of useful app with traction).

Suggestions off the top of my head in ascending order of cost:
* we could use the existing `commutativeBalanceDeltas` to make all add balances commutative. this _should_ fix the hot-recipient traffic pattern.
* group the same sender txs into one task - this is the same pattern that I think executeBlockSequential usues. this fixes the nonce conflict issue
* keep applying to a second, "optimistic" stateAccessIndex past blockedPrefix so later conflicts are detected in the same walk, and rerun each conflicting tx against the freshest prefix you have for it.

giga/evmonly/occ.go
```diff
@@ -0,0 +1,934 @@
...
+func (e *Executor) validateBlockSTMRound(
+	ctx context.Context,
+	runner occSpeculativeRunner,
+	results []occTxExecution,
+	incarnations []int,
+	sourcePrefixes []int,
+	validation *occValidationResult,
+) (StateChangeSet, []occExecutionTask, error) {
+	prefix := newBlockSTMState(e.state)
+	writes := newStateAccessIndex()
```
masih: A fresh prefix+index per round, full O(n) walk re-applying every validated changeset and re-inserting every write key. All of this happens in a single gorutine. that boils down to millons of serial map operations before we hit 3 digit tps throughput. it then is multiplied by rounds.

I worry that this will exhaust the executor very quickly specially given [this](https://github.com/sei-protocol/sei-chain/pull/3583/changes#r3683168556) comment.

`blockedPrefix` never moves backwards between rounds, right? we can use that to significantly reduce the code's resource footprint: keep prefix, writes, and cumulativeGasUsed alive across rounds and resume the walk from the previous blockedPrefix _instead_ of index 0. This means the cost drops from O(rounds × n) to O(n + rerun tails).

giga/evmonly/state.go
```diff
@@ -0,0 +1,236 @@
...
+func (s *lockedStateReader) ConcurrentReads() {}
+
+func parallelSafeStateReader(source StateReader) StateReader {
+	if source == nil {
+		return NewMemoryState()
+	}
+	if _, ok := source.(ConcurrentStateReader); ok {
+		return source
+	}
+	return &lockedStateReader{source: source}
```
masih: That seems expensive: a single mutex around get* everything plus extra clones. this means every OCC worker could block on cold read for example.

The issue i see with this apprach is that it's silent: I can see how it could be surprising where everything suddenly slows down and it's not obvious why.

It is reasonably straightforward to be a bit more forgiving here with big gains; for example, if OCCWorkers > 1 and the reader isn't a ConcurrentStateReader, either refuse in NewExecutor or have useOCC return false and log loudly. Keep lockedStateReader as an explicit test-only wrapper.

giga/evmonly/occ.go
```diff
@@ -0,0 +1,934 @@
...
+			nextToValidate = blockedPrefix
+		}
+		needsRerun, err := needsSTMRerun(validation, writes, result, cumulativeGasUsed, runner.blockGasLimit, sourcePrefix, txIndex, nextToValidate)
+		if err != nil {
+			return StateChangeSet{}, nil, err
+		}
+		if needsRerun {
+			if blockedPrefix < 0 {
+				blockedPrefix = txIndex
+				blockedSource = prefix.clone()
```
masih: Once blockedPrefix is set, prefix.apply is never called again that round. IIUC, this means prefix.clone() deep-copies balances, nonces, and full contract bytecode to protect something that's already immutable. 

why?

giga/evmonly/state_db.go
```diff
@@ -0,0 +1,1330 @@
...
+	s.recordAccount(addr)
+	s.markWrite(stateAccessKey{kind: stateAccessNonce, address: addr})
+	acct.Nonce = nonce
+}
+
+func (s *nativeStateDB) GetCodeHash(addr common.Address) common.Hash {
+	s.markRead(stateAccessKey{kind: stateAccessCode, address: addr})
+	acct := s.account(addr)
+	if len(acct.Code) > 0 {
+		return crypto.Keccak256Hash(acct.Code)
```
masih: this rehashes on every call and is expensive. IIRC geth calls it per tx.

this seems reasonably straightforward to cache ? 

Similar story for `GetCode`.

giga/evmonly/state_db.go
```diff
@@ -0,0 +1,1330 @@
...
+	}
+	snapshot := &s.snapshots[len(s.snapshots)-1]
+	if _, ok := snapshot.journaledAddrs[addr]; ok {
+		return
+	}
+	snapshot.journaledAddrs[addr] = struct{}{}
+	s.journal = append(s.journal, nativeJournalEntry{
+		kind:    nativeJournalAccount,
+		address: addr,
+		account: s.account(addr).clone(),
```
masih: Thanks for fixing the snapshot complexity; this clone still hurts deep call trees on hot contracts. we can further optimise here but please feel free to take a note and defer extra work on this one.

giga/evmonly/result_pool.go
```diff
@@ -0,0 +1,143 @@
...
+		return nil
+	}
+	p := &blockResultPool{free: make(chan *BlockResult, size)}
+	for range size {
+		p.free <- &BlockResult{}
+	}
+	return p
+}
+
+func (p *blockResultPool) acquire(ctx context.Context, txCapacity int) (*BlockResult, error) {
```
masih, line 33: IIUC, this could block forever: If a ResultSink ever forgets its release(), acquire waits on <-p.free with what is usually context.Background() ...and the chain stops, silently. 

For safety I would fall back to allocating a fresh BlockResult after abounded wait, plus a metrics etc. we also need to be careful with the context handling which depends on other pending PRs.

giga/evmonly/occ.go
```diff
@@ -0,0 +1,934 @@
...
+		result.txResult.CumulativeGasUsed = blockResult.GasUsed
+		result.receipt.CumulativeGasUsed = blockResult.GasUsed
+		for _, log := range result.receipt.Logs {
+			log.Index = logIndex
+			logIndex++
+		}
+		blockResult.Txs[i] = result.txResult
+		blockResult.Receipts[i] = result.receipt
+	}
+	changeSet.cloneInto(&blockResult.ChangeSet)
```
masih: We "own" blockresult in this closure; why the "clone into"? That seems superfluous?

masih (commented):
No benchmarks yet, the tests with conflicts test with small number of txs and would love to see some metrics on fundamentals: rerun, time spent in lock, sequential fallback etc.

3 potential blockers: validateBlockSTMRound, validateBlockSTMRound footprint, lockedStateReader mutex surprise. 

Thank you for reflecting on my previous comments Tony; this is taking shape nicely! 🏎️

giga/evmonly/occ.go
codchen: Made an update such that incarnation is attached to each transaction result and incremented only when that particular transaction reruns. That way the frontier never moves backward:
- Conflict-free transaction: validated once.
- Stale/conflicting transaction: generally validated twice, once before and once after its rerun.
- A fully dependent chain of n transactions therefore performs approximately 2n-1 validations, rather than repeated full-block rounds.
I didn't make every balance addition commutative b/c arbitrary balance credits are not safely commutative when later transactions can read or spend the credited balance.

Loadtest results on my laptop:

Scenario | TPS | Gas/s
-- | -- | -- 
Native transfer, unique recipients | 193,456 | 4.06B 
Native transfer, hot recipient | 64,721 | 1.36B 
Native transfer, 10% paired recipient conflicts | 164,272 | 3.45B
Native transfer, same sender nonce chain | 71,059 | 1.49B 
ERC20 transfer, unique recipients | 186,749 | 8.59B
ERC20 transfer, 10% paired recipient conflicts | 146,288 | 6.60B 
Snapshot/revert contract call | 195,186 | 8.98B

giga/evmonly/occ.go
codchen: see response above but tldr the O(n) walk is removed

giga/evmonly/state.go
codchen: removed the mutex and assume the state reader must be concurrent inherently

giga/evmonly/occ.go
codchen: removed the cloning

giga/evmonly/state_db.go
codchen: Added code-hash caching and immutable bytecode sharing.

giga/evmonly/state_db.go
codchen: added nested snapshotting + related testing

giga/evmonly/occ.go
codchen: removed the cloning

giga/evmonly/occ.go
codchen: the loadtest is done with branch `https://github.com/sei-protocol/sei-chain/pull/3658` with `--prebuild-blocks --blocks=20 --txs-per-block=1000 --executor-workers=12 --gas-price-wei=0 --min-gas-price-wei=0 --metrics-addr= --report-interval=0`

giga/evmonly/result_pool.go
codchen: changed receive to be non-blocking: if a pooled result is available, return it; otherwise allocate an unpooled (meaning it has no lease) result

giga/evmonly/occ.go
masih: Great, looking much better thanks Tony!

giga/evmonly/executor.go
```diff
@@ -0,0 +1,458 @@
...
+
+func WithResultSink(sink ResultSink) Option {
+	return func(e *Executor) {
+		e.resultSink = sink
+	}
+}
+
+// NewExecutor constructs an EVM-only executor. Call Close to disable future OCC
+// execution on this executor.
+func NewExecutor(cfg Config, opts ...Option) *Executor {
```
masih, line 49: non-blocker: from interface design perspective, it seems slightly odd to take bot Config and Options. I understand this is likely because of compatibility with the current config ecosystem in the repo but later on we should clean this up.

giga/evmonly/executor.go
```diff
@@ -0,0 +1,458 @@
...
+	release := result.retain()
+	if err := e.resultSink.StoreBlockResult(ctx, height, result, release); err != nil {
+		release()
+		return fmt.Errorf("store block result for block %d: %w", height, err)
+	}
+	return nil
+}
+
+func (e *Executor) useOCC(txCount int) bool {
+	if e.closed.Load() || e.cfg.OCCWorkers <= 1 || txCount <= 1 {
```
masih, line 156: This is still unsafe for concurrent call as AI review comments below pointed out. Do we care? in that, is this function expected to be safe for concurrent calls @codchen?

giga/evmonly/executor.go
```diff
@@ -0,0 +1,445 @@
...
+	chainConfig := e.chainConfig(req.Context)
+
+	stateDB := newNativeStateDB(e.state)
+	blockCtx := buildBlockContext(req.Context)
+	evm := vm.NewEVM(blockCtx, stateDB, chainConfig, vm.Config{}, customPrecompileMap(e.cfg.CustomPrecompiles))
+	stateDB.SetEVM(evm)
+
+	gasLimit := req.Context.GasLimit
+	if gasLimit == 0 {
+		gasLimit = math.MaxUint64
```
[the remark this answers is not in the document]
masih: yup this is a good nit.

giga/evmonly/executor_test.go
```diff
@@ -0,0 +1,2596 @@
...
+
+func TestExecutorRequiresBlobBaseFeeAfterCancun(t *testing.T) {
+	chainID := big.NewInt(713715)
+	key, err := crypto.GenerateKey()
+	require.NoError(t, err)
+	sender := crypto.PubkeyToAddress(key.PublicKey)
+	recipient := testAddress(0xb6)
+
+	state := NewMemoryState()
+	state.SetBalance(sender, big.NewInt(200_000_000_000_000))
```
masih: non-blocker: extracting some of the repeated constants would make test code easier to read.

giga/evmonly/executor_test.go
```diff
@@ -0,0 +1,2596 @@
...
+	rawTx := signLegacyTx(t, key, chainID, 0, &recipient, big.NewInt(1), nil)
+	ctx := blockContext(chainID)
+	ctx.BlobBaseFee = nil
+
+	result, err := NewExecutor(Config{}, WithState(state)).ExecuteBlock(context.Background(), BlockRequest{
```
masih: Can you do a quick search replace across all the tests in this PR, and use `t.Context()` instead please?

giga/evmonly/occ.go
```diff
@@ -0,0 +1,889 @@
...
+	}
+	targetChunks := workers * 8
+	chunkSize := (txCount + targetChunks - 1) / targetChunks
+	if chunkSize < 1 {
+		return 1
+	}
+	if chunkSize > 256 {
+		return 256
+	}
+	return chunkSize
```
masih: Use built-in `min` and `max` functions to simplify the code here?

Ditto in other places.

Example:

```suggestion
	return min(max(chunkSize, 1), 256)
```

giga/evmonly/parser.go
```diff
@@ -0,0 +1,84 @@
...
+
+	parseErrs := make([]error, len(txs))
+	g, groupCtx := errgroup.WithContext(ctx)
+	g.SetLimit(workers)
+	for i, raw := range txs {
+		i, raw := i, raw
+		g.Go(func() error {
+			if err := groupCtx.Err(); err != nil {
+				return err
+			}
```
masih, line 41: This context error check is redundant. errgorup does that for you automatically.

Ditto in other places.

giga/evmonly/parser.go
```diff
@@ -0,0 +1,84 @@
...
+			prepared, err := parsePreparedTx(raw, signer)
+			if err != nil {
+				parseErrs[i] = err
```
masih: Hm; what's happening here is that we have a perfectly good primitive (errgroup) that allows us to manage parallelism in conjunction with error handling and context as a cherry on top.

What we want is to stop the show if there is an error, as you have programmed below, in that: when any parallel processing results in error, do not bother proceeding.

What we are doing instead is:
* save any errors manually in a slice.
* return nil in errgroup
* then iterate over all the saved errors and return error as soon as you see one.

What this does is that it ignores the features of that nice primitive of ours (errgroup) outright: it executes all and every task to the end regardless of whether any of them have failed, then throws away the results if any one of them have failed.

What am I missing?

If nothing, then you can simply return `fmt.Errorf("parse tx %d: %w", i, err)` etc. here and get rid of the parseErrs slice entirely. This would give you the added performance boost of not wasting compute on a group of work that are tightly connected, where error in one work unit renders any remaining work in the group redundant.

giga/evmonly/occ.go
```diff
@@ -0,0 +1,889 @@
...
+	for _, change := range result.changeSet.Balances {
+		delta := result.commutativeBalanceDeltas[change.Address]
+		_, normalWrite := result.writeSet[stateAccessKey{kind: stateAccessBalance, address: change.Address}]
+		if delta != nil && !normalWrite {
+			balance := s.GetBalance(change.Address)
+			balance.Add(balance, delta)
+			s.balances[change.Address] = balance
+			continue
+		}
+		s.balances[change.Address] = cloneBig(change.Balance)
```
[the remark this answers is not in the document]
masih: This one is worth a look.

giga/evmonly/result_pool.go
```diff
@@ -0,0 +1,161 @@
...
+	if p == nil {
+		result := &BlockResult{}
+		result.prepareForBlock(txCapacity)
+		return result, nil
+	}
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
+	select {
+	case result := <-p.free:
```
masih, line 43: Does `p.free` channel ever gets closed? if yes, we should check that it is indeed selected because it was closed in which case result is nil or there is a genuine result.

giga/evmonly/result_pool.go
```diff
@@ -0,0 +1,161 @@
...
+func (p *blockResultPool) acquire(ctx context.Context, txCapacity int) (*BlockResult, error) {
+	if p == nil {
+		result := &BlockResult{}
+		result.prepareForBlock(txCapacity)
+		return result, nil
+	}
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
+	select {
```
masih, line 42: This should select on `ctx.Done` instead of err check above.

giga/evmonly/state_db.go
```diff
@@ -0,0 +1,1412 @@
...
+	s.markWrite(stateAccessKey{kind: stateAccessBalance, address: addr})
+	acct.Balance.Sub(acct.Balance, amount)
+	return prev
+}
+
+func (s *nativeStateDB) AddBalance(addr common.Address, amount *uint256.Int, reason tracing.BalanceChangeReason) uint256.Int {
+	if reason == tracing.BalanceIncreaseRewardTransactionFee {
+		return s.addCommutativeBalance(addr, amount)
+	}
+	prev := *s.GetBalance(addr)
```
masih, line 292: This does a markRead, which means if we have n transfers to one deposite address we would do n-1 reruns (excluding coninbase-fee case)

I made a couple of suggestions in the previous round of reviews to ease this: either use the existing delta mechanism or group tx executions by senders and execute sequentially to avoid this specific expensive (and likely common) scenario.

Non-blocker at this stage.

masih (commented):
One blocker on error group usage and a bunch of suggestions that's worth taking a look; they are mostly easy fixes.

Thanks tony!

giga/evmonly/executor.go
codchen: i think the ai review is outdated now since `Close` either waits for admitted work or causes sequential fallback which makes it impossible for a flag flip to slip in between useOcc call and any meaningful `Run` call

giga/evmonly/parser.go
codchen: good point. Returning `fmt.Errorf("parse tx %d: %w", i, err)` here instead and removed parseErrs

giga/evmonly/occ.go
```diff
@@ -0,0 +1,938 @@
...
```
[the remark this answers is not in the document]
codchen: Cloned source balances before applying commutative deltas

giga/evmonly/result_pool.go
codchen: The result-pool channel never closes because results may be released after executor shutdown.

giga/evmonly/result_pool.go
codchen: We want result acquisition to be nonblocking, so the ctx.Err() precheck gives cancellation priority without a competing-select race

giga/evmonly/executor.go
masih: The issue i think is that `useOCC` itself is unsafe: We may proceed to return true when executor is closed, because  `useOCC` and `Close` are not atomic.

I see later in the pool we have check for closed that returns `errOCCWorkerPoolClosed` in this case, which is fine? But pointing this out since this means: in the case above execution would fail _instead_ of falling back to sequential.

masih (approved):

PR #3588, opened 2026-06-15
Tests/evm rpc split tests pr:1

kollegian:


amir-deris:
Issues
  
  1. Binary blob with no provenance (contracts/cw20_base.wasm)

  A compiled WASM binary is committed with no source link, no version pin, and no checksum. This is a supply-chain risk. At minimum add to the README: which release/tag it comes from, its SHA-256, and
  a sha256sum -c step in run-ci.sh or the bootstrap.

  2. Mnemonic in plaintext source (config/endpoints.ts:623–626)

  export const AdminMnemonic = env(
      'SEI_ADMIN_MNEMONIC',
      'cover brand danger absent gas worth ...',  // full mnemonic hardcoded
  );

  For a local devnet this is tolerable, but committing a BIP39 mnemonic to source triggers secret scanners and sets a bad precedent. Consider moving the default to a comment or a .env.example file and
  requiring the env var to be explicitly set.

  3. Stale comment in run-ci.sh (line 12289)

  command -v node >/dev/null 2>&1 || die "node is required (the workflow sets up Node 20)."

  This PR upgrades CI to Node 22 but the comment still says "Node 20".

  4. Hardcoded sleep(500) in bootstrap (00_bootstrap.spec.ts:574)

  await sleep(500);
  const seiAfter = await seiRpc().getBlockNumber();

  This is fragile on loaded CI runners. Should use waitUntil (already present in chainUtils.ts) to poll until block height advances, rather than a fixed sleep.

  5. Pool size is tight (POOL_SIZE = 96)
  
  With 8 spec files each claiming 2–4 accounts, the pool is already around 20–25% consumed. More specs land and this silently exhausts. The error message when exhausted is clear, but it's worth bumping
  to 128 or 192 now rather than at the first failure.

  6. PR description is empty

  Both sections ("Describe your changes" and "Testing performed") are blank for a 13k-line PR. This makes triage, revert decisions, and future blame harder.

amir-deris (approved):

masih (approved):

PR #3591, opened 2026-06-15
Propogate on chain consensus params to tendermint params

philipsu522:

  Fix propagation of on-chain consensus parameter updates into Tendermint consensus state.

  Governance proposals can update `baseapp` consensus params such as `TimeoutParams`, but the app was building `ResponseFinalizeBlock.ConsensusParamUpdates` from the legacy EndBlock response, which only carries `Block`, `Evidence`,
  `Validator`, and `Version`. As a result, params like `Timeout`, `Synchrony`, and `ABCI` could be updated in app state but never applied by Tendermint.

  This change reads the full consensus params from the state selected for commit and sends the complete set in `ResponseFinalizeBlock.ConsensusParamUpdates`.

  - Propagate full consensus params in `FinalizeBlocker`.
  - Add `BaseApp.GetConsensusParamsForStateToCommit()`.
  - Deep-copy all consensus param sections before returning finalize response:
    - `Block`
    - `Evidence`
    - `Validator`
    - `Version`
    - `Synchrony`
    - `Timeout`
    - `ABCI`
  - Add regression coverage for full consensus-param propagation.


app/app.go
```diff
@@ -2455,27 +2463,74 @@ func (app *App) getFinalizeBlockResponse(appHash []byte, events []abci.Event, tx
...
+	}
+	if params.Timeout != nil {
+		cp.Timeout = &tmproto.TimeoutParams{
+			Propose:             cloneDuration(params.Timeout.Propose),
+			ProposeDelta:        cloneDuration(params.Timeout.ProposeDelta),
+			Vote:                cloneDuration(params.Timeout.Vote),
+			VoteDelta:           cloneDuration(params.Timeout.VoteDelta),
+			Commit:              cloneDuration(params.Timeout.Commit),
+			BypassCommitTimeout: params.Timeout.BypassCommitTimeout,
+		}
```
philipsu522, line 2516: these need to be committed to tm state from app state

sei-will (approved):

wen-coding (approved):

pompon0 (approved):

PR #3596, opened 2026-06-16
Tests: evm rpc tests phase 2

kollegian:
Adding rpc tests for the eth_rpc endpoints.
The changes are only on the test code.


bdchatham (commented):
Ran this through a multi-specialist cross-review (EVM semantics, Sei RPC contract, test reliability, TS idiom). Overall it's a strong, careful suite — the EVM-semantics encodings are largely exact (intrinsic gas, EIP-2930/7702 surcharges, `effectiveGasPrice`, Bloom-9, the EIP-1559 array shapes all check out), so the inline notes are about a few false-confidence and flakiness gaps, not the EVM math.

Main asks before we lean on this as a contract gate:

- **EVM-only-block assumption** — the block/receipt/feeHistory invariants quietly assume EVM-only blocks, but Sei folds Cosmos + bank txs into `block.gasUsed` and `transactions` (and our own rich-block builder co-locates a CW20 tx when wasm is on). Let's assert the node's inclusion config or scope these to EVM txs so a green run actually proves the Sei contract.
- **Key-set-only parity** — we compare key names but not values, so the hardcoded-zero header fields would pass even with a value-level bug.
- **Vacuous geth allowlist** — `GETH_ONLY_TX_FIELDS=['blockTimestamp']` is dead against modern geth, and we run an unpinned `geth --dev`; let's pin the version.
- **`sharedRichBlock` cache** — it doesn't reset on failure the way `sharedGethTx` does, so one transient build failure cascades the whole suite red (cheap one-line fix).
- A couple **fee-market assertions** assume burst-only traffic, and a known divergence (`blockCount==0` → `null` vs `0x0`) is shipped as `it.skip`.

Nice work overall — these are close-or-accept items, not a rewrite. Details inline. Two non-blocking notes: there's a pervasive `: any` on the RPC surface that reverses phase-1's typed convention (worth typing the high-fanout helpers — flagged one spot), and every fee-market number ultimately rests on `nextBaseFeeSei` in `chainUtils.ts`, which is phase-1 code outside this PR and worth confirming against the `x/evm` keeper.

integration_test/rpc_tests/utils/txUtils.ts
```diff
@@ -0,0 +1,1186 @@
...
+    return Promise.all(hashes.map(h => provider.getTransactionReceipt(h)));
+}
+
+/**
+ * Verify the block's gas accounting: the block's gasUsed equals the sum of every
+ * listed transaction's receipt gasUsed, that per-receipt gasUsed is positive, that
+ * cumulativeGasUsed rises monotonically with transaction index, and that the final
+ * cumulativeGasUsed equals the block's gasUsed. Robust on both chains (pure gas units).
+ */
+export async function assertGasAccounting(
```
bdchatham, line 861: This assumes EVM-only blocks, but Sei folds Cosmos `MsgExecuteContract` + bank `MsgSend` into both `block.gasUsed` and `transactions`, so the final `cumulativeGasUsed == block.gasUsed` check breaks the moment a non-EVM tx shares the block. Our own rich-block builder co-locates a CW20 tx when wasm is enabled. Can we assert the node's inclusion config is off, or scope this to EVM receipts?

integration_test/rpc_tests/eth/eth_getBlockByNumber.spec.ts
```diff
@@ -0,0 +1,548 @@
...
+            assertCanonicalHeader(g, { hasTxs: true });
+
+            const sKeys = Object.keys(s);
+            const gKeys = Object.keys(g);
+            for (const f of CORE_BLOCK_FIELDS) {
+                expect(sKeys, `Sei header has ${f}`).to.include(f);
+                expect(gKeys, `geth header has ${f}`).to.include(f);
+            }
+            const seiExtra = sKeys.filter(k => !gKeys.includes(k));
+            const gethExtra = gKeys.filter(k => !sKeys.includes(k));
```
bdchatham, line 378: This only diffs key *names*, so Sei's hardcoded-zero header fields (`mixHash`/`extraData`/`difficulty`/`nonce`) would pass parity even if a value-level contract bug crept in. Can we diff values for the fields we care about, or at least rename this so it's clear "parity" means key sets only?

integration_test/rpc_tests/utils/txUtils.ts
```diff
@@ -0,0 +1,1186 @@
...
+    'r',
+    's',
+    'to',
+    'transactionIndex',
+    'type',
+    'v',
+    'value',
+    'yParity',
+] as const;
+export const GETH_ONLY_TX_FIELDS = ['blockTimestamp'] as const;
```
bdchatham, line 78: Modern geth doesn't emit `blockTimestamp` and CI runs an unpinned `geth --dev`, so this allowlist filters to empty and the divergence assertion passes without testing anything. Let's pin the geth version and drop/fix this entry — otherwise the whole divergence baseline depends on whatever geth the runner happens to have.

integration_test/rpc_tests/utils/txUtils.ts
```diff
@@ -0,0 +1,1186 @@
...
+let cachedRichBlockPromise: Promise<RichBlock> | undefined;
+
+export async function sharedRichBlock(
+    provider: ethers.JsonRpcProvider,
+    runtime: RuntimeState,
+): Promise<RichBlock> {
+    if (cachedRichBlock) return cachedRichBlock;
+    // Guard against two specs' `before` hooks racing the first build in the same process.
+    if (!cachedRichBlockPromise) {
+        cachedRichBlockPromise = (async () => {
```
bdchatham, line 498: Unlike `sharedGethTx`, this never nulls the cached promise on rejection, so a single transient rich-block build failure gets memoized and cascades into a synchronized red across every spec that awaits it. Mirroring the `sharedGethTx` `.catch` reset here is a cheap, high-leverage fix.

integration_test/rpc_tests/utils/txUtils.ts
```diff
@@ -0,0 +1,1186 @@
...
+    's',
+    'to',
+    'transactionIndex',
+    'type',
+    'v',
+    'value',
+] as const;
+
+/** Assert a full transaction object is canonically encoded and linked to its block. */
+export function assertCanonicalTx(tx: any, block: any): void {
```
bdchatham: The whole assertion API takes `any` for tx/block/receipt, which reverses the phase-1 specs' typed convention. A renamed or typo'd field just reads as `undefined` and fails as an assertion message instead of a compile error, which is exactly the drift these tests exist to catch. Non-blocking, but typing the high-fanout helpers (we already define `SentTx`/`RichBlock`) would pay off.

integration_test/rpc_tests/eth/eth_feeHistory.spec.ts
```diff
@@ -0,0 +1,320 @@
...
+            const peak = fh.baseFeePerGas.reduce((m, v) => (v > m ? v : m), 0n);
+            expect(peak > before, `base fee should rise above ${before}, peaked at ${peak}`).to.equal(
+                true,
+            );
+
+            const rose = fh.baseFeePerGas.some((v, i) => i > 0 && v > fh.baseFeePerGas[i - 1]);
+            expect(rose, 'at least one block raised the base fee').to.equal(true);
+
+            const exactTipBlock = fh.reward!.find(
+                row => row.length === percentiles.length && row.every(r => r === BURST_TIP),
```
bdchatham: This pins the reward to exactly `BURST_TIP` at every percentile assuming the burst is the only traffic in the block, but we run serially against one chain so another spec's tx can land in the same block and shift the percentile → spurious failure. Can we scope this to blocks that only contain our burst senders?

integration_test/rpc_tests/eth/eth_feeHistory.spec.ts
```diff
@@ -0,0 +1,320 @@
...
+            expect(reportedRatio).to.be.closeTo(Number(blk.gasUsed) / Number(blk.gasLimit), 1.01e-4);
+            expect(reportedRatio > targetRatio, 'over-target block exceeds the target ratio').to.equal(
+                true,
+            );
+        });
+    });
+
+    describe('empty / null handling', () => {
+        // SKIP(expected-failure): Sei diverges on blockCount 0 oldestBlock; pending manual reverification.
+        it.skip('[spec] blockCount 0 returns oldestBlock as the canonical quantity 0x0', async () => {
```
bdchatham, line 198: Shipping the `blockCount==0` case as `it.skip` (same at L251) means a real Sei divergence — it returns `null` vs the canonical `0x0` — goes untested while the suite stays green. Let's assert Sei's actual behavior here, or track it as an issue, rather than skip it.

integration_test/rpc_tests/eth/eth_getBlockReceipts.spec.ts
```diff
@@ -0,0 +1,719 @@
...
+                (a, b) => Number(BigInt(a.transactionIndex)) - Number(BigInt(b.transactionIndex)),
+            );
+            let running = 0n;
+            for (const rc of ordered) {
+                running += BigInt(rc.gasUsed);
+                expect(BigInt(rc.cumulativeGasUsed), `cumulativeGasUsed at ${rc.transactionIndex}`).to.equal(
+                    running,
+                );
+            }
+            expect(running, 'final cumulativeGasUsed == block.gasUsed').to.equal(BigInt(block.gasUsed));
```
amir-deris: EVM-only block assumption (medium — false confidence)
Gas accounting asserts that EVM receipt gas sums to block.gasUsed:

eth_getBlockReceipts.spec.ts
Lines 388-403
        it('block.gasUsed equals Σ receipt gasUsed and cumulativeGasUsed is consistent', async () => {
            ...
            expect(running, 'final cumulativeGasUsed == block.gasUsed').to.equal(BigInt(block.gasUsed));
        });
assertGasAccounting in txUtils.ts makes the same assumption. But buildRichSeiBlock intentionally co-locates a Cosmos CW20 transfer when wasm is enabled, and Sei can fold non-EVM txs into block.gasUsed / transactions.

CI passes today (wasm may be off or inclusion behaves differently in the cluster), but the test does not prove what it claims when dual-VM traffic is present.

Suggestion: Assert the node’s inclusion config at bootstrap, or scope invariants to EVM receipts only and document the Sei contract explicitly.

integration_test/rpc_tests/eth/eth_getBlockByHash.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+            const [s, g] = await Promise.all([
+                byHash(sei, seiOne.hash, true),
+                byHash(geth, gethOne.hash, true),
+            ]);
+            assertCanonicalHeader(s, { hasTxs: true });
+            assertCanonicalHeader(g, { hasTxs: true });
+
+            const sKeys = Object.keys(s);
+            const gKeys = Object.keys(g);
+            for (const f of CORE_BLOCK_FIELDS) {
```
amir-deris, line 325: Geth parity for blocks/transactions compares field names, not values. Sei’s hardcoded-zero header fields (mixHash, extraData, difficulty, nonce) would pass even with value-level bugs. Rename helpers to assertKeysetParity or add value checks for fields that should match geth semantics.

amir-deris (approved):

PR #3597, opened 2026-06-16
fix(evmrpc): pass raw tx bytes to DeliverTx in BlockByNumber TraceRunnable (CON-326)

wen-coding:
## Summary

- `BlockByNumber` replays non-traced (Cosmos) transactions via a `TraceRunnable` closure but was passing `abci.RequestDeliverTxV2{}` (empty) to `DeliverTx`
- `BaseApp.DeliverTx` sets `ctx.TxBytes = req.Tx` before calling the ante handler; with an empty `req.Tx`, `ctx.TxBytes()` is nil
- The Cosmos ante handler charges `TxSizeCostPerByte * len(ctx.TxBytes())` gas — nil bytes means 0 gas charged, diverging from actual on-chain execution
- Fix: capture `txBytes := tmBlock.Block.Txs[i]` before the closure (also making the loop variable capture explicit and version-independent) and pass `RequestDeliverTxV2{Tx: txBytes}`
- `replayTransactionTillIndex` already did this correctly; `BlockByNumber` was the only callsite missing it

## Test plan

- [ ] `TestBlockByNumberNonTracedTxPassesTxBytes` — new test using a minimal `BaseApp` with a spy ante handler that captures `ctx.TxBytes()`; asserts it equals the raw tx bytes from the block. Verified it fails without the fix and passes with it.


wen-coding:
Closing — PR was opened against wrong base. Recreating against correct base branch.

PR #3599, opened 2026-06-16
Deflake FlatKV EVM migrate halt coordination

masih:
The pre-flip halt-height stop assumed every validator would stop cleanly at the same height. In practice some stop at the halt height while others stay live behind them, leaving the cluster in a split state that's unsafe to flip.

Add retry coordination to recover from this. When a partial halt is detected, keep the cluster in `memiavl_only` rather than flipping `sc-write-mode`, stop any validators still live, schedule a fresh future halt height from the committed and last-sign state, and restart until every validator is stopped at a safe common boundary.

Also add a repo-local `go-lint-setup` action for the reusable UCI Go lint workflow. It reclaims disk on `ubuntu-latest` before `go mod tidy`, `go vet`, formatting, and `go generate`.


masih:
@claude /review

philipsu522 (approved):

bdchatham (approved):

PR #3601, opened 2026-06-16
conditional deactivation of consensus timeout overrides in config

pompon0:
In 6.6 release we want to disable the overrides support.
Overrides can only be disabled AFTER onchain params are updated via government proposal.
Govenment proposal needs to wait for changes deployed in 6.6.
Disabling overrides needs to be synchronized across the network.
We need ability to reenable overrides for disaster recovery (chain halt).

To achieve that we:
* add a unsafeOverridesEnabled flag
* unsafeOverridesEnabled defaults to true BEFORE the government proposal goes through, and to false AFTERWARDS

### DIFF
--- a/sei-tendermint/config/config.go
+++ b/sei-tendermint/config/config.go
@@ -1158,6 +1158,9 @@ type ConsensusConfig struct {
 	// removed in the v0.37 release of Tendermint.
 	// See: https://github.com/tendermint/tendermint/issues/8188
 
+	// If false, all the Unsafe<..>TimeoutOverride fields are ignored.
+	// Defaults to false.
+	UnsafeOverridesEnabled bool `mapstructure:"unsafe-overrides-enabled"`
 	// UnsafeProposeTimeoutOverride provides an unsafe override of the Propose
 	// timeout consensus parameter. It configures how long the consensus engine
 	// will wait to receive a proposal block before prevoting nil.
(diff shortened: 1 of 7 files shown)


sei-tendermint/config/config.go
```diff
@@ -1201,6 +1204,43 @@ type ConsensusConfig struct {
...
+// to correct values via gov proposal. Until then
+// (i.e. while onchain timeout params are still equal to badParams)
+// overrides are still enabled by default.
+var badParams = types.TimeoutParams{
+	Propose:             1000 * time.Millisecond,
+	ProposeDelta:        500 * time.Millisecond,
+	Vote:                50 * time.Millisecond,
+	VoteDelta:           500 * time.Millisecond,
+	Commit:              50 * time.Millisecond,
+	BypassCommitTimeout: false,
```
wen-coding, line 1219: Is this the same as DefaultTimeoutParams?

sei-tendermint/config/config.go
```diff
@@ -1201,6 +1204,43 @@ type ConsensusConfig struct {
...
+}
+
+func (c *ConsensusConfig) ResolveTimeouts(t types.TimeoutParams) types.TimeoutParams {
+	t = t.Or(types.DefaultTimeoutParams())
+	// Overrides are effective iff UnsafeOverridesEnabled OR t == badParams:
+	// see doc on badParams.
+	if !c.UnsafeOverridesEnabled && t != badParams {
```
wen-coding, line 1226: Our gov proposal will always override all params at the same time right? It would be a headache if proposal updates some args while override updates some.

sei-will (approved):

sei-tendermint/config/config.go
pompon0: proposal updates everything atomically, yes. It wouldn't make sense otherwise.

sei-tendermint/config/config.go
pompon0: huh, indeed. Perhaps timeouts are currently not set at all.

wen-coding (approved):

sei-tendermint/config/config.go
philipsu522: correct, all happens within a block

philipsu522 (approved):

sei-tendermint/config/config.go
```diff
@@ -1201,6 +1204,43 @@ type ConsensusConfig struct {
...
+	ProposeDelta:        500 * time.Millisecond,
+	Vote:                50 * time.Millisecond,
+	VoteDelta:           500 * time.Millisecond,
+	Commit:              50 * time.Millisecond,
+	BypassCommitTimeout: false,
+}
+
+func (c *ConsensusConfig) ResolveTimeouts(t types.TimeoutParams) types.TimeoutParams {
+	t = t.Or(types.DefaultTimeoutParams())
+	// Overrides are effective iff UnsafeOverridesEnabled OR t == badParams:
```
masih: Would be nice to not have to apply De Morgan complement when reading the code relative to this inline comment.

sei-tendermint/config/config.go
```diff
@@ -1201,6 +1204,43 @@ type ConsensusConfig struct {
...
 }
 
+// Timeout params on Sei pacific-1, as of 2026-06-16.
+// Overrides will be disabled by default in release 6.6,
+// but only after the onchain timeout params are set
+// to correct values via gov proposal. Until then
+// (i.e. while onchain timeout params are still equal to badParams)
+// overrides are still enabled by default.
+var badParams = types.TimeoutParams{
+	Propose:             1000 * time.Millisecond,
```
masih: Just to match the `DefaultTimeoutParams`.
```suggestion
	Propose:             1 * time.Second,
```

sei-tendermint/config/config.go
masih: hm..

masih (approved):

PR #3602, opened 2026-06-16
Autobahn hashvault integration

cody-littley:

Wire in the hash vault to autobahn. This prevents the app hash from changing for a particular block without human intervention. This does not impact cosmos, and there is no intention to add this to cosmos.


sei-tendermint/internal/consensus/replay.go
```diff
@@ -401,7 +405,16 @@ func (h *Handshaker) replayBlocks(
...
+			blockExec := sm.NewBlockExecutor(
+				h.stateStore,
+				app,
+				newReplayTxMempool(app),
+				sm.EmptyEvidencePool{},
+				h.store,
+				h.eventBus,
+				sm.NopMetrics(),
+				h.consensusPolicy,
+				h.hashVault)
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-tendermint/internal/state/execution.go
```diff
@@ -421,6 +449,11 @@ func (blockExec *BlockExecutor) ApplyBlock(ctx context.Context, state State, blo
 		} else {
 			logger.Debug("pruned blocks", "pruned", pruned, "retain_height", retainHeight)
 		}
+		// Keep the hash vault's retention aligned with block retention (which already accounts for
+		// state-snapshot retention). A prune failure is GC, not equivocation: log and continue.
+		if err := blockExec.hashVault.Prune(ctx, uint64(retainHeight)); err != nil { //nolint:gosec // retainHeight > 0 here
+			logger.Error("failed to prune hashvault", "retain_height", retainHeight, "err", err)
+		}
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-tendermint/internal/state/execution.go
```diff
@@ -400,6 +409,13 @@ func (blockExec *BlockExecutor) ApplyBlock(ctx context.Context, state State, blo
 	}
 	saveBlockTime := time.Now()
 	state.AppHash = fBlockRes.AppHash
+
+	// Commit this block's hash to the equivocation guard before saving state. See commitHashToVault.
+	// A returned error is a benign shutdown cancellation; genuine faults panic inside the call.
+	if err := commitHashToVault(ctx, blockExec.hashVault, block.Height, block.Hash()); err != nil {
```
wen-coding: I think the intent is to vault AppHash? block.Hash() is the Tendermint block header hash.

I think you should do:
```
if err := commitHashToVault(ctx, blockExec.hashVault, block.Height, state.AppHash); err != nil {
    return state, err
}
```
instead.

sei-tendermint/internal/consensus/replay.go
```diff
@@ -401,15 +405,24 @@ func (h *Handshaker) replayBlocks(
...
```
wen-coding: This only covers non-Autobahn part. Is the intent to do Autobahn part later? I think you would need to:

1. GigaRouter struct — add the vault field:

```
type GigaRouter struct {
    // ... existing fields ...
    hashVault hashvault.HashVault
}
```
2. NewGigaRouter — accept it as a parameter and set the field.

3. executeBlock — vault `resp.AppHash` (the FinalizeBlock output, not the Autobahn header hash) right after FinalizeBlock succeeds, before Commit:

```
resp, err := app.FinalizeBlock(ctx, ...)
if err != nil { ... }

// Seal the computed AppHash before commit — same equivocation guard as the Cosmos path.
if err := commitHashToVault(ctx, r.hashVault, int64(b.GlobalNumber), resp.AppHash); err != nil {
    return nil, err
}

commitResp, err := app.Commit(ctx)
```
4. `buildGigaRouter` in setup.go — thread the vault from makeNode through to NewGigaRouter.

sei-tendermint/internal/state/execution.go
```diff
@@ -697,6 +723,41 @@ func FireEvents(
...
+	if err == nil {
+		return nil
+	}
+	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+		logger.Info("HashVault commit aborted by context cancellation during shutdown; not recording hash",
+			"height", height, "err", err)
+		return fmt.Errorf("hashvault CommitToHash aborted at height %d: %w", height, err)
+	}
+	if errors.Is(err, hashvault.ErrHashMismatch) {
+		logger.Error("FATAL: HashVault detected a block-hash mismatch — the node has equivocated. "+
```
wen-coding: Chatted with Greg today, since re-execution of a block should only happen during restart, it's fine to just panic here. I think you can print previous and current hash, then hint if the human is really really sure, he can remove <insert path> directory to proceed, but warn that this may lead to slashing etc

sei-tendermint/internal/state/execution.go
```diff
@@ -697,6 +723,41 @@ func FireEvents(
...
+			"Halting. DO NOT RESTART WITHOUT HUMAN INTERVENTION.",
+			"height", height, "hash", fmt.Sprintf("%X", hash), "err", err)
+	} else {
+		logger.Error("FATAL: HashVault could not commit the block hash (operational error, not a "+
```
wen-coding: Could a transient I/O error land here?

What's the behavior if the on-disk file is corrupted? If the on-disk file is corrupted maybe we should just ask human to remove or just proceed? I just don't want one corrupted bit on disk to take down the whole validator. What do you feel, log an error and proceed, but don't overwrite the corrupted entry?

sei-tendermint/internal/state/execution.go
cody-littley: change made

sei-tendermint/internal/state/execution.go
cody-littley: Yes, and this is by design. If we can't read and write to the hashvault, we have no idea if the hash we are providing will cause slashing. The only way we can be 100% confident is if the hash vault is able to read/write the file system and we find no conflict in the data.

The challenge with random file system errors is that there isn't a clear way to recover from them automatically. If the files are corrupted, the correct solution is probably going to be for a human operator to delete the hashvault archive. But if the error is from a full disk or bad file permissions, the recovery pathway will look very different.

Let me know if you'd like to discuss.

sei-tendermint/internal/state/execution.go
cody-littley: Side note, this file is now reverted. We decided not to integrate the hashvault into cosmos pathways.

sei-tendermint/internal/consensus/replay.go
cody-littley: Good point, I think I did this backwards. I've reverted the cosmos changes and re-integrated with autobanhn.

sei-tendermint/internal/state/execution.go
cody-littley: Code now logs both hashes, as well as instructions about how to bypass the problem.

sei-db/state_db/sc/hashvault/pebble_hashvault.go
```diff
@@ -217,9 +217,14 @@ func (p *PebbleHashVault) Close(_ context.Context) error {
 
 func (p *PebbleHashVault) logHashMismatch(blockHeight uint64, existing, incoming []byte) {
 	logger.Error("Hashvault detected hash mismatch; node attempted to change its mind. "+
```
wen-coding: nit: make it clear this is AppHash mismatch

sei-tendermint/config/toml.go
```diff
@@ -157,6 +157,16 @@ node-key-file = "{{ js .BaseConfig.NodeKey }}"
 # would otherwise nest it under the immediately preceding section.
 autobahn-config-file = "{{ .AutobahnConfigFile }}"
 
+# hash-vault-disabled-unsafe disables the block-hash equivocation guard (HashVault).
+# DO NOT set this to true unless you are knowingly running an UNSAFE node as a last-resort
+# recovery measure. A node with this enabled has NO protection against changing its mind about
+# a committed block's hash, and will log error-level warnings on every startup.
```
wen-coding: nit: we can probably tell them something like:
It's safer to not disable hash vault, just remove hash vault files as instructed and let it run. They should only disable hash vault if they are very sure the hash vault is totally wrong and they will keep running into the same panic on new blocks.

sei-tendermint/config/toml.go
```diff
@@ -157,6 +157,16 @@ node-key-file = "{{ js .BaseConfig.NodeKey }}"
...
+#
+# Placed here (as a top-level key, before any [section] header) so the TOML parser sees it at
+# root scope where mapstructure expects it — viper would otherwise nest it under the
+# immediately preceding section.
+hash-vault-disabled-unsafe = {{ .HashVaultDisabledUnsafe }}
```
wen-coding, line 172: what is "Advanced Configuration Options"? does this belong there? (I don't know what are Advanced Configuration Options either, this is a genuine question, not suggestion.)

sei-tendermint/internal/p2p/giga_router_common.go
```diff
@@ -46,6 +48,12 @@ type gigaRouterCommon struct {
 	// by one or two under contention but never over-accepts.
 	inboundFullnodeCount atomic.Int64
 	inboundFullnodeCap   int64
+
+	// hashVault is the block-hash equivocation guard. runExecute owns its lifecycle: it builds the
```
wen-coding: nit: AppHash, we don't check blockHash

sei-tendermint/internal/p2p/giga_router_common.go
```diff
@@ -46,6 +48,12 @@ type gigaRouterCommon struct {
...
+	// vault on entry (durable under PersistentStateDir, or a no-op when disabled / in-memory) and
+	// closes it on exit. Set before executeBlock is ever called, which is the only other reader. See
```
wen-coding: wait, are we setting it before executeBlock is ever called?

sei-tendermint/internal/p2p/giga_router_common.go
```diff
@@ -238,6 +246,16 @@ func (r *gigaRouterCommon) executeBlock(ctx context.Context, b *atypes.GlobalBlo
 	if err != nil {
 		return nil, fmt.Errorf("app.FinalizeBlock(): %w", err)
 	}
+
+	// Commit this height's app hash to the equivocation guard before persisting app state, so the
+	// vault always records our commitment to a height before the state it implies is committed (and
+	// before the hash is proposed for AppQC voting via PushAppHash below). On restart the block is
+	// re-executed and the identical hash is re-committed idempotently. A returned error is a benign
+	// shutdown cancellation; genuine faults panic inside the call. See commitHashToVault.
+	if err := commitHashToVault(ctx, r.hashVault, b.GlobalNumber, resp.AppHash); err != nil {
```
wen-coding: nit: I'm thinking commitAppHashToVault may make the intent more obvious we never want to commit other Hash.

sei-tendermint/internal/p2p/giga_router_common.go
```diff
@@ -248,7 +266,82 @@ func (r *gigaRouterCommon) executeBlock(ctx context.Context, b *atypes.GlobalBlo
 	return commitResp, nil
 }
 
+// buildHashVault constructs the block-hash equivocation guard runExecute owns. By default it
```
wen-coding: nit: AppHash, maybe find and replace?

sei-tendermint/internal/p2p/giga_router_common.go
```diff
@@ -248,7 +266,82 @@ func (r *gigaRouterCommon) executeBlock(ctx context.Context, b *atypes.GlobalBlo
...
+	}
+	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+		logger.Info("HashVault commit aborted by context cancellation during shutdown; not recording hash",
+			"height", height, "err", err)
+		return fmt.Errorf("hashvault CommitToHash aborted at height %d: %w", height, err)
+	}
+	if errors.Is(err, hashvault.ErrHashMismatch) {
+		// The HashVault has already logged the conflicting hashes, its data directory, and the
+		// bypass/slashing guidance immediately before returning this error; don't duplicate it.
+		logger.Error("FATAL: HashVault detected a block-hash equivocation; halting. See the preceding "+
```
wen-coding: There is a tiny risk the logger doesn't flush before panicing, maybe either flush the log buffer or pass the error message to the panic string?

sei-tendermint/internal/p2p/giga_router_common.go
wen-coding: @pompon0 we are now inserting commitHashToVault before commit. I think this is what we said we wanted during the March offsite, but double checking with you.

sei-db/state_db/sc/hashvault/pebble_hashvault.go
wen-coding, line 61: Do we have a log.Info if there are no data on the disk? Might be useful if it is a surprise.

sei-tendermint/internal/p2p/giga_router_common.go
pompon0: It is fine for now. Eventually we would like to abstract out the whole FinalizeBlock -> commontHashToVault -> Commit pipeline to a single Application.CommitBlock call (or equivalent), since autobahn does not care about any intermediate states. Then we will need to pipeline it: split CommitBlock, into CommitBlock() and WaitForAppHash() calls (and perhaps move them from Application to BlockDB API).

sei-db/state_db/sc/hashvault/pebble_hashvault.go
cody-littley: done

sei-tendermint/config/toml.go
cody-littley:
```
# It is safer to leave HashVault enabled: if you hit a startup panic, first remove the HashVault
# files as instructed in the panic message and let the node run. Only disable HashVault if you are
# very sure the stored hashes are totally wrong and you keep hitting the same panic on new blocks.
```

sei-tendermint/config/toml.go
cody-littley: I don't know the initial intent of the Advanced section. But when I see this, my interpretation is immediately `do not touch these config options unless you know what you are doing`. Calling something advanced config is a fairly standard way of describing this sort of config.

Are you ok with me leaving the flag to disable the hash vault here, or would you like me to move it?

wen-coding: Sounds like we should move this arg into Advanced section then... But I'm fine either way, users have been warned.

sei-tendermint/internal/p2p/giga_router_common.go
cody-littley: fixed

sei-tendermint/internal/p2p/giga_router_common.go
cody-littley: This comment is misleading. We do set it before we execute when we are starting up and loading a state from disk, to ensure the hash didn't somehow change when we restarted. But during normal operation, we can't set the hash for block X before executing block X, because the hash is a function of the execution.

I've cleaned up this godoc so it's less confusing.

sei-tendermint/internal/p2p/giga_router_common.go
cody-littley: done

sei-tendermint/internal/p2p/giga_router_common.go
cody-littley: done

sei-tendermint/internal/p2p/giga_router_common.go
cody-littley: The only way to do a guaranteed flush is a `seilog.Close()`. To compensate, I've replicated this information into the panic message.

sei-db/state_db/sc/hashvault/pebble_hashvault.go
cody-littley: added

wen-coding (approved):

PR #3603, opened 2026-06-17
[codex] Fix Giga V2 fallback cache flush

codchen:
## Summary

- Add a `deliverTxWithV2Fallback` bridge that flushes Giga block-level cache state before running the fallback transaction through V2, then flushes V2 block-level state back for subsequent Giga execution.
- Route both synchronous Giga fallback paths through the shared bridge.
- Add a focused regression test proving V2 sees prior Giga state at the fallback boundary, plus a higher-level three-transaction sender-balance scenario.

## Root Cause

Giga and V2 execution use separate block-level cache paths. The fallback path previously ran the V2 transaction without first publishing already-committed Giga block-level writes, so V2 could observe stale nonce/balance state. After V2 completed, subsequent Giga transactions also needed the V2 block-level writes flushed back to the common base.

## Validation

- `go test ./app -run TestDeliverTxWithV2FallbackFlushesGigaBlockCache -count=1`
- `go test ./giga/tests -run TestGigaSequential_FallbackKeepsCumulativeSenderBalance -count=1 -v`

I also verified the focused app regression fails without the Giga pre-flush: tx2 sees sender nonce `0` instead of `1` and fails with `nonce too high`.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1483,6 +1483,15 @@ func (app *App) DeliverTxWithResult(ctx sdk.Context, tx []byte, typedTx sdk.Tx) 
 	}
 }
 
+func (app *App) deliverTxWithV2Fallback(ctx sdk.Context, ms sdk.CacheMultiStore, tx []byte, typedTx sdk.Tx) *abci.ExecTxResult {
+	// Giga and V2 maintain separate block-level caches. Publish the Giga
+	// side before switching, then publish V2 so later Giga txs read it.
+	ctx.GigaMultiStore().WriteGiga()
+	res := app.DeliverTxWithResult(ctx, tx, typedTx)
+	ms.Write()
+	return res
+}
+
 func (app *App) ProcessTxsSynchronousV2(ctx sdk.Context, txs [][]byte, typedTxs []sdk.Tx) []*abci.ExecTxResult {
 	blockProcessStart := time.Now()
 	defer func() {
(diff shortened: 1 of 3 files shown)


app/app.go
wen-coding, line 1550: Do we need to fix this path as well?

giga/tests/giga_test.go
```diff
@@ -1022,6 +1022,71 @@ func TestGigaVsGeth_SelfDestruct(t *testing.T) {
 	CompareReceipts(t, "SelfDestruct", gethCtx, gethResults, gigaCtx, gigaResults)
 }
 
+func TestGigaSequential_FallbackKeepsCumulativeSenderBalance(t *testing.T) {
```
wen-coding, line 1025: did we test ErrIteratorUnsupported somewhere?

app/app.go
philipsu522: +1

philipsu522, line 1853: does this ordering need to be reversed? WriteGiga() changes won't be flushed to evmCache

codchen: evmCache is for non-EVM/non-bank module and GigaMultiStore is for EVM&bank, so the changes are disjoint and the order shouldn't matter

codchen: good point. Updated

codchen:
The premise of the bug missed the fact that giga and v2 cache share the same parent, so the flush at the end of each successful tx would make changes visible to the next tx, regardless of whether it's giga or v2.

PR #3605, opened 2026-06-17
fix(grpc-web): set ReadTimeout, WriteTimeout, IdleTimeout, and MaxOpenConnections

amir-deris:
## Summary

PR #3558 added `ReadHeaderTimeout` to the gRPC-Web `http.Server`, protecting against slow-header attacks. The body-stall DoS vector remained: a client that sends valid headers and then stalls while the server waits indefinitely for the request body. This PR closes that gap and adds a connection cap to bound total resource usage.

### Timeout changes

| Field | Value | Purpose |
|---|---|---|
| `ReadHeaderTimeout` | 10s | already present — slow header protection |
| `ReadTimeout` | 30s | **primary fix** — drops connections that stall after headers |
| `WriteTimeout` | 2m | backstop against slow-read attacks on the response side |
| `IdleTimeout` | 30s | reclaims idle keep-alive connections |

`WriteTimeout` is set to 2 minutes rather than a shorter value because the longest legitimate call through this server is `BroadcastTxCommit`, which is bounded by Tendermint's `TimeoutBroadcastTxCommit` (default 10s). Paginated query RPCs are already capped at the application layer (PLT-361). The 2-minute window gives a large margin above any real workload while still providing a backstop against response-side connection holds.

### MaxOpenConnections

Wraps the gRPC-Web listener with `netutil.LimitListener` to cap the number of simultaneously open TCP connections. New connections block at the OS accept queue once the limit is reached rather than being accepted and left waiting inside the server.

- Default: **1000** (matches the existing default for the API and gRPC servers)
- Operator-configurable via `grpc-web.max-open-connections` in `app.toml`; set to `0` to disable the cap

**Files changed:** `sei-cosmos/server/grpc/grpc_web.go`, `sei-cosmos/server/config/config.go`

### Tests

`TestStartGRPCWebTimeouts` — starts a real gRPC-Web server on a free loopback port and asserts all four timeout fields on the returned `*http.Server`.

## Test plan

- [ ] `go test ./sei-cosmos/server/grpc/... -run TestStartGRPCWebTimeouts -v` passes
- [ ] `gofmt -s -l .` prints nothing
- [ ] `make build` succeeds


sei-cosmos/server/grpc/grpc_web.go
```diff
@@ -28,6 +28,9 @@ func StartGRPCWeb(grpcSrv *grpc.Server, config config.Config) (*http.Server, err
 		Addr:              config.GRPCWeb.Address,
 		Handler:           wrappedServer,
 		ReadHeaderTimeout: 10 * time.Second,
+		ReadTimeout:       30 * time.Second,
+		WriteTimeout:      30 * time.Second,
```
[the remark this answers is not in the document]
amir-deris: Good catch on the original 30s value. Updated in the latest commit to 2 minutes.

The concern is valid in general — Go's `WriteTimeout` clock starts when request headers are read and includes handler processing time, not just wire-write time. However, for this server the practical risk is negligible: the longest-running call class is `BroadcastTxCommit`, which is already bounded by Tendermint's `TimeoutBroadcastTxCommit` (10s default). All paginated query RPCs have application-level size caps from PLT-361. No legitimate call through this server is expected to approach 2 minutes under any normal or degraded condition.

The 2-minute value gives a large margin above real workload while still providing a backstop against response-side slow-read connection holds.

sei-cosmos/server/grpc/grpc_web.go
```diff
@@ -28,6 +28,9 @@ func StartGRPCWeb(grpcSrv *grpc.Server, config config.Config) (*http.Server, err
 		Addr:              config.GRPCWeb.Address,
 		Handler:           wrappedServer,
 		ReadHeaderTimeout: 10 * time.Second,
+		ReadTimeout:       30 * time.Second,
+		WriteTimeout:      2 * time.Minute,
+		IdleTimeout:       120 * time.Second,
```
bdchatham: Idle timeout strikes me as a bit long. What do you think @sei-will? 

Also, do we have any data on how long standard requests take? Ideally we keep these tightly scoped to what we expect standard workloads to use. For instance, if the largest range query only take 10 seconds based on x years historical data, then we can tighten this to 15 seconds for example. 

We will never get this exactly right so not looking for a perfect magic number, just curious if we have those inputs. My hunch is 2 minutes for a write is more lenient than we need to be

amir-deris: Thanks for feedback. Reduced the idle timeout to 30s.

masih (approved):

PR #3606, opened 2026-06-17
feat(wireguard): per-instance MaxCount + extend to Autobahn fields

wen-coding:
## Summary

Extends the pre-decode wire scanner from #3458 to Autobahn messages, and generalises the wiring so adding protection to a new message type requires only a proto annotation — no per-channel or per-call plumbing.

### Proto annotations

`autobahn.proto` gains `(wireguard.max_count)` on all repeated fields:

| Field | Cap |
|---|---|
| `LaneQC.sigs`, `PrepareQC.sigs`, `CommitQC.sigs`, `AppQC.sigs` | 100 |
| `TimeoutQC.votes` | 100 |
| `FullProposal.lane_qcs`, `Proposal.lane_ranges`, `FullCommitQC.headers` | 100 |
| `Payload.txs` | 2 000 (matches `MaxTxsPerBlock`) |

Caps are checked **per instance** of the containing message, not summed globally. A `FullProposal` with 10 `LaneQC`s each carrying 10 sigs is fine; total memory at any nesting depth is bounded by the product of the caps along the path (`outer_cap × inner_cap × …`). To raise a cap later, update `(wireguard.max_count)` in `autobahn.proto` and re-run:
```bash
GOWORK=off go run github.com/bufbuild/buf/cmd/buf@v1.58.0 generate --template sei-tendermint/internal/buf.gen.yaml
```

### Automatic wiring (main change)

Previously, each channel had to set `PreDecode` explicitly in the reactor. Now the schema is wired directly to the proto type:

- **Plugin** emits `WireguardScan([]byte) error` on each schema-bearing type (instead of `init()`/registry calls).
- **`protoutils.Unmarshal[T]`** asserts `wireguard.Scanner` and scans before `proto.Unmarshal` — zero cost for types without a schema.
- **`transport.go`** asserts `wireguard.Scanner` on the channel `MessageType`, replacing the `PreDecode` field on `ChannelDescriptorT`.
- **`ChannelDescriptorT.PreDecode` removed** — all four explicit assignments (blocksync, consensus, evidence, statesync) dropped; protection follows from the type.

### wireguard package changes

- `wireguard.Scanner` interface exported from the wireguard package (was an unexported duplicate in `protoutils/msg.go`).
- `MaxCount` semantics tightened to per-instance: each nested-message occurrence gets a fresh counter map, so the cap is "at most N per outer element" rather than "N total across the payload".
- Wireguard plugin entry folded into `sei-tendermint/internal/buf.gen.yaml` alongside the hashable plugin (separate `autobahn.wireguard.buf.gen.yaml` deleted).

### Plugin bug fix

Proto3 `optional` fields produce a synthetic oneof in the descriptor but no wrapper struct in Go. The plugin was incorrectly using the `Message_Field` wrapper name for them. Fixed with `!f.ContainingOneof().IsSynthetic()`.

## Test plan

- [ ] `go test ./sei-tendermint/internal/autobahn/...` — all autobahn packages pass
- [ ] `go test ./sei-tendermint/internal/blocksync/... ./sei-tendermint/internal/statesync/... ./sei-tendermint/internal/consensus/... ./sei-tendermint/internal/evidence/...` — wiring tests pass including no-op assertions
- [ ] `go test ./sei-tendermint/internal/protoutils/...` — wireguard scanner + plugin tests pass, including per-instance cap and proto3 optional tests
- [ ] `go build ./sei-tendermint/...` — clean build


sei-tendermint/internal/autobahn/autobahn.proto
```diff
@@ -179,7 +180,7 @@ message PersistedAvailPruneAnchor {
 
 message AppQC {
   AppProposal vote = 1;
-  repeated Signature sigs = 2;
+  repeated Signature sigs = 2 [(wireguard.max_count) = 100];
```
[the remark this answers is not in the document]
wen-coding: Currently we have 40, we can raise if this is indeed too low.

autobahn.wireguard.buf.gen.yaml
```diff
@@ -0,0 +1,11 @@
+version: v2
```
pompon0: this can be just added to the internal/buf.gen.yaml

sei-tendermint/internal/protoutils/msg.go
```diff
@@ -26,8 +26,20 @@ func Marshal[T Message](t T) []byte {
 	return utils.OrPanic1(proto.Marshal(t))
 }
 
+// wireScanner is implemented by proto types whose generated *.wireguard.go
+// adds a WireguardScan method. Unmarshal calls it automatically before
+// proto.Unmarshal so no per-call wiring is needed.
+type wireScanner interface {
```
pompon0: this should be exposed by wireguard

sei-tendermint/internal/evidence/wiring_test.go
```diff
@@ -5,17 +5,26 @@ import (
...
+}
+
+// TestWiring_EvidenceAcceptsAtCap asserts that a payload at exactly the cap
+// is accepted.
+func TestWiring_EvidenceAcceptsAtCap(t *testing.T) {
+	ev := wgtest.EvidenceWithCommit(wgtest.CommitWith(wgtest.MaxCommitSignatures))
+	require.NoError(t, ev.WireguardScan(wgtest.Marshal(t, &ev)))
+}
+
+func init() {
```
pompon0: init wrapper is useless here

pompon0 (approved):

autobahn.wireguard.buf.gen.yaml
wen-coding: done

sei-tendermint/internal/protoutils/msg.go
wen-coding: done

sei-tendermint/internal/evidence/wiring_test.go
wen-coding: removed

PR #3609, opened 2026-06-18
additional wireguard bounds for autobahn

pompon0:
* changed the semantics of max_count to limit field arity per message instance, rather than globally
* changed the generated code to populate a global registry of schemas (eventually I would like to build the schemas lazily at runtime from protoreflect, but for now we stay with generated code because of gogoproto).
* changed the api to use wireguard.Scan\[T\]() rather than access the schema directly
* added additional extensions for wireguard: sized, max_size, max_total_size, so that not only arity but also size of the field can be bounded and to enforce bounded max size of the message.
* whether the hardcoded bounds in proto are correct should be tested by constructing a message which saturates concrete limit in test and doing marshal -> scan (or unmarshal, since it triggers scan). This is especially important for things like hashes/signatures/keys, which have concrete sizes regardless of what protobuf says (i.e. we need to maintain compatible bounds between proto files and the rest of the code).
* merged wireguard into protoutils, since wireguard.Scan is now used every time protoutils.Unmarshal is called (it is not optional).
* extracted protoutils/runtime package which contains symbols which should be accessed only by the generated code.
* refactored wireguard_plugin tests - plugin error codes are now using proto file fixtures, other tests have been moved to scan_test.go and made to observe the generated code behavior instead of syntax.

In the next PR, I'll implement computation of the max message size, which then can be validated against the buffer.


sei-cosmos/x/staking/types/staking.pb.go
```diff
@@ -1274,783 +1274,786 @@ func (this *Pool) Description() (desc *github_com_gogo_protobuf_protoc_gen_gogo_
 func StakingDescription() (desc *github_com_gogo_protobuf_protoc_gen_gogo_descriptor.FileDescriptorSet) {
 	d := &github_com_gogo_protobuf_protoc_gen_gogo_descriptor.FileDescriptorSet{}
 	var gzipped = []byte{
-		// 12410 bytes of a gzipped FileDescriptorSet
-		0x1f, 0x8b, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x02, 0xff, 0xec, 0xbd, 0x7b, 0x94, 0x1b, 0xe7,
```
wen-coding, line 1278: Why did this file change?

sei-tendermint/autobahn/types/committee.go
```diff
@@ -37,6 +37,8 @@ type Committee struct {
 	genesisTimestamp time.Time
 }
 
+const MaxValidators = 100
```
wen-coding, line 40: Do we have a common place for all these constraints?

sei-tendermint/internal/autobahn/autobahn.proto
```diff
@@ -12,6 +12,7 @@ option go_package = "github.com/sei-protocol/sei-chain/sei-tendermint/internal/a
 // See `google.protobuf.Timestamp` for more detailed specification.
 message Timestamp {
   option (hashable.hashable) = true;
+  option (wireguard.sized) = true;
```
wen-coding, line 15: What happens if someday someone adds a new field and forgot to add this field?

sei-tendermint/internal/protoutils/wireguard/plugin/main.go
```diff
@@ -1,10 +1,14 @@
...
+// annotated message type. The output sits next to the .pb.go files as
+// `<name>.wireguard.go`.
 //
-// Only one annotation is needed:
+// Supported annotations:
 //
+//	(wireguard.sized) = true      // require structurally bounded message size
 //	(wireguard.max_count) = N      // cap on a repeated field's occurrences
+//	(wireguard.max_size) = N       // cap on one string/bytes/message instance
+//	(wireguard.max_total_size) = N // cap on summed bytes across field instances
```
wen-coding: nit: specify clearly what happens if you have nested repeated field

sei-tendermint/internal/protoutils/testonly.proto
```diff
@@ -15,3 +17,39 @@ message TestonlyMsg {
 message TestonlyChild {
   string value = 1;
 }
+
+message TestonlyCountedLeaf {
+  repeated string items = 1 [(wireguard.max_count) = 3];
+}
+
+message TestonlySizedLeaf {
+  repeated bytes items = 1 [(wireguard.max_total_size) = 3];
```
wen-coding: nit: add max_size case?

sei-tendermint/proto/wireguard/wireguard.proto
```diff
@@ -13,12 +13,25 @@ option go_package = "github.com/sei-protocol/sei-chain/sei-tendermint/proto/wire
...
-// message type that has (max_count) somewhere in its reachable subtree.
-// Descent through message-typed fields is automatic — if the target type
-// has a schema, the parent's rule descends into it.
+// message type that has a wireguard annotation somewhere in its reachable
+// subtree. Descent through message-typed fields is automatic — if the
+// target type has a schema, the parent's rule descends into it.
+extend google.protobuf.MessageOptions {
+  // sized marks messages that are required to have a bounded maximal wire
+  // size according to the wireguard plugin's structural checks.
+  bool sized = 414126221;
```
wen-coding, line 22: feel free to grab 414126218 and change max_count

sei-tendermint/autobahn/types/committee.go
pompon0: MaxValidators is committee property, payload constraints are with the payload type

sei-tendermint/internal/autobahn/autobahn.proto
pompon0: this annotation is enforcing boundedness of the message size. Adding it later is backward compatible, removing it is not.

sei-tendermint/proto/wireguard/wireguard.proto
pompon0: I mean, do we care? It is not like we have a central registry of extension numbers in our repo, so adding a new number requires inspecting all the proto files in the repo. I would feel more comfortable by assigning a random number.

sei-cosmos/x/staking/types/staking.pb.go
pompon0: because gogoproto code embeds transitive proto file descriptors, wireguard.proto in this case.

wen-coding (approved):

sei-tendermint/internal/protoutils/testonly.proto
pompon0: added more cases

PR #3610, opened 2026-06-18
Tests: Rpc Tests Phase 3

kollegian:
Rpc tests for filters, logs and storage
Only testing code added


integration_test/rpc_tests/eth/eth_newFilter.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+            const id = await sei.send('eth_newFilter', [sceneFilter()]);
+            expect(await sei.send('eth_uninstallFilter', [id]), 'first uninstall').to.equal(true);
+            expect(await sei.send('eth_uninstallFilter', [id]), 'second uninstall').to.equal(false);
+
+            const after = await rawSei('eth_getFilterLogs', [id]);
+            expectJsonRpcError(after, -32000, /filter does not exist/);
+        });
+    });
+
+    describe('rich criteria (multi-address / multi-topic), parity with eth_getLogs', () => {
```
amir-deris, line 76: The topic-filter matrix (OR-sets, wildcards, indexed positional topics, multi-address) is duplicated across three specs: `eth_getFilterChanges`, `eth_newFilter`, and `eth_subscribe` (`assertSubLogsMatchGetLogs`). The per-RPC oracle pattern is defensible, but if filter semantics change you will need to touch all three files.

Consider a follow-up that extracts a shared case table in `logsUtils` (e.g. `LOG_FILTER_MATRIX`) and parameterizes each spec over it. Not a merge blocker — just a maintainability note for when this surface evolves.

integration_test/rpc_tests/eth/eth_subscribe.spec.ts
```diff
@@ -0,0 +1,297 @@
...
+            const logs = await ws.waitFor(subId, 2);
+            for (const log of logs) {
+                expectLogShape(log, 'subscription log');
+                expect(log.address, 'log address is the filtered token').to.equal(address.toLowerCase());
+                expect(log.topics[0], 'log is a Transfer').to.equal(TRANSFER_TOPIC);
+                expect(log.removed, 'live log is not removed').to.equal(false);
+            }
+
+            // The mint topics0->emitter, then emitter->sink: subscription order is emission order.
+            expect(BigInt(logs[0].topics[2]), 'first Transfer credits the emitter').to.equal(
```
amir-deris: This assertion works because `BigInt` ignores leading-zero padding in hex strings, so the padded topic word and the bare address compare equal. That is easy to misread as a bug during review.

Prefer the explicit form used elsewhere in the suite:

```ts
expect(logs[0].topics[2]).to.equal(addressTopic(emitter.address));
```

Same semantics, clearer intent.

integration_test/rpc_tests/eth/eth_newFilter.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+            expect(blockId, 'block filter id is opaque hex').to.match(FILTER_ID);
+            expect(blockId).to.not.equal(logId);
+            await Promise.all([
+                sei.send('eth_uninstallFilter', [logId]),
+                sei.send('eth_uninstallFilter', [blockId]),
+            ]);
+        });
+    });
+
+    describe('wrong params / error handling (parity with geth)', () => {
```
amir-deris, line 120: Filter and WebSocket coverage in this PR is mostly Sei-only. Only a few cases cross-check geth (`eth_newBlockFilter` handle shape, `eth_newFilter` malformed-topic error). That is reasonable — geth WS is not in CI and polling-filter semantics can diverge — but it is worth calling out explicitly in the PR description or README so future reviewers do not expect symmetric geth asserts across every filter/subscription spec.

A short note under the filter/WS specs (e.g. "Sei functional coverage; geth parity limited to error envelopes where applicable") would help.

amir-deris (approved):

integration_test/rpc_tests/eth/eth_newFilter.spec.ts
kollegian: updated

integration_test/rpc_tests/eth/eth_subscribe.spec.ts
kollegian: updated

integration_test/rpc_tests/eth/eth_newFilter.spec.ts
kollegian: updated

integration_test/rpc_tests/eth/eth_subscribe.spec.ts
```diff
@@ -0,0 +1,234 @@
...
+            expect(oracle.length, `${ctx}: scene produced matching logs`).to.be.greaterThan(0);
+
+            const subId = await ws.subscribe(['logs', criteria]);
+            try {
+                const got = await ws.waitFor(subId, oracle.length);
+                // A bounded, address-scoped subscription must not over-deliver: poll for one extra
+                // log and require the wait to time out (no straggler beyond the eth_getLogs set).
+                let overDelivered = true;
+                try {
+                    await ws.waitFor(subId, oracle.length + 1, 2_000);
```
bdchatham, line 186: This one worries me a little. We're proving "no straggler beyond the `eth_getLogs` set" by waiting 2s for an extra log and treating the timeout as success — but a timeout can't tell "there is no extra log" apart from "the extra log just hasn't arrived yet." On a loaded CI runner I think this can pass green while the node actually over-delivers.

Could we instead advance the head ~2 blocks past `lastEventBlock` and then assert `buf.length === oracle.length` directly? That proves absence because the producing window has closed, rather than because we gave up waiting.

integration_test/rpc_tests/utils/logsUtils.ts
```diff
@@ -0,0 +1,407 @@
...
+ * Deploy a dedicated ERC20 and emit a fixed scene of four events:
+ *   1. mint     -> Transfer(0x0      -> emitter)
+ *   2. transfer -> Transfer(emitter  -> alice)
+ *   3. transfer -> Transfer(emitter  -> bob)
+ *   4. approve  -> Approval(emitter  -> alice)
+ *
+ * Each lands in its own block on Sei, giving callers exact block bounds and
+ * counts to assert against.
+ */
+export async function emitLogScene(
```
bdchatham, line 119: The "each event lands in its own block" assumption is load-bearing for the single-block count assertions downstream (the `fromBlock == toBlock` and `mint..bob` cases in `eth_getLogs.spec.ts`), but nothing enforces it — if two of these txs ever land in the same ~400ms block, those counts break and it'll look like a Sei bug rather than a test assumption.

Could we add a quick guard that `mintBlock < aliceBlock < bobBlock < approveBlock` are all distinct, and fail loudly with a clear "scene assumption violated" message? Cheap insurance.

integration_test/rpc_tests/utils/logsUtils.ts
```diff
@@ -0,0 +1,407 @@
...
+    const deadline = Date.now() + timeoutMs;
+    while (collected.length < want && Date.now() < deadline) {
+        const delta = await provider.send('eth_getFilterChanges', [filterId]);
+        collected.push(...delta);
+        if (collected.length < want) await new Promise(r => setTimeout(r, 300));
+    }
+    return collected;
+}
+
+export async function assertFilterChangesMatchGetLogs(
```
bdchatham, line 390: Small inconsistency in rigor: the hand-written `eth_getFilterChanges` test correctly drains and then asserts the next poll is `[]` (proving "exactly once") — but the shared helper the *matrix* runs through never does that follow-up empty poll, so the matrix cases prove "matches the oracle" but not "delivered exactly once." Worth one more poll-empty assertion here so the guarantee holds everywhere the matrix runs.

integration_test/rpc_tests/README.md
```diff
@@ -41,6 +41,16 @@ namespace dirs like `debug/`, `sei/`, etc.) answers one or more of:
   is the only reference — and it is **not** reliable for error envelopes
   (anvil/Hardhat reimplement the RPC layer and diverge from geth).
 
+> **Filter & subscription coverage is deliberately Sei-only.** The log-filter and
```
bdchatham, line 44: Really like this caveat being spelled out. While you're here: it now slightly contradicts a couple of the pre-existing rules above/below it, and since this is the PR that adds the specs those rules describe, might be worth reconciling them in the same change —

- the "What this suite proves" bullet says wrong-params are "asserted byte-for-byte against geth," which reads as universal; a quick inline qualifier ("…for methods with geth parity; see the caveat below") would match this. Same for Rule 5 under "Authoring a new spec."
- minor, unrelated: the authoring example only shows two of the four mandated sections, so anyone scaffolding from it quietly drops two. Either stub all four or soften the rule to "the applicable subset."

Not blocking — just easy to fix while it's all fresh.

integration_test/rpc_tests/utils/logsUtils.ts
```diff
@@ -0,0 +1,407 @@
...
+    title: string;
+    /**
+     * Build the eth_getLogs-style criteria and its expectations from an emitted scene. Every case is
+     * bounded to the scene's blocks + token so topic-only filters can't match unrelated chain history
+     * and bounded subscriptions replay a finite set.
+     */
+    build: (scene: LogScene) => {
+        criteria: Record<string, unknown>;
+        expectedCount: number;
+        check?: (logs: any[]) => void;
```
bdchatham, line 173: Not blocking, but since `expectLogShape` / `assertLogMatchesReceipt` / this `check` callback are about to be the canonical log oracles for the whole suite, could we type them with the `RpcLog` / `RpcReceipt` interfaces we already have in `txUtils.ts` instead of `any`? The fields line up exactly, it costs nothing, and it buys back compiler checking on every `l.topics[0]` / `l.logIndex` downstream — worth doing before these patterns set the Phase-3 convention.

integration_test/rpc_tests/eth/eth_subscribe.spec.ts
```diff
@@ -0,0 +1,234 @@
...
+            for (const head of heads) {
+                assertNewHeadSchema(head);
+                assertNewHeadInapplicableZeros(head);
+                const block = await sei.send('eth_getBlockByNumber', [head.number, false]);
+                expect(block, `canonical block exists for ${head.number}`).to.not.equal(null);
+                assertNewHeadMatchesBlock(head, block);
+                await assertNewHeadBaseFeeTracksChain(head, sei);
+            }
+
+            await ws.unsubscribe(subId);
```
bdchatham, line 62: A few of the newHeads `it` bodies unsubscribe outside a `try/finally`, so a failed assertion leaks the server-side subscription on the shared socket for the rest of the run (the matrix path already does this right). Also worth wiring `onclose`/`onerror` on the live socket in `WsClient` so a dropped WS rejects the pending call instead of hanging to the 120s suite timeout. Minor, but they compound as the suite grows.

integration_test/rpc_tests/utils/logsUtils.ts
```diff
@@ -0,0 +1,407 @@
...
+ * own precompile address, so eth_getLogs can index them like any contract's events.
+ */
+export const STAKING_IFACE = new ethers.Interface([
+    'function delegate(string valAddress) payable returns (bool)',
+    'event Delegate(address indexed delegator, string validator, uint256 amount)',
+]);
+export const DELEGATE_TOPIC = STAKING_IFACE.getEvent('Delegate')!.topicHash;
+
+/** First bonded validator's `seivaloper…` operator address, via the in-container CLI. */
+export async function firstBondedValidator(): Promise<string> {
```
bdchatham, line 56: This shells into `docker exec` with no timeout and falls back to `vals[0]` even if it's jailed/unbonding — which couples a *log-indexing* test to staking liveness and the local docker node. If a validator unbonds, this fails for a reason that has nothing to do with logs. Not urgent, but a timeout + a skip-guard (like the CW20 path has) would make it more portable.

bdchatham (commented):
Great work! 🙌 A few inline comments below — mostly test-robustness on a live chain, nothing structural.

PR #3611, opened 2026-06-18
tool: add full-scan EVM logical digest

blindchaser:
## Summary

Add an `evm-logical-digest` seidb operation for comparing EVM state across FlatKV and memIAVL at the same height. The command normalizes both backends into FlatKV physical keys, strips height-dependent value metadata, reports per-bucket `bucket_digest` values, and emits one `FINAL_DIGEST` line for backend comparison.

- `sei-db/tools/cmd/seidb/operations/evm_logical_digest.go`: Adds the `evm-logical-digest` command with FlatKV native scanning and memIAVL snapshot scanning. FlatKV reads use `RawGlobalIterator`; memIAVL reads stream snapshot `kvs` records sequentially so scan order does not affect correctness.
- `sei-db/tools/cmd/seidb/operations/evm_logical_digest.go`: Enforces an order-independent bucket accumulator over `sha256(len(key)||key||len(value)||value)`. The final digest combines account, code, storage, and marker-adjusted legacy bucket digests so a FlatKV-only migration-version row does not create a false mismatch.
- `sei-db/tools/cmd/seidb/operations/evm_logical_digest.go`: Supports semantic memIAVL normalization by default and an opt-in translator mode through `--memiavl-normalization translator`. Semantic mode decodes raw EVM leaves directly; translator mode routes leaves through `flatkv.ImportTranslator` to validate the migration mapping.
- `sei-db/tools/cmd/seidb/operations/evm_logical_digest.go`: Adds `--inspect-bucket`, prefix sharding, row listing, backend metadata details, and `--find-hash` support for isolating mismatched entries. memIAVL inspect honors the same normalization flag as the global digest, so diagnostics match the selected digest path.

## Test plan

- `sei-db/tools/cmd/seidb/operations/evm_logical_digest_test.go`: `TestSemanticMemiavlDigestMatchesTranslatorForCoreEVMKeys` verifies semantic normalization matches translator normalization for account, code, storage, and legacy buckets, including delete-equivalent zero storage and empty code rows.
- `sei-db/tools/cmd/seidb/operations/evm_logical_digest_test.go`: `TestSemanticMemiavlInspectMatchesTranslatorForCoreEVMKeys` verifies inspect bucket results match translator output for all normalized buckets.
- `sei-db/tools/cmd/seidb/operations/evm_logical_digest_test.go`: `TestInspectMemiavlRejectsUnknownNormalizationBeforeOpeningSnapshot` guarantees invalid `--memiavl-normalization` values are rejected before filesystem access.
- Manual verification: `go test ./sei-db/tools/cmd/seidb/operations`.


cody-littley (approved):
It looks like this code reads all account data into memory. If we have too much account data that risks OOMing the node we're using to compare. But since this tooling is temporary and won't be needed after the migration, I think it's probably ok for now as long as we can fit mainnet account state into memory on the machine where we are doing the testing.

blindchaser:
> It looks like this code reads all account data into memory. If we have too much account data that risks OOMing the node we're using to compare. But since this tooling is temporary and won't be needed after the migration, I think it's probably ok for now as long as we can fit mainnet account state into memory on the machine where we are doing the testing.

good point. this does not keep all evm state in memory, it does keep the account merged state keyed by unique account address so nonce/codehash can be normalized into the flatkv account buckets. storage/code/legacy are streamed into the digest. this is a good call out and we shuld watch memory usage during the pacific-1 comparision

Issue #3578, opened 2026-06-11
Dockerfile needs to be updated for GigaExecutor


m3diumrare:
Current Dockerfile image build (on tag v6.5.2) dies with the following missing shared object error when enabling Giga:

```
panic: failed to load evmone: evmc.Load("/go/src/sei-chain/giga/executor/lib/libevmone.0.12.0_linux_amd64.so"): EVMC loading error: /go/src/sei-chain/giga/executor/lib/libevmone.0.12.0_linux_amd64.so: cannot open shared object file: No such file or directory
```

masih:
Please note that this issue is resolved on main already. But a complimentary PR is open to improve the UX [here](https://github.com/sei-protocol/sei-chain/pull/3624)

PR #3613, opened 2026-06-20
Don't use hardcoded path to the evmc lib

qezz:

The runtime.Caller(0) returns paths to the source code, not picking them up during runtime.

The nodes are starting with giga executor enabled, instead of crashing trying to pick up the wrong hardcoded path to the .so file.

`panic: failed to load evmone: evmc.Load("/go/src/github.com/sei-protocol/sei/giga/executor/lib/libevmone.0.12.0_linux_amd64.so"): EVMC loading error: /go/src/github.com/sei-protocol/sei/giga/executor/lib/libevmone.0.12.0_linux_amd64.so: cannot open shared object file: No such file or directory`

### DIFF
--- a/giga/executor/lib/evmlib.go
+++ b/giga/executor/lib/evmlib.go
@@ -2,8 +2,6 @@ package lib
 
 import (
 	"fmt"
-	"path/filepath"
-	"runtime"
 
 	"github.com/ethereum/evmc/v12/bindings/go/evmc"
 )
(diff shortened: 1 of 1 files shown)


m3diumrare:
This looks like it likely resolves #3578

masih:
This issue is fixed on main, in that the default executor used is geth and evmone load failure is scilently ignored. A [separate PR](https://github.com/sei-protocol/sei-chain/pull/3624) improves the UX as well as encorporating the improvement kindly submitted by this PR.

Closing and many thanks for pointing this issue out + PR to fix 🙌

PR #3615, opened 2026-06-21
feat(protoutils): UnmarshalWithLimit — pre-decode allocation estimate

wen-coding:
## Summary

Adds allocation-bounded protobuf unmarshaling to bound heap allocation during decode.

- `UnmarshalWithLimit[T](bz []byte, limitBytes int)` — google v2 proto
- `UnmarshalGogoWithLimit(bz []byte, msg gogoproto.Message, limitBytes int)` — gogoproto (Tendermint P2P types)

Both functions run `Scan`/`ScanAny` (wireguard schema checks) first, then walk the raw wire bytes with `allocEstimate` before calling `Unmarshal`. If the estimated heap allocation exceeds `limitBytes`, the message is rejected without allocating anything.

## How `allocEstimate` works

The function walks raw wire bytes using `protoreflect.MessageDescriptor` for the field schema and `protoregistry`/`gogoproto.MessageType` for Go struct sizes. It accumulates:

| Wire construct | What is counted |
|---|---|
| Message field (repeated or singular) | `sizeof(GoStruct)` from type registry |
| Repeated message field | +`pointerSize` per element (backing array pointer) |
| `bytes` field | `sliceHeaderSize + len(val)` |
| `string` field | `stringHeaderSize + len(val)` |
| Packed repeated varint | element count × Go element size (1/4/8 bytes) — accounts for up to 8× wire-to-Go size difference |
| Map field | `mapEntryOverhead` per entry (runtime map internals) + key/value content |
| Unknown field (any wire type) | full wire record: `tagLen + n` |
| Wire-type mismatch | counted as unknown bytes (`tagLen + n`) — no panic |
| Singular field repeated N times | all N occurrences accumulated — accounts for gogoproto merge behavior |

The estimate is conservative (may over-count). A 10× estimation error is acceptable — the goal is to distinguish legitimate messages from wire payloads whose decoded size far exceeds their wire size.

## Gogoproto bridge

Tendermint P2P types (`tmproto.*`) are generated by `protoc-gen-gogofaster` and do not implement `google.golang.org/protobuf/proto.Message`. `UnmarshalGogoWithLimit` uses `github.com/golang/protobuf/proto.MessageReflect` (deprecated but the only bridge) to obtain the `protoreflect.MessageDescriptor`, and falls back to `gogoproto.MessageType` for struct size lookups.

## Singular field merge behavior

gogoproto merges repeated wire occurrences of a singular field (e.g. `Proposal.last_commit`) by appending sub-fields. Per-occurrence wireguard checks pass individually, but `allocEstimate` recurses into every wire occurrence and accumulates the totals, so the full merged allocation is estimated correctly.

## Map fields

Map fields are counted at `mapEntryOverhead` per entry (64 bytes) for runtime overhead plus key/value content from the recursive walk. This covers the `hmap` struct and bucket array slots conservatively. Currently no wireguard-protected message has map fields; support is included for future messages.

## Test plan

- `SmallMessageAccepted` — small repeated-message field passes
- `ManyEmptyEntriesRejected` — 10k empty repeated entries caught before Unmarshal
- `LargePayloadRejected` — large bytes field caught
- `UnknownBytesFieldCounted` — unknown bytes field counted toward limit
- `SmallUnknownFieldsAccepted` — small unknown scalars pass cleanly
- `ResultIsCorrect` — decoded result is correct when within limit
- `PackedVarintAmplificationCounted` — 200k packed uint64 (1 byte/wire, 8 bytes/Go) rejected
- `WireTypeMismatchNoPanic` — mismatched wire type: no panic, counted as unknown
- `TruncatedInputReturnsError` — truncated mid-field returns error, not panic
- `ZeroLimitPanics` — `limitBytes <= 0` panics (programming error)
- `LargeMapRejected` — 50k map entries exceed 1MB limit
- `SmallMapAccepted` — 2-entry map passes
- `GogoManyEmptySignaturesRejected` — gogoproto Commit with 10k signatures caught
- `GogoSmallCommitAccepted` — legitimate gogoproto Commit passes
- `GogoSingularFieldMerge` — 500 × last_commit each with 50 signatures exceeds 1MB

**Load tests** (`alloc_scan_load_test.go`):
- `TestUnmarshalWithLimit_MaxBlockAccepted` — max-sized autobahn Block (2000 txs × 1024 bytes) accepted at 4MB limit
- `BenchmarkUnmarshalWithLimit_MaxBlock` — ~242µs for 2MB block, 8.5 GB/s
- `BenchmarkUnmarshalWithLimit_AmplifiedPayload` — ~316ns for 20KB payload rejected before Unmarshal

### DIFF
--- a/sei-tendermint/internal/protoutils/alloc_scan.go
+++ b/sei-tendermint/internal/protoutils/alloc_scan.go
@@ -82,7 +82,7 @@ func allocEstimate(data []byte, desc protoreflect.MessageDescriptor) (int, error
 				// stores it verbatim in the unknown fields blob.
 				total += tagLen + n
 			} else if fd.IsList() {
-				total += scalarElementSize(fd.Kind())
+				total += sliceHeaderSize + scalarElementSize(fd.Kind())
 			}
 
 		case protowire.Fixed32Type:
@@ -96,7 +96,7 @@ func allocEstimate(data []byte, desc protoreflect.MessageDescriptor) (int, error
 				// stores it verbatim in the unknown fields blob.
 				total += tagLen + n
 			} else if fd.IsList() {
-				total += scalarElementSize(fd.Kind())
+				total += sliceHeaderSize + scalarElementSize(fd.Kind())
 			}
 
 		case protowire.Fixed64Type:
@@ -110,7 +110,7 @@ func allocEstimate(data []byte, desc protoreflect.MessageDescriptor) (int, error
 				// stores it verbatim in the unknown fields blob.
 				total += tagLen + n
 			} else if fd.IsList() {
-				total += scalarElementSize(fd.Kind())
+				total += sliceHeaderSize + scalarElementSize(fd.Kind())
 			}
 
 		case protowire.StartGroupType:
--- a/sei-tendermint/internal/protoutils/alloc_scan.go
+++ b/sei-tendermint/internal/protoutils/alloc_scan.go
@@ -148,9 +148,16 @@ func bytesFieldSize(tagLen, n int, val []byte, fd protoreflect.FieldDescriptor) 
 	}
 
 	total := 0
-	if fd.IsList() {
-		// One slice header per field; adding it per occurrence over-counts by
-		// (N-1)*sliceHeaderSize across N elements — noise near a 1MB limit.
+	if fd.IsMap() {
+		// Map fields: Go allocates a runtime map (hmap), not a slice. Add
+		// per-entry overhead for hmap fields, bucket slots, and tophash.
+		// Over-counts the fixed hmap header across N entries, which is
+		// conservative.
+		total += mapEntryOverhead
+	} else if fd.IsList() {
+		// Repeated fields: one slice header per field. Adding it per
+		// occurrence over-counts by (N-1)*sliceHeaderSize across N elements
+		// — noise near a 1MB limit.
 		total += sliceHeaderSize
 	}
 
@@ -158,12 +165,6 @@ func bytesFieldSize(tagLen, n int, val []byte, fd protoreflect.FieldDescriptor) 
 	case protoreflect.MessageKind, protoreflect.GroupKind:
 		if fd.IsList() {
 			total += pointerSize // pointer element in the backing array
-		}
-		if fd.IsMap() {
-			// Add per-entry map runtime overhead (hmap fields, bucket slots,
-			// tophash). Over-counts the fixed hmap header across N entries,
-			// which is conservative.
-			total += mapEntryOverhead
 		}
 		sub, err := allocEstimate(val, fd.Message())
 		if err != nil {
--- a/sei-tendermint/internal/protoutils/alloc_scan.go
+++ b/sei-tendermint/internal/protoutils/alloc_scan.go
@@ -1,0 +1,301 @@
+package protoutils
+
+import (
+	"fmt"
+	"reflect"
+
+	gogoproto "github.com/gogo/protobuf/proto"
+	"google.golang.org/protobuf/encoding/protowire"
+	"google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/reflect/protoregistry"
+)
+
+var (
+	pointerSize      = int(reflect.TypeFor[*byte]().Size())  // 8 on 64-bit
+	sliceHeaderSize  = int(reflect.TypeFor[[]byte]().Size()) // 24 on 64-bit
+	stringHeaderSize = int(reflect.TypeFor[string]().Size()) // 16 on 64-bit
+
+	// mapEntryOverhead is added per map entry to account for Go runtime map
+	// internals (hmap struct, bucket array slots, tophash bytes) that are not
+	// captured by the map-entry message struct size. Adding it per entry
+	// over-counts the fixed hmap header for multi-entry maps, which is
+	// conservative. 8×pointerSize matches the number of fields in runtime.hmap.
+	mapEntryOverhead = 8 * int(reflect.TypeFor[*byte]().Size())
+)
+
+// allocEstimate walks raw protobuf wire bytes and returns a conservative
+// upper-bound on the heap bytes that proto.Unmarshal would allocate.
+//
+// The estimate accounts for:
+//   - the Go struct for each message occurrence (looked up from protoregistry)
+//   - backing arrays for bytes and string fields
+//   - pointer elements in repeated-message slices
+//
+// Unknown fields (field numbers not in the descriptor) are stored verbatim by
+// proto.Unmarshal in a single raw []byte blob on the struct without decoding.
+// Their allocation cost equals their wire size, so we add the wire bytes for
+// each unknown field occurrence. This is exact for bytes-type unknown fields
+// and a slight over-count for scalar unknown fields (which store tag+value
+// together), but both are correct in the conservative direction.
+//
+// The function returns an error on corrupt or truncated wire bytes. Truncation
+// (n < 0 from a Consume* call) surfaces as protowire.ParseError("unexpected
+// end of data") and is treated identically to corruption — the caller receives
+// an error and Unmarshal is never called.
+func allocEstimate(data []byte, desc protoreflect.MessageDescriptor) (int, error) {
+	total := msgStructSize(desc)
+
+	for len(data) > 0 {
+		num, typ, tagLen := protowire.ConsumeTag(data)
+		if tagLen <= 0 {
+			return 0, fmt.Errorf("tag: %w", protowire.ParseError(tagLen))
+		}
+		if num == 0 {
+			// Field number 0 is reserved and illegal in the protobuf spec.
+			return 0, fmt.Errorf("invalid field number 0")
+		}
+		data = data[tagLen:]
+
+		fd := desc.Fields().ByNumber(num)
+
+		switch typ {
+		case protowire.BytesType:
+			val, n := protowire.ConsumeBytes(data)
+			if n <= 0 {
+				return 0, fmt.Errorf("field %d bytes: %w", num, protowire.ParseError(n))
+			}
+			data = data[n:]
+			add, err := bytesFieldSize(tagLen, n, val, fd)
+			if err != nil {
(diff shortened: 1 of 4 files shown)


sei-tendermint/internal/protoutils/msg.go
```diff
@@ -40,6 +42,45 @@ func Unmarshal[T Message](bytes []byte) (T, error) {
...
+func UnmarshalGogoWithLimit(bz []byte, msg gogoproto.Message, limitBytes int) error {
+	desc := golangproto.MessageReflect(msg).Descriptor()
+	est, err := allocEstimate(bz, desc)
+	if err != nil {
+		return fmt.Errorf("protoutils: malformed wire bytes: %w", err)
+	}
+	if est > limitBytes {
+		return fmt.Errorf("protoutils: message would allocate ~%d bytes, limit is %d", est, limitBytes)
+	}
+	return gogoproto.Unmarshal(bz, msg)
```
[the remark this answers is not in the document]
wen-coding: Here the caller will never pass a pre-populated msg.

sei-tendermint/internal/protoutils/alloc_scan.go
```diff
@@ -0,0 +1,233 @@
...
+		}
+	}
+	// Gogoproto types register under their full name via proto.RegisterType.
+	if t := gogoproto.MessageType(string(desc.FullName())); t != nil {
+		if t.Kind() == reflect.Ptr {
+			t = t.Elem()
+		}
+		return int(t.Size())
+	}
+	return 256
```
pompon0: why do we allow that? Shouldn't we panic instead?

sei-tendermint/internal/protoutils/alloc_scan.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+const (
+	pointerSize      = 8  // pointer on 64-bit
```
pompon0: reflect.TypeFor\[...\]().Size() ?

sei-tendermint/internal/protoutils/alloc_scan.go
```diff
@@ -0,0 +1,233 @@
...
+	case protoreflect.Fixed32Kind, protoreflect.Sfixed32Kind, protoreflect.FloatKind:
+		return len(bz) // 4 bytes wire == 4 bytes Go, exact
+	case protoreflect.Fixed64Kind, protoreflect.Sfixed64Kind, protoreflect.DoubleKind:
+		return len(bz) // 8 bytes wire == 8 bytes Go, exact
+	case protoreflect.BoolKind:
+		return countVarintsInPacked(bz) * 1 // bool = 1 byte in Go
+	case protoreflect.Int32Kind, protoreflect.Uint32Kind,
+		protoreflect.Sint32Kind, protoreflect.EnumKind:
+		return countVarintsInPacked(bz) * 4
+	default: // int64, uint64, sint64
```
pompon0: make it explicit case, panic on default

sei-tendermint/internal/protoutils/alloc_scan.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+// countVarintsInPacked counts the number of varint-encoded elements in a
+// packed repeated field payload.
+func countVarintsInPacked(bz []byte) int {
+	count := 0
+	for len(bz) > 0 {
+		_, n := protowire.ConsumeVarint(bz)
+		if n < 0 {
+			break // malformed; allocEstimate will catch the outer error
```
pompon0: wdym by "will catch the outer error"?

sei-tendermint/internal/protoutils/alloc_scan.go
```diff
@@ -0,0 +1,233 @@
...
+				// values in 1 byte on the wire while occupying 4 or 8 bytes in
+				// the Go slice — up to 8× amplification. We count elements and
+				// multiply by the Go element size to avoid undercounting.
+				total += packedAllocSize(val, fd.Kind())
+			}
+
+		case protowire.VarintType:
+			_, n := protowire.ConsumeVarint(data)
+			if n < 0 {
+				return 0, fmt.Errorf("alloc scan: malformed varint field %d: %w", num, protowire.ParseError(n))
```
pompon0: nit: this shared error message prefix looks like it can be prepended at the call site.

sei-tendermint/internal/protoutils/alloc_scan.go
```diff
@@ -0,0 +1,233 @@
...
+			}
+			data = data[n:]
+			if fd == nil {
+				// Unknown scalar: proto.Unmarshal stores tag+value verbatim in
+				// the unknown fields blob, allocating exactly tagLen+n bytes.
+				total += tagLen + n
+			} else if fd.IsList() {
+				total += sliceHeaderSize + 8 // over-counts slice header per element, noise at 1MB scale
```
pompon0: technically this can cause up to 32x higher estimate than the real thing.

pompon0:
Note that this impl is meaningfully slower than the current wireguard approach because of the use of descriptors (note that proto.Unmarshal does not use protoreflect on the hot path). Probably still acceptable overhead though, just saying.

sei-tendermint/internal/protoutils/alloc_scan.go
```diff
@@ -0,0 +1,277 @@
...
+}
+
+// countVarintsInPacked counts the number of varint-encoded elements in a
+// packed repeated field payload.
+func countVarintsInPacked(bz []byte) int {
+	count := 0
+	for len(bz) > 0 {
+		_, n := protowire.ConsumeVarint(bz)
+		if n < 0 {
+			break // partial count on malformed input; proto.Unmarshal will reject it
```
philipsu522: should we return an error here, rather than a partial count?

philipsu522 (approved):

sei-tendermint/internal/protoutils/alloc_scan.go
```diff
@@ -0,0 +1,273 @@
...
+			}
+			data = data[n:]
+			if fd != nil && (fd.Kind() == protoreflect.GroupKind) {
+				sub, err := allocEstimate(val, fd.Message())
+				if err != nil {
+					return 0, err
+				}
+				total += sub
+			} else {
+				total += len(val)
```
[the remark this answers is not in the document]
wen-coding: done

sei-tendermint/internal/protoutils/alloc_scan.go
```diff
@@ -0,0 +1,273 @@
...
+				// sizes so len(val) is correct. Varint kinds (bool, int32,
+				// int64, uint32, uint64, sint32, sint64, enum) can encode small
+				// values in 1 byte on the wire while occupying 4 or 8 bytes in
+				// the Go slice — up to 8× amplification. We count elements and
+				// multiply by the Go element size to avoid undercounting.
+				packed, err := packedAllocSize(val, fd.Kind())
+				if err != nil {
+					return 0, err
+				}
+				total += packed
```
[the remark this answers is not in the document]
wen-coding: packedAllocSize on a singular scalar sent as BytesType counts varint elements × element size, which equals or exceeds the actual unknown-field blob size

sei-tendermint/internal/protoutils/msg.go
```diff
@@ -40,6 +42,51 @@ func Unmarshal[T Message](bytes []byte) (T, error) {
...
+// UnmarshalGogoWithLimit is the gogoproto variant of UnmarshalWithLimit.
+// It uses github.com/golang/protobuf's reflection bridge to obtain the
+// protoreflect.MessageDescriptor from a gogoproto-generated type (which does
+// not implement google.golang.org/protobuf/proto.Message directly), allowing
+// the same allocEstimate walk to protect Tendermint P2P messages.
+func UnmarshalGogoWithLimit(bz []byte, msg gogoproto.Message, limitBytes int) error {
+	if err := ScanAny(bz, msg); err != nil {
+		return err
+	}
+	desc := golangproto.MessageReflect(msg).Descriptor() //nolint:staticcheck
```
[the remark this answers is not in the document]
wen-coding: fixed

sei-tendermint/internal/protoutils/alloc_scan.go
wen-coding: changed to panic

sei-tendermint/internal/protoutils/alloc_scan.go
wen-coding: done

sei-tendermint/internal/protoutils/alloc_scan.go
wen-coding: done

sei-tendermint/internal/protoutils/alloc_scan.go
wen-coding: removed

sei-tendermint/internal/protoutils/alloc_scan.go
wen-coding: done

sei-tendermint/internal/protoutils/alloc_scan.go
wen-coding: fixed, scalarElementSize now returns 1 for bool, 4 for 32-bit kinds, 8 for 64-bit kinds. The 32× overcount for bool (1 byte wire, was returning 32) is gone.

sei-tendermint/internal/protoutils/alloc_scan.go
wen-coding: done, returning an error now

sei-will (approved):

sei-tendermint/internal/protoutils/alloc_scan.go
pompon0: aren't we now undercounting the slice headers for repeated varints?

pompon0 (approved):

sei-tendermint/internal/protoutils/alloc_scan.go
wen-coding: Good catch, now adding the slice header byte counts

PR #3619, opened 2026-06-22
MaxSize computation for sized proto messages

pompon0:
For each message with "wireguard.sized = true", we generate "MaxSize() int" method. It will be used to bound buffers intended for caching those messages. Additionally I have made wireguard apply implicit max_count = 1 to every singular field, because proto.Unmarshal silently merges recursively instances of singular fields (if there are many), therefore making Scan() useless (which applies bounds field arity by instance). We have no use case for encoding singular fields as multiple partitions, so we simply disallow such encoding instead.


sei-tendermint/internal/protoutils/wireguard_plugin/main.go
```diff
@@ -201,18 +219,209 @@ func run(p *protogen.Plugin) error {
...
+
+func (s *sizedMessageMaxState) scalarValueSize(fd protoreflect.FieldDescriptor) int {
+	switch fd.Kind() {
+	case protoreflect.BoolKind:
+		return 1
+	case protoreflect.EnumKind:
+		size := 1
+		values := fd.Enum().Values()
+		for i := range values.Len() {
+			size = max(size, protowire.SizeVarint(uint64(int64(values.Get(i).Number())))) // nolint: gosec // WAI, negative enums are encoded as uint64
```
wen-coding, line 284: nit: this is the size known to receiving side, but the sending side could define a much larger enum value and send it over?

sei-tendermint/internal/protoutils/wireguard_plugin/main.go
```diff
@@ -375,28 +584,22 @@ func newEmitIdents(g *protogen.GeneratedFile) emitIdents {
...
 			continue
 		}
 
-		genDir, err := filepath.Rel(*moduleFlag, string(file.GoImportPath))
-		if err != nil {
-			return fmt.Errorf("filepath.Rel(): %w", err)
-		}
 		genFileName := strings.TrimSuffix(filepath.Base(file.Desc.Path()), ".proto") + ".wireguard.go"
-		g := p.NewGeneratedFile(filepath.Join(genDir, genFileName), file.GoImportPath)
+		g := p.NewGeneratedFile(filepath.Join(string(file.GoImportPath), genFileName), file.GoImportPath)
```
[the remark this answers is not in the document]
wen-coding: why is module param dropped here?

sei-tendermint/internal/protoutils/wireguard_plugin/main.go
```diff
@@ -201,18 +219,209 @@ func run(p *protogen.Plugin) error {
...
+
 // validateRuleValues rejects explicit zero-valued wireguard annotations,
 // which would silently mean "no cap" at runtime. An explicit zero is almost
 // certainly a mistake; pick a positive cap or drop the annotation if the
 // field is genuinely unbounded.
-func validateRuleValues(byName map[protoreflect.FullName]protoreflect.MessageDescriptor, inSchema map[protoreflect.FullName]bool, exts wireguardExts) error {
-	for fullName := range inSchema {
-		d := byName[fullName]
+func validateRuleValues(byName map[protoreflect.FullName]protoreflect.MessageDescriptor, exts wireguardExts) error {
+	for _, d := range byName {
```
wen-coding, line 382: would this iterate on all messages including external importas as well?

sei-will (approved):

sei-tendermint/internal/protoutils/wireguard_plugin/main.go
pompon0: if received does not know the received value, they cannot interpret it. Same would apply to fields which receiver does not know about. It is expected that MaxSize bound will NOT be backward compatible in many cases.

sei-tendermint/internal/protoutils/wireguard_plugin/main.go
pompon0: no, these bounds apply only to the protos compiled. Fields of message types from external imports are not validated at all. If such a field belongs to a sized message, plugin will protest unless the field has max_size set. For the sake of autobahn I presume we will rely on bounds of sized messages only (Scan/MaxSize). But if we want to, we can also enrich the Scan to generate (and cache) the schema in runtime, so that for example max_count = 1 for singular fields always always apply, even on external imports.

sei-tendermint/internal/protoutils/wireguard_plugin/main.go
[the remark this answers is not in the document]
pompon0: it turns out that protogen is the one responsible for trimming the module path, so the new approach is more standard.

wen-coding (approved):

PR #3620, opened 2026-06-22
ReceiptStore Add littidx backend (LittDB bodies + pebble tag index)

Kbhat1:
**Description**

- Adds a littidx receipt-store backend that stores receipt bodies in LittDB with a tx-hash secondary key for single-lookup GetReceipt
- Add a pebble tag index for getLogs range queries
- FilterLogs seeks index entries, intersects candidates across the filter criteria, and point-reads only matching receipts
- Pruning is a single O(1) pebble range tombstone with litt TTL reclaiming bodies behind a block-height retention floor

**Testing**

  - Added unit tests
  - Verified on node in cryptosim + verifying on node in cluster


sei-db/ledger_db/receipt/litt_tag_index.go
```diff
@@ -0,0 +1,316 @@
...
+					break
+				}
+				th := common.HexToHash(topic)
+				scratch = appendLittTagKey(scratch[:0], blockNumber, tagKindTopic0+byte(p), th[:], txIndex, firstLogIndex, txHash) //nolint:gosec // p < maxIndexedTopics
+				if err := batch.Set(scratch, nil); err != nil {
+					return err
+				}
+			}
+		}
+		firstLogIndex += uint32(len(record.Receipt.Logs)) //nolint:gosec // log counts fit within uint32
```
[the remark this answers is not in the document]
Kbhat1: We are going to freshly state sync nodes (no migration). A block is always writen in one `SetReceipts` call, so there's a single part and `firstLongIndex` is correct

RPX layer also rebuilds `LogIndex` from block data anyway, so wouldn't surface

sei-db/ledger_db/receipt/litt_receipt_store.go
```diff
@@ -0,0 +1,478 @@
...
+	if s.keepRecent <= 0 || s.pruneInterval <= 0 {
+		return
+	}
+	s.backgroundWg.Add(1)
+	go func() {
+		defer s.backgroundWg.Done()
+		for {
+			pruneBefore := s.latestVersion.Load() - s.keepRecent
+			if pruneBefore > 0 {
+				if err := s.pruneBlocksBelow(uint64(pruneBefore)); err != nil {
```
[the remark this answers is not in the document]
Kbhat1: Updated PR (cutoff now matches pebble backend)

sei-db/ledger_db/receipt/litt_receipt_store.go
```diff
@@ -0,0 +1,478 @@
...
+	maxBlock := blockNumbers[len(blockNumbers)-1]
+	newLatest := s.latestVersion.Load()
+	if int64(maxBlock) > newLatest { //nolint:gosec // block heights fit within int64
+		newLatest = int64(maxBlock) //nolint:gosec // block heights fit within int64
+		if err := batch.Set(receiptLatestVersionKey, encodeBlockNumber(maxBlock)); err != nil {
+			return err
+		}
+	}
+	if err := batch.Commit(dbtypes.WriteOptions{}); err != nil {
+		return err
```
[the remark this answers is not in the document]
Kbhat1: This is self-healing: `SetReceipts` error panics in pre-commit so block never commits and replays on restart. No permanent orphans

sei-db/ledger_db/receipt/litt_receipt_store.go
```diff
@@ -0,0 +1,482 @@
...
```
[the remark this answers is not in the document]
Kbhat1: Not a real issue. On replay of litt, last write wins on tx-hash secondary key. those re-executed receipt bytes are identical and the duplicate part just re-points the key to the same content

sei-db/ledger_db/receipt/litt_receipt_store.go
```diff
@@ -0,0 +1,491 @@
...
+	receiptBackendLittIdx = "littidx"
+
+	littReceiptTableName = "receipts"
+	littFlushInterval    = 100 * time.Millisecond
+	// littTTLPerBlock converts the KeepRecent block count into litt's wall-clock
+	// TTL (KeepRecent * littTTLPerBlock). Set above Giga block times so the TTL
+	// over-retains; only a sustained block time above this would expire a body
+	// still inside the height-based KeepRecent window, which reads mask as
+	// not-found (the earliest-version floor is authoritative).
+	littTTLPerBlock = 2 * time.Second
```
cody-littley, line 88: If you want more precise control over when LittDB deletes data, consider using the following:

```
type TableConfig struct {
    // ...

	// A function that is called to determine if a key is eligible for garbage collection. Keys are GC eligible once
	// their TTL has expired, and once this function returns true. If nil, only TTL determines GC eligibility.
	GCFilter GCFilter
}
```

```
type GCFilter func(key []byte, isPrimaryKey bool) (bool, error)
```

LittDB deletes data when the following are both true:

- TTL has expired
- the GCFilter returns true for the key (ignored if unset)

sei-db/ledger_db/receipt/litt_receipt_store.go
```diff
@@ -0,0 +1,491 @@
...
+	binary.BigEndian.PutUint64(key, blockNumber)
+	binary.BigEndian.PutUint32(key[blockNumLen:], part)
+	return key
+}
+
+func newLittReceiptStore(cfg dbconfig.ReceiptStoreConfig, storeKey sdk.StoreKey) (ReceiptStore, error) {
+	if err := os.MkdirAll(cfg.DBDirectory, 0o750); err != nil {
+		return nil, fmt.Errorf("failed to create receipt store directory: %w", err)
+	}
+	littConfig, err := litt.DefaultConfig(filepath.Join(cfg.DBDirectory, "littdb"))
```
cody-littley, line 106: Is `cfg.DBDirectory` scoped to the receipt store? If it is, then no complaints. If not, then we might want to better qualify this name, since a node might be running multiple LittDB instances.

sei-db/ledger_db/receipt/litt_receipt_store.go
```diff
@@ -0,0 +1,491 @@
...
+	}
+	// Receipt-workload tuning: this is a small-value, many-keys workload (every
+	// tx hash is a key), the opposite of litt's few-large-values default. The
+	// stock seal triggers fire every few thousand tiny keys / 2MB of key file
+	// and stall throughput, so raise the key-count and key-file caps past a
+	// retention window and let only the value-file size bind segment seals.
+	littConfig.MaxSegmentKeyCount = 100_000_000
+	littConfig.TargetSegmentFileSize = 512 * unit.MB
+	littConfig.TargetSegmentKeyFileSize = 5 * unit.GB
+	littConfig.KeymapType = keymap.PebbleDBKeymapType
```
cody-littley, line 118: PebbleDB is the default keymap type, but ok to leave it as is if you'd rather have this be explicit.

sei-db/ledger_db/receipt/litt_tag_index.go
```diff
@@ -0,0 +1,319 @@
...
+
+// littTagTagKey is the scan prefix for one (block, kind+tag).
+func littTagTagKey(blockNumber uint64, kindTag []byte) []byte {
+	return append(littTagBlockKey(blockNumber), kindTag...)
+}
+
+// appendLittTagKey writes a full tag key into dst[:0], reusing its capacity so
+// the write path allocates no per-key slice (pebble's batch.Set copies the key,
+// so the buffer is free to reuse immediately after).
+func appendLittTagKey(dst []byte, blockNumber uint64, kind byte, tag []byte, txIndex, firstLogIndex uint32, txHash common.Hash) []byte {
```
cody-littley, line 74: nit, might be a good candidate for line wrapping

cody-littley (approved):

sei-db/ledger_db/receipt/litt_receipt_store.go
yzang2019: I think we need a more precise/accurate way to prune receipt/logs for the sake of watermark management

yzang2019 (approved):

PR #3621, opened 2026-06-22
fix(evmrpc): cap eth_getBlockReceipts fan-out and concurrent logs subscriptions (PLT-701)

amir-deris:
## Summary

Part of the RPC endpoint protection workstream. Three hardening fixes in `evmrpc`, each bounding the resources a single client request can consume:

1. **Hard cap on `eth_getBlockReceipts` goroutine fan-out** (`evmrpc/block.go`)
2. **Make the `eth_subscribe(logs)` poll loop stop promptly on disconnect** (`evmrpc/subscribe.go`)
3. **Cap the number of concurrent `eth_subscribe(logs)` subscriptions** (`evmrpc/subscribe.go`, `evmrpc/config`)

Linear: [PLT-701](https://linear.app/seilabs/issue/PLT-701/)

## 1. Cap `eth_getBlockReceipts` goroutine fan-out

The receipt loop previously spawned **one goroutine per transaction** with a bare `sync.WaitGroup`, so a block containing a very large number of txs would create an unbounded number of goroutines from a single request.

Replaced the manual `WaitGroup`/`Mutex` loop with an `errgroup` whose limit is a hard cap:

```go
const maxBlockReceiptsConcurrency = 100
...
g, ctx := errgroup.WithContext(ctx)
g.SetLimit(maxBlockReceiptsConcurrency)
```

- `errgroup.SetLimit` blocks `Go()` until a slot frees, so it bounds the number of **live goroutines** (not just concurrent work). This covers the `eth_`, `sei_`, and `sei2_` variants since they share `GetBlockReceipts`. Underlying DB reads are already bounded by the worker-pool DB semaphore, so additional goroutines beyond this cap only add scheduler/memory pressure without improving throughput.
- `errgroup.WithContext` derives a `ctx` that is cancelled as soon as the first goroutine returns an error (or the caller cancels). Each goroutine checks `ctx.Err()` before starting work and bails early, so once one receipt fails the remaining queued goroutines don't do work the group is about to discard.

Existing error handling is preserved (the "not found" receipt skip and first-error propagation, now via `g.Wait()`); the dead `if returnErr != nil` tail check was removed since errors now flow through `g.Wait()`. The result is all-or-nothing: on any error `GetBlockReceipts` returns `nil, err` rather than a partial receipt list.

## 2. `eth_subscribe(logs)` poll loop stops promptly on disconnect

The poll loop ended each iteration with an unconditional `time.Sleep(SleepInterval)`, so it kept polling and notifying on a fixed cadence even after the subscriber was gone. Replaced the sleep with a select that also wakes on the subscription's error channel:

```go
select {
case <-rpcSub.Err():     // client disconnected / unsubscribed
    return
case <-time.After(SleepInterval):
}
```

This mirrors the `newHeads` handler. Note: the per-call `ctx` is **not** used to tear down the loop — the RPC framework cancels it as soon as the `eth_subscribe` call returns, so it would kill the long-lived subscription immediately; `rpcSub.Err()` is the correct disconnect/unsubscribe signal.

## 3. Cap concurrent `eth_subscribe(logs)` subscriptions

`NewHeads` already caps concurrent subscriptions via `MaxSubscriptionsNewHead`, but `Logs` had no equivalent — every `eth_subscribe(logs)` spawned a long-lived polling goroutine with no ceiling. Added a parallel `MaxSubscriptionsLogs` knob (default `10000`) and enforced it:

- New config field `MaxSubscriptionsLogs` wired through `config.go` (struct field, default, flag `evm.max_subscriptions_logs`, parse, `app.toml` template) and into `SubscriptionConfig.logLimit` in `server.go`.
- `Logs` reserves a slot via `acquireLogSub()` **before** allocating the subscription, returning the same `"no new subscription can be created"` error as `NewHeads` when the cap is reached, so a rejected client never spawns a polling goroutine. The slot is released via `defer releaseLogSub()` at the top of each subscription goroutine, so every exit path frees it.

## Testing

- `go build ./evmrpc/...`, `go vet ./evmrpc/...`, `gofmt -s` all clean.
- Existing suites pass: `evmrpc` and `evmrpc/tests` `Subscribe|Logs|Receipt|Block` selectors, including the multi-tx `TestGetBlockReceipts` integration test that exercises the rewritten fan-out path.
- New tests:
  - `evmrpc/subscribe_logcap_internal_test.go` — covers the logs-subscription cap: acquire/release at the limit, `limit == 0` rejecting all (parity with `NewHeads`), release-underflow guard, and a 500-goroutine concurrent stress test asserting the CAS acquire admits exactly `limit` (passes under `-race`).
  - `evmrpc/config/config_test.go` — `TestReadConfigMaxSubscriptionsLogs` covers the happy-path parse and the bad-type error branch for the new knob.


evmrpc/subscribe.go
```diff
@@ -278,7 +278,16 @@ func (a *SubscriptionAPI) Logs(ctx context.Context, filter *filters.FilterCriter
 			}
 			begin = lastToHeight
 			filter.FromBlock = big.NewInt(lastToHeight + 1)
-			time.Sleep(SleepInterval)
+			// Wait before the next poll, but stop promptly if the client
+			// disconnects (rpcSub.Err()) or the request context is cancelled /
+			// its deadline expires (ctx.Done()).
+			select {
+			case <-ctx.Done():
```
bdchatham: I think this one will actually break the subscription. The ctx we get here gets cancelled as soon as the eth_subscribe call returns its subscription id — the handler in the go-ethereum fork we're on (v1.15.7-sei-16) does a `defer cancel()` on it in handleNonBatchCall (rpc/handler.go:283) and again in startCallProc (rpc/handler.go:384). Since Logs hands back rpcSub right away, by the time we loop back to this select the ctx is already done, so the client gets the first backfill batch and then the goroutine just exits — no more logs, no error, no unsubscribe.

That's why geth's own filter subscriptions (and NewHeads right here in this file) only watch rpcSub.Err() and never the method ctx. I'd drop the ctx.Done() case and keep just rpcSub.Err() + the time.After, so it lines up with NewHeads. If we actually want a disconnect to cancel an in-flight fetch, that's a separate change — we'd need a connection-lifetime ctx, not this per-call one.

Worth noting too that the logs-subscription integration test is skipped, so CI won't catch this.

evmrpc/block.go
```diff
@@ -333,34 +339,37 @@ func (a *BlockAPI) GetBlockReceipts(ctx context.Context, blockNrOrHash rpc.Block
...
+	// Go() until a slot frees, bounding the number of live goroutines rather
+	// than just the number doing concurrent work.
 	allReceipts := make([]map[string]interface{}, len(txHashes))
+	g := new(errgroup.Group)
+	g.SetLimit(maxBlockReceiptsConcurrency)
 	for i, hash := range txHashes {
-		wg.Add(1)
-		go func(i int, hash typedTxHash) {
-			defer wg.Done()
+		g.Go(func() (err error) {
```
bdchatham: Tiny nit — the named err return on this closure isn't doing anything (no naked return, nothing in a defer uses it), so plain func() error would read a bit cleaner.

bdchatham (commented):
Cross-review of the two fixes. The receipts cap looks good; one blocking item on the `eth_subscribe(logs)` change plus a few non-blocking notes, all inline.

evmrpc/subscribe.go
amir-deris: Thanks for great feedback. I will remove the `ctx.Done()`.  Regarding the tests passing, actually there seems to be a tests that is failing: [subscribe test](https://github.com/sei-protocol/sei-chain/actions/runs/27983438458/job/82820662605?pr=3621)
which is a test that is added by @kollegian in this pr [Link](https://github.com/sei-protocol/sei-chain/pull/3610/changes#diff-a31a3ad926664ef796fb7d9d4ca6d524c198141f55652293347acf51e88fbee4:~:text=throw%20new%20Error(%60WsClient.waitFor(%24%7BsubId%7D)%3A%20got%20%24%7Bbuf.length%7D/%24%7Bwant%7D%20notifications%20before%20timeout%60)%3B) 
Thanks to Yasin for improving the test coverage.

sei-will (approved):

evmrpc/block.go
```diff
@@ -333,34 +339,37 @@ func (a *BlockAPI) GetBlockReceipts(ctx context.Context, blockNrOrHash rpc.Block
...
-	// Get tx receipts for all hashes in parallel
-	wg := sync.WaitGroup{}
-	mtx := sync.Mutex{}
+	// Get tx receipts for all hashes in parallel, with a hard cap on the
+	// goroutine fan-out, so a block with a very large number of txs
+	// cannot spawn an unbounded number of goroutines. errgroup.SetLimit blocks
+	// Go() until a slot frees, bounding the number of live goroutines rather
+	// than just the number doing concurrent work.
 	allReceipts := make([]map[string]interface{}, len(txHashes))
+	g := new(errgroup.Group)
```
masih: Use errgroup with context since we have one? 

There is no point waiting in the group if context is cancelled IIUC.

masih (approved):

bdchatham (approved):

PR #3628, opened 2026-06-23
HashLogger

cody-littley:

Creates a new utility called the HashLogger. Purpose of this utility is to allow us to record the hash of each module for each block, as well as other related information. Produces CSV formatted files and rotates files according to configurable policy.

This is first and foremost a debugging utility designed to help us when there is an AppHash. In these situations, triage time is critical, so it's super useful to have this information in an easily digestable format that doesn't require re-execution.

There are two important follow up items to address. These will happen in follow-up PRs once this PR merges.

- integrate the hash logger
- CLI tooling for reading and comparing hash logger archives

unit tests

### DIFF
--- a/sei-db/state_db/sc/hashlog/hash_log_reader.go
+++ b/sei-db/state_db/sc/hashlog/hash_log_reader.go
@@ -1,9 +1,67 @@
 package hashlog
 
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+)
+
+// A hash log file in an archive, paired with its parsed name.
+type archiveFile struct {
+	name   string
+	parsed parsedFileName
+}
+
+// List the hash log files in an archive directory, sorted by file index (ascending).
+func listArchiveFiles(path string) ([]archiveFile, error) {
+	entries, err := os.ReadDir(path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read archive directory %s: %w", path, err)
+	}
+	var files []archiveFile
+	for _, entry := range entries {
+		parsed, ok := parseFileName(entry.Name())
+		if !ok {
+			continue
+		}
+		files = append(files, archiveFile{name: entry.Name(), parsed: parsed})
+	}
+	sort.Slice(files, func(i int, j int) bool {
+		return files[i].parsed.index < files[j].parsed.index
+	})
+	return files, nil
+}
+
 // Looks for a particular block in a hash log archive. May return multiple hashes if the block
-// number in question was executed multiple times (i.e. if there was a rollback).
+// number in question was executed multiple times (i.e. if there was a rollback). Results are returned in
+// archive order (by file index, then by position within a file).
 func ReadHashForBlock(path string, block uint64) ([]*HashLog, error) {
-	return nil, nil
+	files, err := listArchiveFiles(path)
+	if err != nil {
+		return nil, err
+	}
+
+	var result []*HashLog
+	for _, file := range files {
+		// Sealed files carry their block range in the name, so we can skip files that can't contain the block
+		// without reading them.
+		if file.parsed.sealed && (block < file.parsed.firstBlock || block > file.parsed.lastBlock) {
+			continue
+		}
+		parsed, err := ReadHashLogFile(filepath.Join(path, file.name))
+		if err != nil {
+			return nil, err
+		}
+		for i := range parsed.logs {
+			if parsed.logs[i].BlockNumber == block {
+				log := parsed.logs[i]
+				result = append(result, &log)
+			}
+		}
+	}
+	return result, nil
 }
 
 // A set of hashes from two different hash log archives.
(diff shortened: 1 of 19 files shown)


sei-db/state_db/sc/hashlog/hash_logger_config.go
```diff
@@ -0,0 +1,139 @@
...
+	// fewer than BlocksToRetain blocks.
+	MaxDiskSize uint
+}
+
+// DefaultHashLoggerConfig returns a default configuration for a HashLogger.
+func DefaultHashLoggerConfig(path string, version string) *HashLoggerConfig {
+	return &HashLoggerConfig{
+		Path:              path,
+		Version:           version,
+		HashTypes:         []string{"flatKV", "memIAVL", "root"},
```
yzang2019: It's not clear to me what are the valid strings to put here as hashTypes? I understand here it means flatKV lattice hash, memIAVL merkle hash and root app hash. But from user perspective, it's a bit hard to guess what are the valid values to put in the future if we have more types?

sei-db/state_db/sc/hashlog/hash_logger_config.go
```diff
@@ -0,0 +1,139 @@
...
+		DiffHashType:      defaultDiffHashType,
```
yzang2019: What is diffHashType actual for? To detect hash difference between computed hash and expected hash? What will be the other valid values other than the default one?

yzang2019 (approved):
The concurrency model is sound, this is well-designed concurrent pipeline for recording per-block hash digests to CSV-format .hlog files for AppHash divergence debugging. Nice one!

sei-db/state_db/sc/hashlog/hash_logger_impl.go
```diff
@@ -0,0 +1,788 @@
...
+	case ctrlDiffRequest:
+		h.handleDiffRequest(msg.blockNumber, msg.cs)
+	}
+}
+
+// handleHashReport records a caller-reported hash, discarding it if the block has already been flushed.
+func (h *hashLoggerImpl) handleHashReport(blockNumber uint64, hashType string, hash []byte) {
+	if h.hasFlushedAtLeastOnce && blockNumber <= h.flushedHighWater {
+		return // already on disk: a duplicate/late report, or a re-execution without reopening the logger
+	}
```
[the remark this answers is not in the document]
cody-littley: This is intentional behavior. The hash logger is a best effort sort of thing, and if a block has already been flushed to disk we want to skip a late hash report for that block.

sei-db/state_db/sc/hashlog/hash_logger_config.go
cody-littley: This actually shouldn't have a default, those should all be provided by the constructor. Defaults removed.

The short answer is that any string is legal, with the following conditions:

- any string that contains alphanumeric characters and the special characters `_`, `-`, or `.`.
- any string other than `diff` (this is reserved)

The idea is that the caller decides on the naming schema, then reports hashes. For example, we might adopt a format like this:

- for memIAVL module hashes: `memiavl/MODULE_NAME`
- for flatKV modules: `flatkv/MODULE_NAME`
- for the flatKV DB hashes: `flatkv/db/account`, `flatkv/db/code`, `flatkv/db/legacy`, etc.
- for the block hash: `block`

Internally, the hashlogger doesn't really care about naming as long as you provide all of the correct modules at construction time. If you define a module and forget to report hashes for that module, it's not the end of the world. The hashLogger will eventually write the block to disk, but with an empty space for that particular column.

sei-db/state_db/sc/hashlog/hash_logger_config.go
cody-littley: The diff hash is a special case. It's a non-cryptographic hash of the change sets passed to `ApplyChangeSets()`. I chose a hashing algorithm that was fast, so I don't think it will be a problem.

If there is ever an apphash divergence, it will be useful to know the following things:

- the block hash
- the diff hash
- the DB hash

If we know all of these things, it will help us figure out what part of the system caused the nondeterministic behavior.

blindchaser (approved):

sei-db/state_db/sc/hashlog/hash_logger_impl.go
```diff
@@ -0,0 +1,784 @@
...
+}
+
+// emit writes a single block to the writer and records it as flushed.
+func (h *hashLoggerImpl) emit(blockNumber uint64) {
+	log := h.pendingBlocks[blockNumber]
+	delete(h.pendingBlocks, blockNumber)
+	delete(h.blocksWithPendingHashes, blockNumber)
+	h.blockingSendToWriter(writerMessage{log: log})
+	h.flushedHighWater = blockNumber
+	h.hasFlushedAtLeastOnce = true
```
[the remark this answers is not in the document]
cody-littley: This is a false positive. Context can only be cancelled during terminal shutdown.

sei-db/state_db/sc/hashlog/hash_log_reader.go
```diff
@@ -0,0 +1,376 @@
...
+func (r *archiveReader) loadFile(meta fileMeta) (*loadedFile, error) {
+	file, err := ReadHashLogFile(filepath.Join(r.dir, meta.name))
+	if err != nil {
+		return nil, err
+	}
+	byBlock := make(map[uint64][]*HashLog)
+	for i := range file.logs {
+		log := file.logs[i]
+		byBlock[log.BlockNumber] = append(byBlock[log.BlockNumber], &log)
+	}
```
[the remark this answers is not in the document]
cody-littley: False positive. Wrote a unit test to exercise this exact scenario, and code works as intended.

PR #3629, opened 2026-06-23
revisited autobahn msg buffers

pompon0:
I have separated the message types for the signed messages, so that wireguard can estimate the maximal sized with better precision. I have added init-time checks for the giga.API verifying that buffer sizes are large enough to fit largest messages. The limits are hardcoded (rather than derived from MaxSize) so that buffer sizes cannot silently grow due to proto updates (i.e. limits need to be bumped manually).


sei-will (approved):

sei-tendermint/autobahn/types/consensus.go
```diff
@@ -42,11 +42,11 @@ var ConsensusReqConv = protoutils.Conv[ConsensusReq, *pb.ConsensusReq]{
 			}
 		case *ConsensusReqPrepareVote:
 			return &pb.ConsensusReq{
-				T: &pb.ConsensusReq_PrepareVote{PrepareVote: SignedMsgConv[*PrepareVote]().Encode(m.Signed)},
+				T: &pb.ConsensusReq_PrepareVoteV2{PrepareVoteV2: SignedPrepareVoteConv.Encode(m.Signed)},
```
wen-coding, line 45: dumb question: we haven't launched autobahn yet, can we just replace?

bdchatham (approved):

sei-tendermint/autobahn/types/consensus.go
pompon0: we can, but that would require disabling buf breaking CI for this pr, so why bother?

wen-coding (approved):

PR #3631, opened 2026-06-23
fix(evmrpc): Cap calls in eth_estimateGasAfterCalls

amir-deris:

`eth_estimateGasAfterCalls` replays every entry in its `calls` array
through a full EVM execution before estimating gas for the final transaction.
The `calls` slice was previously unbounded, so a single request could drive
`len(calls) × gasCap` of EVM work — an issue that the
per-request rate limiter does not bound.

This PR validates `len(calls) <= max_estimate_gas_calls` at the very top of the
handler, **before** the rate limiter and any state wrapping, so oversized
requests are rejected early without acquiring a limiter token or loading state.

Changes:
- Add `max_estimate_gas_calls` to the EVM RPC config (`mapstructure` tag, flag
  `evm.max_estimate_gas_calls`, parser entry, and `app.toml` template line),
  defaulting to **100**.
- Thread the value through `SimulateConfig` (both constructions in `server.go`)
  and expose it via `Backend.MaxEstimateGasCalls()`.
- Reject over-sized `EstimateGasAfterCalls` requests with
  `eth_estimateGasAfterCalls: too many calls (%d > %d)`. A value of `0` means
  unlimited (explicit opt-out).

Upgrade safety: `ReadConfig` starts from `DefaultConfig` and only overrides a
field when the key is present, so a node upgraded with an existing `app.toml`
that lacks the key keeps the default of 100 — the cap is not silently disabled.

- `TestEstimateGasAfterCallsMaxCalls` (new): over-cap request is rejected with
  the "too many calls" error; an at-cap request passes the size guard.
- Updated `config_test.go` so `TestReadConfig` recognizes the new key.
- `go test ./evmrpc/ ./evmrpc/config/...` passes; `gofmt` clean.

### DIFF
--- a/evmrpc/simulate.go
+++ b/evmrpc/simulate.go
@@ -131,6 +131,11 @@ func (s *SimulationAPI) EstimateGasAfterCalls(ctx context.Context, args export.T
 	defer func() {
 		recordMetricsWithError(ctx, "eth_estimateGasAfterCalls", s.connectionType, startTime, returnErr, recover())
 	}()
+	// Reject over-sized requests early, before any state wrapping or resource acquisition.
+	if maxCalls := s.backend.MaxEstimateGasCalls(); maxCalls > 0 && len(calls) > maxCalls {
+		returnErr = fmt.Errorf("eth_estimateGasAfterCalls: too many calls (%d > %d)", len(calls), maxCalls)
+		return
+	}
 	/* ---------- fail‑fast limiter ---------- */
 	if s.requestLimiter != nil {
 		if !s.requestLimiter.TryAcquire(1) {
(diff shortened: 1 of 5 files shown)


evmrpc/simulate.go
```diff
@@ -131,6 +131,11 @@ func (s *SimulationAPI) EstimateGasAfterCalls(ctx context.Context, args export.T
 	defer func() {
 		recordMetricsWithError(ctx, "eth_estimateGasAfterCalls", s.connectionType, startTime, returnErr, recover())
 	}()
+	// Reject over-sized requests early, before any state wrapping or resource acquisition.
```
bdchatham, line 134: Nice

bdchatham (approved):
LGTM

sei-will (approved):

PR #3638, opened 2026-06-24
Fix panic during state sync

cody-littley:

Fixes an issue where a node could be forced to panic during a state sync


cody-littley:
Closing, will not merge

PR #3640, opened 2026-06-24
fix(metrics): Prometheus metrics output

m3diumrare:
After the `sei-chain` to `sei_chain` prefix change in #3554, it surfaced that otel labels also had invalid Prometheus label characters ('.')

Tested on local Sei node, with this patch the Prometheus remote write endpoint no longer drops any metrics from the node.

Example before:
```
curl -si -H "Accept: application/openmetrics-text;version=1.0.0;escaping=allow-utf-8" http://127.0.0.1:26660/metrics | grep 'telemetry.sdk.language'
target_info{"service.name"="unknown_service:seid","telemetry.sdk.language"="go","telemetry.sdk.name"="opentelemetry","telemetry.sdk.version"="1.39.0"} 1
```

after:
```
 curl -si -H "Accept: application/openmetrics-text;version=1.0.0;escaping=allow-utf-8" http://127.0.0.1:26660/metrics | grep 'telemetry.sdk.language'
target_info{service_name="unknown_service:seid",telemetry_sdk_language="go",telemetry_sdk_name="opentelemetry",telemetry_sdk_version="1.39.0"} 1

```


m3diumrare:
@philipsu522 this would be great to get into 6.6 release if possible

m3diumrare:
Hi @Kbhat1 - is there a process for becoming a contributor? I'd like to get this reviewed and merged, and I'm concerned because my last PR fell by the wayside. I'm thinking the `Contributor` tag makes a difference in reviews here? Thanks

Also this is non-breaking for app hash, but I can't add that label.

bdchatham (approved):

PR #3641, opened 2026-06-24
feat(grpc): add keepalive, LimitListener, and MaxRecvMsgSize to gRPC server :9090 (PLT-705)

amir-deris:
## Summary

The gRPC server at `:9090` was created with only `grpc.MaxConcurrentStreams(100)` and an unbounded `net.Listen`. It had no cap on connection count, no bound on inbound message size, and no keepalive policy. This PR adds all three and exposes every parameter as an operator config field under `[grpc]` in `app.toml`. (PLT-705)

### MaxRecvMsgSize

Sets `grpc.MaxRecvMsgSize()` explicitly (default **4 MB**, matching gRPC's own default). This bounds per-request memory allocation *before* the rate limiter fires, so an oversized request can't allocate first and rate-limit second.

### MaxOpenConnections

Wraps the listener with `netutil.LimitListener` to cap simultaneously-open TCP connections (mirrors the gRPC-Web change in #3605 and the EVM `LimitListener` pattern). Keepalive alone bounds neither connection count nor message size — this is the actual DoS bound.

- Default: **1000** (matches the API and gRPC-Web defaults)
- Set to `0` to disable the cap.

### Keepalive

Adds `keepalive.ServerParameters` and `keepalive.EnforcementPolicy`. Defaults mirror gRPC's own (i.e. opt-in, no behavior change) with **one deliberate exception**:

| Field | Default | Rationale |
|---|---|---|
| `MaxConnectionIdle` | **5m** | Bounded by default — reclaims abandoned connection slots, which matter now that the listener is capped. Only closes connections with **zero in-flight RPCs**, so it never interrupts active work. 5m (not 30s) avoids churning clients that poll on a sub-minute cadence. |
| `MaxConnectionAge` | 0 (∞) | Applies to *active* connections too and would cut long-lived streams; its main benefit (LB rebalancing) is topology-specific. Left off by default, exposed for fleet operators. |
| `MaxConnectionAgeGrace` | 0 (∞) | Paired with `MaxConnectionAge`; meaningless until age is set. |
| `Time` | 2h | gRPC default. |
| `Timeout` | 20s | gRPC default. |
| `EnforcementPolicy.MinTime` | 5m | gRPC default. |
| `EnforcementPolicy.PermitWithoutStream` | false | gRPC default. |

`MaxSendMsgSize` is intentionally **not** added — PageGuard already bounds row count at the query layer. Revisit if large unpaginated responses surface in the handler inventory.

### Config plumbing

All fields are exposed under `[grpc]` in `app.toml` and wired through `DefaultConfig()` and `GetConfig()`. For the bounded defaults (`max-recv-msg-size`, `max-open-connections`, `max-connection-idle`, and the keepalive durations), `GetConfig` applies the in-code default when the key is **absent**, so a node upgrading with an older `app.toml` stays bounded rather than reverting to unlimited/infinity.

**Files changed:** `sei-cosmos/server/grpc/server.go`, `sei-cosmos/server/config/config.go`, `sei-cosmos/server/config/toml.go`, `sei-cosmos/server/start.go`, `sei-cosmos/testutil/network/util.go`

### Tests

- `TestDefaultGRPCConfig` — asserts all new defaults.
- `TestGetConfigGRPCDefaultsWhenAbsent` — a legacy `app.toml` (missing the new keys) still resolves to the bounded in-code defaults, including the 5m idle default.
- `TestGetConfigGRPCOverrides` — operator-provided values override the defaults.

## Test plan

- [ ] `go test ./sei-cosmos/server/config/... ./sei-cosmos/server/grpc/...` passes
- [ ] `gofmt -s -l .` prints nothing
- [ ] `make build` succeeds


sei-will (approved):

masih:
@amir-deris this one needs a conflict resolution

bdchatham (approved):

PR #3642, opened 2026-06-24
feat(inprocess): in-process N-validator harness

bdchatham:
## in-process N-validator harness

Stands up N sei-chain validators in one Go process — real CometBFT consensus, each node serving its own Tendermint RPC + EVM JSON-RPC (HTTP/WS), with deterministic teardown. The in-process foundation for the SDK "local" provider, and the compute backend for running the YAML query/tx suites in-memory (no docker).

Gated behind the `inprocess` build tag, so the heavy sei-tendermint/sei-cosmos bring-up never enters a normal `seid` build. The only production `app.App` change is retaining the two EVM server handles for teardown (harness-only accessors live in `app/app_inprocess.go` behind the tag); `app.App`'s public surface doesn't widen and the serve goroutines keep their original bare-panic path.

### What's here
- `inprocess/` — `Start(ctx, Options) (*Network, error)`, per-node `Node` handles, `WaitReady`, idempotent `Close`. Handles mirror the SDK `sei.NodeHandle`/`NetworkHandle` by shape (no SDK import — toolchain skew would break the seid build), so a future thin adapter satisfies the interface structurally.
- `integration_test/runner/` — a pluggable `execer` seam: docker stays the default; `WithInProcessNetwork` (build-tagged) runs the YAML suites in-memory.

### Validator count: 1 or ≥ 3 (never 2)
Driven by CometBFT's block-sync→consensus handoff, not a quorum: N=1 proposes solo; N=2 deadlocks (each node has one peer, `IsCaughtUp` needs >1) so `Start` rejects it; N≥3 works. The bring-up invariants (empty valset, gentx-derived peer mesh, EVM-enable, metrics-off, loopback scope) are documented at point-of-use in `inprocess/doc.go`.

### Served surface
TM RPC + EVM JSON-RPC HTTP/WS. No cosmos gRPC (the harness never starts it). REST is an honest parity stub for the SDK handle shape.

### Tests
`TestInProcessNetwork` (N=4) asserts each node serves TM RPC + EVM and round-trips a tx (broadcast on node 0, observed on node 1); plus input-rejection and chain-id guards, and `TestInProcessBankModule` running the bank suite in-memory.

```
go test -tags inprocess ./inprocess/ ./integration_test/runner/
```

### Reviewer notes
- The EVM worker pool is a process-wide singleton; `Close` deliberately doesn't close it (its `sync.Once` won't re-fire), so one network runs per process today. De-globalizing it in `evmrpc` is the fix if repeated Start/Close is ever needed.
- Not yet wired into CI: the harness runs on-demand (`-tags inprocess`) and the pipeline still runs the docker suites. CI adoption is a deliberate follow-up.


bdchatham:
bugbot run

bdchatham:
bugbot run

bdchatham:
bugbot run

bdchatham:
bugbot run

bdchatham:
bugbot run

bdchatham:
bugbot run

bdchatham:
bugbot run

bdchatham:
bugbot run

inprocess/handle.go
```diff
@@ -0,0 +1,162 @@
...
+// the SDK's sei.NodeHandle.WaitReady; the probe HTTP client is an internal
+// default (probeClient).
+func (h Node) WaitReady(ctx context.Context) error {
+	if err := waitHeightAdvances(ctx, probeClient, h.TendermintRPC(), 1); err != nil {
+		return fmt.Errorf("%s tendermint: %w", h.n.moniker, err)
+	}
+	if err := waitEVMServing(ctx, probeClient, h.EVMRPC()); err != nil {
+		return fmt.Errorf("%s evm: %w", h.n.moniker, err)
+	}
+	return nil
```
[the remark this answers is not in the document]
bdchatham: Obsolete — the `ServeErr` channel this references was intentionally removed when the EVM serve-error diversion was dropped (commit 4e143fc). Bind failures now panic the serve goroutine (the production fail-loud path), and `grep -r ServeErr` is empty on HEAD. No action needed; marking resolved.

integration_test/runner/runner_inprocess.go
```diff
@@ -0,0 +1,197 @@
...
+		}
+
+		shim := filepath.Join(dir, "seid")
+		script := shimScript(realBin)
+		if werr := os.WriteFile(shim, []byte(script), 0o700); werr != nil { //nolint:gosec
+			e.setup = werr
+			return
+		}
+	})
+	return e.setup
```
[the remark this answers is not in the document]
bdchatham: Fixed in 06a57e3. The shared `seid` build now runs via a `backendPreparer` seam in `RunFile` against the **parent** test, before the `t.Run` subtest loop — so its `t.Cleanup` (and `binDir`) outlive every case. Good catch; this would have bitten the first multi-case suite.

integration_test/runner/runner_inprocess.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// shimScript builds the `seid` shim. It always prepends --home (a root
+// persistent flag) and appends --node only for client subcommands (query/q/tx/
```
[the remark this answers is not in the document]
bdchatham: Fixed in 06a57e3 — the comment now says "prepend", and the shim is rewritten to a two-branch quoted `exec` that prepends both `--home` and `--node` explicitly. Thanks for the precise catch.

integration_test/runner/runner_inprocess.go
```diff
@@ -0,0 +1,197 @@
...
+		dir, err := os.MkdirTemp("", "sei-inprocess-bin-")
+		if err != nil {
+			e.setup = err
+			return
+		}
+		e.binDir = dir
+		// Remove the build dir at test end so repeated runs don't accrue a binary
+		// per run.
+		t.Cleanup(func() { _ = os.RemoveAll(dir) })
+
```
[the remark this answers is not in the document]
bdchatham: Addressed in 06a57e3 — the shim now passes args positionally and quoted (`--home "$SEID_HOME"` …), dropping the unquoted word-splitting, so a home path containing a space is safe. The `$1`-as-subcommand assumption remains (the suites use the `seid <subcommand> …` form) and is now stated explicitly in the shim doc.

inprocess/appoptions.go
```diff
@@ -0,0 +1,41 @@
...
+	chainID  string
+	httpPort int
+	wsPort   int
+}
+
+func (o appOptions) Get(key string) interface{} {
+	switch key {
+	case "chain-id":
+		return o.chainID
+	case "evm.http_enabled":
```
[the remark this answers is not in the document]
bdchatham: Good catch — confirmed and fixed in the latest commit. Verified the path: `NewEVM*Server` → `InitRPCTracker/InitWSTracker(ctx, RPCStatsInterval)`, which guards `if interval > 0` and `go t.run()`s on a reassigned package global that `Stop` never cancels. Rather than document it, `appOptions.Get("evm.rpc_stats_interval")` now returns `time.Duration(0)`, so the tracker is skipped entirely (the harness needs no RPC-stats reporting) — teardown is deterministic with zero orphaned tracker goroutines. The worker-pool singleton remains the one documented per-process caveat.

inprocess/appoptions.go
```diff
@@ -0,0 +1,41 @@
+//go:build inprocess
+
```
amir-deris, line 2: Should the name of this file be `app_options.go`? (`appoptions` seems not standard)

app/app.go
```diff
@@ -478,6 +478,14 @@ type App struct {
 	httpServerStartSignalSent bool
 	wsServerStartSignalSent   bool
 
+	// evmHTTPServer/evmWSServer hold the EVM JSON-RPC HTTP and WebSocket listeners
+	// constructed in RegisterLocalServices so an embedding orchestrator (the
+	// in-process harness) can Stop() them at teardown. Nil when the respective
+	// listener is disabled. Production seid does not read these; its process exit
+	// reaps the listeners.
+	evmHTTPServer evmrpc.EVMServer
```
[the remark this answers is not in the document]
bdchatham: Confirmed clean. `gofmt -s -l` is empty on both files, and the untagged build passes lint in CI — both the `Go / Lint` and `lint` jobs are green on this commit. They compile without the `inprocess` tag (so the accessors in app_inprocess.go aren't seen), yet staticcheck U1000 does not flag the write-only fields, and no `//nolint` is needed. The fields' doc already notes production seid doesn't read them (its process exit reaps the listeners). Resolving as verified.

integration_test/runner/runner_inprocess.go
```diff
@@ -0,0 +1,210 @@
...
+// Assumes the subcommand is $1 (the suites' `seid <subcommand> …` form). Args are
+// quoted and passed positionally, so a home path containing a space is safe.
+func shimScript(realBin string) string {
+	return "#!/bin/sh\n" +
+		"case \"$1\" in\n" +
+		"  q|query|tx|status)\n" +
+		"    exec \"" + realBin + "\" --home \"$SEID_HOME\" --node \"$SEID_NODE\" \"$@\" ;;\n" +
+		"  *)\n" +
+		"    exec \"" + realBin + "\" --home \"$SEID_HOME\" \"$@\" ;;\n" +
+		"esac\n"
```
amir-deris: This script can be written using go literal string (with back tick) to make it easier to read:
fmt.Sprintf(`#!/bin/sh ....

inprocess/harness.go
```diff
@@ -0,0 +1,534 @@
...
+// bare `seid` with no --keyring-backend flag, so keyring-backend=test is resolved
+// from this file (the shim only injects --home and --node). A failed write would
+// silently fall the keyring back to the OS default and break signing — so the
+// error is returned, not swallowed.
+func writeClientConfig(path, chainID, rpcAddr string) error {
+	return os.WriteFile(path, []byte(fmt.Sprintf(clientConfigTemplate, chainID, rpcAddr)), 0o600)
+}
+
+// freePort allocates a free loopback TCP port via server.FreeTCPAddr.
+func freePort() (int, error) {
```
amir-deris, line 557: **Port allocation is racy — likely flake source.** `server.FreeTCPAddr` probes `net.Listen("tcp", "localhost:0")` then closes, but the harness binds the returned port on `tcp://127.0.0.1:PORT` (TM RPC/P2P) and `0.0.0.0:PORT` (EVM). Two gaps: (a) on a dual-stack host where `localhost` resolves to `::1`, the port is verified free on IPv6 but bound on IPv4 (independent namespaces); (b) `freePort` is called 4×N times, each close-then-rebind, so two probes can return the same port intra-process. Either yields a `node.Start` error or a panic in the EVM serve goroutine → flaky `WaitReady` timeout. `doc.go` acknowledges only the external-process TOCTOU. A single hold-open allocator (keep the `net.Listener` and hand its fd to the binder) closes the window.

inprocess/handle.go
```diff
@@ -0,0 +1,171 @@
...
+	// evmrpc is the proper fix for repeated Start/Close in one process.
+
+	if net.ownBaseDir && net.baseDir != "" {
+		_ = os.RemoveAll(net.baseDir)
+	}
+}
+
+// stopNode shuts one node's tendermint service and EVM listeners. Each step is
+// guarded so a nil (never-started) field on a partial-start path is a no-op.
+func stopNode(n *node) {
```
amir-deris, line 150: **EVM serve goroutine leaks when a node never produces its first block.** The goroutine from `RegisterLocalServices` parks on `<-app.httpServerStartSignal`, which only fires from the first `ProcessBlock` (app.go:1922). On a partial-start or `WaitReady`-timeout teardown, `stopNode` stops `tmNode` first (so no block ever commits, signal never sent) and `EVMHTTPServer().Stop()` is a no-op (listener nil) — the goroutine and its app references leak until process exit. There's also a narrow stop-before-start race: if `Close` runs after the signal fired but before `Start()` binds, `Stop()` no-ops and the listener then binds and leaks. Reaped at process exit, so low severity for one run, but it compounds the one-network-per-process constraint.

inprocess/handle.go
```diff
@@ -0,0 +1,171 @@
...
+func (net *Network) Close() {
+	if net.closed {
+		return
+	}
+	net.closed = true
+
+	for _, n := range net.nodes {
+		stopNode(n)
+	}
+	// The EVM worker pool (evmrpc.GetGlobalWorkerPool) is a process-wide
```
amir-deris: **"One network per process" rests on more singletons than documented.** `doc.go` names only `evmrpc.GetGlobalWorkerPool`, but EVM `HTTPServer.Stop()` also calls the process-wide `StopMetricsPrinter()` (`metricsStopOnce sync.Once`, evmrpc/worker_pool_metrics.go:341), which can never restart. A second `Start` in the same binary inherits both a closed pool and a dead metrics printer, and nothing in `Start` guards against it — the limitation is documentation-only and silently violated. Consider a package-level guard in `Start` that fails loudly on a second concurrent network, or extend the doc caveat to cover the metrics printer.

integration_test/runner/runner_inprocess.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// shimScript builds the `seid` shim. Rule: always prepend --home; prepend --node
+// only for client subcommands (q/query/tx/status). Passing --node to a non-client
+// subcommand (keys, etc.) fails cobra flag parsing — only the client leaves
+// register it (AddQueryFlagsToCmd / AddTxFlagsToCmd / StatusCommand).
+//
+// Assumes the subcommand is $1 (the suites' `seid <subcommand> …` form). Args are
+// quoted and passed positionally, so a home path containing a space is safe.
+func shimScript(realBin string) string {
```
amir-deris, line 192: **`--node` injection keyed on a hardcoded subcommand allowlist is brittle.** Only `q|query|tx|status` as `$1` get `--node`. Today's suites (`q`/`tx`/`status`/`keys`) are safe, but a command with a leading global flag (`seid --output json q ...`) or any future RPC-reading subcommand falls into the `*)` branch with no `--node` and silently targets the default RPC (`:26657`) instead of the node's ephemeral port — a confusing downstream verifier failure, not an error at the seam. This re-encodes cobra's per-command flag knowledge as a hand-maintained case statement.

integration_test/runner/runner_inprocess.go
```diff
@@ -0,0 +1,210 @@
...
+	// integration_test/utils/_tx_helpers.sh` resolves (docker runs with the repo
+	// mounted at the container CWD; `go test` runs with CWD = the package dir).
+	root, err := repoRoot()
+	if err != nil {
+		return "", err
+	}
+	c.Dir = root
+	c.Env = append(os.Environ(), envMapSlice(envMap)...)
+	c.Env = append(c.Env,
+		"PATH="+e.binDir+string(os.PathListSeparator)+os.Getenv("PATH"),
```
amir-deris, line 80: **`binDir` read unsynchronized; the no-`t.Parallel` contract is unenforced.** `ensureBin` writes `binDir` inside `sync.Once.Do`; `run` reads it here without synchronization, safe only because cases run serially. The comment at ensureBin states this, but nothing enforces it — adding `t.Parallel()` to a case later races the build/read and fails as an opaque "seid not found."

inprocess/appoptions.go
```diff
@@ -0,0 +1,51 @@
...
+		return true
+	case "evm.ws_port":
+		return o.wsPort
+	case "evm.rpc_stats_interval":
+		// Disable the stats tracker. A positive interval (the unset default is 10s)
+		// makes each EVM server spawn a reporter goroutine on a package-global that
+		// EVMServer.Stop never cancels and that's reassigned per node — so N nodes
+		// would orphan N-1. 0 skips it, keeping teardown deterministic.
+		return time.Duration(0)
+	case app.FlagSCEnable:
```
amir-deris, line 44: **Cleanup: duplicates four SeiDB flag answers verbatim from `app.TestAppOpts`.** `FlagSCEnable`/`FlagSCSnapshotInterval`/`FlagSSEnable`/`FlagSSBackend` return the exact values in `app/test_helpers.go`. Embedding `app.TestAppOpts` and overriding only the EVM-enable keys + chain-id would drop these 8 lines, which otherwise silently drift if the SeiDB defaults change.

inprocess/readiness.go
```diff
@@ -0,0 +1,144 @@
...
+		Result *struct {
+			SyncInfo syncInfo `json:"sync_info"`
+		} `json:"result,omitempty"`
+		SyncInfo syncInfo `json:"sync_info"`
+	}
+	if json.Unmarshal(body, &s) != nil {
+		return 0, false
+	}
+	si := s.SyncInfo
+	if s.Result != nil && s.Result.SyncInfo.LatestBlockHeight != "" {
```
amir-deris, line 101: **Cleanup: dead `/status` branch + unused field.** sei-tendermint always envelopes the result under `result`, so the unwrapped top-level `SyncInfo` field and the `si := s.SyncInfo` default are never reached — the `s.Result != nil` branch always wins. `syncInfo.CatchingUp` is also parsed but never used. The "both shapes the Sei fork emits" comment is misleading; collapsing to the enveloped shape simplifies.

amir-deris (commented):
## Review — in-process N-validator harness

Thorough, well-documented PR. The production-facing change (`app.App` retaining two EVM server handles) is clean and additive; the docker-arm refactor in `runner.go` is byte-for-byte behavior-preserving; `gofmt -s`, `go vet -tags inprocess`, and errcheck are all clean; and the genesis/client.toml reimplementations are justified (the cosmos helpers they mirror are genuinely unexported).

All findings below are in the new `inprocess`-tagged code — none block the production path. The two I'd act on before merge are the **port-allocation flake** and a **`Start` guard for the one-network-per-process limitation** so it fails loudly instead of silently misbehaving.

**CI coverage gap (not a rule violation):** `.golangci.yml` uses `build-tags: [codeanalysis]` / `tests: false`, so none of the new `//go:build inprocess` code is linted in CI, and the PR notes it isn't in the test pipeline either — so this code currently has no automated lint or test gate. If the harness is going to be relied on, an `-tags inprocess` lint+test job is the deeper fix.

Inline comments follow, most-severe first.

amir-deris (approved):

bdchatham:
Thanks for the thorough review, Amir — all addressed in `b55917c` (also ran it through our standards-champion review). Dispositions:

**Acted on**
- **Port-allocation flake** — `freePort` now probes `127.0.0.1` + tracks a process-wide allocated set, closing the dual-stack `::1` mismatch and intra-process reuse; residual cross-process TOCTOU documented.
- **One-network-per-process guard** — a `networkStarted` one-shot in `Start` fails loudly on a second network. Also documented the `StopMetricsPrinter` singleton you flagged alongside the worker pool.
- **`--node` allowlist brittleness** — the shim now errors (exit 64) on a leading global flag instead of silently targeting `:26657`; the allowlist is documented as the maintained contract.
- **Shim raw-string literal**, **serve-goroutine leak / `binDir` unsync** — done / documented as caveats tied to the one-network limit.

**Rejected (with evidence)**
- **Dead `/status` branch** — the unwrapped branch is the *live* path: the in-process node's HTTP `/status` returns the unwrapped shape, and an enveloped-only parse hangs `WaitReady`. Confirmed by bisect (enveloped-only → `TestInProcessNetwork` times out in `WaitReady`; dual-shape → 18s pass). Kept both; did drop the genuinely-unused `CatchingUp` field per your note.

**Declined (with reason)**
- **Embed `app.TestAppOpts`** — good catch that it dedups the SeiDB flags, but embedding silently flips the app off the production **Giga execution engine**: an unset giga flag defaults Giga **ON** (prior harness + production), while `TestAppOpts`' zero value forces Giga **OFF**. Pinned the SeiDB flags explicitly instead, with a forward-guard comment.

**Your CI-coverage-gap point is the most important one** and exactly right: the `inprocess`-tagged code isn't linted or run in CI, and that gap let *both* the giga flip and a readiness regression pass CI green this round — only the local test caught them. Filing a follow-up to wire an `-tags inprocess` lint+test job.

inprocess/harness.go
```diff
@@ -0,0 +1,572 @@
...
+	// input is validated, so a rejected N never burns it. Not released on failure —
+	// a partially built network may already have touched the EVM singletons.
+	if !networkStarted.CompareAndSwap(false, true) {
+		return nil, fmt.Errorf("inprocess: a network was already started in this process; only one is supported (EVM worker pool / metrics printer / Prometheus registries are process-global singletons)")
+	}
+
+	baseDir, ownBaseDir, err := resolveBaseDir(opts.BaseDir)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
bdchatham: Fixed in 0c875ce — moved the slot claim (CompareAndSwap) to just before the first `app.New` in the startNode loop, so a pre-singleton failure (bad base dir, out-of-range ExtraKey, genesis error) no longer burns the slot. Thanks both.

inprocess/harness.go
```diff
@@ -0,0 +1,572 @@
...
+	// >1 peer to ever report caught-up, so neither node leaves block-sync. Reject it
+	// loudly rather than hang. N=1 (solo proposer via onlyValidatorIsUs) and N>=3
+	// (>=2 peers each) both work — see startNode and doc.go.
+	if opts.Validators == 2 {
+		return nil, fmt.Errorf("inprocess: Options.Validators == 2 deadlocks in CometBFT block-sync (BlockPool.IsCaughtUp requires >1 peer); use 1 or >= 3")
+	}
+	// One network per process (see networkStarted): claim the slot only after the
+	// input is validated, so a rejected N never burns it. Not released on failure —
+	// a partially built network may already have touched the EVM singletons.
+	if !networkStarted.CompareAndSwap(false, true) {
```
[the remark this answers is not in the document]
bdchatham: Fixed in 0c875ce — moved the slot claim (CompareAndSwap) to just before the first `app.New` in the startNode loop, so a pre-singleton failure (bad base dir, out-of-range ExtraKey, genesis error) no longer burns the slot. Thanks both.

inprocess/appoptions.go
bdchatham: Declined with reason — embedding `app.TestAppOpts` also adopts its giga-OFF default, which flips the in-process app off the production **Giga execution engine** (an unset giga flag defaults Giga ON). Pinned the SeiDB flags explicitly with a forward-guard comment instead (b55917c). Good catch on the dup — it just isn't safe to dedup via TestAppOpts.

inprocess/readiness.go
bdchatham: Rejected with evidence — the unwrapped branch is the *live* path: the in-process node's HTTP /status returns the unwrapped shape, and an enveloped-only parse hangs WaitReady (bisect: enveloped-only → `TestInProcessNetwork` times out; dual-shape → 18s pass). Kept both shapes; did drop the genuinely-unused `CatchingUp` field per your note.

inprocess/handle.go
bdchatham: Documented as a bounded caveat in `stopNode` (b55917c): the parked serve goroutine is reaped at process exit and bounded by the one-network-per-process guard. The proper fix (a cancellable serve goroutine) touches production `app.go`'s RegisterLocalServices lifecycle, so it's deferred rather than done in this harness PR.

inprocess/handle.go
```diff
@@ -0,0 +1,179 @@
...
+// it compounds that limit rather than breaking teardown.
+func stopNode(n *node) {
+	if n.tmNode != nil && n.tmNode.IsRunning() {
+		n.tmNode.Stop()
+		n.tmNode.Wait()
+	}
+	if n.app != nil {
+		stopEVMServer(n.app.EVMHTTPServer())
+		stopEVMServer(n.app.EVMWebSocketServer())
+	}
```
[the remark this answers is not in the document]
bdchatham: Documented as a bounded caveat in `stopNode` (b55917c): the parked serve goroutine is reaped at process exit and bounded by the one-network-per-process guard. The proper fix (a cancellable serve goroutine) touches production `app.go`'s RegisterLocalServices lifecycle, so it's deferred rather than done in this harness PR.

inprocess/handle.go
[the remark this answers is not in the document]
bdchatham: Documented as a bounded caveat in `stopNode` (b55917c): the parked serve goroutine is reaped at process exit and bounded by the one-network-per-process guard. The proper fix (a cancellable serve goroutine) touches production `app.go`'s RegisterLocalServices lifecycle, so it's deferred rather than done in this harness PR.

PR #3643, opened 2026-06-25
fixed the header number limit for FullCommitQC (PLT-764)

pompon0:
Moved the limit on headers per commitQC to autobahn types, to make it an explicit API property. Added missing tests on types with wireguards limits.


sei-tendermint/internal/autobahn/autobahn.proto
```diff
@@ -157,7 +157,7 @@ message CommitQC {
 message FullCommitQC {
   option (wireguard.sized) = true;
   CommitQC qc = 1;
-  repeated BlockHeader headers = 2 [(wireguard.max_count) = 100];
+  repeated BlockHeader headers = 2 [(wireguard.max_count) = 1000];
```
wen-coding, line 160: Oops, thanks for fixing this. Should we add comment here about how 1000 was calculated?

wen-coding (approved):

sei-tendermint/internal/autobahn/autobahn.proto
pompon0: we have now a test which validates this limit

PR #3648, opened 2026-06-25
feat(evmrpc): pre-decode request size admission control (PLT-295)

amir-deris:
## What

Adds pre-decode admission control for HTTP EVM JSON-RPC requests, bounding the
peak memory that concurrent large simulation payloads (`eth_call`,
`eth_estimateGas`, `debug_trace*`, …) can allocate while their bodies are
buffered and JSON-decoded.

A new middleware (`requestSizeLimiter`) wraps the HTTP handler and enforces two
limits **before** the body is read or decoded:

- **Per-request body cap** — requests whose declared `Content-Length` exceeds
  the limit are rejected with `413 Request Entity Too Large`. The body reader is
  also wrapped in `http.MaxBytesReader`, so chunked / mis-declared bodies cannot
  exceed the cap either.
- **Global concurrent-bytes budget** — a size-weighted semaphore bounds the
  total request bytes admitted for processing at once. Requests that would
  exceed the budget are rejected fast with `429 Too Many Requests`.

## Config

Two new `evm` config options (with sensible defaults):

| Option | Default | Meaning |
| --- | --- | --- |
| `max_request_body_bytes` | `5 MiB` | Per-request body cap. `0` = go-ethereum default (5 MiB). Also wired into the inner go-ethereum server via `SetHTTPBodyLimit`. |
| `max_concurrent_request_bytes` | `128 MiB` | Total request bytes admitted concurrently. `0` = disabled. Raised to `max_request_body_bytes` if set smaller, so one max-size request always fits. |

## Observability

New OTel counter `evmrpc_requests_rejected_total`, labeled by `reason`
(`oversize` / `busy`). No endpoint dimension — rejection happens before the
JSON-RPC method is decoded.

## Tests

- `request_limiter_test.go` — in-budget admission, oversize rejection before the
  inner handler reads the body, default-cap normalization, chunked-body
  enforcement, budget exhaustion + release, and the budget-disabled path.
- `config_test.go` — defaults flow through and custom/zero values are honored.


amir-deris:
@seidroid is there anything else left to address on this pr?

sei-will (approved):

masih (approved):

PR #3650, opened 2026-06-26
Add gov proposal based migration trigger

yzang2019:
## Summary
The state-commitment (SC) store's in-flight `memiavl → flatkv` migration was previously paced by a **node-local** config (`sc-keys-to-migrate-per-block`). Because migration writes feed the AppHash, a per-node rate is consensus-relevant and a divergence risk: two validators draining at different rates fork the chain.
This PR makes the per-block migration rate a **module-agnostic governance parameter** (`migration.NumKeysToMigratePerBlock`) that every validator reads from chain state each `BeginBlock` and applies identically. The gov param also serves as the migration *trigger*: it defaults to `0` (paused), and raising it via a param-change proposal starts the drain fleet-wide at a deterministic height.
The old node-local config and its fallback are removed entirely — the gov param is now the **sole** source of the rate.
### New generic governance parameter
- `app/migration/params.go` — new module-agnostic `migration` params subspace defining `NumKeysToMigratePerBlock` (default `0`), its `KeyTable`, and validation. Deliberately *not* EVM-specific so future module migrations can reuse it.
- `app/app.go` — registers the `migration` subspace; stores the `*rootmulti.Store` on the app and fails fast if the (unsupported) legacy commit multistore is in use.
- `app/abci.go` — `BeginBlock` reads `NumKeysToMigratePerBlock` from chain state and pushes it into the SC store before the block's first write (`applyMigrationBatchSize`).
### Plumbing: push the rate down to the migration router
- `sei-db/state_db/sc/types/types.go` — `Committer` interface gains `SetMigrationBatchSize(int) error`.
- `sei-cosmos/storev2/rootmulti/store.go` — forwards `SetMigrationBatchSize` to the SC store; adds `GetMigrationBatchSize` for observability/tests.
- `sei-db/state_db/sc/composite/store.go` — holds the gov-set batch size (`atomic.Int64`); `buildRouter` and `SetMigrationBatchSize` use it directly. **Removed the `effectiveMigrationBatchSize` config fallback** — `0` means paused, full stop.
- `sei-db/state_db/sc/migration/*` — `Router` interface gains `SetMigrationBatchSize`; all router types implement it. Only the migration manager acts on it; every other router (module/passthrough/dual-write/thread-safe) treats it as a no-op. `router_builder.go` now allows a `0` batch size (paused) instead of rejecting it.
### Removed the node-local config + fallback
- `sei-db/config/sc_config.go`, `sei-db/config/toml.go`, `docker/localnode/config/app.toml`, `app/seidb.go` — dropped the `sc-keys-to-migrate-per-block` field, default, validation, TOML entry, and flag parsing.
### Tests
- `app/migration/params_test.go` — unit tests for the new param (default, key-table registration, validation).
- `app/abci_test.go` — subspace registration, `applyMigrationBatchSize` (defaults/set/clamp), and cross-block "param set in block N takes effect in N+1" coverage.
- SC/rootmulti unit tests now seed the rate via `SetMigrationBatchSize(...)` after construction instead of the removed config field. The random-migration framework re-applies the rate on every store (re)open, faithfully mirroring how production re-pushes the gov param after each restart.
### Integration test
- `integration_test/gov_module/` — new `ParameterChangeProposal` test that raises `NumKeysToMigratePerBlock` and asserts it takes effect.
- `integration_test/contracts/verify_flatkv_evm_migrate.sh` — rewritten to drive the migration via a governance param-change proposal (submit → deposit → quorum vote → poll for `PASSED` → verify on-chain) instead of injecting the removed config.
### Docs
- `AGENTS.md` — Code style now mandates running both `gofmt` and `goimports` on every touched `.go` file.


app/app.go
```diff
@@ -549,6 +551,16 @@ func New(
 		option(app)
 	}
 
+	// The storev2 rootmulti store is the only supported commit multistore; its
+	// composite SC backend drives the in-flight memiavl->flatkv migration that
+	// BeginBlock paces via the migration gov param. Fail fast if the legacy
+	// root multistore is somehow in use.
+	rs, ok := app.CommitMultiStore().(*rootmulti.Store)
```
cody-littley: If we need to cast like this, would it make sense to change the type returned by `app.CommitMultiStore()`?

app/app.go
```diff
@@ -469,6 +470,7 @@ type App struct {
 	genesisImportConfig genesistypes.GenesisImportConfig
 
 	stateStore   seidb.StateStore
+	rs           *rootmulti.Store
```
cody-littley: nit, would it make sense to use a more descriptive name like `rootStore`? Out of context, it's not obvious what `rs` stands for.

sei-db/state_db/sc/migration/migration_manager.go
```diff
@@ -98,8 +98,12 @@ func NewMigrationManager(
...
 		return nil, errors.New("iterator must not be nil")
 	}
-	if migrationBatchSize <= 0 {
-		return nil, fmt.Errorf("migration batch size must be positive, got %d", migrationBatchSize)
+	// A batch size of 0 is a valid "paused" state: the migration manager
+	// is wired up and routes caller writes, but advances no keys per block
+	// until SetMigrationBatchSize raises it above 0 (the governance param
+	// acts as the migration trigger). Only a negative size is rejected.
+	if migrationBatchSize < 0 {
+		return nil, fmt.Errorf("migration batch size must not be negative, got %d", migrationBatchSize)
```
cody-littley: Could we make migration batch size an unsigned integer?

cody-littley (approved):

sei-db/state_db/sc/migration/migration_manager.go
yzang2019: Yeah, that's a good point

app/app.go
yzang2019: good suggestion, will change

app/app.go
yzang2019: Make sense

sei-db/state_db/sc/migration/migration_manager.go
yzang2019: Tried that, looks like it would require a big refactory, will actually do a fallback here to 0 if it's negative

sei-cosmos/storev2/rootmulti/store.go
```diff
@@ -649,6 +649,62 @@ func (rs *Store) SetInitialVersion(version int64) error {
 	return rs.scStore.SetInitialVersion(version)
 }
 
+// SetMigrationBatchSize forwards the governance-controlled number of keys
+// to migrate per block to the SC store, and — when migration has been
+// requested (batch size > 0) while the store is still in the pre-migration
+// steady state — kicks it off by transitioning memiavl_only -> migrate_evm.
+// Only a composite store under types.Auto acts on either; every other SC
+// store treats the forward as a no-op and never reports a memiavl_only
```
blindchaser: GetWriteMode() returns cs.currentWriteMode unconditionally (composite/store.go:510), so a node with an explicit sc-write-mode = "memiavl_only" (a fixed config) does report memiavl_only. The kick-off branch is therefore entered, and SetWriteMode rejects it because the config isn't auto

That error propagates up and is only logged, then execution continues.


should we fail it loud and fix the comments?

sei-cosmos/storev2/rootmulti/store.go
```diff
@@ -649,6 +649,62 @@ func (rs *Store) SetInitialVersion(version int64) error {
...
+// The app calls this from BeginBlock before the block's first write, so the
+// empty-buffer precondition of SetWriteMode (which rebuilds the cached
+// views) holds. Forwarding the batch size first means the migration router
+// built by the transition already observes the new rate.
+func (rs *Store) SetMigrationBatchSize(batchSize int) error {
+	if err := rs.scStore.SetMigrationBatchSize(batchSize); err != nil {
+		return fmt.Errorf("failed to set SC store migration batch size: %w", err)
+	}
+	if batchSize > 0 {
+		if mode, ok := rs.GetWriteMode(); ok && mode == sctypes.MemiavlOnly {
```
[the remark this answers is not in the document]
yzang2019: For fixed MemiavlOnly node, we intentionally want to diverge the app hash and skip the migration. So the switch should only work for auto mode

sei-cosmos/storev2/rootmulti/store.go
yzang2019: Yeah this will be fixed properly next, for fixed memiavl_only mode, we will just treat it as a node that doesn't wanna migrate and wanna diverge

integration_test/gov_module/gov_proposal_test.yaml
```diff
@@ -34,6 +34,41 @@
     - type: eval
       expr: NEW_ABCI_PARAM == "true"
 
+- name: Test migration param change proposal should update NumKeysToMigratePerBlock
```
blindchaser, line 45: it looks the migration gov tests triggers a real evm migration but only asserts the param value, not the migration. want to make sure it's on purpose

yzang2019: There's no good way to check if real migration started or not, I decided to add a check to look for logs signals

blindchaser (approved):

sei-db/state_db/sc/migration/migration_manager.go
```diff
@@ -285,8 +288,15 @@ func (m *MigrationManager) ApplyChangeSets(changesets []*proto.NamedChangeSet, f
 	oldDBPairsByStore := make(map[string]map[string]*proto.KVPair)
 	newDBPairsByStore := make(map[string]map[string]*proto.KVPair)
 
+	// advanceMigration gates the once-per-block boundary advance. It is
+	// suppressed when the batch size is 0 (migration paused): caller writes
+	// still route below, but no keys are pulled forward and no boundary
+	// metadata is rewritten, so the migration holds at its current cursor
+	// until the batch size is raised again.
+	advanceMigration := firstBatchInBlock && m.migrationBatchSize > 0
```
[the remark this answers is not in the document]
yzang2019: It only leaves memiavl_only when batchSize > 0. On that very block, migrationBatchSize is already stored (line 705) before the transition, so that block's commit runs advanceMigration == true and writes the boundary → migrationStarted becomes true and is latched for the store's lifetime.
Batch can only change in BeginBlock, so there is no sub-block window where migrate_evm coexists with batch==0 before the boundary is written.
Pausing later (batch → 0) is safe because migrationStarted is already latched true → the lattice stays in the AppHash.
An auto node that hasn't been triggered is memiavl_only (FlatKV closed, not participating), not migrate_evm.

PR #3652, opened 2026-06-26
Parallelize littidx eth_getLogs across blocks

Kbhat1:
- previously filterLogsByTags walked its block range one block at a time. 
- Fan them across a bounded errgroup 

- Verified on node filter runs 
- Unit tests

### DIFF
--- a/sei-db/ledger_db/receipt/litt_tag_index.go
+++ b/sei-db/ledger_db/receipt/litt_tag_index.go
@@ -207,18 +207,13 @@ func (s *littReceiptStore) filterLogsByTags(fromBlock, toBlock uint64, crit filt
 	groups := criteriaTagGroups(crit)
 	nBlocks := int(toBlock - fromBlock + 1) //nolint:gosec // bounded by the caller's range cap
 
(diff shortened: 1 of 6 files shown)


sei-db/ledger_db/receipt/litt_tag_index.go
```diff
@@ -201,24 +205,54 @@ func (s *littReceiptStore) filterLogsByTags(fromBlock, toBlock uint64, crit filt
 	}
 
 	groups := criteriaTagGroups(crit)
+	nBlocks := int(toBlock - fromBlock + 1) //nolint:gosec // bounded by the caller's range cap
+
+	par := s.logFilterParallelism
+	if par < 1 {
```
yzang2019: There's one fallback in the constructor and another one here, shall we only do the fallback in one place?

yzang2019 (approved):

sei-db/ledger_db/receipt/litt_tag_index.go
Kbhat1: Updated

PR #3660, opened 2026-06-29
Per-block littidx flush + single shard (gated on #3645)

Kbhat1:
## Description
- NOTE: Only Merge post [3645](https://github.com/sei-protocol/sei-chain/pull/3645): That makes per-block flush affordable, reclaiming near-per-block crash durability for free.
- `littFlushInterval` 100ms -> **5ms** (~one block at Giga throughput) — bounds crash loss to ~a single block (near-WAL durability for receipt bodies).
- `ShardingFactor` 16 -> **1**. flushing one segment file is cheaper; sharding mainly helps across multiple disks.

## Testing
- reads-off write ceiling **167k** (vs 168k prior)
- reads-on **~153–157k** writes / 58k point-reads/s / 520 getLogs/s @ 7ms
- 3h GC soak **flat** (old −27% control-loop step-down gone), 0 panics

### DIFF
--- a/sei-db/ledger_db/receipt/litt_receipt_store.go
+++ b/sei-db/ledger_db/receipt/litt_receipt_store.go
@@ -44,11 +44,12 @@ import (
 // metadata (m:latest / m:earliest).
 //
 // Durability: a background flusher bounds litt durability lag to
-// littFlushInterval without putting fsync on the commit path; Close flushes the
-// remainder. Accepted tradeoff: a hard crash can lose up to littFlushInterval
-// of the most recent receipt bodies (litt buffers in memory, no WAL) while the
-// index still lists them — tolerable for auxiliary, non-consensus RPC data,
-// since reads return not-found for a missing body. Retention: receipt values
+// littFlushInterval (~one block) without putting fsync on the commit path;
+// Close flushes the remainder. A hard crash can lose up to littFlushInterval of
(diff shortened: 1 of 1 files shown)


sei-db/ledger_db/receipt/litt_receipt_store.go
```diff
@@ -80,7 +81,12 @@ const (
 	receiptBackendLittIdx = "littidx"
 
 	littReceiptTableName = "receipts"
-	littFlushInterval    = 100 * time.Millisecond
+	// littFlushInterval is roughly one flush per block at Giga throughput (a
+	// block is ~7ms), bounding crash loss to about a single block. Flushing this
+	// often is only cheap because litt flushes its keymap asynchronously, off the
+	// control loop; without that, a per-block flush regresses write throughput
+	// badly (≈-48% observed before the async keymap landed).
+	littFlushInterval = 5 * time.Millisecond
```
cody-littley, line 89: Are we requiring one flush per block for crash recovery safety? If so, doing it via wall clock time seems a bit fragile.

cody-littley (approved):
Comments are true only once I merge my branch, but IMO not a big deal since that's merging soon.

My only concern is that timing might not provide an ironclad crash safety guarantee. But this change by itself doesn't make the problem any worse, so no need to block it while we have this discussion.

PR #3682, opened 2026-07-01
prometheus metrics for autobahn/avail, autobahn/data and p2p/mux

pompon0:
It will give us insight into consensus state and rpc performance.


bdchatham (commented):
**xreview summary** (opentelemetry / sei-networking / systems / idiom lenses). Nice, well-scoped instrumentation — most of it is clean. Three points worth resolving before merge, plus a few smaller ones inline.

**🔴 MeterProvider wiring** — where is the global `MeterProvider` set up? `otel.Meter()` is a no-op unless a provider with a Prometheus/OTLP reader and a `service.name` resource is registered in the binary, and I don't see it in this diff. If it isn't already wired in main, every instrument here silently records nothing — worth confirming before merge.

On your weighted-histogram question: you don't need to call `Record` more than once. You have exactly one latency value per commit, so it's a single `Record` into a `Float64Histogram` — the sum+count counter split was solving a problem the code doesn't have. Details in the inline comment on `commitToCommitLatency`.

sei-tendermint/internal/p2p/mux/metrics/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+	return &Stream{opts: metric.MeasurementOption(attrs)}
+}
+
+func (s *Stream) Open() {
+	if s.start.IsPresent() {
+		return
+	}
+	s.start = utils.Some(time.Now())
+	ctx := context.Background()
+	inFlight.Add(ctx, 1, s.opts)
```
bdchatham: 🔴 `inflight` will drift upward and never recover under normal peer churn. `Open()` adds 1, but the matching `-1` only runs on the graceful `close()` path — when `Mux.Run` exits on ctx-cancel or a connection error it drops still-open streams without closing them. The giga streams are long-lived, so every disconnect leaks a count. Either close outstanding streams when `Run` exits, or drive `inflight` off the stream create/delete lifecycle instead of Open/Close.

sei-tendermint/internal/autobahn/avail/metrics/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+var appGlobalBlockNumber = utils.OrPanic1(meter.Int64Gauge(
+	"tendermint_internal_autobahn_avail__app_global_block_number",
+	metric.WithDescription("global block number of the highest observed appQC"),
+))
+var proposalToCommitLatency = utils.OrPanic1(meter.Float64Histogram(
+	"tendermint_internal_autobahn_avail__proposal_to_commit_latency",
+	buckets(0.01, 1.2, 35),
+	metric.WithDescription("latency from proposal being constructed to commit being observed"),
+))
+var commitToCommitLatencySum = utils.OrPanic1(meter.Float64Counter(
```
bdchatham: 🔴 Make this one `Float64Histogram` rather than a sum + count counter pair. As counters you can only recover the mean — you lose the p99 inter-commit tail, which is the liveness signal you most want here. It also breaks on export: the counter `_total` suffix turns these into `..._sum_total` / `..._count_total`, which reads as a malformed histogram. You already have one value per commit, so it's a single `Record` call — no efficiency cost, and it matches `proposalToCommitLatency` right above.

sei-tendermint/internal/autobahn/avail/metrics/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+))
+var commitGlobalBlockNumber = utils.OrPanic1(meter.Int64Gauge(
+	"tendermint_internal_autobahn_avail__commit_global_block_number",
+	metric.WithDescription("global block number of the highest observed commitQC"),
+))
+var appGlobalBlockNumber = utils.OrPanic1(meter.Int64Gauge(
+	"tendermint_internal_autobahn_avail__app_global_block_number",
+	metric.WithDescription("global block number of the highest observed appQC"),
+))
+var proposalToCommitLatency = utils.OrPanic1(meter.Float64Histogram(
```
bdchatham: 🟡 Add `metric.WithUnit("s")` on the avail latency instruments — the mux-side `latency` sets it but these don't, so the exporter won't append `_seconds` and the two packages end up with inconsistently-named latency series.

sei-tendermint/internal/p2p/rpc/rpc.go
```diff
@@ -86,8 +88,16 @@ func Register[API any, Req, Resp protoutils.Sized](kind mux.StreamKind, limit Li
 	if err := resp.Verify(); err != nil {
 		panic(fmt.Errorf("RPC %v: %w", kind, err))
 	}
-	service[kind] = &rpcConfig{limit: limit}
-	return &RPC[API, Req, Resp]{kind, limit, req, resp}
+	service[kind] = &rpcConfig{
+		limit: limit,
+		name:  fmt.Sprintf("%T", utils.Zero[Req]()),
```
bdchatham: 🟡 `fmt.Sprintf("%T", ...)` bakes the fully-qualified Go type into the `rpc_name` label, so renaming or moving the request type silently breaks dashboard continuity. Prefer an explicit, stable name string.

sei-tendermint/internal/autobahn/avail/metrics/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+func newObserved[T any]() utils.Mutex[*utils.Option[observed[T]]] {
+	return utils.NewMutex(utils.Alloc(utils.None[observed[T]]()))
+}
+
+var observedCommitQC = newObserved[*types.CommitQC]()
+var observedAppQC = newObserved[*types.AppQC]()
+
+// ObserveCommitQC observes the CommitQC latency.
+func ObserveCommitQC(c *types.Committee, qc *types.CommitQC) {
+	ctx := context.Background()
```
bdchatham: 🟡 Recording with `context.Background()` means these can never carry trace exemplars. The callers (`PushCommitQC`, the stream methods) already have a `ctx` — thread it through so we keep that option open.

sei-tendermint/internal/p2p/mux/stream.go
```diff
@@ -91,6 +92,7 @@ func (s *Stream) Send(ctx context.Context, msg []byte) error {
 			f.Header.PayloadSize = utils.Alloc(uint64(len(msg)))
 			f.Header.MsgEnd = utils.Alloc(true)
 		}
+		inner.metrics.Send(len(msg))
```
bdchatham, line 95: 🟡 Each `.Add` allocates ~16 B via the variadic `AddConfig`, and all same-kind streams share one aggregator mutex that's taken here while the stream lock is held — ~64 B / 4 allocs / 4 shared-lock acquisitions per Send+Recv round-trip. Fine at today's rates; if a mux ever gets hot, batch these or move the `Add` outside the stream lock.

sei-tendermint/internal/p2p/mux/metrics/metrics.go
pompon0: every successfully opened stream is closed eventually, even if mux exits. I don't understand this finding.

bdchatham:
### Weighted histograms in OTel — summary of the thread

TL;DR on "how do I migrate the weighted histograms without calling `Record` N times":

**There's no weighted `Record` in the OTel Go API** — a histogram counts observations, so weighting a single latency by N blocks means either N `Record` calls or emitting a pre-aggregated point yourself. Two things worth separating: *is looping actually a problem* (no), and *what's the right metric* (probably not a weighted histogram at all).

**1. Looping `Record` is cheap at our scale.** 1–120 blocks at single-digit commits/sec is single-digit µs/commit — invisible next to the sequencing work. The only real footgun is allocating the attribute set inside the loop, so hoist it:

```go
recordOpts := metric.WithAttributeSet(attrSet) // build once, outside the loop
for i := uint64(0); i < blocks; i++ {
    seqLatency.Record(ctx, latencySec, recordOpts)
}
```

**2. But I'd reconsider the weighted histogram itself.** Replaying one shared per-commit measurement into N identical points inflates the statistical support and quietly redefines `_count` (blocks, not commits). The honest signal keeps latency and load orthogonal:

```go
seqLatency.Record(ctx, latencySec, opts)            // per-commit, unweighted → commit-health p99
blocksSequenced.Add(ctx, int64(blocks), opts)       // blocks_sequenced_total → throughput
// optional: blocksPerCommit.Record(ctx, float64(blocks), opts) // batch-size distribution
```

Now `rate(blocks_sequenced_total)` gives throughput, the histogram stays meaningful, and you can correlate "latency rose when blocks/commit rose" on one panel — strictly more information than one weighted blob.

**3. If you genuinely need a pre-aggregated / weighted histogram in OTel**, it's doable natively via a custom `metric.Producer` — the SDK analogue of Prometheus `MustNewConstHistogram`. You register it on the reader (`metric.WithProducer(...)`, passed into `prometheus.New(...)` — verify the exact symbol on our pinned SDK) and hand back a point whose `Count`/`Sum`/`BucketCounts` you set directly, so weighting is "add N to the bucket, N to count":

```go
func (p *seqLatencyProducer) Observe(latencySec float64, blocks uint64) {
    p.mu.Lock(); defer p.mu.Unlock()
    i := sort.SearchFloat64s(p.bounds, latencySec) // first bound >= value
    p.bucketCounts[i] += blocks                    // weight by N, O(1)
    p.count += blocks
    p.sum += latencySec * float64(blocks)
}

func (p *seqLatencyProducer) Produce(ctx context.Context) ([]metricdata.ScopeMetrics, error) {
    p.mu.Lock()
    bc := append([]uint64(nil), p.bucketCounts...) // snapshot under lock
    dp := metricdata.HistogramDataPoint[float64]{
        Attributes: p.attrs, StartTime: p.start, Time: time.Now(),
        Count: p.count, Bounds: p.bounds, BucketCounts: bc, Sum: p.sum,
    }
    p.mu.Unlock()
    return []metricdata.ScopeMetrics{{
        Scope: instrumentation.Scope{Name: "consensus"},
        Metrics: []metricdata.Metrics{{
            Name: "consensus_block_sequencing_latency", Unit: "s",
            Data: metricdata.Histogram[float64]{
                Temporality: metricdata.CumulativeTemporality,
                DataPoints:  []metricdata.HistogramDataPoint[float64]{dp},
            },
        }},
    }}, nil
}
```

The catch: you now own the cumulative-temporality invariants (bucket counts monotonically non-decreasing + a stable `StartTime`, or Prometheus reads a reset and `histogram_quantile` breaks), your own concurrency, your own attribute sets, and you lose Views + exemplars (you've bypassed the SDK aggregator). That's worth it only when the bucket counts **already exist upstream** (a pre-aggregated subprocess/library/batch source) or the per-observation path is genuinely hot — i.e. folding in an aggregation, not synthesizing one. Not our case here, so I'd go with option 1 or 2.

On the API-ergonomics point: yes, client_golang's `MustNewConstHistogram` is more ergonomic for pre-aggregated data — that's the price of OTel being backend/temporality-agnostic (push+pull, delta+cumulative, Views, exemplars). If we ever want the raw ergonomics for a few metrics, we can register a plain client_golang `Collector` on the same Prometheus registry and it scrapes alongside the OTel series — but only under a pull/Prometheus exporter, not OTLP push.

sei-tendermint/internal/autobahn/avail/metrics/metrics.go
pompon0: it is an efficiency cost, because we would have a separate histogram per timeouts value, which gives 700 additional buckets.

sei-tendermint/internal/autobahn/avail/metrics/metrics.go
pompon0: I find referencing ctx for the sake of bumping a metric super annoying. Metric updates are supposed to be superfast - over the last 2 years we have wasted ton of time debugging inefficient metrics.

sei-tendermint/internal/autobahn/avail/metrics/metrics.go
bdchatham: Keying a histogram on timeouts fans out to ~700 series, which the sum/count pair avoids, so counters are a reasonable call if we only need the mean here

bdchatham (approved):

sei-tendermint/internal/autobahn/avail/metrics/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+	now := time.Now()
+	for mLast := range observedCommitQC.Lock() {
+		if last, ok := mLast.Get(); ok {
+			if last.val.Index() >= qc.Index() {
+				return
+			}
+			// Constructed once per CommitQC, which we should afford.
+			attrs := metric.WithAttributeSet(attribute.NewSet(
+				// Timeouts capped: 20 means [20,inf)
```
wen-coding: I suppose we almost never have 20+ timeouts on one RoadIndex? What happens if the cluster gets stuck?

sei-tendermint/internal/autobahn/avail/metrics/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+	time time.Time
+	val  T
+}
+
+func newObserved[T any]() utils.Mutex[*utils.Option[observed[T]]] {
+	return utils.NewMutex(utils.Alloc(utils.None[observed[T]]()))
+}
+
+var observedCommitQC = newObserved[*types.CommitQC]()
+var observedAppQC = newObserved[*types.AppQC]()
```
wen-coding, line 43: Do we need the full QC? We are only using RoadIndex, View Index, and timestamp I suppose?

sei-tendermint/internal/autobahn/avail/inner.go
```diff
@@ -163,9 +164,11 @@ func (i *inner) prune(c *types.Committee, appQC *types.AppQC, commitQC *types.Co
 		return false, nil
 	}
 	i.latestAppQC = utils.Some(appQC)
+	metrics.ObserveAppQC(appQC)
```
wen-coding, line 167: This could be much later than we actually observe AppQC for this RoadIndex right? Because the block proposal packing AppQC is optional.

sei-tendermint/internal/autobahn/avail/inner.go
```diff
@@ -163,9 +164,11 @@ func (i *inner) prune(c *types.Committee, appQC *types.AppQC, commitQC *types.Co
...
 	i.commitQCs.prune(idx)
 	if i.commitQCs.next == idx {
 		i.commitQCs.pushBack(commitQC)
+		metrics.ObserveCommitQC(c, commitQC)
```
wen-coding, line 171: Sorry I'm confused, why do we observe these QCs inside prune() instead of right after the QCs are verified?

sei-tendermint/internal/autobahn/avail/metrics/metrics.go
pompon0: then we don't care about this metric, I suppose.

sei-tendermint/internal/autobahn/avail/metrics/metrics.go
pompon0: this is a simplification, there is no need to restrain it down.

sei-tendermint/internal/autobahn/avail/inner.go
pompon0: true, these metrics are describing the avail.State content.

wen-coding (approved):

sei-tendermint/internal/autobahn/avail/metrics/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+			}
+			// Constructed once per CommitQC, which we should afford.
+			attrs := metric.WithAttributeSet(attribute.NewSet(
+				// Timeouts capped: 20 means [20,inf)
+				attribute.Int64("timeouts", int64(min(qc.Proposal().View().Number, 20))), // nolint: gosec
+			))
+			commitToCommitLatencySum.Add(ctx, now.Sub(last.time).Seconds(), attrs)
+			commitToCommitLatencyCount.Add(ctx, 1, attrs)
+		}
+		proposalToCommitLatency.Record(ctx, now.Sub(qc.Proposal().Timestamp()).Seconds())
```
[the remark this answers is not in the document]
pompon0: WAI

sei-tendermint/internal/autobahn/avail/inner.go
pompon0: metrics track newest commitQC - the commitQC from the anchor is newest iff the anchor wipes the whole data state.

sei-tendermint/internal/p2p/mux/stream.go
pompon0: addressed - with prometheus we don't have that problem

sei-tendermint/internal/p2p/mux/stream.go
```diff
@@ -100,6 +102,7 @@ func (s *Stream) close(inner *streamStateInner) {
 		return
 	}
 	inner.closed.local = true
+	inner.metrics.Close()
```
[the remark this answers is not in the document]
pompon0: local Close is ALWAYS required by the contract on Stream object - if mux.Connect/Accept succeeds, then caller is required to call Close eventually.

sei-tendermint/internal/autobahn/avail/inner.go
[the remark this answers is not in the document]
pompon0: WAI

sei-tendermint/internal/autobahn/avail/metrics/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// Global block number of the highest observed commitQC.
+	commitGlobalBlockNumber prometheus.GaugeIntVec
+	// Global block number of the highest observed appQC.
+	appGlobalBlockNumber prometheus.GaugeIntVec
+
+	// Latency from proposal being constructed to commit being observed.
+	proposalToCommitLatency prometheus.HistogramVec `metrics_buckets:"exp(0.01, 1.2, 35)"`
+	// Latency between consecutive commits being observed.
+	commitToCommitLatency prometheus.HistogramVec `metrics_labels:"timeouts" metrics_buckets:"none"`
```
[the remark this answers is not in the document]
pompon0: WAI

PR #3684, opened 2026-07-01
Add offline rollback utility for LittDB

Kbhat1:
- RollbackLittDB(dataDirs, filter) — offline tool that rewinds a littDB to a chosen point (e.g. a block height) while the DB is stopped
- Walks each table's key files newest -> oldest, calls filter(key, isPrimary); first match = cutoff -> keeps that key's group + everything older, deletes everything newer from the segment files
- Discards the keymap + snapshot dirs so the DB rebuilds them from the truncated segments on next start

- Unit tests
- Currently verifying on node


sei-db/db_engine/litt/rollback/rollback.go
```diff
@@ -0,0 +1,282 @@
...
+// primaries and primaries that own secondary keys) and false for secondary keys. The first record for
+// which the filter returns true is the rollback point: that record's group is kept along with everything
+// written before it, and everything written after the group is discarded.
+type RollbackFilter func(key []byte, isPrimary bool) (bool, error)
+
+// RollbackLittDB performs an offline rollback of the LittDB instance stored across the given data
+// directories (the same paths passed to the database as its storage roots).
+//
+// For every table found under dataDirs, RollbackLittDB walks the key files from newest to oldest and
+// invokes rollbackFilter for each key. The first key for which the filter returns true marks the rollback
```
cody-littley: If this walks multiple tables, then we need to add a table argument to the RollbackFilter. Key schema may be different across different tables.

sei-db/db_engine/litt/rollback/rollback.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+	return 0, fmt.Errorf("key group starting at record %d has no terminating final-secondary record", i)
+}
+
+// discardDerivedState removes the keymap and snapshot directories for a table from every root. Both are
+// derived entirely from the segment files: the database rebuilds the keymap (via reloadKeymap) and the
+// snapshot on its next start, so deleting them forces both back into sync with the truncated segments.
+// Leaving them would let a stale keymap reference discarded keys, or let a snapshot's hard links pin the
+// rolled-back data on disk. Removing a directory that does not exist is a no-op.
+func discardDerivedState(roots []string, tableName string) error {
```
cody-littley, line 256: Optional:

Instead of dropping the keymap directory wholesale, we can prune the keymap to avoid the cost of rebuilding it. To do this, we'd iterate the keys we are removing, and issue batch deletion operations to the keymap. We'd need to make sure we delete the keys before the segments though, in order to ensure crash safety.

This may be a premature optimization, as rolling back LittDB data is likely going to be rare. I'll leave it up to you if you want to do this or not.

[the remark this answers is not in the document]
cody-littley: It's possible you started this branch before my keymap threading optimizations landed. The `gc-watermark` file was something introduced in my branch.

cody-littley (approved):
LGTM once comments are addressed

sei-db/db_engine/litt/rollback/rollback.go
Kbhat1: Done. RollbackFilter now takes tableName as its first arg

sei-db/db_engine/litt/rollback/rollback.go
Kbhat1: Kept the drop and rebuild since rollback is rare. happy to switch to keymap pruning later if it ever matters

PR #3695, opened 2026-07-02
docs(receipt): littidx receipt store README

Kbhat1:
One clean doc for the `littidx` receipt backend, co-located at `sei-db/ledger_db/receipt/README.md`.

Covers:
- **Architecture** — receipt bodies in LittDB (append-only segments, point lookup by tx-hash secondary key) + a Pebble **tag index** for `eth_getLogs`; parallel per-block fan-out.
- **Durability & retention** — off-commit-path background flush (~1 block), no WAL, TTL + block-range prune + KeepRecent floor.
- **Limits & trade-offs** — auxiliary (non-consensus) durability; `getLogs` ranges have no legacy fallback (archive-node/migration caveat); cold-range latency.
- **Configuration** — the `[receipt-store]` knobs (`rs-backend`, `log-filter-parallelism`, etc.) and that KeepRecent derives from `min-retain-blocks`.
- **Rollout** — no operator data migration: config + state sync (litt is a local side store, rebuilt from live blocks), legacy backfill is chain-level / gov-triggered.


sei-db/ledger_db/receipt/README.md
```diff
@@ -0,0 +1,130 @@
...
+`KeepRecent` is **not** a receipt-store key — it is always derived from the
+global `min-retain-blocks` flag at the app layer.
+
+Example:
+
+```toml
+[receipt-store]
+rs-backend = "littidx"
+log-filter-parallelism = 16
+prune-interval-seconds = 600
```
yzang2019: We don't wanna expose these configs to the users yet, my suggestions is to remove these and use whatever the recommended default we set for them

yzang2019 (approved):

PR #3696, opened 2026-07-02
refactor of sei-tendermint metrics

pompon0:
Metrics cleanup:
* explicit global metric registrations instead of implicit registration in constructor
* removed go-kit dependency which brought no value
* custom GaugeInt/CounterInt types for metrics with integer only values (which allows for faster updates (atomic add, no interface calls)
* custom (copy) of Histogram which supports ObserveWithWeight
* moved populating chain_id label to exporter to ensure all prometheus metrics have it
* hardcoded the Namespace, instead of making it configurable (it doesn't make sense to have it configurable)
* safer metric accessors (number of labels is verifed in compilation time)

No breaking changes to existing metrics are expected: names, labels, types stay the same.


sei-tendermint/node/node.go
```diff
@@ -688,36 +730,36 @@ type NodeMetrics struct {
...
-				evidence:  evidence.PrometheusMetrics(cfg.Namespace, "chain_id", chainID),
+				consensus: consensus.Global,
+				eventlog:  eventlog.Global,
+				indexer:   indexer.Global,
+				mempool:   mempool.Global,
+				p2p:       p2p.Global,
+				proxy:     proxy.Global,
+				state:     sm.Global,
+				statesync: statesync.Global,
+				evidence:  evidence.Global,
```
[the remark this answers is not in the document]
pompon0: WAI

sei-tendermint/internal/consensus/common_test.go
```diff
@@ -39,6 +39,14 @@ import (
 	"github.com/sei-protocol/sei-chain/sei-tendermint/types"
 )
 
+func mustGenesisChainID(cfg *config.Config) string {
```
wen-coding: nit: I think I've seen this function more than once, can they share the same util function?

sei-tendermint/internal/eventlog/metrics.go
```diff
@@ -1,14 +1,18 @@
 package eventlog
 
-import "github.com/go-kit/kit/metrics"
+import tmmetrics "github.com/sei-protocol/sei-chain/sei-tendermint/libs/utils/prometheus"
 
-const MetricsSubsystem = "eventlog"
+const (
+	// MetricsNamespace is the namespace shared by all Tendermint Prometheus metrics.
+	MetricsNamespace = "tendermint"
```
wen-coding, line 7: Just curious: Do we plan to put Giga metrics under a different namespace?

sei-tendermint/internal/state/execution_test.go
```diff
@@ -110,21 +114,22 @@ func TestApplyBlockProposerPriorityHash(t *testing.T) {
...
-	}
 
-	// Expect exactly one emission at the interval boundary.
-	require.Len(t, hashGauge.sets, 1, "hash metric should fire exactly once at height %d", interval)
-	require.Len(t, heightGauge.sets, 1, "height metric should fire exactly once at height %d", interval)
+		if height < interval {
+			require.Zero(t, metricValue(t, testMetrics.ProposerPriorityHashAt()))
+			require.Zero(t, metricValue(t, testMetrics.ProposerPriorityHashHeightAt()))
+		}
+	}
```
wen-coding: nit: originally it checks it fires exactly once, should you check that some metrics value is 1 here?

sei-tendermint/libs/utils/prometheus/prometheus.go
```diff
@@ -0,0 +1,105 @@
...
+type CounterInt struct {
+	value      atomic.Int64
+	desc       *prometheus.Desc
+	labelPairs []*dto.LabelPair
+}
+
+func (c *CounterInt) Desc() *prometheus.Desc { return c.desc }
+func (c *CounterInt) Add(val int64) {
+	if val < 0 {
+		panic(errors.New("counter cannot decrease in value"))
```
wen-coding: nit: should we just print error log here instead of panic? Just in case in the future some packet online is able to pass down a negative value here.

sei-tendermint/internal/state/execution.go
```diff
@@ -228,7 +228,7 @@ func (blockExec *BlockExecutor) ApplyBlock(ctx context.Context, state State, blo
 			Header:              block.Header.ToProto(),
 		},
 	)
-	blockExec.metrics.FinalizeBlockLatency.Observe(float64(time.Since(finalizeBlockStartTime).Milliseconds()))
+	blockExec.metrics.FinalizeBlockLatencyAt().Observe(float64(time.Since(finalizeBlockStartTime).Milliseconds()))
```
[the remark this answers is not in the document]
pompon0: that's preexisting issue

sei-tendermint/scripts/metricsgen/metricsgen.go
```diff
@@ -282,23 +352,78 @@ func extractHelpMessage(cg *ast.CommentGroup) string {
...
+	}
+	args := make([]string, 0, len(labels))
+	for _, label := range labels {
+		args = append(args, labelToParamName(label))
+	}
+	return strings.Join(args, ", ")
+}
+
+func labelToParamName(label string) string {
+	name := nonIdentifierChar.ReplaceAllString(label, "_")
```
wen-coding: nit: theoretically this can map different labels to the same name? e.g. "count", and "count_".

Guess we almost never use that though, "count_a" and "count_b" is more likely.

To make the theoretical chance even slimer, I guess we can use a very rarely used char here rather than "_", but you don't have to do it.

sei-tendermint/internal/eventlog/metrics.go
pompon0: well, it would be misleading to keep "tendermint" infix for autobahn, but for that we should first rename the "sei-tendermint" directory to sth more generic

wen-coding (approved):

bdchatham (approved):

sei-tendermint/internal/state/execution_test.go
pompon0: the problem is that the metrics is set once, but the value it is set to is 0. Checking that metric is set specific number of times is not very useful property to test. I'll remove the test altogether, given that no other metrics have test coverage at all.

sei-tendermint/libs/utils/prometheus/prometheus.go
```diff
@@ -0,0 +1,113 @@
...
+// CounterInt is a Metric that represents a single int64 value that only ever
+// goes up.
+type CounterInt struct {
+	value      atomic.Int64
+	desc       *prometheus.Desc
+	labelPairs []*dto.LabelPair
+}
+
+func (c *CounterInt) Desc() *prometheus.Desc { return c.desc }
+func (c *CounterInt) Add(val int64) {
```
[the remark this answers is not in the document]
pompon0: wai

PR #3701, opened 2026-07-06
State WAL replacement

cody-littley:

This WAL is intended to replace all existing WALs in both SS and SC. Instead of having multiple WALs for multiple different services, we can maintain just a single WAL which is used universally across all parts of the storage stack.

This PR does not integrate the changes, it just creates a WAL utility. Integration work will happen in future PRs. We've got a pre-existing WAL, but that implementation has some serious issues. Namely, the WAL library we currently use has very inefficient garbage collection (it rewrites garbage collected files, as opposed to dropping files as a whole when data in file is all stale).

unit tests


sei-db/state_db/statewal/state_wal_impl.go
```diff
@@ -0,0 +1,797 @@
...
+
+// Write schedules a changeset record for the given block number.
+func (w *stateWALImpl) Write(blockNumber uint64, cs []*proto.NamedChangeSet) error {
+	if w.closed.Load() {
+		return fmt.Errorf("state WAL is closed")
+	}
+	if err := w.enforceWriteOrdering(blockNumber); err != nil {
+		return fmt.Errorf("write rejected: %w", err)
+	}
+	if err := w.sendToSerializer(dataToBeSerialized{entry: NewEntry(blockNumber, cs)}); err != nil {
```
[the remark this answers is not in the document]
cody-littley: updated godoc to reflect that modification is unsafe after method call

sei-db/state_db/statewal/state_wal_impl.go
```diff
@@ -0,0 +1,797 @@
...
+			return nil, fmt.Errorf("failed to create iterator: %w", resp.err)
+		}
+		return resp.iterator, nil
+	case <-w.ctx.Done():
+		if err := w.asyncError(); err != nil {
+			return nil, fmt.Errorf("iterator creation aborted: %w", err)
+		}
+		return nil, fmt.Errorf("iterator creation aborted: %w", w.ctx.Err())
+	}
+}
```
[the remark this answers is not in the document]
cody-littley: updated godoc to reflect current behavior

sei-db/state_db/statewal/state_wal_entry.go
```diff
@@ -0,0 +1,151 @@
...
+	case kindEndOfBlock:
+		return NewEndOfBlockEntry(blockNumber), true, nil
+	case kindChangeset:
+		count, n := binary.Uvarint(rest)
+		if n <= 0 {
+			return nil, false, nil
+		}
+		rest = rest[n:]
+
+		changeset := make([]*proto.NamedChangeSet, 0, count)
```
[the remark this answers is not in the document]
cody-littley: added sanity check

sei-db/state_db/statewal/state_wal_impl.go
blindchaser: Write keeps caller-owned changesets and serializes them later. Because Write returns after enqueueing, callers can mutate or reuse cs before serializerLoop runs, causing the WAL to persist changed data or race with the caller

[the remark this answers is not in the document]
cody-littley: Changed StateWAL to be a non-threadsafe utility and documented this fact. The way we need to use this, the no-concurrent-access rule is not going to be a problem.

sei-db/state_db/statewal/state_wal_file.go
```diff
@@ -0,0 +1,585 @@
...
+		payloadStart := offset + lenN
+		remaining := uint64(len(data) - payloadStart) //nolint:gosec // payloadStart <= len(data), so non-negative
+		if remaining < 4 || length > remaining-4 {
+			break // torn record: payload or checksum truncated (4 trailing bytes are the CRC32)
+		}
+		payloadLen := int(length) //nolint:gosec // bounded above by remaining-4, which is <= len(data)
+		payload := data[payloadStart : payloadStart+payloadLen]
+		recordEnd := payloadStart + payloadLen + 4
+		gotCRC := binary.BigEndian.Uint32(data[payloadStart+payloadLen : recordEnd])
+		if gotCRC != crc32.ChecksumIEEE(payload) {
```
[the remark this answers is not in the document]
cody-littley: Sealed files with corruption now cause the constructor to return an error, and are not automatically recoverable.

sei-db/state_db/statewal/state_wal_impl.go
```diff
@@ -0,0 +1,797 @@
...
+	walFilesSealed.Add(w.ctx, 1)
+
+	mutable, err := newWalFile(w.config.Path, w.nextIndex)
+	if err != nil {
+		return fmt.Errorf("failed to open new mutable WAL file during rotation: %w", err)
+	}
+	w.mutableFile = mutable
+	w.nextIndex++
+	return nil
+}
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/state_db/statewal/state_wal_impl.go
cody-littley: This pattern is intentional for performance, but not well documented.

I updated all godocs to say that it is unsafe to modify slices passed into methods or slices received from methods.

sei-db/seiwal/metrics.go
```diff
@@ -0,0 +1,95 @@
...
+		metric.WithUnit("{count}"),
+	))
+
+	// The time spent serializing a payload in the generic serializing WAL.
+	walSerializeDuration = must(walMeter.Float64Histogram(
+		"seiwal_serialize_duration_seconds",
+		metric.WithDescription("Time spent serializing a payload in the generic WAL"),
+		metric.WithUnit("s"),
+		metric.WithExplicitBucketBoundaries(commonmetrics.LatencyBuckets...),
+	))
```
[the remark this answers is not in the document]
cody-littley: false positive

sei-db/seiwal/seiwal_impl.go
```diff
@@ -0,0 +1,754 @@
...
+// stored index also establishes the invariant pruneSealedFiles relies on: a reservation never falls below the
+// lowest stored index, so a file entirely below the lowest reservation is one every iterator has moved past.
+func (w *walImpl) pinLowestReadableIndex(startIndex uint64) uint64 {
+	pinned := startIndex
+	if r := w.bounds(); r.ok && r.first > pinned {
+		pinned = r.first
+	}
+	w.indexRefs[pinned]++
+	return pinned
+}
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/state_db/statewal/state_wal.go
```diff
@@ -0,0 +1,77 @@
...
+	// Signal that there will be no more writes for the current block number. Attempting to write additional
+	// changes for the same block number after calling this method may result in an error.
+	//
+	// Similar to Write(), this method is asynchronous. Calling this method does not, by itself, make data immediately
+	// crash durable.
+	SignalEndOfBlock() error
+
+	// Flush the WAL to disk. All data previously passed to Write() before this call will be crash durable
+	// after this call returns.
+	Flush() error
```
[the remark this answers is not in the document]
cody-littley: fixed godoc

sei-db/seiwal/seiwal.go
```diff
@@ -0,0 +1,92 @@
...
+
+// WAL is a generic, index-keyed, append-only write-ahead log over payloads of type T.
+//
+// Each record is tagged with a caller-provided monotonic index. The index is what makes garbage
+// collection ("drop everything below N"), iteration ("start at N"), and rollback ("drop everything
+// above N") expressible without the WAL ever interpreting a payload.
+//
+// A WAL instance is not safe for concurrent use: its methods must not be called from multiple
+// goroutines simultaneously. Callers that share a WAL across goroutines must serialize access
+// themselves.
```
yzang2019, line 21: Shall we consider making it goroutine concurrency safe then? It would things much easier from user perspective. My concern is if we implement a non thread safe library, it might makes future integration error prune especially when someone who is not familiar with the library.

I could also imagine sometimes we have to pass WAL around to different objects and call functions in different goroutines, it might be hard to always serialize the function calls of the WAL for both reads and writes.

But if the current usage doesn't require any concurrency call, then I think I'm good for the first pass and we can add that later when we really have concurrency usage

sei-db/seiwal/seiwal.go
```diff
@@ -0,0 +1,92 @@
...
+	// The index must be strictly greater than the index of the most recently appended record (indices
+	// need not be contiguous, but they must strictly increase).
+	//
+	// This method only schedules the append; it does not block until the record is durable. Durability is
+	// achieved by a subsequent Flush.
+	//
+	// data, and every slice reachable through it, must not be modified after this call: the payload may be
+	// retained and serialized asynchronously, so mutating it races the WAL and can corrupt what is
+	// persisted. Callers that need to reuse or mutate the buffer must copy it first.
+	Append(index uint64, data T) error
```
yzang2019, line 43: What's the point of allowing Index to not be contiguous here? 
Is there any scenario that we want to do:
append(1, xxx), append(3, xxx), append(4, xxx), append(10, xxx)

On the other hand, do we really need the caller of this API to pass in the index? If the write pattern is always append only, I wonder if we can just internally manage this index and auto increment it? The benefit is that the caller doesn't need to think about how to get and increment the index any more during writes. So the suggested API would look like this:

`Append(data T) error`

sei-db/seiwal/seiwal.go
```diff
@@ -0,0 +1,92 @@
...
+		err error,
+	)
+
+	// Prune removes all records with an index less than lowestIndexToKeep.
+	//
+	// This method merely schedules the prune; it does not block until the prune is complete. Pruning is
+	// async and lazy, and implementations are free to delay it arbitrarily long. Pruning removes whole
+	// sealed files only, so records may survive above the requested threshold until their containing file
+	// is fully below it.
+	Prune(lowestIndexToKeep uint64) error
```
yzang2019: This Prune (recommend to rename to PruneBefore) is equivalent to truncate from first to lowestIndexToKeep. 
I think we need another API PruneAfter() that is equivalent to truncate from lastIndexToKeep to last.

yzang2019: One use case I think think of PruneAfter useful is for mannual rollback

sei-db/seiwal/seiwal.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// Iterator returns an iterator over the WAL starting at the given index.
+	//
+	// The iterator reads a consistent, point-in-time snapshot of the WAL taken at some instant between the
+	// start and the return of this call. Records appended before that instant are included; records
+	// appended after it are not. For records appended concurrently with this call, whether they are
+	// included is unspecified.
+	Iterator(startIndex uint64) (Iterator[T], error)
```
yzang2019: Personally I would recommend to replace this API or add another API for Replay(startIndex, endIdex) to match our existing usage pattern.

sei-db/seiwal/seiwal_impl.go
```diff
@@ -0,0 +1,780 @@
...
+
+// validateSealedFiles reads and checksums every sealed file, confirming each file's content exactly covers the
+// [first, last] index range its name promises. This surfaces corruption (bit-rot, truncation) at open — where
+// it demands human intervention — rather than lazily at iteration time, by which point Bounds/GetStoredRange
+// would already be reporting a range that iteration cannot deliver. Cost is O(total sealed bytes): every sealed
+// file is read and CRC-verified on open.
+func validateSealedFiles(directory string, sealedFiles *util.RandomAccessDeque[*sealedFileInfo]) error {
+	for _, info := range sealedFiles.Iterator() {
+		contents, err := readWalFile(filepath.Join(directory, info.name))
```
yzang2019: Startup validates every sealed file by reading it entirely into memory which could be very slow and could evict page cache which could impact performance.

readWalFile does this:
`data, err := os.ReadFile(path)`

It is not uncommon that the WAL files could be accumulated into GB or even TB large, at that scale, starting up could take minutes/hours to finish the scanning

sei-db/seiwal/seiwal_impl.go
```diff
@@ -0,0 +1,780 @@
...
+
+// NewWAL opens (or creates) a byte-oriented WAL in the configured directory, recovering any files left
+// behind by a previous session. Operates on []byte payloads.
+func NewWAL(config *Config) (WAL[[]byte], error) {
+	return newWAL(config, nil)
+}
+
+// NewWALWithRollback opens a byte-oriented WAL and deletes all records with an index greater than
+// rollbackIndex before returning, so the WAL contains no record with an index greater than rollbackIndex.
+func NewWALWithRollback(config *Config, rollbackIndex uint64) (WAL[[]byte], error) {
```
yzang2019: Not a big fan of adding rolbackIndex to the constructor, I mean it's fine to keep, but we should also support ondemand rollbacks, not just auto rollback during initialization.

sei-db/seiwal/seiwal_serializing.go
```diff
@@ -0,0 +1,420 @@
...
+	// Fail fast on a bricked WAL, so a fire-and-forget append cannot win the submit select against the
+	// cancelled senderCtx and be silently dropped onto a dead serializer (see walImpl.Append for detail).
+	if err := s.asyncError(); err != nil {
+		return fmt.Errorf("WAL failed: %w", err)
+	}
+	req := serAppend{
+		index:     index,
+		serialize: func() ([]byte, error) { return s.serialize(data) },
+	}
+	if err := s.submit(req); err != nil {
```
yzang2019, line 168: The inner walImpl.Append enforces strict index ordering via appendMu, but serializingWAL.Append does not. Since serialization is deferred to a background goroutine, two rapid calls with indices (5, 3) would both be accepted and enqueued. The inner WAL catches the out-of-order index later, but that fires asynchronously — calling w.fail() which bricks the WAL. The caller of serializingWAL.Append(3, ...) already received a nil error and believes the append succeeded.

There's no appendMu equivalent to catch the misordering synchronously. Either mirror the appendMu gate here, or clearly document that the serializing wrapper provides strictly weaker ordering guarantees than the byte WAL.

sei-db/seiwal/seiwal_serializing.go
```diff
@@ -0,0 +1,420 @@
...
+		return fmt.Errorf("WAL is closed")
+	}
+	// Fail fast on a bricked WAL, so a fire-and-forget append cannot win the submit select against the
+	// cancelled senderCtx and be silently dropped onto a dead serializer (see walImpl.Append for detail).
+	if err := s.asyncError(); err != nil {
+		return fmt.Errorf("WAL failed: %w", err)
+	}
+	req := serAppend{
+		index:     index,
+		serialize: func() ([]byte, error) { return s.serialize(data) },
```
yzang2019, line 166: The closure func() ([]byte, error) { return s.serialize(data) } captures data by value. For T = SomeStruct with slice fields, or T = []byte, the caller could mutate the underlying backing array between Append returning and the serializer goroutine running. The byte-oriented walImpl.Append is safer because it calls frameRecord(index, data) synchronously, copying the payload before enqueueing.

The doc already warns about this, but the asymmetry between the two WAL flavors (byte WAL copies eagerly, serializing WAL defers) is a sharp edge. Consider serializing eagerly on the caller's goroutine and enqueueing the already-serialized bytes, then it would match walImpl's behavior and also let you validate index ordering synchronously.

sei-db/seiwal/seiwal_iterator.go
```diff
@@ -0,0 +1,269 @@
...
+	return it.resultIndex, it.resultPayload
+}
+
+func (it *walIterator) Close() error {
+	it.closeOnce.Do(func() {
+		close(it.stop)    // tell the reader to stop if it is mid-read
+		<-it.readerExited // wait for it to actually exit before releasing resources
+		it.wal.unpinIndex(it.pinnedIndex)
+	})
+	it.done = true
```
yzang2019, line 174: Recommend to use atomic.Bool defensively — it costs nothing and prevents a real data race.
done is read by Next without any synchronization, so there could be data race if Close and Next overlap

sei-db/seiwal/seiwal.go
yzang2019: The iterator exists to support streaming reads concurrent with ongoing writes and pruning. That's why it needs all the machinery:
- Sealing the mutable file — snapshot the write frontier so the iterator sees a consistent point-in-time view
- Pinning (indexRefs) — prevent Prune() from deleting files the iterator hasn't read yet
- Prefetch goroutine — read ahead so the consumer isn't blocked on I/O
- Close lifecycle — unpin so pruning can resume

The core question is: does the WAL actually need streaming reads concurrent with writes and pruning? How a WAL is typically used in a blockchain node?

In practice, WAL replay happens in two scenarios:
1. Startup recovery — replay all records to rebuild uncommitted state. This happens before new writes begin. No concurrency with appends or pruning.
2. Periodic checkpointing — read records since last checkpoint, apply them, then prune. This could be concurrent with new appends, but pruning only removes records before the read range, not within it.

In neither case do you need a long-lived iterator that coexists with concurrent writes and pruning to the same range. You just need: "give me all records from index X to the end."

What a replay function would look like? The simplest version:
`func (w *walImpl) Replay(startIndex uint64, fn func(index uint64, data []byte) error) error
`
Implementation-wise, you'd send a replay request through the writer channel (so it's ordered relative to appends). The writer goroutine would:
1. Flush (not seal) the mutable file
2. Collect the list of sealed files + the mutable file path and its current size
3. Return that snapshot to the caller

Then the caller reads through those files sequentially, calling fn for each record. No background goroutine, no pinning, no lifecycle management.
The pruning concern vanishes for two reasons. On the pruning-during-replay race — you have two clean options:

Option A (simplest): Open all relevant file handles upfront before returning from the writer goroutine. On Linux/macOS, an open file handle survives unlink, so even if Prune() deletes the file from the directory, the handle keeps the data accessible. No pinning needed at all.

Option B: Just document that the caller must not prune indices it's replaying. Since the caller controls both replay and prune, this is trivially enforceable — you prune after replay completes, not during.

What you'd delete? With a replay function, you can remove:
- walIterator struct and the entire seiwal_iterator.go file
- The prefetch goroutine, stop channel, readerExited channel
- indexRefs map and the entire pinning mechanism (pinLowestReadableIndex, unpinIndex, lowestReservation)
- sealForIterator — no more forced rotation on reads
- The iteratorRequest message type in the writer channel
- The it.done data race issue
- The IteratorPrefetchSize config field

That's a substantial reduction in surface area. The core WAL (append, flush, seal, prune, recover) stays the same.

What you'd lose
The only thing you give up is streaming reads with bounded memory during concurrent operation. Specifically:

If you use Option A (open handles upfront), you hold file descriptors for the duration of replay. With many sealed files this could be a lot of FDs, but in practice the count is small.

If the replay callback is slow, new appends continue normally (the mutable file keeps growing), but the replayed snapshot doesn't include them. This is fine — replay is point-in-time by definition.
You lose the ability to do lazy/partial reads (stop iteration early). But with a callback, fn can return a sentinel error to stop early — same effect, simpler mechanism.

sei-db/seiwal/seiwal_impl.go
yzang2019: The sealed files already have per-record CRCs baked into the format. So any corruption — bit-rot, truncation, garbled bytes — will be caught when that record is actually read during replay. The startup validation is doing the exact same CRC checks again, ahead of time, for records you may never even read (if they get pruned before the next replay).

And reading GB of WAL files at startup has a second nasty effect: it floods the OS page cache with cold WAL data, evicting hot pages from the actual database (memiavl, flatKV, pebble, etc.). The first minutes of operation after startup become slow as the real working set faults back in from disk.

Recommendation: three tiers of startup validation
Tier 1 — Metadata-only (default, microseconds)
At startup, only stat() each sealed file. Verify structural invariants without reading any file contents. This catches: missing files, zero-length/truncated files, naming inconsistencies, index gaps. Cost is one stat syscall per file — microseconds even with thousands of files.

Tier 2 — Bookend validation (optional, milliseconds)
Read only the header + first record + last record of each file. This catches filename-content mismatches without reading the interior. 
Cost: two small reads per file (first few bytes + last ~1MB scanned for the final record). For 1000 files this is still sub-second.

Tier 3 — Full validation (optional, background)
Keep the current full-read CRC validation but run it after startup, in a background goroutine, and use fadvise to avoid polluting the page cache.

sei-db/seiwal/seiwal_config.go
```diff
@@ -0,0 +1,81 @@
...
+
+// DefaultConfig returns a default WAL configuration for the WAL at path, identified by name.
+func DefaultConfig(path string, name string) *Config {
+	return &Config{
+		Path:                  path,
+		Name:                  name,
+		WriteBufferSize:       16,
+		SerializerBufferSize:  16,
+		TargetFileSize:        64 * unit.MB,
+		FsyncOnFlush:          true,
```
yzang2019: Would recommend default to false though for performance concern

sei-db/seiwal/seiwal_impl.go
```diff
@@ -0,0 +1,780 @@
...
+	return newWAL(config, &rollbackIndex)
```
blindchaser: an interrupted rollback is not resumed on a plain reopen, and the contract doesn't say so

sei-db/state_db/statewal/state_wal_impl.go
```diff
@@ -0,0 +1,163 @@
...
+		return fmt.Errorf("no block in progress to end")
+	}
+	blockNumber := w.currentBlock
+	w.currentBlockEnded = true
+	changeset := w.buf
+	w.buf = nil // hand ownership to the WAL; the next block starts a fresh buffer
+
+	if err := w.wal.Append(blockNumber, changeset); err != nil {
+		return fmt.Errorf("failed to append block %d: %w", blockNumber, err)
+	}
```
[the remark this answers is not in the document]
cody-littley: the StateWAL now bricks itself on the first error, making this scenario no longer a problem.

sei-db/seiwal/seiwal_impl.go
```diff
@@ -0,0 +1,780 @@
...
+		return fmt.Errorf("append rejected: index %d is not greater than last appended index %d", index, last)
+	}
+	w.lastAppendIndex = index
+	w.hasAppended = true
+	w.appendMu.Unlock()
+
+	record := frameRecord(index, data)
+	if err := w.sendToWriter(dataToBeWritten{record: record, index: index}); err != nil {
+		return fmt.Errorf("failed to schedule append for index %d: %w", index, err)
+	}
```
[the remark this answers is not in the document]
cody-littley: stale comment

sei-db/seiwal/seiwal_serializing.go
blindchaser: +1. two implementation notes if we add the gate here: (1) the gate needs to be seeded from inner.Bounds() at construction (same as newWAL does for lastAppendIndex), otherwise the first append after a reopen goes unchecked; (2) TestGenericWALAppendOrdering currently asserts the async-brick behavior as expected, so its assertions need to flip to a synchronous rejection.

sei-db/seiwal/seiwal.go
cody-littley: This was discussed offline, summarizing here for the record. The end decision was to leave the code as-is (Yiming, please correct me if I misread.)

There are several reasons why I resist adding the mutexes to these WAL classes that would make them "safe" for concurrent callers.

- Even if we added mutexes, the code is still not threadsafe for multiple writers. Because it's critical to the domain that data be written in a very exact order, allowing multiple writers would be inherently unsafe even if the low level library protected variables against concurrent access.
- The StateWAL that will hold diffs for FlatKV will actually be composed of three layers: the StateWAL wrapper that adds logic specific to changesets, the WAL wrapper that does serialization on a background thread, and the WAL impl that writes data to disk on a background thread. If we added mutexes so that each layer of this stack was threadsafe, that means we'd be drilling through three layers of mutexes on each operation.
- The use case we intend to use these WALs for does not require concurrent access.

sei-db/seiwal/seiwal.go
cody-littley: Renamed to `PruneBefore()`.

I intentionally did not put `PruneAfter()` into this interface. The SeiDB implementation does support this workflow. But in order to do so, you call a special constructor, not a method on the instantiated object. The key observation is that pruning the front of the WAL is only ever done at system startup during the recover-and-rebuild phase of system setup. Handling the `PruneBefore()` at construction time allows for significantly simpler implementation.

We currently have the following constructors for the WAL:

```
func NewWAL(config *Config) (WAL[[]byte], error)
func NewWALWithRollback(config *Config, rollbackIndex uint64) (WAL[[]byte], error)
```

sei-db/seiwal/seiwal.go
cody-littley: > What's the point of allowing Index to not be contiguous here?
Is there any scenario that we want to do:
append(1, xxx), append(3, xxx), append(4, xxx), append(10, xxx)

There is no reason not to allow non-contiguous appends for the general use case, so long as the index always increases monotonically. I.e. nothing mechanically breaks in the data structure if we skip an index. For the StateWAL implementation, I think this restriction makes sense, since when recording block changesets it doesn't make sense to skip a block. Even with an empty block, we should be recording an empty changeset.

I added a new flag to the seiwal config: `permitGaps`. If true, it allows non-sequential entries as long as those entries monotonically increase. If false (the default), it will return errors if non-default entries are provided.

> On the other hand, do we really need the caller of this API to pass in the index? If the write pattern is always append only, I wonder if we can just internally manage this index and auto increment it? The benefit is that the caller doesn't need to think about how to get and increment the index any more during writes. So the suggested API would look like this: Append(data T) error

The main reason why I prefer to require an explicit block number with each write is because it forces a tight coupling between the outer context's block number and the WALs block number. We can assert every write that the two stay in sync.  If the WAL's block number and the outer context's block number drift out of sync for whatever reason, then continuing to write more data will result in a corrupted on-disk state.

I'm open to discussing this further if you disagree.

sei-db/seiwal/seiwal_impl.go
yzang2019: Discussed we can move this to a static function instead

sei-db/seiwal/seiwal.go
cody-littley: As we discussed offline, `PruneAfter()` is now a method that is called offline (i.e. when the WAL is not started). It is not a special constructor (like before), it simply edits the data files.

sei-db/seiwal/seiwal.go
cody-littley: Note from when we discussed: this index, at least for the StateWAL use case, will be the block height. It's not some special derived value, and so we don't need any complex mapping between block height and WAL index.

sei-db/seiwal/seiwal.go
cody-littley: We discussed this offline.

The end result of that discussion:

- Function to build iterator remains in the API, since we need a thread safe way to create iterators at runtime while the WAL is being used for other things.
- Iterators DO NOT seal the latest mutable file. Instead, iterators cause some data to be flushed to disk and for some hard links to be created. This new way is much cheaper.

sei-db/seiwal/seiwal_impl.go
cody-littley: Yikes! I did not notice this. 100% agree this is not appropriate for this workload. I'm thinking in terms of TB of data. This would have been very obvious if we'd left it in, wouldn't be surprised to see 10+ minute WAL construction time for real workloads. 😓

I've removed the checksum iteration from the constructor and added in a static helper for doing offline data validation.

sei-db/seiwal/seiwal_impl.go
cody-littley: Yup, is now a static function called `PruneAfter(path, highestIndexToKeep)`

sei-db/seiwal/seiwal_serializing.go
cody-littley: This comment is partially stale, as there are no longer any mutexes in `seiwal_serializing.go`. Defending lack of mutexes below, repeating some points in case some readers are missing context from prior discussions.

- The non-threadsafe part of this code is just for sanity checking. If the code is misused, we want to return an error at the point where it was misused, if possible.
- Even if the sanity check logic fails because somebody called this code concurrently, the background worker goroutine will still catch the error. This is a strong guarantee, it's impossible for incorrect use to bypass this check. The main downside with catching this on the worker thread is that we won't return an error at the site that actually provided the bad input, the error will be returned asynchronously at a different call site, which makes it harder debug.
- Adding mutexes to this code is insufficient to make it thread safe. The core problem is that out of order input is invalid, since it's critical that data be written in-order. Even if the variables are protected by mutexes, it's still possible to send bad input to the WAL if we allow concurrent writes.
- Adding mutexes is a bit expensive, since StateWAL is a composition of 3 layers. If we added mutexes, that would be an extra 3 mutex acquisitions per write.

sei-db/seiwal/seiwal_config.go
cody-littley: Set to false by default. Initially we will want to hard code this to `true` for StateWAL, at least until BlockDB takes over as the source of truth for crash recovery.

sei-db/seiwal/seiwal_iterator.go
cody-littley: The iterator as a whole is documented to be not thread safe. There are a bunch of race conditions if the user calls Close() concurrently with other methods, and making this an `atomic.Bool` would not resolve those race conditions. I'm hesitant to implement partial fixes, since then I'd have to document why I'm using threading primitives in code that is not thread safe.

Thoughts?

yzang2019: Sure, that make sense to me

sei-db/seiwal/seiwal_config.go
yzang2019: Sounds good, Although the tendermint blockstore today already guarantees block is persisted on disk before we commit the state, so I think there's no need to enable fsync tbh

sei-db/seiwal/seiwal_serializing.go
cody-littley: Discussed offline, repeating here for visibility.

The `serializingWAL` struct's whole purpose for existing is to put serialization on its own dedicated goroutine. For high volume data, serialization can sometimes become a performance hotspot. I've seen this specifically with google protobufs, which is something we're very likely going to be writing to this WAL utility. Doing serialization on the caller's thread or on the writer's thread would invalidate the need for `serializingWAL`, so if we did that we'd want to also delete `serializingWAL`.

The conclusion of our discussion was to leave serialization the way it is currently.

sei-db/seiwal/seiwal_config.go
cody-littley: Makes sense.

Outside the scope of this immediate PR, but the reason I advocate for fsync on the StateWAL, at least temporarily, is because the machinery for replaying from block storage still has a few known (hard to provoke) edge cases we are ironing out.

sei-db/seiwal/seiwal_impl.go
cody-littley: This function no longer exists.

sei-db/state_db/statewal/state_wal.go
```diff
@@ -0,0 +1,86 @@
...
+	// is not. For data written concurrently with this call, whether it is included is unspecified.
+	//
+	// The iterator yields one entry per block in ascending block order. Its Entry() returns (blockNumber,
+	// changesets), where changesets are all the changes written for that block (across one or more Write
+	// calls) combined in write order. Blocks that were never ended with SignalEndOfBlock are not yielded.
+	// The returned changesets, and every byte slice reachable through them, must be treated as read-only.
+	Iterator(startingBlockNumber uint64) (seiwal.Iterator[[]*proto.NamedChangeSet], error)
+
+	// Close the WAL, flushing any pending writes and releasing resources.
+	Close() error
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/seiwal/seiwal_iterator.go
```diff
@@ -0,0 +1,289 @@
...
+	// Otherwise wait for the next result, but don't block forever if the WAL is torn down. The reader
+	// goroutine watches the same context (see send) and stops producing when it fires, so the results
+	// channel would never advance again; surface the shutdown as an error rather than hang.
+	select {
+	case result, ok := <-it.results:
+		return it.deliver(result, ok)
+	case <-it.wal.ctx.Done():
+		it.done = true
+		return false, fmt.Errorf("WAL shut down during iteration: %w", it.wal.ctx.Err())
+	}
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/seiwal/seiwal_file.go
```diff
@@ -0,0 +1,584 @@
...
+// whole (records are framed and written atomically), so there is nothing to truncate. A file with no records
+// (including one that never received a record) is removed rather than sealed. Idempotent. Returns the sealed
+// file name, or "" if the file was removed.
+func (f *walFile) seal() (string, error) {
+	if f.sealed {
+		return "", nil
+	}
+	if err := f.flush(true); err != nil {
+		return "", fmt.Errorf("failed to flush before sealing: %w", err)
+	}
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/seiwal/seiwal_file.go
```diff
@@ -0,0 +1,584 @@
...
+		}
+	}
+
+	sealedName := sealedFileName(f.fileSeq, f.firstIndex, f.lastIndex)
+	sealedPath := filepath.Join(f.directory, sealedName)
+	if err := util.AtomicRename(unsealedPath, sealedPath, true); err != nil {
+		return "", fmt.Errorf("failed to seal WAL file: %w", err)
+	}
+	f.sealed = true
+	return sealedName, nil
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/seiwal/seiwal_serializing.go
```diff
@@ -0,0 +1,398 @@
...
+			ok, first, last, err := s.inner.Bounds()
+			m.reply <- serBoundsResult{ok: ok, first: first, last: last, err: err}
+		case serPrune:
+			if err := s.inner.PruneBefore(m.through); err != nil {
+				s.fail(fmt.Errorf("failed to prune below index %d: %w", m.through, err))
+				return
+			}
+		case serIterator:
+			it, err := s.inner.Iterator(m.startIndex)
+			m.reply <- serIteratorResult{it: it, err: err}
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/seiwal/seiwal_impl.go
```diff
@@ -0,0 +1,741 @@
...
+	// The serial number to assign the next iterator, naming its hard-link snapshot directory
+	// (iterator/<serial>/). Mutated only by the writer goroutine.
+	nextIteratorSeq uint64
+}
+
+// recoverDirectory brings a WAL directory into a clean, consistent on-disk state: it removes crash remnants
+// from an interrupted rollback and seals any unsealed file left behind by a prior session. After it returns,
+// every record lives in a sealed file whose name matches its content, with no orphans remaining. Shared by
+// the WAL constructor and the offline GetRange/PruneAfter utilities, so all three run the same sanity pass.
+func recoverDirectory(path string) error {
```
blindchaser, line 161: a second wal instance or an offline utility opens the same path while one is live may corrupts the log

can we add a FileLock to the dir?

sei-db/seiwal/seiwal_serializing.go
```diff
@@ -0,0 +1,406 @@
...
+	select {
+	case r := <-reply:
+		if r.err != nil {
+			return nil, fmt.Errorf("failed to create iterator: %w", r.err)
+		}
+		return &serializingIterator[T]{inner: r.it, deserialize: s.deserialize}, nil
+	case <-s.ctx.Done():
+		if err := s.asyncError(); err != nil {
+			return nil, fmt.Errorf("iterator creation aborted: %w", err)
+		}
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/seiwal/seiwal_impl.go
```diff
@@ -0,0 +1,741 @@
...
+	if !r.ok {
+		// Nothing stored: an empty iterator with no snapshot directory.
+		return iteratorStartResponse{iterator: newWalIterator(w, startIndex, 0, "", nil, w.config.IteratorPrefetchSize)}
+	}
+	maxIndex := r.last
+
+	// Gather the files to read: sealed files reaching startIndex, then the mutable file if it holds records at
+	// or above startIndex. Files entirely below startIndex are never opened, so they are not linked. The
+	// mutable snapshot's range is capped at maxIndex; the reader drops anything the writer appends past it.
+	var sources []iteratorFile
```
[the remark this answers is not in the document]
cody-littley: will not do

sei-db/seiwal/seiwal_impl.go
```diff
@@ -0,0 +1,741 @@
...
+		case pruneRequest:
+			if err := w.pruneSealedFiles(m.through); err != nil {
+				w.fail(err)
+				return
+			}
+		case iteratorStartRequest:
+			resp := w.startIterator(m.startIndex)
+			m.reply <- resp
+			if resp.err != nil {
+				w.fail(resp.err)
```
blindchaser, line 462: Iterator construction failure bricks the entire write path, and hard-link support becomes a hard filesystem requirement

should we distinguish the different failure cases? like for MkdirAll/os.Link error, just return error instead of bricking the entire write path

sei-db/state_db/statewal/state_wal_impl.go
```diff
@@ -0,0 +1,265 @@
...
+	if ok {
+		w.currentBlock = last
+		w.currentBlockEnded = true
+		w.hasCurrentBlock = true
+	}
+	return w, nil
+}
+
+// Write accumulates a set of changes for the given block number in memory.
+func (w *stateWALImpl) Write(blockNumber uint64, cs []*proto.NamedChangeSet) error {
```
blindchaser, line 121: the nil changeset check is performed asynchronously in state_wal_serialization.go. As a result, Write() and potentially SignalEndOfBlock() can return successfully even when the input contains a nil changeset. The serialization goroutine discovers the error later and bricks the WAL, so the error surfaces at an unrelated subsequent operation. 

can we validate nil entries synchronously in Write() ? so the invalid input is rejected at the call site without bricking the WAL.

sei-db/seiwal/seiwal_file.go
```diff
@@ -0,0 +1,581 @@
...
+		parsed, ok := parseFileName(entry.Name())
+		if !ok || !parsed.sealed {
+			continue
+		}
+		prev, seen := kept[parsed.fileSeq]
+		if !seen {
+			kept[parsed.fileSeq] = parsed
+			continue
+		}
+		// Two sealed files share this sequence. Keep the one with the smaller last index (the rolled-back
```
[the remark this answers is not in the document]
cody-littley: no, this is a corruption vector that we don't need to automate. Mess with the files directory, accept consequences of handling a loaded footgun.

sei-db/seiwal/seiwal_file.go
```diff
@@ -0,0 +1,581 @@
...
+	hasRecords bool
+
+	// The number of bytes written to this file so far, including the header.
+	size uint64
+}
+
+// Create a new mutable WAL file on disk, writing its header, ready to accept records.
+func newWalFile(directory string, fileSeq uint64) (*walFile, error) {
+	path := filepath.Join(directory, unsealedFileName(fileSeq))
+	file, err := os.Create(path) //nolint:gosec // path derived from a validated directory
```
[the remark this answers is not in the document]
cody-littley: no

sei-db/seiwal/seiwal_impl.go
cody-littley: added file locks

sei-db/seiwal/seiwal_impl.go
cody-littley: Iterator failures are no longer fatal for failure modes that do not lead to corrupted state.

sei-db/state_db/statewal/state_wal_impl.go
cody-littley: change made

sei-db/seiwal/seiwal_impl.go
```diff
@@ -0,0 +1,760 @@
...
+	// rejects structural corruption (a gap in the sealed sequence).
+	sealedFiles, nextFileSeq, err := scanSealedFiles(config.Path)
+	if err != nil {
+		return nil, err
+	}
+
+	mutable, err := newWalFile(config.Path, nextFileSeq)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open mutable WAL file: %w", err)
+	}
```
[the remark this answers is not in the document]
cody-littley: comment is nonsensical as of the latest commit

sei-db/seiwal/seiwal_impl.go
```diff
@@ -0,0 +1,760 @@
...
+
+	if w.hasAppended {
+		if w.config.PermitGaps {
+			if index <= w.lastAppendIndex {
+				return fmt.Errorf(
+					"append rejected: index %d is not greater than last appended index %d",
+					index, w.lastAppendIndex)
+			}
+		} else if index != w.lastAppendIndex+1 {
```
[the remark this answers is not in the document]
cody-littley: no

sei-db/seiwal/seiwal_offline.go
```diff
@@ -0,0 +1,101 @@
...
+		if entry.IsDir() {
+			continue
+		}
+		parsed, ok := parseFileName(entry.Name())
+		if !ok || !parsed.sealed {
+			continue
+		}
+		sealed = append(sealed, parsed)
+		names[parsed.fileSeq] = entry.Name()
+	}
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/seiwal/seiwal_impl.go
```diff
@@ -0,0 +1,760 @@
...
+		}
+		sealed = append(sealed, parsed)
+		names[parsed.fileSeq] = entry.Name()
+	}
+	sort.Slice(sealed, func(i int, j int) bool { return sealed[i].fileSeq > sealed[j].fileSeq })
+
+	for _, parsed := range sealed {
+		if parsed.lastIndex <= rollbackThrough {
+			// This file lies entirely at or below the rollback point; so does every lower-sequence file. Done.
+			break
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/seiwal/seiwal_impl.go
```diff
@@ -0,0 +1,760 @@
...
+		})
+	}
+	// Flush the mutable file only when it is actually needed to cover the range. When endIndex is fully served
+	// by sealed files, the mutable file is left untouched.
+	if w.mutableFile.hasRecords && w.mutableFile.lastIndex >= startIndex && w.mutableFile.firstIndex <= endIndex {
+		if err := w.mutableFile.flush(false); err != nil {
+			return iteratorStartResponse{err: fmt.Errorf("failed to flush mutable file for iterator: %w", err)}
+		}
+		sources = append(sources, iteratorFile{
+			fileSeq: w.mutableFile.fileSeq, name: unsealedFileName(w.mutableFile.fileSeq),
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/seiwal/seiwal_impl.go
```diff
@@ -0,0 +1,760 @@
...
+			}
+		case iteratorStartRequest:
+			resp := w.startIterator(m.startIndex, m.endIndex)
+			m.reply <- resp
+			// A rejected range is the caller's error and leaves the WAL healthy; only a genuine I/O failure
+			// (a failed flush or hard-link) bricks the pipeline.
+			if resp.err != nil && !errors.Is(resp.err, ErrIteratorRange) {
+				w.fail(resp.err)
+				return
+			}
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/state_db/statewal/state_wal_impl.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+	if w.fatalErr != nil {
+		return fmt.Errorf("state WAL failed: %w", w.fatalErr)
+	}
+	if err := w.enforceWriteOrdering(blockNumber); err != nil {
+		return fmt.Errorf("write rejected: %w", err)
+	}
+	w.buf = append(w.buf, cs...)
+	return nil
+}
```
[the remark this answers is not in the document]
cody-littley: fixed

.gitignore
```diff
@@ -67,4 +67,5 @@ sei-db/state_db/bench/cryptosim/data/**
 sei-db/state_db/bench/cryptosim/bin/
 sei-db/state_db/bench/cryptosim/logs/
 sei-db/ledger_db/block/blocksim/bin/
-sei-db/db_engine/litt/bin/
\ No newline at end of file
+sei-db/db_engine/litt/bin/
+sei-db/seiwal/walsim/bin/
```
blindchaser: it seems there is no such folder

sei-db/seiwal/seiwal_impl.go
```diff
@@ -0,0 +1,809 @@
...
+			return fmt.Errorf(
+				"append rejected: index %d is not contiguous with last appended index %d (expected %d)",
+				index, w.lastAppendIndex, w.lastAppendIndex+1)
+		}
+	}
+	w.lastAppendIndex = index
+	w.hasAppended = true
+
+	record := frameRecord(index, data)
+	if err := w.sendToWriter(dataToBeWritten{record: record, index: index}); err != nil {
```
blindchaser, line 300: what happens if the w.sendToWriter() fails but `lastAppendIndex` and `hasAppended` already advances. should we move `w.lastAppendIndex = index / w.hasAppended = true` before `w.sendToWriter`?

blindchaser (approved):

yzang2019 (approved):

.gitignore
cody-littley: Lol, this is from another PR, not sure how it made it into this one. Removing from this PR for the sake of cleanliness, will be reintroduced soon.

sei-db/seiwal/seiwal_impl.go
cody-littley: We can only fail `sendToWriter()` if the WAL is in a terminal state, i.e. it's shut down or it crashed. If the WAL is in a terminal state, it doesn't matter if `lastAppendIndex` is corrupted. The on-disk files can never be modified after this point, and the contract is that a WAL in this state should be discarded and is unusable.

PR #3708, opened 2026-07-06
feat(indexer): push result cap and ordering into KV tx search (PLT-748)

amir-deris:
## Summary

Follow-up to #3689 (PLT-748), which pushed the result cap and ordering down
into the KV **block** indexer but left the **tx** scan path unbounded, marked
`TODO(PLT-748)`. `TxSearch` still fetched the entire match set, sorted it, and
only then applied `MaxTxSearchResults` at the RPC layer — so a broad tx query
materialized and sorted far more results than the caller would ever see.

This PR bounds the tx scan path the same way the block indexer is bounded, and
de-duplicates the shared bounding helpers into the `indexer` package so both
indexers reuse one implementation.

## What changed

- **KV tx indexer (`tx/kv`) — bounded search:**
  - *Fast path* (`planBounded` / `searchBounded`): eligible when every condition
    is an equality (point-probeable) or a `tx.height` range (evaluable from the
    candidate height), with at least one equality to drive the scan. It scans the
    driver equality's secondary-index prefix `(tag, value)` — which orders by
    `(height, index)` — in `order_by` order, point-probes the remaining
    equalities per candidate via `store.Has(secondaryKey(...))`, evaluates
    height ranges from the candidate height, dedups by hash, and stops at
    `Limit`. Memory is bounded by results kept, not by total match cardinality.
  - *Fallback path* (`intersect` + `collectBounded`): queries with
    `CONTAINS`/`MATCHES`/`EXISTS`, non-height ranges, or **only** a `tx.height`
    range cannot drive an in-order point-probeable scan — the `tx.height`
    secondary index stores the height as a decimal string, so its key order is
    not numeric. These materialize the intersection as before, then order by
    `(height, index)` and cap.
- **RPC `TxSearch`:** validates `order_by` up front, pushes the cap and ordering
  into the indexer, and keeps the post-sort cap as a safety net for sinks that
  ignore the limit (mirrors `BlockSearch`). Removes the `TODO(PLT-748)`.
- **Shared bounding helpers:** hoisted `BoundedCap`, `HeightInRange`, and
  `PrefixUpperBound` (plus the `maxBoundedPrealloc` const) from `block/kv` into
  the `indexer` package as exported helpers, and updated both `block/kv` and
  `tx/kv` to use them. `parseHeightIndexFromKey` stays tx-local (tx keys carry an
  `index` component; block keys do not).

## Tests

- `TestTxSearchBounded` covers the fast path (equality driver, multi-equality
  probe, equality + `tx.height` range, `tx.height` equality driver) and the
  fallback path (`tx.height`-range-only, `CONTAINS`), across asc/desc and
  various limits. It also asserts a cancelled context returns partial results
  without error, and that the bounded top-N equals the first N of the same query
  run unbounded — guarding against divergence between the fast and fallback
  paths.
- Existing `TestTxSearch*` suites are unchanged (behavior for the zero-value,
  unbounded opts is preserved).

## Known limitation / follow-up (PLT-786)

The fallback path still fully materializes its intersection before the cap is
applied — this covers `CONTAINS`/`MATCHES`/`EXISTS`, non-height value ranges, and
`tx.height`-range-only queries. These query shapes can't drive an in-order,
point-probeable scan in the current index format: the `tx.height` secondary
index stores the height as a decimal string (`orderedcode(TxHeightKey, "N", …)`),
so its key order is lexicographic, not numeric, and there is no equality to drive
off. The RPC layer still caps the *returned* set, so this is a transient
memory/CPU bound (scales with match cardinality), not an unbounded result.

Bounding these internally needs a height-ordered secondary index
(`orderedcode(compositeKey, height, value, index)`), which is a storage-format
change (write amplification + reindex/migration on upgrade) and out of scope
here. That work is tracked in **PLT-786**, which also makes numeric `tx.height`
ranges seekable for free. This PR intentionally bounds only the equality /
`tx.height`-point cases that the existing index can serve in order.


sei-tendermint/internal/state/indexer/tx/kv/kv_test.go
```diff
@@ -324,6 +324,204 @@ func TestTxSearchMultipleTxs(t *testing.T) {
...
+	}
+
+	// A cancelled context makes the bounded scan return the results gathered so
+	// far without an error, rather than failing the whole query.
+	t.Run("cancelled context returns partial results", func(t *testing.T) {
+		ctx, cancel := context.WithCancel(t.Context())
+		cancel()
+		results, err := idx.Search(ctx, query.MustCompile(`app.name = 'sei'`), indexer.SearchOptions{Limit: 5, OrderDesc: true})
+		require.NoError(t, err)
+		require.Empty(t, results)
```
bdchatham, line 448: This cancels the context before calling Search and then asserts the result is empty, so it exercises the top-level guard that returns early rather than the in-loop path in searchBounded that gathers rows and returns them on cancellation. The behavior the name describes, partial results before the cancel, never runs here, and the test would still pass if that path were broken. Worth refining it to cancel mid-scan after a few rows are gathered and assert a non-empty ordered prefix, so it actually guards the partial-return path.

sei-tendermint/internal/rpc/core/tx.go
```diff
@@ -65,40 +65,51 @@ func (env *Environment) TxSearch(ctx context.Context, req *coretypes.RequestTxSe
...
+				OrderDesc: orderDesc,
 			})
 			if err != nil {
 				return nil, err
 			}
 
 			// sort results (must be done before cap and pagination)
-			switch req.OrderBy {
-			case DescendingOrder, "":
+			if orderDesc {
```
bdchatham: Is this re-sort still needed now that the indexer returns the KV results already ordered and capped in this exact order, and TxSearch only ever handles the KV sink?

amir-deris: That's a good point, and it seems redundant. I will remove it.

sei-tendermint/internal/state/indexer/tx/kv/kv.go
```diff
@@ -231,16 +270,202 @@ hashes:
...
+		}
+
+		hash := it.Value()
+		if _, dup := seen[string(hash)]; dup {
+			continue
+		}
+
+		height, index, err := parseHeightIndexFromKey(it.Key())
+		if err != nil {
+			continue
```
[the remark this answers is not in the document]
amir-deris: Good catch on the behavioral delta being real, but I think this branch is actually unreachable for a well-formed store rather than just unlikely.

The driver prefix is `prefixFromCompositeKeyAndValue(tag, value)` = `orderedcode.Append(nil, compositeKey, value)` — two complete orderedcode segments. Because orderedcode terminates each segment, a key only matches this prefix if it begins with exactly those two segments. Every key we write is either a `secondaryKey` (4 segments) or a `primaryKey` (2 segments, `TxHashKey`-prefixed), so any key landing under an event `(tag, value)` prefix is necessarily a `secondaryKey` and parses cleanly into `(compositeKey, value, height, index)`. And the driver tag is never `TxHashKey` — hash conditions return early in `lookForHash` — so `primaryKey`s can't collide into the driver prefix either.

There's also an extra hurdle for the "old path returned it, fast path drops it" case: the old path would itself have dropped a foreign entry unless its value happened to be a valid tx hash pointing to a real tx (otherwise `Get` returns nil and it's filtered). Combined with the orderedcode argument, that can't be constructed against this store.

On the remedy: `TxIndex` only holds `store dbm.DB` with no logger, so a debug log would mean threading a logger through `NewTxIndex` and every caller for a branch that can't fire — not worth it. Instead I'll leave the `continue` as a defensive guard and add a short comment explaining why a parse failure here is unreachable, which covers the observability intent without the plumbing.

sei-will (approved):

PR #3710, opened 2026-07-07
fix(feegrant): bound DenomsSubsetOf cost and cap allowance denoms

amir-deris:
## Summary

Fixes a quadratic-complexity issue in fee-grant validation . An
allowance carrying a large coin list forced `Coins.DenomsSubsetOf` to run in
O(n·m), so a single crafted grant/allowance could burn disproportionate CPU
during `ValidateBasic`. This PR closes the hole on two fronts:

1. **Linear `DenomsSubsetOf`** — replace the per-element `AmountOf` membership
   check (O(n·m)) with a two-pointer merge walk (O(n+m)). Both coin sets are
   already sorted and duplicate-free per the `Coins` invariant, so the result is
   identical while the worst case drops from quadratic to linear.

2. **Bounded allowance denom count** — cap allowance coin lists at
   `MaxAllowanceDenoms = 100` and reject anything larger with a new
   `ErrTooManyDenoms`. Enforced in `BasicAllowance.ValidateBasic` (spend limit)
   and `PeriodicAllowance.ValidateBasic` (period spend limit + period can spend).


sei-cosmos/x/feegrant/basic_fee.go
```diff
@@ -7,6 +7,10 @@ import (
 
 var _ FeeAllowanceI = (*BasicAllowance)(nil)
 
+// MaxAllowanceDenoms bounds the number of coins allowed in a single allowance
+// spend-limit list.
+const MaxAllowanceDenoms = 5000
```
bdchatham: nit: I'd lower this to 100

bdchatham (approved):

sei-will (approved):

PR #3720, opened 2026-07-07
fix(mock_balances): mempool readiness gate must see the mock top-off (empty-blocks regression)

bdchatham:
## Problem

Since the TxMempool rewrite (CON-305, #3476, 2026-05-26), ready-promotion gates on the sender's **check-state balance** (`App.EvmBalance`, `app/abci.go`) vs the tx's `requiredBalance` (`internal/mempool/tx.go` ~426), and **gossip + block-reaping consume only ready txs**. The `mock_balances` top-off runs inside CheckTx on a cache branch that is discarded (`app/legacyabci/check_tx.go` — `anteCtx, _ := contextCacher(ctx)`, no `Write()`), so the balance the gate reads never reflects it.

Result on every multi-node `mock_balances` chain since the rewrite: `eth_sendRawTransaction` returns hashes (CheckTx passes via the ephemeral top-off), the tx is inserted **pending** and never promoted, never gossiped, never proposed — **zero transactions reach blocks** while load tools report healthy submission TPS. Before CON-305 there was no balance readiness gate, so mock txs gossiped and the delivery-path top-off (which does persist) funded them — i.e., this worked for ~2.5 years and regressed silently in late May.

Reproduced end-to-end on a harbor dev chain (4 validators + 1 RPC follower, `mock-nightly-20260706-068afb8`): seiload accepted 60k+ txs at ~1000 TPS; every block in the window had `num_txs=0` / `gasUsed=0x0`; a single traced tx sat in the follower's mempool and TTL-evicted while all four validators' `tendermint_mempool_cache_size` stayed 0 lifetime (nothing ever gossiped). The nightly harness did not catch this because `TestBenchmark` asserts Job-completion + liveness only (separate issue being filed).

## Fix

Floor the balance `EvmBalance` reports with `state.TopOffAmount`, via the existing `mock_balances` build-tag pair in package `app` (`mempoolBalanceFloor`). `EvmBalance`'s only consumer is the mempool readiness gate (`sei-tendermint/internal/mempool/tx.go:368` via the proxy), so consensus/execution semantics are untouched. Production (`!mock_balances`) builds get the identity function — zero behavior change.

## Verification
- `go build ./app/` and `go build -tags mock_balances ./app/` both clean
- `go test -tags mock_balances ./app/ -run TestMempoolBalanceFloor` passes
- Post-merge plan: respin the harbor repro chain on the fixed image and re-run the seiload profile — expect nonzero `num_txs`/`gasUsed` and receipts (will report on this PR)

### DIFF
--- a/app/mock_balances_buildtag.go
+++ b/app/mock_balances_buildtag.go
@@ -2,5 +2,25 @@
 
 package app
 
+import (
+	"math/big"
+
+	"github.com/sei-protocol/sei-chain/x/evm/state"
+)
+
 // MockBalancesEnabled indicates whether mock_balances build tag is set.
 const MockBalancesEnabled = true
+
+// mempoolBalanceFloor lifts the balance the mempool readiness gate sees to at
+// least the mock top-off amount. The CheckTx top-off writes to a discarded
+// cache, so the check-state balance behind EvmBalance never reflects it —
+// without this floor a mock-funded sender never satisfies requiredBalance,
+// its txs are never promoted to ready, and ready-only gossip/reaping means
+// they can never reach a block. Execution is unaffected: the real top-off
+// runs in the delivery path, where writes persist.
+func mempoolBalanceFloor(balance *big.Int) *big.Int {
+	if balance.Cmp(state.TopOffAmount) < 0 {
+		return new(big.Int).Set(state.TopOffAmount)
+	}
+	return balance
+}
--- a/app/abci.go
+++ b/app/abci.go
@@ -192,7 +192,7 @@ func (app *App) EvmBalance(evmAddr common.Address, seiAddrBz []byte) uint256.Int
 	if !seiAddr.Equals(sdk.AccAddress(evmAddr[:])) {
 		balance = new(big.Int).Add(balance, app.EvmKeeper.GetBalance(ctx, seiAddr))
 	}
-	return bigIntToUint256(balance)
+	return bigIntToUint256(mempoolBalanceFloor(balance))
 }
 
 func bigIntToUint256(x *big.Int) uint256.Int {
--- a/app/mock_balances_buildtag_default.go
+++ b/app/mock_balances_buildtag_default.go
@@ -2,5 +2,13 @@
 
 package app
 
+import "math/big"
+
 // MockBalancesEnabled indicates whether mock_balances build tag is set.
 const MockBalancesEnabled = false
+
+// mempoolBalanceFloor is the identity in production builds: the mempool
+// readiness gate sees the real check-state balance.
+func mempoolBalanceFloor(balance *big.Int) *big.Int {
+	return balance
+}
--- a/app/mock_balances_buildtag_test.go
+++ b/app/mock_balances_buildtag_test.go
@@ -1,0 +1,22 @@
+//go:build mock_balances
+
+package app
+
+import (
+	"math/big"
+	"testing"
+
+	"github.com/sei-protocol/sei-chain/x/evm/state"
+)
+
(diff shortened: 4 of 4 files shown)


app/mock_balances_buildtag.go
```diff
@@ -2,5 +2,25 @@
...
 package app
 
+import (
+	"math/big"
+
+	"github.com/sei-protocol/sei-chain/x/evm/state"
+)
+
 // MockBalancesEnabled indicates whether mock_balances build tag is set.
 const MockBalancesEnabled = true
```
bdchatham, line 12: Where was this used before the mempool rewrite? Was this logic somehow lost in that process?

bdchatham: Traced it through git — two-part answer:

**Where was `MockBalancesEnabled` used before?** One place only, ever: `app/invariance.go:49` (introduced 2026-01-30, #2779) — it skips the total-supply light-invariance check on mock builds because the top-off "fakes balances without updating the actual store, which would fail the supply check." It was never referenced by the mempool, before or after the rewrite. This PR's floor is the second consumer of the same pattern, and the same species of accommodation: an app-layer read made coherent with the mock contract.

**Was logic lost in the rewrite?** No — there was nothing to lose. The pre-rewrite mempool (`0d2452431^`) gossiped from a FIFO `gossipIndex` clist of everything that passed CheckTx; `git grep` shows zero `balance`/`EvmBalance`/`requiredBalance`/mock references in the old package. `mock_balances` was *implicitly* compatible with that model ('CheckTx passed ⇒ gossip'). The TxMempool rewrite (#3410 → #3476) introduced a brand-new balance-gated readiness model that never had mock parity. So this isn't restoring dropped code — it's adding the mock accommodation the new gate was born without.

**Convention check:** (a) extends the existing `mock_balances_buildtag{,_default}.go` pair rather than new plumbing; (b) keeps sei-tendermint mock-agnostic — the accommodation lives in package `app`, same layer as the invariance skip, rather than teaching the mempool about mocks; (c) no extra `pacific-1` guard needed here: the top-off itself panics on `pacific-1` at first `GetBalance` in the ante, which fires before mempool insertion could ever matter.

yzang2019 (approved):

bdchatham:
**Post-merge verification on harbor (as promised)** — identical chain spec + load profile, only the image changed (`mock-nightly-20260706-068afb8` → `mock-50c63c52…`):

| Signal (same probes both runs) | Before (v1) | After (v2, this fix) |
|---|---|---|
| Blocks with txs during 10-min seiload window | **0 of ~1,050** (num_txs=0, ~900B blocks, gasUsed=0x0) | **274 blocks**, up to **2,380 txs / ~50M gas in a single block** |
| Total included (validator block-store sum) | **0** of ~600k submitted | **7,968** of ~1.06M submitted |
| Validator `tendermint_mempool_cache_size` | 0 lifetime (nothing ever gossiped) | receiving + proposing |
| seiload block sampler | `gas(p50=0)` always | `gas(p50=13.8M…49.98M)` mid-run |

The readiness→gossip→proposal path is restored: mock-funded txs now become ready, gossip, and fill blocks to the gas ceiling.

**Honest caveat for anyone reading these numbers:** included/submitted is ~0.75% — that is overload attrition on a 4-validator + 1-follower dev chain being fed ~2,000 TPS (5s mempool TTL + nonce-contiguous readiness means evictions create nonce gaps that wedge accounts; the follower also lagged execution and its RPC saturated late-run). Separate perf characterization, not this bug — v1's number was zero regardless of rate. The nightly-harness inclusion assertion being filed separately would have caught the regression this PR fixes on night one.

PR #3727, opened 2026-07-08
fix(docker): giga-mixed cluster silently ran all-giga; write explicit executor config + role guard

bdchatham:
## The bug

The "EVM GIGA Mixed (Determinism)" job has never actually run a mixed cluster. Three facts combine:

1. `docker-compose.giga-mixed.yml` sets `GIGA_EXECUTOR=false` for nodes 1-3 (the intended V2 nodes).
2. `step4_config_override.sh` only wrote the `[giga_executor]` section when `GIGA_EXECUTOR=true` — the `false` nodes got no section at all.
3. `giga/executor/config.DefaultConfig` is `Enabled: true`, and `ReadConfig` only overrides when the key is present.

So the three "V2" nodes silently ran giga. The cluster was homogeneous all-giga, and the determinism test has been comparing giga against itself.

## The fix

- **`step4_config_override.sh`**: write `[giga_executor]` with an explicit `enabled` / `occ_enabled` for **every** node, so section-absent can never silently select an executor. The env defaults are aligned with the app's `DefaultConfig` (`GIGA_EXECUTOR=${GIGA_EXECUTOR:-true}`, same for OCC): an unset env keeps the executor an unconfigured seid would run, so **every other matrix leg keeps its current effective engine** — only an explicit `GIGA_EXECUTOR=false` (the giga-mixed compose's V2 nodes) selects V2. The in-place update is range-scoped to the giga block so it cannot touch `[telemetry] enabled`.
- **`evm_giga_mixed_tests.sh`**: a pre-test role guard that asserts each node's *effective* runtime mode — the `Giga Executor ... ENABLED/DISABLED` startup line each node logs after reading its config — matches its compose role (node 0 giga, nodes 1-3 V2). It polls up to 60s for the signal (a slow boot is not a failure), and on mismatch or a missing signal it stops the cluster and exits 1 naming the node, instead of running a vacuous test. Checking the runtime signal (not app.toml) also catches any future default flip.

## Blast radius

`step4` runs for every matrix leg, so the always-write change was checked against all of them at executor AND OCC granularity. Legs with `GIGA_EXECUTOR=true` (GIGA/Autobahn) and legs with it unset (everything else) keep their current effective engine — unset resolves to the app default (giga+OCC on), which is what section-absent legs were already running. One leg needed an explicit pin: Autobahn RPC .io/.iox set `GIGA_EXECUTOR=true` without `GIGA_OCC`, and the old shell default (`false`) made it the matrix's only sequential-giga leg — it now carries `GIGA_OCC=false` explicitly, preserving that coverage rather than leaving it to a shell default. Net: every leg keeps its pre-PR effective mode; the only behavior change is the giga-mixed job's V2 nodes, which now genuinely run V2.

## Heads-up for reviewers

Once merged, this job runs genuinely mixed for the first time. If giga and V2 are not byte-for-byte consistent on this workload, node 0 will halt on a real consensus mismatch that the homogeneous config was masking — i.e. the job may start failing for a legitimate reason. (An in-process mixed giga/V2 run passed direct state-root equality on a send/deploy workload, so consistency is plausible — but this suite's load differs.)

Verified by hand-trace on synthetic configs (GNU/BSD/BusyBox sed) and synthetic startup logs: explicit-false nodes get `enabled = false`; the guard passes correct roles and fails a silently-flipped node, a signal-absent node, and a missing log.

### DIFF
--- a/docker/localnode/scripts/step4_config_override.sh
+++ b/docker/localnode/scripts/step4_config_override.sh
@@ -1,8 +1,10 @@
 #!/usr/bin/env sh
 
 NODE_ID=${ID:-0}
-GIGA_EXECUTOR=${GIGA_EXECUTOR:-false}
-GIGA_OCC=${GIGA_OCC:-false}
+# Defaults mirror the app's DefaultConfig (giga+OCC on): unset runs what an
+# unconfigured seid would; only an explicit false selects V2.
+GIGA_EXECUTOR=${GIGA_EXECUTOR:-true}
+GIGA_OCC=${GIGA_OCC:-true}
 AUTOBAHN=${AUTOBAHN:-false}
 GIGA_STORAGE=${GIGA_STORAGE:-false}
 # GIGA_FLATKV_ONLY=true boots the cluster directly in the terminal v3
--- a/docker/localnode/scripts/step4_config_override.sh
+++ b/docker/localnode/scripts/step4_config_override.sh
@@ -105,28 +105,34 @@ if [ "$GIGA_STORAGE" = "true" ] && [ "$GIGA_MIGRATE_FROM_MEMIAVL" != "true" ] &&
   sed -i 's/^evm-ss-split[[:space:]]*=.*/evm-ss-split = true/' ~/.sei/config/app.toml
 fi
 
-# Enable Giga Executor if requested
+# Write the [giga_executor] section explicitly for every node. The Go config
+# default is Enabled=true and ReadConfig only overrides keys that are present,
+# so a node that never writes the section silently runs giga regardless of its
+# intended role. Emitting enabled/occ_enabled unconditionally keeps each node's
+# executor mode self-describing in app.toml and immune to that default flip.
 if [ "$GIGA_EXECUTOR" = "true" ]; then
+  GIGA_ENABLED_VALUE=true
   echo "Enabling Giga Executor for node $NODE_ID..."
-  if grep -q "\[giga_executor\]" ~/.sei/config/app.toml; then
-    # If the section exists, update enabled to true
-    sed -i 's/enabled = false/enabled = true/' ~/.sei/config/app.toml
-  else
-    # If section doesn't exist, append it
-    echo "" >> ~/.sei/config/app.toml
-    echo "[giga_executor]" >> ~/.sei/config/app.toml
-    echo "enabled = true" >> ~/.sei/config/app.toml
-    echo "occ_enabled = false" >> ~/.sei/config/app.toml
-  fi
+else
+  GIGA_ENABLED_VALUE=false
+  echo "Configuring node $NODE_ID with the V2 executor (Giga disabled)..."
+fi
+if [ "$GIGA_OCC" = "true" ]; then
+  GIGA_OCC_VALUE=true
+else
+  GIGA_OCC_VALUE=false
+fi
 
-  # Set OCC based on GIGA_OCC flag
-  if [ "$GIGA_OCC" = "true" ]; then
-    echo "Enabling OCC for Giga Executor on node $NODE_ID..."
-    sed -i 's/occ_enabled = false/occ_enabled = true/' ~/.sei/config/app.toml
-  else
-    echo "Disabling OCC for Giga Executor (sequential mode) on node $NODE_ID..."
-    sed -i 's/occ_enabled = true/occ_enabled = false/' ~/.sei/config/app.toml
-  fi
+if grep -q "\[giga_executor\]" ~/.sei/config/app.toml; then
+  # Section already present: rewrite its keys in place. The range address scopes
+  # the substitutions to the giga block so an identically-named key in another
+  # section (e.g. [telemetry] enabled) is never rewritten.
+  sed -i "/^\[giga_executor\]/,/^\[/{s/^enabled = .*/enabled = $GIGA_ENABLED_VALUE/;s/^occ_enabled = .*/occ_enabled = $GIGA_OCC_VALUE/;}" ~/.sei/config/app.toml
+else
+  echo "" >> ~/.sei/config/app.toml
+  echo "[giga_executor]" >> ~/.sei/config/app.toml
+  echo "enabled = $GIGA_ENABLED_VALUE" >> ~/.sei/config/app.toml
+  echo "occ_enabled = $GIGA_OCC_VALUE" >> ~/.sei/config/app.toml
 fi
 
 # Override receipt store backend if requested
--- a/integration_test/evm_module/scripts/evm_giga_mixed_tests.sh
+++ b/integration_test/evm_module/scripts/evm_giga_mixed_tests.sh
@@ -57,22 +57,24 @@ sleep 10
 # this checks the effective runtime mode rather than what the config intended.
 assert_giga_mode() {
     node_id="$1"
-    expected="$2" # "enabled" or "disabled"
+    expected="$2" # "occ", "sequential", or "disabled"
     log="build/generated/logs/seid-${node_id}.log"
 
-    # Poll: a slow boot must not read as a missing signal.
+    # Poll: a slow boot must not read as a missing signal. Match the LAST
+    # "Giga Executor" line so a restart under a different config can't leave an
+    # earlier, stale signal to read ambiguously. OCC must be tested before the
+    # bare ENABLED pattern — "with OCC is ENABLED" also contains "is ENABLED".
     deadline=60
     waited=0
     actual=""
     while [ "$waited" -lt "$deadline" ]; do
         if [ -f "$log" ]; then
-            if grep -q "Giga Executor is DISABLED" "$log"; then
-                actual="disabled"
-                break
-            elif grep -q "Giga Executor.*is ENABLED" "$log"; then
-                actual="enabled"
-                break
-            fi
+            signal=$(grep "Giga Executor" "$log" | tail -1)
+            case "$signal" in
+                *"with OCC is ENABLED"*) actual="occ"; break ;;
+                *"is ENABLED"*)          actual="sequential"; break ;;
+                *"is DISABLED"*)         actual="disabled"; break ;;
+            esac
         fi
         sleep 2
         waited=$((waited + 2))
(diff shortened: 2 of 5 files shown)


docker/localnode/scripts/step4_config_override.sh
```diff
@@ -105,28 +105,34 @@ if [ "$GIGA_STORAGE" = "true" ] && [ "$GIGA_MIGRATE_FROM_MEMIAVL" != "true" ] &&
...
-    sed -i 's/occ_enabled = false/occ_enabled = true/' ~/.sei/config/app.toml
-  else
-    echo "Disabling OCC for Giga Executor (sequential mode) on node $NODE_ID..."
-    sed -i 's/occ_enabled = true/occ_enabled = false/' ~/.sei/config/app.toml
-  fi
+if grep -q "\[giga_executor\]" ~/.sei/config/app.toml; then
+  # Section already present: rewrite its keys in place. The range address scopes
+  # the substitutions to the giga block so an identically-named key in another
+  # section (e.g. [telemetry] enabled) is never rewritten.
+  sed -i "/^\[giga_executor\]/,/^\[/{s/^enabled = .*/enabled = $GIGA_ENABLED_VALUE/;s/^occ_enabled = .*/occ_enabled = $GIGA_OCC_VALUE/;}" ~/.sei/config/app.toml
```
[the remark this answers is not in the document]
bdchatham: Tested empirically — this is a false positive. In sed (POSIX/GNU/BSD), a range's end address is only tested starting from the line *after* the start match, so the `[giga_executor]` header cannot terminate its own range. On this input:

```toml
[telemetry]
enabled = true

[giga_executor]
enabled = true
occ_enabled = true

[receipt-store]
rs-backend = "legacy"
```

the exact command from step4 produces `enabled = false` / `occ_enabled = false` under `[giga_executor]` with `[telemetry] enabled` untouched, on all three of: macOS BSD sed, GNU sed 4.9, and `ubuntu:24.04`'s sed (the localnode base image). When `[giga_executor]` is the last section (the CI append path writes it at EOF), the range runs to EOF and still rewrites both keys. CI always takes the append branch anyway — the committed `docker/localnode/config/app.toml` has no `[giga_executor]` section; the in-place branch exists only for idempotent re-runs.

integration_test/evm_module/scripts/evm_giga_mixed_tests.sh
```diff
@@ -49,6 +49,62 @@ fi
...
+    # Poll: a slow boot must not read as a missing signal.
+    deadline=60
+    waited=0
+    actual=""
+    while [ "$waited" -lt "$deadline" ]; do
+        if [ -f "$log" ]; then
+            if grep -q "Giga Executor is DISABLED" "$log"; then
+                actual="disabled"
+                break
+            elif grep -q "Giga Executor.*is ENABLED" "$log"; then
```
amir-deris: **[nit]** The guard verifies executor enabled/disabled but not OCC. Node 0 is documented as `GIGA_OCC=true`; consider additionally asserting the `"with OCC is ENABLED"` signal for node 0 so the concurrent-executor role (the stated purpose of the mixed job) is also covered, not just giga-vs-V2.

For example, require `"with OCC is ENABLED"` for node 0 instead of the broader `Giga Executor.*is ENABLED` pattern.

integration_test/evm_module/scripts/evm_giga_mixed_tests.sh
```diff
@@ -49,6 +49,62 @@ fi
...
+    expected="$2" # "enabled" or "disabled"
+    log="build/generated/logs/seid-${node_id}.log"
+
+    # Poll: a slow boot must not read as a missing signal.
+    deadline=60
+    waited=0
+    actual=""
+    while [ "$waited" -lt "$deadline" ]; do
+        if [ -f "$log" ]; then
+            if grep -q "Giga Executor is DISABLED" "$log"; then
```
amir-deris: **[nit]** `grep -q` over the whole log file means a node restart mid-run with a different config could leave stale ENABLED/DISABLED lines and produce an ambiguous read. Fine for the current fresh-cluster CI flow; if this ever runs on a long-lived cluster, matching the *last* occurrence (e.g. `grep … | tail -1`) would be more robust.

amir-deris (commented):
Non-blocking follow-ups from review (no blockers — looks good to merge):

**`app/app.go` (~827–832):** Consider a one-line comment near the `Giga Executor … ENABLED/DISABLED` startup log strings noting that `evm_giga_mixed_tests.sh` depends on them. The guard fails closed on rewording, but a comment prevents accidental breakage in a future log cleanup.

**Blast-radius nuance:** One matrix leg sets `GIGA_EXECUTOR=true` without `GIGA_OCC` — "Autobahn RPC .io/.iox (spec fixtures)" in `integration-test-matrix.json` line 239. Previously (shell default `GIGA_OCC=false`) that leg ran sequential giga; after this PR (shell default `GIGA_OCC=true`) it runs OCC giga. CI passed for that leg, so not a blocker — just worth being aware of.

**`integration_test/evm_module/rpc_io_test/RPC_IO_README.md` line 44:** Says unset `GIGA_EXECUTOR` means V2, but Go `DefaultConfig` has `Enabled: true` — so unset has meant giga-on for a while. Out of scope here, but worth a follow-up doc fix so local devs aren't surprised.

integration_test/evm_module/scripts/evm_giga_mixed_tests.sh
amir-deris: **[nit]** The guard verifies executor enabled/disabled but not OCC. Node 0 is documented as `GIGA_OCC=true`; consider additionally asserting the `"with OCC is ENABLED"` signal for node 0 so the concurrent-executor role (the stated purpose of the mixed job) is also covered, not just giga-vs-V2.

For example, require `"with OCC is ENABLED"` for node 0 instead of the broader `Giga Executor.*is ENABLED` pattern.

integration_test/evm_module/scripts/evm_giga_mixed_tests.sh
amir-deris: **[nit]** `grep -q` over the whole log file means a node restart mid-run with a different config could leave stale ENABLED/DISABLED lines and produce an ambiguous read. Fine for the current fresh-cluster CI flow; if this ever runs on a long-lived cluster, matching the *last* occurrence (e.g. `grep … | tail -1`) would be more robust.

amir-deris (approved):

integration_test/evm_module/scripts/evm_giga_mixed_tests.sh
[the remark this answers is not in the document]
bdchatham: Done in 5d706b8 — node 0 now asserts the exact `with OCC is ENABLED` signal (expected mode `occ`), so a giga-without-OCC boot fails the guard rather than passing on the broad `is ENABLED` match.

bdchatham: Done in 5d706b8 — node 0 now asserts the exact `with OCC is ENABLED` signal (expected mode `occ`), so a giga-without-OCC boot fails the guard rather than passing on the broad `is ENABLED` match.

bdchatham: Done in 5d706b8 — node 0 now asserts the exact `with OCC is ENABLED` signal (expected mode `occ`), so a giga-without-OCC boot fails the guard rather than passing on the broad `is ENABLED` match.

integration_test/evm_module/scripts/evm_giga_mixed_tests.sh
bdchatham: Done in 5d706b8 — the guard now reads `grep "Giga Executor" "$log" | tail -1` and case-matches the last line, so a stale ENABLED/DISABLED line from a prior config can no longer produce an ambiguous read. Poll loop + 60s deadline are unchanged.

bdchatham: Done in 5d706b8 — the guard now reads `grep "Giga Executor" "$log" | tail -1` and case-matches the last line, so a stale ENABLED/DISABLED line from a prior config can no longer produce an ambiguous read. Poll loop + 60s deadline are unchanged.

bdchatham:
@amir-deris thanks — follow-ups addressed in 5d706b8:

- **app.go:** added a one-line comment by the `ENABLED/DISABLED` startup logs noting `evm_giga_mixed_tests.sh` matches these strings, so a future log cleanup doesn't silently break the guard.
- **RPC_IO_README:** corrected — leaving `GIGA_EXECUTOR` unset runs giga (`DefaultConfig` is `Enabled: true`), not V2; `GIGA_EXECUTOR=false` selects legacy. Pulled it in here since it's a one-liner.
- **Autobahn .io/.iox leg:** intentional consequence of aligning the shell defaults with the app's `DefaultConfig` — that leg (`GIGA_EXECUTOR=true`, no `GIGA_OCC`) now runs OCC giga, which is the prod path, and CI passed on it. If sequential-giga coverage is wanted there, an explicit `GIGA_OCC=false` on that matrix entry restores it — happy to add if you'd prefer.

blindchaser (approved):

docker/localnode/scripts/step4_config_override.sh
```diff
@@ -1,8 +1,10 @@
 #!/usr/bin/env sh
 
 NODE_ID=${ID:-0}
-GIGA_EXECUTOR=${GIGA_EXECUTOR:-false}
-GIGA_OCC=${GIGA_OCC:-false}
+# Defaults mirror the app's DefaultConfig (giga+OCC on): unset runs what an
+# unconfigured seid would; only an explicit false selects V2.
+GIGA_EXECUTOR=${GIGA_EXECUTOR:-true}
+GIGA_OCC=${GIGA_OCC:-true}
```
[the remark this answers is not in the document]
bdchatham: Fixed in 033b202 via your second option — but note the first option (defaulting `GIGA_OCC` back to `false`) would itself change ~19 legs: every section-absent leg ran the app default `OCCEnabled: true` pre-PR, so a global `false` default flips those from OCC to sequential. No single default preserves the whole matrix, because the pre-PR effective OCC depended on which branch a leg took (section-absent → app default `true`; `GIGA_EXECUTOR=true` → shell default `false`).

The exact-preserving fix is the explicit pin: `GIGA_OCC=false` on the Autobahn RPC .io/.iox leg. Every leg now keeps its pre-PR effective mode, and the matrix's one sequential-giga leg is explicit rather than an accident of shell defaults. PR body updated to state the blast radius at OCC granularity.

PR #3745, opened 2026-07-10
Align FlatKV snapshot config with MemIAVL

yzang2019:
Two main changes:
1. We set the default snapshot-keep-recent to 1, so don't allow 0 any more in order to keep more history for snapshot creation
2. FlatKV snapshot and keep-recent should be aligned with MemIAVL when CompositeSC is being used 
3. FlatKV also default to 10000 interval and keeprecent of 1 to match with memIAVL

Also removed the check and validation for state sync, because state sync snapshot creation doesn't really require SS to be enabled at all.

### DIFF
--- a/app/seidb.go
+++ b/app/seidb.go
@@ -80,7 +80,6 @@ func SetupSeiDB(
 			"separateDBs", ssConfig.SeparateEVMSubDBs,
 		)
 	}
-	validateConfigs(appOpts)
 	gigaExecutorConfig, err := gigaconfig.ReadConfig(appOpts)
 	if err != nil {
 		panic(fmt.Sprintf("error reading giga executor config due to %s", err))
@@ -217,16 +216,4 @@ func parseSSConfigs(appOpts servertypes.AppOptions) config.StateStoreConfig {
 	ssConfig.SeparateEVMSubDBs = cast.ToBool(appOpts.Get(FlagEVMSSSeparateDBs))
 	ssConfig.EVMSplit = cast.ToBool(appOpts.Get(FlagEVMSSSplit))
 	return ssConfig
-}
-
-func validateConfigs(appOpts servertypes.AppOptions) {
-	scEnabled := cast.ToBool(appOpts.Get(FlagSCEnable))
-	ssEnabled := cast.ToBool(appOpts.Get(FlagSSEnable))
-	snapshotExportInterval := cast.ToUint64(appOpts.Get(FlagSnapshotInterval))
-	// Make sure when snapshot is enabled, we should enable SS store
-	if snapshotExportInterval > 0 && scEnabled {
-		if !ssEnabled {
-			panic(fmt.Sprintf("Config validation failed, SeiDB SS store needs to be enabled when snapshot interval %d > 0", snapshotExportInterval))
-		}
-	}
 }
--- a/app/seidb.go
+++ b/app/seidb.go
@@ -114,6 +114,10 @@ func parseSCConfigs(appOpts servertypes.AppOptions) config.StateCommitConfig {
 	scConfig.MemIAVLConfig.SnapshotWriteRateMBps = cast.ToInt(appOpts.Get(FlagSCSnapshotWriteRateMBps))
 	scConfig.FlatKVConfig.EnableReadWriteMetrics = cast.ToBool(appOpts.Get(FlagSCFlatKVReadWriteMetrics))
 
+	// Now that the raw flags are parsed, floor memIAVL snapshot-keep-recent and
+	// mirror the (unexposed) FlatKV snapshot cadence onto the memIAVL settings.
+	scConfig.AlignFlatKVWithMemIAVL()
+
 	// sc-write-mode-enable-auto (default true) decides whether the node may run
 	// in auto. An ABSENT key keeps the default (true): nodes provisioned by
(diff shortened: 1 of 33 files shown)


app/seidb.go
```diff
@@ -80,7 +80,6 @@ func SetupSeiDB(
 			"separateDBs", ssConfig.SeparateEVMSubDBs,
 		)
 	}
-	validateConfigs(appOpts)
```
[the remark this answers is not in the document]
yzang2019: That's intentional

sei-db/config/sc_config.go
```diff
@@ -125,3 +125,18 @@ func (c StateCommitConfig) Validate() error {
...
+// AlignFlatKVWithMemIAVL applies the FlatKV-follows-memIAVL policy that the app
+// config layer imposes because FlatKV's snapshot knobs are not independently
+// exposed in app.toml. It is applied by app config parsing after the raw flags
+// are read, so the underlying config layers stay a faithful mapping of
+// app.toml/flags:
+//   - the memIAVL snapshot-keep-recent is floored to memiavl.MinSnapshotKeepRecent
+//     (a configured 0, "keep only the current snapshot", becomes 1); and
+//   - FlatKV's snapshot cadence mirrors the (floored) memIAVL settings so both
+//     backends checkpoint and prune in lockstep.
+func (c *StateCommitConfig) AlignFlatKVWithMemIAVL() {
```
blindchaser: `sc-snapshot-interval = 0` will the lockstep invariant and silently disables FlatKV snapshots

parseSCConfigs mirrors the raw memIAVL interval onto FlatKV before store construction, but the two backends normalize 0 differently: memIAVL treats a configured 0 as "use the default 10000" (applied later, in DB.Load), while FlatKV treats 0 as "auto-snapshots disabled"

sei-cosmos/server/config/config.go
```diff
@@ -457,7 +457,14 @@ func GetConfig(v *viper.Viper) (Config, error) {
 	}
 	scWriteMode = config.ApplyWriteModeAuto(scWriteModeEnableAuto, scWriteMode)
 
+	// FlatKV knobs are not rendered in the default app.toml template. By default
+	// FlatKV's snapshot cadence mirrors the memIAVL (SC) settings so both backends
+	// checkpoint/prune in lockstep, but explicit state-commit.flatkv.* keys are
+	// still parsed and honored if an operator adds them by hand (they take
```
blindchaser: it looks the running node's store is built from app/seidb.go parseSCConfigs (via SetupSeiDB → rootmulti.NewStore), which never reads those keys

sei-cosmos/server/config/config.go
```diff
@@ -457,7 +457,14 @@ func GetConfig(v *viper.Viper) (Config, error) {
...
+	// precedence over the memIAVL mirror).
 	flatKVConfig := config.DefaultStateCommitConfig().FlatKVConfig
+	flatKVConfig.SnapshotInterval = v.GetUint32("state-commit.sc-snapshot-interval")
```
blindchaser: if the `state-commit.sc-snapshot-interval` and `state-commit.sc-keep-recent` keys absent, line 465 default config will be overwrite by the line 466-467 to be 0/0. before this pr, it will uuse the default value even absent

sei-cosmos/server/config/config.go
yzang2019: Only snapshot is derived from MemIAVL, other FlatKV configs can still parsed from config.toml -> Flags -> app/seidb.go

sei-db/config/sc_config.go
yzang2019: What do you think should the correct behavior be when we set 0 for FlatKV? Do we allow snapshot to be disabled or shall we just override it to default and not allow 0

cody-littley (approved):

sei-cosmos/server/config/config.go
```diff
@@ -457,6 +457,12 @@ func GetConfig(v *viper.Viper) (Config, error) {
 	}
 	scWriteMode = config.ApplyWriteModeAuto(scWriteModeEnableAuto, scWriteMode)
 
+	// FlatKV knobs are not rendered in the default app.toml template. GetConfig
+	// is a faithful parse of app.toml/flags: it only reads the explicit
+	// state-commit.flatkv.* keys (if an operator adds them by hand) on top of the
+	// in-code defaults. The FlatKV-follows-memIAVL mirror (and snapshot cadence
+	// normalization) is applied later by StateCommitConfig.AlignFlatKVWithMemIAVL
```
blindchaser: There is no StateCommitConfig.AlignFlatKVWithMemIAVL method. The real function is the unexported package-level composite.alignFlatKVSnapshotWithMemIAVL. Same wrong name appears in config_test.go (lines ~484, ~513). Minor, but it will mislead the next reader searching for it.

sei-cosmos/server/config/config.go
blindchaser, line 419: new GetConfig tests only cover the FlatKV branch. suggest to add testings asserting memIAVL keeps its in-code defaults when the sc-* keys are absent

blindchaser (approved):

sei-cosmos/server/config/config.go
yzang2019: Yeah will fix

PR #3747, opened 2026-07-10
Add shared scan budget for KV tx_search and block_search

amir-deris:
## Summary

Adds a process-wide **scan budget** that bounds the total number of KV index
entries all in-flight `tx_search` and `block_search` requests may visit at
once. Where `max-tx-search-results` caps the *result set* per query, this caps
the *work done* across every concurrent scan — a complementary safety valve
that bounds peak memory (and scan CPU) under broad or highly concurrent search
load.

Builds on the earlier tx/block KV search bounding work (PLT-786): the fast
paths already stream and cap results, but a broad fallback query — or many of
them at once — can still walk a large slice of the index. The shared budget is
the backstop for that case.

## How it works

- New `indexer.ScanBudget` (`budget.go`): a single atomic-counter ceiling on
  concurrently-charged entries. `nil` / zero / non-positive max all mean
  unlimited.
- Each `Search` opens a `ScanLease` and calls `lease.Visit(1)` at every index
  entry it walks (driver-entry scans in the bounded fast path; each iterator
  step in the `match`/`matchRange` fallbacks). When the aggregate in-flight
  charge exceeds the ceiling, the in-flight search aborts with
  `ErrScanBudgetExceeded` rather than continuing to accumulate.
- `lease.Release()` (via `defer`) returns the charge to the shared pool when
  the search returns, so the budget tracks *concurrent* work, not cumulative.
- The tx and block indexers share **one** budget (wired in
  `sink.EventSinksFromConfig`), so the cap is truly process-wide across both
  search types.

## Config

New RPC config `max-search-scan-budget` (default `100_000`; `0` disables,
not recommended on public nodes). Validated as non-negative in `ValidateBasic`
and documented in the generated `config.toml`.

## Other changes

- Tx KV indexer `match`/`matchRange` now **return errors** instead of
  `panic`-ing on iterator/DB failures, so a scan failure (budget exceeded or
  otherwise) surfaces cleanly to the caller instead of crashing the node.
- `kv.NewEventSink` gains a `*indexer.ScanBudget` parameter. Call sites that
  only write (event reindexing) and tests pass `nil` (unlimited).

## Testing

- `go test ./sei-tendermint/internal/state/indexer/...`


sei-tendermint/internal/state/indexer/budget.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"errors"
+	"sync/atomic"
+)
+
+// ErrScanBudgetExceeded is returned by a KV index Search when the aggregate
+// number of index entries visited by all in-flight searches would exceed the
+// configured scan budget. It signals overload: the caller should narrow the
+// query or retry later.
+var ErrScanBudgetExceeded = errors.New("kv indexer scan budget exceeded; narrow the query or retry later")
```
bdchatham, line 12: This is a load-shedding control with no telemetry on the shed, which I'd want to close before it fronts public-node traffic. Two things are missing. `ErrScanBudgetExceeded` is a plain error and both RPC handlers bare-return it, so it serializes the same as a real iterator or DB fault, and an operator can't tell benign back-pressure from a regression that should page someone. Nothing counts or logs when the cap bites either, so `InFlight` (the "for tests and metrics" comment above) is wired to nothing outside the tests.

I'd surface it as a distinct, retryable error classified with `errors.Is` at the tx_search/block_search handlers, and add a counter at the `Visit` error sites plus an `InFlight` gauge, so the headroom is visible before the ceiling trips and the shedding is visible after. Happy to leave the exact metric wiring to the observability side, the ask here is that the budget stops being a silent control.

sei-tendermint/config/config.go
```diff
@@ -536,6 +536,13 @@ type RPCConfig struct {
 	// Maximum number of results returned by tx_search and block_search.
 	// 0 disables the cap (not recommended on public nodes).
 	MaxTxSearchResults int `mapstructure:"max-tx-search-results"`
+
+	// Maximum number of KV index entries that all in-flight tx_search and
+	// block_search requests may visit at once. This is a process-wide safety
+	// cap that bounds peak memory (and scan CPU) under broad or highly
```
bdchatham, line 534: "bounds peak memory" overstates what the counter measures on the tx fallback path. The charge is one per index entry visited, but `collectBounded` then calls `Get` on every entry in `filteredHashes`, materializing the full transaction body and events, and only trims to `opts.Limit` after the whole set is built and sorted. So peak is bounded by the entries visited times the average tx size, in entries rather than bytes, and the result cap does not bound it. A `someattr EXISTS` over ~90k txs stays under the 100k default and materializes ~90k full EVM bodies before truncating to 10k. Block search is fine here since it yields int64 heights.

Two ways to close it, either is fine. Reword the claim to what the counter actually bounds, index entries scanned and thereby the fallback match-set, or apply `opts.Limit` as a hard cap on `filteredHashes` before the `Get` loop so the materialization is bounded by results kept. I lean towards the second since it makes the comment true rather than softening it.

sei-will (approved):

PR #3749, opened 2026-07-13
Fix static seid SIGSEGV (pin pre-gcc-12 libgcc unwinder), add boot smoke gate, build binaries on manual tag pushes

monty-sei:

Fixes the SIGSEGV that made every statically linked `seid` crash on ~70% of boots (the reason the v6.6.0-rc2 binaries were pulled), and closes the two release pipeline gaps that let it ship and then blocked its replacement.

### The bug

gcc >= 12 replaced libgcc's DWARF unwind-frame registry with a lock-free b-tree (`libgcc/unwind-dw2-btree.h`). Wasmer registers unwind frames for every JIT-compiled wasm module (`__register_frame`, from `UnwindRegistry::publish`), and under that pattern the b-tree corrupts and dereferences a null/garbage node inside `btree_insert`. A static musl build links the registry from the build image's toolchain (current Alpine ships gcc 15), so every static `seid` carried it and crashed probabilistically at the genesis wasm store, the first wasmer compile. Caught live in gdb:

```
#0 version_lock_lock_exclusive (vl=0x0)              gcc-15 libgcc/unwind-dw2-btree.h:81
#2 btree_insert (t=<registered_frames>, ...)
#3 wasmer_compiler::engine::unwind::systemv::UnwindRegistry::publish
#9 cosmwasm_vm compile -> #12 do_save_wasm           (genesis pointer-contract store)
```

Dynamic glibc builds (the Docker image) use the shared `libgcc_s` runtime path and never hit this; upstream wasmd's static muslc binaries link a pre-b-tree libgcc, which is what this PR restores. Symbol hygiene, thread stacks, and Go signal settings were all ruled out experimentally (the crash reproduced identically with each changed in isolation).

### Changes

1. **Pin the pre-b-tree libgcc at the static link** (`third_party/alpine-gcc10-libgcc/`, vendored from Alpine 3.15's gcc 10.3.1 apk with full provenance and sha256 pins in its README; `scripts/build-static.sh` verifies the checksums each build and `Makefile` gains a `STATIC_EXTRA_LDFLAGS` hook to inject the search path). The build then asserts the final binary contains no `version_lock_lock_exclusive` symbol, so a toolchain upgrade cannot silently reintroduce the b-tree.
2. **Repeated-boot gate** (`scripts/boot-smoke.sh`): boots the built binary N times against throwaway genesis homes and asserts each run gets through the genesis wasm store to the ABCI handshake. Wired into the `linux-amd64-static` CI job (8 boots) and the goreleaser `before:` hook (4 boots). A ~30% survival rate passed our original single-boot verification; a repeated gate makes that class of luck vanishingly unlikely.
3. **Build binaries on manual tag pushes** (`uci-release-publish.yml`): the "Restrict tagging" ruleset blocks this workflow's token from creating tags, so the publish job fails on version.json pushes and a maintainer pushes the tag by hand, which left the chained goreleaser job gated off (`release_created` never true). This happened for rc1, rc2 and rc3. The goreleaser job now also fires on manually pushed `v*` tags, and the concurrency group is keyed per event+ref with `cancel-in-progress: false` so the tag push can't cancel an in-flight build (they share a SHA).

### Notes

- rc1/rc2/rc3 trees predate this fix, so their binaries cannot be rebuilt via the manual dispatch; the first good binaries ship with the next tag cut after this lands on `release/v6.6` (backport to follow, manual since this touches workflow files).
- Follow-ups tracked separately: report the b-tree bug upstream (gcc/wasmer), wasmvm archive symbol localisation to drop `-z muldefs` (hygiene, experimentally not this crash), optional musl-vs-glibc wasm benchmark.

- Root cause and fix validated on real amd64 runners against the v6.6.0-rc2 source: broken recipe crashed 7/10 boots (and 8/8 under a deterministic reproducer); with only the libgcc pin changed, 18/18 boots clean and the reproducer eliminated.
- Full verification of the pinned build: 3-node localnet committed 450+ blocks, wasm store/instantiate/execute transactions all succeeded (`hackatom.wasm`, tx code 0), Tendermint and EVM RPC endpoints served, full and seed modes ran clean, zero crash signals across all node logs.
- `goreleaser check` and `actionlint` pass; `nm` on the pinned build confirms zero b-tree symbols.
- A dress-rehearsal run of this branch's exact `build-static.sh` + `boot-smoke.sh` sequence on a runner is linked in the first comment.


monty-sei:
Verification runs for this branch:

* **This PR's own CI**: the new paths gate routed this PR through the full static job — build with pinned inputs (checksums verified), zero b-tree symbols asserted, and the 8-boot gauntlet, all green: [Linux AMD64 (static)](https://github.com/sei-protocol/sei-chain/actions/runs/29218497592/job/86718948195). The built binary is downloadable from that run's artifacts (`seid-linux-amd64-static`, 14 day retention) if you want to boot it yourself.
* **Dress rehearsal** of the exact `build-static.sh` + `boot-smoke.sh` sequence on a clean runner, same result (private harness repo `monty-sei/seid-binary-smoke`, run 29217827294; happy to add access for anyone who wants the full investigation trail, including the gdb captures of the crash and the 18/18 fix proof).

alexander-sei (approved):

PR #3750, opened 2026-07-13
Use DBImpl for giga mock balance validation

codchen:
## Summary

Routes giga executor balance validation reads through the giga EVM `DBImpl` so `mock_balances` builds follow the same `ensureMinimumBalance` behavior as V2's `StateTransition.BuyGas`.

## Root Cause

`validateGigaEVMTx` was checking balances with `GigaEvmKeeper.GetBalance`, bypassing `DBImpl`. In V2, `BuyGas` reads through `DBImpl.GetBalance`, so mock-balance builds can apply the 100 SEI minimum top-off before the affordability comparison. Giga's manual validation precheck missed that StateDB boundary.

## Changes

- Create the giga `DBImpl` before validation and pass it into `validateGigaEVMTx`.
- Use DB-backed `GetBalance` during validation, including the unassociated-address balance case.
- Add a `mock_balances` regression test covering an unfunded giga transfer within the minimum top-off amount.

## Validation

- `go test ./giga/deps/xevm/state -run TestSubBalance -count=1`
- `go test ./giga/tests -run TestGigaValidation_InsufficientBalance -count=1`
- `go test -tags mock_balances ./giga/tests -run TestGigaMockBalancesValidationUsesDBImpl -count=1`

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -2061,13 +2061,17 @@ func (app *App) executeEVMTxWithGigaExecutor(ctx sdk.Context, msg *evmtypes.MsgE
 		}, nil
 	}
 
-	_, isAssociated := app.GigaEvmKeeper.GetEVMAddress(ctx, seiAddr)
-
-	// Run validation checks (fee/nonce/balance - stateless checks done earlier)
-	validation := app.validateGigaEVMTx(ctx, ethTx, sender, seiAddr, isAssociated)
-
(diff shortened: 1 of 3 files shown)


pompon0 (approved):
Are the seidroid findings not relevant?

codchen:
> Are the seidroid findings not relevant?

not really since mock balance is only built with non-noop with load test

PR #3751, opened 2026-07-13
feat(scripts): add SC read-path RPC probe (consistency check + load)

blindchaser:
## Summary
- Adds `scripts/rpc-sc-read-probe.sh`, a dev/testing tool for the SC (live state-commitment) read path.
- Drives EVM JSON-RPC reads with the `latest` block tag so requests hit the live SC store (memIAVL / FlatKV), not the historical state store.
- Three modes: `check` (cross-node, height-synchronized response consistency), `monitor` (looped check), and `load` (per-target latency/throughput with percentiles).
- `check` is **latency-independent**: each sample is one batch of `latest` reads bracketed by `eth_blockNumber` sentinels, so reads are only compared when the node served them at a single height — works even over high-RTT links. It emits a verdict with distinct exit codes: `CONSISTENT`=0, `MISMATCH`=1, `INCONCLUSIVE`=2 (nothing comparable → not a pass).
- Runner-agnostic: runs locally (python3) or in an ephemeral in-cluster pod; supports `http(s)://`, bare `host:port`, and `k8s://svc|pod` targets (auto port-forward for local); optional JSONL output.

## Motivation
Used to validate SC read-path correctness across backends (e.g. base vs migrated/FlatKV nodes) and to benchmark read latency/throughput during the FlatKV migration.

## Test plan
- [x] `RUNNER=local MODE=check TARGETS='a=...,b=...'` reports a verdict + matched/mismatched across nodes (verified local IP↔DNS: CONSISTENT; in-cluster arctic-1: 130 matched)
- [x] `MODE=load LOAD_SC_ONLY=1 TARGETS=...` prints per-method p50/p95/p99 (verified ip/DNS/k8s; RPS cap + JSONL output)
- [x] `MODE=monitor` loops and reports a non-`ok` round on any `MISMATCH` or `INCONCLUSIVE` check
- [x] `RUNNER=k8s` (in-cluster pod) and `k8s://` targets (in-cluster DNS + local port-forward) resolve and run
- [x] `--help` prints usage

### DIFF
--- a/scripts/rpc-sc-read-probe.sh
+++ b/scripts/rpc-sc-read-probe.sh
@@ -322,11 +322,14 @@ resolve_spec() {
   esac
 }
 
-# derive_name SPEC -> a sanitized short name from the host portion.
+# derive_name SPEC -> a sanitized short name from the host[:port] portion. The
+# port is kept (rendered as a '-') so two targets on the same host but different
+# ports (e.g. 10.0.0.1:8545 and 10.0.0.1:9545) derive distinct names instead of
+# colliding -- a collision would silently drop one target in check mode.
 derive_name() {
   local s=$1
   s=${s#*://}; s=${s#svc/}; s=${s#pod/}
-  s=${s%%/*}; s=${s%%:*}
+  s=${s%%/*}
   printf '%s' "$s" | tr -c 'A-Za-z0-9_.-' '-' | sed 's/-*$//'
 }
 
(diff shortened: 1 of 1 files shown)


cody-littley (approved):
LGTM, but I largely lack context to grok this sort of testing script, so we'll need to rely on the LLM to do the deep review work. I'm ok with this merging once the LLM gives the green light.

yzang2019 (approved):

PR #3755, opened 2026-07-14
feat(ratelimiter): add MethodParser for pre-decode JSON-RPC method extraction (PLT-800)

amir-deris:
## Summary

Implements `MethodParser` (PLT-800), for the RPC Endpoint Protection work. 

`MethodParser` extracts the JSON-RPC `"method"` field from a request body via a **streaming partial read**, *before* the full body is decoded into a Go request struct. This lets per-method rate-limit overrides and, later, weighted admission make a decision based on which method is being called without first paying the memory cost of materialising the whole request.

## What it does

- `NewMethodParser(maxProbeBytes)` → `Parse(io.Reader) (methods []string, batch bool, err error)`.
- Uses a `json.Decoder` and **skips every non-`method` field via a depth-tracked scan** — nested objects/arrays in `params` are read off the stream but never decoded into Go structures. Each request object is read to its closing `}` so a duplicate `"method"` key can be detected and rejected (rather than returning at the first match, which could disagree with the last-wins value `encoding/json` would decode downstream — letting a body report one method while dispatching another).
- Matches `"method"` case-insensitively, the same way `encoding/json` struct-tag decoding does downstream, so a body cannot report one method to the rate limiter while dispatching another.
- Handles both JSON-RPC framings that need it: **EVM JSON-RPC** (`:8545`/`:8546`) and **CometBFT RPC** (`:26657`), which share the JSON-RPC 2.0 shape and the `"method"` field.
- Supports **batches** (top-level arrays), returning one method per element in order with `batch=true`.
- `maxProbeBytes` (default 1 MiB) bounds the work an adversarial body (a giant `params`) can force; exceeding it returns `ErrProbeLimit`, distinguished from a genuinely truncated/malformed body. Because each object is read to its end, a legitimate body larger than the probe limit also yields `ErrProbeLimit` — the wiring middleware (follow-up PRs) must treat that as "method undetermined, fall back to full decode" rather than a rejection.
- Takes a plain `io.Reader` — the body tee/rewind decision stays with the wiring middleware.

## Scope

- **No request-path wiring**. Consumed by EVM (1b) and CometBFT (1c) HTTP/WS middleware in follow-up PRs.
- **gRPC (1d) is out of scope** — gRPC already exposes the full method name in the interceptor, so no partial read is needed.

## Tests

`ratelimiter/method_parser_test.go` — EVM/CometBFT single requests, both field orders, nested/empty/large `params`, escaped and case-insensitive method keys, batches (multi/single/method-after-params), duplicate-method rejection (single + batch + mixed-case), trailing-data rejection, every error sentinel (`ErrNoMethod`, `ErrMethodNotString`, `ErrNotObject`, `ErrEmptyBatch`, `ErrDuplicateMethod`, `ErrTrailingData`, `ErrProbeLimit`), truncated-vs-probe-limit disambiguation (including the exact-probe-limit boundary), and reading from a streaming `io.Reader`.


ratelimiter/method_parser.go
```diff
@@ -0,0 +1,231 @@
...
+	return &MethodParser{maxProbeBytes: maxProbeBytes}
+}
+
+// Parse reads a JSON-RPC request body from r and returns the method name(s) it
+// carries. A single request object yields a one-element slice with batch=false;
+// a top-level array yields one method per element in order with batch=true.
+func (p *MethodParser) Parse(r io.Reader) (methods []string, batch bool, err error) {
+	// N is maxProbeBytes+1 so that lr.N reaching 0 unambiguously means the body
+	// exceeded the budget.
+	lr := &io.LimitedReader{R: r, N: p.maxProbeBytes + 1}
```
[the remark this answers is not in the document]
bdchatham: This is outdated I think since you subtract 1 in the constructor

bdchatham:
Ran a cross-review of this at head `edde4ea` across four lenses (security, systems, networking/RPC semantics, and Go idiom), reviewing independently before comparing notes. Summary up top, full ledger in the toggle.

The parser is in good shape. The security pass could not find a token-level report-one-method-dispatch-another differential. The case-fold match against `encoding/json` is correct, and rejecting duplicate method keys is fail-closed. The probe-budget invariant holds against decoder read-ahead, the code is reentrant, and `skipValue` is iterative so there's no deep-nesting blowup. Most of the earlier bot findings (duplicate-first, missing `]` validation, the probe `-1`) are already fixed at this head.

Three things I'd like to close, filed as inline comments:
1. The fail-closed contract that makes this safe lives only in the PR description. Worth writing into the package doc so 1b/1c build on it.
2. `skipValue` boxes every scalar via `Token()`, which the comment says it avoids, measured around 600k allocs per within-budget request. A `RawMessage` skip fixes it.
3. CometBFT serves methods in the URL path for GET requests, which the body parser can't see. Needs a scope decision before the :26657 wiring.

None of these block the parser as a standalone piece, they're mostly about the contract the wiring will inherit.

<details>
<summary>Full review ledger (boundaries, findings, what was verified sound)</summary>

**Boundaries checked**

| Interface / Boundary | Status | Evidence |
|---|---|---|
| Method extraction vs downstream `encoding/json` (duplicate key) | COMPATIBLE (fail-closed) | `:161-164` rejects dup; `evmrpc/sei_legacy_http.go:91` decodes `Method string` last-wins. Reject can only over-reject, never report-A-run-B. |
| Case-insensitive `"method"` match | COMPATIBLE | `strings.EqualFold` (`:161`); all 6 letters of "method" are ASCII-fold-only. |
| `ErrProbeLimit` vs wiring "fall back to full decode" | MISSING | Safety depends on consumer error mapping; lives only in PR prose. → F1 |
| CometBFT :26657 URI/GET (method in path) | MISSING | `Parse` reads body only (`:64`). → F3 |
| Trailing-data rejection | MISMATCH (fail-closed-safe) | `expectEOF` (`:100-108`) stricter than go-eth HTTP, matches CometBFT. |
| Batch atomicity vs per-element dispatch | MISMATCH (fail-closed-safe) | `readBatchMethods` aborts whole batch (`:116-127`); dispatchers run valid elements. |
| Shared probe budget across batch | MISMATCH (behavioral) | One `LimitedReader` (`:61`); a padded early element can push a later method past budget. |

**Findings requiring action:** F1 (doc the fail-closed contract), F2 (`skipValue` boxing / false comment, ~600k allocs, RawMessage fix), F3 (CometBFT method-in-path scope).

**Deferrable (fail-closed-safe or wiring-time):** `ErrProbeLimit` not returned when an over-budget body truncates into a `SyntaxError`; trailing-whitespace at exactly budget+1 flips a found method to a misleading `ErrProbeLimit`; tee must be independent of `Parse` on fallback; no batch item cap; test gaps (chunked reader, alloc-bound guard).

**Verified sound:** no report-A/run-B differential (exhaustive fold analysis); probe-budget invariant robust against read-ahead; concurrent-safe; iterative `skipValue`; null/number/wrong-type method, empty batch, top-level scalar all classify safely; reads as native Go (idiom pass ratified, advisory nits only).

Reviewed independently, blinded, at `edde4ea`.
</details>

ratelimiter/method_parser.go
```diff
@@ -0,0 +1,236 @@
...
+		//preventing overflow error
+		maxProbeBytes = math.MaxInt64 - 1
+	}
+	return &MethodParser{maxProbeBytes: maxProbeBytes}
+}
+
+// Parse reads a JSON-RPC request body from r and returns the method name(s) it
+// carries. A single request object yields a one-element slice with batch=false;
+// a top-level array yields one method per element in order with batch=true.
+func (p *MethodParser) Parse(r io.Reader) (methods []string, batch bool, err error) {
```
bdchatham, line 63: The safety of this gate rests on how the caller handles these errors, and right now that contract lives only in the PR description. Every error other than `ErrProbeLimit` has to mean reject or fall back to a full decode, and `ErrProbeLimit` has to mean fall back rather than reject. If a caller ever maps a parse error to admit-with-default, the duplicate-key and trailing-data rejections stop being fail-closed, and a body that reports a cheap method while decoding to an expensive one gets through, which is the case this gate exists to catch.

There's nothing to fix at runtime in 1a since there's no wiring yet. I'd like the fail-closed contract written into the package doc so 1b and 1c build on it explicitly, something like every error other than `ErrProbeLimit` means reject and `ErrProbeLimit` means fall back to full decode. Worth carrying as an acceptance criterion into those PRs too.

ratelimiter/method_parser.go
```diff
@@ -0,0 +1,236 @@
...
+	if _, err := dec.Token(); err != nil {
+		return "", err
+	}
+	if !found {
+		return "", ErrNoMethod
+	}
+	return method, nil
+}
+
+// skipValue consumes exactly one complete JSON value from dec — the value whose
```
bdchatham, line 195: The comment here says the skipped contents are read off the stream but never decoded into Go structures, and that isn't quite what happens. `dec.Token()` decodes every token, allocating a string for each string and boxing each scalar into an `interface{}`. Since we drain the whole object to catch a duplicate method, every request pays that on the rate-limiter path. A within-budget scalar-heavy `params` array measured around 600k allocations and 11 MB of garbage per request, which works against the streaming-partial-read goal.

The non-method branch can skip with a `RawMessage` decode instead, which still validates and still respects the `LimitedReader` budget but drops the per-scalar boxing (roughly 600k allocs down to 23 in the same benchmark):

```go
var raw json.RawMessage
if err := dec.Decode(&raw); err != nil {
    return err
}
```

At a minimum I'd correct the comment, but the `RawMessage` swap looks cheap and worth doing here.

ratelimiter/method_parser.go
```diff
@@ -0,0 +1,236 @@
...
+// Parse reads a JSON-RPC request body from r and returns the method name(s) it
+// carries. A single request object yields a one-element slice with batch=false;
+// a top-level array yields one method per element in order with batch=true.
+func (p *MethodParser) Parse(r io.Reader) (methods []string, batch bool, err error) {
+	// N is maxProbeBytes+1 so that lr.N reaching 0 unambiguously means the body
+	// exceeded the budget.
+	lr := &io.LimitedReader{R: r, N: p.maxProbeBytes + 1}
+	dec := json.NewDecoder(lr)
+
+	tok, err := dec.Token()
```
bdchatham, line 69: One coverage gap to flag for the :26657 path. CometBFT serves every RPC as a URI route as well, so `GET /broadcast_tx_sync?tx=...` carries the method in the URL path with no method field in the body. This parser reads the body first, so those requests would come back as `ErrNoMethod` and the real method never gets classified.

Not necessarily something to solve in 1a. I'd like us to decide the scope explicitly, either this parser is only ever fed body-form requests and we say so in the doc, or 1c pulls the method from the path before falling back to the body parser. Either is fine, I just want it written down before the CometBFT wiring lands.

yzang2019 (approved):

ratelimiter/method_parser.go
amir-deris: Thanks for great feedback. Updated design doc to mention this point for 1c.

ratelimiter/method_parser.go
amir-deris: Thanks for great feedback. For now, updated the comment on the method to reflect this.

PR #3758, opened 2026-07-15
feature(sei-db) LTHash refactory for adding per module hash and stats

yzang2019:
## Summary

Refactors FlatKV's lattice-hash (LtHash) subsystem and extends per-DB metadata to track **per-module** hashes and storage stats. Also renames the `Legacy*` catch-all DB to `Misc*` for clarity. FlatKV is not yet in production, so no on-disk migration/backward-compat is provided (fresh stores only).

### 1. `LegacyDB` → `MiscDB` rename

The catch-all DB (everything that isn't account/code/storage — i.e. all non-EVM cosmos modules, plus some EVM misc keys) is renamed throughout:

- On-disk directory `legacy/` → `misc/`
- `vtype.LegacyData` → `vtype.MiscData`, `keys.EVMKeyLegacy` → `keys.EVMKeyMisc`
- All identifiers, comments, tooling, and integration-test scripts updated.

### 2. Per-module LtHash metadata

Each data DB now decomposes its per-DB root LtHash into **additive per-module hashes**, persisted under `_meta/x:<module>/hash` in that DB's LocalMeta:

- `account`/`code`/`storage` DBs only ever carry the `evm` module.
- `misc` DB may carry several (`evm` plus cosmos modules such as `gov`, `bank`).
- The per-DB root is the homomorphic sum of its module hashes; the global store hash (and consensus AppHash) is unchanged — per-module hashes are consensus-irrelevant bookkeeping.
- Loaded on startup, maintained on live commit, and populated by the state-sync importer + persisted in `FinalizeImport`.

### 3. LtHash computation refactor + threading model

Extracted the whole per-block LtHash pipeline into a new `lthash.HashCalculator`:

- Owns a single **fixed worker pool** (`threading.FixedPool`) shared by both the live-commit path and the state-sync importer.
- Work is split into independent, self-terminating per-chunk tasks, so it's safe to call concurrently from multiple goroutines (the importer runs one per DB) with no deadlock/oversubscription risk.
- Reads previous values through an injected `OldValueReader` interface and extracts module names via an injected `ModuleFunc`, keeping `lthash` decoupled from `pebbledb`/`ktype`.
- `ApplyChangeSets` is now a thin orchestrator (classify → read-old → apply EVM value semantics → compute); EVM/cosmos domain semantics stay in `flatkv`.

### 4. Per-module storage stats

Alongside each module's hash, we now accumulate **key count** and **total bytes** (physical key + serialized value = logical footprint), persisted under `_meta/x:<module>/stats`:

- Add: +1 key, +key+value bytes. Update: count unchanged, byte delta only. Delete: −1 key, −key+old-value bytes. Delete-of-absent: no-op.
- Counting predicates are gated on exactly the same conditions the hash uses to mix/unmix, so stats always track the set the hash represents.
- Crash-recoverable (loaded on init), maintained on live commit, and produced identically by the importer.
- Persisted per-(DB, module) only; per-DB/global totals are derivable on demand.

### 5. Config

- Added `LtHashThreadsPerCore` (float) to size the dedicated LtHash pool: `workers = int(LtHashThreadsPerCore * NumCPU)`, clamped to ≥ 1. Default `1.0` (LtHash is CPU-bound). Mirrors the existing reader/misc pool config convention.

### 6. Tests

- `lthash`: `ModuleStats` marshal round-trip / bad-length, `foldChunk` add/update/delete accounting, pooled aggregation path.
- `ktype`: `_meta/x:<module>/hash` and `/stats` key round-trips, mutual non-aliasing, malformed-key rejection.
- `flatkv`: per-module hash & stats "incremental equals full scan" invariants, add→update→delete transitions, persistence-after-reopen, and state-sync-import-survives-restart.


sei-db/state_db/sc/flatkv/snapshot.go
```diff
@@ -179,7 +179,7 @@ func updateCurrentSymlink(root, snapshotDir string) error {
 }
 
 // snapshotDBDirs lists the DB subdirectory names included in a snapshot.
-var snapshotDBDirs = []string{accountDBDir, codeDBDir, storageDBDir, legacyDBDir, metadataDir}
+var snapshotDBDirs = []string{accountDBDir, codeDBDir, storageDBDir, miscDBDir, metadataDir}
```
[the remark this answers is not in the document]
yzang2019: We don't need any migration, no existing FlatKV is created at all

sei-db/state_db/sc/flatkv/lthash/hash_calculator.go
```diff
@@ -0,0 +1,443 @@
...
+			cur = New()
+			modBucket[key.Module] = cur
+		}
+		cur.MixIn(delta.Hash)
+		statBucket[key.Module] = statBucket[key.Module].Add(ModuleStats{KeyCount: delta.KeyCount, Bytes: delta.Bytes})
+		touched[key.Dir] = struct{}{}
+	}
+	for dir := range touched {
+		newPerDB[dir] = SumModuleHashes(newPerModule[dir])
+	}
```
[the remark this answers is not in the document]
yzang2019: We don't need any backfill, no existing FlatKV is created at all

sei-db/tools/cmd/seidb/operations/dump_flatkv_test.go
```diff
@@ -61,7 +61,7 @@ func TestDumpFlatKVFromStoreAllBuckets(t *testing.T) {
 		"account": {lines: 2}, // 2 nonces -> 2 account rows
 		"code":    {lines: 1}, // 1 code
 		"storage": {lines: 3}, // 3 storage slots
-		"legacy":  {lines: 1}, // 1 bank row
+		"misc":  {lines: 1}, // 1 bank row
```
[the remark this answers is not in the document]
yzang2019: This is fixed

sei-db/state_db/sc/flatkv/lthash/hash_calculator.go
```diff
@@ -0,0 +1,446 @@
...
+// module) key-count / byte totals accumulated alongside the hash.
+type Result struct {
+	PerDB          map[string]*LtHash
+	PerModule      map[string]map[string]*LtHash
+	PerModuleStats map[string]map[string]ModuleStats
+	Global         *LtHash
+}
+
+// HashCalculator owns the CPU-bound lattice-hash worker pool and encapsulates
+// the per-block lattice-hash pipeline:
```
cody-littley: We might want to share the hashing thread pool between different HashCalculator instances.

The main reason why we'd want to use a single pool is so that we could limit maximum CPU utilization for hashing, while at the same time not requiring us to assume certain ratios of workload between different DBs. If we've got more hashing threads than cores we can actually end up degrading hashing speed due to competition over CPU time and lots of extra thread context switching. And it may be the case that we want to set the number of hashing threads to be fewer than the total core count in order to avoid starving out mission critical threads like execution and consensus.

At Hedera we had a similar problem where too many hashing threads were producing noticeable impact on other latency sensitive code that was being temporarily starved of CPU resources every time block hashing started. When we decreased the number of hashing threads, we actually observed a net increase in system throughput.

sei-db/state_db/sc/flatkv/lthash/hash_calculator.go
```diff
@@ -0,0 +1,446 @@
...
+	// irrelevant.
+	merged := make(map[ModuleKey]*ModuleHashInfo)
+	for i, task := range tasks {
+		if acc := merged[task.key]; acc != nil {
+			mergeDelta(acc, results[i])
+		} else {
+			merged[task.key] = results[i]
+		}
+	}
+	return merged
```
cody-littley: Instead of waiting on a wait group, what if instead each job returned its answer via a chan? Then, the merging loop could get started immediately as soon as the first result is ready, as opposed to waiting for all results to be ready. Each iteration of the loop, we'd do `task <- resultChan` and then merge it in.

sei-db/state_db/sc/flatkv/lthash/hash_calculator.go
```diff
@@ -0,0 +1,446 @@
...
+			results[idx] = foldChunk(tasks[idx].pairs)
+		})
+	}
+	wg.Wait()
+
+	// Merge per-chunk results. Multiple chunks of the same (db, module) bucket
+	// sum into one delta; MixIn/addition are commutative so chunk order is
+	// irrelevant.
+	merged := make(map[ModuleKey]*ModuleHashInfo)
+	for i, task := range tasks {
```
cody-littley: If merging ever becomes a hotspot, in theory we could farm out merge work to the hashing pool, where each job merges two pairs of results.

This is very much a premature optimization, so I don't think we should make it in this PR. I expect merging to generally be a fast operation. Bit it might be worth a comment mentioning it's something we can do if merging becomes a hotspot.

sei-db/state_db/sc/flatkv/lthash/hash_calculator.go
cody-littley: Looking at flatKV, I think we may actually be instantiating just one `HashCalculator` object. If that's the case, then disregard the above comment. It's only important to inject a shared pool if there are multiple hashers.

sei-db/state_db/sc/flatkv/store_iteration.go
```diff
@@ -307,10 +307,10 @@ func buildLane[T vtype.VType](
 	return transformedIterator, nil
 }
 
-/* Data flow: buildLegacyDBLane
+/* Data flow: buildMiscDBLane
 
   ┌────────────────────────┐       ┌───────────────────┐
-  │ legacyWrites (pending) │       │ legacyDB (pebble) │
+  │ miscWrites (pending) │       │ miscDB (pebble) │
```
cody-littley: nit, diagram out of alignment

sei-db/state_db/sc/flatkv/store_iteration.go
```diff
@@ -83,7 +83,7 @@ func (s *CommitStore) Iterator(store string, start []byte, end []byte, ascending
 
 buildCodeLane ──────────────┐
 buildStorageLane ───────────┤
-buildLegacyDBLane (evm/) ───┼──► merge iterator ──► memiavl keys + values
+buildMiscDBLane (evm/) ───┼──► merge iterator ──► memiavl keys + values
```
cody-littley: diagram out of alignment

cody-littley (commented):
One big missing element is HashLogger integration. But I think it would be ok if we did the HashLogger as a follow up PR since it's not going to be a 5 line fix. If you'd prefer to do it later, we should probably file a ticket.

cody-littley (approved):
Giving my "Approve" review now since none of my comments are blocking. Might want to give @blindchaser an opportunity to review if he wants, since I think he was the original author of the hashing logic.

sei-db/state_db/sc/flatkv/store_meta.go
```diff
@@ -68,9 +150,74 @@ func writeLocalMetaToBatch(batch types.Batch, version int64, ltHash *lthash.LtHa
...
+// written store never has one without the other; an empty module map alongside
+// a non-zero root can only mean the store predates per-module hashing.
+//
+// Migration of such stores is intentionally unsupported (fresh stores only). If
+// we proceeded, the first write would route through HashCalculator.Compute,
+// which rebuilds each touched DB's root as SumModuleHashes(perModule[dir]) — a
+// map holding only that block's delta — silently discarding the pre-existing
+// keys' contribution to the per-DB root and thus the global store hash /
+// AppHash. Fail loudly here instead of corrupting state.
+func validatePerModuleMetadata(dbDir string, meta *ktype.LocalMeta) error {
```
blindchaser, line 186: +1 something like: 

```
if len(meta.ModuleLtHashes) == 0 {
    if meta.LtHash.IsZero() { return nil }
    return fmt.Errorf("...predates per-module hashing...")
}
sum := lthash.SumModuleHashes(meta.ModuleLtHashes)
if !meta.LtHash.Equal(sum) {
    return fmt.Errorf("flatkv: %s per-module hashes do not sum to per-DB root (root=%x sum=%x)", dbDir, meta.LtHash.Checksum(), sum.Checksum())
}
```

sei-db/state_db/sc/flatkv/lthash/hash_calculator.go
yzang2019: Make sense, will move the workerpool creation out to upper layer like FlatKV constructor, and pass the workerpool to the HashCalculator constructor

sei-db/state_db/sc/flatkv/store_iteration.go
yzang2019: Fixed

sei-db/state_db/sc/flatkv/store_iteration.go
yzang2019: Fixed

sei-db/state_db/sc/flatkv/lthash/hash_calculator.go
yzang2019: Good call, let me take a stab at this next

sei-db/state_db/sc/flatkv/lthash/hash_calculator.go
yzang2019: Agreed, adding a comment

sei-db/state_db/sc/flatkv/importer.go
```diff
@@ -110,9 +124,29 @@ func (w *dbWorker) flush() (err error) {
 			metric.WithAttributes(dbAttr(w.dir), successAttr(err)))
 	}()
 
-	// TODO:In theory, we could offload lattice hash calculation to a work pool and get parallelism between DB operations and hash calculations. Cryptosim performance makes me think we could probably get a 2-3x speedup from this, assuming receiving data from the network isn't the bottleneck.
-	newHash, _ := lthash.ComputeLtHash(w.ltHash, w.ltPairs)
-	w.ltHash = newHash
+	// Per-module hashes are the primitive: distribute this batch's pairs across
```
blindchaser, line 127: Not triggered by any current feeder, both the flatkv→flatkv export (RawGlobalIterator over committed Pebble rows) and the memiavl importer (translator merges accounts, injective per kind) emit each physical key once. 

But the importer's LtHash accumulation is non-idempotent by construction (assumes empty target, only MixIn, never MixOut), so a duplicate key would double-mix the hash and double-count stats while Pebble keeps one row, and FinalizeImport persists it without VerifyLtHash.  (not introduced by this PR, it's the existing behavior)

Suggest a cheap fail-loud dedup check in dbWorker.run() to guard the invariant against future feeders / malformed snapshots. (Pre-existing on main; the refactor widens it to per-module hashes + stats.)

sei-db/state_db/sc/flatkv/verify.go
```diff
@@ -38,45 +42,158 @@ func verifyLtHashInternal(cs *CommitStore) error {
...
+	return hashes, stats, nil
+}
+
+// verifyDBModuleMetadata checks the maintained per-module hashes and stats for
+// one DB against a fresh scan, verifies they homomorphically sum to the
+// maintained per-DB root, and returns that (scan-derived) per-DB root for the
+// global accumulation. It fails on: a scanned module missing/mismatched in the
+// maintained map, a maintained module absent from disk yet not zeroed, a
+// stats mismatch, or the per-module sum not equaling the per-DB root.
+func (cs *CommitStore) verifyDBModuleMetadata(
```
blindchaser, line 134: verification iterates scanned modules and maintained hash modules, but never iterates workingStats on its own.

is it possible that an orphan non-zero stats entry (a module with no on-disk keys and no module hash) passes despite the doc claiming all per-module stats are validated?

sei-db/state_db/sc/flatkv/lthash/hash_calculator.go
yzang2019: Fixed with a merging loop

sei-db/state_db/sc/flatkv/store_meta.go
yzang2019: Good suggestion!

sei-db/state_db/sc/flatkv/verify.go
yzang2019: Good catch!

sei-db/state_db/sc/flatkv/importer.go
yzang2019: Under what scenario would we export duplicate keys to the snapshot? If we add a dedup logic, the concern is that we have to fit every key into memory during the import process, which could fill up memory overhead and cause OOM

yzang2019: If we hit a bad snapshot today which does contains duplicate keys, then the calculated lattice hash should detect the difference and error out the import at the end right?

blindchaser: true

sei-db/state_db/sc/flatkv/store_apply.go
blindchaser, line 217: The empty module name isn't newly accepted here, main's classifyAndPrefix has no guard either. But on main it was harmless (just a /key row folded into the single per-DB hash). 

With per-module metadata it becomes fatal: it writes _meta/x:/hash, which ParseModuleLtHashKey rejects on reload, so the new sum validation then fails and the store won't reopen. So the refactor turns a latent no-op into a store-won't-open failure. Cheap guard: reject nil changeset / empty cs.Name at the top of the loop. (Only reachable via malformed input — module names are never empty in normal operation.)

worth to add:

```
if cs == nil {
        return nil, fmt.Errorf("flatkv: nil changeset")
    }
    if cs.Name == "" {
        return nil, fmt.Errorf("flatkv: changeset with empty module name")
    }

```

blindchaser (approved):

sei-db/state_db/sc/flatkv/store_apply.go
yzang2019: So through the normal rootmulti → SC path, this is indeed unreachable in practice — confirming the reviewer's "module names are never empty in normal operation" note. But FlatKV could also receive changesets from other feeders that don't necessarily go through that constructor (WAL replay of a possibly-corrupted changelog entry, import/migration tooling, any future direct ApplyChangeSets caller). 

Given the failure mode is "store can never be opened again" rather than an immediate loud error at the mistake's source, a cheap guard at the boundary is worth it as defense-in-depth.

PR #3762, opened 2026-07-15
ci: prevent runner disk exhaustion in integration + lint jobs (PLT-807)

amir-deris:
Recently CI jobs have been failing the merge queue more frequently. The failures are caused by GitHub-hosted runner running out of disk on `/` and crashing.

Every failing job (EVM RPC Parity, EVM Module, Autobahn GIGA, and
`Go / Lint`'s `Vet Go files`) shows the same runner-level annotation:

```
System.IO.IOException: No space left on device :
  '/home/runner/actions-runner/cached/2.335.1/_diag/Worker_...log'
```

The runner's own worker process can't even write its diagnostic log. When it
dies mid-step, the step never records completion, log streaming stops, and the
job hangs until GitHub's backend force-fails it — which looks like a timeout and
fails the merge group. Different, unrelated jobs failing the same way = an
environmental (disk) problem, not a test bug.

**Why it fills:** these are GitHub-hosted runners (fresh VM per job), so each job
fills its own `/`. The integration jobs pull two GHCR images + keep the seid
tarball + run a 5-container Sei cluster whose db/logs grow during the run, all
alongside Docker's overlay dir and the Go build cache. The lint job's
`go vet ./...` build cache is enough to tip `ubuntu-latest`'s ~14 GB `/` on its
own. The margin was thin, so recent growth pushed a fraction of runs over the
edge (hence intermittent, and "retry works").

**The fix** — reclaim disk and move the biggest consumers off `/` before the
heavy work, via three small, reusable, self-guarding scripts:

- `.github/scripts/ci-free-disk.sh` — remove preinstalled toolchains Sei CI
  never uses (dotnet, android, ghc, ghcup, CodeQL, boost, swift); frees
  ~20-30 GB on `/`. Prints `df` before/after.
- `.github/scripts/ci-relocate-docker-data-root.sh` — move Docker's `data-root`
  to the `/mnt` ephemeral disk so image + container layers accumulate off `/`.
  Merges into any existing `daemon.json`.
- `.github/scripts/ci-relocate-go-cache.sh` — point `GOCACHE`/`GOMODCACHE`/
  `GOTMPDIR` at `/mnt`. `GOPATH` is deliberately left untouched: the Makefile
  mounts a hardcoded `$HOME/go/pkg/mod` into the build container but pre-creates
  it via `go env GOPATH`, so overriding `GOPATH` would break `prepare-cluster`'s
  in-container build with a root-owned mount.

Wiring:
- `integration-test.yml` — the three prep steps run right after checkout in
  `prepare-cluster`, `integration-tests`, and `autobahn-integration-tests`
  (before any image pull/build).
- Lint job — added `.github/actions/go-lint-setup/action.yml`, a composite
  action automatically picked up by the shared `sei-protocol/uci` go-lint
  workflow's repo-specific setup hook (previously skipped because the repo had
  none), so the lint job gets the disk reclaim without changing the `uci` repo.
  The hook **reclaims disk only** and deliberately does not relocate the Go
  caches: it runs *after* `uci`'s `setup-go`, which has already restored the Go
  cache to `/`, so relocating to `/mnt` would strand that copy and force
  `go vet` to run cold. Reclaim alone leaves enough headroom on `/` while
  keeping `setup-go`'s cache warm.
- In-run cleanup: drop `integration-build.tar.gz` after extracting `build/seid`,
  and `docker image rm` the redundant run-id-tagged images once retagged to
  `sei-chain/*` (verified `ensure-integration-ci-images` only needs the
  `sei-chain/*` tags).

All relocation scripts **no-op cleanly** when `/mnt` is unavailable, so the
worst case degrades to the ~25 GB bloat reclaim on `/` — strictly better than
today.

**Runner-topology caveat (important for future readers):** on `ubuntu-large` —
the class the integration jobs actually run on — `/mnt` is *not* a separate
ephemeral volume. GitHub larger runners use a single Premium-SSD OS disk, so
`/mnt` is just a directory on the same `/dev/root` filesystem. This PR's own
`df` output confirms it (`/` and `/mnt` both resolve to `/dev/root`, 145 GB).
On that class the load-bearing win is entirely `ci-free-disk.sh`'s ~24 GB
reclaim (58G→34G used) on top of the large 145 GB root; the Docker/Go-cache
relocations move data *within the same disk* and add no capacity. The
relocations only become load-bearing on runner classes that expose a genuine
separate `/mnt` (e.g. standard `ubuntu-latest`). To avoid a pointless Docker
daemon restart and misleading logs, both relocation scripts now compare the
device of `/mnt` and `/` (`stat -c '%d'`) and skip relocation when they are the
same filesystem — so on `ubuntu-large` the relocation is an explicit no-op, and
on a real split-disk runner it still relocates.

Not included (deliberately deferred): relocating `build/generated` (fights the
Makefile's `rm -rf` and risks test mechanics), lowering seid log verbosity /
`INVARIANT_CHECK_INTERVAL` (would trade away coverage + diagnostics), and
bumping the `ubuntu-large` runner-group disk size (org runner setting, not a
repo change).

- `python3 -c 'yaml.safe_load(...)'` on `integration-test.yml` and
  `action.yml` — both parse.
- `bash -n` on all three scripts — clean; embedded `python3` heredoc compiles.
- Verified `ensure-integration-ci-images` inspects only `sei-chain/localnode`
  and `sei-chain/rpcnode`, so removing the run-id tags is safe.
- Real validation is this PR's own CI run: the **Reclaim runner disk** step's
  `df` output on `ubuntu-large` shows `/` at 145 GB dropping from 58G→34G used
  (~24 GB reclaimed), with `/mnt` resolving to the same `/dev/root` device — see
  the runner-topology caveat above. The relocation steps therefore log the
  same-device no-op on this class; the reclaim is what keeps `/` from filling.


yzang2019 (approved):

bdchatham:
Per this PR's own run on `ubuntu-large`, the `df` output in the **Reclaim runner disk** step shows `/mnt` is a plain directory on the same 145G root filesystem (`/dev/root` listed twice) — so both relocation scripts move data within the same disk, and the realized win comes entirely from `ci-free-disk.sh`'s ~24GB reclaim (58G→34G used) plus the large root. Not blocking since the relocations degrade safely, but consider noting this in the PR body (and perhaps a follow-up capacity-aware guard, e.g. only relocate when `/mnt` free > `/` free) so future readers don't assume the docker data-root move + daemon restart is load-bearing.

bdchatham (approved):
Cross-reviewed across platform, systems, security, and idiom lenses — all boundaries check out: the relocate/reclaim steps run before any docker or go usage in all three jobs, the uci@v0.0.9 setup hook fired on this PR's own lint run, the in-run cleanup is untag-only with no downstream consumers of the removed refs, and the sudo/daemon.json/GITHUB_ENV surfaces take only static inputs under plain pull_request semantics on ephemeral hosted runners. Remaining notes are advisory only (see my comment on the /mnt topology). LGTM.

PR #3764, opened 2026-07-15
test(mempool): pin drain-race recheck behavior

bdchatham:
## Summary

Characterization tests pinning the mempool's "drain-race" behavior, where a same-sender transaction is affordable at admission but unaffordable by execution because an earlier transaction drains the sender's balance first.

## Why

The team observed such a transaction committed as an empty failure (status=0, gasUsed=0) on arctic-1, while atlantic-2 drops it. Two things drive that, and I verified both against the code.

The inclusion itself is a same-block reap. Readiness promotes tx1 (nonce N) and tx2 (nonce N+1) against the same, undrained account balance, and never subtracts what an earlier nonce will spend, so a single reap returns both into one block. tx1 drains the balance, tx2 then fails its fee at delivery. There is no cumulative-balance reservation across a sender's pending transactions, and there never was. This is not a code regression in the v6.6 mempool rewrite, and the empty-failure receipt is version-invariant EVM-antehandler behavior.

The live difference between the two chains is a consensus parameter, not code. arctic-1 carries `recheck_tx=false` (the current default), atlantic-2 carries `true`. With recheck on, atlantic-2 re-evaluates a became-invalid transaction that survives to a later block and drops it; arctic-1 never re-evaluates. Recheck is a cross-block mechanism, so it does not affect the same-block reap that includes tx2 in the first place.

These tests lock the behavior into CI so any future change surfaces in review. They are characterization only. Whether recheck should be enabled, or whether the mempool should reserve balance across pending transactions, is a design decision for the mempool owners.

## What's tested

- `TestTxMempool_DrainRaceSameBlockIncludesInvalidTx` is the real path. Both transactions are ready and reaped together in one block, and the test asserts tx2 is in the reaped set (inclusion), before any post-block balance recompute can demote it.
- `TestTxMempool_DrainRaceRecheckGatesEviction` covers the cross-block case, where a transaction that survives to a later Update is evicted under recheck=true (the app re-CheckTx) or demoted to pending under recheck=false (the balance recompute). This is not the same-block inclusion path.
- `TestTxMempool_DrainRaceRecheckScopeSkipsPendingTx` documents that a still-pending transaction sits outside the recheck loop's `ReadyTxs()` scan, so recheck cannot reach it.

## For reviewers

No production code changed. The shared `evmNonceApp` mock's `CheckTx` gains an insufficient-funds rejection scoped to `CheckTxTypeV2Recheck`, modeling `EVMFeeCheckDecorator.BuyGas` (`x/evm/ante/fee.go`), which runs on recheck today. A comment flags that gating `BuyGas` behind `!IsReCheckTx()` would make the mock pin fiction. Full mempool package is green, including `-race`.

### DIFF
--- a/sei-tendermint/internal/mempool/recheck_drain_test.go
+++ b/sei-tendermint/internal/mempool/recheck_drain_test.go
@@ -22,9 +22,13 @@ import (
 
 // evmNonceApp models a Sei-like EVM antehandler for mempool tests:
 //   - tracks the next-expected nonce per sender (the "mined" nonce frontier)
+//   - tracks the per-sender balance used for affordability/recheck decisions
 //   - returns nonce metadata used by mempool-side pending evaluation
 //
-// Test format: "evm=<sender>=<nonce>=<priority>".
+// Test format is either 4-part or 5-part:
+//   - "evm=<sender>=<nonce>=<priority>"                    (no balance requirement)
+//   - "evm=<sender>=<nonce>=<priority>=<requiredBalance>"  (5th field is the
+//     EVMRequiredBalance, i.e. the balance the tx must be able to afford)
 type evmNonceApp struct {
 	abci.Application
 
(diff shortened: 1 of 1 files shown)


bdchatham:
Closing without merging. The thread with Greg and Yasin concluded this is expected behavior, not a regression.

With nonces N then N+1 from one sender where N drains the balance, both are selected into the block. The old mempool silently drops N+1 once it becomes invalid after N. The new mempool sequences N+1 and commits an empty stub, consuming the nonce and reporting gasUsed=0, status=0. This is a known tradeoff, not net-new: a tx is sequenced before execution finishes so the block can be voted on, and the free on-chain storage for the failed tx is the accepted cost of that design.

Two things settled the call. The old-vs-new difference is insertion-order dependent, so both old-mempool outcomes are reachable and the test pins a repro-sensitive seam rather than a behavior regression. And making the mempool accumulate requiredBalances across pending txs would trade these false positives for false negatives, which risks breaking users who stamp a uniform gas limit, so that is not being pursued now.

The characterization value does not justify carrying a repro-order-sensitive test. If we later decide to bound the free-storage vector, this can be reopened as the behavioral anchor.

PR #3768, opened 2026-07-16
fix(giga): fall back to v2 on execution errors

philipsu522:

This fixes a live RPC-level divergence, not only a defensive edge case. On `main`, the parity test reaches matching consensus fields but fails receipt parity: Giga stamps `TxType=2` and `gasUsed=gasLimit`, while V2 writes an EndBlock synthetic receipt with zero values. Mixed Giga/V2 fleets can therefore return different receipts for the same transaction.

State-transition errors now route through the existing V2 fallback instead of duplicating V2's fee, nonce, receipt, deferred-info, and ABCI result handling. This adds an `execution_failed` abort sentinel and fallback metric reason. Because the Giga StateDB has not been finalized when `ExecuteTransactionFeeCharged` returns an error, its transaction-local cache is discarded before V2 re-runs the transaction. Both sequential and OCC Giga paths use the existing fallback plumbing.

The EIP-7623 floor-data-gas parity test covers deterministic ABCI fields, LastResultsHash, nonce, receipt, sender balance, fee-collector balance, and per-transaction coinbase balance in sequential and OCC modes.

Operational tradeoff: this creates the first deterministic single-transaction trigger that passes CheckTx at roughly 27.5k gas but reverts an entire OCC batch to V2. Operators should alert on `app_giga_fallback_to_v2{reason="execution_failed"}`; per-transaction OCC fallback should be tracked as follow-up work.

- `go test ./giga/tests -run '^TestGiga_FailedExecutionFallsBackToV2$' -count=1`
- `go test ./giga/tests -count=1`
- `go test ./app -run 'TestGiga|TestOCC|TestProcess' -count=1`
- `gofmt -s` and `goimports` on all changed Go files

[Slack Thread](https://sei-network-io.slack.com/archives/C069QPFHCBC/p1779899204638349?thread_ts=1779899204.638349&cid=C069QPFHCBC)

<div><a href="https://cursor.com/agents/bc-f72b2eec-d8bc-5dd2-a495-59d4be7656d1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-f72b2eec-d8bc-5dd2-a495-59d4be7656d1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


bdchatham (commented):
This is a live bugfix, not just defense: the new test fails on main at receipt parity (giga stamps `TxType=2, gasUsed=gasLimit`; v2 writes the EndBlock synthetic receipt with zeros) while consensus fields match. State the red-proof in the description so the severity (RPC-level receipt divergence on mixed fleets) is on record.

giga/tests/giga_test.go
```diff
@@ -1951,32 +1932,51 @@ func TestGiga_FailedExecution_ProducesReceipt(t *testing.T) {
...
+	v2 := run(ModeV2Sequential)
+	require.Equal(t, uint64(1), v2.nonce)
+	for _, mode := range []ExecutorMode{ModeGigaSequential, ModeGigaOCC} {
+		got := run(mode)
+		name := mode.String()
+		CompareDeterministicFields(t, name, v2.results, got.results)
+		CompareLastResultsHash(t, name, v2.results, got.results)
+		assertReceiptsEqual(t, name, 0, v2.ctx.Mode, got.ctx.Mode, v2.receipt, got.receipt)
+		require.Equal(t, v2.nonce, got.nonce, "%s: nonce", name)
+		require.True(t, v2.usei.Equal(got.usei), "%s: usei balance", name)
```
bdchatham: The test compares sender balances but not where the fee went — the old branch swept via the success-path surplus/deferred-info plumbing while v2's error path skips the sweep, so please also assert fee-collector and per-tx coinbase balances match; a mismatch there would be an AppHash divergence, not just receipt drift.

app/app.go
```diff
@@ -2126,61 +2129,11 @@ func (app *App) executeEVMTxWithGigaExecutor(ctx sdk.Context, msg *evmtypes.MsgE
...
-		}
-		bloom := ethtypes.Bloom{}
-		app.EvmKeeper.AppendToEvmTxDeferredInfo(ctx, bloom, ethTx.Hash(), surplus)
-
-		return &abci.ExecTxResult{
-			Code:      1,
-			GasWanted: int64(ethTx.Gas()), //nolint:gosec
-			Log:       fmt.Sprintf("giga executor apply message error: %v", execErr),
-		}, nil
+		// State-transition failures are rare and their canonical fee, nonce,
```
bdchatham: Keep the EIP-7623 floor-data-gas example in this comment — every other `Execute()` failure class is pre-checked upstream, so it's the only known natural trigger and the reason the test needs just one class.

app/app.go
```diff
@@ -2016,6 +2016,7 @@ func (app *App) ProcessBlock(ctx sdk.Context, txs [][]byte, req *BlockProcessReq
 // Fallback-cause labels for the app_giga_fallback_to_v2 metric.
 const (
 	gigaFallbackValidationFailed  = "validation_failed"
+	gigaFallbackExecutionFailed   = "execution_failed"
```
bdchatham, line 2019: Note the tradeoff: this adds the first deterministic single-tx trigger (passes CheckTx, ~27.5k gas) that reverts a whole OCC batch to v2. Fine for correctness, but let's alert on `app_giga_fallback_to_v2{reason="execution_failed"}` and track per-tx OCC fallback as a follow-up.

bdchatham:
## Live mixed-fleet validation on harbor

Validated this PR end-to-end on an ephemeral harbor chain: 4 validators + one v2 RPC node (`giga_executor.enabled=false, occ_enabled=false`) on `nightly-20260715-65411b1` (= #3753 merge, pre-this-PR), plus one giga+OCC RPC node swapped between the pre-fix image and `sei/sei-chain:pr3768-6ab6c68` (built from head `6ab6c68` via `ecr.yml` workflow_dispatch).

Probe: EIP-7623 floor-data-gas tx (1000 zero calldata bytes, gasLimit 27500 — clears the ante's intrinsic check, fails inside geth's `Execute()`), exercised via qa-testing's `tests/chain_tests/floorDataGas.spec.ts` (sei-protocol/qa-testing#143/#145).

**Pre-fix giga node — divergence reproduced.** Same tx hash, `eth_getTransactionReceipt` on both nodes:

| field | giga (pre-fix) | v2 |
|---|---|---|
| `gasUsed` | `0x6b6c` (gas limit) | `0x0` |
| `cumulativeGasUsed` | `0x6b6c` | `0x0` |
| `effectiveGasPrice` | `0x165a0bc00` | `0x0` |
| `status` / `type` | `0x0` / `0x2` | `0x0` / `0x2` |

Consensus held throughout (nonce, sender balance, and fee charge matched v2 on both executors) — receipt-only divergence, invisible to the AppHash/LastResultsHash gate, per the RCA.

**This PR's image — fixed.** Giga node on `pr3768-6ab6c68`, peered against unfixed validators and the unfixed v2 node:
- qa-testing floor-data-gas spec: **4/4 passing** (was 3/4, failing on exactly `gasUsed` 27500 vs 0)
- Fresh same-tx cross-node diff: receipts **byte-identical** — `{status: 0x0, gasUsed: 0x0, cumulativeGasUsed: 0x0, type: 0x2, effectiveGasPrice: 0x0}`
- Fallback observable on the node's metrics: `sei_chain_app_giga_fallback_to_v2_total{reason="execution_failed", scope="batch"} 2` — one per probe tx, `scope="batch"` as expected under OCC
- Node synced ~4,300 blocks produced under the old image and stayed in consensus with the unfixed fleet

The qa-testing spec now guards this class continuously: it locks the canonical v2 receipt shape at the RPC surface, so a regression on any executor fails it regardless of which node serves the query.

bdchatham:
**Live A/B validation on harbor** (ephemeral chains, 4 v2 validators + 1 RPC follower with `giga_executor.enabled/occ_enabled=true` from birth, qa-testing parity spec `tests/chain_tests/floorDataGas.spec.ts` @ `b43a61ed`):

- **Pre-fix** (`ca3bcfca`, giga executor ENABLED — attested via seid log): parity spec **red** — failing EIP-7623 floor tx got a receipt with `gasUsed = 27500` (= gas limit) where v2 canon writes `0`. Receipt-exists / nonce-bumped-once / full-fee assertions all green.
- **This PR's head** (`6ab6c68c`, same topology, executor ENABLED): **4/4 green** — canonical v2 failure receipt served from the giga executor path.

Same spec, image, topology, and platform on both legs; only the seid commit differs. Confirms the divergence reproduces on a live giga-serving node and that this PR resolves it to v2 canon.

stevenlanders (approved):

bdchatham (approved):

codchen (approved):

sei-will (approved):

PR #3773, opened 2026-07-16
special error when reading below watermark

cody-littley:

Return a special error when reading a block below the watermark.

### DIFF
--- a/sei-db/ledger_db/block/block_db_test.go
+++ b/sei-db/ledger_db/block/block_db_test.go
@@ -311,10 +311,16 @@ func testPruneRefusesBelowWatermark(t *testing.T, build builder) {
 }
 
 // testPrunedDistinctFromNotFound is the crux of the ErrPruned contract: after a
-// prune, a below-watermark by-number read reports ErrPruned (gone for good — a
-// peer should stop retrying), while a never-written height at or above the
-// watermark reports a plain utils.None with a nil error (absent, but a future
-// write may fill it). Both implementations must agree.
+// prune, a below-watermark by-number read reports ErrPruned (not served while
+// below the watermark), while a never-written height at or above the watermark
+// reports a plain utils.None with a nil error (absent, but a future write may
+// fill it). Both implementations must agree.
+//
+// The prune point is placed strictly *inside* a QC's range. A QC's cohort of
+// blocks must change readability atomically — the watermark may never split a
+// cohort — so the effective watermark rounds down to the cohort boundary: the
+// whole straddled cohort stays readable, and only the fully-below first cohort
+// becomes pruned.
 func testPrunedDistinctFromNotFound(t *testing.T, build builder) {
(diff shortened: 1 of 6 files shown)


sei-db/ledger_db/block/block_db_test.go
```diff
@@ -309,6 +310,41 @@ func testPruneRefusesBelowWatermark(t *testing.T, build builder) {
...
+// watermark reports a plain utils.None with a nil error (absent, but a future
+// write may fill it). Both implementations must agree.
+func testPrunedDistinctFromNotFound(t *testing.T, build builder) {
+	committee, keys := buildCommittee()
+	batches := generateBatches(committee, keys)
+	db, _ := openFresh(t, build)
+	defer func() { _ = db.Close() }()
+	writeAll(t, db, batches)
+
+	watermark := batches[1].first
```
wen-coding: nit: can you prune in the middle of a QC to make the watermark case more interesting?

sei-tendermint/autobahn/types/block_db.go
```diff
@@ -21,6 +21,14 @@ var ErrQCNonContiguous = errors.New("block: WriteQC non-contiguous")
 // before that block (see the BlockDB ordering contract).
 var ErrBlockMissingQC = errors.New("block: WriteBlock without covering QC")
 
+// ErrPruned is returned by the by-number read methods (ReadBlockByNumber and
+// ReadQCByBlockNumber) when the requested GlobalBlockNumber is strictly below
+// the retention watermark: the record has been pruned (or is being reclaimed)
+// and will never become readable again. It is distinct from a utils.None
```
wen-coding: nit: we don't write the watermark onto disk, so never say never...

sei-tendermint/autobahn/types/block_db.go
wen-coding, line 166: do we need to update this comment?

wen-coding (approved):

sei-tendermint/autobahn/types/block_db.go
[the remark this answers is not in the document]
cody-littley: fixed

cody-littley: godoc updated

sei-db/ledger_db/block/block_db_test.go
cody-littley: This test was missing. When I added it, it exposed a real bug where we would allow the prune watermark in memory be in the middle of a QC's range. I also fixed the bug.

sei-tendermint/autobahn/types/block_db.go
cody-littley: This is still technically accurate. If there is a QC spanning blocks `100`-`110` and you ask to `PruneBefore(105)`, internally it will clamp the watermark to block `100`.

wen-coding (approved):

PR #3776, opened 2026-07-17
Part 1: Bigtable client and Kafka ingestion

Kbhat1:
## Description
- Bigtable client with frozen MVCC row-key encoding (shard|store|key|inverted-version) 
- Kafka consumer that ingests the node changelog stream into Bigtable: per-partition ordering, redelivery dedupe, offsets committed only after durable writes
- Zero node-behavior change

## Testing
- Verified end to end in GCP


sei-db/state_db/ss/offload/historical/bigtable.go
```diff
@@ -0,0 +1,749 @@
...
+			return err
+		}
+		if sleepErr := sleepWithContext(ctx, backoff); sleepErr != nil {
+			return err
+		}
+		backoff *= 2
+	}
+}
+
+func sleepWithContext(ctx context.Context, d time.Duration) error {
```
yzang2019: Identical implementations in both consumer/consumer.go and historical/bigtable.go. Extract to a shared util package.

sei-db/state_db/ss/offload/consumer/config/example-bigtable.json
```diff
@@ -0,0 +1,25 @@
+{
+  "Kafka": {
+    "Brokers": ["bootstrap.my-cluster.us-east1.managedkafka.my-gcp-project.cloud.goog:9092"],
+    "Topic": "historical-offload",
+    "GroupID": "historical-bigtable",
+    "StartOffset": "first",
+    "TLSEnabled": true,
+    "SASLMechanism": "plain",
+    "Username": "kafka-client@my-gcp-project.iam.gserviceaccount.com",
+    "Password": "<base64-encoded service account key JSON>"
```
yzang2019, line 10: Kafka SASL passwords live in a plain JSON file on disk. No secret manager integration, no envelope encryption. For a production system handling chain state, this is a security concern. At minimum, support reading credentials from environment variables or a secrets manager.

sei-db/state_db/ss/offload/historical/bigtable.go
```diff
@@ -0,0 +1,749 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/oauth"
+	"google.golang.org/grpc/keepalive"
+	"google.golang.org/grpc/status"
+)
+
+const (
+	DefaultBigtableFamily = "state"
+	DefaultBigtableShards = 256
```
yzang2019: The shard value is baked into every mutation row key: FNV32a(storeName + \0 + key) % Shards. If an operator changes Shards from 256 to 512, every Get and Has query computes a different shard for the same (storeName, key), so it looks up the wrong row prefix and returns ErrNotFound. All previously ingested data becomes unreachable with no error. This needs to either be persisted in a metadata row and validated on startup, or enforced as immutable in the config.

sei-db/state_db/ss/offload/consumer/bigtable.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+func bigtableBulkError(rows []historical.BigtableRowMutation, errs []error, err error) error {
+	if err != nil {
+		return err
+	}
+	if len(errs) != len(rows) {
+		return fmt.Errorf("bigtable returned %d mutation results for %d rows", len(errs), len(rows))
+	}
+	for i, rowErr := range errs {
```
yzang2019, line 236: If 50 out of 1024 rows fail, you see one error and no indication of how widespread the failure was. This makes it much harder to distinguish "one hot tablet is overloaded" from "everything is broken." 

Consider Include the failure count:
`return fmt.Errorf("%d of %d rows failed, first: row %q: %w", failCount, len(rows), rows[firstIdx].RowKey, errs[firstIdx])
`

sei-db/queue/kafka/consumer.go
```diff
@@ -0,0 +1,287 @@
...
+			lag = l
+		}
+	}
+	return lag
+}
+
+func (c *Consumer) writeBatchWithRetry(ctx context.Context, records []Record) error {
+	backoff := sinkBaseBackoff
+	var lastErr error
+	for attempt := 1; attempt <= sinkMaxAttempts; attempt++ {
```
yzang2019, line 233: writeBatchWithRetry retries the entire batch even after partial success, If 1023 out of 1024 mutation rows succeed and 1 fails, the retry re-sends all 1024. Since SetCell is idempotent at the same timestamp this is correct, but it amplifies write cost and latency proportional to batch size. A targeted retry of only failed rows would be more efficient — ApplyBulk already returns per-row errors.

sei-db/state_db/ss/offload/consumer/consumer.go
```diff
@@ -0,0 +1,287 @@
...
+		}
+	}
+	return msgs, true
+}
+
+func (c *Consumer) processBatch(ctx context.Context, msgs []kafka.Message) error {
+	records := make([]Record, 0, len(msgs))
+	var firstVersion, lastVersion int64
+	for i, msg := range msgs {
+		entry, err := DecodeEntry(msg.Value)
```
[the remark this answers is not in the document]
yzang2019: +1, If a Kafka message contains invalid protobuf, DecodeEntry returns an error, processBatch returns the error, the worker kills the errgroup, Run returns, and main calls log.Fatalf. On restart, the consumer re-fetches the same uncommitted message and crashes again — infinite loop. There's no dead-letter mechanism, no skip-after-N-failures, no offset advance for undecodable messages.

sei-db/state_db/ss/offload/consumer/config.go
```diff
@@ -0,0 +1,76 @@
...
+	// bigtable_bytes_written_total, bigtable_mutate_latency_seconds, ...) can be
+	// scraped. Empty disables the endpoint.
+	MetricsAddr string
+}
+
+func (c *Config) Validate() error {
+	if err := c.Kafka.Validate(); err != nil {
+		return fmt.Errorf("kafka: %w", err)
+	}
+	bigtable := c.Bigtable
```
yzang2019, line 34: BigtableConfig is a struct (not a pointer), so bigtable := c.Bigtable creates a copy. 
ApplyDefaults() fills in Family = "state" and Shards = 256 on the copy. Then Validate() passes because the copy has those defaults. But the original c.Bigtable is untouched — c.Bigtable.Family is still "".

It happens to work in practice because NewBigtableSink calls cfg.ApplyDefaults() again before using it. But if any code between LoadConfig and NewBigtableSink reads cfg.Bigtable.Family (e.g., for logging or metrics tagging), it sees an empty string despite validation having passed.

The fix is simple — apply defaults on the original, not a copy.

sei-db/state_db/ss/offload/historical/bigtable.go
```diff
@@ -0,0 +1,749 @@
...
+	}
+	if c.Shards == 0 {
+		c.Shards = DefaultBigtableShards
+	}
+}
+
+func (c BigtableConfig) Configured() bool {
+	return strings.TrimSpace(c.ProjectID) != "" ||
+		strings.TrimSpace(c.InstanceID) != "" ||
+		strings.TrimSpace(c.Table) != ""
```
yzang2019: Setting just Project without Instance or Table returns true, and the code proceeds to build a Bigtable reader with empty connection parameters. This should be AND (all required fields present)

sei-db/state_db/ss/offload/consumer/kafka.go
```diff
@@ -0,0 +1,126 @@
+package consumer
```
yzang2019: Would recommend to put Kafka consumer and publisher together into a separate package, not under ss. Something like sei-db/queue/kafka

sei-db/state_db/ss/offload/consumer/bigtable.go
```diff
@@ -0,0 +1,233 @@
+package consumer
```
yzang2019, line 1: Would recommend bundle all BigTable libraries together into a DB library, similar to littDB, and move them into sei-db/db_engine/bigtable/

PR #3779, opened 2026-07-20
PLT-819: Wire fail-closed rate limiter + MethodParser into EVM JSON-RPC HTTP (:8545)

amir-deris:
## Summary

Implements Phase 1b-core of the RPC endpoint protection work: wires `ratelimiter.Registry` + `ratelimiter.MethodParser` (from PLT-800 / #3755) into the EVM JSON-RPC HTTP admission path (`:8545`). WebSocket (`:8546`) is out of scope — tracked as 1b-ws follow-up.

Also hardens `MethodParser` and its HTTP wiring to a **fail-closed** contract (no fallback on oversize or parse failure).

### Rate limiting

- **`ratelimiter.Allow`** now takes a `method` argument; rejections record `rpc_rate_limit_rejected_total{plane,method_namespace}` with low-cardinality namespace bucketing (`ratelimiter/method_bucket.go`).
- **New config fields** (`evmrpc/config`): `rate_limiting_enabled` (default **`false`** — ships dark; operators opt in per node) and `trusted_proxy_cidrs` (default empty / trust-none). A `Config.RateLimiterConfig()` helper builds the `ratelimiter.Config` used by admission.
- **Rollout plan**: this PR ships with rate limiting **disabled by default**. The admission wiring, metrics, and config plumbing land now so it can be reviewed and tested end-to-end, but nodes must explicitly set `rate_limiting_enabled = true` (and configure `trusted_proxy_cidrs` if behind a proxy/LB) to enable it. Flipping the default to `true` fleet-wide is a follow-up once we've validated behavior in staging/canary.
- **Default change**: `ip_rate_limit_burst` default raised **400 → 1000** (aligned with `batch_request_limit`) so a full-size batch can succeed when the bucket is full. Existing nodes with `ip_rate_limit_burst = 400` already materialized in `app.toml` must bump burst (or lower `batch_request_limit`) before enabling rate limiting — `ReadConfig` refuses to start when `burst < batch_request_limit`.
- **Body cap**: method extraction uses the existing **`max_request_body_bytes`** (default 5 MiB) — no separate probe-bytes knob. One cap for size limiting, rate-limit parsing, and go-ethereum's HTTP body limit.
- **New `rateLimitMiddleware`** (`evmrpc/rate_limit_middleware.go`) + **`RateLimitGate`** (`evmrpc/rate_limit.go`):
  - Handler stack (outer → inner): `requestSizeLimiter → rateLimitMiddleware → wrapSeiLegacyHTTP → NewHTTPHandlerStack`
  - Reads the **entire** request body (bounded by `max_request_body_bytes`), extracts JSON-RPC method name(s) via `MethodParser`, applies the per-IP token bucket (method is passed through only to label the rejection metric — see per-method note below), then replays the buffered body to downstream handlers.
  - **Batch requests — admission vs. charging (important):**
    - **Admission:** the whole batch is rejected (HTTP **429**) if *any* element would exhaust the bucket; nothing is dispatched downstream.
    - **Charging:** tokens are consumed **one per batch element** via sequential `Allow()` calls — **not** atomically. If element *k* fails, tokens for elements `0..k-1` are already spent even though the batch is rejected. This matches `requestSizeLimiter`'s existing pattern (partial side effects on atomic reject) and is acceptable for Phase 1b because:
      - Well-behaved clients can split oversized batches or back off on 429.
      - `ReadConfig` enforces `ip_rate_limit_burst >= batch_request_limit`, so a **full** batch can succeed whenever the bucket is full (the “permanently impossible max batch” case is ruled out at startup).
      - With defaults (`burst = batch_request_limit = 1000`), a client sending batches ≤1000 calls only hits partial drain when the bucket is already low — not on every request.
  - **Rejections** (plain HTTP, pre-decode):
    - HTTP **413** `"request body too large"` — body exceeds cap (`rejectReasonOversize`); charges one `MethodInvalid` token so oversize floods cannot bypass the per-IP bucket
    - HTTP **400** `"bad request"` — malformed JSON-RPC / duplicate `"method"` / other parse errors (`rejectReasonUnparseable`); parse failures charge `MethodInvalid` as well
    - HTTP **429** `"too many requests"` — per-IP bucket exhausted (`rejectReasonRateLimited`)
  - Non-JSON-RPC traffic (empty GET/HEAD health checks, OPTIONS preflight) bypasses the gate; GET/HEAD **with a body** is rate-limited like POST (see `evmrpc/tx_test.go` precedent).
- One `*ratelimiter.Registry` is constructed per `HTTPServer` instance in `NewEVMHTTPServer`.

### MethodParser hardening (`ratelimiter/`)

- Package doc (`doc.go`) documents the fail-closed contract.
- Duplicate / mixed-case `"method"` keys → `ErrDuplicateMethod` (reject, not charge a cheap method while dispatching an expensive one).
- `ErrProbeLimit` means body exceeds cap; callers reject (HTTP 413), never fall back to full decode.
- `IsBodyTooLarge(err)` helper for HTTP mapping.

## Test plan

- [x] `ratelimiter`: `Allow` with method label on OTel counter; duplicate/mixed-case method rejection; body-over-limit → `ErrProbeLimit`
- [x] `evmrpc`: allow under limit, burst then 429, per-IP isolation, batch counts all methods, whole-batch reject, oversize → 413, parse-error → 400, disabled bypass, GET-with-body rate-limited like POST, body drained on rejection, oversize/read-error charge per-IP bucket
- [x] Composed stack: rate limit distinct from size budget; Content-Length oversize rejected before body read; chunked oversize charges then rate-limits
- [x] Config: `rate_limiting_enabled` / `trusted_proxy_cidrs` parse correctly; `RateLimiterConfig()` helper; burst ≥ batch guard; golden/fuzz characterization updated
- [x] `go test ./ratelimiter/... ./evmrpc/...`


evmrpc/config/config.go
```diff
@@ -631,6 +652,16 @@ func ReadConfig(opts servertypes.AppOptions) (Config, error) {
...
+	if v := opts.Get(flagRateLimitingEnabled); v != nil {
+		if cfg.RateLimitingEnabled, err = cast.ToBoolE(v); err != nil {
+			return cfg, err
+		}
+	}
+	if v := opts.Get(flagTrustedProxyCIDRs); v != nil {
+		if cfg.TrustedProxyCIDRs, err = cast.ToStringSliceE(v); err != nil {
+			return cfg, err
+		}
+	}
```
bdchatham, line 682: Note for later: as part of the configuration manager project we are creating comprehensive test coverage that asserts the behavior of config values. I know you have added lots of new config. It would be good to have a follow-up action item on these to make sure we have full test coverage over these values. I've created an AGENTS.md in https://github.com/sei-protocol/sei-chain/pull/3816 that should make this simple

amir-deris: Thanks for feedback. Created https://linear.app/seilabs/issue/PLT-874/ to make sure we have full coverage for the newly added config fields.

sei-will (approved):

PR #3787, opened 2026-07-21
storage manager

cody-littley:

Create a utility responsible for managing the lifecycle of various data stores.


sei-db/storagemanager/snapshot_store.go
```diff
@@ -0,0 +1,15 @@
+package storagemanager
+
+// A store that is persisted via snapshots (e.g. SC, SS).
+type SnapshotStore interface {
+
+	// Fetch a list of all block numbers that this store has snapshots for, in ascending sorted order.
+	GetStoredBlocks() ([]uint64, error)
```
yzang2019: Make sense, but maybe rename to GetPersistedBlockRange? I was initially thought this is gonna return the whole list of block numbers like [1,2,3,4,5,6]. Maybe add a comment since it's gonna return different things based on different backend

yzang2019 (approved):

sei-db/storagemanager/storagemanager_config.go
```diff
@@ -0,0 +1,36 @@
...
+	return &StorageManagerConfig{
+		RollbackWindow:       10_000,
+		PruneIntervalSeconds: 60,
+	}
+}
+
+// Validate the storage manager's config.
+func (c *StorageManagerConfig) Validate() error {
+	// A zero rollback window is legal: it means the system prunes as aggressively as possible.
+	if c.PruneIntervalSeconds == 0 {
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/storagemanager/snapshot_store.go
cody-littley: Discussed offline, leaving context for others.

We currently have two interfaces: one for things that store data in WALs, and another for things that store data as snapshots. Each API has slightly different semantics to capture the difference between a list of snapshots and a range that is on a WAL.

PR #3788, opened 2026-07-21
Add CommitBlock API for FlatKV

yzang2019:
## Summary

- Change FlatKV write APIs to take an explicit version (block height):
  - `ApplyChangeSets(version, changesets)`
  - `Commit(version)`
- Add Giga-only `CommitBlock(version, changesets)` that applies and commits a batch at the target height in one call.
- Stamp row last-modified heights with the apply version so `GetBlockHeightModified` matches the committed height under gapped/batch commits.
- Reject `Commit` when its version does not match the height used in `ApplyChangeSets`; empty commits may still jump ahead.
- Update composite, migration writer, wrappers, catchup, and tests to pass the version through.


sei-db/state_db/sc/flatkv/api.go
```diff
@@ -34,13 +34,19 @@ type Store interface {
...
-	// initialVersion. Must be called after LoadVersion, on a truly fresh
-	// store (no prior commits) and before any writes. Returns an error on
-	// a read-only store, on a non-fresh store, or for initialVersion <= 0.
+	// Commit persists buffered writes at the given version (block height).
+	Commit(version int64) (int64, error)
+
+	// CommitBlock is a Giga only API for commiting all changesets till a given block height.
+	// version is the target version we want to commit till. In batch commit mode,
+	// we should be passing the last block height as well as all changesets in the batch.
+	CommitBlock(version int64, cs []*proto.NamedChangeSet) (int64, error)
```
cody-littley: What is the `int64` return value returning?

sei-db/state_db/sc/flatkv/api.go
```diff
@@ -34,13 +34,19 @@ type Store interface {
...
+
+	// SetInitialVersion seeds the store so that Commit(initialVersion) is
+	// accepted as the first commit. Must be called after LoadVersion, on a
+	// truly fresh store (no prior commits) and before any writes. Returns an
+	// error on a read-only store, on a non-fresh store, or for
+	// initialVersion <= 0.
 	SetInitialVersion(initialVersion int64) error
```
cody-littley, line 60: Do we still need this if we require that callers provide the block number? We could do something like this:

- if DB is completely empty, permit any positive initial block number
- if DB is not empty, assert that next block number follows required rules (e.g. is previous block number +1, or if we allow gaps that it's strictly greater than the previous one)

sei-db/state_db/sc/flatkv/api.go
yzang2019: That was the committed block number, removed since we don't need for Giga

sei-db/state_db/sc/flatkv/api.go
yzang2019: Let me look into this

yzang2019: we can't remove it yet.

When a chain that's been running memiavl-only switches on flatkv, flatkv needs to be seeded to memIAVL.Version()+1 immediately at load time, not "whenever its first write happens to land." There can be an arbitrary number of blocks after the migration boundary with zero EVM writes, and during that gap several things depend on flatkv already reporting the seeded height/earliest-version, not 0:

EarliestVersion() — persisted durably by SetInitialVersion, distinct from "first committed version." It's the boundary used to tell "this height predates flatkv entirely" (expected) apart from "flatkv has this era but the snapshot was pruned/corrupted"

cody-littley: Sounds good to me, we can remove it once we dump memIAVL

sei-db/state_db/sc/flatkv/api.go
```diff
@@ -29,18 +29,31 @@ type Store interface {
...
-	// a read-only store, on a non-fresh store, or for initialVersion <= 0.
+	// ApplyChangeSets buffers EVM changesets (x/evm memiavl keys) and updates
+	// LtHash, stamping row last-modified heights with version. Non-EVM modules
+	// are routed into misc storage under their module prefix.
+	// May be called multiple times before a single Commit: either repeatedly
+	// at the same version (e.g. one block's writes split across several
+	// calls) or at strictly increasing versions (e.g. batching several
+	// blocks' writes together); version must never decrease from the
+	// previous pending call (see PendingVersion). The final Commit must be
+	// called with the highest version applied since the last Commit.
```
cody-littley: Nit, I suggest omitting implementation details such as where specific keys are routed internally.

sei-db/state_db/sc/flatkv/api.go
```diff
@@ -29,18 +29,31 @@ type Store interface {
...
+	ApplyChangeSets(version int64, cs []*proto.NamedChangeSet) error
+
+	// Commit persists buffered writes at the given version (block height).
+	// If ApplyChangeSets has buffered writes, version must equal the highest
+	// height those rows were stamped with (see PendingVersion).
+	Commit(version int64) (int64, error)
+
+	// CommitBlock is a Giga-only helper that applies changesets and commits
+	// them at version in one call.
+	CommitBlock(version int64, cs []*proto.NamedChangeSet) error
```
cody-littley, line 53: I suggest documenting that callers should either call `ApplyChangeSets()`+`Commit()` or `CommitBlock()` but not both, and that `ApplyChangeSets()`+`Commit()` will eventually be deprecated.

sei-db/state_db/sc/flatkv/store_catchup.go
```diff
@@ -149,17 +149,16 @@ func (s *CommitStore) catchup(targetVersion int64) (err error) {
...
-		if off, err := s.walOffsetForVersion(s.committedVersion + 1); err == nil && off > startOff {
-			if off > lastOff {
-				return nil
+		if walFirstVer <= s.committedVersion {
+			// The WAL is guaranteed to still hold the entry for
+			// committedVersion: every Commit writes it in Step 1 before
+			// advancing committedVersion in Step 4, and truncation never
+			// removes entries at or after committedVersion. Locate it
+			// directly; the replay loop below skips it
+			// (entry.Version <= s.committedVersion) and applies from there.
```
cody-littley, line 158: Do we document what `Step 1` and `Step 4` is? We should probably avoid referencing steps, and instead give a short description, e.g. "after flushing to the DB".

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -15,10 +15,21 @@ import (
...
-// Commit persists buffered writes and advances the version.
+// CommitBlock applies changesets and commits them at version in one call.
+// Giga-only helper for committing all state changes of a block.
+// TODO: make this async and pipelined
+func (s *CommitStore) CommitBlock(version int64, changesets []*proto.NamedChangeSet) error {
+	if err := s.ApplyChangeSets(version, changesets); err != nil {
+		return err
+	}
+	_, err := s.Commit(version)
+	return err
```
cody-littley: nit, I suggest wrapping errors in returns like this. It makes it a lot easier to trace the call path of an error if we do.

cody-littley (approved):
Overall LGTM, although I suggest resolving the LLM comments before merging (either by fixing issues or by commenting something like "will not do" and then hiding the comment)

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -52,8 +63,13 @@ func (s *CommitStore) Commit() (version int64, err error) {
...
 	}
-	// Auto-increment version
-	version = s.committedVersion + 1
+	if version <= s.committedVersion {
+		return 0, fmt.Errorf("flatkv: committing bad version: got %d, current %d", version, s.committedVersion)
+	}
+	if s.pendingBlockHeight != 0 && version != s.pendingBlockHeight {
+		return 0, fmt.Errorf("flatkv: commit version %d does not match applied block height %d",
+			version, s.pendingBlockHeight)
+	}
```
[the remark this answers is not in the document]
yzang2019: Not a concern today, we'll refactor WAL to support batch commit in the future properly

sei-db/state_db/sc/flatkv/api.go
```diff
@@ -29,18 +29,31 @@ type Store interface {
...
+	// ApplyChangeSets buffers EVM changesets (x/evm memiavl keys) and updates
+	// LtHash, stamping row last-modified heights with version. Non-EVM modules
+	// are routed into misc storage under their module prefix.
+	// May be called multiple times before a single Commit: either repeatedly
+	// at the same version (e.g. one block's writes split across several
+	// calls) or at strictly increasing versions (e.g. batching several
+	// blocks' writes together); version must never decrease from the
+	// previous pending call (see PendingVersion). The final Commit must be
+	// called with the highest version applied since the last Commit.
+	ApplyChangeSets(version int64, cs []*proto.NamedChangeSet) error
```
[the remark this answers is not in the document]
yzang2019: Not a concern today, we'll refactor WAL to support batch commit in the future properly

PR #3789, opened 2026-07-21
Add new Storage interface for Giga executor

yzang2019:
### `sei-db/proto` — fixed dangling `iavl` reference & flattened layout

- **`changeset.proto`** *(new)* — recreated with `package iavl;`, defining `KVPair` and `ChangeSet`. This replaces the dangling `iavl/changeset.proto` import left over from the removed `sei-iavl` module (only `changeset.pb.go` had been carried into `sei-db`). Wire names remain `iavl.KVPair` / `iavl.ChangeSet`.
- **`changelog.proto`** and **`commit_info.proto`** — moved out of `memiavl/` into flat `sei-db/proto/`, matching where the generated `.pb.go` files live. Imports updated to flat paths; `NamedChangeSet.changeset` uses `iavl.ChangeSet`.
- **`.pb.go` path sync (no full regen):** `// source:` and `RegisterFile` updated to flat names; embedded file-descriptor `name`/`dependency` strings rewritten to match. Marshal/Unmarshal code untouched.

### `sei-db/state_db/giga/api.go` — storage API for the Giga EVM executor

Top-level type is `giga.Store` (avoids colliding with SS “state store” naming).

| Type | Role |
|---|---|
| `Address [20]byte` / `Hash [32]byte` | Dependency-free local types (no `go-ethereum`/`evmc` import); convertible to/from `common.Address`/`common.Hash`/`evmc.Address`/`evmc.Hash`. `Hash` covers storage slots, balances, and code hashes — not block hashes |
| `Store` | Commit + open snapshots |
| `StateSnapshot` | Full read-only snapshot: raw KV + lifecycle + embedded EVM reads |
| `EVMStateSnapshot` | EVM account-state leaf (`AccountExists`, `GetStorage`, `GetBalance`, `GetNonce`, `GetCodeSize`, `GetCodeHash`, `GetCode`) |

Block hashes are intentionally **not** on this API: EVMC `GetBlockHash` is served from chain `BlockContext`, not SC/SS account state.

#### `Store`

- `CommitStateChanges(blockNum, changeset) error` — write into both SC and SS after each block; changes must be visible on return (hash/disk I/O may still finish async).
- `OpenSnapshot() StateSnapshot` — current block (SC ephemeral snapshot). Caller must `Close`.
- `OpenSnapshotAt(blockNum) (StateSnapshot, bool)` — historical committed height (SS). `false` if missing; when `true`, caller must `Close`.

#### `StateSnapshot`

Embeds `EVMStateSnapshot`, plus:

- `GetBlockHeight() int64`
- `Get(key) ([]byte, bool)` — point-in-time; no error return (DB failures panic / crash)
- `Close()` — sole lifecycle method (no `Reserve`/`Release`); not closing can leak

#### `EVMStateSnapshot`

Account-state reads used by the executor / keeper (code, balance, nonce, storage). None of these methods return errors: underlying DB failures panic so the process does not continue on corrupt/incomplete state.

`GetCodeHash` matches EXTCODEHASH / `keeper.GetCodeHash`: empty-code hash (`keccak256("")`) for existing no-code accounts; zero `Hash` only when the account does not exist.

## Test plan

- [ ] Confirm flat `sei-db/proto/{changelog,changeset,commit_info}.proto` and no `memiavl/` proto sources
- [ ] Confirm `.pb.go` `// source:` / `RegisterFile` / descriptor paths use flat names
- [ ] `go build ./sei-db/state_db/giga/ ./sei-db/proto/ ./sei-db/wal/ ./sei-db/state_db/statewal/`
- [ ] `go test ./sei-db/wal/ ./sei-db/state_db/statewal/`
- [ ] `gofmt -s -l` / `goimports -l` clean on touched Go files


codchen (approved):

sei-db/state_db/giga/api.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// Reserve pins the snapshot, incrementing its reference count so its
+	// underlying resources are not reclaimed while still in use. Must be
+	// paired with a corresponding call to Release.
+	Reserve()
+
+	// Release unpins the snapshot, decrementing its reference count.
+	// Once the last reference is released, the underlying resources may
+	// be reclaimed. Must only be called after a matching Reserve.
+	Release()
```
cody-littley: I suggest we document a slight tweak in semantics. We should say that a snapshot is always created with an initial reservation. So in order to garbage collect a snapshot, if we've called `Reserve()` `N` times, we must call `Release()` `N+1` times.

We should also document that it's illegal to call methods on a snapshot that has been fully released.

sei-db/state_db/giga/api.go
```diff
@@ -0,0 +1,101 @@
...
+type StateSnapshot interface {
+	EVMStateSnapshot
+
+	// GetBlockHeight returns the block height of this snapshot.
+	GetBlockHeight() int64
+
+	// Get returns the raw value stored under key in this snapshot, and
+	// whether it was found. It never observes writes made after the
+	// snapshot was taken.
+	Get(key []byte) ([]byte, bool)
```
cody-littley, line 60: Should getter methods have an error field, e.g. if the database is corrupted, IO errors, etc? Otherwise these sorts of things will need to be raised as `panic()`.

If we decide to allow these methods to panic, then I think we need strong confirmation from consensus team that all DB panics are guaranteed to crash the node. In cosmos this is not the case, which means that when the DB panics we just skip the transaction, which is extremely unsafe and likely to lead to apphash issues.

yzang2019: Discussed that we start with DB always panics on any internal error, and the up layer should agree never ignore the panics from the database

sei-db/state_db/giga/api.go
yzang2019: Simplified the interface so that caller only need to worry about closing the snapshot

sei-db/proto/changelog.proto
```diff
@@ -2,8 +2,8 @@ syntax = "proto3";
 package memiavl;
 
 import "gogoproto/gogo.proto";
-import "iavl/changeset.proto";
-import "memiavl/commit_info.proto";
+import "changeset.proto";
+import "commit_info.proto";
```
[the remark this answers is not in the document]
yzang2019: Fixed

cody-littley (approved):

PR #3791, opened 2026-07-23
Harden account key validation

codchen:
## Summary
- Add defensive validation for account key metadata before active ante handling persists it.
- Keep invalid account key metadata from being written into account state.
- Add regression coverage for invalid account key metadata in the active app ante path.

## Validation
- `go test ./app -run 'Test(CosmosStatelessChecksRejectsInvalidNestedMultisigKey|CheckPubKeysRejectsInvalidNestedMultisigKey)' -count=1`
- `go test ./app/ante -count=1`
- `gofmt -s -l .`
- `goimports -l app/ante/cosmos_checktx.go app/cosmos_checktx_test.go`

Note: `goimports -l .` reports existing generated files and nested local worktree files unrelated to this branch.

### DIFF
--- a/sei-cosmos/x/auth/ante/sigverify.go
+++ b/sei-cosmos/x/auth/ante/sigverify.go
@@ -198,6 +198,11 @@ func (sgcd SigGasConsumeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simula
 // Verify all signatures for a tx and return an error if any are invalid. Note,
 // the SigVerificationDecorator will not check signatures on ReCheck.
 //
+// NOTE: Deprecated for normal Sei App transaction handling. The active CheckTx
+// and DeliverTx ante logic lives under app/ante/; update that path for App tx
+// validation changes. This decorator remains wired for BaseApp-driven flows and
+// should not be treated as the primary App ante implementation.
+//
 // CONTRACT: Pubkeys are set in context for all signers before this decorator runs
 // CONTRACT: Tx must implement SigVerifiableTx interface
 type SigVerificationDecorator struct {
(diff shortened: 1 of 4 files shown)


sei-cosmos/x/auth/ante/sigverify.go
```diff
@@ -127,6 +133,63 @@ func (spkd SetPubKeyDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate b
...
+	defer func() {
+		if r := recover(); r != nil {
+			err = sdkerrors.Wrapf(sdkerrors.ErrInvalidPubKey, "invalid public key: %v", r)
+		}
+	}()
+
+	return validatePubKeyNoPanic(pubKey)
+}
+
+func validatePubKeyNoPanic(pubKey cryptotypes.PubKey) error {
```
masih: We call this function exactly once; I would avoid over refactoring it and would inline it into `validatePubKey`.

sei-cosmos/x/auth/ante/sigverify.go
```diff
@@ -127,6 +133,63 @@ func (spkd SetPubKeyDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate b
...
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidPubKey, "multisig threshold %d exceeds public key count %d", threshold, len(pubKeys))
+	}
+
+	for i, pk := range pubKeys {
+		if err := validatePubKey(pk); err != nil {
+			return sdkerrors.Wrapf(err, "invalid multisig public key at index %d", i)
+		}
+	}
+
+	if len(pubKey.Address()) == 0 {
```
masih: This would accept any byte slice of length 33. Can we harden the check? `Address()` just wraps the bytes into type.

sei-will (approved):

PR #3798, opened 2026-07-23
fix(evmrpc): floor `earliest` state height at genesis InitialHeight (SEI-10383)

bdchatham:
## Problem

On a fresh (never-pruned) SS-enabled RPC node, EVM state reads at the `earliest` tag and block 0 (`eth_getTransactionCount` / `eth_getCode` / `eth_getBalance` / `eth_getStorageAt` / `eth_call`) return the chain tip instead of genesis. Surfaced by the nightly release-test suite on seid `nightly-20260723-96fa55d`: `eth_getTransactionCount(addr, 'earliest')` returned the tip nonce instead of 0.

## Cause

On a never-pruned state store `GetEarliestVersion()` returns `0` (the earliest-version key is only written by pruning or state-sync). `ResolveHeight` returned that `0` for the `earliest` tag, and `CreateQueryContext` coerces height `0` to the latest committed height — so the read ran against the tip. #3739 introduced this by resolving `earliest` to the state floor (`GetEarliestVersion()`), where the prior code fell back to `blockEarliest`.

## Fix

Floor the earliest **state** height at the chain's genesis `InitialHeight`, in `Watermarks()` (which backs every read path), rather than special-casing the `earliest` tag in `ResolveHeight`:

```go
if stateEarliest < latest {
    stateEarliest = max(stateEarliest, min(m.genesisInitialHeight, latest))
}
```

- **Sources the real `InitialHeight`, not a literal `1`.** A chain started above height 1 never committed version 1, so a literal-`1` floor would make `CreateQueryContext(1)` fail (`failed to load state at height 1`). Using the genesis `InitialHeight` resolves `earliest` to the actual committed genesis height.
- **The `stateEarliest < latest` guard** preserves the pre-commit window: with no blocks yet, `stateEarliest == latest` and the floor is skipped, so `earliest` resolves to `0` and reads the checkState genesis.
- **The `min(InitialHeight, latest)` clamp** keeps the floor from ever exceeding the safe latest. A zero `InitialHeight` disables the floor.
- **Autobahn-safe.** Because the floor no longer derives from `Status.EarliestBlockHeight` (`blockEarliest`, which is `0` under Autobahn), the earliest-state tag is now correct on Autobahn nodes too — a case the earlier `min(blockEarliest, latest)` variant missed. It also drops the block/SS retention-lockstep assumption.

### Sourcing `InitialHeight`

Injected at construction from the node's parsed `GenesisDoc`, read via the already-exposed `NodeService.RPCEnvironment().GenDoc` at each `RegisterLocalServices` call site (`start.go`, `testutil/network`, `inprocess` harness). This needs:

- **no `Genesis()` RPC** — that call fails with `genesis response is large` on archive nodes whose genesis exceeds 16 MiB (and the in-process local client hits the same check);
- **no genesis-file re-read** — avoids the `DefaultNodeHome` vs `--home` path-mismatch risk;
- **no sei-tendermint API change** — `RPCEnvironment()` is already on the interface.

The value threads as an `int64` through `RegisterLocalServices` (interface + 4 impls) and the two `evmrpc` server constructors into `NewWatermarkManager`.

## Tests

`Watermarks`/`ResolveHeight` unit coverage for: the unpruned-zero floor (SEI-10383 regression guard), a chain started above height 1 (the key win over literal-`1`), the `Watermarks`-level floor, the pre-commit clamp, the state-prune floor left untouched, explicit reads at/above the floor admitted, and explicit reads below the genesis floor rejected as pruned (the hardening from moving the floor upstream). All existing watermark, genesis-committed-state, simulate, and rollback tests pass.

Refs: SEI-10383. Regressing PR: #3739. Addresses @gregprusak's comprehensive-floor feedback (the earlier `ResolveHeight`-only variant regressed `InitialHeight > 1` chains and missed Autobahn).

### DIFF
--- a/evmrpc/watermark_manager.go
+++ b/evmrpc/watermark_manager.go
@@ -127,6 +127,10 @@ func (m *WatermarkManager) ResolveHeight(ctx context.Context, blockNrOrHash rpc.
 			// to the earliest retained block (genesis) instead: height 0 is
 			// coerced to the tip by CreateQueryContext. Clamp to latest so the
 			// pre-commit window (no blocks yet) stays at 0 and reads checkState.
+			//
+			// Under Autobahn, blockEarliest stays 0 (Status leaves
+			// EarliestBlockHeight unpopulated), so earliest still resolves to the
+			// tip there; that node class is tracked separately in SEI-10385.
 			return min(blockEarliest, latest), nil
 		}
 		return stateEarliest, nil
--- a/evmrpc/watermark_manager.go
+++ b/evmrpc/watermark_manager.go
@@ -64,16 +64,20 @@ func (m *WatermarkManager) Watermarks(ctx context.Context) (int64, int64, int64,
 	stateEarliest := latest // SS disabled: only the tip is servable.
 	if m.stateStore != nil {
 		latest = min(latest, m.stateStore.GetLatestVersion())
-		// GetEarliestVersion() is 0 until the store first prunes or state-syncs,
-		// so treat that sentinel as unpruned-from-genesis and floor at
-		// blockEarliest. A nonzero floor is the real state-prune boundary, used
-		// as-is: it can sit below blockEarliest (state retained deeper than
-		// blocks), where raising it to the block floor would hide retained state.
+		// GetEarliestVersion() is 0 until the store first prunes or state-syncs.
+		// An unpruned store retains state from genesis, so its earliest servable
+		// state is height 1. A nonzero value is the real state-prune floor, used
+		// as-is: it is a state-availability boundary independent of block pruning
+		// and is not coupled to blockEarliest.
 		stateEarliest = m.stateStore.GetEarliestVersion()
 		if stateEarliest == 0 {
-			stateEarliest = blockEarliest
+			stateEarliest = 1
 		}
 	}
+	// State is never servable above the tip, and before the first commit
+	// blockEarliest can lead latest. Clamp so `earliest` stays on the servable
+	// (pre-commit checkState) path rather than a future height.
+	stateEarliest = min(stateEarliest, latest)
 	return blockEarliest, stateEarliest, latest, nil
 }
 
--- a/evmrpc/watermark_manager.go
+++ b/evmrpc/watermark_manager.go
@@ -61,29 +61,11 @@ func (m *WatermarkManager) Watermarks(ctx context.Context) (int64, int64, int64,
 	)
 
 	// State store heights (historical state DB) may lag behind block pruning.
-	stateEarliest := latest // SS disabled: only the tip is servable.
+	stateEarliest := latest // no historical storage => just the current state.
 	if m.stateStore != nil {
 		latest = min(latest, m.stateStore.GetLatestVersion())
-		// GetEarliestVersion() is 0 until the store first prunes or state-syncs.
-		// An unpruned store retains state from genesis, so its earliest servable
-		// state is the chain's genesis height. A nonzero value is the real
-		// state-prune floor, used as-is: it is a state-availability boundary
-		// independent of block pruning and is not coupled to blockEarliest.
 		stateEarliest = m.stateStore.GetEarliestVersion()
-		if stateEarliest == 0 {
-			// Source genesis from InitialHeight (as getBlockNumber does), not a
-			// literal 1, so chains started above height 1 resolve correctly.
-			genesisRes, err := m.tmClient.Genesis(ctx)
-			if err != nil {
-				return 0, 0, 0, err
-			}
-			stateEarliest = genesisRes.Genesis.InitialHeight
-		}
 	}
-	// State is never servable above the tip, and before the first commit
-	// blockEarliest can lead latest. Clamp so `earliest` stays on the servable
-	// (pre-commit checkState) path rather than a future height.
-	stateEarliest = min(stateEarliest, latest)
 	return blockEarliest, stateEarliest, latest, nil
 }
 
@@ -111,7 +93,7 @@ func (m *WatermarkManager) EarliestStateHeight(ctx context.Context) (int64, erro
 // an error explaining whether it is too old (pruned) or too new (not yet
 // available).
 func (m *WatermarkManager) ResolveHeight(ctx context.Context, blockNrOrHash rpc.BlockNumberOrHash) (int64, error) {
-	_, stateEarliest, latest, err := m.Watermarks(ctx)
+	blockEarliest, stateEarliest, latest, err := m.Watermarks(ctx)
 	if err != nil {
 		return 0, err
 	}
@@ -140,6 +122,13 @@ func (m *WatermarkManager) ResolveHeight(ctx context.Context, blockNrOrHash rpc.
 	case rpc.SafeBlockNumber, rpc.FinalizedBlockNumber, rpc.LatestBlockNumber, rpc.PendingBlockNumber:
 		return latest, nil
 	case rpc.EarliestBlockNumber:
+		if stateEarliest == 0 {
+			// An unpruned store reports earliest state as 0. Resolve `earliest`
+			// to the earliest retained block (genesis) instead: height 0 is
+			// coerced to the tip by CreateQueryContext. Clamp to latest so the
+			// pre-commit window (no blocks yet) stays at 0 and reads checkState.
+			return min(blockEarliest, latest), nil
+		}
 		return stateEarliest, nil
 	}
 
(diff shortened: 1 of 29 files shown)


bdchatham:
Updated in `2a5ad54d` after an independent multi-lens review. Summary of what changed and why:

**Correction (blocking): `max()` → zero-sentinel.** The first commit used `stateEarliest = max(blockEarliest, GetEarliestVersion())`. That over-couples the block and state floors: on a mature full node where state is retained deeper than blocks (`ss-keep-recent=100000` > block `keep-recent=86400`), `GetEarliestVersion() = tip-100000 < blockEarliest = tip-86400`, so `max` would raise `stateEarliest` to `tip-86400` and reject explicit historical state reads in `[tip-100000, tip-86400)` — state that is physically retained — as `"pruned"`. That's a fail-closed regression versus `main`.

The fix now falls back to `blockEarliest` **only when `GetEarliestVersion() == 0`** (a never-pruned store, where the earliest-version key is unwritten); a nonzero state-prune floor is used as-is. This still fixes the fresh-node `earliest`→tip bug (SEI-10383) without the regression. `TestStateEarliestUsesPruneFloorBelowBlockEarliest` guards it (fails under `max()`, passes with the sentinel).

**Re the bot notes:**
- The `Codex` "block 0" High is a false positive, as `seidroid` noted: `rpc.EarliestBlockNumber == BlockNumber(0)`, so `"0x0"` is caught by the `earliest` case and returns `stateEarliest` directly (never reaches the pruned path). Covered by the existing test.
- The receipt-store zero-clamp in the same function is tracked separately in SEI-10384 (out of scope here).

**Known limitation (noted, out of scope):** the floor is only as strong as `blockEarliest`. If CometBFT ever reports `EarliestBlockHeight == 0` (a pre-first-block window), `earliest` can still resolve to 0; that window overlaps SEI-10384 and is not addressed here.

One open question for a maintainer: is the block-pruned / state-retained-deeper config (state queryable below the earliest retained block) officially supported? The zero-sentinel preserves today's behavior there; if that config should instead reject such reads, the semantics belong in a follow-up.

bdchatham:
Both Bugbot findings were real and are fixed in `a43b79a2e`.

**1. Pre-commit future-height panic (the more serious one).** Confirmed by tracing: with no committed blocks, `latest = 0` while `blockEarliest = 1`, so flooring `stateEarliest` to `blockEarliest` made `stateEarliest > latest`; `earliest` returned that future height and `RPCContextProvider` panics. I introduced this. Fix: clamp `stateEarliest = min(stateEarliest, latest)`, so before the first commit `earliest` resolves to `0` and stays on the checkState path. (One correction: this did not affect `genesis_latest_state_test` — those tests build the `WatermarkManager` with a nil stateStore, i.e. the SS-disabled path this change never touches — but the panic mechanism was real for an SS-enabled node in that window.)

**2. Zero-sentinel re-coupling block/state floors.** Right — flooring the unpruned sentinel to `blockEarliest` reintroduced the coupling this PR removed for nonzero floors. Fixed by flooring `GetEarliestVersion()==0` to genesis (height 1) instead, so retained state below a pruned block floor stays queryable.

New guards: `TestStateEarliestFloorsAtGenesisWhenUnpruned` (unpruned SS with blocks pruned to 5 → `stateEarliest` 1, and a retained height 3 still resolves) and `TestEarliestClampedToLatestBeforeFirstCommit` (latest 0 → earliest clamps to 0). Both fail against the prior commit and pass now; all existing watermark and genesis-committed-state tests still pass.

bdchatham:
Simplified substantially in `eb25ef9e8` after review pushback (and a good nudge to keep parity with the old behavior).

The earlier commits changed `stateEarliest` inside `Watermarks()`. That value does double duty — it resolves the `earliest` tag **and** gates explicit historical heights — and it backs the entire read path, so each variant I tried regressed something main handled: `max()` over-coupled the floors, `blockEarliest` reintroduced a pre-commit future-height panic, and sourcing `InitialHeight` via `Genesis()` made the whole read path fail on archive nodes with a >16 MiB genesis (the systems dissent).

`main` already handled all of that. The only real bug is that the `earliest` tag resolved to `0`, which `CreateQueryContext` coerces to the tip. So the fix is now confined to `ResolveHeight`'s `rpc.EarliestBlockNumber` branch, and `Watermarks()` is byte-identical to main. Net production diff: one line plus one guard. Everything else — explicit-height gating, the pre-commit path, no `Genesis()` on the hot path — is main's behavior unchanged.

bdchatham:
On the Autobahn finding (cursor "Earliest still tip under Autobahn", claude 🔴): the mechanism is correct — under Autobahn `Status().SyncInfo.EarliestBlockHeight` is permanently 0 (status.go feeds it only from the CometBFT BlockStore, which Autobahn does not populate), so `min(blockEarliest, latest)` would return 0 → tip for state reads.

But the node this PR fixes runs **classic CometBFT**, not Autobahn — verified live: the harbor nightly rpc node reports `earliest_block_height = 1` (Autobahn would report 0). So `blockEarliest = 1` there, and this PR resolves `earliest → 1 → genesis`, fixing the reported SEI-10383 failure. `evm-ss-split=true` on that node is giga *storage*, which is independent of the giga *consensus* router; sei-config does not enable Autobahn consensus.

The Autobahn RPC case is real but a separate node class (its EVM state reads route through `ctxProvider`, so they are reachable and would hit this), and Autobahn block-data RPC is still WIP (broadly non-functional per the status.go `TODO(autobahn)`). Tracked as a follow-up — **SEI-10385** — with the reviewers' recommended direction: source the floor from the immutable genesis `InitialHeight` (which survives Autobahn's empty block store), resolved once at construction. Keeping this PR scoped to the minimal classic fix.

evmrpc/watermark_manager_test.go
```diff
@@ -183,6 +183,74 @@ func TestStateWatermarksCanLagBlocks(t *testing.T) {
...
+// unpruned node), never a literal 1.
+func TestResolveEarliestToGenesisOnUnprunedNode(t *testing.T) {
+	tmClient := &fakeTMClient{
+		status: &coretypes.ResultStatus{SyncInfo: coretypes.SyncInfo{LatestBlockHeight: 200, EarliestBlockHeight: 100}},
+	}
+	stateStore := &fakeStateStore{latest: 200, earliest: 0}
+	wm := newTestWatermarkManager(tmClient, 200, stateStore, 200)
+
+	earliest := rpc.EarliestBlockNumber
+	resolved, err := wm.ResolveHeight(context.Background(), rpc.BlockNumberOrHash{BlockNumber: &earliest})
```
amir-deris: Could use `t.Context()` here instead of `context.Background()`

amir-deris (approved):

sei-will (approved):

evmrpc/watermark_manager_test.go
masih: yep for test context we should be using t.Context()

Kbhat1 (approved):

yzang2019 (approved):

evmrpc/watermark_manager.go
```diff
@@ -66,6 +74,19 @@ func (m *WatermarkManager) Watermarks(ctx context.Context) (int64, int64, int64,
...
+	// its earliest version as 0 (the earliest-version key is only written by
+	// pruning or state-sync), which would otherwise resolve `earliest` to the
+	// tip via CreateQueryContext's height-0 coercion. Guarding on
+	// stateEarliest < latest leaves the pre-commit window (no blocks yet, where
+	// stateEarliest == latest) reading the genesis checkState, and the clamp to
+	// latest keeps the floor from ever exceeding the safe latest. Using the real
+	// InitialHeight (not a literal 1) keeps this correct for chains started
+	// above height 1, whose version 1 was never committed.
+	if stateEarliest < latest {
+		stateEarliest = max(stateEarliest, min(m.genesisInitialHeight, latest))
```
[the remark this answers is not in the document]
bdchatham: Resolving as a non-issue — the routing premise doesn't hold. `rpc.EarliestBlockNumber == BlockNumber(0)` (go-ethereum `rpc/types.go`), and `0x0` unmarshals to `BlockNumber(0)` — the same value as `earliest`. So an explicit `0x0` is caught by the `case rpc.EarliestBlockNumber` branch and returns the genesis-floored `stateEarliest`; it never reaches `getBlockNumber`/`ensureWithinWatermarks`. Block `0` and `earliest` resolve identically (to genesis), so there's no inconsistency, and the PR description is accurate.

`TestExplicitReadBelowGenesisFloorRejected` covers a different case: an explicit *numeric* height (50) below `InitialHeight` (100) on a chain started above height 1 — which is the intended pruned-error hardening, not block 0.

pompon0 (approved):

PR #3806, opened 2026-07-24
Allow iteration to start at a specific key in littDB/blockDB

cody-littley:

Allow iteration to start at any key in either LittDB or BlockDB.

### DIFF
--- a/sei-db/db_engine/litt/disktable/disk_table.go
+++ b/sei-db/db_engine/litt/disktable/disk_table.go
@@ -1308,6 +1308,22 @@ func (d *DiskTable) IteratorAt(key []byte, reverse bool) (litt.Iterator, bool, e
 	}
 
 	// Scan the located segment's keys (insertion order) for the exact key to fix the start position.
+	//
+	// TODO(perf): positioning is O(keys in segment). The key-file read itself is not wasted — the
+	// iterator walks this same slice — but the scan is pure overhead on top of it. Two ways to get to
+	// O(log n), neither free:
+	//
+	//   - Fix the key size, or enforce a maximum so a key record can occupy a fixed number of bytes.
+	//     Records are variable-stride today (kind, uint16 key length, key, address — see
+	//     keyFile.readKeys), so the file cannot be indexed arithmetically; a fixed stride would allow
+	//     binary searching the file directly. Only sound where insertion order is also key order,
+	//     since the key file is written in insertion order.
+	//   - Build a small per-segment index over the key file, shaped like a binary search tree. Works
+	//     regardless of key ordering, at the cost of another file per segment.
+	//
+	// Both trade one sequential key-file read for random reads, so either needs benchmarking before
+	// adoption. Deferred until there is evidence it matters: if iteration is mostly a startup
+	// activity, reading one segment's key file is noise.
 	keys, err := segs[segPos].GetKeys()
 	if err != nil {
 		return nil, false, errors.Join(
--- a/sei-db/db_engine/litt/disktable/disk_table.go
+++ b/sei-db/db_engine/litt/disktable/disk_table.go
@@ -1242,21 +1242,45 @@ func (d *DiskTable) IteratorAt(key []byte, reverse bool) (litt.Iterator, bool, e
 			"cannot process IteratorAt() request, DB is in panicked state due to error: %w", err)
 	}
 
-	// Locate the segment that holds key via the keymap, exactly as Get does. The keymap is the index
-	// whose job is this lookup; we trust it to identify the segment and do not cross-check it against
-	// the other segments in the snapshot. A key still present in the keymap has a live entry, so its
-	// segment cannot be garbage collected out from under the snapshot opened below.
-	address, ok, err := d.keymap.Get(key)
+	// Existence check, cache before keymap — the same oracle Get uses. The keymap lags writes: a key
+	// lives in the unflushed data cache from Put until its keymap entry is durable, and is pruned from
+	// the cache only after the keymap put succeeds, so a key absent from both is genuinely absent. This
+	// keeps the not-found case cheap: no segment is sealed and no barrier is paid for a key that was
+	// never written. A key present in the keymap has a live entry, so its segment cannot be garbage
+	// collected out from under the snapshot opened below.
+	_, inCache := d.unflushedDataCache.Load(util.UnsafeBytesToString(key))
+	address, inKeymap, err := d.keymap.Get(key)
 	if err != nil {
 		return nil, false, fmt.Errorf("failed to look up key address: %w", err)
 	}
-	if !ok {
+	if !inKeymap && !inCache {
 		return nil, false, nil
 	}
 
 	segs, err := d.openSnapshot()
 	if err != nil {
 		return nil, false, err
+	}
+
+	if !inKeymap {
+		// The key is in the unflushed data cache but not yet in the keymap. openSnapshot sealed the
+		// mutable segment, which handed every pending key to the keymap manager before returning, so
+		// draining the manager's buffered puts makes the keymap lookup authoritative. A miss after the
+		// barrier means the key was deleted concurrently.
+		if err = d.keymapManager.syncPuts(); err != nil {
+			return nil, false, errors.Join(
+				fmt.Errorf("failed to sync keymap manager puts: %w", err),
+				d.abandonSnapshot(segs))
+		}
+		address, inKeymap, err = d.keymap.Get(key)
+		if err != nil {
+			return nil, false, errors.Join(
+				fmt.Errorf("failed to look up key address: %w", err),
+				d.abandonSnapshot(segs))
+		}
+		if !inKeymap {
+			return nil, false, d.abandonSnapshot(segs)
+		}
 	}
 
 	// Find the snapshot position of the keymap-identified segment. segs is ordered by ascending segment
--- a/sei-db/db_engine/litt/disktable/disk_table.go
+++ b/sei-db/db_engine/litt/disktable/disk_table.go
@@ -1,6 +1,7 @@
 package disktable
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"log/slog"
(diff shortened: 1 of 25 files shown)


sei-db/db_engine/litt/disktable/disk_table.go
```diff
@@ -1223,23 +1224,117 @@ func (d *DiskTable) Iterator(reverse bool) (litt.Iterator, error) {
...
+
+	segs, err := d.openSnapshot()
+	if err != nil {
+		return nil, false, err
+	}
+
+	// Find the snapshot position of the keymap-identified segment. segs is ordered by ascending segment
+	// index; a linear scan suffices here. (This could be a binary search if it ever showed up as a
+	// hotspot — it won't; even tens of thousands of segment pointers is negligible next to the key-file
+	// read below.)
```
wen-coding, line 1296: Dunno whether this matters for your performance, but the most common use case on consensus side is to read all blocks > lastExecuted and their companion QCs, under normal case this means ~100 blocks.

I would imagine for this use case a reverse lookup  with a stop would be most efficient, but I don't know enough about littDB internals to compare alternatives.

sei-tendermint/autobahn/types/block_db.go
```diff
@@ -189,6 +189,35 @@ type BlockDB interface {
...
+	// served). With reverse == false the iterator yields the block at the
+	// (clamped) start and every higher block, ascending; with reverse ==
+	// true it yields the start block and every lower retained block,
+	// descending.
+	//
+	// If no block exists at the (clamped) start — e.g. n is past the last
+	// written block — the iterator is empty (Next immediately returns
+	// false). Same snapshot, single-goroutine, and must-close semantics as
+	// Blocks.
+	BlocksAt(n GlobalBlockNumber, reverse bool) (BlockIterator, error)
```
wen-coding: Good, we probably don't need Blocks() and QCs() after this.

nit: BlocksAt sounds like getting one block

sei-db/db_engine/litt/disktable/disk_table.go
```diff
@@ -1223,23 +1224,117 @@ func (d *DiskTable) Iterator(reverse bool) (litt.Iterator, error) {
...
+	}
+
+	if reverse {
+		return newReverseIterator(d, segs), nil
+	}
+	return newForwardIterator(d, segs), nil
+}
+
+// IteratorAt returns an iterator positioned at key. See litt.Table.IteratorAt for the contract.
+func (d *DiskTable) IteratorAt(key []byte, reverse bool) (litt.Iterator, bool, error) {
```
yzang2019, line 1239: Do we support pass in a key prefix here?
In the case where ReceiptDB primary index is (BlockNum, TxIndex), how do we create an iterator just for a given BlockNumber if txIndex is unknown to find All Receipts by block number?

sei-db/ledger_db/block/littblock/litt_block_db.go
```diff
@@ -357,6 +357,56 @@ func (s *blockDB) QCs(reverse bool) (types.QCIterator, error) {
...
+	start := n
+	if uint64(start) < watermark {
+		start = types.GlobalBlockNumber(watermark)
+	}
+
+	// A QC is stored under its First as the primary key, with covered-number aliases for the rest of its
+	// range. Positioning at qcKey(start) for a mid-range start would land on an alias and skip past the
+	// covering QC's primary, so resolve the covering QC and position at its First instead. A missing
+	// covering QC means start is past the last written QC — nothing to yield.
+	covering, err := s.ReadQCByBlockNumber(start)
```
yzang2019: QCsAt first calls ReadQCByBlockNumber(start) (which opens its own snapshot internally to resolve the covering QC), then calls IteratorAt(qcKey(first), reverse) (which opens a second snapshot) Between the two, a concurrent PruneBefore could remove the covering QC. The ReadQCByBlockNumber succeeds and returns first, but IteratorAt then can't find qcKey(first) in the new snapshot and returns an empty iterator — silently dropping a QC the caller knows exists.

Sounds like an edge case but do you forsee this is a problem or not?

sei-db/db_engine/litt/disktable/disk_table.go
```diff
@@ -1223,23 +1224,117 @@ func (d *DiskTable) Iterator(reverse bool) (litt.Iterator, error) {
...
+		}
+	}
+	if segPos < 0 {
+		// The key's segment is not part of the readable snapshot (e.g. a concurrent prune/GC boundary).
+		// Report not found; surface any error releasing the snapshot.
+		return nil, false, d.abandonSnapshot(segs)
+	}
+
+	// Scan the located segment's keys (insertion order) for the exact key to fix the start position.
+	keys, err := segs[segPos].GetKeys()
```
yzang2019, line 1327: GetKeys() reads all keys from the target segment, then linear-scans for the match. The keys are reused by the iterator so the I/O isn't wasted, but for segments with many keys, this is O(keys_in_segment) work just to position. If the iteratorAt is called frequently, this might be a performance concern unless each segment is small enough.

For BlockDB or ReceiptDB where keys are sorted, is there a way to achive O(logN) for the key search?

sei-db/db_engine/litt/disktable/disk_table.go
cody-littley: Reverse lookup is actually really bad from an efficiency standpoint. Reading keys oldest to newest causes sequential file reads. Reading in the other direction results in lots of random IO. Much better to do regular iteration.

In this particular case, the DB has to to do a linear scan of several thousand keys. This is probably less than a megabyte of data to inspect. Additionally, the DB still has to parse this whole file anyways, so we can't escape an O(n) touch of the keys in the target segment.

sei-db/db_engine/litt/disktable/disk_table.go
cody-littley: No support for prefixes. This is a fundamental limitation of the storage engine.

PebbleDB allows iteration to start at a prefix because keys are inherently ordered in lexicographical order inside the LSM tree. LittDB, on the other hand, orders keys by order of insertion, not by alphabetical order. Requesting to iterate after a prefix doesn't map well. There may be 50 keys with a given prefix, and those 50 keys may appear on completely different parts of the data store.

> how do we create an iterator just for a given BlockNumber if txIndex is unknown to find All Receipts by block number?

We can iterate in order written. So if you know the first transaction in a block, you can iterate by starting with the receipt at transaction index 0, then continuing until you find the first transaction receipt from the next block.

sei-db/db_engine/litt/disktable/disk_table.go
```diff
@@ -1223,23 +1224,117 @@ func (d *DiskTable) Iterator(reverse bool) (litt.Iterator, error) {
...
+	// whose job is this lookup; we trust it to identify the segment and do not cross-check it against
+	// the other segments in the snapshot. A key still present in the keymap has a live entry, so its
+	// segment cannot be garbage collected out from under the snapshot opened below.
+	address, ok, err := d.keymap.Get(key)
+	if err != nil {
+		return nil, false, fmt.Errorf("failed to look up key address: %w", err)
+	}
+	if !ok {
+		return nil, false, nil
+	}
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/db_engine/litt/disktable/disk_table.go
```diff
@@ -1223,23 +1224,117 @@ func (d *DiskTable) Iterator(reverse bool) (litt.Iterator, error) {
...
+	if ok, err := d.errorMonitor.IsOk(); !ok {
+		return nil, false, fmt.Errorf(
+			"cannot process IteratorAt() request, DB is in panicked state due to error: %w", err)
+	}
+
+	// Locate the segment that holds key via the keymap, exactly as Get does. The keymap is the index
+	// whose job is this lookup; we trust it to identify the segment and do not cross-check it against
+	// the other segments in the snapshot. A key still present in the keymap has a live entry, so its
+	// segment cannot be garbage collected out from under the snapshot opened below.
+	address, ok, err := d.keymap.Get(key)
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/ledger_db/block/littblock/litt_block_db.go
```diff
@@ -357,6 +357,56 @@ func (s *blockDB) QCs(reverse bool) (types.QCIterator, error) {
...
+
+	it, found, err := s.table.IteratorAt(qcKey(first), reverse)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open qcs iterator at %d: %w", first, err)
+	}
+	if !found {
+		return &qcIterator{it: nil, watermark: watermark}, nil
+	}
+	return &qcIterator{it: it, watermark: watermark}, nil
+}
```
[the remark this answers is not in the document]
cody-littley: stale comment

sei-tendermint/autobahn/types/block_db.go
cody-littley: Simplified to a single method.

sei-db/ledger_db/block/littblock/litt_block_db.go
cody-littley: I ended up simplifying iteration in a way that also accidentally resolves this issue. There is now a single iterator that walks both blocks and QCs, meaning there is no possibility for the two to be out of sync.

sei-db/db_engine/litt/disktable/disk_table.go
cody-littley: Several strategies, but both would require nontrivial LittDB changes. If we do one of these, I'd suggest we leave them for a follow up PR.

Although both of these changes would bring time down to O(log n) on average, the danger is that it would also result in more random read IO. If we were to adopt one of these strategies, we'd probably need to carefully benchmark.

Overall, I'm wondering if we might want to delay this optimization until we have evidence of it being a perf problem. If we are mostly just iterating at startup, we probably won't care much about time to read a single segment's keyfile.

# Option A: fixed size keys

Right now keys area of variable size. If we made keys fixed sized or enforced a maximum key size (allowing us to represent on disk with fixed number of bytes), we could do a binary search on the key file.

# Option B

Build a little index over each segment. Would look like a binary  search tree.

yzang2019: Not a blocker, that comment options are good enough, maybe left a TODO for future optimiazations

sei-db/db_engine/litt/disktable/disk_table.go
yzang2019: But we don't know for sure if the transaction always start at index 0, we just wanna start from any first transaction in a block

yzang2019: Not a blocker actually, we can still get this by querying BlockDB

yzang2019 (approved):

sei-tendermint/autobahn/types/block_db.go
wen-coding: Sorry, which ones are simplfied to a single method?

sei-db/db_engine/litt/disktable/disk_table.go
```diff
@@ -1223,23 +1224,117 @@ func (d *DiskTable) Iterator(reverse bool) (litt.Iterator, error) {
...
+	if err != nil {
+		return nil, false, fmt.Errorf("failed to look up key address: %w", err)
+	}
```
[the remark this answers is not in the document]
cody-littley: should be fixed now

sei-tendermint/autobahn/types/block_db.go
```diff
@@ -153,41 +159,38 @@ type BlockDB interface {
...
+	// and is not promised here.
 	//
-	// Same snapshot, single-goroutine, and must-close semantics as
-	// Blocks.
-	QCs(reverse bool) (QCIterator, error)
+	// The iterator captures a snapshot of the records present when it is
+	// created; records written afterward are not observed. It is NOT safe
+	// for concurrent use and MUST be closed when no longer needed (see
+	// BlockDBIterator.Close).
+	Iterator(n GlobalBlockNumber) (BlockDBIterator, error)
```
wen-coding, line 198: Yes, I guess we have lastBlockHeight from App.Info(). Theoretically since App.Info() and BlockDB follow different persistent path, it's possible App.Info() tip is ahead of BlockDB tip....

Talked with @pompon0 about this, we decided to just hard-panic in that case for now. In the future can we expect storage to reconcile between last executed and last committed and give us a coherent picture?

sei-tendermint/internal/autobahn/data/state.go
```diff
@@ -204,86 +204,64 @@ func NewState(cfg *Config, blockDB types.BlockDB) (*State, error) {
...
-				}
-				// updateNextBlock only advances nextBlock through contiguous present
-				// entries, so runPersist always writes [persistedBlock, nextBlock)
-				// fully populated. A gap here means BlockDB corruption.
-				if n != nextExpect {
-					return fmt.Errorf("%w: expected %d, got %d", types.ErrBlockGap, nextExpect, n)
+				if !pos.HasBlock {
+					// The iteration tail: the covering QC is persisted but this block is
+					// not (lost in a crash, or written ahead of its blocks).
+					continue
```
wen-coding, line 270: It will never happen that we have block N and N+2 but no N+1 right?

sei-tendermint/internal/autobahn/data/state.go
```diff
@@ -204,86 +204,64 @@ func NewState(cfg *Config, blockDB types.BlockDB) (*State, error) {
...
-// TODO: Push gap / first-block / QC-coverage consistency checks down into the
-// BlockDB implementation so loadFromBlockDB can assume a consistent view.
 func (s *State) loadFromBlockDB(blockDB types.BlockDB) error {
 	for in := range s.inner.Lock() {
-		// Restore QCs from BlockDB. On the first QC, skipTo its GlobalRange.First
-		// to advance past any pruned prefix. Subsequent QCs must be consecutive —
-		// insertQC errors on any gap.
 		err := func() error {
-			it, err := blockDB.QCs(false)
+			it, err := blockDB.Iterator(0)
```
wen-coding, line 228: I think one iterator for both QC and block is fine. If it's more convenient for you, we don't even need to associate QC with blocks when we return here. This API is for startup only, so if you return me a dump of contents in WAL, every element can either be QC or block, that's also fine.

We can read the whole thing first, then do a loop to check "every block has qc covered" later. And non-contiguous QC/block can already be rejected when we insert.

wen-coding (approved):

sei-tendermint/autobahn/types/block_db.go
cody-littley: There is now a single method that builds an iterator. Used to be method for blocks, method for QCs, and two variations of each that would iterate either from the start or from a specific block height.

sei-db/db_engine/litt/disktable/disk_table.go
cody-littley: Added a TODO in case we need to revisit this

sei-tendermint/autobahn/types/block_db.go
cody-littley: Yes, storage can provide this info.

sei-db/ledger_db/block/littblock/litt_block_iterator.go
```diff
@@ -5,123 +5,264 @@ import (
...
-func (b *blockIterator) Close() error {
-	if err := b.it.Close(); err != nil {
-		return fmt.Errorf("failed to close blocks iterator: %w", err)
+	if l.heldBlock && l.heldNumber != next {
+		// Blocks are written densely, so a missing number below the highest persisted block can
+		// only be corruption.
+		return types.Position{}, false,
+			fmt.Errorf("%w: corrupt store: block gap at %d (next persisted block is %d)",
+				types.ErrBlockGap, next, l.heldNumber)
 	}
```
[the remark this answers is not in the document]
cody-littley: slop

sei-tendermint/internal/autobahn/data/state.go
cody-littley: Correct, this should be impossible. The iterator takes a consistent snapshot of the DB to walk over, and the database does not permit out of order deletion. Meaning the iterator will never be able to observe gaps like this.

sei-tendermint/internal/autobahn/data/state.go
cody-littley: My personal preference would be to leave this as it is for now. My understanding of the code in this file is thin, and I'd like to reduce the amount of changes I make to it in this PR.

cody-littley: Also, it's probably inefficient to iterate all blocks. Leaving it at iterating all blocks because that's what it currently does, but I'd suggest we fix that in the future. At autobahn scale, this will be a very expensive iteration at startup time.

sei-db/ledger_db/block/blocksim/blocksim.go
```diff
@@ -175,39 +175,30 @@ func NewBlockSim(
...
-			return 0, 0, fmt.Errorf("failed to advance block iterator: %w", err)
+			return 0, 0, fmt.Errorf("failed to advance ledger iterator: %w", err)
 		}
 		if !ok {
 			break
 		}
-		blocks++
-	}
-
-	qcIt, err := db.QCs(false)
```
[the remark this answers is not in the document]
cody-littley: will not fix, benchmark cares about perf not boundary conditions like this

sei-db/ledger_db/block/littblock/litt_block_db.go
```diff
@@ -341,20 +389,90 @@ func (s *blockDB) Status() types.DBStatus {
...
+	}
+	if hasBlocks && start < firstBlock {
+		start = firstBlock
+	}
+
+	if start >= nextQC {
+		// Nothing is covered at or above start. NextQC is zero on an empty store, which this
+		// comparison also classifies as past-coverage.
+		return &blockDBIterator{}, nil
+	}
```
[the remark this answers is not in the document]
cody-littley: will not fix

sei-db/ledger_db/block/littblock/litt_block_db.go
```diff
@@ -341,20 +389,90 @@ func (s *blockDB) Status() types.DBStatus {
 	return tips
 }
 
-func (s *blockDB) Blocks(reverse bool) (types.BlockIterator, error) {
-	it, err := s.table.Iterator(reverse)
+func (s *blockDB) Iterator(n types.GlobalBlockNumber) (types.BlockDBIterator, error) {
+	watermark := s.watermark.Load()
+	nextQC, oldestQCStart := s.qcCoverage()
+	firstBlock, hasBlocks := s.blockFloor()
```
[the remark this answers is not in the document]
cody-littley: edge case closed

sei-db/ledger_db/block/blocksim/blocksim.go
```diff
@@ -175,39 +175,30 @@ func NewBlockSim(
...
-	if err != nil {
-		return 0, 0, fmt.Errorf("failed to open QC iterator: %w", err)
-	}
-	defer func() { _ = qcIt.Close() }()
-	for {
-		ok, err := qcIt.Next()
-		if err != nil {
-			return 0, 0, fmt.Errorf("failed to advance QC iterator: %w", err)
+		if pos.QC.QC().GlobalRange().First == pos.Number {
```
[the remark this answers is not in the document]
cody-littley: benchmark, will not fix

sei-tendermint/internal/autobahn/data/state.go
```diff
@@ -202,88 +202,78 @@ func NewState(cfg *Config, blockDB types.BlockDB) (*State, error) {
...
 				}
-				nextExpect++
-				blk, err := it.Block()
+				blkOpt, err := it.Block()
 				if err != nil {
-					return err
+					return fmt.Errorf("read block %d from BlockDB: %w", n, err)
 				}
-				qc := in.qcs[n]
+				blk := blkOpt.OrPanic(fmt.Sprintf("block %d absent at a HasBlock position", n))
```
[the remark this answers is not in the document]
cody-littley: replay is startup only, this is trivial cost

sei-db/ledger_db/block/littblock/litt_block_iterator.go
```diff
@@ -5,123 +5,267 @@ import (
...
+	// A closed iterator holds no position, so Block() reports misuse rather than
+	// reading through a released snapshot.
+	l.positioned = false
+	if l.it == nil {
+		return nil
+	}
+	if err := l.it.Close(); err != nil {
+		return fmt.Errorf("failed to close ledger iterator: %w", err)
 	}
 	return nil
```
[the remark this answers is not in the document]
cody-littley: fixed

PR #3808, opened 2026-07-24
Snapshot Engine

cody-littley:

Implements an in-memory MVCC engine that will allow us to take O(1) snapshots of flatKV state. This is a prerequisite for pipelining hashing.

This PR does not integrate this change, that will be a future PR.


sei-db/db_engine/snapshot/read_cache.go
```diff
@@ -0,0 +1,498 @@
...
+	entry := c.entryLocked(key, true)
+	entry.status = statusAvailable
+	entry.value = value
+	// Overwriting a failed entry is deliberate: this data comes from the engine's own retired
+	// writes, not from the failed DB read, and the original error has already been propagated.
+	entry.err = nil
+
+	size := uint64(len(key)) + uint64(len(value)) + c.overheadPerEntry
+	c.gcQueue.Push(key, size)
+}
```
bdchatham, line 515: **Read-failure latch is cleared by retirement (fix 1 of 2).**

The `statusFailed` latch is documented as permanent because a read that fails and later succeeds could fork the chain (L96). Retirement doesn't preserve it, this overwrites the entry and zeroes `err`, and `deleteRetiredLocked` below does the same. A read failure also never bricks the engine, so it stays live and reaches this path on its own. The net is that one key can return an error and then, a few blocks later, return a value in the same process, the sequence the latch exists to prevent.

This is the contained, fail-closed fix, keep a failed entry through retirement. The complete fix is to brick on the read failure itself, but that needs a brick hook plumbed through `newReadCache` and fired outside the shard lock, since `injectValue` runs under it and `Commit` already establishes `versionLock` then shard-lock ordering. I'd take this now and do brick-on-failure as a follow-up.

```suggestion
func (c *readCache) setRetiredLocked(key []byte, value []byte) {
	entry := c.entryLocked(key, true)
	if entry.status == statusFailed {
		// Never resurrect a permanently failed key (see statusFailed). The read error was
		// already propagated; letting a later retired write flip it back to available would
		// let the same key return an error and then a value within one process, which is the
		// fork risk the latch exists to prevent. The flushed value still lives in the DB, but
		// the latch keeps it from being served.
		return
	}
	entry.status = statusAvailable
	entry.value = value
	entry.err = nil

	size := uint64(len(key)) + uint64(len(value)) + c.overheadPerEntry
	c.gcQueue.Push(key, size)
}
```

sei-db/db_engine/snapshot/read_cache.go
```diff
@@ -0,0 +1,498 @@
...
+	entry.status = statusDeleted
+	entry.value = nil
+	// Overwriting a failed entry is deliberate: this tombstone comes from the engine's own
+	// retired writes, not from the failed DB read, and the original error has already been
+	// propagated.
+	entry.err = nil
+
+	size := uint64(len(key)) + c.overheadPerEntry
+	c.gcQueue.Push(key, size)
+}
```
bdchatham, line 531: **Same latch-clearing issue as `setRetiredLocked` above (fix 2 of 2).**

A retired tombstone must not clear a permanent read-failure latch either. A freshly created entry is `statusUnknown`, not `statusFailed`, so only genuinely-failed keys are held, and the accounting doesn't drift since failed entries were never in `gcQueue` to begin with.

```suggestion
func (c *readCache) deleteRetiredLocked(key []byte) {
	entry := c.entryLocked(key, false)
	if entry == nil {
		// Key is not in the cache, so nothing to do.
		return
	}
	if entry.status == statusFailed {
		// See setRetiredLocked: a permanent read-failure latch is never cleared by retirement.
		return
	}
	entry.status = statusDeleted
	entry.value = nil
	entry.err = nil

	size := uint64(len(key)) + c.overheadPerEntry
	c.gcQueue.Push(key, size)
}
```

sei-db/db_engine/snapshot/snapshot_engine_impl.go
```diff
@@ -0,0 +1,1117 @@
...
+		return fmt.Errorf("version (%d) not found", version)
+	}
+
+	if counter.referenceCount == 0 {
+		return fmt.Errorf("version (%d) has already been dropped", version)
+	}
+
+	if counter.referenceCount == 1 && counter.hash == nil {
+		return fmt.Errorf("version (%d) was fully released without first being hashed", version)
+	}
```
bdchatham, line 527: **Final release without a hash should brick.**

The interface calls releasing the final reservation on an unhashed snapshot a fatal error (`snapshot_engine.go` L151). This path returns an error but doesn't brick and doesn't decrement, so the snapshot stays pinned at refcount 1. Flush and retire both stop at the first referenced version, so that one stuck snapshot stalls flush and retire of every later version indefinitely, and memory grows behind it. The method already holds `versionLock`, so `brickLocked` is in-bounds here.

```suggestion
	if counter.referenceCount == 1 && counter.hash == nil {
		// Releasing the last reservation on an unhashed snapshot is fatal (see Snapshot). Latch
		// the failure so the engine stops rather than leaving the snapshot pinned at refcount 1,
		// which would stall flush and retire of every later version indefinitely.
		err := fmt.Errorf("version (%d) was fully released without first being hashed", version)
		c.brickLocked(err)
		return err
	}
```

sei-db/db_engine/snapshot/shard.go
```diff
@@ -0,0 +1,428 @@
...
+	outcome := s.cache.lookupLocked(key, updateLru)
+	s.lock.Unlock()
+
+	return s.cache.resolve(key, outcome)
+}
+
+// validateVersionLocked checks that the given version is within the valid range.
+//
+// The Locked postfix indicates that the caller must hold the shard lock.
+func (s *shard) validateVersionLocked(version uint64) error {
```
bdchatham, line 161: **A released-but-unretired version can be read for a silent wrong value.**

`doLifecycleWork` computes the retire set before flush marks `flushedToDisk`, so there's a window where a version is fully released and flushed but not yet retired, and `validateVersionLocked` still admits it. Holding a reservation makes this unreachable, flush breaks at the first referenced version and the retire transition is atomic under the shard lock, so a compliant caller can't hit it. The reason I'd still guard it, if a caller reads a released version whose old value has been evicted from the cache, the read falls through to the DB and returns the newer flushed value rather than an error. Refusing reads on `referenceCount == 0` versions would turn that into a detected halt.

sei-db/db_engine/snapshot/snapshot_engine_impl.go
```diff
@@ -0,0 +1,1117 @@
...
+
+	return snapshot, nil
+}
+
+// This method blocks if the lifecycle runner is not keeping up. It is assumed that the caller already holds the
+// versionLock. When this method returns, it will still hold the versionLock, but it may release and then
+// re-acquire versionLock internally as it awaits for the lifecycle runner to catch up.
+func (c *snapshotEngine) lifecycleBackpressureLocked() error {
+	for c.unflushedCount > c.config.MaxUnflushedVersions {
```
bdchatham, line 434: **Backpressure leaves three unbounded growth vectors, and the size metric can't see them.**

This bounds only hashed-but-unflushed versions (the scan breaks at the first unhashed version at L542). No ceiling on three vectors:
- a stalled hasher, unhashed versions accumulate and `Commit` never blocks
- a leaked reservation, retirement stalls behind the oldest still-referenced version
- `statusFailed`/`statusScheduled` cache entries, never pushed to `gcQueue` so unbounded by `MaxSize`

This matches the documented contract, the caller owns pacing. What concerns me operationally is that `sizeInfoLocked` (L496) reports only `gcQueue`, so an operator watching `snapshot_engine_size_bytes` sees the cache under budget while the process walks into an OOM. I'd want metrics for `len(versionMap)`, the unhashed and unreleased counts, and `len(entries)` against `gcQueue.GetCount()` before this is wired to anything that can stall independently of `Commit`.

sei-db/db_engine/snapshot/snapshot_engine_impl.go
```diff
@@ -0,0 +1,1117 @@
...
+			} else {
+				if err := batch.Set([]byte(key), value); err != nil {
+					return fmt.Errorf("flush failed to set key: %w", err)
+				}
+			}
+		}
+
+		// Len is non-negative, so the conversion is safe.
+		if uint64(batch.Len()) >= c.config.TargetBytesPerFlush { //nolint:gosec
+			commitErr := batch.Commit(types.WriteOptions{Sync: c.config.FlushSync})
```
bdchatham, line 1009: **Cross-batch crash durability depends on an unstated DB guarantee.**

A flush range can split across multiple `Batch.Commit` calls at version boundaries, and the default is `Sync=false`. Within one batch the diff and the `HashKey` write are atomic. Across two commits, the property that the on-disk hash always matches the on-disk data holds only if the backing DB guarantees prefix-consistent durability, which the `KeyValueDB`/`Batch` contract in `types.go` doesn't state. On a lazy-fsync or reordering backend, a crash could leave `hash(V6)` durable with V1 through V3 missing, and `InitialHash` on restart returns a hash that doesn't match the recovered state. I'd state the prefix-durability requirement in the DB interface and either fsync the terminal batch of a range or write the hash only after the data batches are durable.

sei-db/db_engine/snapshot/snapshot_impl.go
```diff
@@ -0,0 +1,97 @@
...
+func (s *snapshotImpl) SetHash(hash []byte) error {
+	return s.parentEngine.SetSnapshotHash(s.version, hash)
+}
+
+func (s *snapshotImpl) AwaitHash(ctx context.Context) ([]byte, error) {
+	return s.parentEngine.AwaitSnapshotHash(ctx, s.version)
+}
+
+func (s *snapshotImpl) Iterator() Iterator {
+	iter, err := s.parentEngine.requestIterator(s.version)
```
bdchatham: **`Iterator()` doesn't hold the reservation its own doc requires.**

`requestIterator`'s doc says the caller, typically this method, must hold a reservation across the call (impl L684). This calls it with no `Reserve`/`Release`, and the public `Snapshot.Iterator` doc doesn't state the precondition either. Not a live bug, construction is serialized on the lifecycle goroutine and a retired version yields a clean `errIterator`, but the contract contradicts itself. I'd either take the reservation here or state the requirement on the public method.

sei-db/db_engine/snapshot/snapshot_engine_config.go
```diff
@@ -0,0 +1,123 @@
...
+	FlushSync bool
+}
+
+// Default configuration for a production snapshot engine.
+func DefaultSnapshotEngineConfig() *SnapshotEngineConfig {
+	return &SnapshotEngineConfig{
+		ShardCount:                   8,
+		MaxSize:                      unit.GB / 2,
+		EstimatedOverheadPerEntry:    256,
+		MetricsEnabled:               true,
```
bdchatham, line 75: **`DefaultSnapshotEngineConfig()` can't construct an engine.**

`MetricsEnabled` is true with an empty `MetricsName`, and `HashKey` is empty. `Validate` rejects both (L107 and L119), so `NewSnapshotEngine` with the production default always returns `invalid snapshot engine config`. The only caller is `DefaultTestSnapshotEngineConfig`, which patches both fields and disables metrics. Clean error rather than a crash, but a documented production default that can't build an engine is a footgun. I'd populate valid defaults or make the two required fields explicit.

sei-db/db_engine/snapshot/snapshot_engine_impl.go
```diff
@@ -0,0 +1,1117 @@
...
+	if firstVersion >= lastVersion {
+		// Nothing to retire. Callers in doLifecycleWork pass the same first/last when
+		// no version is retirement-eligible.
+		return nil
+	}
+
+	for i, shard := range c.shards {
+		err := shard.DropVersions(firstVersion, lastVersion)
+		if err != nil {
+			return fmt.Errorf("failed to drop versions from shard %d %w", i, err)
```
bdchatham: **Missing `: ` separator in the error string.** Renders as `shard 0failed to...`; every other wrap in the file uses `": %w"`.

```suggestion
			return fmt.Errorf("failed to drop versions from shard %d: %w", i, err)
```

sei-db/db_engine/snapshot/shard.go
```diff
@@ -0,0 +1,428 @@
...
+	"github.com/sei-protocol/sei-chain/sei-db/db_engine/types"
+	"github.com/sei-protocol/sei-chain/sei-db/proto"
+)
+
+// A single shard of a SnapshotEngine. The shard owns the MVCC layer: versioned in-memory data
+// awaiting flush, per-version diffs, and version bookkeeping. Reads that miss the versioned data
+// fall through to the shard's read-through DB cache (see readCache).
+type shard struct {
+	// A lock to protect the shard's data. Shared with the read cache (see the cache field).
+	lock sync.Mutex
```
yzang2019, line 22: Propose to add a TODO to make it RW lock if we hit this as a bottleneck

sei-db/common/structures/deque.go
```diff
@@ -0,0 +1,238 @@
...
+	"iter"
+	"math/bits"
+)
+
+const minDequeCapacity = 8
+
+// Deque is a generic double-ended queue backed by a growable circular buffer. It supports amortized O(1)
+// push/pop/peek at both ends and O(1) random access by index (including Python-style negative
+// indexing). Not thread safe.
+type Deque[T any] struct {
```
yzang2019, line 14: Just curious why not use any 3rd party Deque library such as https://pkg.go.dev/github.com/gammazero/deque
Any specific feature we need from the custom deque struct?

sei-db/db_engine/snapshot/shard.go
```diff
@@ -0,0 +1,428 @@
...
+	return diffs, nil
+}
+
+// materializeOverridesAtVersion returns every in-memory override visible to this shard
+// at the given version. The result is unsorted.
+//
+// Keys whose most recent entry is at a version greater than version are
+// skipped (the key did not yet exist in the cache at that version).
+func (s *shard) materializeOverridesAtVersion(version uint64) ([]kvPair, error) {
+	s.lock.Lock()
```
yzang2019, line 411: buildIterator calls it per shard; each iterates the entire versionedData map under s.lock, blocking all reads/writes to that shard for the scan. It runs on the lifecycle goroutine, so it also stalls flush/retire progress. Bounded by working-set size. This might become a real perf concern. A snapshot of deque heads, or releasing the lock during the sort/merge, would probably help?

sei-db/db_engine/snapshot/snapshot_engine_config.go
```diff
@@ -0,0 +1,123 @@
...
+	// are not individually fsync'd: on a hard OS/power crash the most recent unsynced flushes may be
+	// lost (but the DB is not corrupted), and crash durability is instead provided by an upstream
+	// fsync'd WAL / block replay. Set true for deployments that want per-flush durability regardless.
+	FlushSync bool
+}
+
+// Default configuration for a production snapshot engine.
+func DefaultSnapshotEngineConfig() *SnapshotEngineConfig {
+	return &SnapshotEngineConfig{
+		ShardCount:                   8,
```
yzang2019, line 71: Why only 8 shards to start with? Wouldn't that lead to high contention?

sei-db/db_engine/snapshot/snapshot_engine_config.go
```diff
@@ -0,0 +1,123 @@
...
+	if c.MetricsEnabled && c.MetricsScrapeIntervalSeconds <= 0 {
+		return fmt.Errorf("MetricsScrapeIntervalSeconds must be positive when MetricsEnabled is true")
+	}
+	if c.MaxUnflushedVersions == 0 {
+		return fmt.Errorf("MaxUnflushedVersions must be greater than 0")
+	}
+	if c.TargetBytesPerFlush == 0 {
+		return fmt.Errorf("TargetBytesPerFlush must be greater than 0")
+	}
+	if c.HashKey == "" {
```
yzang2019, line 121: The production default sets MetricsEnabled: true with MetricsName: "", and HashKey: "". Validate() rejects both (MetricsName must be non-empty when MetricsEnabled is true, HashKey must be non-empty). So NewSnapshotEngine(DefaultSnapshotEngineConfig(), ...) errors out of the box — only DefaultTestSnapshotEngineConfig() pass validation

sei-db/db_engine/snapshot/snapshot_engine_impl.go
```diff
@@ -0,0 +1,1117 @@
...
+// already prevents this race today, but the order is a cheap belt to go with
+// the suspenders.)
+//
+// The returned iterator excludes the engine's metadata hash key (see
+// SnapshotEngineConfig.HashKey). The value stored under that key in the DB is
+// the most recently *flushed* hash, which is generally stale relative to this
+// snapshot and varies with flush timing; exposing it would let a consumer pair
+// data-at-V with hash-at-W. Consumers that need the snapshot's hash should use
+// Snapshot.AwaitHash, which is guaranteed to match the iterated data.
+func (c *snapshotEngine) buildIterator(version uint64) (Iterator, error) {
```
yzang2019: EVM execution has no concept of iteration. So not really sure if we need to support this iteration for snapshot, and whether there's anything we can simplify when removing iteration.

sei-db/common/structures/deque.go
cody-littley: My thought process is this. Potentially worth a short chat during our next parking lot.

- Adding additional third party libraries increases our attack surface to supply chain attacks.
    - Although the risk from any particular library is small, I prefer to avoid third party libraries unless they are saving us a lot of effort.
    - My default stance when working on a blockchain project is "extreme paranoia". I feel like this is a defensible world view. ;)
    - The engineering effort to properly vet a library and future library updates is nontrivial, and we have to continue paying that tax forever. The engineering effort to vet our own impl is a one time cost.
- The deque is super simple, one of those things an LLM can one shot in 3 minutes. We pay very little engineering cost to maintain this data structure.
- This deque is used by highly performance sensitive code. If in the future we need to fine tune low level things like allocations or add detailed metrics, it's much easier to do so when the source code is in the repo.

sei-db/db_engine/snapshot/snapshot_engine_metrics.go
```diff
@@ -0,0 +1,169 @@
...
+		metric.WithDescription("Total number of cache misses"),
+		metric.WithUnit("{count}"),
+	)
+	missLatency, _ := meter.Float64Histogram(
+		"snapshot_engine_miss_latency",
+		metric.WithDescription("Time taken to resolve a cache miss from the backing store"),
+		metric.WithUnit("s"),
+		metric.WithExplicitBucketBoundaries(metrics.LatencyBuckets...),
+	)
+	snapshotPhaseTimer := metrics.NewPhaseTimer(meter, fmt.Sprintf("snapshot_engine_snapshot_%s", cacheName))
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/db_engine/snapshot/snapshot_impl.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		return &errIterator{err: fmt.Errorf("failed to build snapshot iterator: %w", err)}
+	}
+	return iter
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/db_engine/litt/util/random_access_deque.go
```diff
@@ -1,481 +1,14 @@
...
 
-import "math"
+import "github.com/sei-protocol/sei-chain/sei-db/common/structures"
 
-// The minimum initial capacity of a RandomAccessDeque.
-const minimumInitialCapacity = 32
+// RandomAccessDeque is a backward-compatibility alias for structures.Deque. The deque
+// implementation moved to the shared structures package; this alias is retained so existing
+// callers compile unchanged. Prefer structures.Deque directly in new code.
+type RandomAccessDeque[T any] = structures.Deque[T]
```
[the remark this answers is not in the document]
cody-littley: Don't care, those methods aren't currently used. I wrote both, I'm just unifying disparate impls.

sei-db/db_engine/snapshot/read_cache.go
cody-littley: I reworked the "brick" path and updated godocs.

The purpose of the DB bricking itself is to make it operationally difficult to keep using a database after there is evidence of an unrecoverable error. The specific contract of "when does the database begin to act like a brick" is ok to be loosely defined. The goal is to ensure the system tears itself down, not to specify the exact chain of events that leads to that teardown. Putting too strict of a contract on the teardown sequence ends up requiring additional synchronization primitives, which is costly in this part of the code in particular.

sei-db/db_engine/snapshot/read_cache.go
cody-littley: see comment above, this has been reworked

sei-db/db_engine/snapshot/snapshot_engine_impl.go
cody-littley: fixed

sei-db/db_engine/snapshot/snapshot_engine_impl.go
cody-littley: > a stalled hasher, unhashed versions accumulate and Commit never blocks

It's not the job of this layer to apply hashing backpressure. Consensus is the consumer of the hash, and eventually consensus should stop producing new blocks if it can't get an older block's hash.

> a leaked reservation, retirement stalls behind the oldest still-referenced version

Reservation leak is terminal (in latest commit), and not solvable via backpressure.

> statusFailed/statusScheduled cache entries, never pushed to gcQueue so unbounded by MaxSize

Read backpressure should be enforced by the external context. e.g. by limiting the maximum number of executor threads, limiting the maximum number of concurrent RPCs, etc.

sei-db/common/structures/deque.go
yzang2019: make sense to me

sei-db/db_engine/snapshot/snapshot_engine_impl.go
cody-littley: added documentation that specifies that commits must become crash durable in write order

sei-db/db_engine/snapshot/snapshot_impl.go
cody-littley: doc was out of date, fixed

sei-db/db_engine/snapshot/snapshot_engine_config.go
cody-littley: add some required params so that default config returned by method is now viable

sei-db/db_engine/snapshot/snapshot_engine_impl.go
cody-littley: fixed

sei-db/db_engine/snapshot/shard.go
cody-littley: added

sei-db/db_engine/snapshot/shard.go
cody-littley: I agree this has potential to be a performance hotspot if iteration ever becomes a load-bearing workload. Unfortunately, the solutions for fixing this problem are potentially even worse than the original problem, if we aren't careful.

We don't currently hold locks when we do the sorting, just when we iterate the map to extract the values from the shard. Taking snapshots deque heads at specific versions will be quite expensive if we do it, essentially equally as expensive as the implementation are now. Would make the transition from block-to-block `O(size of working set)`. Currently, transitioning to the next block is `O(1)`, and we force the expensive `O(size of working set)` operation onto the build-iterator workflow.

My hope is that iterators aren't really a common thing once we fully ditch cosmos and live fully in an eth world. If that doesn't end up being a valid assumption, my suggestion is that we delay optimizing this code until we've got a workload to optimize it against.

sei-db/db_engine/snapshot/snapshot_engine_config.go
cody-littley: I chose 8 via trial and error. More than 8 and `cryptosim` performance stopped getting better. Lots more than 8 and it gets much slower.

This is an area I will revisit once I get this fully integrated. IMO, this is something that can really only be dialed in via direct experimentation.

sei-db/db_engine/snapshot/snapshot_engine_config.go
cody-littley: I updated this. The default config builder now has a required param for the name and hash key. 

```
DefaultSnapshotEngineConfig(metricsName string, hashKey string)
```

sei-db/db_engine/snapshot/snapshot_engine_impl.go
cody-littley: This is a really good suggestion. I dropped iteration from snapshots, and now only support iteration on the mutable state, which will satisfy cosmos. Resulted in very large simplifications.

sei-db/db_engine/snapshot/shard.go
cody-littley: Will not fix, reading a released version is explicitly a violation of contract.

sei-db/db_engine/snapshot/snapshot_engine_impl.go
```diff
@@ -0,0 +1,1136 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create snapshot iterator: %w", err)
+	}
+
+	// Block writes only now that construction has fully succeeded, so a failed construction cannot
+	// leave the engine permanently unwritable.
+	for _, s := range c.shards {
+		s.iteratorOpened()
+	}
+	return &writeBlockingIterator{Iterator: iter, engine: c}, nil
```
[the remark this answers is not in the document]
cody-littley: Behavior is not the problem, but documentation was over promising. Updated the godoc to properly capture intended contract.

```
	// An iterator must be closed before the engine is next written: writing to the engine while an
	// iterator is open — via Set, Delete, BatchSet, or Commit — is illegal. The engine makes a
	// best-effort attempt to detect such a write and return an error from it, but that detection is
	// inherently race prone and must never be relied upon; it exists to catch the mistake in a
	// sequential caller, not to synchronize a concurrent one. A leaked iterator leaves the engine
	// permanently unwritable.
```

sei-db/db_engine/snapshot/snapshot_engine_impl.go
```diff
@@ -0,0 +1,1136 @@
...
+
+	return diff, nil
+}
+
+func (c *snapshotEngine) Iterator() (Iterator, error) {
+	// Overrides first, DB iterator second, and the order is load-bearing: a concurrent flush+retire
+	// that moved data out of versionedData and into the DB between the two steps would drop those
+	// keys entirely if the DB snapshot were taken first. In this order the same race can only yield a
+	// key twice, which the merge resolves in favor of the override.
+	overrides, err := c.materializeCurrentOverrides()
```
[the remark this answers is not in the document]
cody-littley: Updated the godoc:

```
	// An iterator must be closed before the engine is next written: writing to the engine while an
	// iterator is open — via Set, Delete, BatchSet, or Commit — is illegal. The engine makes a
	// best-effort attempt to detect such a write and return an error from it, but that detection is
	// inherently race prone and must never be relied upon; it exists to catch the mistake in a
	// sequential caller, not to synchronize a concurrent one. A leaked iterator leaves the engine
	// permanently unwritable.
```

sei-db/db_engine/snapshot/read_cache.go
```diff
@@ -0,0 +1,548 @@
...
+			}
+			continue
+		}
+		if result.value != nil {
+			results[pending[i].key] = result.value
+		}
+	}
+
+	c.metrics.reportCacheMissLatency(time.Since(startTime))
+	go c.bulkInjectValues(pending)
```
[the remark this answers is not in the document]
cody-littley: nah, way to late in the process to be suggesting new nit fixes

sei-db/db_engine/snapshot/read_cache.go
```diff
@@ -0,0 +1,548 @@
...
+
+	size := uint64(len(key)) + uint64(len(value)) + c.overheadPerEntry
+	c.gcQueue.Push(key, size)
+}
+
+// Delete a retired value.
+//
+// The Locked postfix indicates that the caller must hold the shared lock.
+func (c *readCache) deleteRetiredLocked(key []byte) {
+	entry := c.entryLocked(key, false)
```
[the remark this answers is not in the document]
cody-littley: will not fix

sei-db/db_engine/snapshot/snapshot_engine_impl.go
```diff
@@ -0,0 +1,1136 @@
...
+			closeErr := batch.Close()
+			batch = nil
+			if commitErr != nil {
+				return fmt.Errorf("flush failed to commit batch: %w", commitErr)
+			}
+			if closeErr != nil {
+				return fmt.Errorf("flush failed to close batch: %w", closeErr)
+			}
+			c.versionLock.Lock()
+			c.unflushedCount -= versionsInBatch
```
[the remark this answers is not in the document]
cody-littley: added sanity check

sei-db/db_engine/snapshot/snapshot_engine_impl.go
```diff
@@ -0,0 +1,1136 @@
...
+	c.lifecycleExit <- struct{}{}
+	<-c.lifecycleExited
+
+	// Release everyone blocked on the engine's future: AwaitHash, AwaitFlush, backpressured
+	// Snapshot callers, and reads still awaiting results. The cancel happens under versionLock
+	// because backpressure waiters re-check the context under that lock before parking on the
+	// cond; a lockless cancel could slip between a waiter's check and its Wait, losing the
+	// wakeup.
+	c.versionLock.Lock()
+	c.cancel()
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/db_engine/snapshot/read_cache.go
```diff
@@ -0,0 +1,548 @@
...
+// for a stored empty value would silently turn that key into a tombstone.
+func (c *readCache) readFromDB(key []byte) (value []byte, found bool, err error) {
+	val, err := c.db.Get(key)
+	if err != nil {
+		if errors.Is(err, errorutils.ErrNotFound) {
+			return nil, false, nil
+		}
+		return nil, false, fmt.Errorf("failed to read value from database: %w", err)
+	}
+	return val, true, nil
```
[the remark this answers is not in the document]
cody-littley: nah

sei-db/db_engine/snapshot/read_cache.go
```diff
@@ -0,0 +1,548 @@
...
+			}
+			continue
+		}
+		if result.value != nil {
+			results[pending[i].key] = result.value
+		}
+	}
+
+	c.metrics.reportCacheMissLatency(time.Since(startTime))
+	go c.bulkInjectValues(pending)
```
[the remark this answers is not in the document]
cody-littley: updated godoc to match what actually happens

sei-db/db_engine/snapshot/shard.go
```diff
@@ -0,0 +1,503 @@
...
+	// the lock.
+	if err := s.writableLocked(); err != nil {
+		return err
+	}
+	for i := range entries {
+		if entries[i].Delete {
+			// A delete is stored as a nil-valued (tombstone) entry at the current version.
+			s.setLocked(entries[i].Key, nil)
+		} else {
+			s.setLocked(entries[i].Key, entries[i].Value)
```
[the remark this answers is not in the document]
cody-littley: will not fix

sei-db/db_engine/snapshot/snapshot_engine_impl.go
```diff
@@ -0,0 +1,1160 @@
...
+	counter, ok := c.versionMap[version]
+	if !ok {
+		return fmt.Errorf("version (%d) not found", version)
+	}
+
+	if counter.hash != nil {
+		return fmt.Errorf("hash already set for version %d", version)
+	}
+
+	counter.hash = hash
```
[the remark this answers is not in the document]
cody-littley: will not fix

sei-db/db_engine/litt/util/random_access_deque.go
```diff
@@ -1,481 +1,14 @@
...
-	} else if cmp < 0 {
-		// value < element, so missing value should go to the left of it
-		return left, false
-	}
-	// value > element, so missing value should go to the right of it
-	return left + 1, false
+// NewRandomAccessDeque creates a structures.Deque with the given initial capacity. Retained as a
+// compatibility shim for the former util deque constructor.
+func NewRandomAccessDeque[T any](initialCapacity uint64) *structures.Deque[T] {
+	return structures.NewDequeWithCapacity[T](int(initialCapacity)) //nolint:gosec // capacity fits int
```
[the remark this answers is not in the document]
cody-littley: updated comment

yzang2019 (approved):

PR #3815, opened 2026-07-27
Add GetSubrange for LittDB

yzang2019:
## Summary

Adds a sub-range read primitive to LittDB and a block-store primitive built on top of it, so a single transaction can eventually be read out of a stored block (or a receipt's fixed-position index field) without paying for a full-value read + full protobuf decode.

- **`litt.Table.GetSubrange(key, offset, length)`** — reads only `[offset, offset+length)` of a value instead of the whole thing.
- **`littblock.(*blockDB).ReadBlockSubrange(n, offset, length)`** — reads a byte range out of a stored block's marshalled body via `GetSubrange`, skipping the full-block disk read and protobuf unmarshal.

## `GetSubrange` implementation

Threaded through every layer of the read stack:

| Layer | Change |
|---|---|
| `litt.Table` (interface) | New `GetSubrange` method, fully documented (semantics, bounds, caching note). |
| `dbcache.cachedTable` | Consults the write and read caches the same way `Get` does, slicing the requested range out of a cached value; falls through to the base table on a miss. Never *populates* either cache (rationale below). |
| `disktable.DiskTable` | Slices the unflushed in-memory value when present; otherwise resolves the key via the keymap and delegates to the segment. |
| `segment.Segment.ReadSubrange` | The actual I/O optimization: for an **uncompressed** segment, seeks to `value.Offset()+offset` and reads exactly `length` bytes (cost scales with `length`, not the full value). For a **compressed** segment, a compressed blob can't be sliced on disk, so it falls back to `Read` (full read + decompress) and slices the plaintext — still correct, just without the I/O win. |

### Cache policy: consult, never populate

- **Populating would be unsafe.** Both caches are keyed by the full key and hold the full value, so storing a sub-range under that key would hand a later `Get` a partial value it is entitled to treat as the whole thing.
- **Consulting is safe**, because every path agrees that a key's offsets are relative to that key's own logical value: `Put` stores `value[sk.Offset:sk.Offset+sk.Length]` under a secondary key, the base table addresses that secondary at the same aliased region on disk, and the read cache holds whatever `Get` returned. A hit therefore slices exactly the bytes the base table would have read.
- **Consulting is also strictly cheaper.** A hit means the value is already materialized in memory, so slicing costs no I/O, where the base table would pay a keymap lookup plus a disk read. Bypassing the caches would make `GetSubrange` slower than plain `Get` for precisely the hot, recently-written values a sub-range read is meant to serve.

Because hits come from the same caches `Get` uses, the two agree on whether a key exists — including for a key already reclaimed from the base table by GC or TTL expiry but still cache-resident.

### Returned slices are capacity-capped

A sub-range may alias a buffer holding the whole value — the live unflushed entry, or a cache entry shared by every later reader. Every path therefore returns a three-index slice, so `cap == len` and an `append` by the caller allocates instead of overwriting whatever follows the range in that shared buffer. "NOT safe to mutate" alone didn't cover `append`, so the guarantee is now stated in the `litt.Table` contract and asserted in the tests.

Caching sub-ranges *themselves*, under specially structured keys, would speed up misses as well, but that adds complexity we don't yet know we need; it's documented as a revisit-if-profiling-says-so. Worth noting that every table currently runs with caching disabled (`DefaultTableConfig` sets both cache sizes to 0), so cache consultation only starts paying off once a caller enables a cache on the table it reads.

## `ReadBlockSubrange`

Added as a method on littblock's concrete `*blockDB` (alongside `ReadBlockByNumber`, `Close`, etc.) rather than on the `types.BlockDB` interface, since the offset space is defined by this implementation's serialization (`[version:1][GlobalBlockNumber:8][proto(Block)]`) and wouldn't be portable to `memblock` (which holds blocks by pointer with no byte layout). It mirrors `ReadBlockByNumber`'s watermark/pruning semantics (`types.ErrPruned` below the watermark, `None` for an absent block), and the watermark is checked before argument validation so retention state is what a caller observes for a pruned block either way.

**Offsets are body-relative.** `offset` is measured from the start of the marshalled `proto(Block)` body, and the method adds the fixed prefix itself. That is the frame a writer naturally computes in — it locates each transaction inside the marshalled body — so there is no conversion step to forget, and the framing stays a private detail of `littblock` instead of being baked into every offset persisted in another subsystem. The prefix addition is overflow-checked: an offset near the top of the `uint32` range would otherwise wrap and quietly return bytes from the start of the block.

**It returns bytes, not transactions.** The range is handed back verbatim with no framing validation, so the name says what it does; a tx-shaped wrapper belongs with whatever layer holds recorded transaction locations.

## Testing

- `disktable/disk_table_subrange_test.go` — table-driven across all four disk-table configurations (mem/pebble keymap × single/multi shard), before and after flush: full range, assorted sub-ranges, zero-length reads, composition with secondary keys, out-of-bounds errors, missing-key handling.
- `test/table_test.go` — exercises the `litt.Table` surface including the cached wrappers:
  - `TestGetSubrangeParity` pins that a sub-range read is indistinguishable between the cached and the base path — primary key, secondary key (offsets relative to its own aliased region), zero-length range, out-of-bounds against both the full value and the shorter alias, missing key — and that a sub-range read never poisons a later full `Get`. It also runs against a **cache-disabled** wrapper (zero-max-weight caches, i.e. the `DefaultTableConfig` production shape), because the existing cached builders keep every written value resident and so would only ever exercise hits, leaving the wrapper's fall-through to the base table untested.
  - `TestGetSubrangeServedFromCache` proves the caches are genuinely consulted, by seeding a cache entry for a key the base table has never seen so that only a cache hit can answer it. Also covers the reclaimed-but-still-cached case, where `GetSubrange` and `Get` must agree that the key exists, and asserts that appending to a returned range leaves the cached value untouched.
- `disktable/compression_test.go` (`TestGetSubrangeCompressed`) — proves the compressed fallback path is correct for both a compressible (S2-tagged) and an incompressible (`CompressionNone`-tagged) value on the same compressed segment.
- `littblock/litt_block_db_test.go` — round-trips a block's whole body, extracts real transactions by locating their range within the body, and verifies the overflow guard, out-of-bounds ranges, absent blocks, and `ErrPruned` below the watermark.

## Not in scope

Wiring the receipt-side producer — computing per-tx `(offset, length)` during `WriteBlock` and threading it into `ReceiptRecord` — is left for a follow-up. That work also depends on resolving the open block-numbering question (EVM height vs. `GlobalBlockNumber`).

Since `*blockDB` is unexported and `NewBlockDB` returns `types.BlockDB`, `ReadBlockSubrange` is currently reachable only from within the `littblock` package; the follow-up producer PR is where the exported access point (an interface to type-assert for, or a package-level helper) belongs.


cody-littley (approved):
Straight forward change, LGTM

PR #3820, opened 2026-07-28
v6.5.1 Stackoverflow patch

yzang2019:
### **Root cause**
nextReverse implements "skip backwards until you find a key visible at the target version" as self-recursion. Go has no tail-call optimization, so recursion depth grows one frame per skipped key, and the depth is bounded only by how many keys are in the scanned range.

The MVCC layout is what makes that depth enormous. Keys are stored as userKey || 0x00 || version sorted ascending by (userKey, version), so all versions of all keys are interleaved in one keyspace. To iterate backwards at version V, the iterator must physically walk over every key created after V. There is no index that lets it skip them in bulk.

The trigger was this request chain, reading up the stack:

debug_traceBlockByNumber(0x73d8857) → tracers.API.traceBlock → evmrpc.Backend.BlockByNumber → BaseApp.DeliverTx → wasm ExecuteContract → cgo into the wasm VM → cQueryExternal155 callback → QuerySmart → a contract smart-query that does a descending range scan.

0x73d8857 is block 121,473,111, while the node was at 222,485,912 — a block roughly 101 million heights (about two years) old, confirmed by the interleaved x/epoch/keeper logs showing current=2024-06-18T13:11:43. CacheMultiStoreWithVersion serves all historical queries from the SS PebbleDB store, so that scan ran against a two-year-old version. The contract's map had grown by ~1.77M keys since then, and because those keys sort after the old ones (monotonic IDs), they formed one contiguous run that a single logical step had to skip — 1.77M frames in one go.

### A second bug in the same code
While reproducing this I found the reverse iterator also silently returns wrong results. After SeekLT(nextKey@version+1) the code checks only the version it landed on, never whether it landed on the key it asked for. When a key has no visible version, the seek overshoots onto the previous key, and the recursion then skips that previous key entirely — even if it does have an older visible version.

Added a test against unmodified v6.5.1: keys keyA@5, keyB@30, keyA@100, keyB@200, reverse-iterating at version 50. It should yield keyB=B@30, keyA=A@5:

expected: []string{"keyB=B@30", "keyA=A@5"}
actual  : []string{"keyA=A@5"}
keyB vanishes because it also has a version 200. This directly violates the rule in evmrpc/AGENTS.md that "RPC responses for historical heights should never change" — historical debug_trace* and archive queries have been quietly dropping keys. It is confined to queries, though: CacheMultiStoreWithVersion is documented as "used to createQueryContext, abci_query or grpc query service", while consensus block execution reads from SC/memiavl, so consensus determinism was never at risk from this.

### DIFF
--- a/sei-db/db_engine/pebbledb/mvcc/iterator.go
+++ b/sei-db/db_engine/pebbledb/mvcc/iterator.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"math"
 	"sync"
 
 	"github.com/cockroachdb/pebble/v2"
@@ -21,10 +22,7 @@ var _ types.DBIterator = (*iterator)(nil)
 // in the provided domain for a given version. If a key has been written at the
 // provided version, that key/value pair will be iterated over. Otherwise, the
 // latest version for that key/value pair will be iterated over s.t. it's less
-// than the provided version. Note:
-//
-// - The start key must not be empty.
-// - Currently, reverse iteration is NOT supported.
+// than the provided version. The start key must not be empty.
 type iterator struct {
 	source             *pebble.Iterator
 	prefix, start, end []byte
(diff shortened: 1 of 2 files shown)


cody-littley (approved):

yzang2019:
Closed and moving this patch to v6.6 going forward

PR #3830, opened 2026-07-30
Enable seidroid xreview on sei-chain

bdchatham:
Enables **seidroid xreview** on sei-chain — the on-demand, sandbox-backed agentic review, a third seidroid capability alongside the `ai-review` / `ai-assist` this repo already runs.

Comment **`seidroid xreview`** on a PR and the `sei-droid` agent reviews it inside a managed omnigent Kubernetes sandbox (real `git`/`gh` toolchain, so it can build, test, and inspect the tree) and posts a structured verdict as one sticky comment.

## What this adds

One file, `.github/workflows/seidroid-xreview.yml` — a thin caller that `uses:` the uci reusable workflow, so the review logic lives in `sei-protocol/uci` and updates are a ref bump here:
- Pinned to the **uci v0.0.15 commit SHA** (`6590124`) on both `uses:` and `uci-ref` — a fixed ref, never a moving tag.
- Passes **only** `OMNIGENT_M2M_CLIENT_SECRET` (not `secrets: inherit`), so the cross-repo reusable workflow never receives this repo's release/ECR/GHCR/Slack secrets.
- A local `if:` gate mirrors the reusable workflow's trust check (PR comment, non-bot, `OWNER`/`MEMBER`/`COLLABORATOR`, command mentioned) so an unrelated or untrusted comment never dispatches it.

## Posting behavior

The verdict is **always posted** as a single sticky comment when a run produces one (a no-verdict run posts nothing). uci v0.0.15 removed the earlier dry-run mode: a review that posts nothing isn't useful, and dry-run was never a security boundary — the only real risk (a prompt injection driving the sandbox agent's vended `pull_requests: write` token) is independent of whether the driver posts.

## Safety

- **Trusted commenters only** — the guard admits `OWNER`/`MEMBER`/`COLLABORATOR`; an untrusted PR author cannot trigger it.
- Runs on the in-cluster `uci-default` runner; one session per trigger, torn down on exit; verdict is a single sticky upsert.
- **Operative constraint:** keep xreview on first-party / trusted content until a server-side shell gate is verified for the `sei-droid` agent. sei-chain PRs are first-party, so this is within scope.

sei-chain already has the seidroid app installed, `uci-default` runner-group access, and a rotator token scoped to it, and the `OMNIGENT_M2M_CLIENT_SECRET` secret is set — so xreview is live on merge.

### DIFF
--- a/.github/workflows/seidroid-xreview.yml
+++ b/.github/workflows/seidroid-xreview.yml
@@ -1,10 +1,8 @@
 name: seidroid xreview
-# Comment `seidroid xreview` (or `... --dry-run`) on a PR to run the on-demand,
-# sandbox-backed agentic review. Thin caller: the logic is the reusable workflow in
-# sei-protocol/uci; bump both refs below (to the next release's commit SHA) to adopt a
-# new xreview release. Runs dry-run until the SEIDROID_XREVIEW_DRY_RUN repo variable is
-# set to 'false' (read as `vars.SEIDROID_XREVIEW_DRY_RUN` by the reusable workflow, which
-# resolves it against this repo — there is no `with:` input for it here).
+# Comment `seidroid xreview` on a PR to run the on-demand, sandbox-backed agentic review.
+# Thin caller: the logic is the reusable workflow in sei-protocol/uci; bump both refs below
+# (to the next release's commit SHA) to adopt a new xreview release. The verdict is posted to
+# the PR as one sticky comment when a review produces one.
 on:
   issue_comment:
     types: [created]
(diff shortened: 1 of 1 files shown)


bdchatham:
Addressed / resolving:
- **`secrets: inherit` too broad:** good catch, especially here — switched to passing only `OMNIGENT_M2M_CLIENT_SECRET` explicitly, so the reusable workflow no longer receives this repo's release/ECR/GHCR/Slack secrets. The reusable workflow declares that single secret as `required`, so the explicit pass is sufficient.
- **Dry-run var not referenced in-file:** added a clause noting it's read as `vars.SEIDROID_XREVIEW_DRY_RUN` by the reusable workflow, resolved against this repo (no `with:` input here).
- **`MEMBER`/`COLLABORATOR` broader than ai-assist's team gate:** acceptable while dry-run is the default; a team-scoped `allowed-team` input on the reusable workflow is tracked as a before-real-post hardening (with the server-side shell gate) — won't flip `SEIDROID_XREVIEW_DRY_RUN=false` without it.
- **`contains` vs `startsWith`:** keeping `contains` deliberately — the reusable workflow's guard matches the command as its own line anywhere in the body, so `startsWith` would false-negative a valid invocation placed on a later line. The pre-filter must be a superset of the guard; the occasional wasted dispatch is cheap and the guard no-ops.

bdchatham:
All three are addressed / resolvable without a code change:

1. **Secret defined for sei-chain** — confirmed: `OMNIGENT_M2M_CLIENT_SECRET` is set as a repo secret on sei-chain (set when this was wired). And the reusable workflow declares it `required` and hard-fails on empty (a `${VAR:?}` guard before the mint), so a missing value surfaces loudly at the mint step, not silently as a sandbox auth error. Good flag that dropping `secrets: inherit` makes this load-bearing.
2. **`contains` unanchored** — deliberate. The reusable workflow's guard matches the command as its own whole line, so anchoring here (`startsWith`) would false-negative a valid invocation placed on a later line. The caller pre-filter must be a superset of the guard; a quoted/mentioned command just wastes a cheap dispatch that the guard then rejects.
3. **`COLLABORATOR` breadth** — accurate, and the uci-side guard uses the same author-association check (not a team check), so both admit read/triage collaborators. Acceptable for first-party/trusted content — removing dry-run doesn't widen this, since the real risk is content-based (injection via the sandbox agent's token), not trigger-based. A team-scoped `allowed-team` gate on the reusable workflow is tracked as a hardening to add before pointing xreview at less-trusted content; resolving against that follow-up.

yzang2019 (approved):

PR #3831, opened 2026-07-31
fix(autobahn): prevent block-sync livelock after restart catch-up

wen-coding:
## Summary
- Stop Autobahn validators from consuming catch-up `GetBlock` on the self dial (`RunSelfClient`), which always returns empty for missing local heights and can starve the contiguous prefix.
- Treat in-memory gap-fills as already fetched via `HaveBlock`, so the block fetcher does not keep re-requesting heights it already holds while a lower hole is open.
- Make the Autobahn major-upgrade post-downgrade height wait assert `WAIT_FOR_HEIGHT_PASS` explicitly, so a live stalled node cannot false-green as the expected upgrade panic.

## Test plan
- [x] `go test ./autobahn/... ./internal/autobahn/...` (253 passed)
- [x] Autobahn `TestUpgradeMajor` on local docker cluster
- [x] Autobahn `TestUpgradeMinor` on fresh local docker cluster
- [ ] CI Autobahn upgrade-module major/minor jobs

### DIFF
--- a/sei-tendermint/internal/autobahn/data/state.go
+++ b/sei-tendermint/internal/autobahn/data/state.go
@@ -529,18 +529,18 @@ func (s *State) TryBlock(n types.GlobalBlockNumber) (*types.Block, error) {
 	return s.blockFromDB(n)
 }
 
-// HaveBlock reports whether height n is already held in memory for catch-up:
-// either in the contiguous prefix (n < nextBlock) or as a gap-fill above
-// nextBlock. Unlike TryBlock, gap-fills count as present so the block fetcher
-// does not keep re-requesting heights it already holds while waiting for a
-// lower contiguous hole to close. Does not consult BlockDB.
-func (s *State) HaveBlock(n types.GlobalBlockNumber) bool {
+// NeedBlock reports whether catch-up still needs to fetch height n.
+// False when n is already past nextBlock (including heights pruned or
+// evicted from RAM) or an in-memory gap-fill is present. Unlike TryBlock,
+// gap-fills count as satisfied so the fetcher does not keep re-requesting
+// them while a lower contiguous hole is open. Does not consult BlockDB.
+func (s *State) NeedBlock(n types.GlobalBlockNumber) bool {
 	for inner := range s.inner.Lock() {
 		if n < inner.nextBlock {
-			return true
+			return false
 		}
 		_, ok := inner.blocks[n]
-		return ok
+		return !ok
 	}
 	panic("unreachable")
 }
--- a/sei-tendermint/internal/autobahn/data/state.go
+++ b/sei-tendermint/internal/autobahn/data/state.go
@@ -506,6 +506,21 @@ func (s *State) TryBlock(n types.GlobalBlockNumber) (*types.Block, error) {
 	return s.blockFromDB(n)
 }
 
+// HaveBlock reports whether height n is already held in memory for catch-up:
+// either in the contiguous prefix (n < nextBlock) or as a gap-fill above
+// nextBlock. Unlike TryBlock, gap-fills count as present so the block fetcher
+// does not keep re-requesting heights it already holds while waiting for a
+// lower contiguous hole to close. Does not consult BlockDB.
+func (s *State) HaveBlock(n types.GlobalBlockNumber) bool {
+	for inner := range s.inner.Lock() {
+		if n < inner.nextBlock {
+			return true
+		}
+		return inner.blocks[n] != nil
+	}
+	panic("unreachable")
+}
+
 // assembleGlobalBlock builds a GlobalBlock from a block and its covering QC.
 // Callers must supply non-nil b and fqc for height n. In-memory paths look up
 // maps only for heights still indexed there (including gap-fills); BlockDB
--- a/sei-tendermint/internal/autobahn/data/state.go
+++ b/sei-tendermint/internal/autobahn/data/state.go
@@ -539,7 +539,8 @@ func (s *State) HaveBlock(n types.GlobalBlockNumber) bool {
 		if n < inner.nextBlock {
 			return true
 		}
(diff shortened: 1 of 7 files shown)


codchen (approved):

philipsu522 (approved):
this looks good, i just wonder if we need to bound outgoing requests, given previously we had a pseudo-limit via holding the permit for an existing block? otherwise would we oom / overwhelm the node with this new (more performant) behavior?

sei-tendermint/internal/p2p/giga/service.go
```diff
@@ -83,19 +83,38 @@ func (x *Service) RunServer(ctx context.Context, server rpc.Server[API]) error {
...
+	return scope.Run(ctx, func(ctx context.Context, s scope.Scope) error {
+		x.spawnConsensusClientStreams(ctx, s, client)
+		return nil
+	})
+}
+
+// spawnConsensusClientStreams starts the streams shared by peer and self
+// clients. Block-sync streams (FullCommitQCs / GetBlock) are added only by
+// RunClient so the self dial cannot drift back into catch-up fetches.
+func (x *Service) spawnConsensusClientStreams(ctx context.Context, s scope.Scope, client rpc.Client[API]) {
```
pompon0: don't take scope as an arg, just make it a run...() command and create another scope internally.

pompon0 (commented):
The mismatch between TryBlock and fetcher semantics was definitely worth fixing, thank you. Special handling of loopback-connection is something that I'd really like to avoid, because it obfuscates deeper problems: like here - we fix the specific case where node asks itself about the block that it doesn't have, but the problem persists if we have a peer which also doesn't have the block. Indeed the fetching strategy has a lot of room for improvements, but let's make them uniform.

sei-tendermint/internal/autobahn/data/state.go
```diff
@@ -529,6 +529,22 @@ func (s *State) TryBlock(n types.GlobalBlockNumber) (*types.Block, error) {
 	return s.blockFromDB(n)
 }
 
+// HaveBlock reports whether height n is already held in memory for catch-up:
+// either in the contiguous prefix (n < nextBlock) or as a gap-fill above
+// nextBlock. Unlike TryBlock, gap-fills count as present so the block fetcher
+// does not keep re-requesting heights it already holds while waiting for a
+// lower contiguous hole to close. Does not consult BlockDB.
+func (s *State) HaveBlock(n types.GlobalBlockNumber) bool {
```
pompon0: name is misleading, since the block could have been already pruned. One option would be to invert the semantics and call it "NeedBlock", which does not imply the block is accessible (pruned/ahead of nextBlock)

sei-tendermint/internal/p2p/giga/service.go
```diff
@@ -83,19 +83,38 @@ func (x *Service) RunServer(ctx context.Context, server rpc.Server[API]) error {
 
 func (x *Service) RunClient(ctx context.Context, client rpc.Client[API]) error {
 	return scope.Run(ctx, func(ctx context.Context, s scope.Scope) error {
-		s.Spawn(func() error { return x.clientPing(ctx, client) })
-		s.Spawn(func() error { return x.clientConsensus(ctx, client) })
+		x.spawnConsensusClientStreams(ctx, s, client)
```
pompon0: until we implement a robust retry strategy (please add a TODO), for now it would be fine to add a flag to RunClient deciding whether to run GetBlock or not, as a minimal asymmetry addressing the issue at hand.

ranchalp (approved):

wen-coding:
> this looks good, i just wonder if we need to bound outgoing requests, given previously we had a pseudo-limit via holding the permit for an existing block? otherwise would we oom / overwhelm the node with this new (more performant) behavior?

@philipsu522 we still have MaxConcurrentBlockFetches=100 cap the in flight requests, there is just less waste.

sei-tendermint/internal/p2p/giga/service.go
wen-coding: done

sei-tendermint/internal/autobahn/data/state.go
wen-coding: done

sei-tendermint/internal/p2p/giga/service.go
wen-coding: done

sei-tendermint/internal/p2p/giga/service.go
```diff
@@ -81,12 +81,24 @@ func (x *Service) RunServer(ctx context.Context, server rpc.Server[API]) error {
 	})
 }
 
-func (x *Service) RunClient(ctx context.Context, client rpc.Client[API]) error {
+func (x *Service) RunClient(ctx context.Context, client rpc.Client[API], getBlock bool) error {
+	// TODO: implement a uniform robust GetBlock peer-selection / retry strategy
+	// so connections that lack a height (including self) do not need a separate
+	// getBlock=false path to avoid starving the shared fetch queue.
+	return scope.Run(ctx, func(ctx context.Context, s scope.Scope) error {
+		s.Spawn(func() error { return x.runConsensusClientStreams(ctx, client) })
```
pompon0: nit: inline back runConsensusClientStreams

sei-tendermint/internal/p2p/giga/service.go
```diff
@@ -81,12 +81,24 @@ func (x *Service) RunServer(ctx context.Context, server rpc.Server[API]) error {
...
+		if getBlock {
+			s.Spawn(func() error { return x.clientStreamFullCommitQCs(ctx, client) })
```
pompon0: clientStreamFullCommitQCs can be unconditional, no need for asymmetry here

pompon0 (approved):

PR #3840, opened 2026-08-03
fix(seidb): stop evm-logical-digest replay mode from mutating a live memiavl

blindchaser:
## Summary

Make memIAVL replay reads fail closed instead of repairing a live changelog or copying it. Read-only opens now take an immutable view of the existing WAL segment files. A torn tail, an in-progress recovery marker, a pruned coverage gap, or a short replay returns an actionable retry error without changing the source.

- `sei-db/wal/readonly.go`, `sei-db/wal/changelog.go`: add a changelog reader that opens segment files read-only, indexes the bytes present at open time, retains file descriptors for a stable point-in-time view, and rejects every mutation API. It never invokes tidwall's writable open or recovery path.
- `sei-db/state_db/sc/memiavl/db.go`: route `Options.ReadOnly` through the immutable WAL reader while keeping writable opens on the existing auto-repair path. Read-only opens reject torn tails, interrupted recovery markers, snapshot-to-WAL coverage gaps, and explicit targets they cannot reach. Failed opens close the tree, WAL, and exclusive lock from one cleanup path.
- `sei-db/tools/cmd/seidb/operations/evm_logical_digest.go`: keep replay mode on the live read-only view without allocating a changelog clone. Failures state that the live WAL was not modified and tell the operator to rerun; persistent corruption requires an offline repair.

## Test plan

- `sei-db/wal/readonly_test.go`: replay a valid WAL without mutation; reject writes and truncations; preserve a point-in-time view while the writer appends; reject a torn tail and both recovery-marker forms without modifying them; keep empty or missing directories unchanged; stress concurrent append and front-prune operations against repeated read-only opens.
- `sei-db/state_db/sc/memiavl/db_test.go`: verify read-only open leaves a torn tail untouched while writable open still repairs it; reject a WAL pruned past the selected snapshot; reject a clean WAL that ends before the requested target; verify a failed writable open releases its exclusive lock.
- `sei-db/tools/cmd/seidb/operations/memiavl_open_test.go`: verify replay mode reports that the live WAL was not modified and asks the operator to rerun.
- `go test -race ./sei-db/wal ./sei-db/state_db/sc/memiavl ./sei-db/tools/cmd/seidb/operations`
- `golangci-lint run ./sei-db/wal/... ./sei-db/state_db/sc/memiavl/... ./sei-db/tools/cmd/seidb/operations/...`


blindchaser:
@seidroid review

blindchaser:
Superseded by #3983, which starts from the latest main branch with one clean implementation commit.

PR #3844, opened 2026-08-03
Replace legacy WAL

cody-littley:

Replaces legacy autobahn WAL with new implementation.


sei-tendermint/internal/autobahn/consensus/persist/wal.go
```diff
@@ -1,211 +1,123 @@
...
+	"github.com/sei-protocol/sei-chain/sei-db/common/unit"
+	"github.com/sei-protocol/sei-chain/sei-db/seiwal"
 )
 
+// targetFileSize is the size a WAL file may reach before it is sealed and a fresh one is opened.
+//
+// This sits far below seiwal's default because pruning only deletes whole sealed files and never the
+// file currently being written, so the volume of pruned-but-still-present records scales with this
+// value — and autobahn keeps one WAL per lane, multiplying that cost by the committee size.
+const targetFileSize = 4 * unit.MB
```
wen-coding: We didn't expect these WAL files to be too big. Normally they are consumed quickly in new proposals and then pruned, which is why we didn't consider cutting them in chunks. What is the main reason we want a max file size now?

sei-tendermint/internal/autobahn/consensus/persist/blocks.go
```diff
@@ -17,95 +18,87 @@ var logger = seilog.NewLogger("tendermint", "internal", "autobahn", "consensus",
 
 const blocksDir = "blocks"
 
+// blocksWALName identifies every lane WAL in metrics. All lanes share it so their counters aggregate;
+// a per-lane name would make a metric label out of a lane's public key.
+const blocksWALName = "autobahn_blocks"
```
wen-coding, line 27: nit: maybe rename so it reads more like a metrics label.

sei-tendermint/internal/autobahn/consensus/persist/commitqcs.go
```diff
@@ -147,12 +169,18 @@ func (cp *CommitQCPersister) MaybePruneAndPersist(
 				return err
 			}
 		}
-		fn, hasFn := afterEach.Get()
```
wen-coding, line 150: We originally added afterEach so that avail gets notified as soon as each commitqc is persisted. Guess we don't need it if you flush all then notifies.

@pompon0 hmm I'm suddenly confused, under which case will we have more than one commitqcs in this batch? I think we won't enqueue the next commitqc until the previous one is persisted?

pompon0: we don't participate in consensus until the previous commitqc is persisted. But it does not mean that we won't receive multiple CommitQCs from peers, which will go into a single batch. For commitQCs specifically we can flush once after all commitQCs from the batch are persisted, and only then notify: consensus does not need to be notified about non-last persisted commitQC, because it doesn't make sense to participate in consensus instance which already has finished. For LaneBlocks however we want to have the notification ASAP, so that lane blocks can be included in consensus.

sei-tendermint/internal/autobahn/consensus/state_test.go
```diff
@@ -272,11 +272,15 @@ func TestVoteTimeoutPrepareQC_PersistedRestart(t *testing.T) {
 	pqc0 := makePrepareQC(keys, types.GenProposalForEpoch(rng, registry.LatestEpoch(), view0))
 
 	// Session 1: push PrepareQC + TimeoutQC, let runOutputs persist.
+	// Hoisted so session 1's WALs can be released after its goroutines have stopped: they hold an
```
pompon0, line 275: @wen-coding how did it manage to work so far? We didn't have a lock on the directory? Either way we should add a todo to make WALs work with structured concurrency - they should be opened and closed from within Run() function of the owner type.

pompon0 (approved):

sei-tendermint/internal/autobahn/consensus/state_test.go
wen-coding: The old WAL library didn't take a directory lock at all, so restart tests can reopen the same path without Close. seiwal brought in the concept of wal.lock, now we need to handle Close() carefully.

Yes we can add TODO to open and close within Run().

sei-tendermint/internal/autobahn/consensus/persist/commitqcs.go
wen-coding: I agree with that policy, but right now in PushCommitQC we wait for CommitQC to be durable before ingesting the next one, so will we be able to ingest CommitQC on N+1, N+2, ... before CommitQC on N is persisted?

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -185,6 +185,16 @@ func NewState(key types.SecretKey, data *data.State, stateDir utils.Option[strin
 	}, nil
 }
 
+// Close releases the WALs this state owns, and with them the exclusive lock each holds on its
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-tendermint/internal/autobahn/consensus/persist/wal.go
```diff
@@ -1,211 +1,123 @@
...
+//
+// name identifies the WAL in its metrics and has no effect on the on-disk layout, so instances whose
+// metrics should aggregate may share one.
+//
+// fileSize is the size a file may reach before it is sealed, which is also the granularity at which
+// pruning reclaims space. Production callers pass targetFileSize.
+//
+// Appends to the returned WAL are only durable once Flush returns; a caller that reports a record as
+// persisted must flush first.
+func openWAL[T any](dir string, name string, codec codec[T], fileSize uint) (seiwal.WAL[T], error) {
```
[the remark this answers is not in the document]
cody-littley: Not deployed anywhere yet, not real bug. Will not fix.

sei-tendermint/internal/autobahn/avail/state.go
```diff
@@ -185,6 +185,16 @@ func NewState(key types.SecretKey, data *data.State, stateDir utils.Option[strin
...
+// directory.
+//
+// Production does not call this: a node exits by rugpull and the OS reclaims everything. It exists so
+// that a process which opens the same state directory more than once in its lifetime — a test
+// simulating a restart — can release the first State before constructing the second.
+func (s *State) Close() error {
```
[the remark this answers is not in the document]
cody-littley: fixed

masih (approved):

sei-tendermint/internal/autobahn/consensus/persist/blocks.go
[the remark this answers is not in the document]
cody-littley: disabled metrics for lane WAL

sei-tendermint/internal/autobahn/consensus/persist/wal.go
```diff
@@ -1,211 +1,123 @@
...
-	if err := w.wal.TruncateAfter(walIdx - 1); err != nil {
-		return fmt.Errorf("truncate after WAL index %d: %w", walIdx-1, err)
+// contiguousSuffix returns the longest run of entries that ends at the final entry and whose indices
+// each exceed their predecessor by exactly one.
+//
+// A gap marks the boundary between garbage and live data. Pruning only deletes whole sealed files, so
+// records the prune anchor has moved past survive until every record in their file is below the
+// threshold; appending at the anchor's index then leaves a hole behind it. Everything before the last
+// hole was pruned and merely not yet reclaimed.
+func contiguousSuffix[T any](entries []walEntry[T]) []walEntry[T] {
```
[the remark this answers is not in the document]
cody-littley: documented

sei-tendermint/internal/autobahn/consensus/persist/wal.go
cody-littley: This is primarily a GC mechanism. In order to be efficient, we wait to delete a WAL file until it is safe to delete the entire file. We never delete just half of a WAL file. We can make these much larger, at the cost of slightly delayed disk space reclamation. 

4mb does seem kind of small, now that I think about it though. Adjusting this upwards to 1gb.

sei-tendermint/internal/autobahn/consensus/persist/blocks.go
cody-littley: I actually ended up disabling metrics for these WALs. Since we have one WAL per lane, we actually need a unique name per lane if we want to enable metrics. We could get that by putting the lane hash into each lane's name, but I wasn't sure if that was clean.

Let me know if you'd like me to turn the metrics back on. If we leave metrics off, this name is functionally ignored.

sei-tendermint/internal/autobahn/consensus/persist/wal.go
wen-coding: Err, does that mean on average we need to read 1GB of data before restart?

sei-tendermint/internal/autobahn/consensus/persist/commitqcs.go
cody-littley: I don't understand this code well enough to make a judgement call on this issue. @wen-coding @pompon0 can you work it out and let me know how you'd like it to work?

sei-tendermint/internal/autobahn/consensus/state_test.go
cody-littley: Added the TODO.

If a WAL lock is inconvenient, I can add an optional config to optionally disable them. Let me know if this is a route you'd like me to take.

wen-coding: Nah I think it's fine, this is only a problem for tests.

sei-tendermint/internal/autobahn/consensus/persist/commitqcs.go
wen-coding: I think for now what you have is fine. I'd argue if the normal case is <=1 CommitQC then how you call forEach at the end is okay.

sei-tendermint/internal/autobahn/consensus/persist/wal.go
wen-coding: I calculate one CommitQC is about 4KB, so we can fit ~1000 CommitQCs in each 4MB, I'd argue 4MB is enough, since you only purge the whole file when all entries are out of date, I'd prefer to read 4MB rather than 1GB on restart.

wen-coding (approved):

sei-db/seiwal/seiwal_config.go
```diff
@@ -66,6 +72,7 @@ func DefaultConfig(path string, name string) *Config {
 		PermitGaps:            false,
 		IteratorPrefetchSize:  32,
 		MetricsSampleInterval: 15 * time.Second,
+		MetricsEnabled:        true,
```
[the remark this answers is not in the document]
cody-littley: fixed

PR #3859, opened 2026-08-05
Migrate precompiles/bank and giga/deps metrics to OTel (PLT-912)

amir-deris:
## Summary

Part of the sei-chain OTel metrics migration ([PLT-912](https://linear.app/seilabs/issue/PLT-912)). Extends OTel dual-emit for `bank_new_account` and the OCC scheduler's `scheduler_retries`/`scheduler_incarnations` counters to the bank precompile and Giga fork paths. The `sei-cosmos` scheduler already had OTel dual-emit wired; this PR threads real request context into those calls and fixes the `scheduler_incarnations` description. Legacy (`telemetry.IncrCounter`) output is preserved on every path.

- `precompiles/bank/bank.go`: `sendNative`'s new-account path now calls `metrics.RecordBankNewAccount(ctx.Context())`, a new helper in `utils/metrics` that dual-emits to a `bank_new_account` OTel counter and the legacy counter, wrapped in a panic recover so a telemetry fault can't escape into precompile execution.
- `sei-cosmos/x/bank/keeper/metrics.go`: documents that its `bank_new_account` instrument is mirrored by the precompile helper above and by the Giga fork's copy below, so all three merge into a single series; description/unit must stay byte-identical across all three declarations.
- `giga/deps/xbank/keeper/`: new `metrics.go` declaring the Giga fork's own `bank_new_account` counter (kept import-free of chain-app code), wired into `send.go`'s `InputOutputCoins`/`SendCoins` new-account paths.
- `giga/deps/tasks/`: new `metrics.go` mirroring the sei-cosmos scheduler's `scheduler_retries`/`scheduler_incarnations` OTel instruments into the Giga fork, wired into `scheduler.go`'s `emitMetrics`.
- `sei-cosmos/tasks/scheduler.go`: `emitMetrics` now takes a `context.Context` threaded from `ProcessAll`'s `ctx.Context()` instead of using `context.Background()` (OTel dual-emit was already present).
- `sei-cosmos/tasks/metrics.go`: corrected the `scheduler_incarnations` description to "Sum of per-round maximum incarnations in the OCC scheduler" (it sums per-round maxes, not a single running max).

No precompile execution/state logic changed — purely additive OTel instrumentation. Legacy metric output is preserved on every path.

## Test plan
- [x] `go build` / `gofmt -l` / `goimports -l` clean on all touched files
- [x] `go test` passes for `giga/deps/tasks`, `giga/deps/xbank/keeper`, `utils/metrics`, `precompiles/bank`, `sei-cosmos/tasks`, `sei-cosmos/x/bank/keeper`


utils/metrics/metrics_util.go
```diff
@@ -52,6 +72,21 @@ func SetupOtelMetricsProvider(chainID string) error {
...
+// into a consensus-critical path.
+func RecordBankNewAccount(ctx context.Context) {
+	defer func() {
+		if e := recover(); e != nil {
+			debug.PrintStack()
+		}
+	}()
+	bankNewAccountCounter.Add(ctx, 1)
+	// TODO(PLT-353): remove once bank_new_account verified
+	SafeTelemetryIncrCounter(1, "new", "account")
```
bdchatham: `RecordBankNewAccount` runs the OTel `Add` before `SafeTelemetryIncrCounter` under one function-scoped recover, so a panic in the OTel emit drops the legacy counter silently and the PLT-353 parity comparison skews with it. On main the precompiles emitted the legacy counter unconditionally, so ordering legacy first, or giving each emit its own recover, would hold that guarantee.

masih (approved):

PR #3861, opened 2026-08-05
test(config): extend golden value test coverage (PLT-893)

bdchatham:
## Motivation

An earlier sweep renamed 27 operator-facing `app.toml` key names with the whole configuration suite green. "The suite covers these keys" and "the suite would notice a change to them" turned out to be different claims, and a test that cannot fail reads exactly like one that works.

## Test Specs & Approach

These suites pin how a `seid` node resolves configuration, meaning which value each setting ends up with once the config files, the environment and the built-in defaults have had their say. They record how that behaves today rather than how it was meant to, because the path has shipped and changing it would move what running nodes do. The versioned manager corrects those cases later, using these tests as evidence nothing else moved with them.

Three mechanisms carry the work, and each had a way to pass while asserting nothing.

**Checked-In Expected Values**

One file per group of settings, holding its defaults and its key names. Comparing a reader against the package's own defaults struct moves both sides together when a default changes, so a second copy the code does not derive from is what makes that change fail.

**Fuzz Seeds Per Setting**

Under an ordinary `go test` a Go fuzz test replays its hand-written examples and nothing else, so the seed set is the coverage. A setting whose seeds all resolve the way an absent key would is a setting nothing exercises.

**Per-Package Coverage Record**

One file per package listing which checks cover each group, read from the package's own test files. Deleting a check call was otherwise silent, because the checks that remain still pass.

## New Test Coverage

**Added Here**

Nine rows declared that a malformed value is rejected rather than quietly kept while no seed reached that path, and the harness now requires one. Five groups gained a defaults file. Five of the six helpers every row assertion flows through had no test of their own. Deleting a check call now fails a comparison naming what went missing, across all 11 wired packages. Each closure was checked by breaking the reader on purpose and confirming the tests noticed.

**Recorded Rather Than Repaired**

A node mode's state-store settings are discarded before a node reads them, so an archive node prunes at the default instead of keeping all history. Correcting which value wins changes the `app.toml` a newly initialised node receives, so it is tracked as **PLT-955**. Three further oddities are filed under a new `config-resolution-oddity` label as **PLT-952**, **PLT-953** and **PLT-954**.

**Coverage Today**

Counted from the records rather than claimed. Fourteen groups of settings are covered: defaults 14, key names 13, per-row resolution 13, seed discrimination 13, table completeness 12, absent-key baseline 8.

**Remaining Gaps**

Three, in the order worth closing.

1. The absent-key baseline reaches 8 of 14 groups. `genesis`, `light_invariance`, `server_config`, `state-commit`, `state-store` and `state-sync` have none, so nothing pins what those readers return when handed no keys.
2. Sixteen rows declare that omitting a setting yields something other than the built-in default, and nothing verifies that column, because the check that would needs the table passed to it.
3. `GetConfig` reads 78 distinct literal keys against a 23-row table, and 17 of them appear in no test anywhere. Their values are anchored by a defaults file, so the gap is names, casts and guards.

`gofmt` and `goimports` clean, `golangci-lint` (v2.8.0, CI's pin) reports 0 issues, all 12 affected packages pass. Nothing new runs in CI and no workflow changes, since these assertions live in suites that already run.


bdchatham:
Worked all review threads. Each claim was verified against the tree before acting, and the behavioural ones were checked by mutation. Resolving them now, with one left open deliberately.

**The one that mattered most: an assertion that could not fail.** The full-mode sub-case of the mode-overlay test was inert. `srvconfig.DefaultConfig` already carries `Enable: true` and `KeepRecent: 100000`, and `setFullnodeTypeAppConfig` assigns exactly those, so deleting both of its `StateStore` lines left the case green. I reproduced that by mutation rather than taking it on reading. The coincidence cannot be asserted away, since nothing about a resolved value separates an assignment from its absence when the two agree, so the case now asserts explicit literals and the comment states which failures they reach. A changed assignment reddens and so does the sei-db default drifting while the assignment is missing, both mutation-verified. Deletion alone stays undetectable and is documented as a property of the two values coinciding.

**The alias gap was real, not latent.** An earlier reviewer called it low severity. It is not: the aliased-and-never-calls-`CheckWiring` path reaches green with no report, because the package contributes zero pairs and the bootstrap then classifies it as unwired. The local name is resolved from the file's import specs now, in `wiringIn` and `callsCheckWiring` both. Verified by aliasing a real package's import and deleting its `CheckWiring` call: the bootstrap names it.

**The deferral rationale was wrong.** `SetAppConfigByMode` has one non-test caller, `init.go`, whose result reaches `WriteConfigFile` during `seid init`, and `initAppConfig` builds its template without the mode. So repairing the ordering changes the `app.toml` a newly initialised node receives, not what a running fleet prunes on restart. Corrected, along with the failure message that said "the config a node runs on". The deferral itself stands on the standing decision to pin resolution here and correct it in the versioned manager. Also recorded that shadowing reaches only the redeclared fields, so `MinRetainBlocks` and the API and GRPCWeb enables do take effect and archive mode fails partially.

**Also landed.** `recordRewriteInProgress` dropped from `CheckWiring`, which restores its doc's claim and removes a guard at a call site the choke point already covers. The record's set semantics are stated in the doc and the header, with every pair in the tree confirmed to be a single call today. The CI refusal names the variable and gives the `env -u CI` form. `aRejectedSeed` is a plain bool. `golden_contract_test.go` no longer uses `ParseDir`. Two consequences of allowlisting `CI` are recorded next to the entry.

**Performance.** The bootstrap walk prefilters on bytes before parsing, which took it from 1.02s to 0.11s. Splitting the scanned-versus-wired maps removed 1,011 redundant `ParseDir` calls and 24,509 redundant file parses per run.

**Left open on purpose.** The `AGENTS.md` thread asking to split the Structural corrections section into its own PR. I fixed its second point, the paragraph narrating this branch's own intermediate drafts, which no reader of merged history could see. Whether the normative guidance ships here or separately is the maintainer's call, so that stays open rather than resolved by me.

amir-deris (approved):

PR #3864, opened 2026-08-06
Integrate evmonly executor with giga store

codchen:
## Summary

- make the evmonly executor store-only: every block opens a giga snapshot and commits through `CommitStateChanges`
- require a store-specific `NamedChangeSetEncoder` and preserve storage-prefix clears
- add the loadtest `MemoryStore` implementation over the existing immutable `StateReader`
- encode state changes directly into typed, fixed-width `NamedChangeSet` key/value pairs with contiguous backing allocations
- apply direct pairs without an RLP encode/decode round trip
- retain historical state through per-block version slabs and linked version nodes, avoiding independently growing slices and duplicate account-touch indexes
- keep result sinks post-commit and preserve cleanup/error behavior across sequential and OCC execution

## Why

The evmonly executor now has one persistence model: a giga `Store`. The concrete store implementation can vary, but execution no longer has a separate non-giga state path.

The first loadtest adapter wrapped the complete native changeset in RLP and decoded it immediately inside `CommitStateChanges`. The direct format removes that redundant work while continuing to exercise the real giga interface.

## Loadtest

Configuration: 400 blocks, 1,000 transactions/block, one ordered block worker, 12 executor workers, zero gas price, and discard result sink. Values are three-run medians in tx/s.

| Workload | Pre-MemoryStore | Giga MemoryStore | Difference | OCC reruns/block |
|---|---:|---:|---:|---:|
| Transfer, unique | 254,684 | 227,601 | -10.6% | 0 |
| Transfer, 10% recipient conflicts | 224,630 | 198,971 | -11.4% | 50 |
| Transfer, hot recipient | 74,961 | 71,220 | -5.0% | 999 |
| Transfer, same sender | 84,601 | 81,371 | -3.8% | 999 |
| ERC20 transfer, unique | 194,834 | 175,273 | -10.0% | 0 |
| ERC20 transfer, 10% recipient conflicts | 163,868 | 146,445 | -10.6% | 50 |
| ERC20 transfer, hot recipient | 34,256 | 32,986 | -3.7% | 999 |
| Snapshot/revert | 262,522 | 233,799 | -10.9% | 0 |

Every run completed 400,000/400,000 transactions successfully with zero execution errors and zero OCC fallbacks.

The pre-MemoryStore comparison is not an equivalent persistence implementation: it executes against `WithState` and discards block state changes, while the Giga run encodes, commits, and retains current and historical state for later snapshots. It is therefore a useful lower bound on commit overhead, not evidence that the Giga interface itself costs 10% in production.

In an 800-block snapshot/revert profile, `EncodeMemoryStoreChangeSet` and `CommitStateChanges` each represented about 0.1% of sampled CPU. Most MemoryStore-specific allocation was the retained versioned storage map. A pointer-free indexed-history experiment did not improve end-to-end throughput and was reverted.

## Validation

- `go test ./giga/... ./sei-db/state_db/giga`
- `go test -race ./giga/evmonly/...`
- `go vet ./giga/evmonly/... ./sei-db/state_db/giga`
- touched Go files pass `gofmt -s -l` and `goimports -l`
- `git diff --check`

The full-tree `goimports -l .` reports pre-existing untouched generated and test files.

### DIFF
--- a/giga/evmonly/executor.go
+++ b/giga/evmonly/executor.go
@@ -26,7 +26,7 @@ type Executor struct {
 	resultPool       *blockResultPool
 	stateDBPool      sync.Pool
 	storeMu          sync.Mutex
-	store            gigastore.Store
+	store            gigastore.StateDB
 	changeSetEncoder NamedChangeSetEncoder
 	closed           atomic.Bool
 }
@@ -42,7 +42,7 @@ func WithResultSink(sink ResultSink) Option {
 // WithStore selects the giga store implementation used for all state reads and
 // commits. The encoder owns the implementation-specific conversion from the
 // executor's EVM-native StateChangeSet to the store's protobuf changesets.
-func WithStore(store gigastore.Store, encoder NamedChangeSetEncoder) Option {
+func WithStore(store gigastore.StateDB, encoder NamedChangeSetEncoder) Option {
 	return func(e *Executor) {
 		e.store = store
 		e.changeSetEncoder = encoder
--- a/giga/evmonly/executor.go
+++ b/giga/evmonly/executor.go
@@ -15,17 +15,21 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/params"
 	"github.com/sei-protocol/sei-chain/giga/evmonly/precompiles"
+	gigastore "github.com/sei-protocol/sei-chain/sei-db/state_db/giga"
 )
 
 // Executor runs raw EVM transactions against an EVM-native state backend.
 type Executor struct {
-	cfg         Config
-	state       StateReader
-	resultSink  ResultSink
-	occPool     *occWorkerPool
-	resultPool  *blockResultPool
-	stateDBPool sync.Pool
-	closed      atomic.Bool
+	cfg              Config
+	state            StateReader
+	resultSink       ResultSink
+	occPool          *occWorkerPool
+	resultPool       *blockResultPool
+	stateDBPool      sync.Pool
+	storeMu          sync.Mutex
+	gigaStore        gigastore.Store
+	changeSetEncoder NamedChangeSetEncoder
+	closed           atomic.Bool
 }
 
 type Option func(*Executor)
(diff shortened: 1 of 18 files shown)


giga/evmonly/executor.go
```diff
@@ -121,14 +125,14 @@ func (e *Executor) ExecutePreparedBlock(ctx context.Context, req PreparedBlock)
 	return result, nil
 }
 
-func (e *Executor) executePreparedBlock(ctx context.Context, req PreparedBlock) (*BlockResult, error) {
+func (e *Executor) executePreparedBlock(ctx context.Context, req PreparedBlock, source StateReader) (*BlockResult, error) {
```
yzang2019, line 128: Is source properly closed after the execution is done? The way how StateSnapshot works require the underlying goroutine to close the reader after usage is done

sei-db/state_db/giga/api.go
```diff
@@ -45,6 +45,8 @@ type StateDB interface {
 //
 // Until Close, the underlying resources (e.g. an ephemeral SC snapshot or a
 // pinned SS version) stay alive, even concurrently with later writes/commits.
+// All read methods must be safe for concurrent calls because EVM executor
+// workers may share one snapshot while executing a block.
```
yzang2019: We will rename this to StateView in the future PR, just FYI, no action needed for now

giga/evmonly/executor.go
codchen: Yes. `source` is only an adapter around the snapshot opened in `executePreparedBlockWithStore`, and that function registers `defer snapshot.Close()` immediately after the nil check. Sequential execution is synchronous; on the OCC path, `occWorkerPool.Run` returns only after `errgroup.Wait()` has joined every worker. Each worker also releases its `nativeStateDB` and resets its source before `executePreparedBlock` returns. Therefore the opening goroutine closes the snapshot only after all executor goroutines have stopped using it, including error paths, and before `ExecutePreparedBlock` proceeds to the result sink. `TestExecutorGigaStoreSnapshotFeedsOCCExecution` covers closure on the OCC path.

yzang2019 (approved):

PR #3866, opened 2026-08-06
fix(sei-cosmos): repair pagination scan-cap logic and make pagination bounds configurable (PLT-956)

amir-deris:

Fixes the LCD pagination regression.

**Root cause:** PLT-361 added hardcoded scan/limit/offset caps to `sei-cosmos/types/query`. `ValidatorDelegations` (and structurally the same for authz `GranteeGrants` / feegrant granter-allowances) has no validator-keyed index, so a filtered query must scan the *entire* delegation store. Hitting the raw-scan cap before the page filled discarded every hit already found and returned a hard error — failing *below* the caller's actual match count, not above it. A scan-cap hit past a filled page could also silently return `NextKey: nil`, which a caller reads as "last page" even though a real match exists beyond the scan window.

**What this PR does instead:**

1. **Fixes the actual logic bugs**, keeping the caps:
   - Hitting the scan cap before a filtered page fills (`FilteredPaginate` / `GenericFilteredPaginate`) now returns the partial page already found plus a resumable `NextKey`, instead of discarding results behind an error. A client that keeps following `NextKey` makes bounded progress even on an unindexed, sparse filtered query.
   - Hitting the scan cap past a filled page now also returns a resumable `NextKey` instead of `nil`, closing the silent-data-loss case.
   - The one deliberate exception: `count_total=true` still errors when the scan is cut short, since an approximate `Total` would silently violate the API's exactness contract.
   - `offset+limit` is now computed with a saturating add (`paginationEnd`), since raising the caps via config (below) makes the old bare addition capable of overflowing `uint64`.
2. **Bounds `GetBlockWithTxs`'s allocation by `min(limit, blockTxsLn)`** instead of trusting the raw requested limit — correct regardless of how high `MaxLimit` is configured.
3. **Makes `MaxLimit` / `MaxOffset` / `MaxScanLimit` configurable** via a new `[pagination]` app.toml section (`max-limit`, `max-offset`, `max-scan-limit`), defaulting to today's `1000` / `10000` / `10000`. Wired in at node startup via `query.SetPaginationLimits`. Deliberately no "disable" flag — an operator who wants an effectively unlimited scan sets an explicit large value in their own app.toml, a visible and deliberate choice rather than a one-line code-level bypass of the audited DoS fix.

**Explicitly out of scope:** adding a validator→delegations secondary index (and equivalent grantee/granter indexes) to turn these filtered queries from O(store) into O(matches). That's the structural fix for the underlying scalability defect, but requires a state migration and is tracked separately.

Refs: PLT-956, PLT-361 (#3494, cause), #3865 (revert-under-review, superseded by this PR), PLT-348 (umbrella), CON-310 (origin, Immunefi 79786)

- New unit tests: partial-page-on-sparse-filter and resumable-key-on-silent-loss for both `FilteredPaginate` and `GenericFilteredPaginate`, `count_total` still erroring on a cut-short scan, `SetPaginationLimits` zero-means-unchanged, and `offset+limit` overflow saturation.
- `gofmt -s` / `goimports` clean on all changed files.
- `go build ./...` clean.
- `go test` green across: `sei-cosmos/types/query`, `sei-cosmos/server`, `sei-cosmos/server/config` (including the `testutil/configtest`-adjacent characterization suite, golden files regenerated and reviewed), `sei-cosmos/x/auth/tx`, `sei-cosmos/x/authz(/keeper)`, `sei-cosmos/x/feegrant(/keeper)`, `sei-cosmos/x/staking/keeper`, `sei-cosmos/x/bank/keeper`, `sei-cosmos/x/gov/keeper`, `sei-cosmos/x/distribution/keeper`, `sei-cosmos/x/evidence/keeper`, `sei-cosmos/x/slashing/keeper`, `x/tokenfactory/keeper`, `app`, `cmd/seid/cmd`, and the `sei-ibc-go` / `sei-wasmd` keeper packages that call these pagination helpers.


amir-deris:
@seidroid can configuration of query pagination params cause app hash divergence, as raised by an earlier blocker comment?

PR #3868, opened 2026-08-06
Implement new Giga GarbageCollector interface

yzang2019:
## Summary

Retention for Giga's storage components is currently decided by each store's own
pruner, independently and with no knowledge of the others. This PR implements
`gc.PrunableStore` for all four of them so a single `StorageGarbageCollector`
can manage the pruning logic for the whole fleet.

Each cycle the collector asks every store for its ingest height and for the
oldest block it must keep in order to serve the shared `RollbackWindow`, takes
the minimum of those answers, and prunes every store to it. The stores therefore
stay mutually consistent — a rollback target that one store can serve is one they
can all serve, which per-store pruners could not guarantee.

## What's included

| Store | Change |
| --- | --- |
| `BlockDB` | New `litt_block_gc.go`; adds a `RetentionWindow` config field |
| `ReceiptDB` | New `litt_receipt_gc.go`; adds an `ExternalPruning` config field that gates the existing background pruner; adds a litt `GCFilter` so reclamation follows the retention floor |
| `StateWAL` | New `state_wal_gc.go`; `PruneBelow` calls `seiwal` directly |
| `FlatKV` (SC) | New `store_gc.go`; snapshot-aware pruning boundary; adds an `ExternalPruning` config field |

### `ExternalPruning() bool` on `PrunableStore`

Two stores keep a pruner of their own — FlatKV (`SnapshotKeepRecent`) and
ReceiptDB (`KeepRecent`) — because both still run without a collector: FlatKV in
the seidb tools and bench paths, ReceiptDB on any node with
`rs-backend = "littidx"`. Both pruners active at once is unsafe: the local one
would delete the very data the collector is holding to serve the rollback window.
Standing the local one down with nothing to replace it is equally unsafe, and
fails silently — the retention floor simply stops advancing.

`ExternalPruning` makes both combinations unrepresentable rather than merely
discouraged. Each store answers it from the same `config.ExternalPruning` field
its own pruner reads to stand down, so "the collector prunes this store" and "the
store does not prune itself" are one fact instead of two settings that can
disagree. Stores with no pruner of their own return `true` unconditionally.

A store that reports `false` is still asked for its boundary and still holds the
shared minimum down — it just never receives `PruneBelow`. Dropping it from the
vote instead would prune the WAL out from under the snapshots it replays from.

Neither `ExternalPruning` field is reachable from `app.toml` (both are
`mapstructure:"-"`). Enabling one is only correct when the store is registered
with a running collector, which is a property of how the process was wired and
not something an operator can assert. Where it is checkable it is checked:
`newReceiptBackend` rejects `pebbledb` + `ExternalPruning`, because that backend
is not a `gc.PrunableStore` and would end up with no pruner at all.

### The retention floor gates reclamation, and the TTL is only an age failsafe

Both litt-backed stores now pair their TTL with a `GCFilter`, so a record is
reclaimed only when it is **both** below the store's retention floor **and**
older than the TTL. Previously ReceiptDB had no filter, which made the TTL the
sole reclamation mechanism, and it was sized as `KeepRecent × 2s`. Under
`ExternalPruning` the enforced retention is `RollbackWindow + KeepRecent`, so a
TTL sized for `KeepRecent` alone expired receipt bodies for blocks the collector
still considered live — `ErrNotFound` inside the rollback window, which is the
exact cross-store guarantee the collector exists to provide.

With the filter in place the TTL no longer has to know how many blocks anything
is, so `littTTLPerBlock` is gone and both stores take a flat duration
(`RetentionTime` / `littRetentionTime`), defaulting to 1 hour. Visible retention
follows the floor and reclamation can no longer lead it.

### Retention semantics

With `R` = a store's `GetRetentionWindow` and `F = LatestBlock - RollbackWindow - R`,
collection guarantees, per managed store:

1. Nothing needed to roll back to any block in `[LatestBlock - RollbackWindow, LatestBlock]` is deleted.
2. No data at or above `F` is deleted — so even after rolling back to
   `LatestBlock - RollbackWindow`, the most recent `R` blocks are still readable.
3. Data below `F` is eventually deleted, each store reclaiming on its own schedule.

Guarantee 2 is why pruning is to the shared minimum rather than to each store's
own boundary: a retained snapshot is only restorable if the blocks that follow it
survive in the contiguous stores. This is recorded on
`StorageGarbageCollectorConfig.RollbackWindow` and, in block terms, on
`BlockDBConfig.RetentionWindow`.

`GetRetentionWindow` reports extra retention *beyond* the shared
`RollbackWindow`, with `InfiniteRetentionWindow` (`-1`) meaning never prune.
Note that `ReceiptStoreConfig.KeepRecent == 0` already means "keep everything",
which is the opposite of what `0` means to the collector, so it is mapped to
`InfiniteRetentionWindow` rather than passed through.

StateWAL answers `0` unconditionally, and has no retention config of its own. Its
depth is not its to declare: it is a replay source, and SC/SS already express how
far back it must reach by answering their oldest live snapshot as a boundary. A
window here would be additive on top of the shared minimum, retaining every
managed store further back rather than the WAL alone — a fleet-wide decision
wearing a per-store name, which is what `RollbackWindow` already is.

### Snapshot stores

FlatKV restores only at a snapshot boundary and replays the WAL forward from
there, so what it must retain is not a block range but the newest snapshot at or
below the target. `GetPruningBoundary` reports that, which is what holds the WAL
back for it.

## Config changes

- `littblock.BlockDBConfig.Retention` → `RetentionTime`, default `24h` → `1h`.
  It is an age floor, not a retention policy; how much history BlockDB keeps is
  `RetentionWindow`. The `AutobahnBlockDBConfig.Retention` override keeps its
  name, since its `retention` JSON key is a persisted config format.
- `littblock.DefaultConfig` now leaves `RetentionWindow` at `0` (was `10000`).
  It is an input to a minimum shared across every managed store, so a non-zero
  default here would have held ReceiptDB, the state WAL and the SC snapshots
  10k blocks further back on BlockDB's say-so. Every other store reports 0; a
  deployment wanting deeper block history sets it at the call site.
- New `ExternalPruning` on `ReceiptStoreConfig` and `FlatKVConfig`, both
  `mapstructure:"-"` and both defaulting to `false`.

`ReceiptStoreConfig.KeepRecent` is deliberately left at `0` (keep everything).
Nothing here couples it to `RetentionWindow`, because the two fields disagree
about `0`: BlockDB folds only negatives to `InfiniteRetentionWindow`, so `0`
there is the most aggressive setting, while ReceiptDB folds `<= 0` to infinite,
so `0` there means never prune. `KeepRecent` cannot express "nothing beyond the
rollback window" at all. Reconciling the two is left to the wiring PR, along
with what the shared window should actually be.

## Also in this PR

- Renames `littblock.LittBlockConfig` to `BlockDBConfig` (ripples into `sei-tendermint`).
- Renames the BlockDB table from `ledger` to `blocks`. No migration is needed —
  BlockDB is not deployed on any network and no such data exists. Because the
  table name is persisted layout rather than an identifier, `NewBlockDB` would
  otherwise open a fresh empty table beside the old data, and an empty store is
  indistinguishable from a correct one until something asks for history. A
  `refuseLegacyTable` check at open turns that into a startup error naming the
  directory, for dev/CI/devnet homes written before the rename; it can be deleted
  once no such directory remains.
- Documents `seiwal.WAL.PruneBefore` as safe to call off the WAL owner's
  goroutine, unlike most of the interface, which is what lets the collector prune
  the WAL from its own goroutine.

## Not in this PR

Not moving all stores to use `StorageGarbageCollector` yet. We expect to
construct it in a future PR when we decide to unify the pruning for mainnet. Both
`ExternalPruning` fields deliberately default to `false`: no behavior change by
default.

That wiring PR is also where the remaining guard belongs. The receipt path can
reject its unsupported combination at startup, but nothing on the FlatKV path can
validate that a collector exists — "a collector exists" is not knowable from that
package. Keeping the field unreachable from config is what stands in for the
check until then.

Also worth knowing at wiring time: enabling `ExternalPruning` changes the shape of
snapshot retention rather than just its depth. Snapshot count becomes roughly
`RollbackWindow / SnapshotInterval` instead of `SnapshotKeepRecent + 1`.

Not managing SS yet in this PR since SS doesn't have snapshot capability yet.

## Testing

- A `*_gc_test.go` suite per store, plus collector coverage for the self-pruning
  path (a store reporting `false` keeps its vote but receives no `PruneBelow`).
- `litt_receipt_gcfilter_internal_test.go` covers the filter as a predicate and
  end-to-end: blocks below the floor are reclaimed by a real litt GC pass, those
  at or above are retained. It fails if the filter is removed.
- `litt_receipt_pruner_internal_test.go` pins when the local pruner runs, as a
  pure predicate rather than a timing assertion.
- `litt_block_legacy_table_test.go` covers the pre-rename refusal, including that
  a refused open leaves the directory exactly as it found it.
- Config defaults that moved are re-recorded in `testdata/*.golden`, so each new
  value lands in a diff.
- Run under `-race`.


sei-db/ledger_db/block/littblock/litt_block_db.go
```diff
@@ -12,18 +12,19 @@ import (
 	"github.com/sei-protocol/sei-chain/sei-tendermint/libs/utils"
 )
 
-// ledgerTableName is the single table holding both blocks and QCs. They share
+// TableName is the single table holding both blocks and QCs. They share
 // one table so a crash leaves a contiguous write-order prefix spanning both
 // record kinds (see NewBlockDB), which is what guarantees a persisted block is
 // always covered by a persisted QC.
-const ledgerTableName = "ledger"
+const TableName = "blocks"
```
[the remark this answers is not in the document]
yzang2019: That's fine since there's no existing data for BlockDB, it's not used yet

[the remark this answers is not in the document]
yzang2019: Not a real blocker since this is not deployed in production

sei-db/ledger_db/block/littblock/litt_block_config.go
```diff
@@ -20,23 +21,45 @@ type LittBlockConfig struct {
...
+	// aggressive setting this field has; "keep everything" is -1. Assigning a KeepRecent
+	// value to this field inverts the retention it asks for.
+	//
+	// This is an input to a minimum shared across every managed store, not a policy applied
+	// to this store alone: a deep window here also holds back receiptDB and the SC/SS
+	// snapshots. Must be >= gc.InfiniteRetentionWindow.
+	//
+	// Independent of Retention, which is a wall-clock TTL failsafe underneath the watermark.
+	// Both must permit reclamation before any record is dropped.
+	RetentionWindow int64
```
cody-littley: Might be useful to call out the following invariants:

```
Storage garbage collection guarantees the following invariants:

1. Garbage collection will not delete any data that is necessary to roll back to any block 
   between LatestBlock and (LatestBlock - RollbackWindow), inclusive.
2. Garbage collection will not delete block DB data that is before 
   (LatestBlock - RollbackWindow - RetentionWindow). This ensures that even if the 
   system rolls back to block (LatestBlock - RollbackWindow), it is still possible to read any
   block from the last RetentionWindow blocks.
3. Garbage collection will eventually delete block data older than 
   (LatestBlock - RollbackWindow - RetentionWindow).
```

sei-db/seiwal/seiwal.go
```diff
@@ -64,6 +64,17 @@ type WAL[T any] interface {
...
+	// owner, concurrently with any method including Append and Close, and implementations must support
+	// that without external serialization. Retention is driven by a garbage collector on its own
+	// goroutine, and requiring it to take the writer's turn would mean either blocking the writer or
+	// deferring the prune until the writer next runs — the latter stalling reclamation indefinitely on a
+	// WAL that has stopped receiving appends.
+	//
+	// Concurrent calls are unordered with respect to appends: whether a record appended around the same
+	// instant is pruned is unspecified. This costs nothing, because which records a prune actually
+	// reclaims is already approximate — it drops whole sealed files, and may defer the work arbitrarily.
 	PruneBefore(lowestIndexToKeep uint64) error
```
cody-littley: LGTM. Technically Flush() is also legal to call from anywhere, but it's ok for the godocs to be more restrictive than the code when it comes to threading model.

sei-db/state_db/sc/flatkv/store_gc.go
```diff
@@ -0,0 +1,180 @@
...
+	// snapshot it names — but the cost of the check is one readlink against making a wrong answer
+	// elsewhere unbootable here.
+	_, activeVersion, err := currentSnapshotDir(dir)
+	if err != nil {
+		return fmt.Errorf("resolve active snapshot before pruning: %w", err)
+	}
+
+	var errs error
+	pruned := 0
+	scanErr := traverseSnapshots(dir, true, func(version int64) (bool, error) {
```
cody-littley: The following would probably simplify this code a lot:

- Create a method `getSnapshotBlocks() ([]uint64, error)` that returns a slice of block numbers in sorted order for all the snapshots on disk.
- Create a method `deleteSnapshot(block uint64) error` that deletes a single snapshot with the given block number

This lets us avoid the lambda function, and splits apart the logic for traversing the directory structure and the logic for deciding which blocks to keep and which ones to drop.

sei-db/ledger_db/block/littblock/litt_block_config.go
cody-littley: I suggest we paste this block of invariants at each RetentionWindow or RollbackWindow config (with wording adjusted a little, the above is specific to block storage).

cody-littley (approved):

sei-db/ledger_db/receipt/litt_receipt_gc.go
```diff
@@ -0,0 +1,82 @@
...
+//
+// Negative values are folded into the same case: KeepRecent is never negative in practice, and
+// this store has historically read <= 0 as "pruning off", so folding them keeps that reading
+// intact rather than passing an out-of-contract value to the collector.
+func (s *littReceiptStore) GetRetentionWindow() int64 {
+	if s.keepRecent <= 0 {
+		return gc.InfiniteRetentionWindow
+	}
+	return s.keepRecent
+}
```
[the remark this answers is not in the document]
yzang2019: Fixed by adding a gc filter

sei-db/ledger_db/block/littblock/litt_block_config.go
```diff
@@ -7,44 +7,59 @@ import (
...
 	littConfig, err := littdb.DefaultConfig(dir)
 	if err != nil {
 		return nil, fmt.Errorf("failed to build litt config: %w", err)
 	}
-	return &LittBlockConfig{
-		Litt:      littConfig,
-		Retention: 24 * time.Hour,
+	return &BlockDBConfig{
+		Litt:          littConfig,
+		RetentionTime: time.Hour,
```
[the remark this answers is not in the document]
yzang2019: that's fine

sei-db/management/gc/api.go
cody-littley, line 1: Nit, not blocking. Perhaps we could instead name this file `prunable_store.go`? If we name it like this, then it becomes easy to search for this file by name. If I search "api" in my IDE, I get lots of hits.

sei-db/management/gc/api.go
```diff
@@ -1,124 +1,102 @@
...
-// answers this is not idle — it will replay forward once its first snapshot lands, so pruning the
-// others now can delete the range it will replay from. The collector therefore stops:
+// PrunableStore is a store whose old data may be dropped by the StorageGarbageCollector.
 //
-//	RollbackWindow > 0  → the whole cycle is abandoned; nothing is pruned
-//	RollbackWindow == 0 → this store is ignored and the others are pruned
+// Deletion is split in two because the two halves are pruned to different depths. Snapshots go down
+// to the deepest rollback the fleet still owes, since a restore point below that is one nothing can
+// ask for. History goes one lookback window deeper, because the floor snapshot is only restorable if
+// the blocks above it survive, and because that history is still readable where the snapshot is not.
```
cody-littley: I suggest explaining it like something like this:

> There are two dimensions of garbage collection: snapshots, and history. Snapshots are a mechanism utilized to roll back the chain. History is what lets data stores answer queries about historical blocks. Not all stores have a concept of "history" (e.g. SC), and not all stores have "snapshots" (e.g. the state WAL).

sei-db/management/gc/api.go
cody-littley: This interface has overly verbose. Claude really likes generating long godocs that are not extremely helpful for understanding the code.

My general rules I give my LLM when generating godocs are as follows:

- Explain WHAT, not WHY or HOW.
- Never explain the history of design decisions in a godoc.
- Multi-paragraph godocs should be rare, in general. Most functions don't deserve multiple paragraphs of explanation.
- When editing a godoc to conform to these rules, it is often better to rewrite the godoc instead of making incremental edits. LLMs often struggle keeping a godoc cohesive when editing, which often leads to rambling paragraphs.
- Although sometimes appropriate, in general it's the wrong move to describe the system as a whole in a godoc. Most godocs should focus on the subject matter at hand, not encode an entire system's design doc.

sei-db/management/gc/api.go
```diff
@@ -1,124 +1,102 @@
...
+	// blockNumber is calculated with the consideration of both lookback and rollback window.
+	// For example, WAL can only prune the block number up till the last snapshot height.
 	//
-	// RollbackWindow is shared so every store stays consistent under rollback. Retention is
-	// optional history on top, so a store can still serve queries after a rollback has consumed
-	// the whole window.
+	// It is the minimum across every participating store less the lookback window, so it can sit
+	// below what this store alone would need — and above this store's head, since a store may lag
+	// the ones that set the minimum. Both are the store's to absorb: a request outside what it
+	// holds must clamp to a no-op rather than empty the store.
```
cody-littley: I suggest avoiding details about how the garbage collection works in this godoc. Implementers of this godoc don't need to know how the garbage collector aggregates this data together with other stores, they just need to know the required contract for these methods.

sei-db/management/gc/api.go
```diff
@@ -1,124 +1,102 @@
...
+	// off" are one fact rather than two settings that can disagree. A store with an internal
+	// pruner answers from the same field that pruner consults, which is what makes the two
+	// mutually exclusive by construction instead of by wiring discipline. A store with no
+	// pruner of its own returns true unconditionally.
+	ExternalPruning() bool
+
+	// GetRollbackFloor returns the earliest height a rollback may target: given rollbackWindow,
+	// the deepest height this store could still be asked to restore to, and so the oldest block
+	// it needs the fleet to keep. The collector takes the minimum across stores to get the
+	// snapshot cut line, then subtracts LookbackWindow from that for the history cut line.
```
cody-littley: Suggest avoiding details about the top level GC algorithm, and instead focusing on the specific contract required to be implanted by this method.

e.g.

```
If this store does not have the concepts of snapshots, it should return X. If it does manage snapshots,
it should return the block number of its highest snapshot that occurs on or before `rollbackWindow`.
If it has no snapshots (e.g. we are starting from genesis or a state sync), it should return Y.
```

sei-db/management/gc/api.go
```diff
@@ -1,124 +1,102 @@
...
-	// nothing about whether the store may be pruned around: one that cannot serve a rollback
-	// still stops the cycle, via GetPruningBoundary.
+	// GetLatestBlock returns the highest block this store has ingested, 0 when it has ingested
+	// nothing.
 	//
-	// 0 does not mean "disabled". A store disabled for this node is not instantiated and never
-	// reaches the collector.
+	// It is the head GetRollbackFloor measures the rollback window against. The collector does
+	// not call it — every height it acts on comes from GetRollbackFloor — so this is here as the
+	// store's ingest position for operators and tests.
```
cody-littley: Same comment, avoid talking about what the garbage collector does, and instead focus on what implementations of this interface are required to do.

sei-db/management/gc/storage_garbage_collector.go
```diff
@@ -15,49 +16,22 @@ var logger = seilog.NewLogger("db", "gc")
...
 //
-//  1. head = min non-zero GetLatestBlock across stores
-//  2. per store: cutLine = head - RollbackWindow - GetRetentionWindow, skipping the store
-//     when that is 0 (infinite retention, or head still inside its own window)
-//  3. ask GetPruningBoundary(cutLine): a positive answer votes, and CannotServeRollback
-//     abandons the cycle while RollbackWindow > 0
-//  4. pruneHeight = min of the positive answers; PruneBelow(pruneHeight) on every store
-//     that answered positively
+//  1. ask every store GetRollbackFloor(RollbackWindow) — the earliest height it could still be
+//     asked to roll back to, which it resolves against its own head
```
cody-littley: It's unclear what `"which it resolves against its own head"` means

sei-db/management/gc/storage_garbage_collector.go
```diff
@@ -15,49 +16,22 @@ var logger = seilog.NewLogger("db", "gc")
...
+//  2. snapshotCutLine = min of those answers, the deepest rollback the fleet still owes
```
cody-littley: unclear what `"the fleet still owes"` means

sei-db/management/gc/storage_garbage_collector.go
```diff
@@ -15,49 +16,22 @@ var logger = seilog.NewLogger("db", "gc")
...
+//  3. historyCutLine sits LookbackWindow below snapshotCutLine, so the lookback window falls
```
cody-littley: unclear what `historyCutLine sits LookbackWindow below snapshotCutLine` means

sei-db/management/gc/storage_garbage_collector_config.go
```diff
@@ -7,51 +7,75 @@ import (
...
+	// One window covers every managed store. With a finite LookbackWindow and
+	// F = LatestBlock - RollbackWindow - LookbackWindow, collection guarantees (a LookbackWindow of
+	// -1 keeps everything, so F is genesis):
+	//
+	//  1. Nothing needed to roll back to any block in
+	//     [LatestBlock - RollbackWindow, LatestBlock] is deleted.
+	//  2. No data at or above F is deleted.
+	//  3. Data below F is eventually deleted, on each store's own reclamation schedule.
+	//
+	// Snapshots stop one window short of history, at the lowest rollback floor the stores report,
```
cody-littley: unclear meaning: `"Snapshots stop one window short of history"`

sei-db/state_db/sc/flatkv/store_gc.go
```diff
@@ -0,0 +1,297 @@
...
+//
+// The snapshot this store reported from GetRollbackFloor is never a candidate: blockNumber is the
+// minimum across every store, so it sits at or below that answer. Keeping the reported restore point
+// therefore follows from the collector's reduction and needs no guard here. The active snapshot is a
+// separate question, and pruneCutLine is where that one is answered.
+//
+// Deletion is not all-or-nothing: a snapshot that fails to delete is reported and the rest are still
+// attempted, since each snapshot directory is independent and a single undeletable one must not strand
+// the disk space of the others. Already-gone snapshots are not an error — the SnapshotKeepRecent
+// pruner in WriteSnapshot deletes from the same set, and losing that race means the work is done.
```
cody-littley: On this file and on the other files that implement `gc.PrunableStore`, I suggest doing a godoc pass. In general, the godocs in this file are quite verbose. I suggest skipping information about the how the high level garbage collector works, and to instead focus on what each implementation specifically is doing. A danger of encoding the high level algorithm in each of these implementations is that it's a doc that can drift out of date over time if we ever change the garbage collector implementation.

cody-littley (approved):

sei-db/state_db/sc/flatkv/store_gc.go
yzang2019: Make sense, will simplify

sei-db/management/gc/storage_garbage_collector_config.go
yzang2019: Rewording

sei-db/management/gc/storage_garbage_collector.go
yzang2019: It means historyCutline = SnapshotCutLine - LookbackWindow

sei-db/management/gc/storage_garbage_collector.go
yzang2019: Fixed

PR #3874, opened 2026-08-07
ci: shard Go race tests into a round-robin 3-way matrix (PLT-959)

amir-deris:
## Summary

- Race Detection previously ran all ~420 packages serially in a single job (~18 min baseline). It's now sharded into a 3-way matrix in `go-test.yml`, with an aggregate check that keeps the **"Race Detection"** status name so branch protection doesn't need to change.
- Packages are split **round-robin** (`package i → shard i % NUM_SPLIT`) via the Makefile's `split-test-packages` target.
- A short list of consistently slow packages (e.g. IBC keeper packages, `sei-db` engine tests, `giga/tests`, `evmrpc/tests`) is round-robined separately, ahead of the rest, so they can't all land on the same shard by coincidence.
- `make test-group-N` reproduces a given CI shard locally.
- Preserves `-race`, `-tags=ledger,test_ledger_mock`, and the `occ_tests` `-parallel=1` special case per shard.

## Results

Latest run: shard 0 = 13m0s, shard 1 = 12m28s, shard 2 = 10m41s (max ~13 min), down from the ~18 min single-job baseline, showing about 28% CI runner time reduction for this job.

## Context

Linear: [PLT-959](https://linear.app/seilabs/issue/PLT-959)

Starting at `NUM_SPLIT=3`; can scale up further once we've confirmed shared-runner capacity handles it.

## Test plan

- [x] Confirm the run fans out into "Race Detection (shard N)" jobs plus the aggregate "Race Detection" check
- [x] Confirm all shards pass and the aggregate check reflects failures correctly
- [x] Confirm wall-clock improvement over the single-job baseline
- [x] Confirm `NUM_SPLIT=3 make test-group-0` locally reproduces shard 0's package set


amir-deris:
Will create another PR from same changes to cleanup the AI feedback.

PR #3883, opened 2026-08-07
Retire the IBC precompile and drop dead IBC references (step 1 of sei-ibc-go removal)

masih:

IBC inbound and outbound are disabled on chain and are not coming back, and all funds have been reviewed. This is the first step toward deleting `sei-ibc-go/` entirely at the v6.7 coordinated upgrade. It removes every IBC dependency that can be removed without touching the module wiring, so the later steps have a smaller surface.

**Two commits:**

1. **`precompiles/ibc`: retire the precompile at every version.** `0x…1009` now reverts with `"ibc precompile is retired; ibc transfers are disabled"` instead of building a `MsgTransfer`. A new `pcommon.NewRetiredPrecompile` holds the single retired implementation; the root package and all 15 legacy tags are thin wrappers over it, each supplying its own embedded `abi.json`.

2. **Drop IBC references that were already dead.** `x/evm/keeper` and `giga/deps/xevm/keeper` each stored an `ibctransferkeeper.Keeper` that no method in either package ever read; `tools/utils` listed the `ibc`/`transfer` stores; and three archived legacy common packages declared IBC keeper interfaces that nothing implements once the precompile is retired.

### Three decisions worth reviewing

**The address stays registered.** Unregistering it would let a call to `0x…1009` succeed and return empty data instead of reverting, quietly turning a caller that fails today into one that believes it succeeded. This follows the retired oracle precompile at `0x…1008`.

**Every tag is retired, not just the current one.** The oracle retirement kept its legacy versions executing real logic, because the oracle module still exists. Here the `transfer` store is being unmounted at v6.7, so a legacy version replayed at an old height would have no keeper to call — preserving the original code would buy nothing. The practical cost is that `debug_traceTransaction` on a historical block that called `0x…1009` reports a revert rather than the original transfer. That fidelity is forfeited by the store removal regardless of what this PR does.

**Three archived legacy packages were edited.** `precompiles/common/legacy/{v605,v606,v610}/expected_keepers.go` had their IBC keeper interfaces removed. This is against `scripts/bump_version`'s rule that archived tags are immutable, so it is called out rather than slipped in. The edit is compile-only: nothing referenced those interfaces once the precompile stopped needing them.

### Remaining steps (not in this PR)

- Unwire wasm IBC (`sei-wasmd/x/wasm/ibc.go`, `keeper/{ibc,relay}.go`, the handler and query plugins, `types/expected_keepers.go`, `wasmbinding`'s `IBCRawPacketHandler`).
- Unwire `app` (keepers, IBC router, `ModuleBasics`, `maccPerms`, gov client-proposal routes, `SetOrderInitGenesis`, param subspaces, `app/ante.go`, the `CheckMessage` block in `app/ante/cosmos_checktx.go`, the `ibcclient` begin blocker in `app/legacyabci/begin_block.go`).
- Drop `ibc`/`transfer`/`capability` from `app.kvStoreKeyNames` and `sei-db/common/keys.MemIAVLStoreKeys`, and add `StoreUpgrades{Deleted: ...}` for `v6.7` in `SetStoreUpgradeHandlers`, the same shape as `"dex"` at v5.8.0.
- `git rm -r sei-ibc-go` (~109k Go lines, roughly half generated `.pb.go`).

**One product decision blocks the wasm step.** `Keeper.instantiate` and `Keeper.migrate` call `ensureIbcPort` when `report.HasIBCEntryPoints`. With no port keeper, uploading or instantiating a contract that exports IBC entry points either has to be rejected outright or silently skip port binding. Rejecting is more honest — such a contract cannot work — but it is a consensus behavior change for deployers, so it should be an explicit call rather than an implementation detail.

**One thing worth doing before the module goes.** `seid query ibc-transfer denom-trace` is the only way to resolve an `ibc/<hash>` denom to its origin chain and base denom, and it dies with the module. If any of those denoms still carry supply, consider dumping the denom-trace table into the repo first so the mapping survives.

- `go build ./...` — clean.
- `go vet ./...` across the whole repository — clean (this compiles test packages too).
- `go test ./precompiles/...` — all pass.
- `go test ./x/evm/keeper/...` — passes, including `TestGetCustomPrecompiles`, the completeness check that maps every entry in each precompile's `versions` file to an upgrade height. The IBC `versions` file is unchanged, so all 15 tags still resolve.
- `gofmt -s -l .` and `goimports -l .` report nothing for any touched file.

Not covered: there is no integration test for the IBC precompile to update — `integration_test/precompile_tests/README.md` records that it is out of scope, and `0x…1009` is absent from `PRECOMPILE_ADDRESSES`.

<div><a href="https://cursor.com/agents/bc-e19ebff9-1963-4667-a213-b11a58c2e008?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-e19ebff9-1963-4667-a213-b11a58c2e008&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


masih:
Superseded by a cleaner more maintainable approach here https://github.com/sei-protocol/sei-chain/pull/3884

PR #3885, opened 2026-08-10
feat(seeds): ship Sei Labs seeds as the default bootstrap-peers

monty-sei:
## What

A fresh `seid init --chain-id pacific-1` (or `atlantic-2`) now writes a `config.toml` with `bootstrap-peers` already populated with the Sei Labs seed nodes, so a node bootstraps peer discovery **with no other config set**. Today the field defaults to `""` and operators have to source a peer list out of band.

## How

Adds `app/seeds`, mirroring the existing `app/genesis` pattern for well-known chain data, and calls `applyDefaultBootstrapPeers` in `InitCmd` after the chain-id is resolved and before `WriteConfigFile`.

## Decisions worth reviewing

- **`bootstrap-peers`, not `persistent-peers`.** Seeds populate the address book via PEX and may then be dropped. Holding operator connections open against our seeds indefinitely is wrong for them and a load multiplier for us.
- **arctic-1 is deliberately excluded.** It is a devnet, it has no Cosmos chain-registry entry, and a devnet is the most likely network to be reset or re-keyed — the worst case for an address baked permanently into a release. It stays well-known for genesis; it just ships no seeds. A test asserts both halves of that.
- **Init-time only, no runtime injection.** Writing config at start-up would mutate a file the operator owns.

## Behaviour on existing configs — please read before release notes

An earlier revision of this description said "the operator value always wins". That was imprecise, and the review was right to flag it. Precisely:

- `seid init` builds its config from `tmcfg.DefaultConfig()` and **exposes no flag for `bootstrap-peers`**, so the field is always empty at that point. The empty check in `applyDefaultBootstrapPeers` is **defensive, not an operator-precedence mechanism** — it keeps the behaviour correct for any future caller that pre-populates the field.
- **Without `--overwrite`**, `init` refuses to touch an existing config at all, so a hand-edited `bootstrap-peers` is safe.
- **With `--overwrite`**, `config.toml` is rewritten wholesale, so a hand-edited `bootstrap-peers` is now replaced **by the Sei seeds instead of by `""`**. Not a regression, and arguably an improvement, but it is a behaviour change and belongs in the release notes.

Verified empirically: hand-edit `bootstrap-peers`, run `init --overwrite`, and the seeds replace it; run `init` without `--overwrite` and it errors out leaving the file untouched.

**Nodes that already ran `init` do not retroactively get seeds** — they are covered by the docs update and a separate chain-registry submission.

## Permanence

These strings ship inside released binaries and operators pin them; the secret-connection handshake verifies the NodeID, so a changed ID is a rejected dial rather than a degraded one, and a release in the wild cannot be recalled. The inputs are final: the DNS pattern is settled, all node keys are pinned in encrypted secrets, and every instance-target NLB port is pinned in the infrastructure repo. Retiring an address means keeping it dialable until every release carrying it is out of use — noted in the package doc.

## Drive-by

Corrects the `--chain-id` flag help, which claimed *"if left blank will use sei"* while the code panics on an empty value.

The other half of that mismatch — `panic()` printing a Go stack trace for what is plain operator error, where `RunE` returns errors for the neighbouring invalid-mode case — is left for a separate PR, per the review. It is tracked.

## Testing

**Wiring coverage (new).** `TestInitCmdWritesDefaultBootstrapPeers` executes the real `InitCmd` against a temp home and asserts the written `config.toml`. This closes a gap the review identified: previously nothing exercised `RunE`, so deleting the wiring left the entire suite green. Confirmed by mutation — removing the call now fails this test.

**Data coverage.** `app/seeds` tests hold each NodeID against `types.NodeID.Validate` rather than a local regex (so the check cannot drift from CometBFT's definition), require the `:26656` port, and assert NodeID **and** host uniqueness across the whole table — a per-chain check would miss a `pacific-1` entry pasted into the `atlantic-2` block. A further test asserts every seeded chain is well-known per `genesis.IsWellKnown`, catching a typo'd chain-id that would otherwise be a silent no-op.

End to end with a locally built binary:

| `--chain-id` | `bootstrap-peers` |
|---|---|
| `pacific-1` | 3 entries |
| `atlantic-2` | 3 entries |
| `arctic-1` | `""` |
| `my-private-chain` | `""` |

Existing `cmd/seid/cmd` tests pass unchanged.

## Related

The operator docs update is held as a draft until this ships in a release, since it documents the defaulted behaviour.


alexander-sei (approved):

app/seeds/seeds.go
```diff
@@ -0,0 +1,62 @@
...
+// ForChain returns the Sei Labs seed addresses for a well-known chain, or nil
+// when the chain-id is not recognised (private and local chains included).
+// The returned slice is a copy; callers may not mutate package state.
+func ForChain(chainID string) []string {
+	s, ok := chainSeeds[chainID]
+	if !ok {
+		return nil
+	}
+	out := make([]string, len(s))
+	copy(out, s)
```
masih: use built-in Go SDK `slices` to copy.

app/seeds/seeds.go
```diff
@@ -0,0 +1,62 @@
...
+	return out
+}
+
+// BootstrapPeers returns the seeds for a chain as the comma-separated value
+// CometBFT's `bootstrap-peers` expects, or "" when the chain is unrecognised.
+func BootstrapPeers(chainID string) string {
+	return strings.Join(ForChain(chainID), ",")
```
masih: non-blocker: This is an anti-pattern: over refactored exported API that is only used once.

Why export `ForChain` at all. Why refactor it to its own function that is only used once?

app/seeds/seeds_test.go
```diff
@@ -0,0 +1,107 @@
...
+// across the whole table, not per chain: the likeliest copy/paste error is a
+// pacific-1 entry pasted into the atlantic-2 block, which a per-chain check
+// would miss.
+func TestSeedAddressesAreWellFormed(t *testing.T) {
+	seenID := map[string]string{}
+	seenHost := map[string]string{}
+
+	for chainID, addrs := range chainSeeds {
+		if len(addrs) == 0 {
+			t.Errorf("%s: no seeds configured", chainID)
```
masih: Use testify utilities to reduce verbosity.

app/seeds/seeds_test.go
```diff
@@ -0,0 +1,107 @@
...
+// A malformed NodeID is rejected at the secret-connection handshake, so the
+// seed silently never connects — no error, no metric (seed mode serves none).
+// This is the test that stops that shipping.
+//
+// The ID is held against types.NodeID.Validate rather than a local pattern, so
+// this cannot drift from CometBFT's actual definition. Uniqueness is checked
+// across the whole table, not per chain: the likeliest copy/paste error is a
+// pacific-1 entry pasted into the atlantic-2 block, which a per-chain check
+// would miss.
+func TestSeedAddressesAreWellFormed(t *testing.T) {
```
masih: is there a tendermint address parser we can use instead of string slicing? that would make a stronger test and avoid the tests here diverging from what actually matters: tendermint p2p understands the addresses.

app/seeds/seeds_test.go
```diff
@@ -0,0 +1,107 @@
...
+			id, hostPort, ok := strings.Cut(a, "@")
+			if !ok {
+				t.Errorf("%s: seed %q is not in NodeID@host:port form", chainID, a)
+				continue
+			}
+			if err := types.NodeID(id).Validate(); err != nil {
+				t.Errorf("%s: seed %q has an invalid NodeID: %v", chainID, a, err)
+				continue
+			}
+			if !strings.HasSuffix(hostPort, ":26656") {
```
masih: Why should one cate about the port? That is unclear from the tests alone. Technically, that port can be any valid port. no?

app/seeds/seeds_test.go
```diff
@@ -0,0 +1,107 @@
...
+	first[0] = "tampered"
+	if second := ForChain("pacific-1"); second[0] == "tampered" {
+		t.Fatal("ForChain leaked the package-level slice to callers")
+	}
+}
+
+func TestBootstrapPeers(t *testing.T) {
+	got := BootstrapPeers("pacific-1")
+	if n := len(strings.Split(got, ",")); n != 3 {
+		t.Errorf("expected 3 comma-separated entries, got %d (%q)", n, got)
```
masih: what is magical about the length 3 other than the fact that it so happens to be the number of seed nodes we have added?

I would encourage tests to capture the essence of business logic and rationale, more than mirroring what so happens to be an incarnation of it.

Unless I have missed something?

masih (approved):
Integration tests would be awesome to have where we assert those endpoints are connectable and are usable as seed, i.e. new peers are discovered from them. Those tests should be disabled by default and only run on CI since they would be making remote calls.

Excited to get this moving! 🙌

app/seeds/seeds.go
pompon0: and there is no need to special-case nil - slices.Clone already does that

app/seeds/seeds.go
pompon0: +1, if ForChain was unexported, then even cloning the slice makes no sense.

app/seeds/seeds_test.go
pompon0: p2p.ParseNodeAddress is internal to sei-tendermint, but I'd recommend to reexport it from sei-tendermit/config.

app/seeds/seeds_test.go
pompon0: +1

app/seeds/seeds_test.go
pompon0: +1, this is a change detector. At best you can assert that the list is non-empty. The other problem of this test is that it dupes TestForChain

app/seeds/seeds_test.go
```diff
@@ -0,0 +1,107 @@
...
+	if got := ForChain("arctic-1"); got != nil {
+		t.Fatalf("arctic-1 is a devnet and must not ship seeds, got %v", got)
+	}
+	// Guard the premise of the exclusion: arctic-1 is still initialisable.
+	if !genesis.IsWellKnown("arctic-1") {
+		t.Error("arctic-1 is expected to remain a well-known chain for genesis")
+	}
+}
+
+func TestForChain(t *testing.T) {
```
pompon0: these test are not useful as soon as you make ForChain private.

app/seeds/seeds_test.go
```diff
@@ -0,0 +1,107 @@
...
+func TestSeedChainsAreWellKnown(t *testing.T) {
+	for chainID := range chainSeeds {
+		if !genesis.IsWellKnown(chainID) {
+			t.Errorf("chain %q has seeds but is not a well-known chain (typo?)", chainID)
+		}
+	}
+}
+
+func TestArcticIsDeliberatelyExcluded(t *testing.T) {
+	if got := ForChain("arctic-1"); got != nil {
```
pompon0: nil check is too strong, just assert it is empty

app/seeds/seeds_test.go
```diff
@@ -0,0 +1,107 @@
...
+		t.Fatalf("arctic-1 is a devnet and must not ship seeds, got %v", got)
+	}
+	// Guard the premise of the exclusion: arctic-1 is still initialisable.
+	if !genesis.IsWellKnown("arctic-1") {
```
pompon0: make the chain names constants instead of copying them all over the place

app/seeds/seeds_test.go
```diff
@@ -0,0 +1,107 @@
...
+		t.Fatal("ForChain leaked the package-level slice to callers")
+	}
+}
+
+func TestBootstrapPeers(t *testing.T) {
+	got := BootstrapPeers("pacific-1")
+	if n := len(strings.Split(got, ",")); n != 3 {
+		t.Errorf("expected 3 comma-separated entries, got %d (%q)", n, got)
+	}
+	if strings.Contains(got, " ") {
```
pompon0: Does the config parser really complain about spaces?

app/seeds/seeds.go
monty-sei: Hey @masih! Yep you're right, and @pompon0's follow up nails the rest of it. I checked and `ForChain` had no caller outside the package at all, so exporting it (and defensively cloning the slice for callers that didn't exist) was API for nobody. I've dropped it entirely, so the package is now just the address table plus `BootstrapPeers`, which is the only thing `seid init` actually uses. `BootstrapPeers` reads straight off the map now, a missing chain gives nil and `strings.Join` renders that as "", so the nil special casing went away too!

app/seeds/seeds.go
monty-sei: Done, it uses `slices` now, and @pompon0 is right that the nil case needed no special handling. It actually folded away entirely once `ForChain` went (see the thread below), so `BootstrapPeers` just joins straight off the map!

app/seeds/seeds_test.go
monty-sei: Great call, and @pompon0 pointed me at the route for it. `p2p.ParseNodeAddress` is under `sei-tendermint/internal` so `app/seeds` can't reach it, but I had a look and `sei-tendermint/config` already imports `internal/p2p` (it surfaces `p2p.NodePublicKey` in `AutobahnValidator`), and `internal/p2p` doesn't import config, so there's no cycle. I've added a small re-export there and the tests now parse every address with the same parser the router dials with, instead of splitting on "@" and re-encoding the format locally.

Happy to trim that to validation only (just returning the error) if you'd rather not widen config's surface, just let me know!

app/seeds/seeds_test.go
monty-sei: Fair point, 26656 is our convention rather than anything the protocol requires. Parsing the address covers the port properly now (the parser rejects an invalid one) so I've dropped the specific 26656 assertion, which I think gets the actual invariant without the magic number.

app/seeds/seeds_test.go
monty-sei: Nothing magical about it at all, you and @pompon0 are both right that it's just a change detector for how many seeds we happen to run. It now asserts the list is non-empty and that every entry parses, and I've removed the overlap with the other test that @pompon0 spotted.

app/seeds/seeds_test.go
monty-sei: Done! I had a look and `app/genesis` and `app/params` both use testify and there's no rule either way in the root AGENTS.md, so it matches the neighbours now and the file is a fair bit shorter for it.

app/seeds/seeds_test.go
monty-sei: Good catch, it doesn't! I went and checked and `setup.go` calls `SplitAndTrimEmpty(cfg.P2P.BootstrapPeers, ",", " ")`, which trims the cutset off each element, so spaces are tolerated and that assertion was testing an invariant that doesn't exist. Removed.

app/seeds/seeds_test.go
monty-sei: Agreed, empty is the contract rather than nil specifically, so it asserts that now.

app/seeds/seeds_test.go
monty-sei: Done, they're constants now instead of being copied through every test.

app/seeds/seeds_test.go
monty-sei: Yep, both of those went with it. `ForChain` is unexported (actually removed entirely) so the copy semantics test had nothing left to assert.

monty-sei:
Hey @masih! This is the comment I was most keen to action, so I've added a build tagged suite (`go test -tags=integration ./app/seeds/...`) that's off by default and dials every published seed, asserting it actually sends its preface rather than just accepting the TCP connection.

Worth flagging that it currently fails on two of the six, both in the prod cell, and that's a real finding rather than a flaky test. Those endpoints complete the TCP handshake and then send nothing at all, even holding the connection open for three minutes, while the ones in prod-euw1 and prod-use2 respond in under a second. I'm still digging into why and @bdchatham is looking into it as well, it needs a look from inside the cluster to tell whether the pod is serving at all or whether it's something in the load balancer path. Because seed mode publishes no metrics and the pods report Ready, there's currently nothing that would have told us, which I think makes a pretty good case for exactly the tests you're asking for!

I'd like to get to the bottom of those two and have the full suite green before we merge, so the seeds are genuinely serving by the time the defaults go out to operators.

app/seeds/seeds_test.go
monty-sei: Small correction to my last reply, I've had to put a port assertion back in, though for a different reason than the original one.

Bugbot caught that `ParseNodeAddress` substitutes 26657 (the RPC port) when the port is missing or zero, so parsing on its own never fails on a dropped `:26656`. That meant my "the parser covers it" answer was wrong, an entry pasted without the port would have parsed clean, passed every test and shipped pointing at the RPC port, which is exactly the sort of thing we can't recall once it's in a release.

So it now asserts the port explicitly, but on the basis that a missing one silently becomes something wrong rather than 26656 being a protocol rule. Your underlying point still stands and the magic number is gone from everywhere else!

Makefile
```diff
@@ -174,6 +174,9 @@ lint:
 	go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.8.0 run
 	go fmt ./...
 	go vet ./...
+	# Build-tagged files are skipped by the vet above, so they would rot
+	# unnoticed against API changes. Compile-check them here.
+	go vet -tags=integration ./app/seeds/...
```
masih: This seems too specific of a lint; i would either remove it entirely, or run it for `./...`

app/seeds/seeds_integration_test.go
```diff
@@ -0,0 +1,75 @@
...
+func TestSeedAddressesAreDialableForm(t *testing.T) {
+	for chainID, addrs := range chainSeeds {
+		for _, entry := range addrs {
+			addr, err := config.ParseNodeAddress(entry)
+			require.NoErrorf(t, err, "%s: %q", chainID, entry)
+			require.Lenf(t, string(addr.NodeID), 40, "%s: %q", chainID, entry)
+			// Not NotZero: ParseNodeAddress fills a missing port with 26657,
+			// so that assertion could never fail and a dropped ":26656" would
+			// ship pointing at the RPC port.
+			require.EqualValuesf(t, 26656, addr.Port,
```
masih: Ditto re explicitly asserted port number.

Tests should read as executable documentation of business logic. What I read here is that the business demands port to be exactly 26656. What we mean, howerver, is that the port should match whatever the `entry` port was... which is implicitly asserted by `ParseNodeAddress` and dial test above... at which point I would simply remove this entire test.

app/seeds/seeds_integration_test.go
```diff
@@ -0,0 +1,75 @@
+//go:build integration
```
masih, line 1: Is `integration` an existing tag used elewhere? If yes and CI is already hooked up to running them then great. 

If not, i recommend to:
* pick an existing tag to save yourself from having to set up a CI for it, unless there is a really good reason that I might be missing for keeping it separate.
* at the very least this PR should hook these tests to some CI job that runs them.

app/seeds/seeds_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+		for _, entry := range addrs {
+			addr, err := config.ParseNodeAddress(entry)
+			require.NoErrorf(t, err, "%s: %q", chainID, entry)
+
+			// Parsing alone does not cover the port: ParseNodeAddress
+			// substitutes 26657, the RPC port, when one is missing or zero.
+			// A dropped ":26656" would therefore parse clean, pass every other
+			// assertion here, and ship a permanently wrong port to operators.
+			require.EqualValuesf(t, 26656, addr.Port,
```
masih: Ditto re what tests should be and i would probably remove these tests too.

cmd/seid/cmd/bootstrap_peers_test.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+}
+
+// The public networks must actually get a usable set, not just a non-empty
+// string — a silently truncated table would still satisfy the table test above.
+func TestApplyDefaultBootstrapPeersPopulatesPublicNetworks(t *testing.T) {
+	for _, chainID := range []string{"pacific-1", "atlantic-2"} {
+		cfg := tmcfg.DefaultConfig()
+		applyDefaultBootstrapPeers(cfg, chainID)
+		if n := len(strings.Split(cfg.P2P.BootstrapPeers, ",")); n != 3 {
```
masih: ditto re testing rationale.

sei-tendermint/config/node_address.go
```diff
@@ -0,0 +1,19 @@
+package config
+
+import "github.com/sei-protocol/sei-chain/sei-tendermint/internal/p2p"
+
+// NodeAddress is a peer address in `NodeID@host:port` form, as the p2p router
+// understands it. It is re-exported here because the parser lives under
+// sei-tendermint/internal and is unreachable from packages outside the
```
masih: This is unfortunate. 

I would move those basic functions out of internal. My sense is that Tendermint owning P2P _and_ consensus is a historical categorisation that doesn't quite track in general, nor for Sei going forward. P2P transport etc. should have its own root package and exported APIs. 

This is beyond the scope of this PR; no actionables here.

masih (changes requested):
One blocker on `integration` test tag and comments.

app/seeds/seeds_test.go
monty-sei: Yep, fair, and I've swapped it for something that documents the actual rule. It now asserts the address survives a parse round-trip rather than pinning 26656, so the test says "a published address must come back out of the parser unchanged" instead of "the business demands this port".

The reason I couldn't just drop it: `ParseNodeAddress` substitutes 26657 (the RPC port) when a port is missing, so a dropped `:26656` parses clean and would have shipped pointing at the wrong port. The round-trip catches that without the magic number. I sanity checked it by removing a port from the table and watching it fail!

app/seeds/seeds_integration_test.go
monty-sei: Agreed, removed the whole test. @seidroid made the same point from another angle, that it made no network calls at all so the `integration` tag meant it never actually ran, and its parse and port assertions just duplicated the unit test. Its one unique check (that we publish a DNS name rather than a bare host) has moved into `seeds_test.go` where it runs by default, and the integration file is now just the reachability test.

cmd/seid/cmd/bootstrap_peers_test.go
monty-sei: Done, these now assert that init writes `seeds.BootstrapPeers(chainID)` unaltered, which is the bit the wiring is actually responsible for. Cardinality and address shape have moved to `app/seeds` where the table and its rationale live, so adding or retiring a cell doesn't mean editing tests in two places.

Makefile
monty-sei: Yep, removed entirely. @seidroid pointed out the other half of it too, `make lint` isn't a CI job (golangci.yml calls the action directly), so that line only ever fired locally and didn't give the compile check it looked like it was giving. Wiring the tagged package into CI properly is the real fix and I've tracked it separately.

sei-tendermint/config/node_address.go
monty-sei: Appreciate that, and agreed on the direction. For what it's worth the dependency edge already exists today, `sei-tendermint/config` imports `internal/p2p` in autobahn.go, so this doesn't add a new one, but that's a bit beside your point which is that p2p transport probably shouldn't be living behind `internal` in the first place. Happy to leave it as is here.

Makefile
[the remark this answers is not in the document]
monty-sei: Removed the line entirely. You're right that it did not achieve anything: `make lint` is not a CI job (golangci.yml calls the action directly), so it only ever fired locally, and the tagged file still was not linted. @masih independently said the same. Wiring the tagged package into a scheduled job is the real fix and is tracked as a follow up.

sei-tendermint/config/node_address.go
```diff
@@ -0,0 +1,19 @@
+package config
+
+import "github.com/sei-protocol/sei-chain/sei-tendermint/internal/p2p"
+
+// NodeAddress is a peer address in `NodeID@host:port` form, as the p2p router
+// understands it. It is re-exported here because the parser lives under
```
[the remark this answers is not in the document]
monty-sei: Trimmed, the alias doc is now just what the type is. Kept the port-substitution paragraph on `ParseNodeAddress` since, as you say, that one is a caller contract, and it's exactly the footgun that bit this PR.

app/seeds/seeds_integration_test.go
```diff
@@ -0,0 +1,75 @@
...
+				require.NotZerof(t, n, "%s sent an empty preface", hostPort)
+			})
+		}
+	}
+}
+
+// The published address must round-trip through the parser the router uses, and
+// the NodeID must be the 40-hex form the handshake pins. A mismatch here is a
+// dial every operator makes and every operator loses.
+func TestSeedAddressesAreDialableForm(t *testing.T) {
```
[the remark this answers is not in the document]
monty-sei: Good catch, deleted. You're right that it made no network calls, so tagging it meant it never ran, and the parse/port assertions duplicated the unit test. The hostname check has moved into `seeds_test.go` so that invariant now actually runs, and the integration file holds only the reachability test.

cmd/seid/cmd/bootstrap_peers_test.go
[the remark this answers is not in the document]
monty-sei: Done, both literals are gone. The cmd tests now assert only that init writes `seeds.BootstrapPeers(chainID)` unaltered, and cardinality lives in `app/seeds` next to the table.

app/seeds/seeds.go
```diff
@@ -0,0 +1,50 @@
...
+// Seeds are dialled to populate the address book via the PEX reactor and may
+// then be dropped, which is why they belong in `bootstrap-peers` rather than
+// `persistent-peers` — an operator should not hold connections open against
+// them indefinitely.
+package seeds
+
+import "strings"
+
+// chainSeeds maps a well-known chain-id to its Sei Labs seed nodes, each in
+// CometBFT's `NodeID@host:port` form. Three per network, one per cell
```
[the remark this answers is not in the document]
monty-sei: Added, the comment now spells out that unsuffixed `prod` is eu-central-1 alongside prod-euw1 and prod-use2.

cmd/seid/cmd/bootstrap_peers_test.go
```diff
@@ -0,0 +1,128 @@
...
+		{"devnet gets none", "arctic-1", ""},
+		{"unknown chain gets none", "my-private-chain", ""},
+		{"empty chain-id gets none", "", ""},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			cfg := tmcfg.DefaultConfig()
+			applyDefaultBootstrapPeers(cfg, tt.chainID)
+			if cfg.P2P.BootstrapPeers != tt.want {
```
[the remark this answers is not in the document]
monty-sei: Unified, the file uses `require.*` throughout now.

sei-tendermint/config/node_address.go
```diff
@@ -0,0 +1,19 @@
+package config
+
+import "github.com/sei-protocol/sei-chain/sei-tendermint/internal/p2p"
```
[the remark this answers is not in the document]
monty-sei: One correction here: this doesn't add that dependency, it already exists. `sei-tendermint/config` imports `internal/p2p` on main today in autobahn.go (`p2p.NodePublicKey` in `AutobahnValidator`), so the transitive graph and the cycle constraint are the same with or without this file.

The "no production consumer" half is fair though, it is test-only API on a widely imported package. @masih landed on the same file and called it unfortunate but out of scope for this PR, so I've left it in place.

app/seeds/seeds_integration_test.go
monty-sei: Good question, and I went and checked: no, `integration` is a new tag. The repo currently uses `norace, linux, inprocess, rocksdb, dummy, cgo, darwin, mock_balances, ledger, littdb_wip, windows`, and the existing integration suite runs under a different one, `yaml_integration`, driven by `.github/workflows/integration-test-matrix.json`.

I looked at reusing that tag and decided against it: that suite is docker based (`docker exec sei-node-0 ...`) against a locally spun chain, whereas this one dials live public endpoints. Folding them together means either the docker matrix starts making external calls, or this gets dragged into a container lifecycle it does not need.

So for now I have gone the honest route rather than the half wired one: the file header now says plainly that nothing runs or compile checks it and that it is an on demand tool rather than coverage, and I have raised a follow up to add a scheduled (cron plus workflow_dispatch) job.

Scheduled rather than a PR gate, since it depends on live external infrastructure and would otherwise block unrelated PRs whenever a seed hiccups or CI egress is flaky. Sequencing it after the two silent seeds are fixed as well, because wiring it today means the job is red on its first run, which is the quickest way to train everyone to ignore it. Happy to bring it into this PR instead if you would rather it did not land separately!

app/seeds/seeds_integration_test.go
```diff
@@ -0,0 +1,71 @@
+//go:build integration
```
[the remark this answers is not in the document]
monty-sei: Good question, and I went and checked: no, `integration` is a new tag. The repo currently uses `norace, linux, inprocess, rocksdb, dummy, cgo, darwin, mock_balances, ledger, littdb_wip, windows`, and the existing integration suite runs under a different one, `yaml_integration`, driven by `.github/workflows/integration-test-matrix.json`.

I looked at reusing that tag and decided against it: that suite is docker based (`docker exec sei-node-0 ...`) against a locally spun chain, whereas this one dials live public endpoints. Folding them together means either the docker matrix starts making external calls, or this gets dragged into a container lifecycle it does not need.

So for now I have gone the honest route rather than the half wired one: the file header now says plainly that nothing runs or compile checks it and that it is an on demand tool rather than coverage, and I have raised a follow up to add a scheduled (cron plus workflow_dispatch) job.

Scheduled rather than a PR gate, since it depends on live external infrastructure and would otherwise block unrelated PRs whenever a seed hiccups or CI egress is flaky. Sequencing it after the two silent seeds are fixed as well, because wiring it today means the job is red on its first run, which is the quickest way to train everyone to ignore it. Happy to bring it into this PR instead if you would rather it did not land separately!

app/seeds/seeds_integration_test.go
monty-sei: Hey @masih! I've hooked it up rather than leaving it unrun, so there's now a `Seed Reachability` workflow that runs the tagged tests and vets the package so it can't rot. It's manual only, partly because it dials live endpoints and I'm assuming we wouldn't want it gating unrelated PRs.

To get it green today it tolerates one unreachable seed per network, which I believe is actually the more honest assertion anyway, since three per network exist for redundancy and a node bootstraps fine on two (I saw one reach 21 peers with a seed down). Anything tolerated still gets named in the output so nothing quietly hides, and I've raised a follow up to tighten it to zero once the prod seeds are serving again.

Whenever you get a chance it'd be great to get your eyes on it again, and just let me know if you'd rather it went the other way!

app/seeds/seeds_integration_test.go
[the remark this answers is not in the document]
masih: @monty-sei the reply to this comment is identical to the one below. AI burp?

app/seeds/seeds_integration_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+			require.LessOrEqualf(t, len(unreachable), maxUnreachablePerNetwork,
+				"%s: %d of %d seeds are not serving P2P (tolerating up to %d):\n  %v",
+				chainID, len(unreachable), len(addrs), maxUnreachablePerNetwork, unreachable)
+		})
+	}
+}
+
+// speaksP2P dials the seed and waits for it to send its handshake preface.
+func speaksP2P(addr config.NodeAddress) error {
```
masih, line 72: I recommend convention of `requireXYZ` where the helper takes `t` and asserts things for you instead of returning error which then one asserts in the test.

Then the author has the duty to write small sharp `require` assertions that are assembleable across different tests.

.github/workflows/seed-reachability.yml
```diff
@@ -0,0 +1,49 @@
...
+          cache: true
+
+      # Compile-checked as well as run: build-tagged files are skipped by the
+      # untagged vet and by golangci-lint, so without this the package can rot
+      # against an API change and stay green.
+      - name: Vet
+        run: go vet -tags=integration ./app/seeds/...
+
+      - name: Seed reachability
+        run: go test -tags=integration -v -count=1 -timeout=5m ./app/seeds/...
```
masih, line 49: Thanks for adding this monty, we can tidy this up later in terms of requirements for PR landing or alerting if seed nodes are down.

masih (approved):

app/seeds/seeds_integration_test.go
monty-sei: Good convention and I agree with it generally! The only reason this one returns an error is that the tolerance needs to collect failures across all three seeds before it can decide, so a require style helper would end the test at the first failure and never get to count them.

Happy to restructure it as something like `requireQuorum(t, chainID, addrs)` so the assertion lives in the helper instead, just let me know if you'd prefer that!

app/seeds/seeds_integration_test.go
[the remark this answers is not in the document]
monty-sei: Hah that's funny, yep definitely an AI burp here. Please ignore!

PR #3889, opened 2026-08-10
feat(seidb): add state store snapshots

blindchaser:
# feat(seidb): add exact-version online state store snapshots

## Summary

Add opt-in, online State Store (SS) snapshots. Every snapshot is a set of Pebble checkpoints (hardlink trees) whose label is exact: the image contains every version up to the label and nothing above it. The label is pinned by a barrier placed in each backend's apply queue at enqueue time, so the commit path never waits for a checkpoint. This PR covers generation and retention only; SS rollback is out of scope and does not compose with snapshots yet.

- `sei-db/db_engine/types/types.go`: define the capability contract — `Checkpointable`, `DrainBarrier`, `CheckpointVersionSetter`, and the `CheckpointScheduler` composition. `ScheduleCheckpoint` rejects any backend that lacks a capability instead of degrading.
- `sei-db/db_engine/pebbledb/mvcc/db.go`: add `ScheduleAtDrain`, an ordered barrier on the single-writer FIFO. A barrier for version H runs after H is applied and before H+1, without blocking the caller. `Checkpoint` uses a flushed-WAL Pebble checkpoint; `SetCheckpointVersion` stamps the version marker inside the checkpoint copy only and never touches the live marker.
- `sei-db/state_db/ss/cosmos/store.go`, `sei-db/state_db/ss/evm/store.go`: implement checkpoint scheduling for the Cosmos store, the unified EVM store, and separate EVM sub-databases. An idle sub-database is stamped at the label, which is its correct state for that block. Publication waits for every required checkpoint; the first error wins.
- `sei-db/state_db/ss/composite/snapshot.go`: the `snapshotManager` enforces one snapshot in flight, the interval boundary, and the minimum time gate. Publication stages into a `tmp-` directory, renames atomically, fsyncs the parent directory, and moves the `current` symlink forward only. A failure past the barrier abandons that boundary rather than relabeling later state; recovery is the next boundary. Startup removes stale staging directories, resumes from the newest snapshot on disk, restores `current`, and applies retention (`1 + keepRecent`). Enablement is fail-closed: a backend without checkpoint support, or a live database that cannot hardlink into the snapshot root, rejects boot.
- `sei-db/state_db/ss/composite/store.go`, `sei-cosmos/storev2/rootmulti/store.go`: give the trigger one owner. `rootmulti.flush` calls `ScheduleSnapshot` for every block, populated or empty; `ApplyChangesetAsync` deliberately schedules nothing, so import, recovery, pruning, and benchmark callers cannot publish a partial snapshot. A compile-time assertion pins the capability on `CompositeStateStore`, so a wrapper that drops the method fails the build.
- `sei-db/config/ss_config.go`, `sei-db/config/sc_config.go`, `sei-db/config/toml.go`, `app/seidb.go`, `sei-cosmos/server/config/config.go`: one operator knob, `ss-snapshot-enable`, default `false`. Both read sites guard the key, so an `app.toml` rendered before this key existed keeps the default. When enabled, SS mirrors state-commit's effective snapshot interval, minimum time interval, and retention through `AlignSSSnapshotWithSC`; the shared cadence helper heals unset SC values to the memIAVL defaults, so a raw zero can never silently disable SS snapshots. The three derived fields are `mapstructure:"-"` and cannot be set by any key.
- `sei-db/state_db/ss/composite/snapshot_metrics.go`: export `ss_snapshot_*` metrics — attempts, skips by reason, completions by outcome, duration, in-flight, published height, retained count, and apparent bytes (persisted per snapshot in an `.apparent-size` file).
- `sei-db/common/utils/path.go`: default snapshot root is `<home>/data/state_store/snapshots`; a custom SS directory `<db>` moves it to the sibling `<db>-snapshots` so hardlinks stay on one filesystem. Managed snapshot directories have no lease; consumers must coordinate with generation and pruning.

## Test plan

- `sei-db/state_db/ss/composite/snapshot_test.go`: label exactness (every version below the label present, versions written after the boundary excluded while writes continue), empty boundary blocks, idle EVM sub-databases, Cosmos-only and EVM-split layouts, one-in-flight and minimum-time gates, cancellation at the barrier during close, restart resumption from the newest snapshot, stale staging cleanup, out-of-order publication, retention after a failed publish, hardlink preflight rejection across filesystems, and rejection of backends without checkpoint or barrier support.
- `sei-cosmos/storev2/rootmulti/store_test.go`: `flush` schedules a snapshot at a boundary for both a populated and an empty block, pinning the trigger at its single owner.
- `sei-db/db_engine/pebbledb/mvcc/db_test.go`: a checkpoint stamp preserves the live version marker, and a queued checkpoint cancels before work starts.
- `sei-db/config/ss_config_test.go`, `sei-db/config/toml_test.go`, `app/config_fuzz_test.go`, `sei-cosmos/server/config` goldens: default-off rollout, guarded reads, cadence mirroring with zero-healing, and the derived fields pinned as unreachable-by-key.
- Verified locally with `go test -race` on `sei-db/state_db/ss/composite`, `sei-db/config`, `sei-db/db_engine/pebbledb/mvcc`, and `sei-cosmos/storev2/rootmulti`, plus the configuration characterization suites; `gofmt -s` and `goimports` are clean on all changed files.


sei-db/db_engine/pebbledb/mvcc/db.go
```diff
@@ -278,6 +286,58 @@ func (db *Database) PebbleMetrics() *pebble.Metrics {
...
+func (db *Database) SetCheckpointMarkers(destDir string, latest, earliest int64) error {
+	if latest < 0 || earliest < 0 {
+		return fmt.Errorf("versions must be non-negative")
+	}
+	if earliest > latest {
+		return fmt.Errorf("earliest version %d is above latest version %d", earliest, latest)
+	}
+
+	opts := newPebbleOptions(db.config, nil)
+	opts.DisableAutomaticCompactions = true
```
cody-littley: This is a dangerous setting, is it intentional? If we disable pebble compaction, the likely outcome is that the data on disk will continue growing forever, even if we intend to prune it.

app/seidb.go
```diff
@@ -48,6 +48,7 @@ const (
 	FlagSSPruneInterval     = "state-store.ss-prune-interval"
 	FlagSSImportNumWorkers  = "state-store.ss-import-num-workers"
 	FlagSSReadWriteMetrics  = "state-store.ss-enable-read-write-metrics"
+	FlagSSSnapshotEnable    = "state-store.ss-snapshot-enable"
```
yzang2019, line 51: Why do we need this config at all? We should make SS default always taking snapshots right? (In order to support rollback)

sei-cosmos/storev2/rootmulti/store.go
```diff
@@ -48,10 +49,24 @@ var (
...
+// assertion at startup. This pins the only implementation, so wrapping the state
+// store without carrying the method through fails the build here rather than
+// silently ending SS snapshots at runtime.
+var _ stateStoreSnapshotScheduler = (*sscomposite.CompositeStateStore)(nil)
+
 type Store struct {
 	mtx            sync.RWMutex
 	scStore        sctypes.Committer
 	ssStore        seidbtypes.StateStore
+	ssSnapshots    stateStoreSnapshotScheduler
```
yzang2019, line 69: In the future, we will have a global snapshot scheduler to manage both SC and SS snapshot, but I think it's fine to keep as it is today until MemIAVL is fully gone

sei-db/config/toml.go
```diff
@@ -140,6 +140,31 @@ ss-import-num-workers = {{ .StateStore.ImportNumWorkers }}
...
+# Retained snapshots pin referenced SSTs, so compaction cannot reclaim them.
+# Expect steady-state disk overhead on the order of the compaction churn over
+# one snapshot interval per retained snapshot, which is substantial on a
+# multi-TB state store.
+# Managed snapshot directories have no lease in this release. Do not pack an
+# archive or serve state sync directly from them while the node is running:
+# retention can remove a directory during use. Stop the node, or use external
+# coordination that prevents pruning, before consuming a snapshot. Snapshot
+# attempts, skips, outcomes, duration, in-flight state, height, count, and
+# apparent bytes are exported through ss_snapshot_* metrics. Default: false.
```
yzang2019: Let's simplify the comments here

sei-db/config/toml.go
```diff
@@ -140,6 +140,31 @@ ss-import-num-workers = {{ .StateStore.ImportNumWorkers }}
...
+ss-snapshot-enable = {{ .StateStore.SnapshotEnable }}
```
yzang2019, line 155: Suggest to remove that config

sei-db/db_engine/types/types.go
```diff
@@ -120,6 +122,73 @@ type Checkpointable interface {
...
+// SetCheckpointMarkers makes a completed checkpoint self-describing without
+// changing the live database.
+func SetCheckpointMarkers(db StateStore, destDir string, latest, earliest int64) error {
+	setter, ok := db.(CheckpointMarkerSetter)
+	if !ok {
+		return fmt.Errorf("state store backend %T cannot set checkpoint markers", db)
+	}
+	return setter.SetCheckpointMarkers(destDir, latest, earliest)
+}
+
```
yzang2019, line 146: Might be worth to move all these to a separate checkpoint_scheduler.go under management/ folder

sei-db/state_db/ss/evm/store.go
```diff
@@ -370,6 +372,106 @@ func (s *EVMStateStore) Close() error {
...
+	storeTypes := AllEVMStoreTypes()
+	var (
+		mu        sync.Mutex
+		remaining = len(storeTypes)
+		firstErr  error
+	)
+	for _, storeType := range storeTypes {
+		name := StoreTypeName(storeType)
+		dest := filepath.Join(destDir, name)
+		types.ScheduleCheckpoint(s.subDBs[storeType], dest, shouldRun, func(err error) {
```
yzang2019: How do we make sure that all underlying DBs are gonna take the checkpoint for the same height?

sei-db/state_db/ss/composite/snapshot.go
```diff
@@ -0,0 +1,743 @@
...
+	}
+	now := time.Now()
+	m.mu.Lock()
+	previous := m.lastRequested
+	previousRequestAt := m.lastRequestAt
+	var skipReason string
+	accepted := false
+	switch {
+	case m.stopped || version <= m.lastRequested:
+		// A repeated commit-path call is expected and is not a skipped attempt.
```
yzang2019, line 315: Might be worth to log out that we are skipping due to what reason

PR #3891, opened 2026-08-11
fix(config): correct the minimum-gas-prices help text separator

bdchatham:
Stacked on #3890. Base is `test/plt-851-live-node-config-reads`, so the diff shown here is the two production lines plus the assertion they change. Review #3890 first.

## Motivation

`seid start --help` advertises `0.01photino;0.0001stake` as its `minimum-gas-prices` example, and a semicolon is not a separator `sdk.ParseDecCoins` accepts. `cmd/seid/cmd/root.go:296` hands the key to `baseapp.SetMinGasPrices`, which panics on a value `ParseDecCoins` rejects, and that panic is the whole boot.

So an operator pricing a second fee token by copying the flag's own example cannot start a node. Single-denomination values are unaffected, which is the shipped default and nearly every deployment, and is why this has stayed quiet.

The auth spec carries the identical defect at `sei-cosmos/x/auth/spec/01_concepts.md:29`, in the multi-denomination case it is describing.

## Why the comma is the correct direction

The sibling registration already in this tree uses a comma: `sei-ibc-go/testing/simapp/simd/cmd/testnet.go:90` reads `(e.g. 0.01photino,0.001usei)`. That is what identifies the semicolon as an outlier rather than a convention, so this moves toward the tree's own existing form rather than introducing a new one.

Both corrected examples were checked against the parser before being written: `0.01photino,0.0001stake` and `0.00001stake,0.05photinos` both parse, and their semicolon forms both return `invalid decimal coin expression`.

## Risk

Only the separator changes, so every value that started a node before still starts one. The change strictly adds a working example where a broken one stood. A tree-wide sweep finds no remaining semicolon-separated `minimum-gas-prices` examples.

## The assertion changes with it

The characterization suite pinned this as a defect, so the assertion that recorded it has to move in the same change rather than be loosened. It previously asserted the help text contains the broken literal. It now reads whatever example the help text advertises and hands it to the reader that has to accept it, which is the invariant the fix establishes and re-arms against any later edit to the text.

Verified by mutation: reverting the help text to the semicolon fails the test, naming the value and the panic. Changing the example to a different invalid value fails. Removing the example entirely fails.

### DIFF
--- a/cmd/seid/cmd/mingasprices_config_fuzz_test.go
+++ b/cmd/seid/cmd/mingasprices_config_fuzz_test.go
@@ -14,26 +14,21 @@ import (
 	serverconfig "github.com/sei-protocol/sei-chain/sei-cosmos/server/config"
 )
 
-// minimum-gas-prices is the one key in this tree whose accepted syntax is contradicted by its own
-// documentation, and this file holds all three parties still.
+// minimum-gas-prices is resolved by one reader whose way of rejecting a value is to panic, and this
+// file holds that reader, the example the flag advertises, and the one artifact that still describes
+// a different syntax.
 //
-// One reader governs a node. root.go:296 hands cast.ToString of the key to baseapp.SetMinGasPrices,
-// which calls sdk.ParseDecCoins and panics on anything it cannot parse (options.go:24-28). That
-// panic is the whole boot, and ParseDecCoins separates denominations with a comma.
+// root.go:296 hands cast.ToString of the key to baseapp.SetMinGasPrices, which calls
+// sdk.ParseDecCoins and panics on anything it cannot parse (options.go:24-28). That panic is the
+// whole boot, so an example an operator is shown has to be a value that parses. The assertion below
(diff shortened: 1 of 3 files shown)


bdchatham:
Closing as deferred rather than rejected. The defect is real and the fix is verified; it is simply not worth the review cycle right now against the rest of the queue.

Nothing is lost. The branch \`fix/min-gas-prices-separator\` is left in place at \`490e5ef9d\`, two commits, each vetted and mutation-verified: reverting the help text fails the assertion, a differently-invalid example fails, and removing the example fails. Picking this up is a reopen rather than a rewrite.

One thing whoever picks it up should know: the assertion change has to travel with the production change. #3890 pins the current help text as a recorded defect, so correcting the text without moving that assertion in the same commit leaves a red test. That is the suite's rule about not loosening an assertion when the pinned behaviour is deliberately changed, and it is why this could not be a two-line production diff on its own.

Tracked as item 1 of PLT-976.

PR #3892, opened 2026-08-11
fix: harden mempool proposal gas accounting

codchen:
## What changed

- reject transactions whose `GasWanted` exceeds a positive `MaxGasWanted` proposal cap
- refresh cached priority, gas-wanted, and estimated-gas metadata after a successful mempool recheck
- skip proposal candidates that do not fit the remaining byte or gas budgets instead of ending the entire selection pass
- preserve EVM nonce dependencies when a candidate is skipped

## Why

The original version of this PR addressed a state-dependent mismatch for fee-exempt `MsgAssociate` transactions. Current `main` has removed gasless Cosmos transactions: `CheckTx` now installs the transaction's declared gas meter, and proposal validation counts the declared gas for Cosmos transactions. The stale zero-gas associate path therefore no longer exists.

After rebasing, the associate-specific gas formula, fee-exemption branches, oracle special cases, and execution-meter overrides have all been removed from this PR. The remaining changes are state-agnostic mempool hardening:

- a transaction that can never fit under the effective proposal gas-wanted cap is not retained
- a transaction whose application metadata changes during recheck cannot keep stale gas accounting
- one oversized or otherwise unfittable candidate does not prevent later independent transactions from filling the proposal

For EVM transactions, skipping a nonce blocks later nonces from the same account during that reap pass, while transactions from other accounts remain eligible.

## Validation

- `go test ./internal/mempool/... ./internal/state` in `sei-tendermint`
- `go test --run='^$' ./...` in `sei-tendermint`
- touched Go files pass `gofmt -s` and `goimports`
- `git diff --check`

### DIFF
--- a/sei-tendermint/internal/mempool/mempool.go
+++ b/sei-tendermint/internal/mempool/mempool.go
@@ -33,6 +33,26 @@ const (
 	MinGasEVMTx  = 21000
 )
 
+type checkedTxMetadata struct {
+	priority     int64
+	gasWanted    int64
+	estimatedGas int64
+}
+
+// metadataFromCheckTx normalizes application metadata identically for initial checks and
+// rechecks. Reusing this step prevents proposal selection from retaining stale gas values.
+func metadataFromCheckTx(res *abci.ResponseCheckTxV2) checkedTxMetadata {
+	estimatedGas := res.GasEstimated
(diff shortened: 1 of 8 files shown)


codchen:
@seidroid review

codchen:
closing because the issue no longer exists with gasless txs removed

PR #3894, opened 2026-08-11
ci: shard Go race tests into a round-robin 3-way matrix (PLT-959)

amir-deris:
## Summary

- Race Detection previously ran all ~420 packages in a single job (~18 min baseline). It is now sharded into a 3-way matrix in `go-test.yml`, with an aggregate check that keeps the **"Race Detection"** status name so branch protection does not need to change.
- Packages are split **round-robin** (`package i → shard i % NUM_SPLIT`) via the Makefile's `split-test-packages` target.
- A short list of consistently slow packages (e.g. IBC keeper packages, `sei-db` engine tests, `giga/tests`, `evmrpc/tests`) is round-robined separately, ahead of the rest, so they cannot all land on the same shard by coincidence.
- `NUM_SPLIT=3 make test-group-N` reproduces a given CI shard's **package split** locally (local runs add coverage flags and omit CI's ledger build tags).
- Preserves `-race`, `-tags=ledger,test_ledger_mock`, and the `occ_tests` → `-parallel=1` special case per shard.
- Includes every package from `go list ./...` (not just test-bearing ones), so the race job still compiles the full tree under `-race -tags=ledger,test_ledger_mock`.

## Results

Latest run: shard 0 = 13m0s, shard 1 = 12m28s, shard 2 = 10m41s (max ~13 min), down from the ~18 min single-job baseline — roughly 28% CI runner time reduction for this job.

## Context

Linear: [PLT-959](https://linear.app/seilabs/issue/PLT-959)

Starting at `NUM_SPLIT=3`; can scale up further once shared-runner capacity is confirmed. Also explored duration-aware bin-packing via a `testsplit` tool and reverted — round-robin with a heavy-package pre-pass was simpler and fast enough.

## Test plan

- [x] Confirm the run fans out into "Race Detection (shard N)" jobs plus the aggregate "Race Detection" check
- [x] Confirm all shards pass and the aggregate check reflects failures correctly
- [x] Confirm wall-clock improvement over the single-job baseline
- [x] Confirm `NUM_SPLIT=3 make test-group-0` locally reproduces shard 0's package set


amir-deris:
Closing this — sharding race detection into a 3-way matrix isn't worth the fleet cost right now.

**Why**

- Each PR would consume **3 concurrent `uci-default` slots** (plus coverage) instead of 1 — roughly **2× total runner-minutes** (~36 vs ~18) to save ~5 min of wall clock when runners are free.
- When the pool is saturated (as we've seen on this branch), sharding makes queue depth worse: every open PR needs multiple slots before race tests even start, which slows everyone down.
- The Makefile/local-repro improvements (`split-test-packages`, `test-group-N`, heavy-package round-robin) are useful on their own and don't require CI sharding to land.

**If we revisit**

- Add `uci-default` capacity first, or
- Shard on `main` only and keep a single job on PRs, or
- Land Makefile parity changes without changing CI job count.

Happy to reopen or split out the Makefile-only pieces if that's useful.

PR #3897, opened 2026-08-11
Test the new explict trigger after first review

masih:
Test the new explicit ask trigger review after the initial review by seidroid.

See: https://github.com/sei-protocol/uci/pull/57


masih:
@seidroid review

masih:
* working as expected for PRs.
* explicit comment can't be tested here because it always pulls from the default branch so the UCI version on main needs updating. 

Closing since this PR was for testing only.

PR #3902, opened 2026-08-11
Snapshot Engine + flatKV integration: phase 1

cody-littley:

Integrate flatKV and the new snapshot engine.

As part of the pipelining refactor, our goal is to move three things off of the main execution thread:

- hashing
- pebble checkpointing (i.e. on-disk snapshots)
- flushing data down into pebble

In order to limit the size of this PR, this PR only moves flushing of data off the main thread, although it does lay the groundwork for moving hashing and checkpointing off-thread.


sei-db/db_engine/snapshot/shard.go
```diff
@@ -279,29 +280,15 @@ func (s *shard) iteratorClosed() {
...
-	if s.openIterators > 0 {
-		return fmt.Errorf("cannot write while %d iterator(s) are open; close them first",
-			s.openIterators)
-	}
-	return nil
-}
-
 // Set sets the value for the given key at the current version.
+//
+// A write to a shard that is out of service is refused: it would land in versioned data that no
```
yzang2019, line 300: What  does out of service mean for a shard? When would a shard become out of service?

sei-db/state_db/sc/flatkv/store_apply.go
```diff
@@ -33,6 +40,22 @@ func (s *CommitStore) ApplyChangeSets(version int64, changeSets []*proto.NamedCh
 	}
 	// Blocks are contiguous and the first block is 1, so writes always land at committedVersion+1. See the
 	// Commit contract: a store whose history starts higher is seeded by SetInitialVersion.
+	// An empty batch for a block that is already committed is accepted and does nothing.
+	if version > 0 && version == s.committedVersion {
```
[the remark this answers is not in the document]
cody-littley: Will not fix. This is nitpicking at problems that are not real problems in production code.

sei-db/state_db/sc/flatkv/store.go
```diff
@@ -631,96 +660,289 @@ func (s *CommitStore) acquireFileLock(dir string) error {
...
+		s.localMeta[dir] = meta
 	}
-	toClose = append(toClose, s.codeDB)
+	return nil
+}
 
-	s.storageDB, err = s.openPebbleDB(&s.config.StorageDBConfig, &s.config.StorageCacheConfig)
-	if err != nil {
-		return fmt.Errorf("failed to open storage DB: %w", err)
+// openStores wraps the five already-open PebbleDBs in snapshot engines. It is the last step of
```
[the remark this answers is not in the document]
cody-littley: tweaked comment

sei-db/state_db/sc/flatkv/config/config.go
```diff
@@ -106,6 +111,17 @@ type Config struct {
...
 }
 
+// MetaKeyPrefix is the key namespace FlatKV reserves for per-database metadata, and which each
+// snapshot engine owns: Finalize writes land under it and iteration filters it out. It matches
+// ktype.MetaKeyPrefixBytes, restated here because ktype imports this package's siblings.
+const MetaKeyPrefix = "_meta/"
+
+// defaultStoreConfig returns the snapshot engine defaults for one database, named for the database's
+// directory so metrics and per-database hash bookkeeping can tell the stores apart.
+func defaultStoreConfig(name string) snapshot.SnapshotEngineConfig {
```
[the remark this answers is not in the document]
cody-littley: will not fix

sei-db/state_db/sc/flatkv/store.go
```diff
@@ -774,12 +996,43 @@ func (s *CommitStore) PendingVersion() int64 {
...
+// caller asking for one is therefore asking for the block to be committed, and gets it.
+//
+// This exists for Cosmos, which asks for the hash before it calls Commit. Committing early is safe
+// there because every one of the block's writes has already arrived: rootmulti's GetWorkingHash begins
+// by flushing every buffered changeset into this store, and only then reads the hash. The Commit that
+// follows finds the block already committed and does nothing (see Commit).
+//
+// Post-Cosmos this goes away along with rootmulti: a single call will supply a block's writes and
+// commit them, and nothing will ask for a hash mid-block.
 func (s *CommitStore) RootHash() []byte {
```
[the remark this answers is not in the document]
cody-littley: will not fix

sei-db/db_engine/snapshot/shard.go
```diff
@@ -422,6 +410,12 @@ func (s *shard) materializeCurrentOverrides() ([]kvPair, error) {
 		if deque.IsEmpty() {
 			continue
 		}
+		if lowerBound != nil && key < string(lowerBound) {
```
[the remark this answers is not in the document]
cody-littley: no

sei-db/state_db/sc/flatkv/store.go
```diff
@@ -631,96 +660,289 @@ func (s *CommitStore) acquireFileLock(dir string) error {
...
+		return err
+	}
+	s.miscStore, err = open(&s.config.MiscStoreConfig, dbs.misc)
+	if err != nil {
+		return err
 	}
-	toClose = append(toClose, s.miscDB)
 
-	s.metadataDB, err = s.openPebbleDB(&s.config.MetadataDBConfig, &s.config.MetadataCacheConfig)
+	metaCfg := s.config.MetadataStoreConfig
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -141,265 +137,257 @@ func (s *CommitStore) Commit(version int64) (committed int64, err error) {
...
-		batch types.Batch
-	}
-	var pendingBuf [4]pendingCommit
-	pending := pendingBuf[:0]
+	snapshots := make(map[string]snapshot.Snapshot, len(s.stores))
 	defer func() {
-		for _, p := range pending {
-			_ = p.batch.Close()
+		if retErr != nil {
+			// An error in this function is non-recoverable. Outer scope is responsible for teardown.
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -93,42 +103,28 @@ func (s *CommitStore) Commit(version int64) (committed int64, err error) {
...
-	// Periodic snapshot so WAL stays bounded and restarts are fast.
+	// Periodic snapshot so WAL stays bounded and restarts are fast. A failure fails the commit: the
+	// flush wait inside WriteSnapshot is where a dead store surfaces, and a block whose data will never
+	// reach disk must not be reported as committed. The block is already durable in the WAL, so replay
+	// reconciles whatever the caller's halt leaves behind.
 	if s.config.SnapshotInterval > 0 && version%int64(s.config.SnapshotInterval) == 0 {
 		s.phaseTimer.SetPhase("commit_write_snapshot")
 		if err := s.WriteSnapshot(""); err != nil {
-			logger.Error("auto snapshot failed", "version", version, "err", err)
+			return version, fmt.Errorf("auto snapshot at version %d: %w", version, err)
```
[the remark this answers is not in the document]
cody-littley: no

sei-db/state_db/sc/flatkv/verify.go
```diff
@@ -79,8 +73,10 @@ func verifyLtHashInternal(cs *CommitStore) error {
 // membership predicate foldChunk / serializeKV use for LtHash MixIn — so the
 // scan is directly comparable to the maintained per-module metadata. Module
 // membership uses the same physical-key routing the write path uses.
-func scanDBByModule(db seidbtypes.KeyValueDB) (map[string]*lthash.LtHash, map[string]lthash.ModuleStats, error) {
-	iter, err := db.NewIter(&seidbtypes.IterOptions{})
+func scanStoreByModule(
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/state_db/sc/composite/store_migration_test.go
```diff
@@ -387,9 +387,12 @@ func isZeroTestValue(v []byte) bool {
 
 func pruneZeroStorageViaRoutedGet(t *testing.T, cs *CompositeCommitStore, limit int) (int, int) {
 	t.Helper()
+	// The iterator is closed before the deletes are applied. FlatKV's iterators are a live merge over
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/db_engine/snapshot/shard.go
cody-littley: Two scenarios qualify as "out of service":

- The shard is shutdown. Only reachable if you call shutdown concurrently with an operation that touches a shard (which is illegal).
- The shard has crashed. Only possible when the DB experiences a crash. DB failures are fatal, so we should never try to recover them.

Updated the doc to spell this out.

sei-db/state_db/sc/flatkv/store.go
```diff
@@ -592,96 +660,288 @@ func (s *CommitStore) acquireFileLock(dir string) error {
...
+		height := int64(0)
+		if meta := s.localMeta[dir]; meta != nil {
+			height = meta.CommittedVersion
+		}
+		heights[dir] = height
+		if height < lowest {
+			lowest = height
+		}
+	}
+	// The metadata database records the store-wide committed version, so that is its own height.
```
yzang2019: Shall we just remove the metadata DB?

sei-db/state_db/sc/flatkv/store.go
```diff
@@ -735,12 +995,43 @@ func (s *CommitStore) PendingVersion() int64 {
...
+//
+// This exists for Cosmos, which asks for the hash before it calls Commit. Committing early is safe
+// there because every one of the block's writes has already arrived: rootmulti's GetWorkingHash begins
+// by flushing every buffered changeset into this store, and only then reads the hash. The Commit that
+// follows finds the block already committed and does nothing (see Commit).
+//
+// Post-Cosmos this goes away along with rootmulti: a single call will supply a block's writes and
+// commit them, and nothing will ask for a hash mid-block.
 func (s *CommitStore) RootHash() []byte {
+	if err := s.commitPendingBlock(); err != nil {
```
blindchaser: is looks RootHash commits the pending block. then later composite commit increments FlatKV again? is there a double commits on flatkv?

sei-db/state_db/sc/flatkv/store.go
```diff
@@ -296,6 +287,45 @@ func NewCommitStore(
...
+	if c.StorageDBConfig.DataDir == "" {
+		c.StorageDBConfig.DataDir = filepath.Join(workDir, storageDBDir)
+	}
+	if c.MiscDBConfig.DataDir == "" {
+		c.MiscDBConfig.DataDir = filepath.Join(workDir, miscDBDir)
+	}
+	if c.MetadataDBConfig.DataDir == "" {
+		c.MetadataDBConfig.DataDir = filepath.Join(workDir, metadataDir)
+	}
+	applyPebbleMetricsConfig(c)
```
blindchaser: Config.Fsync is not passed to the five snapshot-engine FlushSync fields. Since FlushSync defaults to false, programmatic callers that set Fsync=true still get unsynchronized normal flushes. can we propagate this value to all five store configurations and add a test.

sei-db/state_db/sc/flatkv/store.go
```diff
@@ -735,12 +995,43 @@ func (s *CommitStore) PendingVersion() int64 {
...
+// caller asking for one is therefore asking for the block to be committed, and gets it.
+//
+// This exists for Cosmos, which asks for the hash before it calls Commit. Committing early is safe
+// there because every one of the block's writes has already arrived: rootmulti's GetWorkingHash begins
+// by flushing every buffered changeset into this store, and only then reads the hash. The Commit that
+// follows finds the block already committed and does nothing (see Commit).
+//
+// Post-Cosmos this goes away along with rootmulti: a single call will supply a block's writes and
+// commit them, and nothing will ask for a hash mid-block.
 func (s *CommitStore) RootHash() []byte {
```
yzang2019: RootHash should be a read-only function, but here we are committing pending block before we return, which breaks the assumption and contract.

sei-db/state_db/sc/flatkv/store.go
cody-littley: This is intentional on my part. The core problem is that comsos wants the hash before the commit, but the proper shape of the storage engine is to provide a hash only after the block is finalized+committed. 

I've set it up so that if we call Commit() multiple times for the same block number, later commits become no-ops. If you try to write changes to the block after the commit, you get an error. Previously it was technically possible to write changes after we get the hash but before the commit, but this would either mean the hash is actually an invalid hash for that block.

Worth a discussion in parking lot.

sei-db/state_db/sc/flatkv/config/config.go
```diff
@@ -66,32 +66,37 @@ type Config struct {
...
+	MiscStoreConfig snapshot.SnapshotEngineConfig
 
 	// MetadataDBConfig defines the PebbleDB configuration for the metadata database.
 	MetadataDBConfig pebbledb.PebbleDBConfig
 
-	// MetadataCacheConfig defines the cache configuration for the metadata database.
-	MetadataCacheConfig dbcache.CacheConfig
+	// MetadataStoreConfig defines the snapshot engine configuration for the metadata database. The store
+	// owns this database's read cache and write staging, so its MaxSize is that database's cache budget.
+	MetadataStoreConfig snapshot.SnapshotEngineConfig
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/state_db/sc/flatkv/store.go
cody-littley: good point, fixed

sei-db/state_db/sc/flatkv/store.go
cody-littley: Refactored. RootHash() no longer commits. Instead, the outer context is responsible for committing block N before fetching that block's hash. I've added an additional block number return value for RootHash() so that the outer scope can assert that its getting the hash for the block its expecting.

sei-db/state_db/sc/flatkv/store.go
cody-littley: removed (separate PR, merged into this branch)

sei-cosmos/storev2/rootmulti/store.go
```diff
@@ -730,7 +759,15 @@ func (rs *Store) SetInterBlockCache(_ types.MultiStorePersistentCache) {}
...
-	return rs.scStore.SetInitialVersion(version)
+	if err := rs.scStore.SetInitialVersion(version); err != nil {
+		return err
+	}
+	// A chain seeded to begin at this height has version-1 behind it, and nextVersion is what tells the
+	// commit store which block is being built. Leaving it at 0 would ask for block 1 on a chain whose
+	// first block is this one. The backends cannot supply this: flatkv reflects the seed immediately
+	// while memiavl does not apply it until its first commit, so they disagree until then.
+	rs.lastCommitInfo.Version = version - 1
+	return nil
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-cosmos/storev2/rootmulti/store.go
```diff
@@ -248,11 +273,20 @@ func (rs *Store) flush() error {
...
-	// nil is normalized to an empty (non-nil) set so an empty block records the stable empty-changeset
-	// hash rather than a nil one.
+	// baseapp requests the working hash in FinalizeBlock and again in Commit before committing, so flush
+	// runs three times per height and PopChangeSet has already drained the block's writes by the second
+	// run. Those later runs must not hand an empty changeset down: the commit store stamps it with a
+	// height it derives from its own last committed block, which the first working-hash request already
+	// advanced, so it would conclude the chain had moved to the next block and commit one that never
+	// existed. A later run that does carry writes is handed down instead of refused here — whether the
+	// block is still open to them is the commit store's to judge, and flatkv refuses a sealed one.
+	if len(changeSets) == 0 && rs.flushedVersion == currentVersion {
```
yzang2019, line 279: FlatKV only refuses when the batch is stamped at the already-committed height N. The writer stamps N+1, so it silently defers instead. Someone reading this comment would conclude the non-empty case is caught downstream, and it isn't.

And the safety property currently rests on a convention — "nothing writes to the multistore in the preCommitHandler" — rather than an invariant. Whoever adds the next preCommitHandler hook, or a module that writes during it, gets silent height drift rather than an error. Given the PR's own stated principle about guarding at the choke point, turning flushedVersion == currentVersion && len(changeSets) > 0 into an error in flush() would make it checkable. That's a small, self-contained follow-up, not something that needs to hold up phase 1.

Might be worth adding a comments for context

sei-db/state_db/sc/composite/store.go
```diff
@@ -1063,23 +1083,54 @@ func (cs *CompositeCommitStore) shouldIncludeMemiavlInfos() bool {
...
+	hash, hashed := cs.flatKV.RootHash()
+	if hashed != version {
+		panic(fmt.Sprintf(
+			"composite: flatkv hashed block %d but the chain is building block %d", hashed, version))
+	}
+	return hash
+}
+
 // LastCommitInfo returns the last commit info
 func (cs *CompositeCommitStore) LastCommitInfo() *proto.CommitInfo {
```
yzang2019, line 1165: This PR moves the FlatKV seal earlier: flatKVWorkingHash now calls flatKV.Commit(version) from inside GetWorkingHash, which runs in FinalizeBlock. memIAVL still commits later, in ABCI Commit. So for the whole gap between those two ABCI calls, FlatKV is at block N while memIAVL is at N-1.

The result is a CommitInfo labeled Version: N-1, carrying memIAVL's store hashes at N-1, with an evm_lattice entry holding block N's hash. That combination corresponds to no committed state at any height. Note the contrast with flatKVWorkingHash, which does compare hashed != version and panics — the height check exists exactly where the race can't happen, and is dropped where it can.

It's reachable. rootmulti.Query recomputes commit info live on the latest path, under no lock, and RPC nodes serve queries concurrently with block execution.
`if latest {
	// latest never needs historical LoadVersion clone
	store = types.Queryable(commitment.NewStore(rs.scStore.GetChildStoreByName(storeName)))
	commitInfo = convertCommitInfo(rs.scStore.LastCommitInfo())
	commitInfo = amendCommitInfo(commitInfo, rs.storesParams)
}`

The race technically predates the PR (memIAVL and FlatKV committed sequentially inside composite.Commit, so they disagreed for a few microseconds), but this widens it from an instruction window inside one function to the full inter-ABCI-call gap, on every block.

yzang2019 (approved):

sei-cosmos/storev2/rootmulti/store.go
cody-littley: Updated docs.

I agree this is fragile, but it's tricky to write pipelined code without assuming a "well behaved" write pattern like this. I'm banking on the fact that we should mostly just be maintaining cosmos modules, not extending them, until its time to switch to autobahn.

sei-db/state_db/sc/composite/store.go
cody-littley: Fixed. Composite layer no longer returns commit info where DBs are out of sync with each other. It does this by making a copy of the commit info at commit time and serving that for queries, instead of re-deriving it at query time from the sub-DBs (which can be inconsistent).

PR #3917, opened 2026-08-13
Sei-load: configurable load generators

kollegian:
Integrates configurable load runners for sei load. 

No source code change.


integration_test/load_generator/deploy/helm/sei-load-generator/templates/_helpers.tpl
```diff
@@ -0,0 +1,42 @@
+{{- define "sei-load-generator.name" -}}
```
masih: I would just move all of these to platform repo (and probably use kustomize) to save the hassle of having to set up helm repository and build process.

masih (commented):
Love that AI has made @kollegian into a SUPER USER! awesome end-to-end work. the only blocker for me is to simply move pieces of this work into different repos and leave the code here to be purely about the implementation of the load generation itself

We also need to figure out where the containers for this are being published and pulled from.

Super thanks to @bdchatham for helping out on this one 🙌

PR #3928, opened 2026-08-14
Remove metadata DB

cody-littley:

Remove the metadata DB.


sei-db/state_db/sc/flatkv/store_meta.go
```diff
@@ -430,74 +348,47 @@ func (s *CommitStore) SetInitialVersion(initialVersion int64) error {
...
-}
-
-// readVersionRecord reads one 8-byte big-endian version record out of the
-// working metadata DB under dir, opening and closing that single PebbleDB
-// around the read. An absent directory or key reads as 0.
-func readVersionRecord(dir string, key []byte) (int64, error) {
-	metaDir := filepath.Join(dir, workingDirName, metadataDir)
-	if _, err := os.Stat(metaDir); err != nil {
+	miscDir := filepath.Join(dir, workingDirName, miscDBDir)
+	if _, err := os.Stat(miscDir); err != nil {
```
yzang2019: why do we only look at miscDB for getting LatestVersion? Shall we look at the minimum of four?

sei-db/state_db/sc/flatkv/store.go
```diff
@@ -713,15 +684,56 @@ func (s *CommitStore) loadGlobalMetadata() error {
...
+func (s *CommitStore) requireAlignedDataDBs() error {
+	for _, dbDir := range dataDBDirs {
+		meta := s.localMeta[dbDir]
+		if meta == nil {
+			return fmt.Errorf("flatkv: %s has no local metadata after load", dbDir)
+		}
+		if meta.CommittedVersion != s.committedVersion {
+			return fmt.Errorf(
+				"flatkv: %s is at version %d but the store is at %d; this store holds a block "+
+					"its write-ahead log lost, which no replay can reconcile (restore from a snapshot)",
```
yzang2019: SetInitialVersion writes the same watermark into four separate databases with four separate commits, and the new requireAlignedDataDBs guard refuses to open a store whose four watermarks disagree — so a crash in the middle of that loop produces exactly the state the guard rejects, and the code that would repair it lives after the load that now fails.

sei-db/state_db/sc/flatkv/store_meta.go
```diff
@@ -358,34 +298,12 @@ func (s *CommitStore) SetInitialVersion(initialVersion int64) error {
...
-		if err := batch.Commit(types.WriteOptions{Sync: s.config.Fsync}); err != nil {
-			_ = batch.Close()
-			return fmt.Errorf("flatkv: SetInitialVersion: persist earliest version: %w", err)
-		}
-		_ = batch.Close()
-		s.earliestVersion = seededVersion
-	}
-
 	syncOpt := types.WriteOptions{Sync: s.config.Fsync}
 	for _, ndb := range s.namedDataDBs() {
```
yzang2019, line 344: SetInitialVersion writes each data DB's watermark in its own separate durable batch — four independent Pebble commits, no atomicity across them. What if the node crash in between?

yzang2019 (approved):

sei-db/state_db/sc/flatkv/store_replay.go
```diff
@@ -13,12 +13,22 @@ import (
...
-// targetVersion <= 0, and then persists the result so a later open does not replay it again.
+// targetVersion <= 0, and rejects a store whose data DBs did not all reach that version.
 //
-// It runs at startup (open/openTo) and during Rollback, never concurrently with live commits, so it reads the
-// WAL unlocked. A nil WAL is legal only if this store already sits at targetVersion — the outer context owns
-// the WAL pipeline in that case.
-func (s *CommitStore) replayIntoMutableStore(targetVersion int64) (err error) {
+// It runs at startup (open/openTo) and during Rollback, never concurrently with live commits.
+func (s *CommitStore) replayIntoMutableStore(targetVersion int64) error {
+	if err := s.catchUpFromWAL(targetVersion); err != nil {
```
yzang2019, line 20: First, replay bypasses the contiguity check that normally protects the watermark. Commit refuses to go backwards — it enforces version != s.committedVersion+1. But replay doesn't go through Commit; it calls commitBatches directly.

Second, prepareBatch only skips a DB when it has both no writes and is already at or past the version. The && is what matters here — if the replayed block touches this DB at all, the batch is built even though the DB is ahead.

Third, the batch stamps the replayed version unconditionally. There is no "only move forward" 

So a DB sitting at version 3 that gets block 2 replayed into it is written back down to version 2.

The check has to happen before catchup, and it has to compare against the WAL rather than against the other DBs. Comparing DBs to each other pre-catchup would be wrong, because a plain torn commit legitimately leaves them disagreeing and that case is fully recoverable.

sei-db/state_db/sc/flatkv/store.go
cody-littley: Fixed.

During startup, the code now checks to see if DBs have seen 1 or more blocks (we can tell due to the hash key). If no DB has observed a block and we don't have equal block numbers, it infers that it is the result of a torn genesis and automatically repairs things.

sei-db/state_db/sc/flatkv/store_meta.go
cody-littley: I've now fixed this. Not by making them atomic (which is nontrivial). Rather, the code detects a "torn genesis" (i.e. when we crash in this region), it automatically cleans up and continues. It's possible to differentiate between a crash when setting up the DBs and a crash later.

sei-db/state_db/sc/flatkv/store_meta.go
cody-littley: Carry over from previously where we were just checking the metadata DB. Fixed now. We now look at the snapshot directory and WAL to figure out which version we will load, not just an individual DB.

yzang2019 (approved):

sei-db/state_db/sc/flatkv/store.go
blindchaser: is the fix pushed yet?

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -441,8 +429,5 @@ func (s *CommitStore) FinalizeImport(version int64) error {
 	s.workingLtHash = globalHash
 	s.committedVersion = version
 	s.committedLtHash = s.workingLtHash.Clone()
-	if err := s.commitGlobalMetadata(version, s.committedLtHash); err != nil {
```
blindchaser, line 444: this write was load-bearing for crash recovery because it ran last. After a crash inside the loop above, the global watermark was still 0, loadGlobalMetadata adopted that, and committedVersion was only ever lowered, never raised, so a DB left at the import height got ignored and the store opened as fresh. With the write gone, that same skew reaches `requireAlignedDataDBs` and the open is refused. Import bypasses the WAL, so no replay can reconcile it.

sei-db/state_db/sc/flatkv/store.go
cody-littley: not yet pushed, will be pushed soon

sei-db/state_db/sc/flatkv/store_replay.go
cody-littley: should be fixed now

sei-db/state_db/sc/flatkv/store_write.go
cody-littley: My LLM reviewer arrived at the same conclusion. I've refactored startup, so we should be safe here now.

blindchaser (approved):

sei-db/state_db/sc/flatkv/store.go
```diff
@@ -499,9 +493,61 @@ func (s *CommitStore) openTo(catchupTarget int64) error {
...
+			unreachable = append(unreachable, fmt.Sprintf("%s at %d", dbDir, meta.CommittedVersion))
+		}
+	}
+	if len(unreachable) == 0 {
+		return nil
+	}
+
+	logger.Warn("FlatKV rebuilding the working copy: a data DB holds blocks nothing can account for",
+		"dataDBs", strings.Join(unreachable, ", "), "highestReachableVersion", reachable)
+	return s.rebuildWorkingCopy()
```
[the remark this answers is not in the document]
cody-littley: Will not fix. It's a very small window of "vulnerability", and the consequence is just that we waste some time repeating the state sync.

PR #3932, opened 2026-08-17
Split the musl cgo link path by architecture

monty-sei:
Prerequisite for shipping a static `linux/arm64` seid binary. It changes three words, adds three five-line files, and adds a test.

## The bug

The glibc link path is arch-split (`link_glibclinux_aarch64.go` and `link_glibclinux_x86_64.go`), but the musl path never was. `link_muslc.go` carried `linux && muslc && !sys_wasmvm` with no arch term, so a `linux/arm64` muslc build selected it, emitted `-lwasmvm_muslc`, resolved the **x86_64** archive, and died at link time with incompatible-archive errors.

The `libwasmvm*_muslc.aarch64.a` archives have been checked in for months, referenced by nothing.

## The change

Constrain the three existing files to `amd64`, and add three aarch64 siblings pointing at the archives already in the tree. The new files are byte-identical to CosmWasm/wasmvm v2.1.0's equivalents, since the vendored tree here predates that upstream fix.

## Validation

Each of the six changes was tested in isolation rather than as one bundle:

| Mutation | `linux/arm64 muslc` selection | Effect |
|---|---|---|
| (correct) | 1 | links |
| remove any one new file | **0** | undefined references |
| drop `amd64` from any one file | **2** | both archives on one link line |

Real links in the pinned release image (Alpine 3.23.3, gcc 15.2.0, native arm64) confirm the model predicts linker behaviour rather than just build-tag arithmetic. The baseline fails with 3x `incompatible`, the "new files but no constraints" variant fails with 6x (double, one per extra archive), "constraints but no new files" fails with `undefined reference to version_str`, and the full change produces `ELF 64-bit LSB executable, ARM aarch64, statically linked`.

**amd64 is provably untouched.** For `GOARCH=amd64` the selected files and the resulting cgo `LDFLAGS` are byte-identical before and after, across all three packages. An actual amd64 static build via unmodified `scripts/build-static.sh` succeeds with the change applied.

**No consensus impact observed.** A mixed fleet of three binaries built from this branch (arm64 static musl, amd64 static musl, arm64 dynamic glibc) on one chain agreed on every app hash compared, including the blocks carrying a wasm store/instantiate/execute cycle. Separately, the seven consensus-relevant packages that swap amd64 assembly for portable Go on arm64 all pass their upstream vector suites on both architectures.

## Reviewer notes

- **This lands inert.** Nothing in CI, goreleaser or the Dockerfile builds `linux/arm64` with the muslc tag yet. It unblocks that work, it does not ship an arm64 binary.
- **Behaviour change for exotic architectures.** 386, riscv64, s390x and ppc64le previously matched the amd64 muslc directive and failed with an incompatible-archive error. They now match no link file and fail with undefined references instead. That matches what the glibc path has always done, and seid cannot build on those architectures anyway, since `giga/executor/lib` rejects them at compile time with a clearer message.
- **The third commit touches a workflow file**, so the backport bot cannot carry it and it needs a manual cherry-pick for release branches. Happy to split it out if you'd prefer, the first two commits stand alone.

## Why the test

`link_directives_test.go` checks each `-l<name>` resolves to a file on disk but never parses `//go:build`, so it passed even while this bug was present. Nothing else covers it either, because no job anywhere builds arm64 muslc. Without the new test, reverting this change is a green build.

It asserts each platform selects exactly one link directive, which file it is, and that an aarch64 archive is used exactly when `GOARCH=arm64`. Selection is evaluated with `go/build`'s own `MatchFile`, so the answer is the toolchain's rather than a reimplementation.

Confirmed to fail rather than merely to pass. All seven ways of breaking this change turn it red, including swapping the `-l` names between two files, which is a case the existing tests still pass.

### DIFF
--- a/link_constraints_test.go
+++ b/link_constraints_test.go
@@ -1,0 +1,149 @@
+package sei_test
+
+import (
+	"go/build"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// linkPkgDirs are the vendored libwasmvm api packages whose link_*.go files carry the
+// cgo directive naming a prebuilt archive.
+var linkPkgDirs = []string{
+	"sei-wasmd/x/wasm/artifacts/v152/api",
+	"sei-wasmd/x/wasm/artifacts/v155/api",
+	"sei-wasmvm/internal/api",
+}
+
+// linkPlatform is a build configuration and the link directive it must resolve to.
+type linkPlatform struct {
+	goos, goarch string
+	tags         []string
+	// wantCount is how many link_*.go files the toolchain selects. 1 is the only
+	// healthy value: 0 leaves the linker with no archive and undefined references,
+	// 2 puts two -l directives on one line and the archives collide.
+	wantCount int
+	// wantFile, when set, is the file that must be selected. It catches a swap that
+	// leaves wantCount at 1.
+	wantFile string
+	why      string
+}
+
+// linkPlatforms enumerates the build configurations this repo produces, plus the
+// linux/arm64 muslc cell the static arm64 binary depends on.
+var linkPlatforms = []linkPlatform{
+	{"linux", "amd64", []string{"muslc"}, 1, "link_muslc.go", "static musl release build, amd64"},
+	{"linux", "arm64", []string{"muslc"}, 1, "link_muslc_aarch64.go", "static musl release build, arm64"},
+	{"linux", "amd64", nil, 1, "link_glibclinux_x86_64.go", "ordinary dynamic build, amd64"},
+	{"linux", "arm64", nil, 1, "link_glibclinux_aarch64.go", "ordinary dynamic build, arm64 (the Docker image)"},
+	{"linux", "amd64", []string{"muslc", "sys_wasmvm"}, 1, "link_system.go", "system-libs escape hatch"},
+	{"linux", "arm64", []string{"muslc", "sys_wasmvm"}, 1, "link_system.go", "system-libs escape hatch, arm64"},
+	{"darwin", "arm64", nil, 1, "link_mac.go", "local development on Apple Silicon"},
+	{"darwin", "amd64", nil, 1, "link_mac.go", "local development on Intel Mac"},
+	{"windows", "amd64", nil, 1, "link_windows.go", "windows"},
+
+	// No archive of either flavour is vendored for the remaining linux architectures,
+	// so selecting nothing is correct and matches the glibc path. seid cannot build
+	// for them regardless: giga/executor/lib rejects them at compile time.
+	{"linux", "riscv64", []string{"muslc"}, 0, "", "unsupported arch, must select nothing"},
+}
+
+// TestLinkConstraintsSelectOneFile asserts that every build configuration selects exactly
+// one cgo link directive per api package, and that the archive it names matches the target
+// architecture.
+func TestLinkConstraintsSelectOneFile(t *testing.T) {
+	for _, p := range linkPlatforms {
+		p := p
+		name := p.goos + "_" + p.goarch
+		for _, tag := range p.tags {
+			name += "_" + tag
+		}
+		t.Run(name, func(t *testing.T) {
+			for _, dir := range linkPkgDirs {
+				dir := dir
+				t.Run(dir, func(t *testing.T) {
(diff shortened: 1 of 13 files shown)


alexander-sei (approved):

link_constraints_test.go
```diff
@@ -0,0 +1,149 @@
...
+		require.NoErrorf(t, err, "match %s", filepath.Join(dir, name))
+		if ok {
+			selected = append(selected, name)
+		}
+	}
+	return selected
+}
+
+// hasTag reports whether tag is present in tags.
+func hasTag(tags []string, tag string) bool {
```
masih: Use `slices.Contains`?

link_constraints_test.go
```diff
@@ -0,0 +1,149 @@
...
+	wantCount int
+	// wantFile, when set, is the file that must be selected. It catches a swap that
+	// leaves wantCount at 1.
+	wantFile string
+	why      string
+}
+
+// linkPlatforms enumerates the build configurations this repo produces, plus the
+// linux/arm64 muslc cell the static arm64 binary depends on.
+var linkPlatforms = []linkPlatform{
```
masih: I would inline this inside the test function itself. That's the predominant Go convention and is the only place this slice is used.

link_constraints_test.go
```diff
@@ -0,0 +1,149 @@
...
+// linkPkgDirs are the vendored libwasmvm api packages whose link_*.go files carry the
+// cgo directive naming a prebuilt archive.
+var linkPkgDirs = []string{
+	"sei-wasmd/x/wasm/artifacts/v152/api",
+	"sei-wasmd/x/wasm/artifacts/v155/api",
+	"sei-wasmvm/internal/api",
+}
+
+// linkPlatform is a build configuration and the link directive it must resolve to.
+type linkPlatform struct {
```
masih: You can also inline this inside the test function as an anonymous struct value.

link_constraints_test.go
```diff
@@ -0,0 +1,149 @@
...
+					if p.wantFile != "" {
+						require.Equalf(t, p.wantFile, selected[0],
+							"%s/%s tags=%v (%s): wrong link directive selected",
+							p.goos, p.goarch, p.tags, p.why)
+					}
+
+					// Naming an archive that exists is already covered by
+					// TestLinkDirectivesResolve, and that holds even if two directives are
+					// swapped. Pin the architecture of the archive as well.
+					if p.goos == "linux" && !hasTag(p.tags, "sys_wasmvm") && len(selected) == 1 {
```
masih: You can separate tests the same way we separate build using tags to only run on certain structures. That's the Go native way to do this, where you can write tests that only run on Linux amd64 for example

link_constraints_test.go
```diff
@@ -0,0 +1,149 @@
...
+					}
+				})
+			}
+		})
+	}
+}
+
+// selectedLinkFiles returns the link_*.go files in dir that the Go toolchain compiles for
+// the given platform.
+func selectedLinkFiles(t *testing.T, dir string, p linkPlatform) []string {
```
masih: This can also be made into architecture specific tests with constant values using build tags which is a Go Native way of doing it.

link_constraints_test.go
```diff
@@ -0,0 +1,149 @@
...
+	t.Helper()
+
+	ctx := build.Default
+	ctx.GOOS = p.goos
+	ctx.GOARCH = p.goarch
+	ctx.BuildTags = p.tags
+	// The link files are pure `import "C"`; with cgo off the toolchain excludes them all.
```
masih: Is this true? Change that value to false and see if the tests pass.

masih (commented):
nice, thanks @monty-sei 

Can I ask you to add checksum tests too please? Take a look at existing checksum tests for other architectures I added. ty!

Left some comments otherwise no blockers 🙌

bdchatham (approved):

link_constraints_test.go
monty-sei: Thanks, went with this, it's split per platform now plus a `Linux ARM64` job so the arm64 half actually runs.

link_constraints_test.go
monty-sei: Thanks, same as the other thread, per platform files using the `_GOOS_GOARCH_test.go` filename convention so no explicit tags needed.

PR #3935, opened 2026-08-17
fix(evmrpc): don't charge an innocent client's per-IP bucket on mid-read budget exhaustion

amir-deris:

`rateLimitMiddleware` and the outer `requestSizeLimiter` both watch the same request body, but disagreed about who owns a mid-read failure. When the global byte budget (`max_concurrent_request_bytes`) or the body-read idle timeout was exhausted while `readBoundedBody` was reading, the error surfaced to `rateLimitMiddleware` as a generic read failure. It responded by charging the requesting client's own per-IP token bucket (`chargeAdmissionRejection`) and recording a `read_error` rejection metric — on top of the `budget_midread`/`slow_body` reason and metric the outer `requestSizeLimiter` had already recorded for the same request.

Two problems followed:
1. A well-behaved client could be pushed toward rate-limiting for a purely server-side capacity event it didn't cause. Under load or an actual DoS that exhausts the shared global budget, an innocent client whose request happens to be mid-read at that moment gets its own bucket debited for someone else's traffic.
2. `evmrpc_requests_rejected_total` double-counted the same rejected request under two different reasons.

Flagged in review on #3836: https://github.com/sei-protocol/sei-chain/pull/3836#discussion_r3735763876

### Fix

`rateLimitMiddleware` now recognizes `errBudgetExhausted` / a read-idle-timeout error from `readBoundedBody` and returns without charging admission or recording its own rejection reason, deferring entirely to `requestSizeLimiter`, which already owns the correct reason and the client-visible response.

- Added `TestComposedStack_BudgetMidreadDoesNotChargeInnocentIP`, which holds the global budget open with one client's in-flight request and verifies a second, distinct client's request that fails mid-read due to budget exhaustion does not consume that client's own per-IP token (a follow-up request from the same IP still succeeds under `burst=1`).
- Verified the new test fails without the fix and passes with it.
- `go test ./evmrpc/...` passes.
- `gofmt -s` / `goimports` clean on changed files.


amir-deris:
@seidroid review

amir-deris:
@seidroid review

evmrpc/rate_limit_middleware.go
```diff
@@ -37,6 +37,15 @@ func (m *rateLimitMiddleware) ServeHTTP(w http.ResponseWriter, r *http.Request)
...
+			return
+		}
+		if errors.Is(err, errSlowBody) {
+			// Client-caused stall: still charge the per-IP bucket.
+			_ = m.gate.chargeAdmissionRejection(r.Context(), ip)
+			return
+		}
 		m.rejectAdmission(r.Context(), w, ip, rejectReasonReadError, http.StatusBadRequest, "bad request")
 		return
 	}
```
bdchatham, line 51: Outside the scope of your PR, but a bit strange that we seem to be swallowing errors and not logging them.

evmrpc/rate_limit_middleware.go
```diff
@@ -37,6 +37,15 @@ func (m *rateLimitMiddleware) ServeHTTP(w http.ResponseWriter, r *http.Request)
 			m.rejectAdmission(r.Context(), w, ip, rejectReasonOversize, http.StatusRequestEntityTooLarge, "request body too large")
 			return
 		}
+		if errors.Is(err, errBudgetExhausted) {
+			// Server-side capacity event; outer limiter already owns the response.
+			return
+		}
+		if errors.Is(err, errSlowBody) {
+			// Client-caused stall: still charge the per-IP bucket.
+			_ = m.gate.chargeAdmissionRejection(r.Context(), ip)
```
bdchatham: > m.gate.chargeAdmissionRejection(r.Context(), ip)

Don't need _ if it's anyways not used at all

evmrpc/rate_limit_middleware.go
amir-deris: Thanks for feedback. Regarding logging, I think that opens the gate for more DOS problems due to writing logs to disk. It seems adding metrics around them could be a more suitable approach. Perhaps we can revisit this once it is rolled out and we see how much rate limiting traffic we get

amir-deris:
@seidroid review

bdchatham (approved):

sei-will (approved):

PR #3936, opened 2026-08-17
Refine review config to consider on chain params

masih:
Update revuew context to avoid false negative blocker comments regarding consensus params and config.

See:
* https://github.com/sei-protocol/sei-chain/pull/3934#discussion_r3796299583

### DIFF
--- a/REVIEW.md
+++ b/REVIEW.md
@@ -100,3 +100,38 @@ transaction to land for a sender that is already associated (or already voted)
 at the time of broadcast — then the fee path does apply and the transaction is
 rejected. When in doubt, check `IsTxGasless` for the authoritative set rather
 than reasoning from `minimum-gas-prices` alone.
+
+## 4. Verify live consensus params before flagging timeout-override removal
+
+Legacy `unsafe-*-timeout-override` handling intentionally remained active while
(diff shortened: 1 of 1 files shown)


REVIEW.md
```diff
@@ -100,3 +100,38 @@ transaction to land for a sender that is already associated (or already voted)
...
+`badParams` and unsafe overrides are disabled. It becomes a real finding when an
+affected live network still has default or unset timeout params, or explicitly
+enabled overrides would stop applying without an announced migration.
+
+This does not excuse local harness regressions. Docker, scripts, and test networks
+that relied on overrides must move their intended values into genesis consensus
+params, and tests must inspect the final genesis after gentx collection because
+genesis export helpers may replace `ConsensusParams`.
+
+## 5. Unknown legacy TOML keys do not require placeholder struct fields
```
pompon0: does this mean that we don't need to preserve the deprecated fields? Perhaps we should error on unknown fields instead then?

pompon0 (approved):

REVIEW.md
masih: will add thanks

PR #3941, opened 2026-08-17
[ConfigManager] Remove unnecessary defensive test coverage validation

bdchatham:
Purely subtractive. **26 files, +23 / −1026.**

`wiring_coverage.txt` recorded which checks each package called, and `CheckWiring` re-derived that list by parsing the package's test sources and compared the two. It asserted nothing about configuration — its whole subject was whether a call had been deleted.

That is a test that other tests exist, and it does not terminate. Deleting the `CheckWiring` call was itself invisible to it, so a second check asserted that every package calling any check also called `CheckWiring`. Tests are assumed to be the shape their author gave them; review is what holds that.

## What it cost

- a 288-line source parser using `go/parser` and `ast.Inspect`
- eleven `testdata/wiring_coverage.txt` files
- a requirement that every section argument be a **string literal**, because a loop variable collapses several sections into one row — which is why coverage in this suite could not be driven from a table

## What does not change

The per-check tests were already plain Go tests and are untouched:

```go
func TestReadReceiptConfigAbsentKeysKeepDefaults(t *testing.T) {
	configtest.CheckAbsent(t, "receipt-store", readReceipt, DefaultReceiptStoreConfig())
}
```

Only the layer watching that function is gone. No check loses a call site, no record of configuration behaviour is removed, and nothing needed converting.

## Also in here

- Nine prose passages justified a literal section name by citing the parse. That constraint is gone, so they are too.
- `repoRoot` moves from the deleted contract test to `golden_contract_test.go`, its only remaining user. `TestNoRecordedKeyIsNamedCI` walks from `go.mod` to read every `*.keys.golden` in the tree, so a key named `ci` added in any package is caught without anyone remembering that test exists. Behaviour is unchanged; the function just moves to the file that calls it.
- The primitive table loses a row, so the count of file-backed checks beneath it goes from three to two.

## Verification

`go build ./...` clean. Green: `testutil/configtest`, `cmd/seid/cmd`, `admin`, `evmrpc/config`, `giga/executor/config`, `sei-db/config`, `sei-wasmd/x/wasm`, `x/evm/{blocktest,querier,replay}`, `sei-cosmos/server/config`, `app`. `gofmt -s` and `goimports` clean.


bdchatham:
Superseded by #3942, which is the same change from a branch on this repo.

This PR was opened from a fork, and fork PRs cannot publish org packages, so `Integration Test Check` fails by construction — `integration-test.yml` lines 57-59 say so directly. `codex_review` and `ai-review` fail for the same reason, since a `pull_request` event from a fork gets no secrets. None of the three red checks was caused by the change.

#3942 also carries two fixes found while reviewing this one:

- `TestGuideListsEveryPrimitive` shared the deleted file and belongs to nothing in this change. It holds the guide's primitive table to the package's exported `Check*` surface, and deleting it would have made that table an unchecked claim in the same commit that removed a row from it. It moves to `guide_contract_test.go`.
- Three counts in the guide were stated against the old table and are corrected.

Apologies for the churn to the reviewers already tagged.

PR #3943, opened 2026-08-17
[ConfigManager] Config Registry 

bdchatham:
A configuration key today needs three things written separately, the reader's lookup, a flag binding, an environment spelling, with nothing tying them together, so a rename moves one and leaves the others. Several such mismatches are already pinned in `testutil/configtest` (#3816, #3837, #3851). A key now enters once, as a field on its owning package's section struct, and all three come from that field's `mapstructure` tag.

**Start with `config/registry/doc.go`** — the problem, the declaration contract, what the package deliberately is not, and how to add a section.

### Scope

`Resolve` answers for declared keys only, so it serves a diagnostic or an authoring check, not the boot. A running node reads a source carrying every key whether a section declares it or not. Delivery, the file format, and per-section validation are later slices.

### Verified

`build`, `vet`, `gofmt -s`, `goimports`, `golangci-lint` clean. 34 tests, 104 assertions, 100% of statements with no unreached block, under `-race`.

`Resolve` reads the registry once, so the defaults and the declared set describe one registry. That property has no test: from outside the package, a section arriving during the call cannot be told from one arriving after it.


config/registry/README.md
```diff
@@ -0,0 +1,103 @@
+# Sei Config Registry
+
+The single declaration point for a stable configuration key. A key enters once, as a field on
```
masih: I would move all of these to `doc.go` or if something specific here for agents I would move them to AGENTS.md in this package ( create if there isn't one ).

config/registry/registry.go
```diff
@@ -0,0 +1,281 @@
...
+// feature package can import. TestModesMatchTheNodeModes holds the two sets against each
+// other, so a mode added there fails here rather than silently having no default.
+type Mode string
+
+// The modes a default may vary on.
+const (
+	ModeValidator Mode = "validator"
+	ModeFull      Mode = "full"
+	ModeSeed      Mode = "seed"
+	ModeArchive   Mode = "archive"
```
masih, line 23: Use iota?

config/registry/registry.go
```diff
@@ -0,0 +1,281 @@
...
+
+// RegisterSection records a section's struct and its per-mode default.
+//
+// The dotted key for every field is derived from the section name and the field's mapstructure
+// tag. Nothing here reads a hand-typed key string, which is the property that makes template
+// and reader drift impossible rather than test-guarded.
+//
+// It never panics. A registration this package cannot use is recorded as a Defect and the
+// section is not registered.
+func RegisterSection(name string, proto any, defaults func(Mode) any) {
```
amir-deris: I was wondering what does `proto` name stands for in this context?

amir-deris (approved):

config/registry/registry.go
```diff
@@ -0,0 +1,281 @@
...
+const envPrefix = "SEID"
+
+// EnvName returns the one environment variable that delivers a key.
+//
+// Derived from the key rather than declared, so a section cannot carry a spelling its
+// environment variable does not match. Dots and hyphens both become underscores, which is the
+// replacer the boot installs, so two keys differing only in that punctuation collide; a check
+// over the registry is what catches that rather than a reader discovering it.
+func EnvName(key string) string {
+	return envPrefix + "_" + strings.ToUpper(strings.NewReplacer(".", "_", "-", "_").Replace(key))
```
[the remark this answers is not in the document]
bdchatham: Added the check. `envNamesAreDistinct` runs inside `RegisterSection` under the lock and refuses a registration whose keys share a spelling with each other or with an already-registered section, so the collision cannot enter the declared set.

`TestAKeySharingAnEnvironmentSpellingIsRefused` covers both directions: within one section, and across two whose names differ only in that punctuation (`a-b` and `a`, which both derive `SEID_A_B_A_B`). A third case asserts distinct spellings still register, so the guard cannot pass by refusing everything. `TestEnvSpellingCollisionsAreDetectable` no longer says a check is owed; it points at the one that exists.

Verified by removing the call and watching the test fail.

config/registry/registry.go
```diff
@@ -0,0 +1,281 @@
...
+				return fmt.Errorf("%s.%s is squashed but is a %s, not a struct", prefix, f.Name, ft.Kind())
+			}
+			if err := walk(ft, prefix, keys); err != nil {
+				return err
+			}
+			continue
+		}
+
+		path := prefix + "." + tag
+		if ft.Kind() == reflect.Struct && !isLeaf(ft) {
```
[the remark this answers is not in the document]
bdchatham: Both guards added, at the two places you named.

Duplicate paths: `deriveKeys` sorts and then refuses an adjacent repeat, naming the path. Your exact shape is the probe — a squashed `Shared{Laddr}` alongside an outer `Laddr` with the same tag.

Empty subtrees: `walk` now recurses through `walkSubtree`, which refuses a struct-typed field that contributed zero keys. Three probes cover the class: an empty struct, one whose fields are all unexported, and a defined type over an opaque leaf (`type stamp time.Time`), which is the `sdk.Dec` case.

While writing the duplicate probe I hit a third: `walk` skipped `!f.IsExported()` before reading the tag, so a lowercase embedded type carrying `,squash` lost its whole subtree with no defect and no other sign. An unexported field carrying a tag is now refused on both walks, since nothing can write to it.

Each guard verified by removing it and watching the covering test fail.

config/registry/spec_test.go
```diff
@@ -0,0 +1,1073 @@
...
+	})
+	for _, d := range registry.Defects() {
+		t.Fatalf("registering the probe section produced a defect: %v", d.Err)
+	}
+
+	for _, m := range registry.Modes() {
+		res, err := registry.Resolve(m)
+		if err != nil {
+			// A named refusal is an answer. A resolution missing a declared key is not.
+			continue
```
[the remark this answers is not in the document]
bdchatham: Correct, and it was worse than not asserting resolution: the `continue` meant the test would also have passed on a `Resolve` that refused everything.

Rewritten as the two fixtures you described. The first registers a default that states every declared key and asserts each one resolves, with a guard that the probe declared something so the loop cannot be empty. The second keeps `TLS` nil and asserts `Resolve` returns an error naming `svc.tls.cert`, since a caller cannot fix a default from a message that does not say which key is missing.

config/registry/registry.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+	sort.Slice(out, func(i, j int) bool { return out[i].Name < out[j].Name })
+	return out
+}
+
+// Lookup returns a registered section.
+func Lookup(name string) (Section, bool) {
+	mu.RLock()
+	defer mu.RUnlock()
+	s, ok := sections[name]
```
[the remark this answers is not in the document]
bdchatham: Both copy now, so the read API is by-value throughout.

`TestAReadKeySliceCannotReachTheRegistry` writes into the slice each accessor returns and then reads again, for `Lookup` and `Sections` separately. Verified by removing the copies and watching both subtests fail.

config/registry/registry.go
```diff
@@ -0,0 +1,281 @@
...
+			continue
+		}
+
+		tag, squash, err := tagOf(f, prefix)
+		if err != nil {
+			return err
+		}
+
+		ft := f.Type
+		for ft.Kind() == reflect.Ptr {
```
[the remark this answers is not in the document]
bdchatham: Confirmed and fixed. `walk` now carries the struct types on the current path and refuses one that reappears.

It is the one refusal that has to run before the recursion rather than after it, for the reason you give, and the comment on `walk` says so. I reproduced it first: `type Cyc struct { Child *Cyc }` took 18 seconds and then killed the test binary, which is what a `Defect` cannot be recovered from. Removing the guard reproduces that; with it, the case is an ordinary row in the refusal table.

config/registry/registry.go
bdchatham: I've updated this since`proto` is a bit overloaded and could read as protobuf everywhere else in this tree, and the argument is a prototype struct. Renamed to `prototype`, and `RegisterSection`'s godoc now says what it is for: read for its fields and their tags, never for its values, which is why it is the reader's own struct rather than a copy. A second struct would be a second statement of the same key set, and the two would disagree the first time somebody edited one.

Leaving this thread for you to close, since it was a question rather than a defect.

config/registry/registry.go
```diff
@@ -0,0 +1,379 @@
...
+// feature package can import. TestModesMatchTheNodeModes holds the two sets against each
+// other, so a mode added there fails here rather than silently having no default.
+type Mode string
+
+// The modes a default may vary on.
+const (
+	ModeValidator Mode = "validator"
+	ModeFull      Mode = "full"
+	ModeSeed      Mode = "seed"
+	ModeArchive   Mode = "archive"
```
yzang2019, line 23: This is nice

config/registry/registry.go
```diff
@@ -0,0 +1,379 @@
...
+// asked, which is why two orders are observable across the key set. Stating it as data is what
+// lets a diagnostic tell a node operator that their environment variable beat their file.
+var precedence = []string{"default", "file", "env", "flag"}
+
+// Precedence returns the declared order in which layers win, lowest to highest.
+//
+// A copy, and a function rather than a variable, because an order any importer could assign to is
+// an order no caller can rely on: one package reordering it at init would change every other
+// package's resolved values with nothing to point at.
+func Precedence() []string { return append([]string(nil), precedence...) }
```
masih: Why is this simply not a constant? Better yet, can we just document the order as implicit convention rather than allowing the code to choose it? We never want to change this order, and there are well established standards for it going back to the birth of Unix.

yzang2019 (approved):

config/registry/registry.go
```diff
@@ -0,0 +1,379 @@
...
+	}
+	return keys, nil
+}
+
+// walk appends the dotted keys a struct declares under prefix.
+//
+// open carries the struct types on the current path, so a self-referential one is refused rather than
+// recursed into. A stack overflow cannot be recovered into a Defect, so this is the one refusal that
+// has to happen before the recursion rather than after it.
+func walk(t reflect.Type, prefix string, keys *[]string, open map[reflect.Type]bool) error {
```
masih, line 219: Unfortunate to use reflection; can we design this such that all the config is verified at compile time? 

I am curious to understand why reflection at all.

config/registry/resolve.go
```diff
@@ -0,0 +1,336 @@
...
+	"strings"
+)
+
+// Layer is one configuration source's contribution, before precedence is applied.
+//
+// Source names which layer this is, and must appear in Precedence. Values holds that source's own
+// keys and nothing else: a layer that folded in a second source would make provenance unrecoverable,
+// which is the property that lets a diagnostic tell a node operator their environment variable beat
+// their file.
+type Layer struct {
```
masih: This seems like an implementation detail in config resolution mechanism. I would avoid exporting this. Ditto for all other conceptual things that a user of registry package does not care about. Unless there is a clear reason for this being an exported type?

masih (commented):
Questions please.

config/registry/README.md
bdchatham: I'd put this here simply for the markdown readability for people, but I'll consolidate

config/registry/registry.go
bdchatham: On iota, it forces this to an int which then requires a translation when used as a formal String form so I'd opted for just this. Maybe I'm overcomplicating it and Go iota has some sugar I don't know about.

config/registry/README.md
bdchatham: Consolidated into `doc.go` and deleted the README, so the PR is four Go files.

Go doc comments carry headings, lists and code blocks, so the structure survived the move and `gofmt` keeps it canonical. `go doc ./config/registry` renders the whole thing as one page.

The package is not registered anywhere yet and has no contributor procedure beyond the four steps under **Adding a Section**, so I left it at `doc.go` rather than also adding a package `AGENTS.md`. If a section-authoring workflow grows past those four steps, that is when it earns its own file.

config/registry/registry.go
masih: all good

config/registry/resolve.go
bdchatham: Yeah, this is for tracking what source a config value came from in the `doctor` command. Partially similar for the Precedence relevance but not entirely for that one - just related. I'd originally wanted it for a diagnostic to introspect into outcomes of the config management, but I'm going to cut it from scope for now and add this back later if it's something we find is missing from the UX

config/registry/registry.go
bdchatham: Yeah, not ideal but it's something we can continue to consolidate complexity out of once we harmonize where we are sourcing configuration from. The downstream readers are the challenge with this, since they ask for a specific dotted config string like viper.GetBool("giga_executor.occ_enabled"). My thinking is we take these improvements in bite-sized chunks. This first step is bridging configuration to the registry pattern and essentially wrapping the app.toml and config.toml plumbing in a single sei.toml source
  
After this lands, I see no reason we can't go to the consumer code paths and move them off the dotted lookup onto the struct field it already decodes into, then remove the lookups all together once there are no remaining read edge cases. The key names themselves mostly stay as they are, since renaming one breaks a file an operator already wrote

config/registry/resolve.go
```diff
@@ -0,0 +1,297 @@
...
+		return nil
+	}
+	out := map[string]any{}
+	for _, key := range Keys() {
+		// An empty value is treated as unset. A variable exported empty is far more often a shell
+		// artefact than a deliberate empty string, and the two are indistinguishable here. The cost is
+		// that clearing a key by exporting it empty reads as touching nothing, and Overrides will not
+		// mention it.
+		if v, ok := lookup(EnvName(key)); ok && v != "" {
+			out[key] = v
```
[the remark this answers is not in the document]
bdchatham: Confirmed and fixed. I had closed this window between the defaults and the declared set and then reopened it in `envValues`, which called `Keys()` for the same thing.

`envValues` now takes the declared set rather than reading it, so one `Sections()` call answers for the whole resolution. There is exactly one read left in the file.

Worth recording how it is tested, because my first attempt did not work. Unlike the defaults window, this one has an external symptom: a key reported as one no section declares, that the registry declares. So I wrote a 40-round concurrent test asserting that, and it passed with the defect still in place. The two reads are adjacent lines, and a goroutine does not reliably land between them.

The trigger that does work is deterministic. `Resolve` calls each section's `Defaults` between reading the registry and reading the environment, and `Defaults` is caller-supplied, so a section registered from inside one lands in the window every time. With the fix the test passes; with `envValues` reading `Keys()` again it fails naming `second.a`.

config/registry/registry.go
```diff
@@ -0,0 +1,365 @@
...
+	}
+	if section != strings.ToLower(section) {
+		return nil, fmt.Errorf("section name %q is not lower case; configuration sources "+
+			"enumerate lower-cased, so a key under it would never match a written one", section)
+	}
+	if strings.Contains(section, ".") {
+		return nil, fmt.Errorf("section name %q carries a dot, and a section is one segment. A dotted "+
+			"name declares keys inside another section's subtree, where the two sections' defaults "+
+			"land in one map and whichever renders last silently wins", section)
+	}
```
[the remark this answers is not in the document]
bdchatham: Confirmed and fixed. Reproduced it first, since the consequence is worth seeing:

```
declared keys: [a b.a]
  "a b.a" -> env "SEID_A B_A"
defects: []
```

That variable cannot be exported, and the key cannot be written in a file unquoted, yet the section registered with no defect.

Fixed at the cause rather than by adding the missing character. The section check and the tag check were the same rule stated twice, which is how they drifted, so both now go through one function:

```go
// unaddressableChar returns the first character in a key segment that no configuration source can
// carry, and whether there was one.
//
// One function for both places a segment enters, a section name and a field tag, because the two
// answer to the same sources. Stated separately they drifted, and a section name accepted a space the
// field rule refused.
func unaddressableChar(segment string) (string, bool)
```

Each caller keeps its own consequence, because they differ: a dot in a section name declares keys inside another section's subtree, where whichever renders last silently wins, while a dot in a tag claims a subtree the struct does not have. The messages now name the offending character rather than the class, so `a b` and `a.b` read differently.

Removing the space from the shared rule fails both the section case and the tag case, where before it could only fail one. That is the property worth having.

`doc.go` had the same gap and described the rule as applying to tags only. It now says a segment is refused if it is upper-case or carries a dot or a space, and that this holds for the section name and a field's tag alike.

PR #3944, opened 2026-08-18
Deprecate oracle message and query handlers

codchen:
## Summary

- return a registered `ErrOracleDeprecated` error from both oracle message handlers
- return the same explicit error from every oracle gRPC query handler
- return a registered `ErrAssociateDeprecated` error from the EVM `MsgAssociate` handler
- remove the oracle module's legacy SDK message/query routes and orphaned legacy handler
- remove the application-level gasless transaction classifier and its fee, gas-meter, metrics, and block-gas exceptions
- make deprecated oracle and `MsgAssociate` transactions use normal fee and gas accounting
- remove oracle-specific spam-prevention, vote-alone, and priority ante handling
- disable oracle transaction simulation operations
- update handler, state-machine, fee, block-gas, priority, and integration coverage for the deprecated behavior

## Why

The oracle module is deprecated, but its Cosmos message and query services still accepted incoming requests. Keeping the gRPC services registered while returning a stable module error gives callers an explicit deprecation result and prevents further oracle message state mutations. The legacy SDK routes are no longer registered.

`MsgAssociate` is also deprecated. It was the final message eligible for the application's gasless transaction path, so that classifier and its special cases can be removed entirely.

## Impact

Incoming aggregate exchange-rate votes, feeder delegation messages, and oracle queries now fail with oracle error code 25 (`oracle module is deprecated`). `MsgAssociate` now fails with EVM error code 2 (`MsgAssociate is deprecated`).

All Cosmos transactions now use ordinary fee deduction, transaction gas meters, gas metrics, and block-gas accounting. Deprecated oracle transactions no longer receive oracle-specific mempool priority; they use ordinary fee-derived priority before dispatch.

The Ethereum-native `AssociateTx` path is unchanged.

Oracle mid-block/end-block processing remains registered for state compatibility. Under the default oracle parameters, both `DefaultSlashFraction` and `DefaultMinValidPerWindow` are zero, so missed oracle votes do not produce slashing. Networks that override those parameters must set them to zero before or with this deployment.

## Validation

- `go test ./app/... ./x/evm/...`
- `go test -race ./app ./app/ante ./app/antedecorators ./x/evm/ante ./x/evm/keeper ./x/evm/types`
- `go vet ./app/... ./x/evm/...`
- parsed the renamed oracle integration YAML scenarios
- `gofmt`, `goimports`, and `git diff --check` on all touched files


codchen:
@seidroid review

codchen:
@seidroid review

masih:
@seidroid review

sei-will (approved):

app/prioritizer_test.go
```diff
@@ -94,9 +142,19 @@ func (s *PrioritizerTestSuite) TestGetTxPriority() {
 			wantPriority: 123,
 		},
 		{
-			name:         "oracle Tx type is oracle priority",
+			name:         "oracle Tx type uses fee priority",
```
philipsu522: should we just remove this test completely?

contracts/test/EVMPrecompileTest.js
```diff
@@ -308,15 +308,15 @@ describe("EVM Precompile Tester", function () {
             oracle = new ethers.Contract(OraclePrecompileContract, contractABI, accounts[0].signer);
         });
 
-        it("Oracle CLI TWAP query should fail without feeder data", async function () {
+        it("Oracle CLI TWAP query should fail when the oracle module is deprecated", async function () {
```
philipsu522: nit - replace "when" with "because"

x/oracle/types/msgs.go
```diff
@@ -30,7 +30,7 @@ func NewMsgAggregateExchangeRateVote(exchangeRates string, feeder sdk.AccAddress
 }
 
 // Route implements sdk.Msg
-func (msg MsgAggregateExchangeRateVote) Route() string { return RouterKey }
+func (msg MsgAggregateExchangeRateVote) Route() string { return ModuleName }
```
philipsu522, line 33: are the changes in this file intentional?

codchen: yeah since we are removing router logics. These two vars are both `oracle`

PR #3946, opened 2026-08-18
Deprecate IBC and transfer query handlers

masih:
Now that IBC in/outbound is disabled as part of SIP-3:
- return explicit deprecation errors from every IBC core and transfer gRPC query handler
- add coverage ensuring all registered query methods reject requests
- preserve capability unchanged because it exposes no query service


masih:
Replaced by #4002

PR #3956, opened 2026-08-19
fix(gov): harden deposit refund handling

codchen:
## Summary

- align governance refund handling with bank-layer recipient eligibility rules
- preserve deposit records and their backing module balance when a payout cannot be completed
- continue processing eligible refunds independently
- expand regression coverage for recipient-policy edge cases

## Behavior

Refund processing now validates recipient eligibility before initiating each account transfer. If a recipient is unavailable, its deposit remains recorded and backed by the governance module balance for migration-based recovery. Other eligible deposits are refunded and removed normally.

## Validation

- `go test ./sei-cosmos/x/gov/...`
- `go vet ./sei-cosmos/x/gov/...`
- `gofmt` and `goimports` on all changed Go files
- `git diff --check`


sei-will (approved):

codchen:
@seidroid review

PR #3960, opened 2026-08-19
Use iterative memiavl snapshot traversal

alexander-sei:
## Summary
- replace recursive snapshot traversal with an explicit, height-sized frame stack while preserving post-order encoding
- keep per-node allocations under stable Go call stacks so tree paths do not create permanent memory-profiler buckets

Linear: [PLT-1013](https://linear.app/seilabs/issue/PLT-1013/go-memory-profiler-bucket-growth-degrades-rpc-latency-on-long-running)

## Test plan
- [x] `go test ./sei-db/state_db/sc/memiavl -count=1`
- [x] `go test -race ./sei-db/state_db/sc/memiavl -run 'TestSnapshotWriterPipeline|TestSnapshotWriterCancellation|TestWriteSnapshotWithBuffer' -count=1`
- [x] `make dblint`


alexander-sei:
Superseeded by PR #4003

PR #3966, opened 2026-08-19
Bump version to v6.6.2 in prep for patch release

masih:


masih:
escalating to fast merge

PR #3967, opened 2026-08-19
feat(sei-db): roll back state store from snapshot and WAL

blindchaser:
## Summary

Make `seid rollback` rewind the state store with Tendermint and state-commit. The state store restores the newest retained snapshot at or below the target and replays its changelog to the exact target height. If the configured state store cannot follow, the command logs a warning and keeps the previous behavior of rolling back Tendermint and state-commit.

- `sei-db/state_db/ss/composite`: adds validation, snapshot selection, WAL replay, and a crash-resumable rollback sequence. Durable markers, directory synchronization, and restart recovery keep the snapshot, changelog, and version watermark consistent after a process crash or power loss. Rollback rejects unsupported EVM-split and non-Pebble state stores.
- `sei-cosmos/storev2/rootmulti`: validates state-store rollback before moving state-commit, then coordinates both stores through the existing rollback command.
- `sei-db/db_engine/pebbledb/mvcc` and `sei-db/state_db/ss/snapshot`: anchor changelog retention to the oldest retained snapshot while preserving the count-based recovery floor. Empty-block version gaps remain replayable.
- `sei-db/wal` and `sei-db/common/utils`: add version-aware WAL searches and durable Pebble directory cloning.
- `sei-db/db_engine/types`: defines optional rollback, rollback-validation, WAL-read, and WAL-pruning capabilities.

## Test plan

- `go build ./...`
- `go vet ./sei-db/state_db/ss/composite/... ./sei-db/db_engine/pebbledb/mvcc/... ./sei-cosmos/storev2/rootmulti/...`
- `golangci-lint run ./sei-db/state_db/ss/composite/... ./sei-db/db_engine/pebbledb/mvcc/... ./sei-cosmos/storev2/rootmulti/...`
- `go test -race -count=1 ./sei-db/state_db/ss/composite/... ./sei-db/db_engine/pebbledb/mvcc/... ./sei-cosmos/storev2/rootmulti/...`
- Covers exact-snapshot rollback, WAL replay, empty blocks, unavailable rollback support, legacy layouts, every crash-recovery step, durable marker and changelog handling, discarded-fork snapshot cleanup, and WAL retention bounds.

### DIFF
--- a/sei-db/state_db/ss/composite/rollback.go
+++ b/sei-db/state_db/ss/composite/rollback.go
@@ -132,6 +132,13 @@ func rollbackBaseVersion(root string, changelog types.SnapshotWALReader, target 
 		return 0, fmt.Errorf("cannot roll back state store to version %d: nearest snapshot is %d and the changelog is empty", target, base)
 	}
 	if next == 0 {
+		// A retained prefix proves that a gap between two retained entries came
+		// from empty blocks, because retention only removes a prefix. It does
+		// not prove that a missing suffix was empty: rollback truncates the WAL
+		// tail, and opening the WAL can repair a corrupt tail. If the live
+		// database is still above that cut, restoring base and advancing only
+		// its watermark to target would silently lose the missing writes.
+		// base == target is different and returns above: no suffix is needed.
 		return 0, fmt.Errorf("cannot roll back state store to version %d: nearest snapshot is %d and the changelog has no newer entries", target, base)
 	}
 	// A first entry above base+1 means the versions in between either wrote
--- a/sei-db/state_db/ss/composite/rollback.go
+++ b/sei-db/state_db/ss/composite/rollback.go
@@ -1,0 +1,489 @@
+package composite
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/sei-protocol/sei-chain/sei-db/common/utils"
+	"github.com/sei-protocol/sei-chain/sei-db/config"
+	"github.com/sei-protocol/sei-chain/sei-db/db_engine/types"
+	sssnapshot "github.com/sei-protocol/sei-chain/sei-db/state_db/ss/snapshot"
+	"github.com/sei-protocol/sei-chain/sei-db/wal"
+)
+
+const rollbackTargetFile = ".ss-rollback-target"
+
+type ssRollbackPlan struct {
+	root        string
+	changelog   string
+	baseVersion int64
+	snapshotDir string
+}
+
+func (s *CompositeStateStore) ValidateRollback(target int64) error {
+	_, err := s.planRollback(target)
+	return err
+}
+
+func (s *CompositeStateStore) Rollback(target int64) error {
+	plan, err := s.planRollback(target)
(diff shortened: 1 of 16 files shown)


sei-db/state_db/ss/composite/rollback.go
```diff
@@ -0,0 +1,489 @@
...
+	}
+
+	oldest, next, err := changelog.WALVersionsAfter(base)
+	if err != nil {
+		return 0, fmt.Errorf("read changelog coverage above snapshot %d: %w", base, err)
+	}
+	if oldest == 0 {
+		return 0, fmt.Errorf("cannot roll back state store to version %d: nearest snapshot is %d and the changelog is empty", target, base)
+	}
+	if next == 0 {
```
yzang2019, line 134: rollbackBaseVersion allows a middle empty block, WALVersionsAfter returns next == 0 when nothing in the log is above base. Retention only truncates a prefix, so a retained entry at or below base (oldest != 0 && oldest <= base) already proves base+1..target were never written. The reopen path in NewCompositeStateStore is built for exactly that: restore the snapshot, replay nothing, then SetLatestVersion(pendingRollbackTarget) from the marker.

Rolling back to the snapshot height is allowed even with an empty changelog (base == target returns early). Rolling back above it by empty heights is not. That is an inconsistency in the same function.

yzang2019 (approved):

sei-db/state_db/ss/composite/rollback.go
blindchaser: I do not think `oldest <= base` is enough evidence for a missing suffix.

It proves a gap **between retained entries** was not removed by prefix retention. It does not prove the WAL tail was never removed. This codebase has two suffix-removal paths: `truncateChangelogAfterVersion` deliberately truncates the tail during rollback, and opening the underlying WAL can repair a corrupt tail.

The existing test now named `changelog tail cut below target` demonstrates the ambiguity. The live store is at version 8 and versions 6–8 wrote data. Its WAL is cut at snapshot 5, which gives exactly `oldest != 0`, `oldest <= base`, and `next == 0`. If target 7 were accepted, rollback would restore snapshot 5, replay nothing, set only the watermark to 7, and silently lose the writes from 6 and 7.

`base == target` is not the same case: snapshot `base` is already the exact requested state, so no suffix evidence is required.

Supporting an all-empty suffix safely needs new durable evidence, such as WAL records for empty blocks or a separate verified coverage marker. The current WAL cannot distinguish it from a missing tail, so the conservative refusal remains. Commit `8b6e603e4` adds this rationale beside the check and makes the corruption case explicit in the test. The composite race suite and lint are clean.

blindchaser: next == 0 also could be the WAL is truncated or corrupted tail. 

we have a test where the store is at version 8, the snapshot is at 5, versions 6–8 wrote data, and the WAL tail is cut at 5. This produces oldest <= base and next == 0. accepting rollback to 7 would restore snapshot 5, replay nothing, and silently lose writes from versions 6 and 7.

base == target is safe because the snapshot is already the exact requested state. For target > base, we need durable evidence for the suffix. The current WAL does not provide it, so we keep the conservative refuse.

PR #3969, opened 2026-08-20
Separate BlockStore and BlockDB interface

yzang2019:
Main goal of this PR is to separate the Block interface to:
- BlockStore (The interface at consensus layer)
- BlockDB (The DB layer interface)
The idea is that DB layer should not be aware about any specific types or concept in the consensus, we should provide a more generic BlockDB interface 

Consensus related concepts and interfaces should all only stays in the consensus package.

Two extra changes included in this PR:
- Rename the package name from management -> controller
- Move the GarbageCollector config to config package to avoid import cycle

Note that there's no migration needed for any table name change because there's no data and this code is not deployed for production env yet


sei-tendermint/autobahn/types/block_store.go
```diff
@@ -0,0 +1,326 @@
...
+	// undefined.
+	Close() error
+}
+
+// SuffixRange represents the suffix of BlockStore data that data.State can append to/would load on recovery.
+// Elements since the last anchor (last full row which contains AppQC,AppProposal,Block,QC) to
+// the tips persisted in the DB. These are the elements that would be loaded by data.State on restart
+// via BlockStore.ReadSuffix.
+// First <= NextAppQC <= NextAppProposal <= NextBlock <= NextQC
+type SuffixRange struct {
```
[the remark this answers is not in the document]
yzang2019: Already fixed

sei-tendermint/internal/autobahn/blockstore/codec.go
```diff
@@ -1,4 +1,4 @@
-package littblock
```
pompon0, line 1: +1 for moving this type-safe logic to sei-tendermint. Do we need the blockstore impl to be public, or can we move it under internal/autobahn ?

yzang2019: We can move to internal actually, that's a good call

yzang2019:
@seidroid review

sei-db/ledger_db/block/littblock/keys.go
cody-littley, line 1: Should this file live in this package if the storage team impl is agnostic to consensus details?

sei-db/ledger_db/block/littblock/litt_block_db.go
```diff
@@ -97,282 +79,151 @@ func NewBlockDB(config *BlockDBConfig) (types.BlockDB, error) {
...
-		Length: uint32(len(value)), //nolint:gosec // value length fits u32 (litt value cap is 2^32)
-	}
-	if err := s.table.Put(blockKey(n), value, hashAlias); err != nil {
-		return fmt.Errorf("failed to put block %d: %w", n, err)
-	}
-	status.NextBlock = n + 1
-	s.status = utils.Some(status)
-	return nil
+// PruneWatermark returns the reclamation floor.
+func (s *blockDB) PruneWatermark() uint64 {
```
cody-littley: Nit, perhaps we can call this `GetPruneWatermark()`? When I first read this method, I thought it was a function that did pruning internally.

sei-db/ledger_db/block/littblock/litt_block_db.go
```diff
@@ -97,282 +79,151 @@ func NewBlockDB(config *BlockDBConfig) (types.BlockDB, error) {
...
-			Offset: 0,
-			Length: uint32(len(value)), //nolint:gosec // value length fits u32 (litt value cap is 2^32)
-		})
-	}
-	if err := s.table.Put(qcKey(gr.First), value, aliases...); err != nil {
-		return fmt.Errorf("failed to put QC [%d,%d): %w", gr.First, gr.Next, err)
+// SetPruneWatermark raises the reclamation floor to n. Reclaiming the data
+// happens later, on LittDB's GC schedule and no earlier than
+// BlockDBConfig.RetentionTime.
+func (s *blockDB) SetPruneWatermark(n uint64) {
```
cody-littley, line 118: Any particular reason why we are changing the name away from `PruneBefore(n)`? Not strictly opposed to the change, just curious.

sei-db/ledger_db/block/littblock/litt_block_db.go
```diff
@@ -97,282 +79,151 @@ func NewBlockDB(config *BlockDBConfig) (types.BlockDB, error) {
...
+		switch _, err := os.Stat(legacy); {
+		case err == nil:
+			return fmt.Errorf(
+				"block db: found a pre-rename %q table at %s; the table is now named %q, so those "+
+					"blocks and QCs would be neither served nor reclaimed. Delete or move the "+
+					"directory aside to start from an empty store",
+				legacyTableName, legacy, tableName)
+		case !os.IsNotExist(err):
+			return fmt.Errorf("block db: check for a pre-rename %q table at %s: %w",
+				legacyTableName, legacy, err)
```
cody-littley: This this migration code for the version of BlockDB currently in the main branch? If so, not sure it's necessary... we've not deployed the new BlockDB to any environments yet.

cody-littley (approved):

sei-db/ledger_db/block/littblock/litt_block_db.go
yzang2019: I think that's proposed by AI's, I guess SetPruneWatermark to me sounds more obvious that we are set setting the pruning watermark, but not actually pruning here. The pruning would happen sometime later in the background.

One thing to notice here, because the pruning logic has dependency on consensus restriction for QC, the GC implementation will be wired up with the BlockStore, not BlockDB

sei-db/ledger_db/block/littblock/litt_block_db.go
yzang2019: Agreed, will fix

sei-db/ledger_db/block/littblock/keys.go
yzang2019: This has to stay in storage layer for now just because we want to avoid the db to import any dependency from consensus which is a layer above. Consensus can have dependency on DB but not the other way. I think in the future if we can remove the need to store different types of data in the same table by folding the qc into blocks, we can probably simplify and getting rid of this as well.

sei-db/ledger_db/block/littblock/litt_block_db.go
yzang2019: Agree that we can remove it. I think it's AI generated

sei-db/ledger_db/block/littblock/keys.go
cody-littley: Does the type need to be an actual type? Or could we have the type be a char/string type?

This is not a blocking issue.

yzang2019: Will defer that change, if we can remove the type in the future, we don't need to change that any more

yzang2019:
@seidroid review

sei-db/ledger_db/block/littblock/litt_block_db.go
```diff
@@ -2,87 +2,58 @@ package littblock
...
-// NewBlockDB opens (or creates) a LittDB-backed types.BlockDB from config. The
+// NewBlockDB opens (or creates) a LittDB-backed blocktypes.BlockDB from config. The
 // underlying LittDB is built from config.Litt, and the table applies
 // config.RetentionTime as a TTL failsafe (pruning never reclaims data younger
 // than that even once the watermark has advanced past it).
-func NewBlockDB(config *BlockDBConfig) (types.BlockDB, error) {
+//
+// Opening a store that holds records but no QC fails: the store lost data out of
+// band and must not be served.
+func NewBlockDB(config *BlockDBConfig) (blocktypes.BlockDB, error) {
```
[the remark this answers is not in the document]
yzang2019: This is not deployed on production yet, so it's not a blocker

sei-tendermint/autobahn/types/block_store.go
wen-coding, line 75: nit: I don't think we allow starting first block in the middle

sei-tendermint/internal/autobahn/blockstore/codec.go
```diff
@@ -138,14 +67,14 @@ func encodeQC(qc *types.FullCommitQC) []byte {
...
+// as specified by types.BlockStore.WriteQC: [First, First+len(Headers())).
 //
 // The upper bound is derived from the header count rather than read from
 // GlobalRange().Next — the two are equal by FullCommitQC's own invariant, but
 // First and Headers are the only two fields the wire format carries explicitly.
 // GlobalRange().Next is recomputed from the proposal's lane ranges on decode,
 // so deriving from the encoded fields is what makes a QC's range identical
-// before and after a round trip through the table.
+// before and after a round trip through the store.
 func coveredRange(qc *types.FullCommitQC) (types.GlobalBlockNumber, types.GlobalBlockNumber) {
```
wen-coding, line 78: We don't allow starting in the middle of QC right now, I think this is unused.

wen-coding (approved):

sei-tendermint/internal/autobahn/blockstore/codec.go
yzang2019: Just saw the comments, that's good to know, will clean it up in following PRs.

PR #3979, opened 2026-08-21
PLT-980: Populate validator-indexed delegation store at the v6.7 upgrade

amir-deris:
## Problem

`MaxScanLimit` (10,000) is enforced by `query.FilteredPaginateV66`, which
`Querier.ValidatorDelegations` selects whenever `ctx.IsABCIQuery()` is false — that is,
on consensus execution. The staking precompile's `validatorDelegations` filters over the
whole `0x31` delegation prefix to find one validator's rows, so at pacific-1's ~82k
delegations it scans past the limit and reverts.

Plain gRPC/LCD queries take the uncapped `FilteredPaginate` path and were never affected.

## Approach

Populate a validator→delegator secondary index (`0x37`) in one pass inside the **v6.7
upgrade handler**, and switch the current staking precompile to read it. The indexed
prefix holds only one validator's delegations, so iteration tracks page size instead of
the total delegation count and the cap is never approached.

Gating is a **readiness marker key** (`0x38`) written by the migration, not a version
comparison. Because the marker is versioned state, `DelegationByValIndexReady(ctx)` is
correct for historical queries, re-traced blocks, and state-synced nodes without the
caller supplying an upgrade name or height. This matters because the read switch changes
transaction results: if nodes flipped it at different heights they would diverge. The
marker makes the flip atomic with the upgrade block by construction.

Below the upgrade height the marker is absent, the read falls through to the existing
scan, and the query reverts exactly as it does at those heights today — which is what
historical replay requires.

## What changed

**Migration** — `Keeper.MigrateDelegationByValIndex` writes an index key per delegation
then sets the marker. Returns an error rather than panicking on a malformed proto or
address, and is a no-op once the marker is set. Called from the `v6.7` handler on an
infinite gas meter.

**Gating** — `DelegationByValIndexReadyKey` (`0x38`) plus `DelegationByValIndexReady`.
The dual-write in `SetDelegation` / `RemoveDelegation` gates on it. The previous
`delegationByValIndexActive` semver/`IsTracing` gate is removed.

**Read path** — new `Querier.ValidatorDelegationsIndexed`. With the marker set it
paginates `GetDelegationsByValIndexKey(valAddr)` and resolves entries through
`GetDelegationKeyFromValIndexKey`, matching the shape `ValidatorUnbondingDelegations`
already uses; without it, it delegates to `ValidatorDelegations`. That method itself is
unchanged.

**Precompile wiring** — only `precompiles/staking/staking.go` calls the indexed query.

**Removed** — the `seid tools staking backfill-delegation-index` benchmark CLI. Its
sizing job is answered (below), and its audit job is now a keeper invariant test.

## Compatibility notes

- **Cursor semantics change.** For `validatorDelegations`, `next_key` becomes an index
  key rather than a delegation key, so a cursor issued before the upgrade will not
  resolve after it.
- **Legacy precompiles are untouched.** `legacy/v65`, `legacy/v630` and the rest keep
  calling the unindexed querier and keep their current behavior at their eras — zero
  diff in those files.
- **`legacy/v67` is changed deliberately.** It is a committed generated snapshot of
  `staking.go` that `setup.go` does not yet reference (`latestUpgrade` still routes to
  the live file). The next `scripts/bump_version` run will wire it in as the v6.7-era
  precompile, and the pre-change snapshot called the unindexed querier — so the fix
  would have regressed at that bump, and v6.7-era heights would behave differently
  across binaries. v67 is this era's snapshot, not a prior era. Regenerating it instead
  of hand-editing would be equivalent; flagging since the file is marked `DO NOT EDIT`.

## Sizing

`max(heatseeker_client_state_analysis_prefix_num_keys{module="staking", prefix="31"})`
on pacific-1 gives ~82k delegations:

image

One linear scan with no nested lookups over 82k rows, writing ~43-byte keys, is
comfortably within the upgrade block's budget — negligible next to the pacific-1 `evm`
module's ~710M keys at ~2M keys/min. No chunked/EndBlocker migration is needed.

## Test plan

```bash
go test ./sei-cosmos/x/staking/... ./precompiles/staking/... ./app/ -count=1
make fmtcheck && go vet ./...
```

- `TestGRPCQueryValidatorDelegationsIndexedBeyondScanLimit` — seeds `MaxScanLimit + 1`
  delegations plus one to the target validator under a `0xff…ff` delegator that sorts
  last, so the scan must traverse the whole keyspace. Pins that the scan reverts both
  before and after the migration, and that the indexed query answers it afterwards.
- `TestGRPCQueryValidatorDelegationsIndexedMatchesScan` — both paths agree on a
  validator small enough for either to answer.
- `TestMigrateDelegationByValIndex` / `…NoOrphans` — migration counts, marker,
  idempotency, and the both-directions invariant: every delegation is indexed and every
  index entry resolves to a delegation.
- `TestDelegationByValIndexNotReadyNoDualWrite` / `…DualWriteAfterMigration` — the
  write gate follows the marker.
- `TestV67PopulatesDelegationByValIndex` — seeds a delegation through the raw store,
  applies the v6.7 plan, asserts the marker and index key exist.

Linear: https://linear.app/seilabs/issue/PLT-980/


amir-deris:
@seidroid review

amir-deris:
@seidroid review

PR #3980, opened 2026-08-21
Remove capability module

masih:
Remove x/capability and capability-based wiring from IBC, transfer, CosmWasm, and application initialization.

Retain the capability KV store for freeze-mode historical state access and remove its module version during the v6.7 upgrade. Restrict transfer sends to the configured port after removing scoped capabilities.


alexander-sei (approved):

sei-will (approved):

yzang2019 (approved):

app/upgrades.go
```diff
@@ -97,6 +97,7 @@ func (app *App) RegisterUpgradeHandlers() {
 				if err != nil {
 					return nil, err
 				}
+				app.UpgradeKeeper.DeleteModuleVersion(ctx, capabilityModuleName)
```
philipsu522, line 100: curious why we want to delete this (and feegrant) from ModuleVersion vs letting it stay as the last value it was?

sei-ibc-go/modules/core/04-channel/keeper/handshake.go
philipsu522, line 27: do we need to continue supporting this function? (I'm not sure either, was just wondering if you knew)

philipsu522: actually do we need to support any of the public functions here going fwd?

codchen (approved):

app/upgrades.go
masih: RunMigrations skips removed modules anyway but the old entries here need an explicit removal. So that the active module set is represented correctly.

Leaving those entries wouldn't affect execution but seems untidy since the module itself is removed since it won't reflect the actual active set.

Happy to leave it be if there is a good reason to keep it.

sei-ibc-go/modules/core/04-channel/keeper/handshake.go
masih: Yup we can remove those, but would like to remove them in a separate PR along with related interfaces. Some channel APIs (e.g. `SendPacket`) are still used by cosmwasm

PR #3981, opened 2026-08-21
Remove live IBC transfer module

masih:
Remove ICS-20 module wiring, APIs, CLI commands, and CosmWasm integration.

Retire the live IBC precompile while preserving legacy implementations and codecs for historical tracing and transaction decoding. Keep the transfer store and module account mounted for historical state safety, and remove its module version during the v6.7 upgrade.


sei-will (approved):

sei-ibc-go/modules/apps/transfer/keeper/keeper.go
```diff
@@ -46,63 +39,16 @@ func NewKeeper(
...
-
-// SetPort sets the portID for the transfer module. Used in InitGenesis
-func (k Keeper) SetPort(ctx sdk.Context, portID string) {
-	store := ctx.KVStore(k.storeKey)
-	store.Set(types.PortKey, []byte(portID))
-}
-
 // GetDenomTrace retreives the full identifiers trace and base denomination from the store.
 func (k Keeper) GetDenomTrace(ctx sdk.Context, denomTraceHash tmbytes.HexBytes) (types.DenomTrace, bool) {
 	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.DenomTraceKey)
```
philipsu522: should we remove this too?

sei-ibc-go/modules/apps/transfer/keeper/metrics.go
```diff
@@ -9,31 +9,19 @@ var (
 	meter = otel.Meter("ibc_transfer_keeper")
 
 	ibcTransferMetrics = struct {
-		txMsgIbcTransfer         metric.Int64Gauge
-		ibcTransferPacketReceive metric.Int64Gauge
-		ibcTransferSend          metric.Int64Counter
-		ibcTransferReceive       metric.Int64Counter
+		txMsgIbcTransfer metric.Int64Gauge
+		ibcTransferSend  metric.Int64Counter
```
philipsu522: curious why we want to keep these?

sei-ibc-go/modules/apps/transfer/keeper/params.go
philipsu522: can probably remove this too if we're removing the other getters for params

sei-wasmd/x/wasm/keeper/handler_plugin_encoders.go
```diff
@@ -263,33 +262,18 @@ func EncodeWasmMsg(sender sdk.AccAddress, msg *wasmvmtypes.WasmMsg) ([]sdk.Msg,
...
-				Sender:           sender.String(),
-				Receiver:         msg.Transfer.ToAddress,
-				TimeoutHeight:    ConvertWasmIBCTimeoutHeightToCosmosHeight(msg.Transfer.Timeout.Block),
-				TimeoutTimestamp: msg.Transfer.Timeout.Timestamp,
-			}
-			return []sdk.Msg{msg}, nil
-		default:
-			return nil, sdkerrors.Wrap(types.ErrUnknownMsg, "Unknown variant of IBC")
-		}
+func EncodeIBCMsg(_ sdk.Context, sender sdk.AccAddress, _ string, msg *wasmvmtypes.IBCMsg) ([]sdk.Msg, error) {
```
philipsu522: woulnd't we want to remove all ibc messages?

masih: I was thinking freeze-mode compatibility. But then we are going to run a node in v6.6. version for that frozen at height "v6.7 upgrade"; in which case we can remove a whole lot more. hmm 🤔 

@codchen does that track?

sei-ibc-go/modules/apps/transfer/keeper/keeper.go
masih: Agreed. We initially retained this for replaying pre-v6.7 IBC precompile calls on v6.7 nodes. Since pre-v6.7 history is served by v6.6 freeze nodes, that compatibility layer is unnecessary.

sei-wasmd/x/wasm/keeper/handler_plugin_encoders.go
masih: proceeded with offloading query functionality to freezemode which allows us to remove more code. state remains. Marking as resolved and happy to revisit if i have missed anything. ty

codchen (approved):

PR #3984, opened 2026-08-23
feat(query): origin-aware pagination limits for ABCI queries

amir-deris:
## Summary

- Reintroduce pagination DoS/OOM protection on the ABCI/gRPC query path without breaking consensus or EVM precompile semantics
- Untrusted callers get configurable `max-limit`, `max-offset`, and flat `max-iterations` (defaults: 1000 / 10000 / 11000 — default max-iterations covers offset + limit so a request at the offset ceiling still can fill a page in one call); out-of-bound params are rejected upfront
- When the iteration budget is exhausted, return a partial page with `next_key` so clients can resume via key-based pagination
- Operators can bypass limits via `[query] disable-limits`, or per-origin via `trusted-cidrs` on supported query planes; v6.6 frozen paginators remain on all non-ABCI paths

## Behavior

**Untrusted ABCI/gRPC queries** (any origin not in `query.trusted-cidrs`):

- `limit` or `offset` above configured caps → `InvalidArgument` before store iteration
- Flat iteration counter per request; on exhaustion → success with partial results + `next_key` (omit `total` when truncated)

**Trusted origins** (CIDR match) and **kill switch** (`disable-limits = true`): no pagination limits

**`trusted-cidrs` supported planes** — matching uses the direct TCP peer (Tendermint RPC `RemoteAddr` or Cosmos gRPC peer address), not forwarded headers:

- **Tendermint RPC** (`abci_query` on the RPC listener): matched
- **Cosmos gRPC** (`grpc.address`, typically `:9090`): matched
- **REST/LCD** (`api.address`, typically `:1317`, including grpc-gateway routes): **not matched** — queries relay in-process without caller origin; indexers needing higher limits over REST should use gRPC or Tendermint RPC, or set `disable-limits` on nodes not exposed to untrusted callers

**Consensus / precompile path**: unchanged — `FilteredPaginateV66` / `GenericFilteredPaginateV66` retain frozen v6.6 scan semantics

New `[query]` section in `app.toml`:

```toml
[query]
disable-limits = false
trusted-cidrs = []
max-limit = 1000
max-offset = 10000
max-iterations = 11000
```

`0` on any `max-*` field falls back to its package default rather than being taken literally.


amir-deris:
@seidroid review

amir-deris:
@sedroid review

amir-deris:
@seidroid review

sei-will (approved):

PR #3958, opened 2026-08-19
Remove feegrant module

masih:
Remove the feegrant module and its keeper, store, CLI, APIs, protobuf definitions, simulation support, and application wiring.

* Delete the feegrant store during the v6.7 upgrade.
* Reject transactions specifying a distinct fee granter.
* Retain the transaction granter field for wire compatibility.
* Retire precompile 0x1010 with a deterministic revert.
* Remove feegrant from SeiDB, wasmd, tooling, and Swagger.
* Preserve historical genesis files unchanged.
* Regenerate embedded API documentation and update tests.

PR #3893, opened 2026-08-11
feat(precompiles): add scoped module authorizations

codchen:
## Summary

- add scoped authorization flows for staking delegate/redelegate/undelegate, slashing unjail, and distribution reward/commission withdrawals
- preserve the existing vote-only governance authorization and add a separate proposal-submission authorization
- centralize native Cosmos authz grant, execution, and grouped revocation behavior
- update Solidity interfaces/ABIs and add end-to-end coverage

## Why

EVM callers need narrowly scoped ways to let another associated account perform common Cosmos module actions without exposing arbitrary Cosmos message execution. Each EVM-facing permission maps to the native authorization enforced by Cosmos authz.

## Staking permission

`grantStakingAuthorization` requires an explicit allowed-validator list and `maxTokens` in the base denomination. It creates three native `StakeAuthorization` grants for `MsgDelegate`, `MsgBeginRedelegate`, and `MsgUndelegate`.

Each action has an independent cumulative `maxTokens` budget. The validator allowlist applies to delegation and undelegation targets and to redelegation destinations. These constraints apply equally through the EVM precompile and native Cosmos `MsgExec`. Native delegation spends the granter's liquid balance, while the payable EVM delegation method uses the grantee's `msg.value`.

Requests above the remaining budget return a normal insufficient-funds authorization error without consuming the grant or panicking. Denomination mismatches return a distinct invalid-request error. This changes failure behavior for pre-existing native `StakeAuthorization` grants and is an intentional consensus-affecting change covered by the PR's `app-hash-breaking` release process.

Authorized redelegation preserves the existing precompile behavior of pre-withdrawing accrued destination-validator rewards to the delegator's configured withdrawal address before redelegating. The Solidity interface now documents that side effect, and unexpected pre-withdraw failures are logged.

## Governance permissions

Governance authorization is split by capability:

- `grantVoteAuthorization` / `revokeVoteAuthorization` manage only `MsgVote`; they do not authorize proposal submission or weighted votes
- `grantProposalAuthorization` / `revokeProposalAuthorization` manage only `MsgSubmitProposal`

The proposal grant is intentionally a native `GenericAuthorization`. A grantee can use it through the EVM precompile or through a native Cosmos `MsgExec`. The native path can specify an arbitrary initial deposit debited from the granter, and governance deposits can be permanently lost. The Solidity interface therefore warns callers to grant proposal authority only to fully trusted accounts.

## Behavior

[...]

PR #3985, opened 2026-08-23
Tests: Precompile Tests Phase 3 (PLT-372)

alexander-sei:

Phase 3 of the precompile test refactor ([PLT-372](https://linear.app/seilabs/issue/PLT-372/refactor-precompile-tests)), building on the phase 1 framework (#3777) and the phase 2 coverage (#3786).

Phase 2 covered the wasm-free precompiles **as they existed then**. Since that merged, `Generate v6.7 precompiles` (#3961), the scoped module authorizations (#3893, #3846) and the module Msg rpc surface (#3825) landed. This phase brings the suite back in line with the deployed interface.

### New specs

| Precompile | Spec | Coverage |
|---|---|---|
| auth (0x…100D) | `auth.spec.ts` | `account` / `accounts` / `params` / `nextAccountNumber`, LCD parity incl. the nested base account module accounts wrap theirs in |
| authz (0x…100E) | `authz.spec.ts` | `grants` / `granterGrants` / `granteeGrants`; the non-empty fixture is a `staking.grantStakingAuthorization`, since authz itself cannot grant |
| evidence (0x…100F) | `evidence.spec.ts` | `evidence` / `allEvidence` against the module's own list |
| mint (0x…1012) | `mint.spec.ts` | `params` / `minter` vs Sei's own `/seichain/mint/v1beta1/...` routes |
| params (0x…1013) | `params.spec.ts` | `params(subspace, key)` vs the staking module, plus the unknown/empty subspace reverts |
| slashing (0x…1014) | `slashing.spec.ts` | `params` / `signingInfo` / `signingInfos`, grant + revoke of unjail authorization, and the unjail guards |
| upgrade (0x…1015) | `upgrade.spec.ts` | `currentPlan` / `appliedPlan` / `upgradedConsensusState` / `moduleVersions` |

### Existing specs extended

- **bank** — `spendableBalances`, `totalSupply`, `params`, `denomMetadata` (via a tokenfactory denom), `denomsMetadata` (pages until it reaches that denom, so `nextKey` round-trips).
- **staking** — `grantStakingAuthorization` → `delegateWithAuthorization` → Cosmos-side delegation → `revokeStakingAuthorization` → reverts; `redelegate`/`undelegateWithAuthorization`; and the remaining query surface (`validators`, `validatorDelegations`, `unbondingDelegation`, `delegatorDelegations`, `redelegations`, `historicalInfo`, …).
- **gov** — vote and proposal authorizations (`grant`/`voteWithAuthorization`/`submitProposalWithAuthorization`/`revoke`), plus `proposal`, `proposals`, `getVote`, `votes`, `params`, `getDeposit`, `deposits`, `tallyResult`.
- **distribution** — `withdrawValidatorCommission`, the withdraw authorizations, and `params`, `validatorOutstandingRewards`, `validatorCommission`, `validatorSlashes`, `delegationRewards`, `delegatorValidators`, `delegatorWithdrawAddress`, `communityPool`.

### Framework

- New addresses on `PRECOMPILE_ADDRESSES`.
- `cosmosUtils.cosmosRest` — an LCD reader for the modules cosmjs does not wrap (mint, params, upgrade, evidence, slashing, authz, auth). Every path was taken from the module's decoded `pb.gw.go` gateway pattern rather than guessed, and every response field from its proto tag; the API server marshals with `OrigName` + `EmitDefaults`, so the reads are snake_case.
- Bootstrap pool 48 → 80, with the `claimPool` budget (29) recorded next to it — a non-blocking note from the phase 2 review.

Out of scope, unchanged: wasm-gated flows (wasmd, pointer `addCW*`, solo), the ibc precompile, and feegrant (0x…1010), which is not registered and whose module was removed in #3958. oracle stays the phase 2 retirement assertion.

- `npx tsc --noEmit` clean; no lint errors.
- Every assertion was reviewed against the Go executor and the sei-cosmos keeper behind it rather than against assumption. That pass corrected, among others: `historicalInfo` pinning an unreachable executor fallback instead of the querier's `historical info for height %d not found`; `nextAccountNumber` reading `next_account_number` where the response field is `count`; and mint's LCD path, which was three guesses deep with the real route last.
- It also removed tests that could not fail — a `historicalInfo` try/catch asserting success in both branches, `evidence` hardcoding an empty chain in four places, `authz` index-matching expirations across two independently ordered lists, `bank.denomsMetadata` asserting only "is an array" — and replaced receipt-only verification of the authorized reward withdrawal with the `DelegationRewardsWithdrawn` credit its non-authz sibling already pins.

**Not yet run against a live cluster.** The remaining risk is concentrated in real chain responses, so this wants a `make docker-cluster-start` + `npm run precompile:ci` run before merge; I will post the result on this PR.

🤖 Generated with [Cursor](https://cursor.com)


alexander-sei:
Addressed all three of @seidroid's findings, plus the Bugbot note, in f996dac. Each claim was re-verified against the source rather than taken on the review's word.

**[blocker] `slashing.spec.ts` — racing block production.** Confirmed: `infractions.go:72` advances `IndexOffset` for every validator in the last commit, so the two `eth_call`s at `latest` really could straddle a block. Both reads now pin the same `blockTag`, following the precedent in `distribution.spec.ts`. I took the pinning option rather than narrowing `asSigningInfo`, since that keeps the full-struct equality instead of dropping the moving fields.

**[suggestion] `bank.spec.ts` — single-page assumption.** Real latent flake, and self-inflicted: this suite mints a tokenfactory denom on every run, and `factory/…` sorts before `usei`. `totalSupply` now walks pages the way the `denomsMetadata` test below it does. The `supply('usei')` bracketing is unchanged, since supply legitimately grows between reads.

**[suggestion] `distribution.spec.ts` — vacuous on empty.** `expectCoinArray` now requires non-empty by default, with a `mayBeEmpty` opt-out for `communityPool` — sei-cosmos defaults `community_tax` to **0%** (`x/distribution/types/params.go:28`), unlike upstream's 2%, so an empty pool is the correct answer there rather than a defect.

One correction to the reasoning, since the conclusion is right but the mechanism isn't: the accruing delegation does **not** by itself guarantee `validatorCommission` is non-empty — commission is the operator's cut, so a 0%-rate validator would leave it empty regardless of how healthy the delegation is. What actually guarantees it here is that the devnet's genesis validators are created by `seid gentx` with no `--commission-rate` (`docker/localnode/scripts/step1_configure_init.sh:58`), taking the CLI default of 10%, and `DefaultMinCommissionRate` is 5% so no validator can drop to zero. I tightened the assertion and recorded that reason.

**[Bugbot, low] `gov.spec.ts` — self-contradicting dispatch section.** Also correct. `Execute` dispatches all eight query methods *before* the `readOnly` check, so gov views do answer under STATICCALL and only transaction methods are refused; the stale "gov has no view methods" label is now accurate. The duplicate `vote` STATICCALL case is removed, so the suite no longer mines the same failing tx twice.

`npx tsc --noEmit` clean, no lint errors. Still not run against a live cluster — that remains the gate before merge and I'll post the result here.

alexander-sei:
@seidroid review

PR #3989, opened 2026-08-24
Add frozen RPC router and Docker integration cluster

codchen:

- Add `frozen-rpc-router`, an EVM JSON-RPC proxy that routes block-scoped requests to the matching frozen node and all other traffic to the live node.
- Add a Docker Compose topology with non-validating archival nodes frozen before heights 10 and 20, two live validators, and the router exposed on port 8553.
- Add integration coverage that proves heights 9, 15, and 20 reach the height-10 frozen node, height-20 frozen node, and live node respectively, and verifies that cross-node block ranges are rejected.
- Build and package the router alongside `seid`, and add the topology to the integration-test CI matrix.

- `make frozen-rpc-router-integration-test`
- `go test -race ./cmd/frozen-rpc-router`
- `go test -tags frozen_rpc_integration -run '^$' ./integration_test/frozen_rpc_router/...`
- `golangci-lint run --build-tags frozen_rpc_integration ./cmd/frozen-rpc-router/... ./integration_test/frozen_rpc_router/...`
- `make fmtcheck`
- `git diff --check`

### DIFF
--- a/.github/workflows/integration-test-matrix.json
+++ b/.github/workflows/integration-test-matrix.json
@@ -238,6 +238,13 @@
     ]
   },
   {
+    "name": "Frozen RPC Router",
+    "cluster": "frozen-rpc-router",
+    "scripts": [
+      "go test -tags frozen_rpc_integration -v -count=1 -timeout 5m ./integration_test/frozen_rpc_router/..."
+    ]
+  },
+  {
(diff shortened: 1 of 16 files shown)


masih (approved):
nice work Tony 🙌

PR #3990, opened 2026-08-24
Disable mempool traffic in freeze mode

masih:
Freeze mode prevents the node from executing future blocks, so transactions accepted through RPC or P2P cannot be committed. Continuing to validate, retain, and gossip them wastes application and network resources and leaves stale transactions in the mempool.

Keep the internal mempool required by block execution, but do not expose it to RPC or start the mempool reactor when freeze mode is enabled.


sei-will (approved):

masih:
@seidroid review

masih:
@seidroid review

masih:
@seidroid review

masih:
@seidroid review

sei-will (approved):

PR #3991, opened 2026-08-24
fix(metrics): wire pending nonce events in mempool OTel counter (PLT-1017)

amir-deris:
## Summary
- Wire `tendermint_mempool_pending_nonce` OTel counter at the actual pending-nonce lifecycle points in `sei-tendermint/internal/mempool/tx.go` (`added`, `expired`, `rejected`, `accepted`)
- Remove the dead `evm_pending_nonce` counter from `x/evm/ante/metrics.go` (defined but never incremented since #3410 moved pending-nonce tracking into the mempool)

Closes PLT-1017 (sei-chain side).

## Related PR
- **Dashboards:** sei-protocol/platform#1564 — updates App/EVM Monitoring (OTel) panels to query `sei_chain_tendermint_mempool_pending_nonce_total`

Deploy and merge this PR before (or together with) the platform PR so the new series exists when dashboards roll out.


sei-tendermint/internal/mempool/tx.go
```diff
@@ -596,6 +605,9 @@ func (s *txStore) Update(spec updateSpec) {
 			_, executed := spec.TxResults[wtx.Hash()]
 			remove := invalid || executed || (expired && (s.config.RemoveExpiredTxsFromQueue || !inner.isReady(wtx)))
 			if remove {
+				if expired && wtx.evm.IsPresent() {
+					recordPendingNonceExpired()
+				}
```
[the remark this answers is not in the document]
masih: This one seems genuine?

masih (approved):

x/evm/ante/metrics.go
```diff
@@ -26,12 +25,6 @@ var (
 		// Association errors
 		associationError metric.Int64Counter
 	}{
-		pendingNonce: must(meter.Int64Counter(
-			"evm_pending_nonce",
```
bdchatham, line 30: Any concern with this going away? Any alerts?

bdchatham (approved):

x/evm/ante/metrics.go
amir-deris: I believe there shouldn't be any concern, as this metric was already not being emitted. 

```
The metric was already dead
On main, evm_pending_nonce in x/evm/ante/metrics.go is defined but never incremented. Nothing in x/evm calls evmAnteMetrics.pendingNonce:

sig.go → nonceMismatch
fee.go → effectiveGasPrice
preprocess.go → associationError
Instrumentation moved out of ante when #3410 removed mempool callbacks. At the same time, app/ante/evm_checktx.go stopped emitting app_pending_nonce (that metric definition was removed too). Pending-nonce lifecycle now lives entirely in the Tendermint mempool (tx.go), which PR 3991 wires up.

So removing the dead definition is cleanup, not a behavioral change — production has not been emitting either old series since #3410.

PR #3996, opened 2026-08-24
fix(sei-tendermint): address mac/linux test node divergence

shemnon:
Two low level mac/linux test fix. First, SO_REUSEADDR permits listen() reuse on Linux but not on mac. Second, allocate temp files via the go testing framework.

## Summary

- Use released ephemeral loopback addresses for node tests whose servers bind through net.Listen.
- Ensure helper-created nodes receive unique RPC listener addresses.
- Use a test-local directory for the private-validator Unix socket.
- Avoid tcp.TestReserveAddr, whose reserved socket can conflict with net.Listen on macOS.

## Testing

- Verified the sei-tendermint/node test suite passes.

### DIFF
--- a/sei-tendermint/node/node_test.go
+++ b/sei-tendermint/node/node_test.go
@@ -9,7 +9,6 @@ import (
 	"net"
 	"net/http"
 	"os"
-	"path/filepath"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


sei-tendermint/node/node_test.go
```diff
@@ -236,7 +237,7 @@ func TestNodeSetPrivValTCP(t *testing.T) {
 func TestPrivValidatorListenAddrNoProtocol(t *testing.T) {
 	ctx := t.Context()
 
-	addrNoPrefix := tcp.TestReserveAddr().String()
+	addrNoPrefix := "127.0.0.1:1234"
```
wen-coding: Is this guaranteed to never conflict with randomly generated addr?

bdchatham (approved):

wen-coding (approved):

sei-tendermint/node/node_test.go
shemnon: seidroid led me astray

PR #3998, opened 2026-08-24
fix(sei-tendermint): publish Autobahn block events (CON-352)

shemnon:
Publish committed Autobahn blocks and transactions to the event bus so indexer-backed RPC methods can observe inclusion.

- Wire the node event bus into validator and fullnode Giga routers
- Publish block, header, and transaction events after application commit
- Cover transaction indexing, empty blocks, result-count mismatches, and broadcast_tx_commit end to end

## Summary

Autobahn committed blocks were not publishing block and transaction events. As a result, the transaction indexer never observed Autobahn transactions, causing indexer-backed RPC methods such as broadcast_tx_commit and /tx
to wait until timeout.

- Wire the node event bus into validator and fullnode Giga routers.
- Publish NewBlock, NewBlockHeader, and per-transaction events after application commit.
- Validate transaction and ABCI result counts before dispatching events.
- Cover transaction indexing, empty blocks, mismatched results, and broadcast_tx_commit end to end.

## Testing

- go test ./internal/p2p ./internal/rpc/core ./node -count=1
- make fmtcheck

### DIFF
--- a/sei-tendermint/internal/p2p/giga_router.go
+++ b/sei-tendermint/internal/p2p/giga_router.go
@@ -52,6 +52,10 @@ type GigaRouterCommonConfig struct {
 
 	// Whether validator should proxy txs which do not belong to the local node.
 	EnableEvmProxy bool
+
+	// EventBus receives NewBlock, NewBlockHeader, and per-tx events after
+	// each Autobahn commit.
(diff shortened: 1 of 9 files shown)


shemnon:
not using event bus to solve CON-352

PR #4004, opened 2026-08-25
Attach a static linux/arm64 seid binary to releases

monty-sei:
Releases carry `linux/amd64` only. #3932 arch-split the musl link path, which made an arm64 static build possible; this wires it through the build script, the CI gate and goreleaser so the archive actually ships.

## Verified end to end

`goreleaser release --snapshot` on this branch:

```
• archiving  name=dist/sei-chain_..._linux_arm64.tar.gz
• archiving  name=dist/sei-chain_..._linux_x86_64.tar.gz

sei-chain_..._linux_arm64.tar.gz:  OK   -> ELF 64-bit LSB executable, ARM aarch64
sei-chain_..._linux_x86_64.tar.gz: OK   -> ELF 64-bit LSB executable, x86-64
```

Both archives are listed in `checksums.txt`, `sha256sum -c --ignore-missing` reports OK for both, and each archive contains a binary of the matching architecture. That last check matters because `--ignore-missing` exits 0 silently for a file that is present but unlisted, so an out of band upload would give operators a verification step that verifies nothing.

## The toolchain pin is needed on both architectures

The gcc>=12 unwind b-tree that crashed the amd64 binary at the genesis wasm store is not amd64 specific. `ATOMIC_FDE_FAST_PATH` is gated on atomics support rather than a target allowlist, and an unpinned arm64 build SIGSEGVs on the **first** boot under `RAYON_NUM_THREADS=1`, reproduced on native arm64 hardware. So the Alpine 3.15 gcc 10.3.1 libgcc is now vendored for both, each in its own subdirectory, with provenance and checksums in the README.

With the pin applied, the arm64 binary carries zero b-tree symbols and boots 8/8 clean at 4 CPUs.

## Build script

`build-static.sh` takes a target architecture and writes `build/seid-<arch>`, so the two builds do not overwrite each other on one runner. The libgcc directory and the checksums it verifies are both derived from that one argument, so a build cannot verify one architecture's archives while linking another's. It also asserts the ELF machine of the output matches what was requested.

Two guards that were failing open are now closed:

- The `nm` b-tree assertion materialises the symbol table before grepping. Reading `nm` through a pipe reported **grep's** exit status, so a failing `nm` printed "pre-b-tree unwinder confirmed" and passed.
- `boot-smoke.sh` refuses a binary built for another architecture. It previously reached `seid init`, died with an exec-format error, and reported "did not reach the ABCI handshake", which reads as a crashing binary rather than the wrong file being passed in.

## Where the arm64 binary gets booted

The release runner is amd64 and cannot execute the arm64 binary, so the goreleaser hook boots only the amd64 one. The arm64 8-boot gauntlet runs in the new `Linux ARM64 (static)` job on native `ubuntu-24.04-arm` hardware.

The link between them is reproducibility: the static build produces a byte-identical binary for a given commit (measured, identical `sha256` across independent runs), so the binary CI boots is the one that ships. The release job builds arm64 under emulation via binfmt, which costs roughly 45 minutes and is why the boot gate lives on native hardware instead.

## Notes for review

- No `uci` change needed. Docker and privileged runs already work on the release runner, so binfmt registration is just another `before:` hook.
- The workflow file changes cannot be carried by the backport bot and need a manual cherry-pick for release branches.
- Docs are drafted separately in sei-protocol/sei-docs#64, held in draft until an arm64 tarball exists on a published release.


monty-sei:
@seidroid please re-review, all three findings are addressed in the last two commits.

- binfmt image pinned by digest (`tonistiigi/binfmt:qemu-v8.1.5@sha256:2d2918e8...`)
- positive control added ahead of the b-tree assertion, so a stripped or truncated symbol table can no longer satisfy it vacuously
- the arm64 binary now boots in the release hooks under emulation, so the artefact that ships is the one gated rather than one assumed byte-identical to it. `build-static.sh` also prints each binary's sha256 so drift between a CI build and a later release build is visible

monty-sei:
@seidroid please re-review. Rebased onto current main (that also clears the stale `buf breaking`, which was comparing against a base that had moved 20 protos), and both blockers from the last round are fixed in `5eba408`. All 68 checks green including `Linux ARM64 (static)`.

Two things worth calling out, since both are corrections to fixes made during review rather than to the original PR:

**The arm64 boot hook could never have executed.** goreleaser execs a before-hook directly rather than through a shell, so the `BOOT_TIMEOUT=300` prefix was taken as the program name: `exec: "BOOT_TIMEOUT=300": executable file not found in $PATH`. The before pipe aborts on that, so a release would have published a tag with no artefacts at all, including the amd64 tarball that ships today. Reproduced on goreleaser 2.15.4 and fixed by going through `env`.

**BOOT_TIMEOUT was the cost of every boot, not a ceiling**, since `seid start` never exits on its own, so the 300s budget would have been paid on every healthy boot. A boot now ends shortly after the handshake with a brief dwell, which took 2 boots of the release binary from ~50s to 11s and is what makes 4 boots at a 300s ceiling affordable.

On the native-broken-binary case you flagged: the probe now runs for every binary and only the message branches on architecture. Verified each cause reports itself: missing path, non-executable, non-ELF, truncated header, unrunnable foreign binary, and a native binary whose `seid version` fails.

Also re-ran the controls after the early-exit change, since it could have weakened the gate: the preserved b-tree binary is still caught on boot 1, the pinned binary still passes.

monty-sei:
@seidroid please re-review `1d17a384`, which takes findings 1, 2 and 3 plus the README nit. All 68 checks green (one integration job timed out at 35 min and passed on re-run; the diff touches only release tooling, nothing integration tests reach).

**Finding 2 is the one worth looking at.** Added `scripts/check-goreleaser-hooks.sh`, which shell-splits every `before.hooks` entry and asserts `command -v` resolves argv[0], wired into the amd64 static job that the `.goreleaser.yaml` path filter already routes. Verified by reintroducing the original defect: it exits 1 and names the fix. That covers all six hooks rather than the one that broke, which was your point.

Writing it turned up a bug in the check itself, worth mentioning since it is the same class: I used `mapfile`, which is bash 4+, so it would have passed on the CI runner and failed for anyone on macOS. Replaced with a portable read loop.

**Finding 1** — you were right that the argument supports changing the code, not the comment. Hook is now 2 boots.

**Finding 3** — the crash grep reads a snapshot taken before the TERM. I re-ran the controls afterwards since the change could have weakened discrimination: the preserved b-tree binary is still caught on boot 1, the pinned binary still passes, two boots in 10s.

**Finding 5** — checked and it is a non-issue: `uci/.github/workflows/goreleaser-release.yml@v0.0.11` sets no `timeout-minutes` anywhere, so the job inherits GitHub's 360 minute default. I should have said so last round rather than leaving it looking unaddressed.

**Finding 4** left alone as pre-existing, happy to take it if you would rather it landed here.

masih:
@seidroid review

masih (approved):

PR #4005, opened 2026-08-25
Consolidate lint CI: fold UCI go-lint into golangci workflow (PLT-1069)

amir-deris:
## Summary

- Fold the UCI `Go / Lint` checks (`go mod tidy`, `go mod verify`, `go generate ./...` + drift gates) into `.github/workflows/golangci.yml`
- Delete the now-redundant `.github/workflows/uci-go-lint.yml` and `.github/actions/go-lint-setup` (the disk reclaim it did was only needed for UCI's full-tree `go vet ./...`)
- Align `make lint` and `AGENTS.md` with the consolidated gate: drop `go vet ./...` (already covered by golangci's `govet` linter), add `go generate ./...`

Expected CI savings: ~4–5 min PR wall-clock (UCI was the long pole at ~12 min) and ~10 min runner time per PR.

## Follow-up (not in this PR)

Once this merges and the consolidated `lint` job is green on main, **Platform needs to remove `Go / Lint` from branch protection** so PRs no longer wait on the deleted UCI workflow.

## Test plan

- [ ] `lint` job passes on this PR (golangci-lint, fmt, go mod tidy, go mod verify, go generate)
- [ ] Confirm consolidated job runtime is ~7–8 min vs prior ~12 min wall-clock gate
- [ ] `make lint` passes locally


masih (changes requested):
this forks code right out of UCI and that's not what we want in terms of consolidating a unform quality gate across all repos.

The time savings do not matter i would imagine if the slowest job required to pass for PRs takes as long still. right?

What have I missed?

amir-deris:
> this forks code right out of UCI and that's not what we want in terms of consolidating a unform quality gate across all repos.
> 
> The time savings do not matter i would imagine if the slowest job required to pass for PRs takes as long still. right?
> 
> What have I missed?

Thanks for feedback. This PR optimizes sei-chain CI locally, but it works against the cross-repo uniformity UCI is meant to provide. Consolidation of the CI jobs toward UCI is the right approach, will drop this PR for now to work on higher priority items.

PR #4012, opened 2026-08-25
fix(giga): fail-fast broadcast_tx_commit and feed newBlockFilter from the notifier

shemnon:
## Summary
- Autobahn `BroadcastTxCommit` (`-b block`) returns `ErrBroadcastTxCommitUnsupported` before `InsertTx` instead of waiting on an empty EventBus/KV indexer (CON-352 hang). Comet is unchanged.
- `eth_newBlockFilter` / `eth_getFilterChanges` take Autobahn block hashes from `BlockHeaderNotifier` (same FinalizeBlock hash as `newHeads` / `eth_getBlockBy*`), not `/events`. HTTP FilterAPI now gets the notifier.
- CosmJS-style cosmos `/tx`, `tx_search`, TM `subscribe`, and `/events` fail-fast is **not** in this PR (fork-gated).

Linear: [CON-409](https://linear.app/seilabs/issue/CON-409/autobahn-fail-fast-broadcasttxcommit-and-keep-evm-inclusion-off)
Related: [CON-352](https://linear.app/seilabs/issue/CON-352/auditor-feedback-fix-broadcasttxcommit-timeout-under-autobahn) (do not merge EventBus PR 3998 as the fix)

## Test plan
- [ ] `go test ./sei-tendermint/internal/rpc/core/ -count=1 -run 'TestBroadcastTxCommit'` — Autobahn fail-fast; Comet still hits mempool, not the sentinel
- [ ] `go test ./evmrpc/ -count=1 -run 'TestBlockHeaderNotifier_Subscribe|TestFilterAPI_NewBlockFilter|TestFilterBlockFilter'` — notifier fan-out, Autobahn block filter hashes, Comet Events path unchanged
- [ ] Confirm `seid tx … -b block` against an Autobahn node errors immediately; `-b sync` still works
- [ ] Confirm `eth_newBlockFilter` + `eth_getFilterChanges` on Autobahn returns the overlay block hash, not `0x000…0`


sei-tendermint/internal/rpc/core/mempool.go
```diff
@@ -116,9 +116,20 @@ func (env *Environment) BroadcastTx(ctx context.Context, req *coretypes.RequestB
...
+// confirm inclusion from committed state.
+var ErrBroadcastTxCommitUnsupported = errors.New("broadcast_tx_commit is not supported on Autobahn; use broadcast_tx_sync")
+
+// BroadcastTxCommit returns the CheckTx and DeliverTx results after inclusion.
+// Under Autobahn it returns ErrBroadcastTxCommitUnsupported without submitting
+// the transaction.
 // More: https://docs.tendermint.com/master/rpc/#/Tx/broadcast_tx_commit
 func (env *Environment) BroadcastTxCommit(ctx context.Context, req *coretypes.RequestBroadcastTx) (*coretypes.ResultBroadcastTxCommit, error) {
+	if env.gigaRouter().IsPresent() {
+		return nil, ErrBroadcastTxCommitUnsupported
```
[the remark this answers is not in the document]
shemnon: "autobahn mode" detected by wethere blockHeadNotifier is nil. If in autobahn slow mode won't use broadcastTxCommit just broadcastTx

evmrpc/notifier.go
```diff
@@ -138,6 +155,15 @@ func (n *BlockHeaderNotifier) publish(evt blockHeaderEvent) {
 	case n.ch <- evt:
 	default:
 	}
+
+	n.mu.Lock()
+	listeners := make([]func(blockHeaderEvent), len(n.listeners))
+	copy(listeners, n.listeners)
+	n.mu.Unlock()
+	for _, fn := range listeners {
+		defer recoverAndLog()
```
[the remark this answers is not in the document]
shemnon: roger that, done.

evmrpc/notifier.go
```diff
@@ -115,9 +120,21 @@ func (n *BlockHeaderNotifier) OnBlockCommitted(hash []byte, header *tmproto.Head
...
+		return
+	}
+	n.mu.Lock()
+	n.listeners = append(n.listeners, fn)
+	n.mu.Unlock()
+}
+
+// publish pushes evt onto the newHeads channel with overwrite-on-full
+// semantics, then invokes every subscribe listener. Used by both
+// PublishStashed and OnBlockCommitted so the fan-out lives in one place.
```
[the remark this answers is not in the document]
shemnon: Rodger that, done.

shemnon:
@seidroid review

app/app.go
```diff
@@ -430,13 +430,15 @@ type App struct {
...
 	evmRPCConfig         evmrpcconfig.Config
-	// blockHeaderNotifier is non-nil only when Autobahn is enabled. It
-	// owns the FinalizeBlock→Commit pairing for eth_subscribe("newHeads"):
-	// FinalizeBlocker calls Stash with the (hash, header, response)
-	// tuple, App.Commit calls PublishStashed after a successful
-	// BaseApp.Commit, and FinalizeBlocker entry calls ClearStash to
-	// defend against stale tuples from prior failed commits or
-	// non-stashing return paths (EthReplay/EthBlockTest).
+	// autobahnEnabled records whether Autobahn was configured for this node.
+	autobahnEnabled bool
```
wen-coding, line 434: I understand why you added this, but right now I don't think we plan to have a node configured:
autobahnEnabled: true
blockHeaderNotifier: None

evmrpc/config/config.go
```diff
@@ -125,6 +125,12 @@ type Config struct {
 	// timeout for filters
 	FilterTimeout time.Duration `mapstructure:"filter_timeout"`
 
+	// maximum number of active eth_newFilter and eth_newBlockFilter filters
+	MaxFilters uint64 `mapstructure:"max_filters"`
```
wen-coding, line 129: I guess a natural question users will ask is: what should these two numbers be set to?

evmrpc/config/config.go
```diff
@@ -329,6 +335,8 @@ var DefaultConfig = Config{
 	CORSOrigins:                  "*",
 	WSOrigins:                    "*",
 	FilterTimeout:                120 * time.Second,
+	MaxFilters:                   1000,
```
wen-coding, line 338: Just curious: why 1000?

evmrpc/config/config.go
```diff
@@ -840,6 +866,13 @@ ws_origins = "{{ .EVM.WSOrigins }}"
 # timeout for filters
 filter_timeout = "{{ .EVM.FilterTimeout }}"
 
+# maximum number of active eth_newFilter and eth_newBlockFilter filters
+max_filters = {{ .EVM.MaxFilters }}
```
wen-coding, line 871: what is the behavior if we reached this capacity?

evmrpc/config/config_test.go
```diff
@@ -293,6 +303,8 @@ func TestReadConfig(t *testing.T) {
 	require.Nil(t, err)
 	require.False(t, cfg.EnableParallelizedBlockTrace)
 	require.True(t, cfg.EnableSimulation)
+	require.Equal(t, uint64(100), cfg.MaxFilters)
```
wen-coding: nit: can you define the constants for 100/200 somewhere?

evmrpc/filter.go
```diff
@@ -327,23 +340,80 @@ func NewFilterAPI(
...
+	}
+}
+
+// addFilter installs a polling filter when capacity is available.
+func (a *FilterAPI) addFilter(f filter) (ethrpc.ID, error) {
+	a.filtersMu.Lock()
+	defer a.filtersMu.Unlock()
+
+	if uint64(len(a.filters)) >= a.filterConfig.maxFilters {
+		return "", fmt.Errorf("too many filters: maximum allowed is %d", a.filterConfig.maxFilters)
```
wen-coding, line 392: Can an attacker somehow create 1000 filters and blocks the normal users forever?

evmrpc/filter.go
```diff
@@ -327,23 +340,80 @@ func NewFilterAPI(
...
+// BlocksSubscription and invalidates filters whose backlog is full.
+func (a *FilterAPI) appendBlockHash(evt blockHeaderEvent) {
+	hash := common.BytesToHash(evt.hash)
+	a.filtersMu.Lock()
+	defer a.filtersMu.Unlock()
+	for id, f := range a.filters {
+		if f.typ != BlocksSubscription {
+			continue
+		}
+		if uint64(len(f.blockHashes)) >= a.filterConfig.maxBlockFilterHashes {
```
wen-coding, line 374: For BlocksSubscription, when would this map be full?

evmrpc/filter_block_notifier_test.go
```diff
@@ -0,0 +1,177 @@
+package evmrpc
```
wen-coding, line 1: nit: should we fold these into filter_test.go?

evmrpc/filter.go
```diff
@@ -503,6 +573,9 @@ func (a *FilterAPI) GetFilterChanges(
 	result := []*ethtypes.Log{}
 	switch filter.typ {
 	case BlocksSubscription:
+		if a.blockHeaderNotifier != nil {
+			return a.takeBlockHashes(filterID)
```
wen-coding, line 577: So basically this is Autobahn path? Can we naively make a mistake and turn off blockHeaderNotifier for Autobahn in the future maybe?

sei-tendermint/internal/rpc/core/mempool_test.go
```diff
@@ -0,0 +1,136 @@
+package core
```
wen-coding, line 1: Should we fold into rpc/core/mempool_test.go?

evmrpc/config/config.go
shemnon: These are the defaults for geth and besu

evmrpc/filter.go
shemnon: This would only be an attack against a specific RPC. With this config the win is denying others filters, without it the win is OOM panic on the process.  Users who get denied a filter can move onto another RPC node.

app/app.go
shemnon: SeiDroid was very insistent on not relying on blockHEaderNotifier being not nil for autobahn, so I pushed it up higher and triggered off of autobahn config file.  Is there another flag I should look at?

evmrpc/config/config.go
shemnon: This is a devops tuning question for available memory.  Unless they are tuning RPC they should never touch it.  Conseumed memory is number of blocks over 2 minutes * max filters * blockhash size.

evmrpc/config/config.go
shemnon: RPC fails when asking to add a filter.

If blocks per filter exceeds max (if block times go very low) then it's a ring buffer and oldest unpolled block hash disappears.

app/app.go
wen-coding: Yeah, I understand why SeiDroid wants it, but I also don't want to support all possible combinations too.

You can ignore SeiDroid if you really disagrees with its opinions. Or maybe we can panic on combinations we don't support.

evmrpc/filter.go
shemnon: If there are more blocks in the filter window than the hashes, so if block rate goes up (via configuration) or the reaper is slow, or it gets misconfigured.

codchen (approved):

evmrpc/config/config.go
wen-coding: Hmm, I think there is a comment later which says "the filter is invalidated if it falls farther behind." on the blocks per filter.

Anyways, can we maybe comment here what happens when the capacity is reached.

wen-coding (approved):

evmrpc/server.go
```diff
@@ -90,7 +91,7 @@ func NewEVMHTTPServer(
 
 	globalBlockCache := NewBlockCache(3000)
 	cacheCreationMutex := &sync.Mutex{}
-	sendAPI := NewSendAPI(tmClient, txConfigProvider, NewSendConfig(config.Slow, config.EnableSimulation), k, beginBlockKeepers, ctxProvider, homeDir, simulateConfig, app, antehandler, ConnectionTypeHTTP, methodTimeout, globalBlockCache, cacheCreationMutex, watermarks)
+	sendAPI := NewSendAPI(tmClient, txConfigProvider, NewSendConfig(config.Slow, config.EnableSimulation, blockHeaderNotifier != nil), k, beginBlockKeepers, ctxProvider, homeDir, simulateConfig, app, antehandler, ConnectionTypeHTTP, methodTimeout, globalBlockCache, cacheCreationMutex, watermarks)
```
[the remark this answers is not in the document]
shemnon: done

sei-tendermint/internal/rpc/core/mempool_autobahn_test.go
```diff
@@ -0,0 +1,133 @@
...
+			Connection:               conn.DefaultMConnConfig(),
+			IncomingConnectionWindow: utils.Some(time.Duration(0)),
+			MaxAcceptRate:            rate.Inf,
+			MaxDialRate:              rate.Inf,
+			Giga:                     utils.Some[p2p.GigaRouter](giga),
+		},
+	)
+	require.NoError(t, err)
+
+	// Setup: empty KV indexer; TimeoutBroadcastTxCommit stays 0 so a wait regression hangs.
```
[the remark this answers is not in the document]
shemnon: done

evmrpc/config/config_test.go
shemnon: Done

evmrpc/filter.go
```diff
@@ -327,23 +330,60 @@ func NewFilterAPI(
...
+// live BlocksSubscription.
+func (a *FilterAPI) appendBlockHash(evt blockHeaderEvent) {
+	hash := common.BytesToHash(evt.hash)
+	a.filtersMu.Lock()
+	defer a.filtersMu.Unlock()
+	for id, f := range a.filters {
+		if f.typ != BlocksSubscription {
+			continue
+		}
+		f.blockHashes = append(f.blockHashes, hash)
```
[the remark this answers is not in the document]
shemnon: There is a cap on filters, so there is a ceiling. It is also devops configurable.

evmrpc/filter.go
shemnon: sedroid was insistent there was a possible future where v2 has blockHeaderNotifier and not autobahn that we need to keep wired up. The block hash push depends on a non-nill blockHeaderNotifier, so that is the gate.

sei-tendermint/internal/rpc/core/mempool_test.go
shemnon: There is no rpc/core/mempool_test.go.  We have internal/autobahn/producer, internal/consensus, internel/mempool, types, and test/fuzz/tests.  I'll just call it mempool_test.go

evmrpc/filter_block_notifier_test.go
shemnon: filter_test is `package evmrpc_test` which presents scoping problems, it's a black box HTTP test. so a suffixed name in the evmrpc pacakge is the right scope here.

evmrpc/config/config.go
shemnon: There are two axes: one is how recently was it polled (after 120s it starts to fall behind) and the other is how many filters in general have been applied.  so (a) use it or lose it and (b) not too many at once.

Docs tweaked.

app/app.go
shemnon: The reason I leaned into seidroid is it also increased readability, _why_ are we skipping this option? then name is in the variable vs. switching on the impact of the config.

PR #4013, opened 2026-08-25
Fix intermittent failures in buf breaking change check

masih:
Check the comparison base out locally so `buf breaking` never clones
over the network: the buf CLI caps every command at 120s, and its
anonymous HTTPS fetch of this repository does not reliably finish in
that budget, which made the breaking step fail intermittently.

### DIFF
--- a/.github/workflows/buf-check.yml
+++ b/.github/workflows/buf-check.yml
@@ -16,8 +16,20 @@ jobs:
     name: Check
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


bdchatham (approved):

masih:
@seidroid review

masih:
@seidroid review

masih:
@seidroid review

masih:
@seidroid review

PR #4014, opened 2026-08-25
Detach applications from retired IBC core

masih:
Remove IBC keeper, module, genesis, simulation, params, and begin-block
wiring from the production and standalone wasm applications.

Delete dormant CosmWasm IBC lifecycle callbacks and state-backed queries,
returning deterministic unsupported errors for retired contract IBC
operations.

Keep the ibc, transfer, capability, and feegrant stores mounted, preserve
the retired transfer module account, and retain the ibc/103 raw-store
query tombstone. Leave the unreferenced sei-ibc-go tree for a follow-up
mechanical deletion.


masih:
@seidroid review

app/ante/cosmos_checktx.go
```diff
@@ -110,7 +109,7 @@ func CosmosCheckTxAnte(
 	}
 	ctx = DecoratePriority(ctx, priority)
 
-	return ctx, ibcante.RejectMessages(tx)
+	return ctx, nil
```
philipsu522, line 113: i think the error was introduced previously - are we moving it back to nil because ibc is completely removed?

sei-wasmd/app/app.go
```diff
@@ -616,7 +627,9 @@ func (app *WasmApp) InitChainer(ctx sdk.Context, req abci.RequestInitChain) abci
 
 	app.upgradeKeeper.SetModuleVersionMap(ctx, app.mm.GetVersionMap())
 
-	return app.mm.InitGenesis(ctx, app.appCodec, genesisState, genesis.GenesisImportConfig{})
+	response := app.mm.InitGenesis(ctx, app.appCodec, genesisState, genesis.GenesisImportConfig{})
+	app.accountKeeper.GetModuleAccount(ctx, retiredTransferName)
```
philipsu522: why do we need to get this module if it's not being used?

sei-wasmd/app/app.go
```diff
@@ -94,7 +92,9 @@ import (
 const (
 	appName              = "WasmApp"
 	feegrantStoreKeyName = "feegrant"
-	retiredTransferName  = "transfer" // retained store and module account for historical state
+	retiredIBCStoreName  = "ibc"
+	retiredTransferName  = "transfer"
+	capabilityStoreName  = "capability"
```
philipsu522: nit - should we prefix this with "retired" as well for consistency?

philipsu522 (approved):

app/ante/cosmos_checktx.go
masih: yup, in the next PR

masih:
@seidroid review

PR #4015, opened 2026-08-25
Remove unreferenced IBC core

masih:
Delete the remaining IBC core keepers, light clients, CLI, simulations,
specifications, protobuf sources, and generated code after application
wiring has been detached.

Remove obsolete Buf and Swagger artifacts and use dependency-free raw
store constants in tooling.

Keep the ibc, transfer, capability, and feegrant stores mounted, retain
the transfer module account and retired raw-store query tombstone, and
leave CosmWasm IBC wire requests returning deterministic unsupported
errors for v6.7 while v6.6 freeze nodes serve historical IBC data.


masih:
Buf failure is expected since we are dropping a whole bunch of protos.

alexander-sei (approved):

philipsu522 (approved):

PR #4016, opened 2026-08-26
Remove stale ibc from v6.7 module version map

masih:
Remove stale IBC from v6.7 module version map and apply minor cleanups
that are missed in previous IBC removal.

Remove obsolete relayer tests.


masih:
@seidroid review

alexander-sei (approved):

sei-will (approved):

PR #4021, opened 2026-08-26
PLT-1072: Wire rate limiter into native gRPC (:9090) unary+stream interceptors

amir-deris:
## Summary

Wires the shared `ratelimiter.Registry` (from [PLT-411](https://linear.app/seilabs/issue/PLT-411) / [PLT-800](https://linear.app/seilabs/issue/PLT-800)) into native gRPC on `:9090` via unary and stream server interceptors. Rejections return `codes.ResourceExhausted` and emit `rpc_rate_limit_rejected_total{plane="grpc", method_namespace="..."}`.

This is the last Phase 1 plane; EVM HTTP ([PLT-819](https://linear.app/seilabs/issue/PLT-819)), CometBFT HTTP ([PLT-981](https://linear.app/seilabs/issue/PLT-981)), and the core registry/parser are already landed.

- **Interceptors** (`sei-cosmos/server/grpc/rate_limit.go`) — per-IP token bucket at unary call and stream establishment; uses `Registry.IPFromGRPCContext` and `info.FullMethod` (no `MethodParser` / body pre-read needed on gRPC).
- **Server wiring** (`StartGRPCServer`) — interceptors chained at server creation, before BaseApp query handler registration, so admission runs before `sdk.Context` creation.
- **Config** (`[grpc]` in `app.toml`) — `rate_limiting_enabled`, `ip_rate_limit_rps`, `ip_rate_limit_burst`, `trusted_proxy_cidrs`; ships **disabled by default** (same rollout pattern as 1b/1c).
- **Metrics bucketing** (`ratelimiter/method_bucket.go`) — adds `PlaneGRPC` and low-cardinality service-name labels for known protobuf services.

**Out of scope:** gRPC-Web (`:9091`, Phase 4d).

## Test plan

- [x] `go test ./ratelimiter/... ./sei-cosmos/server/grpc/... ./sei-cosmos/server/config/...`
- [x] Unary interceptor: allow under limit, burst then `ResourceExhausted`, per-IP isolation, trusted-proxy XFF
- [x] Stream interceptor: allow then reject at stream open
- [x] Config: defaults, absent-key reads, manifest/golden updates for new `[grpc]` keys
- [ ] Manual: enable `grpc.rate-limiting-enabled = true` on a dev node and confirm 429-equivalent gRPC rejections under burst load


sei-cosmos/server/grpc/server.go
```diff
@@ -41,7 +42,19 @@ func StartGRPCServer(clientCtx client.Context, app types.Application, cfg config
...
+		if err != nil {
+			return nil, fmt.Errorf("grpc rate limiter: %w", err)
+		}
+		serverOpts = append(serverOpts,
+			grpc.ChainUnaryInterceptor(UnaryRateLimitInterceptor(rateLimitRegistry)),
+			grpc.ChainStreamInterceptor(StreamRateLimitInterceptor(rateLimitRegistry)),
+		)
+	}
+
+	grpcSrv := grpc.NewServer(serverOpts...)
```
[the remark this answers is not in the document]
amir-deris: Confirmed and fixed in c6808999a.

Verified the mechanism: grpc-go v1.57.1 passes `s.opts.unaryInt` to `MethodDesc.Handler` (`server.go:1358`) rather than invoking it, and `BaseApp.RegisterGRPCServer` discarded that argument.

The fix is in `sei-cosmos/baseapp/grpcserver.go`, not here — that wrapper handler is the single point every unary query passes through, so threading the interceptor in there covers this limiter and any future server-level interceptor without each registration site re-adding it. New `chainQueryInterceptors` orders recovery -> server interceptor -> query-context, with the query-context interceptor last so admission control rejects before a call costs a query context. Nil-guarded for servers built without one.

Regression guards added in `sei-cosmos/baseapp/grpcserver_test.go` and `sei-cosmos/server/grpc/rate_limit_dispatch_test.go`, both registering through `RegisterGRPCServer` on a real `grpc.Server` and dialling it. Confirmed they fail with the fix reverted.

sei-cosmos/server/grpc/server.go
```diff
@@ -41,7 +42,19 @@ func StartGRPCServer(clientCtx client.Context, app types.Application, cfg config
...
 		}),
-	)
+	}
+	if cfg.RateLimitingEnabled {
+		rateLimitRegistry, err := ratelimiter.New(cfg.RateLimiterConfig())
+		if err != nil {
+			return nil, fmt.Errorf("grpc rate limiter: %w", err)
+		}
+		serverOpts = append(serverOpts,
+			grpc.ChainUnaryInterceptor(UnaryRateLimitInterceptor(rateLimitRegistry)),
```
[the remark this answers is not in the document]
amir-deris: Confirmed and fixed in c6808999a.

Verified the mechanism: grpc-go v1.57.1 passes `s.opts.unaryInt` to `MethodDesc.Handler` (`server.go:1358`) rather than invoking it, and `BaseApp.RegisterGRPCServer` discarded that argument.

The fix is in `sei-cosmos/baseapp/grpcserver.go`, not here — that wrapper handler is the single point every unary query passes through, so threading the interceptor in there covers this limiter and any future server-level interceptor without each registration site re-adding it. New `chainQueryInterceptors` orders recovery -> server interceptor -> query-context, with the query-context interceptor last so admission control rejects before a call costs a query context. Nil-guarded for servers built without one.

Regression guards added in `sei-cosmos/baseapp/grpcserver_test.go` and `sei-cosmos/server/grpc/rate_limit_dispatch_test.go`, both registering through `RegisterGRPCServer` on a real `grpc.Server` and dialling it. Confirmed they fail with the fix reverted.

sei-cosmos/server/config/config_fuzz_test.go
```diff
@@ -1354,7 +1358,7 @@ func TestBaseConfigManifestNamesEveryField(t *testing.T) {
 	)
 }
 
-// grpcKeys covers the three [grpc] keys read as plain casts.
+// grpcKeys covers the four [grpc] keys read as plain casts.
```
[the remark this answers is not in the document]
amir-deris: Fixed in c6808999a. `grpcKeys` godoc now reads four rows and eleven others; `grpcKeysWithTargetsOfTheirOwn` now reads nine behind `v.IsSet` and two through the clamp. Also corrected the drifted `config.go:551-552` reference to `604-605`, where the two unconditional clamp reads now live.

amir-deris:
@seidroid review

amir-deris:
@seidroid review

sei-cosmos/server/grpc/server.go
```diff
@@ -41,7 +42,19 @@ func StartGRPCServer(clientCtx client.Context, app types.Application, cfg config
 			MinTime:             cfg.KeepaliveMinTime,
 			PermitWithoutStream: cfg.KeepalivePermitWithoutStream,
 		}),
-	)
+	}
+	if cfg.RateLimitingEnabled {
```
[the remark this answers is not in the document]
amir-deris: Addressed in fa2ebc6. The check now lives in `rateLimitServerOptions`, the one function every gRPC start passes through, using a package-level `seilog.NewLogger("cosmos", "server", "grpc")` (the pattern used elsewhere in sei-cosmos, e.g. `server/util.go:42`) rather than threading a logger through `start.go` and `testutil/network/util.go`. The message mirrors the CometBFT one.

sei-cosmos/server/grpc/rate_limit.go
```diff
@@ -0,0 +1,39 @@
...
+			return nil, errRateLimited
+		}
+		return handler(ctx, req)
+	}
+}
+
+// StreamRateLimitInterceptor returns a server interceptor that applies per-IP
+// token-bucket rate limiting when a client stream is established. registry must
+// be non-nil.
+func StreamRateLimitInterceptor(registry *ratelimiter.Registry) grpc.StreamServerInterceptor {
```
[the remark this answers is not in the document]
amir-deris: Closed rather than documented, in d6edf04. `StreamRateLimitInterceptor` now wraps the stream and charges one token per inbound `RecvMsg` on top of the establishment charge, so an admitted `ServerReflectionInfo` stream can no longer serve unbounded descriptor lookups. The charge happens after a successful read, so a stream blocked on an idle client holds no token and EOF is free. Covered by unit tests plus a real-server test over an actual reflection stream.

amir-deris:
@seidroid review

masih:
@seidroid review

sei-cosmos/server/config/testdata/grpc.keys.golden
masih, line 1: This and the server config "golden" files are now redundant and should be removed right? 

This is given the changes on main.

sei-cosmos/server/config/config_fuzz_test.go
```diff
@@ -691,6 +691,10 @@ func TestGetConfigAbsentSectionDivergences(t *testing.T) {
 			"grpc.keepalive-permit-without-stream",
 			cfg.GRPC.KeepalivePermitWithoutStream, def.GRPC.KeepalivePermitWithoutStream, false,
 		},
+		{
```
masih, line 662: Can I ask you to review the comments in this doc? it is quite verbose with godoc which is good and bad in that they rot fast if unmaintained.

sei-cosmos/server/grpc/server.go
```diff
@@ -18,14 +20,62 @@ import (
...
+		grpc.ChainUnaryInterceptor(UnaryRateLimitInterceptor(registry)),
+		grpc.ChainStreamInterceptor(StreamRateLimitInterceptor(registry)),
+	}, registry, nil
+}
+
+// registeredMethods returns the "service/Method" name of every method served by
+// srv.
+func registeredMethods(srv *grpc.Server) []string {
+	info := srv.GetServiceInfo()
+	methods := make([]string, 0, len(info))
```
masih: why prealloc by service info? we know the exact size of this slice.

sei-cosmos/server/config/config.go
```diff
@@ -258,6 +259,36 @@ type GRPCConfig struct {
...
+	TrustedProxyCIDRs []string `mapstructure:"trusted-proxy-cidrs"`
+}
+
+// RateLimiterConfig builds the ratelimiter.Config used by native gRPC admission.
+func (c GRPCConfig) RateLimiterConfig() ratelimiter.Config {
+	return ratelimiter.Config{
+		RPS:               c.IPRateLimitRPS,
+		Burst:             c.IPRateLimitBurst,
+		TrustedProxyCIDRs: c.TrustedProxyCIDRs,
+	}
```
masih, line 298: Small note that I think this would end up using a single bucket for all loopback addresses. This is OK for now. Worth documenting specifically.

sei-cosmos/server/config/testdata/grpc.keys.golden
amir-deris: Oh they were left over from merge conflict resolution. My bad, I will remove them.

sei-cosmos/server/config/config.go
amir-deris: Thanks for feedback. Will add documentation.

sei-cosmos/server/config/config_fuzz_test.go
amir-deris: Yes, for sure. I will work on cleaning up the comments.

sei-will (approved):

sei-will (approved):

PR #4026, opened 2026-08-26
Add global checkpoint scheduler

yzang2019:
Two main changes in this PR:
1. Move existing SS checkpointer to ss/snapshot
2. Create a new global checkpoint scheduler and put that under controller

This PR introduce the interface of new global scheduler, which is not reachable from production yet, nothing implements it. A follow-up PR will wire it up with existing stores.


sei-db/controller/checkpoint_scheduler.go
```diff
@@ -4,91 +4,228 @@
...
+// when a store is still writing the last checkpoint, has not reached its scheduled version, or the
+// last one finished too recently.
+func (s *CheckpointScheduler) scheduleNextCheckpoint() {
+	if s.CheckpointInProgress() {
 		return
 	}
-	barrier, ok := db.(types.DrainBarrier)
-	if !ok {
-		done(fmt.Errorf("state store backend %T does not support ordered checkpoint barriers", db))
+	if s.scheduledVersion != 0 && !s.allReached(s.scheduledVersion) {
```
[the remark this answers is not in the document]
yzang2019: Not a huge concern

sei-db/controller/checkpoint_scheduler.go
```diff
@@ -4,91 +4,228 @@
...
+	s.noteCheckpointFinished()
+	if s.withinMinTime() {
+		return
+	}
+
+	targetVersion := nextCheckpointVersion(s.stores, s.config.CheckpointInterval)
+	for _, store := range s.stores {
+		store.ScheduleCheckpoint(targetVersion)
+	}
+	s.scheduledVersion = targetVersion
```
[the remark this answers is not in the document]
yzang2019: Fixed

yzang2019:
@seidroid review

sei-db/controller/checkpoint_scheduler.go
```diff
@@ -4,91 +4,229 @@
...
+
+// CheckpointConfig is the cadence a CheckpointScheduler holds every registered store to.
+type CheckpointConfig struct {
+	// CheckpointInterval is how many blocks apart checkpoints are taken. 0 turns checkpointing off.
+	CheckpointInterval int64
+
+	// MinTimeBetweenCheckpoints is the shortest wall-clock gap allowed between one checkpoint
+	// finishing and the next being scheduled, which bounds how fast a node replaying blocks
+	// checkpoints. 0 leaves CheckpointInterval as the only pacing.
+	MinTimeBetweenCheckpoints time.Duration
```
cody-littley: It seems odd to specify a time between checkpoints instead of specifying the target rate of checkpoints. By specifying time between checkpoints, nodes will have different checkpoint frequencies depending on how fast they can checkpoint.

An alternate strategy might be to do the following:

- specify checkpoint period, e.g. "once per 10 minutes"
- if a checkpoint takes longer than 10 minutes, then do not start a new checkpoint until the old one is finished
- Two possible protocols for handling long checkpoints:
    - Option 1: if a checkpoint takes longer than the checkpoint period, start the next checkpoint right away when the previous one finishes
    - Option 2: if a checkpoint takes longer than the checkpoint period, don't write another checkpoint until we hit the next time that is a clean multiple of the checkpoint period

sei-db/controller/checkpoint_scheduler.go
```diff
@@ -4,91 +4,229 @@
...
-// CheckpointScheduler coordinates checkpoints for stores with in-flight writes.
+var checkpointLogger = seilog.NewLogger("db", "checkpoint")
+
+// checkpointPollInterval is how often the scheduler looks for a boundary to dispatch. It has to stay
+// well inside the time the stores take to cover an interval's worth of versions, or boundaries pass
+// undispatched.
+const checkpointPollInterval = 10 * time.Second
+
+// CheckpointConfig is the cadence a CheckpointScheduler holds every registered store to.
+type CheckpointConfig struct {
```
cody-littley: Optional suggestion, feel free to ignore if you don't agree.

In many other places we split configs into their own files. Does it make sense to do so here as well?

sei-db/controller/checkpoint_scheduler.go
cody-littley, line 1: Are we calling the on disk snapshots "checkpoints" now? Changed the name of the in-memory stuff to View so not to conflict with the name "snapshot". I don't mind calling them one thing or the other, I just want to make sure we settle on a uniform name in the code.

sei-db/controller/checkpoint_scheduler.go
```diff
@@ -4,91 +4,229 @@
...
-// ScheduleCheckpoint checkpoints an engine after all writes already enqueued
-// on it have been applied.
-func ScheduleCheckpoint(db types.StateStore, destDir string, shouldRun func() bool, done func(error)) {
-	cp, ok := db.(types.Checkpointable)
-	if !ok {
-		done(fmt.Errorf("state store backend %T does not support checkpoints", db))
+// scheduleNextCheckpoint runs one cycle: it hands the next boundary to every store, or does nothing
+// when a store is still writing the last checkpoint, has not committed past its scheduled version, or
+// the last one finished too recently.
+func (s *CheckpointScheduler) scheduleNextCheckpoint() {
```
cody-littley: One nice to have property would be if this utility ensured that different checkpoint stores checkpoint the same block number. Not required for correctness of course, we're intentionally designing a system that doesn't break when checkpoints don't align (since a crash can cause us to be missing some checkpoints, even if we intend to checkpoint the same block). But its still nice to align them when we can.

If we wanted this property, we'd have to flip the model a little. Instead of the scheduler saying "it's time to checkpoint your current block", we'd have each of the stores ask the scheduler "should I checkpoint block X?". Via this mechanism we could ensure that we don't have an off-by-one block height difference when we snapshot our different stores.

cody-littley (approved):

sei-db/controller/checkpoint_scheduler.go
yzang2019: Agreed

sei-db/controller/checkpoint_scheduler.go
yzang2019: Changed to a poll model, but poll model has its own edge case and limitation as well.

sei-db/controller/checkpoint_scheduler.go
yzang2019: Checkpoint interval now allows both time based and block based setup.

yzang2019:
@seidroid review

sei-db/controller/checkpoint_scheduler.go
```diff
@@ -4,91 +4,151 @@
...
+//
+// Stores ask ShouldCheckpoint at every version they commit and call MarkCheckpointComplete once
+// that version's checkpoint has finished, whether or not it succeeded: every yes obliges a report.
+// A store joins the schedule by asking, and a height is held until every store registered when it
+// was picked, along with any that took it afterwards, has reported. A store behind the others is
+// handed that same height rather than one that has moved on, and heights arrive no faster than the
+// slowest store reaches them.
+//
+// A no is final and covers every height under it, so a version refused to one store is refused to
+// all of them; a yes holds for every store that reaches that height before it is replaced.
```
cody-littley: This sentence is confusing to me. What does it mean that "no" covers every height under it?

Consider the following example:

Store A gets "yes" for block 100.
Store A gets "no" for block 101.
Store B is slow, and is just now getting to block 100. Does it get a "yes" or a "no"?

sei-db/controller/checkpoint_scheduler.go
```diff
@@ -4,91 +4,151 @@
...
-type CheckpointScheduler interface {
-	SupportsCheckpoint() bool
-	ScheduleCheckpoint(destDir string, shouldRun func() bool, done func(error))
-	SetCheckpointVersion(destDir string, version int64) error
+// A time interval, a block interval, or both may be set. With both set a height has to clear both;
+// a value of 0 or less is unused, and with neither set checkpointing is off. Both are measured from
+// the last checkpoint, and from the scheduler's creation before there is one.
+//
+// Every store has to ask at each version it commits. One asking at only some of them never reaches
+// the height being held, which stops the node checkpointing rather than only that store.
```
cody-littley: In order for this to be safe, we need to make sure snapshotting is not enabled during replay, since not all stores will replay the same blocks. (We don't want snapshotting during replay anyways, just pointing out that this becomes a safety issue instead of a performance issue.)

sei-db/controller/checkpoint_scheduler.go
```diff
@@ -4,91 +4,151 @@
...
+type CheckpointScheduler struct {
+	mu     sync.Mutex
+	config config.CheckpointConfig
+
+	// registered holds every store that has asked, which is how a store joins the schedule.
+	registered map[string]struct{}
+	// awaiting holds the stores the current height is held for that have yet to report it. Empty
+	// means no checkpoint is outstanding.
+	awaiting map[string]struct{}
```
cody-littley: It isn't immediately obvious what "current height" means in context.

sei-db/controller/checkpoint_scheduler.go
cody-littley: Reading through the code, I think this scenario works the way it's intended, i.e. store B gets the correct "yes" answer for block 100. But this wasn't obvious from the godoc.

cody-littley (approved):

sei-db/controller/checkpoint_scheduler.go
```diff
@@ -4,91 +4,151 @@
...
+	blocksElapsed := s.config.BlockInterval <= 0 || version-s.nextCheckpointVersion >= s.config.BlockInterval
+	return timeElapsed && blocksElapsed
+}
+
+// pickCheckpointHeight makes version the height in hand, held for the stores registered now. One
+// registering later is not held for here, at a version it may already have passed, and is held for
+// from the moment it takes a height.
+func (s *CheckpointScheduler) pickCheckpointHeight(version int64) {
+	s.nextCheckpointVersion = version
+	s.awaiting = maps.Clone(s.registered)
```
blindchaser: it a registered store moves from H-1 to H+1, it never reports H. would it stay in awaiting forever?

blindchaser: another thing, it looks registered only grows. If a store stops participating, each future checkpoint still adds it to awaiting. The store cannot report completion, so checkpoint scheduling stops permanently. Can we add store deregistration?

sei-db/controller/checkpoint_scheduler.go
yzang2019: True, the comment is a bit confusing, worth a rewording.

sei-db/controller/checkpoint_scheduler.go
yzang2019: Hmm that's a good point

sei-db/controller/checkpoint_scheduler.go
yzang2019: We should never allow any store to move from H-1 to H+1, it is required for each store to commit on every block and check if it should checkpoint after the commit for every height

yzang2019: But I think it would be good to add a logic to bypass the awaiting anyway if the store somehow skip to the next height (e.g. batch commit multiple blocks in the future)

yzang2019: We should just assume the stores won't change, for example FlatKV and SS should always continuously do checkpoint, there's no need for any deregistration

sei-db/controller/checkpoint_scheduler.go
yzang2019: Fixed

blindchaser (approved):

cody-littley (approved):

yzang2019:
@seidroid review

PR #4028, opened 2026-08-27
Integrate Autobahn with in-memory EVM-only executor

codchen:

This PR builds on the EVM-only Giga store merged in #3864 and is rebased onto that merge commit.

- Add a test-only top-level `evm-only-in-memory` node setting that replaces the Cosmos ABCI application before `proxy.New`, so Autobahn and the executor share the same application.
- Initialize the EVM-only executor from `InitChain`, accept protected raw Ethereum transactions on chain ID 1337, and execute finalized blocks with a deterministic app hash.
- Add a minimal EVM JSON-RPC listener on port 8545 with only `eth_sendRawTransaction`. It validates Ethereum encoding, routes by Autobahn EVM shard ownership, submits through the in-process mempool, and returns the Ethereum transaction hash.
- Do not start Tendermint RPC for EVM-only nodes. The Docker harness also disables the unused Cosmos REST, gRPC, and gRPC-web servers.
- Keep executor and Autobahn consensus state in memory; Docker disables Autobahn persistence with `--persistent-state-dir=`.
- Add `make autobahn-evmonly-integration-test`, which submits 4,000 signed transfers through EVM JSON-RPC, asserts Tendermint RPC is unavailable, and uses internal Prometheus execution counters to wait for every validator.
- Add the `autobahn-e2e` cluster manager with `deploy`, `list`, `forward`, and `teardown` subcommands for local Docker or a managed EC2 host. `list` obtains height from internal Prometheus metrics and does not depend on Tendermint RPC.
- Keep the runtime deliberately ephemeral and load-test-only. Persistence, restart recovery, state sync, receipts, additional EVM RPC methods, and staking or other custom precompiles are out of scope.

- `make autobahn-evmonly-integration-test` (4,000 raw transfers through EVM JSON-RPC; all four validators executed through height 8; 5,010 tx/s observed locally; Tendermint RPC unavailable)
- Real `autobahn-e2e` local lifecycle: deploy four nodes, list live status/Prometheus heights, forward node 2 to `127.0.0.1:18545`, call the EVM RPC through the tunnel, confirm Tendermint `/status` resets, then teardown all containers and the network
- `go test -count=1 ./cmd/autobahn-e2e`
- `go test -count=1 -run "^TestEVMOnlyRPC" ./sei-tendermint/node`
- `go test -c -tags autobahn_integration ./integration_test/autobahn`
- `go vet ./sei-tendermint/node ./cmd/autobahn-e2e`
- `go vet -tags autobahn_integration ./integration_test/autobahn`
- golangci-lint v2.8.0 on the changed packages with `--build-tags autobahn_integration` (0 issues)
- golangci-lint v2.8.0 `fmt --diff` across the repository
- `sh -n docker/localnode/scripts/step4_config_override.sh docker/localnode/scripts/step5_start_sei.sh`
- `git diff --check`


docker/localnode/scripts/step4_config_override.sh
```diff
@@ -165,6 +166,12 @@ if [ "$AUTOBAHN" = "true" ]; then
 
   # Generate autobahn config using seid
   seid tendermint gen-autobahn-config $NODE_DIRS --output "$AUTOBAHN_CONFIG"
+  if [ "$AUTOBAHN_EVMONLY_IN_MEMORY" = "true" ]; then
+    AUTOBAHN_CONFIG_TMP="$AUTOBAHN_CONFIG.tmp"
+    jq '.evm_only_in_memory = true | del(.persistent_state_dir)' "$AUTOBAHN_CONFIG" > "$AUTOBAHN_CONFIG_TMP"
```
wen-coding: You can pass empty string to persistent_state_dir, which I think will disable persistence.

integration_test/autobahn/autobahn_test.go
```diff
@@ -200,6 +214,21 @@ func assertAutobahnEnabled(t *testing.T) {
 	}
 }
 
+func evmOnlyInMemoryEnabled() bool {
```
shemnon, line 222: I acknowledge that this is for integration testing, but are there facilities to auto-map env variables to the json config?  I'm only seeing JSON for AutobahnFileConfig.

sei-tendermint/node/setup.go
```diff
@@ -295,6 +295,14 @@ func buildGigaRouter(
 	if err != nil {
 		return nil, nil, err
 	}
+	if fc.EVMOnlyInMemory {
+		maxGas, err := genesisMaxGas(genDoc)
+		if err != nil {
+			return nil, nil, err
+		}
+		logger.Warn("Autobahn EVM-only in-memory execution enabled; state is ephemeral and unsafe for persistent networks")
+		app = p2p.NewEVMOnlyInMemoryProxy(config.AutobahnEVMOnlyInMemoryChainID, maxGas, genDoc.InitialHeight)
```
wen-coding: I think this only updates buildGigaRouter's local app. Maybe try to do it like MockApp, do it earlier in node.New? https://github.com/sei-protocol/sei-chain/pull/3905

integration_test/autobahn/autobahn_test.go
codchen: There is not an automatic Autobahn JSON env mapper here, so I removed this switch from `AutobahnFileConfig`. It is now the test-only top-level `evm-only-in-memory` TOML setting beside `mock-app`; the Docker env variable is only the harness selector and maps into that node config.

docker/localnode/scripts/step4_config_override.sh
codchen: Updated. The harness now passes `--persistent-state-dir=` to `gen-autobahn-config`, so the generator omits the persistent state directory without a follow-up `jq` deletion.

sei-tendermint/node/setup.go
codchen: Updated to the `MockApp` pattern. `prepareApplication` now replaces the app inside `node.New` before `proxy.New`, so both RPC and GigaRouter receive the same application. The EVM-only executor derives the initial height and gas limit later from `InitChain`.

docker/localnode/config/config.toml
```diff
@@ -17,6 +17,9 @@ moniker = "sei-node-0"
 #   - No priv_validator_key.json, priv_validator_state.json
 mode = "validator"
 
+# Test-only application replacement for Autobahn EVM load tests
+evm-only-in-memory = false
```
philipsu522, line 21: why not just use the "autobahn" flag?

cmd/autobahn-e2e/aws.go
```diff
@@ -0,0 +1,471 @@
+package main
```
philipsu522, line 1: would suggest getting this reviewed by platform cc @amir-deris @bdchatham

giga/evmonly/rpc/server.go
```diff
@@ -0,0 +1,135 @@
+// Package rpc serves the minimal JSON-RPC surface for the EVM-only executor.
```
philipsu522, line 1: is this rpc performant enough to serve 200k tps given it's single threaded?

sei-tendermint/config/autobahn.go
```diff
@@ -84,6 +84,9 @@ type AutobahnFileConfig struct {
 	BlockDB AutobahnBlockDBConfig `json:"block_db,omitzero"`
 }
 
+// AutobahnEVMOnlyInMemoryChainID is the chain ID of the test-only EVM executor.
+const AutobahnEVMOnlyInMemoryChainID uint64 = 1337
```
philipsu522: why do we use prod's evm chain id? seems dangerous - if it's straightforward would suggest changing this to something other than pacific-1's

sei-tendermint/internal/p2p/evmonly_inmemory_app.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func (a *evmOnlyInMemoryApplication) LastBlockHeight() int64 {
+	for state := range a.state.Lock() {
+		return state.committedHeight
+	}
+	panic("unreachable")
+}
+
+func (a *evmOnlyInMemoryApplication) CheckTx(_ context.Context, req *abci.RequestCheckTxV2) *abci.ResponseCheckTxV2 {
```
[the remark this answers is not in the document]
philipsu522: likely ok for tests, but worth adding a comment when/if we want to productionize this @codchen

sei-tendermint/internal/p2p/evmonly_inmemory_app.go
```diff
@@ -0,0 +1,312 @@
...
+	gigastore "github.com/sei-protocol/sei-chain/sei-db/state_db/giga"
+	abci "github.com/sei-protocol/sei-chain/sei-tendermint/abci/types"
+	"github.com/sei-protocol/sei-chain/sei-tendermint/libs/utils"
+)
+
+const evmOnlyInMemoryMinGasPrice = 1_000_000_000
+
+var evmOnlyInMemoryBaseBalance = new(big.Int).Lsh(big.NewInt(1), 200)
+
+type evmOnlyInMemoryApplication struct {
```
philipsu522, line 28: do we need to impl GetValidators? https://github.com/sei-protocol/sei-chain/blob/main/app/app.go#L2427

@wen-coding otherwise when we hit the last road (i.e. when we move to next epoch) we will have a nil val set?

wen-coding:
I mostly checked the consensus/execution API, looks good to me.

sei-tendermint/internal/p2p/evmonly_inmemory_app.go
wen-coding: we can submit this PR first, then I link GetValidators with the correct committee.

wen-coding (approved):

cmd/autobahn-e2e/aws.go
codchen: Requested reviews from both @amir-deris and @bdchatham.

sei-tendermint/internal/p2p/evmonly_inmemory_app.go
codchen: Fixed in 62df3f549. The EVM-only app now returns the validator keys from the same autobahn.json committee with unit voting power, matching BuildDataState, and returns defensive copies. Added coverage for config plumbing and snapshot isolation so the last-road transition does not stage an empty validator set.

docker/localnode/config/config.toml
codchen: The Autobahn flag selects the consensus/router path, but existing Autobahn deployments still run the normal Cosmos application. This separate test-only flag selects the in-memory EVM application and the minimal RPC surface; inferring it from Autobahn would silently replace the application for every Autobahn node. I kept the two switches separate for that reason.

sei-tendermint/config/autobahn.go
codchen: Pacific-1 uses EVM chain ID 1329. This test-only executor uses 1337, the conventional local development chain ID, so signatures from this cluster cannot be replayed on Pacific.

sei-tendermint/internal/p2p/evmonly_inmemory_app.go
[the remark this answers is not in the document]
codchen: Added an explicit TODO at the CheckTx admission boundary in 62df3f549: production use must close the admission/block-validity gap before this app accepts arbitrary traffic.

giga/evmonly/rpc/server.go
codchen: The server is not single-threaded: Go net/http dispatches requests concurrently, and the geth RPC handler runs on those request goroutines. We have now measured 20k submitted TPS end-to-end across four endpoints (p50 3 ms, p99 23-26 ms); all 623,240 transactions admitted during the run were executed by every validator at about 19.8k TPS. We have not demonstrated 200k through this RPC path, so I would not claim that yet. The current PR scope is the bare-minimum load-test integration; a 200k target needs an isolated RPC saturation profile plus removal of the extra shard-proxy HTTP hop and repeated sender recovery.

cmd/autobahn-e2e/aws.go
amir-deris: I was wondering is it better to have this infra provisioning in `Platform` repo instead? There seems to be potential security issues with this setup.  Perhaps with Harbour cluster, we can run the same workload with those security concerns eliminated.

In case this needs to stay in `sei-chain`, here are some recommendations to make it more secure:
```
1. Orphaned AWS resources — no expiry, and untracked-orphan windows

The issue. Teardown is driven entirely by IDs recorded in the local state file, and every create call happens before the save that records it. Three windows:

- aws.go:113 creates the security group; the first save is at aws.go:125. The error return at aws.go:121 is a bare return err, not fail(err), so it doesn't even mark the cluster failed.
- aws.go:128 tags the SG as a separate call after creation. Between 113 and 134 the SG exists with no tag and no state entry.
- aws.go:152 creates the key pair; the save recording ManagedKey is at aws.go:172. The key pair also gets no tags at all — it's the one resource with no discovery path of any kind.

A Ctrl-C, a network blip, or a laptop sleep in any of those windows leaves a resource that teardown cannot see and list will not show. Separately, nothing expires: a successful deploy that's simply forgotten runs a c7g.2xlarge and holds an internet-facing SSH endpoint indefinitely.

Mitigation — make teardown tag-driven, not ID-driven. State becomes a convenience, not the sole source of truth. Three parts:

Tag atomically at creation. Both create-security-group and create-key-pair accept --tag-specifications, so the resource is never observable in an untagged state. Drop the separate create-tags call at aws.go:128:

"ec2", "create-security-group",
"--group-name", securityGroupName,
"--vpc-id", vpcID,
"--tag-specifications", fmt.Sprintf(
    "ResourceType=security-group,Tags=[{Key=sei-autobahn-e2e-cluster,Value=%s},{Key=sei-autobahn-e2e-expires,Value=%s}]",
    options.name, expiresAt.Format(time.RFC3339)),

Same shape on create-key-pair. run-instances (aws.go:191) already tags atomically — just add the expiry tag there too.

Teardown sweeps by tag. After the ID-driven pass, describe-instances / describe-security-groups / describe-key-pairs filtered on Name=tag:sei-autobahn-e2e-cluster,Values=<name> and delete anything found. This closes all three windows at once, because it doesn't care what state recorded. It also gives you a teardown --orphans mode that sweeps every cluster tag in the region — the recovery tool for the crash cases that already happened.

Expiry is a tag, plus something that reads it. The tag alone documents intent; it doesn't reclaim anything. Pair it with either a scheduled Lambda in the account (filter on the tag, terminate past-due) or, cheaper to land now, have list render an EXPIRED status and print the teardown command — that at least makes a forgotten cluster visible the next time the operator touches the tool.

While you're in there: change the bare return err at aws.go:121 (and 125, 172, 203, 225, 241) to fail(err) so a failed deploy is consistently marked, and pass --client-token to run-instances so a retry after a timeout doesn't launch a second instance.

---

2. Unverified Go toolchain download in user-data

The issue. aws.go:391 fetches the Go tarball over TLS and untars it with no integrity check:

curl -fsSL "https://go.dev/dl/go1.25.6.linux-${go_arch}.tar.gz" -o /tmp/go.tgz
tar -C /usr/local -xzf /tmp/go.tgz

The apt-get install above it is not in the same category — apt verifies Debian repo signatures, so those packages are covered. This tarball is the single unverified artifact in the bootstrap, and it's extracted to /usr/local and then used to build the node binary. TLS authenticates the host; it says nothing about whether the bytes you got are the bytes Go published.

Mitigation — pin the SHA256 per architecture. Go publishes a checksum for every tarball on the download page. Both arm64 and amd64 a
    goSHA256AMD64   = "…"
)

and in the script:

case "$(uname -m)" in
  aarch64|arm64) go_arch=arm64; go_sha=%[2]s ;;
  x86_64|amd64)  go_arch=amd64; go_sha=%[3]s ;;
  *) echo "unsupported architecture: $(uname -m)" >&2; exit 1 ;;
esac
curl -fsSL "https://go.dev/dl/go%[1]s.linux-${go_arch}.tar.gz" -o /tmp/go.tgz
echo "${go_sha}  /tmp/go.tgz" | sha256sum -c -

set -euo pipefail is already at the top, so a failed sha256sum -c aborts before the tar and the readiness marker is never written — the deploy fails closed at waitForEC2Bootstrap rather than silently running on a tampered toolchain. The maintenance cost is real but small: the constants change only when you bump the Go version, and they sit next to the version string so the bump is one edit.

---

3. --ssh-cidr accepts 0.0.0.0/0

The issue. resolveSSHCIDR (aws.go:327) validates syntax only:

if _, _, err := net.ParseCIDR(configured); err != nil {
    return "", fmt.Errorf("invalid --ssh-cidr: %w", err)
}
return configured, nil

0.0.0.0/0 and ::/0 parse fine and go straight into authorize-security-group-ingress. The default path is good — a /32 from checkip.amazonaws.com — so this only bites when someone passes the flag, but that's exactly when it bites: the natural reason to reach for --ssh-cidr is that IP detection gave the wrong answer behind a VPN or NAT, and 0.0.0.0/0 is the first thing a frustrated operator tries. The result is an internet-open SSH port that no other guard catches, and it persists for the life of the cluster (which per item 1 may be indefinite).

Mitigation — refuse /0, warn on anything broader than /24. The check belongs in resolveSSHCIDR itself, which is the single function every path passes through, rather than at the call site:

ones, bits := network.Mask.Size()
if ones == 0 {
    return "", fmt.Errorf("--ssh-cidr %q opens SSH to the internet; pass a specific range", configured)
}
if (bits == 32 && ones < 24) || (bits == 128 && ones < 120) {
    // warn to stderr, proceed
}

Hard-refuse only /0 — a corporate /16 is a legitimate answer for someone on a VPN, so that case gets a warning, not a block. Note net.ParseCIDR already returns the *net.IPNet you need; the current code discards it into _.

---

4. SSH client args: known_hosts pollution and IP reuse

The issue. sshBaseArgs (aws.go:458) sets StrictHostKeyChecking=accept-new but no UserKnownHostsFile. Two consequences, and the operational one is worse than the security one:

- Every ephemeral instance's host key is appended to the operator's real ~/.ssh/known_hosts and never removed at teardown.
- EC2 recycles public IPs. When a later cluster lands on an IP a previous cluster used, accept-new doesn't accept — it sees a mismatch against the stale entry and hard-fails. Because waitForEC2Bootstrap (aws.go:408) polls SSH and treats every error as "not ready yet", the operator gets a 20-minute silent hang ending in a timeout, with the real cause buried.

Also missing: IdentitiesOnly=yes. Without it, ssh offers every key in the operator's agent to the box before falling back to -i.

Mitigation — scope the known-hosts file to the tool's state dir. One file per cluster, removed at teardown along with the key:

"-o", "UserKnownHostsFile=" + filepath.Join(stateDir, state.Name+".known_hosts"),
"-o", "IdentitiesOnly=yes",

This keeps first-connect trust-on-first-use (unchanged security posture — pinning genuinely needs ec2 get-console-output fingerprint extraction or SSM Session Manager, both larger changes), but it eliminates the stale-entry failure entirely because the file starts empty for every new cluster, and it stops the tool writing to the operator's personal SSH config. sshBaseArgs takes only clusterState today, so it needs the state dir threaded in — either widen the signature or store the path on awsState at deploy time.

---
```

amir-deris (approved):

PR #4030, opened 2026-08-27
Bound block reference parsing depth

masih:
Limit nested EIP-1898 block references to 16 levels to prevent quadratic parsing time on deeply nested inputs. Add boundary coverage.

Fixes PLT-1083

### DIFF
--- a/cmd/frozen-rpc-router/router.go
+++ b/cmd/frozen-rpc-router/router.go
@@ -22,7 +22,6 @@ const (
 	jsonRPCUnsupportedError = -32000
 	jsonRPCUpstreamError    = -32001
 	rpcRouteHeader          = "Sei-RPC-Route"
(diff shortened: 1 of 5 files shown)


cmd/frozen-rpc-router/router.go
```diff
@@ -22,6 +22,7 @@ const (
 	jsonRPCUnsupportedError = -32000
 	jsonRPCUpstreamError    = -32001
 	rpcRouteHeader          = "Sei-RPC-Route"
+	maxBlockReferenceDepth  = 16
```
codchen: can we make this a flag that can be passed in? just in case we'd need more depth for some blocks

masih:
@seidroid review

codchen (approved):

sei-will (approved):

PR #4036, opened 2026-08-27
Validate snapshots before publication

masih:
Reject malformed node hashes before writing snapshot records and validate
completed snapshots before updating the current symlink. Add regression
coverage for partial leaf records and corrupt snapshot publication.


sei-db/state_db/sc/memiavl/db.go
```diff
@@ -829,26 +831,39 @@ func (db *DB) RewriteSnapshot(ctx context.Context) error {
...
+		return fmt.Errorf("rename snapshot directory to %q: %w", targetPath, err)
+	}
+
+	return updateCurrentSymlink(db.dir, snapshotDir)
+}
+
+func (db *DB) validateSnapshot(ctx context.Context, path string) error {
+	opts := db.opts
+	opts.SnapshotPrefetchThreshold = 0
+	mtree, err := LoadMultiTree(ctx, path, opts)
```
yzang2019, line 890: LoadMultiTree leaks mmaps and fds on partial failure. multitree.go:103 returns nil, err without closing the snapshots it already opened for earlier stores. This is a pre-existing issue, but this PR makes it newly reachable: every validation failure now leaks fds

sei-db/state_db/sc/memiavl/db.go
```diff
@@ -488,18 +488,15 @@ func (db *DB) checkBackgroundSnapshotRewrite() error {
 		db.snapshotRewriteCancelFunc = nil
 
 		if !ok {
-			// channel was closed without sending a result
-			// Still prune old snapshots to prevent accumulation
-			go db.pruneSnapshots()
```
yzang2019, line 493: I think we added that because there were edge causes causing snapshot not being pruned properly and getting accumulated, causing disk fililng up

yzang2019: In the latest main, I think we already made the change so that we will make sure to keep at least 2 memIAVL snapshots

masih:
@seidroid review

masih:
@seidroid review

yzang2019 (approved):

masih:
@seidroid review

PR #4039, opened 2026-08-27
move pebble checkpoints off of the execution goroutine

cody-littley:

Moves pebble checkpointing (what we call snapshots) off of the transaction execution goroutine


sei-db/db_engine/view/view_manager_config.go
```diff
@@ -74,7 +74,7 @@ func DefaultViewManagerConfig(name string, reservedPrefix string) *ViewManagerCo
 		Name:                         name,
 		MetricsEnabled:               true,
 		MetricsScrapeIntervalSeconds: 10,
-		MaxUnflushedVersions:         4,
+		MaxUnflushedVersions:         1024,
```
blindchaser: any OOM risk here to MaxUnflushedVersions to 1024?

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -114,13 +114,12 @@ func (s *CommitStore) Commit(version int64) (committed int64, err error) {
...
-	// flush wait inside WriteSnapshot is where a dead store surfaces, and a block whose data will never
-	// reach disk must not be reported as committed. The block is already durable in the WAL, so replay
-	// reconciles whatever the caller's halt leaves behind.
-	if s.config.SnapshotInterval > 0 && version%int64(s.config.SnapshotInterval) == 0 {
-		s.phaseTimer.SetPhase("commit_write_snapshot")
-		if err := s.WriteSnapshot(""); err != nil {
+	// Step 5: Offer the block to the snapshot writer, which decides whether it becomes a snapshot and,
+	// if so, writes it on its own goroutine. Periodic snapshots are what keep the WAL bounded and
+	// restarts fast.
+	if s.snapshotWriter != nil {
```
blindchaser, line 120: Offer now reports the writer's latched failure, so the error names the current height, not the height whose snapshot failed. The halt is also one block late. Both are acceptable if nothing downstream reads that height for recovery. Please confirm, and consider naming the failed height in the latched error so the log points at the real block.

sei-db/state_db/sc/flatkv/snapshot_writer.go
```diff
@@ -0,0 +1,360 @@
...
+	// Whatever is still queued is owed a hand-back, so nothing is left holding a reservation that would
+	// stall its database for good.
+	defer w.discardQueued()
+
+	for {
+		// Charged for however long the queue stays empty, so a writer that is never idle is one the
+		// cadence is outrunning.
+		w.phaseTimer.SetPhase("idle")
+		select {
+		case <-w.ctx.Done():
```
blindchaser, line 266: select randomly chooses among ready cases, so shutdown competes with the queue instead of taking priority. Once stop() is called, w.ctx.Done() remains ready, causing races on both sides of the channel.

Here, a queued message may win, causing Close to wait for another snapshot and breaking its guarantee that queued work is discarded.

In enqueue at L166, a send may win if the queue has room, so Offer returns nil even though the writer may exit without processing it. Commit then reports success for a snapshot that may never be written.

Both sides need to be fixed: one prevents extra work during shutdown; the other prevents false success after shutdown.

sei-db/state_db/sc/flatkv/config/config.go
```diff
@@ -22,27 +17,32 @@ type Config struct {
...
 	SnapshotKeepRecent uint32 `mapstructure:"snapshot-keep-recent"`
 
+	// MaxSnapshotLagBlocks is how many committed blocks may queue up behind a snapshot that is still
+	// being written before Commit blocks. A value below 1 is treated as 1.
+	//
+	// A snapshot being written holds every database pinned at its own height, so no later block can
+	// reach disk until it completes, and each one is retained in memory meanwhile. This bounds how far
+	// that can run, trading a pause in block production for the memory the backlog would otherwise
+	// consume. It bounds blocks rather than bytes, so it mitigates exhaustion rather than preventing it.
+	MaxSnapshotLagBlocks uint32 `mapstructure:"max-snapshot-lag-blocks"`
```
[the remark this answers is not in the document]
blindchaser: +1

yzang2019 (approved):

app/testdata/state-commit.golden
```diff
@@ -28,7 +29,7 @@ FlatKVConfig.AccountStoreConfig.EstimatedOverheadPerEntry = uint64(256)
 FlatKVConfig.AccountStoreConfig.Name = string("account")
 FlatKVConfig.AccountStoreConfig.MetricsEnabled = bool(true)
 FlatKVConfig.AccountStoreConfig.MetricsScrapeIntervalSeconds = float64(10)
-FlatKVConfig.AccountStoreConfig.MaxUnflushedVersions = uint64(4)
+FlatKVConfig.AccountStoreConfig.MaxUnflushedVersions = uint64(1024)
```
yzang2019: This seems to be a pretty big jump? Why do we need that big of backlog?

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -114,13 +114,12 @@ func (s *CommitStore) Commit(version int64) (committed int64, err error) {
 	// Step 4: Clear per-block bookkeeping
 	s.clearPendingBlock()
 
-	// Periodic snapshot so WAL stays bounded and restarts are fast. A failure fails the commit: the
-	// flush wait inside WriteSnapshot is where a dead store surfaces, and a block whose data will never
-	// reach disk must not be reported as committed. The block is already durable in the WAL, so replay
-	// reconciles whatever the caller's halt leaves behind.
-	if s.config.SnapshotInterval > 0 && version%int64(s.config.SnapshotInterval) == 0 {
-		s.phaseTimer.SetPhase("commit_write_snapshot")
-		if err := s.WriteSnapshot(""); err != nil {
```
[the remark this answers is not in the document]
cody-littley: fixed, snapshot deletion now managed by snapshot writer thread

sei-db/state_db/sc/flatkv/snapshot_writer.go
```diff
@@ -0,0 +1,360 @@
...
+				request.responseChan <- struct{}{}
+			default:
+				err = fmt.Errorf("unknown snapshot writer message type %T", message)
+			}
+			if err != nil {
+				w.brick(err)
+				return
+			}
+		}
+	}
```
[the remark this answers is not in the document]
cody-littley: documentation issue, fixed

sei-db/state_db/sc/flatkv/config/config.go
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/db_engine/view/view_manager_config.go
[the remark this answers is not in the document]
cody-littley: Adjusted to 128. Must be large, based on benchmark results. Will not take performance tuning advice from an LLM here.

cody-littley: Changed this to 128. In benchmarking, I observed 1024 was safe. But probably not necessary to be quite as big as that.

sei-db/state_db/sc/flatkv/store_write.go
cody-littley: Error currently reports both the block that failed and the block where the error bubbles up. Nothing should be attempting to recover this, the block number is for debugging only.

sei-db/state_db/sc/flatkv/snapshot_writer.go
cody-littley: IMO, not a problem if shutdown loses the race. Even if shutdown was guaranteed to win in the select statement, we might receive a shutdown one nanosecond after we start a pebble checkpoint... so it's still possible that shutdown might be delayed by a pebble checkpoint.

Also mitigating this problem is that in production use cases, we don't really shut down a store so much as pull the rug out from under it... so I don't think a slow shutdown really hurts us much.

app/testdata/state-commit.golden
cody-littley: It was helpful during performance testing to absorb bursty behavior. 1024 is larger than it needs to be though, after I did several optimizations. Reduced to 128.

PR #4040, opened 2026-08-27
Add v6.7 retired module upgrade tests

alexander-sei:
## Summary
- Add build-tagged v6.7 upgrade coverage for the feegrant, capability, IBC, and transfer modules retired by v6.7, plus the deprecated oracle handlers.
- Add boundary discovery, scaffolding, and structural checks so the current v6.6 -> v6.7 test set is selected from `app/tags` instead of being hard-coded in CI.
- Exercise the v6.6 -> v6.7 boundary through two independent paths: a persisted two-process Go test and the existing live four-validator binary upgrade.
- Add a dedicated GitHub workflow for the offline Go path while keeping the live-node workflow separate.

## Cross-version coverage
- The offline path compiles its source phase on `release/v6.6`, commits an application database, then compiles its target phase on `release/v6.7` and reopens that same database.
- The v6.7 target applies the real upgrade handler, removes exactly the retired module-version entries, retains oracle, and proves the seeded feegrant, capability, IBC, and transfer stores are unchanged after another reopen.
- The live path builds one `seid` per ref, creates fixtures with v6.6, performs the governance halt and binary replacement across four validators, and runs the tagged v6.7 assertions against the same node homes.
- In-process tests cover retired transaction behavior, fee-granter rejection, retained store data, exported genesis, orphaned module versions, and the remaining oracle blocker behavior.

## Test plan
- [x] `go test ./app -count=1 -run 'Test(V67|LatestUpgradeLeavesNoOrphanedModuleVersions|MountedStoresAreOwnedOrExplicitlyRetained|RetainedStoresRemainMounted|RetiredOracle)'`
- [x] `go test ./app -count=1 -run '^TestOracleKeepsWritingStateAfterV67$'`
- [x] `go test ./testutil/processblock/... -count=1`
- [x] `go test ./upgradetest/...`
- [x] `make upgrade-test-vet`
- [x] `FROM_REF=release/v6.6 TO_REF=release/v6.7 bash .github/scripts/offline-upgrade-test.sh`
- [x] `make fmtcheck`


masih:
This is plenty good! ship it alex!

alexander-sei:
@seidroid review

.github/workflows/offline-upgrade-test.yml
```diff
@@ -0,0 +1,74 @@
+name: Offline Minor Upgrade Test
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
```
masih: There are other paths that can be touched _and_ result in breaking upgrade tests; I am curious why limit to this specific set?

I would just remove this condition entirely?

masih (approved):

.github/workflows/offline-upgrade-test.yml
alexander-sei: Agreed. Removed the `paths` filter in 443a6a763 so the persisted upgrade test runs for every PR to `main`.

PR #4042, opened 2026-08-27
Fix memiavl snapshot race condition

yzang2019:
## Summary

`Tree.Copy()` shares nodes with the live tree but did not freeze them. A reader of the copy and a writer on the live tree therefore share `MemNode` memory with no lock in common, and a torn read of `MemNode.hash` is enough to produce a leaves file that is 32 bytes short of a whole number of records. This PR makes `Copy()` establish the freeze before handing the tree over.

The production crash that motivated the work is:

```
corrupted snapshot, leaves file size 63201088 is not a multiple of 48
```

The byte pattern matches that torn-hash write. **This PR does not claim to have found the production trigger, and should not be treated as a confirmed fix for that incident.** On the affected node's config, every concrete path we could enumerate is closed; see "Scope" below.

## The problem

Snapshots are written in the background so they don't block the chain. To do that safely, `Tree.Copy()` hands the writer its own view of the tree. The two views share the same underlying nodes in memory, and the deal is: the copy is *frozen*, so the live chain will never touch a node the writer is still reading.

`Copy()` was not keeping that deal. Three separate reasons:

- **The freeze line was off by one version.** Writes stamp new nodes at `version + 1`, but `Copy()` set its protection floor at `version`, leaving any node written during the current block unprotected. `DB.Copy` / `CommitStore.Copy` reach this state: `ApplyChangeSets` releases `db.mtx` on return, so a copy taken before the following `Commit` is mid-version.
- **`SaveVersion()` updated the version with no lock held**, so a concurrent copy could read the new version and compute a floor that protected nothing.
- **Nodes weren't guaranteed to be hashed before being handed over**, so a reader would compute and cache hashes in place on nodes the live tree also held.

Each leaf is written as a 16-byte header followed by its 32-byte hash. `writeLeaf` sizes that copy with `len(hash)`. A torn read of the slice header — pointer set, length still zero — yields a zero-length hash, so the writer emits the header and nothing else. That is a leaves file with remainder 16, which is the observed `63201088 % 48 == 16`.

### Why one mutex per copy is the underlying hole

`RootHash` and the proof builders take the *write* lock because `MemNode.Hash` fills the hash cache in place — that was the fix for Immunefi 83246. But `Copy()` gives each tree its **own** mutex, so that lock only serializes hash fills within a single tree instance. Two trees sharing nodes serialize against nothing, which puts the same unsynchronized mutation right back. Freezing the nodes before handing them over is what actually closes it.

## Changes

- `Copy()` now takes the write lock, hashes everything reachable before handing the tree over, and sets the protection floor to the highest version actually present in the tree instead of the last saved version.
- `SaveVersion()` is now a single locked section, so the version bump and the hashing can't be observed apart.
- The snapshot writer reads the root and the version together, so a snapshot's metadata can't claim a version its contents don't match.
- **Separate bug, also fixed:** a copy inherited the live tree's background-write channel, so calling `Close()` on *any* copy shut down the live tree's background writer and made the next async write panic on a closed channel. Copies now get their own.

There is no behavior change outside the race — no existing test needed editing.

## Notes for reviewers

- **Performance:** the added hashing in `Copy()` is free on the rewrite path, since the commit already hashed everything and the call short-circuits. Only a mid-block copy does real work, and that's work the next `SaveVersion` would have done moments later anyway. The tighter freeze line avoids extra node copies rather than adding them.
- **Relationship to #4036:** complementary, not redundant. That PR catches a corrupt snapshot before publishing it; this one stops the known API-reachable way of producing one. Worth having both: the validator is the backstop if another path to this kind of mutation turns up, and it is the only thing that would have turned this incident into a failed rewrite instead of a published-and-unbootable snapshot.
- **Still open:** the production trigger. The mechanism (torn `node.hash` at the write site → remainder 16) is exact. A production path that writes `node.hash` on a node the rewrite is traversing has not been found on this node's config.

### DIFF
--- a/sei-db/state_db/sc/memiavl/tree.go
+++ b/sei-db/state_db/sc/memiavl/tree.go
@@ -30,18 +30,20 @@ type Tree struct {
 	// when true, the get and iterator methods could return a slice pointing to mmaped blob files.
 	zeroCopy bool
 
+	// unsavedWrites reports whether Set or Remove has stamped nodes at
+	// version+1 since the last SaveVersion, so maxNodeVersion knows whether the
+	// tree currently holds nodes above version.
+	unsavedWrites bool
+
 	// mtx guards concurrent access to this tree's mutable state (root, version,
 	// cowVersion, snapshot) AND the lazily-populated MemNode.hash caches reachable
 	// from root. Operations that fill those caches in place — RootHash and the
 	// proof builders (GetProof/GetMembership/GetNonMembership) — take the write
 	// lock; pure reads (Get/Has/Iterator) take the read lock. This serialization
 	// only protects a single tree instance: a tree produced by Copy() gets its
-	// own mtx and shares the underlying nodes copy-on-write. Cross-copy hash
-	// consistency therefore relies on (a) the shared nodes already being fully
-	// hashed before the copy is used for hashing/proofs (Copy is taken between
-	// commits, and the commit path hashes via SaveVersion(true)/RootHash), and
-	// (b) cowVersion cloning any shared MemNode before it is structurally mutated,
-	// so a live-tree write never mutates a node another copy is still reading.
+	// own mtx and shares the underlying nodes copy-on-write. Copy is what makes
+	// that safe across instances, by freezing the shared nodes before handing
+	// them over; see its doc comment.
 	mtx *sync.RWMutex
 
 	pendingChanges chan proto.ChangeSet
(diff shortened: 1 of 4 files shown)


yzang2019:
@seidriod review

sei-db/state_db/sc/memiavl/tree.go
```diff
@@ -108,24 +110,62 @@ func (t *Tree) SetInitialVersion(initialVersion int64) error {
...
+	return t.copyNoLock()
+}
+
+// copyNoLock is Copy without acquiring t.mtx; the caller must already hold the
+// write lock.
+func (t *Tree) copyNoLock() *Tree {
+	// Hash every reachable node now, while we hold the write lock. Otherwise the
+	// copy's reader fills MemNode.hash in place on nodes the live tree also
+	// holds, and the two trees have separate mutexes to serialize with.
+	_ = t.rootHashNoLock()
```
blindchaser, line 133: could this increase EndBlock latency? Trace snapshots call Copy() before commit, so this now hashes all dirty nodes while holding the write lock

yzang2019: Not a concern. Trace capture runs in EVM EndBlock, and that is before the current block is flushed into memiavl. So SnapshotSCStore → Copy() is hashing last-committed trees, not this block’s dirty nodes. rootHashNoLock() therefore stops at the root: one nil-check per tree, not a walk. MultiTree.Copy does that for each of the ~20 stores, under that tree’s write lock. The added EndBlock cost is lock acquire + an O(1) hash hit + the same shallow struct copy as before.

yzang2019:
@seidroid review

masih (approved):

masih:
I will land the corruption check PR after this one lands, thanks @yzang2019

PR #4046, opened 2026-08-27
Derive Autobahn epoch committees from execution stake (CON-358)

wen-coding:
## Summary
- Stage `C_{E+2}` from the bonded validator set after `Commit` at LastRoad(E); keep it pending until a matching AppQC (or eviction past that road) publishes it.
- Persist the live/pending registry as an A/B snapshot. Genesis epochs 0/1 are constructor-only and are never written.
- Consensus no longer seeds later epochs from genesis placeholders (`SetupInitialEpochs` / `AdvanceIfNeeded`).
- AppQC is only persisted if it matches local AppHash

### DIFF
--- a/sei-tendermint/autobahn/types/committee_test.go
+++ b/sei-tendermint/autobahn/types/committee_test.go
@@ -37,15 +37,31 @@ func TestNewCommittee_FiltersOutZeroWeightValidators(t *testing.T) {
 	}
 }
 
-func TestCommitteeEqual_IgnoresZeroWeightsAndMapOrder(t *testing.T) {
+func TestCommitteeEqual_IgnoresMapOrder(t *testing.T) {
 	rng := utils.TestRng()
 	a := GenPublicKey(rng)
 	b := GenPublicKey(rng)
 	c1 := utils.OrPanic1(NewCommittee(map[PublicKey]uint64{a: 3, b: 1}))
-	c2 := utils.OrPanic1(NewCommittee(map[PublicKey]uint64{b: 1, a: 3, GenPublicKey(rng): 0}))
+	c2 := utils.OrPanic1(NewCommittee(map[PublicKey]uint64{b: 1, a: 3}))
 	require.True(t, c1.Equal(c2))
-	c3 := utils.OrPanic1(NewCommittee(map[PublicKey]uint64{a: 3, b: 2}))
-	require.False(t, c1.Equal(c3))
+}
+
+func TestCommitteeEqual_IgnoresZeroWeights(t *testing.T) {
+	rng := utils.TestRng()
+	a := GenPublicKey(rng)
+	b := GenPublicKey(rng)
+	c1 := utils.OrPanic1(NewCommittee(map[PublicKey]uint64{a: 3}))
+	c2 := utils.OrPanic1(NewCommittee(map[PublicKey]uint64{a: 3, b: 0}))
+	require.True(t, c1.Equal(c2))
+}
+
+func TestCommitteeEqual_RejectsDifferentWeights(t *testing.T) {
+	rng := utils.TestRng()
+	a := GenPublicKey(rng)
+	b := GenPublicKey(rng)
+	c1 := utils.OrPanic1(NewCommittee(map[PublicKey]uint64{a: 3, b: 1}))
+	c2 := utils.OrPanic1(NewCommittee(map[PublicKey]uint64{a: 3, b: 2}))
+	require.False(t, c1.Equal(c2))
 }
 
 func TestNewCommittee_RejectsZeroTotalWeight(t *testing.T) {
@@ -329,12 +345,18 @@ func TestCommitteeConv_PreservesJoined(t *testing.T) {
 	rng := utils.TestRng()
 	a := GenPublicKey(rng)
 	b := GenPublicKey(rng)
-	src := utils.OrPanic1(NewCommittee(map[PublicKey]uint64{a: 1, b: 1}))
-	src = utils.OrPanic1(src.DeriveNext(map[PublicKey]uint64{a: 2, b: 3}, 4))
+	c := GenPublicKey(rng)
+	d := GenPublicKey(rng)
+	genesis := utils.OrPanic1(NewCommittee(map[PublicKey]uint64{a: 1, c: 3}))
+	src := utils.OrPanic1(genesis.DeriveNext(map[PublicKey]uint64{a: 2, b: 4, c: 3}, 4))
 	got, err := CommitteeConv.Decode(CommitteeConv.Encode(src))
 	require.NoError(t, err)
 	require.Equal(t, src.Lane(a).OrPanic("a"), got.Lane(a).OrPanic("a"))
 	require.Equal(t, src.Lane(b).OrPanic("b"), got.Lane(b).OrPanic("b"))
+	require.Equal(t, src.Lane(c).OrPanic("c"), got.Lane(c).OrPanic("c"))
 	require.Equal(t, uint64(2), got.Weight(a))
-	require.Equal(t, uint64(3), got.Weight(b))
+	require.Equal(t, uint64(4), got.Weight(b))
+	require.Equal(t, uint64(3), got.Weight(c))
+	require.Equal(t, EpochIndex(4), got.Lane(b).OrPanic("b").Joined)
+	require.False(t, got.HasReplica(d))
 }
--- a/sei-tendermint/autobahn/types/committee_test.go
+++ b/sei-tendermint/autobahn/types/committee_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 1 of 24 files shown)


sei-tendermint/internal/autobahn/data/state.go
```diff
@@ -321,6 +336,30 @@ func (s *State) First() types.GlobalBlockNumber {
...
+	}
+	if err := s.cfg.Registry.ActivateEpoch(idx); err != nil {
+		return fmt.Errorf("activate epoch %d: %w", idx, err)
+	}
+	for inner := range s.inner.Lock() {
+		inner.publishNextCommitEpoch(s.cfg.Registry)
+		return nil
+	}
+	panic("unreachable")
+}
```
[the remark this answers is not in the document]
wen-coding: Not a reachable path. Consensus cannot get more than blocksCacheSize (4000) blocks ahead of the AppQC-certified frontier (PushQC waits on first+blocksCacheSize, and first only advances after a matching AppQC is persisted). An epoch is 108,000 roads, so by the time the tip is at LastRoad(E), AppQC(LastRoad(E-1)) is already far behind first and C_{E+1} is live. The pending committee is C_{E+2}; restore’s EpochAt(tip+1) asks for C_{E+1}.

StageEpoch at LastRoad(E) has the same bound: it needs E+1 live, which the leash already guarantees. One pending slot is enough. We will not queue extra pending committees.

sei-tendermint/autobahn/types/committee_test.go
```diff
@@ -36,6 +37,17 @@ func TestNewCommittee_FiltersOutZeroWeightValidators(t *testing.T) {
 	}
 }
 
+func TestCommitteeEqual_IgnoresZeroWeightsAndMapOrder(t *testing.T) {
+	rng := utils.TestRng()
+	a := GenPublicKey(rng)
+	b := GenPublicKey(rng)
+	c1 := utils.OrPanic1(NewCommittee(map[PublicKey]uint64{a: 3, b: 1}))
+	c2 := utils.OrPanic1(NewCommittee(map[PublicKey]uint64{b: 1, a: 3, GenPublicKey(rng): 0}))
+	require.True(t, c1.Equal(c2))
```
shemnon, line 46: These are two relatively disjoint tests.  Shoudln't they be two different tests?

sei-tendermint/autobahn/types/committee_test.go
```diff
@@ -291,3 +303,38 @@ func TestDeriveNext_StayLeaveRejoin(t *testing.T) {
...
+	}
+	_, err := CommitteeConv.Decode(&pb.Committee{Members: []*pb.EpochMember{member, member}})
+	require.Error(t, err)
+}
+
+func TestCommitteeConv_PreservesJoined(t *testing.T) {
+	rng := utils.TestRng()
+	a := GenPublicKey(rng)
+	b := GenPublicKey(rng)
+	src := utils.OrPanic1(NewCommittee(map[PublicKey]uint64{a: 1, b: 1}))
```
shemnon: Can we add a c case that is unchanged? as well as a D case that is never set?, and make the be case not set in src but is added by derive?

sei-tendermint/internal/autobahn/data/state.go
```diff
@@ -136,6 +140,23 @@ func (i *inner) insertAppQC(appQC *types.AppQC) error {
...
+	if i.first < i.nextQC && i.qcs[i.first].qc.QC().Proposal().Index() > road {
+		return true
+	}
+	for n := i.first; n < i.nextQC; {
+		p := i.qcs[n].qc.QC().Proposal()
+		if p.Index() == road {
+			return n < i.nextAppQC &&
+				bytes.Equal(i.appProposals[n].AppHash(), i.appQCs[n].Proposal().AppHash())
+		}
+		n = max(n+1, p.GlobalRange().Next)
```
shemnon: is there ever a case where p.GlobalRange().Next is less than n+1?  Is it only populated for gaps?

sei-tendermint/autobahn/types/committee_test.go
```diff
@@ -36,6 +37,17 @@ func TestNewCommittee_FiltersOutZeroWeightValidators(t *testing.T) {
 	}
 }
 
+func TestCommitteeEqual_IgnoresZeroWeightsAndMapOrder(t *testing.T) {
```
philipsu522: nit - test is a bit misleading i don't think we're testing ignore zero weights, and zero weights, maybe we should rename remove "ZeroWeights"

sei-tendermint/internal/autobahn/epoch/registry.go
```diff
@@ -130,129 +199,105 @@ func (r *Registry) EpochAt(roadIndex types.RoadIndex) (*types.Epoch, error) {
...
-) (*types.Epoch, error) {
-	for s, ctrl := range r.state.Lock() {
-		if s.dropped(parent) {
-			return nil, fmt.Errorf("epoch %d: %w", parent, types.ErrPruned)
+// StageEpoch derives C_{endEpoch+2} from weights and persists it as pending.
+// A no-op if that epoch is already staged or live with the same committee.
+// An error if it conflicts, if the target is not live.Next, or if endEpoch+1
+// is not live.
+func (r *Registry) StageEpoch(endEpoch types.EpochIndex, weights map[types.PublicKey]uint64) error {
+	target := endEpoch + 2
```
philipsu522, line 229: the reason to do epoch C{e+2} is because C{e+1} cannot wait for epoch e to finish right (given async execution). open question - but do we have enough guarantees that we will **always** finish execution epoch e by the time we need to calculate C{e+2}?

philipsu522: i.e. - do we have backpressure in place?

codchen (approved):

shemnon (approved):
I wish github would have allowed file ordering, so I could start with the committee protobufs instead of end with them.

LGTM, but I'm the new hire who hasn't been indoctrinated into the ways of authbahn;

sei-tendermint/autobahn/types/committee_test.go
wen-coding: done

sei-tendermint/autobahn/types/committee_test.go
wen-coding: done

sei-tendermint/internal/autobahn/data/state.go
wen-coding: That's a good point, actually this shouldn't happen, changed.

sei-tendermint/internal/autobahn/data/state.go
```diff
@@ -867,7 +913,11 @@ func (s *State) runPersist(ctx context.Context) error {
 				status.NextAppProposal = appProposal.GlobalRange().Next
 			}
 			for status.NextAppQC < inner.nextAppQC {
-				appQC := inner.appQCs[status.NextAppQC]
+				n := status.NextAppQC
+				appQC := inner.appQCs[n]
+				if got, want := inner.appProposals[n].AppHash(), appQC.Proposal().AppHash(); !bytes.Equal(got, want) {
```
philipsu522, line 918: just making sure that there's no edge case where we might compare `bytes.Equal.(nil, []byte{})`? i think this isn't possible but want to be extremely sure

sei-tendermint/autobahn/types/committee_test.go
wen-coding: done, it's IgnoresZeroWeights now.

sei-tendermint/internal/autobahn/epoch/registry.go
wen-coding: Yes. Everyone who enters epoch e+1 must already agree on C_{e+1}. Execution of `LastRoad(e)` can still be in flight, so that committee cannot be “stake at end of e”. One epoch of gap: C_{e+2} = Committee(end(e)), used at FirstRoad(e+2).

The cluster backpressure is in avail/: validators do not enter C_{e+1} unless that committee is available (`WaitForEpoch` on activate). Full nodes only run data/ and are observers; they rely on validator backpressure. If a full node sees a QC/block for the next epoch without that committee, PushQC hard-fails (`EpochByIndex`). New joiners are expected to catch up via data sync and snapshot, within one epoch; otherwise they cannot function.

sei-tendermint/internal/autobahn/data/state.go
wen-coding: bytes.Equal.(nil, []byte{}) is true, but inner.appProposals[n].AppHash() is internal data structure, so it should never be nil.

Per style guide, internal data should never be nil, we use Option if the data can be missing, we don't test for nil.

Hmm, but I don't think we currently reject empty AppHash on AppProposal even though the field is marked "//required". I'll do an audit in another PR and fix it.

wen-coding: Check added in https://github.com/sei-protocol/sei-chain/pull/4058

PR #4058, opened 2026-08-28
(Autobahn) Reject missing AppProposal.app_hash on proto decode

wen-coding:
## Summary
- In Autobahn, `AppProposal.app_hash` is proto `// required` but Decode accepted nil/empty hashes.
- Reject `len(app_hash) == 0` at the converter (same choke point as the other AppProposal scalars).
- Cover missing required fields, including nil and empty `app_hash`, in `TestAppProposalConvDecode_RequiredFields`.


shemnon (approved):
LGTM

philipsu522 (approved):

PR #4061, opened 2026-08-31
StateView and StateDB impl

cody-littley:

Implement StateDB APIs with flatKV, using the new view infrastructure now available inside FlatKV.


sei-db/state_db/sc/flatkv/snapshot.go
```diff
@@ -427,31 +434,30 @@ func (s *CommitStore) outOfBandSnapshot() (err error) {
...
 func checkpointDatabases(
 	ctx context.Context,
 	dir string,
-	version int64,
-	views map[string]view.View,
+	blockView *storeView,
 	dbs map[string]types.Checkpointable,
 	phaseTimer *metrics.PhaseTimer,
 ) (_ string, err error) {
+	version := blockView.blockHeight
```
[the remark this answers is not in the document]
cody-littley: fixed

sei-db/state_db/sc/flatkv/state_view.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+// GetCodeSize returns the length of addr's contract code in bytes, or 0 when it has none.
+func (v *flatKVStateView) GetCodeSize(addr giga.Address) int {
+	return len(v.GetCode(addr))
+}
+
+// accountData returns the account row for the 20-byte address in keyBytes, or nil when no account
+// exists in this block.
+func (v *flatKVStateView) accountData(keyBytes []byte) *vtype.AccountData {
```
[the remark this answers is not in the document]
cody-littley: you are wrong

sei-db/state_db/sc/flatkv/store_view.go
```diff
@@ -0,0 +1,99 @@
...
+		if err := dbView.Reserve(); err != nil {
+			return fmt.Errorf("reserve %s view at height %d: %w", dbView.Name(), sv.blockHeight, err)
+		}
+	}
+	return nil
+}
+
+// release() hands back one reservation on every store's view. A failure stops there, for the same
+// reason reserve() does.
+func (sv *storeView) release() error {
```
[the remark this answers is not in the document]
cody-littley: errors are fatal, leaked reservations don't matter in a fatal system, will not fix

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -159,48 +165,45 @@ func (s *CommitStore) clearPendingBlock() {
...
-		if err := s.finalizeStore(sealed, version, alreadyHave); err != nil {
-			return fmt.Errorf("finalize %s: %w", sealed.Name(), err)
+	for _, dbView := range blockView.viewSlice {
+		if err := s.finalizeStore(dbView, version, alreadyHave); err != nil {
+			// Error is fatal; leaking reservations doesn't make it worse.
+			return fmt.Errorf("finalize %s: %w", dbView.Name(), err)
 		}
 	}
 
+	if err := s.lastSealed.set(blockView); err != nil {
```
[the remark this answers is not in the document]
cody-littley: you are wrong, will not change code

sei-db/state_db/gigadb/state_db_impl.go
```diff
@@ -0,0 +1,68 @@
+package gigadb
```
yzang2019: I think we can probably just move these impl to giga package?

cody-littley: This causes a circular dependency. `FlatKV` imports an interface from the `giga` package, and the implementation needs to import `FlatKV`.

We could solve this by promoting FlatKV's API to the `giga` package. Let me know if you'd prefer that.

sei-db/state_db/giga/state_db_impl.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// No need to flush WAL, since this WAL isn't used for crash recoverability safety (that's the BlockDB's job).
+	if err := s.wal.Write(uint64(blockNum), changeset); err != nil {
+		return fmt.Errorf("write block %d to state WAL: %w", blockNum, err)
+	}
+	if err := s.wal.SignalEndOfBlock(); err != nil {
+		return fmt.Errorf("end block %d in state WAL: %w", blockNum, err)
+	}
+
+	if err := s.liveStateDB.CommitStateChanges(blockNum, changeset); err != nil {
```
yzang2019, line 50: Is there any issue if the node crash in between?

sei-db/state_db/giga/state_view.go
```diff
@@ -1,38 +1,32 @@
 package giga
 
-const (
-	AddressLen = 20
-	HashLen    = 32
-)
+import ethcommon "github.com/ethereum/go-ethereum/common"
```
yzang2019: Not sure if we really want to use the geth types here, in case we want to switch to different EVM engine in the future, here I think defining our own types is better to avoid having any dependency on evm

sei-db/state_db/gigadb/state_db_impl.go
yzang2019: I see, in that case maybe we should just move the giga interfaces from `giga` to `types` package?

yzang2019 (approved):

sei-db/state_db/gigadb/state_db_impl.go
cody-littley: Change made. I moved the `flatkv.Store` interface into the giga package. It's now called `giga.LiveStateStore`, since just `Store` doesn't make sense in context.

We might also want to move the interface for SS into this package to. Perhaps call it `giga.HistoricalStateStore`? (Outside of scope for this PR, just thinking ahead.)

sei-db/state_db/giga/state_db_impl.go
cody-littley: No problems if we crash in between, BlockDB is ultimately responsible for crash recoverability. We've designed the system to tolerate both when the StateWAL and FlatKV is ahead of the other. Both the WAL and the StateDB have async commits anyways, so a crash after could produce the same interleaving as a crash in between.

sei-db/state_db/giga/state_view.go
cody-littley: Reverted back to the non-geth types for now. I'd like to discuss this with execution team though, I think we should have a unified type for both storage and execution (i.e. we shouldn't have to cast at the boundary between the two).

sei-db/state_db/gigadb/state_db_impl.go
yzang2019: Make sense to me

PR #4062, opened 2026-08-31
Add giga storage manager

yzang2019:
Couple of changes included in this PR:
1. Add GigaStorageManager as the main bootstrap constructor
2. Wire up existing stores with CheckpointScheduler and Prunnable interface
3. Move BlockStore from internal tendermint to public
4. Add enable config for receiptstore
CrashRecovery logic is todo in next PR


sei-db/bootstrap/recovery.go
```diff
@@ -0,0 +1,38 @@
...
+	db            types.Rollbackable
+	targetVersion int64
+}
+
+// CrashRecover brings the opened stores onto one height after an unclean shutdown. It assumes the
+// BlockDB is always ahead of other state DBs and will make sure all other DBs are reconciled to the same height.
+//
+// NewGigaStorageManager calls it once every store is open and before either the checkpoint schedule
+// or the prune cycle starts. It is a no-op on a node that shut down cleanly, and idempotent, so a run
+// interrupted partway is completed by the next open.
```
cody-littley: Specifically, we want to ensure that after this method is called, the following is true:

- All stores must be at or below the BlockDB's height
- All stores other than block DB must be on the same height
    - it's also ok if BlockDB is at the same height, but also ok if BlockDB is ahead

The way this is written, it seems like we also promise to bring other stores up to the BlockDB height. That wouldn't be "wrong". But if we are writing execution code that knows how to catch up from a block DB that's ahead, then we don't need that replay in here.

sei-db/bootstrap/storage_manager.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+	return m, nil
+}
+
+// start runs the steps that bring storage up, in the only order they permit.
+//
+// Validation comes first so a config no store will accept costs nothing to reject. Recovery comes
+// after every store is open, because it is a comparison across all of them, and
+// before either cycle starts, because both read heights that recovery moves: a prune cycle overlapping
+// a rollback derives its cut line from heights that are still falling.
```
cody-littley: This godoc could use some refactoring. Things like "in the only order they permit" doesn't really help the reader understand the function. And we can probably drop things like saying we validate the config first, since that's an implementation detail not a contract.

sei-db/bootstrap/storage_manager.go
```diff
@@ -0,0 +1,286 @@
...
+func (m *GigaStorageManager) start(ctx context.Context, cfg config.GigaStorageConfig) error {
```
cody-littley: Looking at this method, does it make sense to have this be a stand alone function instead of happening inline inside the constructor? All the constructor does is to call this method and check its error.

sei-db/bootstrap/storage_manager.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+	if err := m.CrashRecover(); err != nil {
+		return err
+	}
+	m.startCheckpointSchedule(cfg.CheckpointConfig)
+	return m.startGarbageCollector(ctx, cfg.PruningConfig)
+}
+
+// openDBs fills in the stores this config asks for, in an order chosen so that each one is reachable
+// by Close as soon as it exists: the block ledger, receipts, the state WAL, SC (which adopts the
```
cody-littley: Seems odd that the ordering matters like this.

sei-db/bootstrap/storage_manager.go
```diff
@@ -0,0 +1,286 @@
...
+// WAL), then SS.
+//
+// Receipts and SS each have an Enable of their own and are left nil when it is false. The block
+// ledger, the state WAL and SC have none: a Giga node is not a node without them.
+//
+// Each store recovers itself to whatever height its own durable state reaches, so on return they can
+// disagree with each other. CrashRecover is what closes that gap.
```
cody-littley: We shouldn't ask each store to recover to their latest height before we do CrashRecover. This could cause us to replay the WAL twice.

Let's say that SC can automatically advance to block 1000 by replaying the WAL from block 500. But during CrashRecover(), we decide we actually want to load block 999. It will have to throw away the block 1000 state and do the WAL recovery a second time.

A better pattern is to do the following:

1. Inspecting files on disk, determine the block height each store can load. This may not require us to construct the stores yet.
2. Decide which height we want to load.
3. Instruct each store to load the height we want to load.

sei-db/bootstrap/storage_manager.go
```diff
@@ -0,0 +1,286 @@
...
+		}
+	}
+
+	// TODO: pass wal, sc and ss to GigaStateDB constructor
+
+	return nil
+}
+
+// startCheckpointSchedule puts the halves of state this node opened on one checkpoint cadence, in
+// place of the per-store intervals they would otherwise each keep.
```
cody-littley: Nit, I suggest we avoid putting design history in godocs. I.e. we shouldn't describe the way things used to work, we should describe the way they currently work in the code. In this case, "in place of the per-store intervals they would otherwise each keep" describes an old design.

sei-db/config/testdata/receipt-store.golden
cody-littley, line 1: Note, I deleted the golden config file tests in main. I suggest merging main, which will let you delete this file (and several others... leaving this one comment instead of flagging all of them).

sei-db/state_db/sc/flatkv/snapshot_writer.go
cody-littley, line 1: FYI, this file may have merge conflicts with https://github.com/sei-protocol/sei-chain/pull/4061

sei-db/state_db/sc/flatkv/store.go
```diff
@@ -360,6 +366,23 @@ func (s *CommitStore) flatkvDir() string {
 	return s.config.DataDir
 }
 
+// SetCheckpointScheduler hands this store the schedule it takes its checkpoint heights from, in place
+// of its own SnapshotInterval. Every committed block is then offered to the schedule, which is what
+// holds this store and every other store on it to the same heights.
+//
+// A nil scheduler returns the store to its interval. Safe to call on an open store, and it survives a
+// later reopen.
+func (s *CommitStore) SetCheckpointScheduler(scheduler *controller.CheckpointScheduler) {
```
cody-littley, line 376: Does an unset scheduler just mean that we should not be taking snapshots?

sei-db/config/giga_config.go
```diff
@@ -8,39 +8,31 @@ import (
...
+// avoid an import cycle with the receipt package.
+const gigaReceiptBackend = "littidx"
+
+// DefaultGigaStorageConfig returns a config rooted at homePath:
 //
 //	data/state_commit/flatkv
-//	data/state_store/evm/{backend}   (sole SS; no Cosmos SS in Giga)
+//	data/state_store/evm/{backend}
 //	data/ledger/receipt/{backend}
 //	data/ledger/block
```
blindchaser, line 31: is the default path correct `data/ledger/block`?

sei-db/bootstrap/recovery.go
yzang2019: It's not implemented yet so haven't add the full comment. But yeah that's what we agreed on.

sei-db/bootstrap/storage_manager.go
yzang2019: Fixed

sei-db/bootstrap/storage_manager.go
yzang2019: Yeah, agreed

sei-db/bootstrap/storage_manager.go
yzang2019: True, fixing the godocs

sei-db/bootstrap/storage_manager.go
yzang2019: Agreed, but I think this can be a future optimization. Since we don't have the interface, right now it's much easier to just open the DB directly to latest height and then call rollback to a specific height.

sei-db/bootstrap/storage_manager.go
yzang2019: Fixed

sei-db/config/testdata/receipt-store.golden
yzang2019: This one seems left over, will delete it

sei-db/state_db/sc/flatkv/snapshot_writer.go
yzang2019: Will address later once you merge your PR

sei-db/config/giga_config.go
yzang2019: yeah, that's the path for giga

sei-db/config/toml.go
```diff
@@ -205,11 +211,6 @@ async-write-buffer = {{ .ReceiptStore.AsyncWriteBuffer }}
 # defaults to 600 seconds
 prune-interval-seconds = {{ .ReceiptStore.PruneIntervalSeconds }}
 
-# EnableReadWriteMetrics emits estimated read/write counters for Pebble-backed
-# receipt storage.
-# Default: false.
-enable-read-write-metrics = {{ .ReceiptStore.EnableReadWriteMetrics }}
```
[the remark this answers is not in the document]
blindchaser: +1

sei-db/state_db/sc/flatkv/store.go
yzang2019: I don't think we support unset, if the CommitStore has a CheckpointScheduler, it means it will use the CheckpointScheduler to create checkpoint. Otherwise it will just follow the legacy config to take snapshot.

app/app.go
```diff
@@ -659,7 +659,9 @@ func New(
 	if err != nil {
 		panic(fmt.Sprintf("error reading receipt store config: %s", err))
 	}
-	if app.receiptStore == nil {
+	// A nil store is the seam tests inject their own through, and it is also what a disabled
+	// receipt store leaves behind: the EVM keeper reads nil as a node that keeps no receipts.
+	if app.receiptStore == nil && receiptConfig.Enable {
```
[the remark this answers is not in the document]
yzang2019: Should be fixed now

sei-db/config/toml.go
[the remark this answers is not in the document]
yzang2019: Fixed, making the config map structure "-"

yzang2019:
@seidroid review

x/evm/keeper/receipt.go
```diff
@@ -140,6 +140,11 @@ func isLegacyReceipt(ctx sdk.Context, receipt *types.Receipt) bool {
 }
 
 func (k *Keeper) flushTransientReceipts(ctx sdk.Context) error {
+	// An absent receipt store means the node keeps no receipts, so there is nothing to flush. This
+	// runs from the pre-commit handler, where a returned error panics the node, so it must not be one.
+	if k.receiptStore == nil {
```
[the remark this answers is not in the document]
yzang2019: That's fine, the migration has already done

yzang2019:
@seidroid review

yzang2019:
@seidroid review

sei-db/bootstrap/storage_manager.go
cody-littley: Ok merging as-is.

cody-littley (approved):

PR #4064, opened 2026-09-01
Tests: eth exec specs

kollegian:
## Summary
- Add a standalone, informational post-merge workflow for Ethereum execution-spec coverage.
- Run pinned `ethereum/tests` TransactionTests and RLPTests through Sei's go-ethereum fork.
- Run Prague EEST across eight isolated devnet shards plus one converted legacy state-test runner.
- Exclude the EIP-2929 precompile family until it has per-vector state isolation.

## Testing performed
- TransactionTests and RLPTests pass locally.
- Verified the generated matrix contains eight regular shards and one ported-state runner.
- Verified deterministic shard coverage with no duplicate or missing vectors.
- Passed Ruff, shell syntax, actionlint, and whitespace checks.


.github/workflows/execution-specs.yml
```diff
@@ -0,0 +1,330 @@
...
+on:
+  push:
+    branches:
+      - main
+      # TEMP: remove after the final CI validation.
+      - tests/exec-specs
+
+concurrency:
+  group: execution-specs-${{ github.ref }}
+  cancel-in-progress: false
```
amir-deris, line 10: Should this one remain `false`? I think it means if newer commits are pushed to branch, it doesn't cancel the older currently running ones. Setting to `true` can save runner time and cost.

integration_test/exec_specs/patches/sei-compat.patch
```diff
@@ -0,0 +1,509 @@
...
+     """
+@@ -79,6 +165,11 @@ class TransactionPost(BaseExecute):
+         balances: Dict[Address, int] = {}
+         for block in self.blocks:
+             for tx in block:
++                # Preserve deliberately underfunded senders for transaction
++                # rejection tests. Only valid transactions should cause the
++                # remote pre-allocation layer to increase explicit funding.
++                if tx.error is not None:
++                    continue
```
masih, line 271: Hm this would avoid funding _every_ tx with error? This means we then mask expected errors with "insufficient funds".

I am curious, why?

An alternative is to fund invalid transactions as normal except for `INSUFFICIENT_ACCOUNT_FUNDS`?

.github/workflows/execution-specs.yml
```diff
@@ -0,0 +1,330 @@
...
+env:
+  GHCR_LOCALNODE: ghcr.io/sei-protocol/sei-chain-exec-specs-localnode
+  GHCR_RPCNODE: ghcr.io/sei-protocol/sei-chain-exec-specs-rpcnode
+
+jobs:
+  transaction-rlp:
+    name: Ethereum Transaction and RLP Tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    continue-on-error: true
```
masih: This means CI is effectively always green? ditto in other places.

What was the rationale for this ?

.github/workflows/execution-specs.yml
```diff
@@ -0,0 +1,330 @@
...
+    steps:
+      # See: https://github.com/actions/checkout/releases/tag/v7.0.0
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0
+      - name: Download execution-spec reports
+        uses: actions/download-artifact@v4
+        with:
+          pattern: exec-specs-*-${{ github.run_id }}
+          path: integration_test/exec_specs/reports/downloaded
+      - name: Write run summary
```
masih, line 285: We should check expliclty the expected report name per chain, otherwise a partial failure gets swallowed up in a normal looking report?

masih (changes requested):
Exciting to see these landing 🙌 

I need your help to understand the rationale behind continue on error and some of the patch work since as is they make the CI look green regardless or can shadow root causes.

integration_test/exec_specs/scripts/prepare_admin.ts
```diff
@@ -0,0 +1,27 @@
+/**
```
amir-deris, line 1: I was wondering do we need to use the mix of `typescript`, `bash` and `python` for these tests, or is it better if everything could be written in `go`?

amir-deris (approved):

.github/workflows/execution-specs.yml
kollegian: Yes, this was an experiment from before and latest runs were all passing so I didnt notice. Getting too AI complacent I am afraid. Updated.

.github/workflows/execution-specs.yml
kollegian: Updated.

integration_test/exec_specs/patches/sei-compat.patch
kollegian: You are correct, updated the tests.

.github/workflows/execution-specs.yml
kollegian: This runs only when PR's are merged to main. I guess it would be good to have this running per merged PR so that we would know which PR is the issue.

.github/workflows/execution-specs.yml
```diff
@@ -0,0 +1,326 @@
...
+          fi
+          for package in sei-chain-exec-specs-localnode sei-chain-exec-specs-rpcnode; do
+            versions="$(
+              gh api --paginate "${base}/${package}/versions" \
+                --jq ".[] | select(.metadata.container.tags | index(\"${RUN_TAG}\")) | .id" ||
+                true
+            )"
+            while read -r version; do
+              [[ -n "$version" ]] || continue
+              gh api --method DELETE "${base}/${package}/versions/${version}" || true
```
masih: I would: remove || true; retry, attempt both packages, then fail if either operation failed.

masih (approved):
🙌

.github/workflows/execution-specs.yml
kollegian: updated

masih (approved):

PR #4070, opened 2026-09-01
fix(wasm): bound legacy Prometheus cardinality for execute and QuerySmart (CON-336, CON-337)

amir-deris:
## Summary
- Remove the labeled `wasmd_execute_contract_latency` defer from `MsgExecuteContract`; execute latency remains on the keeper path via `wasm_contract_execute_duration` / `wasm.contract.execute`.
- Drop raw `contract_address` labels from QuerySmart legacy telemetry and record the invocation counter only after contract lookup succeeds.

Closes the Immunefi RPC OOM vectors in CON-336 and CON-337 where unique valid `sei...` addresses could retain unbounded in-process Prometheus series.


amir-deris:
Will fix the issue by migration to Otel instead.

PR #4072, opened 2026-09-01
fix(mempool): tests for transaction invalidation across v2/giga (CON-375)

shemnon:
## Summary
- The mempool checked each queued nonce's cost against the same static per-account balance snapshot instead of a running total, so two txs from one account could both look individually affordable even though their combined cost exceeded the real balance. The second would then fail ante's real balance check at inclusion and land on-chain as a `status=0, gasUsed=0` stub receipt instead of staying pending.
- Adds a `readyCost` accumulator to `evmAccount` and threads it through the readiness-advance loop, the same-nonce replacement check, and `shouldReject`, so all three agree on what's actually affordable.
- Fixes `testEnv.readyTxs()`'s test oracle in `tx_test.go`, which had the identical bug baked in, and adds two targeted regression tests.
- Adds an EIP-7702 differential test (CON-411): a delegate/drain/undelegate sequence reproduces the same stub-receipt symptom via a completely different mechanism (execution-time SELFBALANCE-triggered balance loss, invisible to any admission-time cost estimate). Confirms Giga has no independent code path for this failure class -- any fee/nonce/balance validation failure makes it unconditionally fall back to V2 -- and that the fallback produces byte-identical results.

## Test plan
- [x] `go test ./sei-tendermint/internal/mempool/...` -- new regression tests fail against the pre-fix code (verified via temporary stash), pass with the fix
- [x] `go test ./app/...` -- CON-411 V2 integration test passes
- [x] `go test ./giga/tests/...` -- CON-411 Giga-differential test passes (V2/Giga byte-identical results)
- [x] `gofmt -s` / `goimports` clean on all changed files
- [x] Manually reproduced the original bug end-to-end against a local v6.6.2 and HEAD docker cluster before the fix, and confirmed the fix corrects it (tx2 stays pending, un-hangs correctly once funded)

Linear: CON-375, CON-411


giga/tests/eip7702_selfbalance_drain_test.go
```diff
@@ -0,0 +1,182 @@
...
+
+// buildEIP7702SelfBalanceDrainTxs builds the delegate/drain/undelegate
+// sequence, associating both accounts first so Giga uses its native EIP-7702
+// path instead of deferring an unassociated authority to V2.
+func buildEIP7702SelfBalanceDrainTxs(t *testing.T, tCtx *GigaTestContext, sponsor, authority utils.TestAcct) (txs [][]byte, drainAddr common.Address) {
+	t.Helper()
+	// Three sponsor txs at up to 500,000 gas * 100 gwei each need up to 4.5e16
+	// wei; fund generously above that.
+	fundAccount(t, tCtx, sponsor.AccountAddress, big.NewInt(1_000_000_000_000_000_000)) // gas money only
+	fundAccount(t, tCtx, authority.AccountAddress, big.NewInt(1_000_000_000_000_000))   // the balance drain() will empty
```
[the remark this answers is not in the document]
shemnon: increased to 1e18

sei-tendermint/internal/mempool/tx.go
```diff
@@ -391,7 +430,7 @@ func (s *txStore) insert(inner *txStoreInner, wtx *WrappedTx, recordAdded bool)
 			oldEvm := old.evm.OrPanic("non-evm tx")
 			oldReady := oldEvm.nonce < account.nextNonce
 			// If the old tx is ready but the new tx is not, then reject the new tx.
-			if oldReady && account.balance.Cmp(&evm.requiredBalance) < 0 {
+			if oldReady && !account.canReplace(&oldEvm.requiredBalance, &evm.requiredBalance) {
```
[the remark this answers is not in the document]
shemnon: This is exposing a different seam outside the scope of this bug, area where behavior is intended to change. Hence there is no need to test.

app/eip7702_selfbalance_drain_test.go
```diff
@@ -0,0 +1,174 @@
+package app_test
+
```
wen-coding, line 2: @codchen how do we normally struct tests and directories in app/? Should we create a test directory here? Do we normally aggregate tests by EIP?

giga/tests/eip7702_selfbalance_drain_test.go
```diff
@@ -0,0 +1,182 @@
+package giga_test
+
```
wen-coding, line 2: nit: I suppose we will have similar tests later, guaranteeing v2 and giga have the same behavior on other EIPs. Do you think we should reorganize tests so they share common fixtures?

sei-tendermint/internal/mempool/tx.go
```diff
@@ -342,7 +381,7 @@ func (inner *txStoreInner) shouldReject(txHash types.TxHash) bool {
 	oldEvm := old.evm.OrPanic("non-evm tx")
 	oldReady := oldEvm.nonce < account.nextNonce
 	// If the old tx is ready but the new tx is not, then reject the new tx.
```
wen-coding, line 344: Is this comment still accurate?

sei-tendermint/internal/mempool/tx.go
```diff
@@ -91,6 +91,45 @@ type evmAccount struct {
...
+	return !overflow && a.balance.Cmp(sum) >= 0
+}
+
+// commitReady records that cost is now committed against a newly-ready nonce.
+func (a *evmAccount) commitReady(cost *uint256.Int) {
+	a.readyCost.Add(&a.readyCost, cost)
+}
+
+// replaceReady swaps the committed cost of an already-ready nonce. Callers
+// must have already validated the swap via canReplace with the same costs.
```
wen-coding: Do you mean this relies on canReplace running first? What if in the future we refactor the code so that's no longer true? How do we catch it in that case?

sei-tendermint/internal/mempool/tx.go
```diff
@@ -391,7 +430,7 @@ func (s *txStore) insert(inner *txStoreInner, wtx *WrappedTx, recordAdded bool)
 			oldEvm := old.evm.OrPanic("non-evm tx")
 			oldReady := oldEvm.nonce < account.nextNonce
 			// If the old tx is ready but the new tx is not, then reject the new tx.
```
wen-coding, line 393: Is this comment still accurate?

app/eip7702_selfbalance_drain_test.go
shemnon: This is in `app_test` as it uses finalizeToBlockProcessReq from ./app_test.go, but that file uses the app_test package. Been that way for 4 years.

giga/tests/eip7702_selfbalance_drain_test.go
shemnon: The common fixtures would only work for what is externally the same, as giga and v2 diverge the specialties will diverge.  I'm of the opinion commonality should exist in what is tested and not how, and sharing common fixtures is a how we test issue.

sei-tendermint/internal/mempool/tx.go
shemnon: Based on the seidroid comment, this enters into changing behavior, which was supposed to be outside the scope of CON-375, so I reverted it.

sei-tendermint/internal/mempool/tx.go
shemnon: code is reverted, so yes.

sei-tendermint/internal/mempool/tx.go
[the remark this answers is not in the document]
shemnon: on second thought I realize I'm introducing changes in what was supposed to be an "assert sameness" issue.  This line of code has been removed from this pr.

sei-tendermint/internal/mempool/tx.go
shemnon: reverted, so yes.

wen-coding (approved):

PR #4078, opened 2026-09-02
PLT-1125: Bound concurrent in-flight RPCs and connections per IP on the gRPC query plane

amir-deris:
Closes [PLT-1125](https://linear.app/seilabs/issue/PLT-1125/bound-concurrent-in-flight-rpcs-and-connections-per-ip-on-the-grpc).

## Impact

Caps how much load a single client IP can place on the gRPC query plane (`:9090` and `:9091`):

- **Connections:** optional per-IP limit via `max-connections-per-ip` (default **0**, unlimited). Set a positive value to cap one address's share of the global connection budget.
- **In-flight RPCs:** when rate limiting is enabled, each IP is limited to 100 concurrent RPCs (`max-in-flight-per-ip`), complementing the per-IP token bucket from #4021.

Both planes share the same per-IP pools. Rejections are counted separately from rate-limit rejections (`rpc_connection_rejected_total`, `rpc_inflight_rejected_total`).

**Operators:** clients behind a shared egress or reaching the node over `127.0.0.1` share one allowance when a positive cap is set.


amir-deris:
@seidroid review

sei-will (approved):

masih (approved):

PR #4079, opened 2026-09-02
Add crash recovery logic for GigaStorageManager

yzang2019:

Adds crash recovery to `GigaStorageManager`, so that after an unclean shutdown:

1. Every store is at or below the block store's height.
2. Every store other than the block store is on the same height.

Getting there needed one structural change: **`giga.StateDB` now owns the state commit store, the EVM state store, and the state WAL those two share.** Previously the manager held all four pieces itself and SC opened a WAL of its own, which left SC and the manager both writing the same WAL — a double-append that makes recovery-by-replay impossible. SC and SS are now each constructed without a WAL, `StateDB` is the only writer, and the replay that brings either half onto a height reads through that one WAL.

### Recovery flow

`OpenDBWithRecovery` opens the block store and the receipt store, computes a target, and brings everything else onto it:

- **`findTargetRecoveryHeight`** — the lowest head among the block store, the state WAL and the receipt store, receipts being skipped when disabled. The WAL tail is read offline through `statewal.GetRange`, with no live WAL open. A target of `0` means there is no height to converge on, and nothing is moved.
- **`recoverReceipt`** — the store has to be closed to be rewritten, so it closes, rolls back offline through `receipt.Rollback`, and reopens.
- **`openStateDB`** — opens SC, the WAL and SS where it finds them, and puts both halves of state on one checkpoint schedule. It converges nothing: the caller names the height, because only the caller knows what the stores outside the StateDB can serve.
- **`recoverState`** — `StateDB.RollbackTo(target)`.

`RollbackTo` is the single convergence primitive, and it is what readies a freshly opened StateDB as much as what rewinds a running one. It:

1. rewinds SC to a snapshot boundary at or below the target, when SC sits above it;
2. truncates the WAL tail to the target, so the next commit is the block after it;
3. rewinds SS to its newest snapshot at or below the target, when SS sits above it, discarding the snapshots above it;
4. replays the WAL forward into SC;
5. replays the WAL forward into SS.

Each half is handled independently of the other, which is what lets them have crashed at different heights.

### New store-level APIs

- `flatkv.CommitStore.RewindToSnapshotAtOrBelow` — `Rollback` for a store whose WAL an outer context owns. It moves only between snapshot boundaries, so no WAL crosses the API and replaying forward from the version it returns is the caller's to do. It shares `repointAtSnapshot` with the existing `Rollback`, which is unchanged.
- `flatkv.StateWALPath` / `StateWALConfig` — how the WAL's owner locates and reopens it without repeating FlatKV's layout convention.
- `evm.EVMStateStore.RewindToSnapshotAtOrBelow` / `ApplyReplayedBlock` — the SS mirror of the split above: the rewind moves between snapshot boundaries and discards the snapshots above the target, and applying one replayed block is the per-block step `StateDB`'s replay drives. No WAL crosses either, so both halves of state reach the same replay and its missing-blocks check.
- `snapshot.Manager.RemoveSnapshotsAbove` — drops the snapshots a rollback has just discarded the history for, repointing `current` at the newest survivor.
- `receipt.Rollback` — offline rollback of the LittDB receipt bodies plus a rewind of the PebbleDB tag index.
- `statewal.GetRange` / `PruneAfter` / `VerifyIntegrity` now take a directory path rather than a `*Config`, so they can run against a WAL directory with no live instance open.
- The `StateDB` contract gains `RollbackTo` and `Close`.

### `giga` contracts moved to `giga/types`

`flatkv` and `composite` imported `giga` for `LiveStateStore` and `StateView`, which meant `giga` could not import `flatkv` to construct SC from config. Those interfaces and the EVM value types in their signatures now live in `sei-db/state_db/giga/types` (imported as `gigatypes`), a package with no implementation. The dependency runs one way again: the `giga` implementation imports `flatkv`, and both sides depend only on the contract package.

### Smaller changes

- `DefaultGigaStorageConfig` returns `*GigaStorageConfig`, and `Validate` takes a pointer receiver and guards nil. Adds the `WithValidatorMode` / `WithFullNodeMode` / `With{Account,Storage,Code}DBCacheSize` builders.
- `GigaStorageManager` is down to the block store, the receipt store, the StateDB and the collector. `SC()` / `SS()` / `StateWAL()` delegate to the StateDB, `prunableStores` takes the state stores from it as a group, and `Close` reports every store's failure rather than stopping at the first.

`sei-db/bootstrap/recovery_test.go` covers each skew a crash can leave behind:

| Test | Skew |
| --- | --- |
| `TestRecoverSCReplaysAMissedWALBlock` | SC behind the WAL |
| `TestRecoverSCRollsBackToTheTarget` | SC above the target |
| `TestRecoverSCAboveTheWALHeadRewindsToASnapshotAndReplays` | SC above the WAL head, via a snapshot |
| `TestRecoverSSReplaysEVMChangesets` | SS behind the WAL |
| `TestRecoverSSRollsBackToTheTarget` | SS above the target |
| `TestRecoverReceiptRewindsTheHead` | receipts above the target |
| `TestRecoverStateDropsWALBlocksAboveTheTarget` | WAL above the target |
| `TestStateDBRollbackToRewindsBothHalvesAndTheWAL` | both halves and the WAL in one call |
| `TestFindTargetRecoveryHeightIsZeroWithoutABlockLedger` | target computation |
| `TestOpenDBWithoutRecoveryOnAFreshHome` | fresh node, every height zero |

`storage_manager_test.go` additionally covers the checkpoint schedule reaching both halves of state, every store joining the prune cycle in order, and `Close` on a partial open.

`TestCatchUpRefusesAWALMissingTheBlocksAStoreNeeds` covers both halves refusing a WAL pruned past a store's head, `TestRecoverSSRemovesSnapshotsAboveTheTarget` the snapshot tree after a rollback, and `TestHealInterruptedRestore` a snapshot restore interrupted between its two renames.

Verified with `scripts/ramtest.sh ./sei-db/...` (56 packages green) and `scripts/ramtest.sh ./giga/...`, plus `make fmtcheck` and golangci-lint v2.8.0 clean on `giga`, `bootstrap` and `flatkv`.

## Known gaps, called out for review

None of these are regressions; they are limits of what this PR reaches.

- **`CommitStateChanges` still does not write SS** (`TODO: Commit changes to SS`). SS recovery is implemented and tested, but on the commit path SS only moves when something else populates it, so the SS half of convergence is not yet exercised by ordinary block production.
- **Rolling SS back needs a snapshot at or below the target.** SC always has one, so its rewind cannot fail this way; SS has one only where a checkpoint landed at or below the target, and there is no fallback to replaying from empty.
- **The checkpoint schedule is live during recovery replay**, so a replay that crosses a boundary snapshots there. That costs some recovery time and is the reason the schedule is documented as being in place before the StateDB is put on a height.
- **A WAL truncation replaces the handle, and nothing re-hands the new one out.** `StateDB.WAL()` must therefore be re-read after any `RollbackTo`, and `RollbackTo` must not run once the prune cycle holds the WAL. Recovery finishing before `startGarbageCollector` is what keeps that safe today; it is not enforced.
- **`receipt.Rollback` supports only the `littidx` backend** and refuses any other.
- **`OpenViewAt` panics** — serving a past height needs the historical state DB, which is not wired into `StateDB`.


yzang2019:
@seidroid review

yzang2019:
@seidroid review

yzang2019:
@seidroid review

sei-db/bootstrap/recovery.go
```diff
@@ -1,10 +1,170 @@
...
+// An empty block store or state WAL instead yields 0, which skips recovery. Neither is unambiguous the
+// way an empty receipt store is: state whose WAL was pruned away behind a snapshot still exists with an
+// empty WAL, and converging on a target derived from the other stores would discard it with no WAL left
+// to replay it from.
+func recoveryTarget(blockHeight, stateHeight, receiptHeight uint64) uint64 {
+	if blockHeight == 0 || stateHeight == 0 {
+		return 0
+	}
+	target := min(blockHeight, stateHeight)
+	if receiptHeight > 0 {
```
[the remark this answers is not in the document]
yzang2019: This is not usually possible since a node will stays in its role and never change

sei-db/bootstrap/recovery.go
```diff
@@ -1,10 +1,170 @@
...
-func (m *GigaStorageManager) CrashRecover() error {
+//  2. Every store other than the block store is on the same height, unless it holds no history at all,
+//     in which case it is left empty to fill forward from that height.
+//
+// The target is the lowest head among the block store, the state WAL and the receipt store. A target of
+// 0 means there is no height to converge on and nothing is moved.
+//
+// The two halves of state and the WAL they share belong to the StateDB, which opens them where it
+// finds them. recoverStores is the step that puts every store on the target and readies state for the
+// block after it.
```
cody-littley: Suggest removing this last godoc paragraph

sei-db/bootstrap/recovery.go
```diff
@@ -1,10 +1,170 @@
...
+	if err != nil {
+		return err
+	}
+	if err := m.openStateDB(ctx); err != nil {
+		return err
+	}
+	return m.recoverStores(targetHeight)
+}
+
+// recoverStores puts the receipt store and the two halves of state on target.
```
cody-littley: nit

I understand what it means by "two halves of the state", but I don't think that's a good way to describe it. Some nodes won't be running an SS. Perhaphs we can say `recoverStores aligns the block height of the the receipt, the live state (SC), and the historical state (SS, if enabled).`

sei-db/bootstrap/recovery.go
```diff
@@ -1,10 +1,170 @@
...
+//
+// A target of 0 is no height to converge on, and every store is left as it was found: rolling back to
+// it would drop every receipt the node holds along with every block in its WAL. This is the single
+// guard for that, which is why the two rollbacks below it carry none of their own.
+func (m *GigaStorageManager) recoverStores(target int64) error {
```
cody-littley, line 48: Should this method also be responsible for rolling back StateWAL?

sei-db/bootstrap/recovery.go
```diff
@@ -1,10 +1,181 @@
...
+		return nil
+	}
+	head := m.receiptDB.LatestVersion()
+	if head <= target {
+		return nil
+	}
+	if err := m.receiptDB.Close(); err != nil {
+		return fmt.Errorf("close receipt store before rolling it back to %d: %w", target, err)
+	}
+	m.receiptDB = nil
```
cody-littley: Once my PR with offline methods is available, we might want to change this so it doesn't startup and then stop the DB for this step. That PR is in the merge queue (as of the moment I type this comment).

sei-db/ledger_db/receipt/rollback.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	return rollbackLittBodies(cfg, uint64(target)) //nolint:gosec // target >= 0
+}
+
+func rollbackLittBodies(cfg dbconfig.ReceiptStoreConfig, target uint64) error {
+	littDir := filepath.Join(cfg.DBDirectory, "littdb")
+	if _, err := os.Stat(littDir); os.IsNotExist(err) {
+		return nil
+	}
+	return rollback.RollbackLittDB([]string{littDir}, func(_ string, key []byte, isPrimary bool) (bool, error) {
```
cody-littley: FYI, the package name for this changes in my PR from `rollback` to `offline`. Mentioning this incase the merge conflict makes it unclear what happened.

cody-littley (approved):

sei-db/bootstrap/recovery.go
yzang2019: Rollback StateWAL will be done in StateDB since WAL lifecycle is managed by StateDB now

sei-db/bootstrap/recovery.go
yzang2019: Sounds good!

sei-db/bootstrap/recovery.go
yzang2019: Fixed

sei-db/bootstrap/recovery.go
yzang2019: Fixed

yzang2019:
@seidroid review

yzang2019:
@seidroid review

yzang2019:
@seidroid review

PR #4086, opened 2026-09-02
(Autobahn) dial peers from the live commit-epoch committee (CON-358)

wen-coding:
## Summary
- Dial outbound giga and EVM RPC from the live commit-epoch committee instead of the static address book. Departing members are cancelled, then awaited, before the slot is reused.
- Inbound giga role is fixed at accept: a membership change closes the connection so the peer's dialer reconnects into the role it now has. The inbound fullnode cap is still decided at accept.
- Skip republishing `NextCommitEpoch` when the epoch index is unchanged so watchers do not thrash on every QC.

Linear: [CON-358](https://linear.app/seilabs/issue/CON-358)

## Test plan
- [ ] `cd sei-tendermint && GOWORK=off go test ./internal/p2p/ -count=1 -run 'TestGigaRouter'`
- [ ] `cd sei-tendermint && GOWORK=off go test ./internal/p2p/... -count=1`
- [ ] Confirm a validator that leaves the committee is dropped from outbound giga/EVM and, if it redials, is admitted as a capped block-sync inbound
- [ ] Confirm a fullnode block-sync connection is not dropped on an epoch tick that does not change the committee

### DIFF
--- a/sei-tendermint/internal/p2p/giga_router_common.go
+++ b/sei-tendermint/internal/p2p/giga_router_common.go
@@ -45,6 +45,9 @@ type gigaRouterCommon struct {
 	// nextCommitEpoch is data.NextCommitEpoch() cached at construction so
 	// EvmProxy can Load() without taking the data lock on every call.
 	nextCommitEpoch utils.AtomicRecv[*atypes.Epoch]
+	// anchor is data.Anchor() cached at construction: the AppQC/CommitQC covering
+	// the lowest row data.State still holds.
+	anchor utils.AtomicRecv[utils.Option[data.Anchor]]
 
 	// inboundFullnodeCount tracks inbound connections currently served the
 	// block-sync subset. Optimistic Add(1) + compare against cap;
@@ -528,12 +531,37 @@ type memberSession struct {
 	done   chan struct{}
 }
 
-// stopDepartingMembers stops sessions for validators outside committee.
-func stopDepartingMembers(ctx context.Context, live map[atypes.PublicKey]*memberSession, committee *atypes.Committee) error {
+// keepReplicas is commitEpoch's committee, plus the committee Anchor covers when present.
+func keepReplicas(anchor utils.Option[data.Anchor], commitEpoch *atypes.Epoch) map[atypes.PublicKey]struct{} {
+	keep := map[atypes.PublicKey]struct{}{}
+	for lane := range commitEpoch.Committee().Lanes().All() {
+		keep[lane.Validator] = struct{}{}
+	}
+	if a, ok := anchor.Get(); ok {
+		for lane := range a.Epoch.Committee().Lanes().All() {
+			keep[lane.Validator] = struct{}{}
+		}
+	}
+	return keep
+}
+
+// stopDepartingMembers stops sessions for validators outside keepReplicas once
+// Anchor is at most one epoch behind commitEpoch.
+func stopDepartingMembers(
+	ctx context.Context,
+	live map[atypes.PublicKey]*memberSession,
+	anchor utils.Option[data.Anchor],
+	commitEpoch *atypes.Epoch,
+) error {
+	a, ok := anchor.Get()
+	if !ok || commitEpoch.EpochIndex() > a.Epoch.EpochIndex()+1 {
+		return nil
+	}
+	keep := keepReplicas(anchor, commitEpoch)
 	var departing []*memberSession
 	// Cancel every departing session before waiting for any of them.
 	for validator, session := range live {
-		if committee.HasReplica(validator) {
+		if _, ok := keep[validator]; ok {
 			continue
 		}
 		session.cancel()
@@ -548,8 +576,7 @@ func stopDepartingMembers(ctx context.Context, live map[atypes.PublicKey]*member
 	return nil
 }
 
-// runPerCommitteeMember runs tasks for each reachable member of the
-// committee covering the next CommitQC.
+// runPerCommitteeMember runs tasks for each reachable member needed by the commit epoch or Anchor.
 func (r *gigaRouterCommon) runPerCommitteeMember(ctx context.Context, tasks ...committeeMemberTask) error {
 	return scope.Run(ctx, func(ctx context.Context, s scope.Scope) error {
 		live := map[atypes.PublicKey]*memberSession{}
@@ -559,13 +586,18 @@ func (r *gigaRouterCommon) runPerCommitteeMember(ctx context.Context, tasks ...c
 				session.cancel()
 			}
 		}()
-		return r.nextCommitEpoch.Iter(ctx, func(_ context.Context, epoch *atypes.Epoch) error {
-			committee := epoch.Committee()
-			if err := stopDepartingMembers(ctx, live, committee); err != nil {
+		// Anchor is republished on every persist batch; only its epoch can
+		// change the keep set.
+		epochOf := func(opt utils.Option[data.Anchor]) utils.Option[atypes.EpochIndex] {
+			return utils.MapOpt(opt, func(a data.Anchor) atypes.EpochIndex { return a.Epoch.EpochIndex() })
+		}
+		for ctx.Err() == nil {
+			commitEpoch := r.nextCommitEpoch.Load()
+			anchor := r.anchor.Load()
+			if err := stopDepartingMembers(ctx, live, anchor, commitEpoch); err != nil {
 				return err
 			}
-			for lane := range committee.Lanes().All() {
-				validator := lane.Validator
+			for validator := range keepReplicas(anchor, commitEpoch) {
 				if _, ok := live[validator]; ok {
 					continue
 				}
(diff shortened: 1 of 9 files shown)


sei-tendermint/internal/p2p/giga_router_common.go
```diff
@@ -557,9 +655,22 @@ func (r *gigaRouterCommon) RunInboundConn(ctx context.Context, hConn *handshaked
...
 			defer Global.gigaConnsAt("in").Add(-1)
-			if err := r.service.RunServer(ctx, server, isCommittee); err != nil {
+			v, ok := validator.Get()
+			if !ok {
+				if err := r.service.RunServer(ctx, server, false); err != nil {
+					return fmt.Errorf("inbound from %v: %w", key, err)
+				}
+				return nil
+			}
+			changed, err := r.runUntilMembershipChange(ctx, v, isCommittee, func(ctx context.Context) error {
```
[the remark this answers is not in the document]
wen-coding: done

sei-tendermint/internal/p2p/giga_router_common.go
```diff
@@ -524,26 +516,132 @@ func (r *gigaRouterCommon) dialAndRunConn(
...
+					continue
+				}
+				addr, ok := r.cfg.ValidatorAddrs[validator]
+				if !ok {
+					logger.Error("committee member has no configured address; not dialing", "validator", validator)
+					continue
+				}
+				taskCtx, cancel := context.WithCancel(ctx)
+				done := make(chan struct{})
+				live[validator] = &memberSession{cancel: cancel, done: done}
```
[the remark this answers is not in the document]
wen-coding: added doc

sei-tendermint/internal/p2p/giga_router_validator.go
```diff
@@ -73,31 +73,32 @@ func (r *gigaValidatorRouter) Run(ctx context.Context) error {
...
 		s.SpawnNamed("producer", func() error { return r.producer.Run(ctx) })
 		s.SpawnNamed("data", func() error { return r.data.Run(ctx) })
 		s.SpawnNamed("execute", func() error { return r.runExecute(ctx) })
 		s.SpawnNamed("service", func() error { return r.service.Run(ctx) })
-		s.SpawnNamed("evmProxies", func() error { return r.runEvmProxies(ctx) })
 		return nil
 	})
 }
 
+// runCommitteePeer maintains an outbound connection to a committee member.
```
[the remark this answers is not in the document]
wen-coding: done

wen-coding:
@seidroid re-review please

sei-tendermint/internal/p2p/giga_router_fullnode.go
```diff
@@ -80,18 +82,43 @@ func (r *gigaFullnodeRouter) EvmProxy(sender common.Address) utils.Option[*ethrp
...
-		addr := r.cfg.ValidatorAddrs[validator]
-		err := r.dialAndRunConn(ctx, utils.Some(addr.Key), addr.HostPort, func(ctx context.Context, client rpc.Client[giga.API]) error {
-			// Consensus PublicKey (map key), not GigaNodeAddr.Key (p2p NodePublicKey).
-			return r.service.RunClient(ctx, client, validator, true)
-		})
-		logger.Info("fullnode giga connection ended; failing over", "addr", addr, "err", err)
-		if err := utils.Sleep(ctx, r.cfg.DialInterval); err != nil {
-			return err
+	for {
+		ep := r.nextCommitEpoch.Load()
```
philipsu522, line 87: what happens if this node stops and restarts when the committee view has changed, and all it knows are stale validators? i.e.
- epoch 1, committee {A, B}
- node stops
- epoch 2, committee {C}
- node starts up, only will try committee {A, B}

(ffti if this edge case is addressed in future work)

sei-tendermint/internal/p2p/giga_router_common.go
```diff
@@ -524,26 +516,134 @@ func (r *gigaRouterCommon) dialAndRunConn(
...
+		live := map[atypes.PublicKey]*memberSession{}
+		// End all sessions before the scope waits for them.
+		defer func() {
+			for _, session := range live {
+				session.cancel()
+			}
+		}()
+		return r.nextCommitEpoch.Iter(ctx, func(_ context.Context, epoch *atypes.Epoch) error {
+			committee := epoch.Committee()
+			if err := stopDepartingMembers(ctx, live, committee); err != nil {
```
philipsu522: will this cause issues if enough validators change at epoch boundary? specifically because of the lag between consensus and execution, i think we run into an issue if 33%+ validator state changes between epochs, because we immediately disconnect from those 33%+ validators but we still need them to reach AppQC. specifically
- epoch 1, committee A,B,C, we know that next commitee is C,D,E
- (disconnect from A,B, but still finalizing App state, we can't reach quorum because we need AppVotes form A,B)
- (unreachable) epoch 2, committee C,D,E

sei-tendermint/internal/p2p/giga_router_fullnode.go
wen-coding: Yes, right now if epoch 1 and epoch 2 have no overlap in committee, this validator will get stuck. I don't know whether we need to handle this case though, because that should be very rare in reality.

As long as there is one member in both epoch 1 and epoch 2 which can serve QCs, this guy will eventually reach epoch 2 and talk with the correct guys.

sei-tendermint/internal/p2p/giga_router_common.go
```diff
@@ -524,26 +516,134 @@ func (r *gigaRouterCommon) dialAndRunConn(
...
+	for v, addr := range r.cfg.ValidatorAddrs {
 		if addr.Key == key {
-			isCommittee = true
+			validator = utils.Some(v)
 			break
 		}
 	}
+	isCommittee := false
+	if v, ok := validator.Get(); ok {
+		isCommittee = r.nextCommitEpoch.Load().Committee().HasReplica(v)
```
codchen, line 684: would this mean if the val is not caught up then inbound will also be refused? (might be okay but want to check if it's intentional)

codchen (approved):

sei-tendermint/internal/p2p/giga_router_common.go
wen-coding: Nice catch, I should disconnect members in epoch 1 only when AppQC has been received, changed

sei-tendermint/internal/p2p/giga_router_common.go
wen-coding: Yes. If we can't confirm you are in the committee we won't accept your inbound connections as a validator, just to be safe. If there is fullnode capacity you can still stream blocks/QCs as a fullnode; that leaves a small door so joiners can catch up.

wen-coding:
@seidroid re-review please

sei-tendermint/libs/utils/mutex.go
```diff
@@ -124,6 +125,35 @@ func (w *atomicWatch[T]) Wait(ctx context.Context, pred func(T) bool) (T, error)
...
+// WaitAny waits for pred to hold, waking on updates to any supplied watch.
+func WaitAny(ctx context.Context, pred func() bool, watches ...AtomicUpdate) error {
+	for {
+		// Every update channel is read before pred, so a Store landing between
+		// these reads and pred still wakes the select rather than being missed.
+		cases := make([]reflect.SelectCase, 1, len(watches)+1)
+		cases[0] = reflect.SelectCase{Dir: reflect.SelectRecv, Chan: reflect.ValueOf(ctx.Done())}
+		for _, watch := range watches {
+			cases = append(cases, reflect.SelectCase{Dir: reflect.SelectRecv, Chan: reflect.ValueOf(watch.updated())})
+		}
```
shemnon, line 139: Why are the watch cases re-built between every select?

Can the Dirs and channels change inside the AtomicUpdate within the duration of the wait? Otherwise this init could live outside the for?

Either way, probably worth a one-line note.

sei-tendermint/internal/p2p/giga_router_fullnode.go
shemnon: Can we force that each new committe have at least one legacy validator?  

Does this show up if it slowly migrates to a new set?  {A, B} -> {B, C} -> {C, D}, node started in 1 was offline for 2 now the world is 3?  Does it read chain history to get the QC or is it only served out of band?

shemnon (approved):

sei-tendermint/libs/utils/mutex.go
wen-coding: It's mostly because of this specific implementation (see line 99~102 in this file), each store replaces updated with a new channel and closes the previous one, so the cases cannot be reused across iteration.

Added comments.

sei-tendermint/internal/p2p/giga_router_fullnode.go
wen-coding: We can, protocol wise either way is feasible, but it will also create various corner cases where stake != committee, which has deeper economical impact.

I can talk with Greg when he comes back to see what safe guards we need to put here.

philipsu522 (approved):

PR #4092, opened 2026-09-03
test(config): record the lookup-side divergences between the two writers

bdchatham:
## What this measures

`seid` has two writers for a node's configuration file. The `init` command
writes one for a kind of node, and a node started without a file gets one
generated by the boot instead. They do not agree, and a declared value is
what the first one writes.

The decode-delivered keys already have a record of where the two disagree.
The lookup-delivered keys had none, so the number was unknown. This adds it.

## What it found

Twelve rows across the four kinds of node, and the count varies by kind
because the declaration varies and the writer does not:

| kind | keys that disagree |
|---|---|
| validator | `api.enable`, `evm.http_enabled`, `evm.ws_enabled`, `grpc.enable`, `pruning`, `state-store.ss-enable` |
| full | `min-retain-blocks`, `pruning` |
| seed | the same six as a validator |
| archive | `pruning`, `state-store.ss-keep-recent` |

`pruning` is the one key that disagrees for every kind, and it is a
different cause from the rest. A flag on the start command carries a default
of its own, and that default answers the key whether a file mentions it or
not. The declaration keeps all state history and the flag prunes.

Every other row is the kind of node. The boot's writer produces one shape
for all of them, so a declaration for a kind that shape does not describe
disagrees with it. A validator and a seed serve no queries and the shape
opens the listeners that do. A full node keeps a hundred thousand blocks
where the shape keeps all of them, and an archive node keeps all state where
the shape keeps the hundred thousand most recent.

Measuring one kind only would have found two of the twelve.

## The keys that cannot be compared

Eleven declared keys are left unanswered by a generated file. Their readers
each start from a default of their own and take the source's answer only
when there is one, so the comparison has nothing to read for them. They are
recorded as a named set rather than compared, and both directions are held:
a key that stops being answered moves into the set, and one that starts
being answered leaves it and belongs in the comparison.

Each of those eleven declarations is taken from the same default its reader
falls back to, so the value a node runs is the declared one. Nothing here
measures that. What reaches it is a writer rendering every declared key into
a file and a node started from that file.

## Verification

The two-boot helper is now shared with the decode-side measurement rather
than held inside it. That measurement is unchanged and passes unchanged,
which is the proof the extraction changed nothing.

Every check was broken on purpose and confirmed to fail for its stated
reason, with the mutation confirmed to compile each time: a dropped row, a
wrong recorded value, an extra row, a dropped kind of node, a row that no
longer disagrees, and both directions of the unanswered set.

Tests only. No production code changes.

### DIFF
--- a/cmd/seid/cmd/node_agreement_test.go
+++ b/cmd/seid/cmd/node_agreement_test.go
@@ -97,16 +97,24 @@ func TestTheDivergencesFromAGeneratedFileAreTheRecordedOnes(t *testing.T) {
 // started with no configuration file of its own.
 func whatTheBootGenerates(t *testing.T) map[string]string {
 	t.Helper()
+	ctx := theBootWithNoFileOfItsOwn(t)
+	return configmanager.DescribeForTest(t, ctx.Config, keysADecodeDelivers())
+}
+
+// theBootWithNoFileOfItsOwn starts a node with no configuration file of its own and no sei.toml, and
+// returns what it holds. Nothing is delivered, so both deliveries read purely what the boot generated.
+//
+// Booted twice, and the second one is returned. The first writes the file, and a flag bound to a key the
+// writer sets overwrites that value before anything reads it back, because the file has not been read yet.
+// From the second boot the file is read and wins, so what a node runs from its second start onward is what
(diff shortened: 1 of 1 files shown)


bdchatham:
Folding this into the generate work, so the record and the writer that depends on it land together.

PR #4093, opened 2026-09-03
balances and bools

cody-littley:

This PR updates FlatKV in two important ways:

- balance EVM keys are now supported
- read API now has an `ok bool` return parameter

### DIFF
--- a/sei-db/state_db/sc/flatkv/store_apply.go
+++ b/sei-db/state_db/sc/flatkv/store_apply.go
@@ -109,8 +109,9 @@ func (s *CommitStore) prepareWrites(
 ) (preparedWrites, error) {
 	var out preparedWrites
 
-	// A nonce or codehash change carries only its own field, so it has to be merged onto the account as
-	// it stands right now — a live read, since anything an earlier call at this height wrote counts.
+	// A nonce, codehash or balance change carries only its own field, so it has to be merged onto the
+	// account as it stands right now — a live read, since anything an earlier call at this height wrote
+	// counts.
 	s.phaseTimer.SetPhase("apply_change_sets_read_accounts")
 	readStart := time.Now()
 	accountOld, err := s.readAccountsForMerge(changesByType)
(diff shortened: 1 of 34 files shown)


yzang2019 (approved):

sei-db/state_db/sc/flatkv/store_read.go
```diff
@@ -55,27 +55,15 @@ func (s *CommitStore) Get(moduleName string, key []byte) ([]byte, bool) {
 		}
 		return value, value != nil
 
-	case keys.EVMKeyNonce, keys.EVMKeyCodeHash:
+	case keys.EVMKeyNonce, keys.EVMKeyCodeHash, keys.EVMKeyBalance:
```
yzang2019, line 58: GetBlockHeightModified and Get disagree about whether a balance key exists.

Get on a balance key for an account whose balance is zero returns (nil, false), but GetBlockHeightModified on the same key returns (height, true).

Is that an issue?

sei-db/state_db/sc/flatkv/store_apply.go
```diff
@@ -125,7 +126,7 @@ func (s *CommitStore) prepareWrites(
 	accountUpdates, err := mergeAccountUpdates(
 		changesByType[keys.EVMKeyNonce],
 		changesByType[keys.EVMKeyCodeHash],
-		nil, // TODO: update this when we add a balance key!
+		changesByType[keys.EVMKeyBalance],
```
yzang2019, line 129: Before this PR the balance bytes were always zero, because we passed nil for the balance argument so "a row exists" implied "nonce ≠ 0 or codeHash ≠ 0". 

This PR replaces that nil with changesByType[keys.EVMKeyBalance] (store_apply.go:129), so a balance write alone now creates a row, and "a row exists" no longer implies anything about the nonce.

Question is will this behavior change cause any issues?

cody-littley:
Marking this PR as a draft until we figure out proper semantics

PR #4106, opened 2026-09-08
Validate block part Merkle proof totals (CON-412)

shemnon:
## Summary
- Reject block parts whose Merkle `Proof.Total` does not match the PartSet size, so a non-power-of-two last leaf cannot be relabelled into an earlier first-writer-wins slot.
- Same guard CometBFT already has in `PartSet.AddPart`; CON-20's index check is not enough because the attacker sets `Part.Index == Proof.Index` on the stolen slot.
- Add a 3-part test where `ValidateBasic` and `Proof.Verify` still succeed, `AddPart` rejects, and the genuine part still fills the slot.

## Test plan
- [x] `go test ./sei-tendermint/types/... -run 'TestAddPartRejectsMismatchedProofTotal|TestBasicPartSet|TestWrongProof|TestAddPartWithSwappedProof'`
- [ ] New test fails without the `Proof.Total` guard and passes with it

### DIFF
--- a/sei-tendermint/types/part_set.go
+++ b/sei-tendermint/types/part_set.go
@@ -302,6 +302,11 @@ func (ps *PartSet) AddPart(part *Part) (bool, error) {
 		return false, nil
 	}
 
+	// The proof should be compatible with the number of parts.
+	if part.Proof.Total != int64(ps.total) {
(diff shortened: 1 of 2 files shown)


shemnon:
operator error

PR #4108, opened 2026-09-09
(Autobahn) Prove commit-committee identity on giga handshake and learn live dial addresses (CON-358)

wen-coding:
## Summary
- Validators sign a tagged giga handshake claim (validator key, Autobahn `SelfAddr`, EVM RPC). No claim, or a claim outside the current commit committee, is served as a fullnode; a verified current member is served as that validator for the life of the socket.
- Inbound advertisements of current members are learned on a live overlay. The configured `ValidatorAddrs` book wins on dial; the overlay only covers members it omits. An unroutable EVM RPC means we learn none of the advertisement. Overlay rows and leaver sessions stay until Anchor is within one epoch of commit.
- Join does not upgrade a block-sync socket. A book-absent member who inbounds too early is recovered by their own redial after `DialInterval`.
- HTTP(S) EVM RPC is checked by `utils.CheckHTTPURL` (config, decode, local offer). Loopback, link-local, and unspecified hosts are refused on the overlay. The local book address must survive `NodeAddress.Validate()` or the validator will not start.

Linear: [CON-358](https://linear.app/seilabs/issue/CON-358)

## Test plan
- [x] `GOWORK=off go test ./sei-tendermint/internal/p2p/ ./sei-tendermint/libs/utils/ ./sei-tendermint/config/ -count=1 -run 'TestHandshake|TestGigaRouter|TestDecodeGigaClaim|TestCheckHTTPURL|TestIsLoopbackOrLinkLocalURL|TestURLUnmarshal'` (32 passed)
- [x] Confirm a book-absent current member who inbounds is dialed back at the advertised Autobahn address (`TestGigaRouterCommon_RunInboundConnLearnsMemberAndClosesOnLeave` learns the overlay row; `validatorAddr` is book-first)
- [x] Confirm a leaver’s inbound closes, and overlay dials stop only after Anchor catches up (`RunInboundConnLearnsMemberAndClosesOnLeave`, `RunPerCommitteeMemberKeepsLeaversWhileAnchorLags`)
- [x] Confirm a giga inbound without a claim is capped block-sync and is not learned as a validator (`TestHandshakeAcceptsFullnodeWithoutClaim`; `acceptInbound` returns None when there is no current-committee claim)
- [x] Confirm localhost / `0.0.0.0` EVM RPC is not learned, and a local `:0` address refuses to start (`AcceptInboundRefusesUnroutableEvmRPC`, `RejectsInvalidSelfAddr`)

PR #4114, opened 2026-09-09
ci: drop unused Chrome/Microsoft apt sources before apt-get update

wen-coding:
## Summary
- GitHub `ubuntu-*` runners ship Google Chrome (and Microsoft) apt sources. `apt-get update` then fetches `dl.google.com` even when the job only installs Ubuntu packages (`build-essential`, RocksDB deps).
- That repo periodically returns a `Packages.gz` whose hash does not match the signed `Release` file, failing Cross-Architecture Build and RocksDB Tests (e.g. attempts on #4108).
- Drop those unused sources before `apt-get update`. We do not install or need Chrome.

## Test plan
- [x] Confirm Cross-Architecture Build Test → Linux AMD64 `Install system dependencies` no longer hits `dl.google.com/linux/chrome-stable`.
- [x] Confirm sei-db → RocksDB Tests `Install RocksDB dependencies` succeeds without hash-sum mismatch.
- [x] Confirm both jobs still install `build-essential` (and RocksDB deps) from Ubuntu.

### DIFF
--- a/.github/scripts/ci-drop-third-party-apt-sources.sh
+++ b/.github/scripts/ci-drop-third-party-apt-sources.sh
@@ -6,20 +6,15 @@
 # `apt-get update` — even one that only wants build-essential — fails the job.
 #
 # Safe on any ubuntu-* runner: every removal is guarded so a change to the
-# runner image layout can never fail the job.
+# runner image layout can never fail the job. Filenames are globbed so a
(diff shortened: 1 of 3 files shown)


wen-coding:
@seidroid re-review please

wen-coding:
@seidroid re-review please

bdchatham (approved):

PR #4119, opened 2026-09-10
Add gigasim for giga storage benchmark

yzang2019:

Adds `gigasim`, an end-to-end benchmark for a Giga node's whole storage stack.

`blocksim` exercises the block ledger alone and `cryptosim` the state DB alone. `gigasim` drives the
block store, the live state DB, the historical state store and the receipt store together, through the
same `GigaStorageManager` a node opens. It is the benchmark to reach for when the question is how the
engines behave *together* — whether pruning, checkpointing or hashing on one store shows up as latency
on another.

Two goroutines split the work the way a node splits consensus from execution: a generator builds each
block and writes it to the ledger, while a consumer executes that block's transactions across a pool
and writes what they produced to the receipt store and the state DB. Every block commits at the same
height in every store, and the staging queue between the two sides bounds how far generation may run
ahead — its depth is the first number to read when interpreting a run, since a queue that stays full
means execution and the state DB are the limit and one that stays empty means generation is.

Execution is simulated rather than real: each transaction replays the reads and writes an ERC20
transfer makes without doing the arithmetic, because what is under measurement is storage traffic
rather than the EVM. That is the same transaction model `cryptosim` uses, so the two stay comparable on
the state DB.

**Included**

- `sei-db/bench/gigasim` — the benchmark, with three shipped configs: `standard.json` (a full node),
  `validator.json` (block store and live state DB only) and `debug.json` (a short local smoke run).
- `docker/monitornode/dashboards/gigasim-dashboard.json` — a Grafana dashboard covering pipeline
  throughput and backpressure, transaction execution phases, the FlatKV commit and read paths, the WAL,
  Pebble internals and host resources. Grafana provisions every dashboard in that directory, so it
  appears without an import step.

The change is additive: nothing outside `sei-db/bench` and the monitoring assets is touched, and no
node or consensus path changes.

- `scripts/ramtest.sh ./sei-db/bench/gigasim/... -race`. The suite covers that every store advances to
  the same height, that disabled stores are never written, that generation runs ahead of execution,
  that a run resumes where the last one stopped, and that an interrupted run drains onto a consistent
  height — plus units for config validation, synthetic receipt hashes and the staged write batch.
- Ctrl-C and resume exercised on the built binary: the interrupted run drains, exits 0, and a following
  run resumes the same data directory rather than refusing it as torn.
- Every panel query in the new dashboard executed against a live Prometheus scraping a running
  benchmark, with the derived panels checked for sane values rather than merely non-empty ones.
- `golangci-lint` v2.8.0 and `make fmtcheck` clean.

Dashboard:
image
image


yzang2019:
@seidroid review

sei-db/bench/gigasim/gigasim.go
```diff
@@ -0,0 +1,651 @@
...
+		if err := cleanDirectories(config); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// cleanDirectories empties the data and log directories, naming each on the console as it goes. It
+// attempts both even when the first fails, and reports every failure.
+func cleanDirectories(config *GigasimConfig) error {
```
blindchaser, line 710: Cleanup accepts any directory without an ownership marker. A root, home, repository, or symlink path can cause major data loss. The debug config also cleans the same directory that the standard config uses.

suggest fix: Require a GigaSim marker, resolve symlinks, reject unsafe paths, and give each shipped config a separate directory.

sei-db/bench/gigasim/account_model.go
```diff
@@ -0,0 +1,279 @@
...
+		a.nextAccountID++
+		if a.rand.Float64() >= a.config.NewAccountDormancyProbability {
+			a.numberOfColdAccounts++
+		}
+		return a.accountAddress(accountID), true, nil
+	}
+
+	// The cold population sits immediately below the newest account that existed before this block, so
+	// the window slides forward as accounts are minted.
+	lastColdAccountID := a.highestSafeAccountID + 1
```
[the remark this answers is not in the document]
blindchaser: new cold and dormant accounts can have mixed IDs

sei-db/bench/gigasim/account_model.go
```diff
@@ -0,0 +1,281 @@
...
+	firstColdAccountID := lastColdAccountID - a.numberOfColdAccounts
+	if firstColdAccountID >= lastColdAccountID {
+		return nil, false, fmt.Errorf("no cold accounts available to select from")
+	}
+	return a.accountAddress(a.rand.Int64Range(firstColdAccountID, lastColdAccountID)), false, nil
+}
+
+// RandomAccountSlot selects one of the ERC20 storage slots an account may touch.
+func (a *accountModel) RandomAccountSlot() []byte {
+	slotID := a.rand.Int64Range(0, int64(a.config.Erc20InteractionsPerAccount)*a.nextAccountID+1)
```
blindchaser: Issue: Slot selection is global. Hot accounts do not produce hot storage slots, so the workload differs from CryptoSim and from the stated ERC20 model.

consider derive each slot from its selected account ID and slot number?

sei-db/bench/gigasim/receipt_writer.go
```diff
@@ -0,0 +1,52 @@
...
+	metrics *GigasimMetrics
+}
+
+// newReceiptWriter wraps the receipt store the storage manager opened.
+func newReceiptWriter(store receipt.ReceiptStore, metrics *GigasimMetrics) *receiptWriter {
+	return &receiptWriter{store: store, metrics: metrics}
+}
+
+// writeBlock stores every receipt a block produced, in one call, as a node does at commit.
+func (w *receiptWriter) writeBlock(number int64, receipts []*evmtypes.Receipt) error {
```
blindchaser, line 41: setup commits many block heights but produces no receipts. writeBlock returns early, so the receipt store does not record these heights. Its latest height remains 0?

sei-db/bench/gigasim/config/standard.json
```diff
@@ -0,0 +1,8 @@
+{
+  "Comment": "Standard gigasim configuration: a full node's stack, with the historical state store and the receipt store both enabled.",
+  "Seed": 1337,
+  "DataDir": "~/gigasim/data",
```
blindchaser: it looks standard.json/validator.json/debug.json are using same DataDir and LogDir. 

the debug.json's cleanDirectories could wipe out DataDir and LogDir without warnings.

how about use different Data/Log folders for each json to avoid override

sei-db/bench/gigasim/state_batch.go
```diff
@@ -0,0 +1,133 @@
...
+	return &b.shards[uint(key[len(key)-1])%batchShards]
+}
+
+// Put stages a write, replacing any earlier write to the same key.
+//
+// Safe to call concurrently with Put and Get, but not with drainToChangeSet.
+func (b *stateBatch) Put(key []byte, value []byte) {
+	shard := b.shardFor(key)
+	shard.mu.Lock()
+	shard.entries[newStagedKey(key)] = stagedWrite{key: key, value: value}
```
blindchaser, line 79: the mutex prevents a data race. It does not enforce transaction order. The last goroutine to get the lock wins. this matters because every transaction writes the fee account, and hot accounts also cause duplicate keys.

so equal seeds can produce different state hashes. it's fine if we don't care about the state hash in this case

sei-db/bench/gigasim/receipt_writer.go
yzang2019: Receipts are generated when we generate block

sei-db/bench/gigasim/gigasim.go
yzang2019: Fixed

sei-db/bench/gigasim/account_model.go
yzang2019: Good call, will make a fix for that

sei-db/bench/gigasim/config/standard.json
yzang2019: Fixed

yzang2019:
@seidroid review

sei-db/common/metrics/queue_meter.go
```diff
@@ -0,0 +1,100 @@
...
+
+// NewQueueMeterFactory creates a factory recording to {prefix}_queue_blocked_seconds_total.
+func NewQueueMeterFactory(meter metric.Meter, prefix string) *QueueMeterFactory {
+	blockedSeconds, _ := meter.Float64Counter(
+		prefix+"_queue_blocked_seconds_total",
+		metric.WithDescription("Seconds producers spent blocked waiting for room on a full queue"),
+		metric.WithUnit("s"),
+	)
+	return &QueueMeterFactory{blockedSeconds: blockedSeconds}
+}
```
cody-littley: Why do we need a factory for this metric type? Not sure how its used, but want to make sure the factory pattern is justified and not just unnecessary complexity.

sei-db/common/metrics/queue_meter.go
```diff
@@ -0,0 +1,100 @@
...
+// scrapes. Blocked time accumulates, so a scrape at any cadence reads all of it.
+type QueueMeter struct {
+	depth          metric.Int64Gauge
+	blockedSeconds metric.Float64Counter
+	attrs          metric.MeasurementOption
+}
+
+// Observe records how full queue is, and must be called before putting anything on it: read after a
+// send, the depth is the queue at its emptiest, a consumer having just made the room the send used.
+func (m *QueueMeter) Observe(queue int) {
```
cody-littley: Unclear from godoc how this is supposed to be used.

sei-db/common/metrics/queue_meter.go
```diff
@@ -0,0 +1,100 @@
...
+// Send puts value on queue, recording the depth the producer found and charging the wait when the queue
+// was full. A send that finds room reads no clock.
+func Send[T any](m *QueueMeter, queue chan T, value T) {
+	m.Observe(len(queue))
+	select {
+	case queue <- value:
+		return
+	default:
+	}
+	_ = m.Blocked(func() error {
```
cody-littley: Both Observe() and Blocked() seem like they are supposed to only be called internally by Send(), but both are exported methods. Would it make sense to instead make those two methods into un-exported methods?

sei-db/common/metrics/queue_meter.go
```diff
@@ -0,0 +1,100 @@
...
+	started := time.Now()
+	err := send()
+	m.blockedSeconds.Add(context.Background(), time.Since(started).Seconds(), m.attrs)
+	return err
+}
+
+// Send puts value on queue, recording the depth the producer found and charging the wait when the queue
+// was full. A send that finds room reads no clock.
+func Send[T any](m *QueueMeter, queue chan T, value T) {
+	m.Observe(len(queue))
```
cody-littley: For high volume queues, it's potentially adding moderate latency by checking the length of the queue at each insertion.

A lower overhead way to monitor queue size is to set up a background thread that wakes up once very few seconds, observes the length, then updates a metric. Prometheus only samples the value once every few seconds anyways, so sampling at each insertion doesn't actually improve the data in the resulting graph.

sei-db/db_engine/litt/disktable/control_loop.go
```diff
@@ -36,6 +37,9 @@ type controlLoop struct {
 	// compression stage in order).
 	inputChannel chan any
 
+	// inputQueue records the time writers spent blocked because inputChannel was full.
+	inputQueue *commonmetrics.QueueMeter
```
cody-littley: nit, perhaps call it `inputChannelMeter`? Calling it `inputQueue` makes me think this is an actual queue, as opposed for something designed to measure channel latency.

sei-db/db_engine/litt/disktable/flush_loop.go
```diff
@@ -22,6 +23,9 @@ type flushLoop struct {
 	// flushChannel is a channel used to enqueue work on the flush loop.
 	flushChannel chan any
 
+	// flushQueue records the time the control loop spent blocked because flushChannel was full.
+	flushQueue *commonmetrics.QueueMeter
```
cody-littley: nit, perhaps call this `flushChannelMeter`?

sei-db/state_db/sc/flatkv/store_write.go
```diff
@@ -124,15 +124,52 @@ func (s *CommitStore) Commit(version int64) (committed int64, err error) {
...
+	window := now.Sub(c.windowStart)
+	if window < commitLogInterval {
+		return
+	}
+
+	logger.Info("FlatKV committing",
+		"version", version,
+		"blocks", c.blocks,
+		"blocks_per_second", int64(float64(c.blocks)/window.Seconds()),
+		"average_commit", (c.spent / time.Duration(c.blocks)).Truncate(time.Microsecond))
```
cody-littley: Anything that logs once per block is going to be quite noisy at giga scale (i.e. 10s to 100s of log entries per second). Is there a reason why we are writing this via logs instead of metrics?

cody-littley (approved):

sei-db/db_engine/litt/disktable/control_loop.go
yzang2019: make sense

sei-db/db_engine/litt/disktable/flush_loop.go
yzang2019: make sense

sei-db/state_db/sc/flatkv/store_write.go
yzang2019: This is actually only logged every interval, but yeah it was more for debugging purpose, we can remove it

sei-db/common/metrics/queue_meter.go
yzang2019: Added a try() before we sample the metrics

sei-db/common/metrics/queue_meter.go
yzang2019: Fixed

sei-db/common/metrics/queue_meter.go
yzang2019: Fixed

sei-db/common/metrics/queue_meter.go
yzang2019: Removed

yzang2019:
@seidroid review

blindchaser (approved):

PR #4130, opened 2026-09-10
Fix flaky TestRateLimitErrorFormat and mock block encode data race

masih:
`TestSimulationAPIRequestLimiter/TestRateLimitErrorFormat` fires 20 concurrent calls and asserts that at least one is rejected by the rate limiter with the expected error text. On an overloaded runner a single burst can be serialized enough by the scheduler that the limiter never trips, so zero errors are observed and the assertion fails. Separately, running the test under `-race` reports a data race on main: `MockClient.mockBlock` calls `Encoder(Tx1)` on every block fetch, and the auth tx wrapper memoizes encoded bytes on first use, so concurrent simulate calls write to the shared `Tx1` concurrently.

The fix retries a synchronized burst against a fresh `testEnv` up to 5 times until a rejection is observed (the pattern `TestDifferentMethodsShareSameLimiter` already uses), and encodes the mock block txs once in fixture setup (`Tx1Bz`, `TxNonEvmWithSyntheticLogBz`) so the mock block only reads immutable byte slices.

Flaked in: https://github.com/sei-protocol/sei-chain/actions/runs/34329926916/job/102395953394

### DIFF
--- a/evmrpc/simulate_test.go
+++ b/evmrpc/simulate_test.go
@@ -971,29 +971,44 @@ func TestSimulationAPIRequestLimiter(t *testing.T) {
 	})
 
 	t.Run("TestRateLimitErrorFormat", func(t *testing.T) {
-		tEnv := newTestEnv(t)
-		// Test the error message format by overwhelming the rate limiter
-		const numRequests = 20
-		results := make(chan error, numRequests)
-		start := make(chan struct{})
-		var wg sync.WaitGroup
+		// Test the error message format by overwhelming the rate limiter.
+		// A single burst can occasionally avoid contention on overloaded CI workers,
+		// so retry a synchronized burst a few times.
+		const (
+			numRequests = 20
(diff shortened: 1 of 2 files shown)


evmrpc/simulate_test.go
```diff
@@ -971,29 +971,44 @@ func TestSimulationAPIRequestLimiter(t *testing.T) {
...
 		}
-		close(start)
-		wg.Wait()
-		close(results)
 
 		var rateLimitErrors []error
-		for err := range results {
-			if err != nil && strings.Contains(err.Error(), "rejected due to rate limit") {
-				rateLimitErrors = append(rateLimitErrors, err)
+		for attempt := 1; attempt <= maxAttempts; attempt++ {
```
masih: Use for range style

alexander-sei (approved):

PR #4132, opened 2026-09-10
docs(evmonly): document Autobahn cluster workflows

codchen:

- expand the Autobahn EVM-only README with copy-pasteable local and AWS cluster lifecycle commands
- document deployment initialization, genesis validator counts and power, EVM execution defaults, and the current `seid start` node entrypoint
- add a checked-in four-endpoint `sei-load` profile and explain workload types, TPS/workers/accounts tuning, transactions-per-block estimates, and producer-side limits
- document the deliberately narrow EVM JSON-RPC surface, verified `cast receipt` usage, and the RPC gaps that limit other `cast` workflows
- document local and AWS teardown behavior

The EVM-only node remains part of `seid`; this PR does not add a standalone `giga` executable.

- `scripts/ramtest.sh ./cmd/autobahn-e2e/... -count=1`
- `jq empty integration_test/autobahn/sei-load.local.json`
- `git diff --check`
- built `autobahn-e2e`, deployed the four-validator cluster locally, and verified all four live node processes were `seid`
- inspected generated genesis and Autobahn configuration: four genesis validators/gentxs, four active Autobahn validators, 35,000,000 block gas, 2,000 transactions per block, 400 ms interval, and empty blocks disabled
- installed `sei-load@v0.0.1` with the documented command and ran the checked-in profile against all four endpoints: 4,302 transfers in 17 seconds at 248.75 average TPS
- published an offline-signed transfer with `cast publish` and fetched its successful finalized receipt with `cast receipt`
- verified `eth_blockNumber` returns JSON-RPC `-32601`, matching the documented `cast` gap
- listed the live cluster at height 177, then ran the documented teardown and verified the containers and cluster metadata were removed

`make fmtcheck` was also attempted, but the local golangci-lint executable was built with Go 1.24 and refuses to load this repository's Go 1.25.6 configuration. The final branch differs from `main` only in the README and checked-in JSON load profile.


masih (changes requested):
Discussed internally, we need to drop the bonus binary to avoid confusion on how to run a giga node for internal testing. Because the bonus binary introduces a secondary way of doing the same thing via the regular seid with env vars set.

masih (approved):

masih:
@seidroid review

PR #4134, opened 2026-09-10
Assert usei supply as deltas in mint and gov integration tests

masih:
`mint_test.yaml` and `gov_proposal_test.yaml` asserted the absolute `usei` total supply (`5000000000333333333333`, genesis plus exactly one scheduled mint), and the mint suite also pinned single-mint minter state (`REMAINING == 666666666666`, `START_DATE == LAST_MINT_DATE`, `LAST_MINT_HEIGHT == 0`). The localnet genesis carries a 3-day release schedule and `x/mint` mints again on the first block after a UTC date change, so any cluster that lives across midnight failed both suites, and packing suites into shared rows (#4133) widens that window since the assertions can now run several minutes after boot.

The mint suite now derives `MINTS_SO_FAR = (total - remaining) / last_mint` and checks the minter state and total supply against that (`1..3` mints, each `333333333333`, remaining and supply consistent with the count, last mint date inside the schedule), so it holds on any day of the schedule. The gov burn case checks `before - after == 20000000` (the two 10M deposits) instead of two absolute values, and the unused `NEW_PARAM` query in that case is dropped. Both sets of expressions were checked against the yaml runner's left-to-right big-int arithmetic for the one-mint and two-mint cases.

Spun out of the review on #4133.

### DIFF
--- a/integration_test/gov_module/gov_proposal_test.yaml
+++ b/integration_test/gov_module/gov_proposal_test.yaml
@@ -142,8 +142,6 @@
     - cmd: seid q bank total --denom usei --output json | jq -r .amount
       env: TOTAL_SUPPLY_AFTER_BURN
   verifiers:
-    # Check if the total supply is reduced or not to verify token burns
+    # The rejected proposal burns both deposits (10000000usei initial + 10000000usei top-up)
     - type: eval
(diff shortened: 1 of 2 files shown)


masih:
@seidroid review

masih:
@seidroid review

kollegian (approved):

PR #4138, opened 2026-09-10
(Autobahn) Cover timeout-vote aggregation (CON-423)

wen-coding:
## Summary
- Rename the lock-held timeout ingest path to `pushVerifiedVote` and drop the unused verify TODO. `PushTimeoutVote` still verifies before taking the lock.
- Add `timeoutVotes` unit tests for QC formation, same-view leftover votes, and stale/newer vote replacement.
- Add `PushTimeoutVote` tests for quorum ingest plus outsider and wrong-epoch rejection.

## Test plan
- [x] `go test github.com/sei-protocol/sei-chain/sei-tendermint/internal/autobahn/consensus -count=1 -run 'TestTimeoutVotes_|TestPushTimeoutVote_'`

### DIFF
--- a/sei-tendermint/internal/autobahn/consensus/state.go
+++ b/sei-tendermint/internal/autobahn/consensus/state.go
@@ -221,7 +221,7 @@ func (s *State) PushTimeoutVote(vote *types.FullTimeoutVote) error {
 		return fmt.Errorf("vote.Verify(): %w", err)
 	}
 	for tv := range s.timeoutVotes.Lock() {
-		tv.pushVote(ep.Committee(), vote)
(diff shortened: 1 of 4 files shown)


shemnon (approved):
LGTM.

state.go and timout_votes.go are only non-functional changes, do they need to stay?

wen-coding:
> LGTM.
> 
> state.go and timout_votes.go are only non-functional changes, do they need to stay?

Yeah, I think I probably should keep the rename, otherwise bots will keep asking "were the votes verified"?

codchen (approved):

PR #4144, opened 2026-09-11
Shard the race-detection run so the merge queue is no longer bound by it

masih:
The merge queue's clean-group floor is about 19–20 minutes, and it is set by a single job: `Race Detection` in the Go Test workflow. It runs `go test -race` over the whole tree in one invocation (~15.7 min wall-clock across roughly 84 package-minutes of test time) and then runs `make upgrade-test` and `make upgrade-test-vet` serially on top (~2.8 min), with no build cache. Every other required workflow finishes in 12–14 minutes or less, so this one job decides how fast a group can land.

This splits the race run into four matrix shards and moves the upgrade tests into their own job. Packages are assigned round-robin over the sorted list rather than in contiguous slices: the slow packages cluster under `sei-db/*` and `sei-cosmos/x/*`, and a contiguous split from the last green run would have put 32 minutes of test time on one shard and 8 on another, whereas round-robin lands at 18–24 minutes each. Coverage is unchanged: the same packages run with the same `-race -tags=ledger,test_ledger_mock` flags, and `state_db` is still excluded because `sei-db-tests.yml` owns it. The shards and the upgrade job roll up into a job that keeps the `Race Detection` name, so the required-check configuration for branch protection and the merge queue does not need to change.

### DIFF
--- a/.github/workflows/go-test.yml
+++ b/.github/workflows/go-test.yml
@@ -20,7 +20,7 @@ env:
   STATE_DB_PKG_PREFIX: github.com/sei-protocol/sei-chain/sei-db/state_db
 
 jobs:
-  # The race run fans out over RACE_SHARDS runners. The shards and the upgrade
+  # The race run fans out over the shard matrix. The shards and the upgrade
   # tests all roll up into the `Race Detection` job below, which is the single
   # check name branch protection and the merge queue require.
   race-shard:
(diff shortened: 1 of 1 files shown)


masih:
@seidroid review

alexander-sei (approved):

cody-littley (approved):

PR #4145, opened 2026-09-11
fix(flatkv): retain 72 checkpoints instead of mirroring memIAVL's count

blindchaser:
## What this changes

FlatKV stops taking its retention count from memIAVL and keeps its own default of
72 old checkpoints. The snapshot *interval* is still mirrored.

| | before | after |
|---|---|---|
| FlatKV checkpoints retained | 1 | 72 |
| Guaranteed reach | 10,000 blocks (74 min) | 720,000 blocks (~89 h) |
| Extra disk at mainnet state size | — | ~20 GiB |

## Why

`alignFlatKVSnapshotWithMemIAVL` overwrote *both* of FlatKV's snapshot knobs from
memIAVL's `sc-*` keys, and `sc-keep-recent` defaults to 1. FlatKV therefore kept one
old checkpoint. `ExternalPruning` is false on the composite path, so nothing else
extended that.

Below FlatKV's reach, `migrate-evm-status`, `dump-flatkv` and a cross-backend digest
cannot open a past version, and a rollback has no base snapshot to rewind to. The EVM
migration window became unobservable 74 minutes after passing.

The shared knob cannot express the setting we want, and this is not a matter of cost.
Measured on the mainnet-sized `pacific-1` shadow node, post-`H_done`, at the tip:

| | one further retained version |
|---|---|
| memIAVL | **56,782 MiB** (snapshots are independent full copies) |
| FlatKV | **~286 MiB** (261 MiB of pinned SSTs + ~25 MiB of state WAL) |

A Pebble checkpoint hardlinks its SSTs, so an old checkpoint only pins the bytes
compaction has since obsoleted. The factor is about 200. At the depth FlatKV wants,
the memIAVL side of a shared count would need 3.68 TiB on a 2 TiB volume, so the
knob has to be split before any number can be chosen.

## Why 72

72 checkpoints at the 10,000-block interval is a guaranteed reach of 720,000 blocks,
about 89 hours at the shadow node's measured 2.247 blocks/s, against an 85-hour drain
at `NumKeysToMigratePerBlock = 1024`. It covers the whole migration window with margin.

The cost is linear in depth, not sublinear: each older checkpoint pins exactly the
files obsoleted during its own interval, and those sets are disjoint by construction.
It only saturates at ~93 checkpoints, where the pinned bytes reach the 24,241 MiB of
live SSTs.

## How operators get it

The production store is built by `parseSCConfigs`, which starts from the in-code
defaults and reads no FlatKV retention key, so raising the default reaches every
existing node with no `app.toml` edit. Nodes carrying a stale explicit
`state-commit.flatkv.snapshot-keep-recent` from the pre-#3745 template are also
unaffected, because that key never reaches a store.

Making the key actually readable is a deliberate follow-up rather than part of this
change, so the v6.7 backport carries no new config read.

## Testing

- `TestAlignFlatKVSnapshotIntervalWithMemIAVL` replaces the test that pinned the
  both-values mirror; it now pins that the interval is mirrored, a zero interval is
  still healed before assignment, and neither a set nor a zero memIAVL keep-recent
  reaches FlatKV.
- `TestFlatKVDefaultRetentionSpansTheMigrationWindow` guards the reach itself, so a
  later depth change that no longer spans the window fails rather than passing quietly.
- `./sei-db/state_db/sc/composite/`, `./sei-db/state_db/sc/flatkv/...`,
  `./sei-db/config/`, `./sei-cosmos/server/config/`, `./app/` all pass.
- `make fmtcheck` and `make dblint` are clean.

## Backport

`release/v6.7` is where this is needed, and it is filed separately because `main` has
since removed the golden config suite and the FlatKV default constants, so the two
diffs are not identical. `v6.7.0-rc1` is already cut, so the backport implies an rc2.


blindchaser:
Superseded by #4158, which lands the same fix on `release/v6.7` with a retention of 10 rather than 72. Closing rather than retargeting so the depth decision lives in one place.

PR #4152, opened 2026-09-11
(Autobahn) Share vote aggregation (CON-423)

wen-coding:
## Summary
- Fold prepare/commit/timeout vote aggregation into one weighted aggregator and one phase adapter. The three phases still differ in bucket key (vote hash vs view), QC constructor, and quorum getter; they no longer duplicate the latest-vote-per-key / monotonic-QC loop.
- Shared tests cover below-quorum, quorum emit, ignore equal/stale, and replace-at-newer-view. Each phase keeps a QC-validity test plus the property that is actually its own: conflicting prepare/commit proposals do not form a QC; timeout votes with differing prepare-QC payloads still do. Same-view votes after a QC at a non-genesis view leave the published QC unchanged.
- Ingest is unchanged: verify outside the lock, then insert. Published QC behavior matches main (`qc.Load()`, then `New*QC`, then `Store`).

## Quorum wiring
`PrepareQuorum`, `CommitQuorum`, and `TimeoutQuorum` are already the same function on main (`totalWeight - Faulty()`; prepare/timeout/app return `CommitQuorum()`). The phase tests size the committee with `CommitQuorum()` and do not pin which getter is wired into each adapter.

If those thresholds diverged later, swapping the wrong getter would not always fail this suite (a too-strict aggregator in particular would still form a QC). We are not adding two-sided per-phase quorum tests for that: we do not expect the getters to diverge in the near future. `QC.Verify` would still catch the too-loose direction.

## Test plan
- [x] `GOWORK=off go test github.com/sei-protocol/sei-chain/sei-tendermint/internal/autobahn/consensus -count=1`
- [x] `make fmtcheck`
- [ ] CI unit / race on this package


shemnon (approved):
LGTM

Non-blocking design question, why a struct of function pointers and not an interface with three impls?  Is this for performance or code size?

wen-coding:
> Non-blocking design question, why a struct of function pointers and not an interface with three impls? Is this for performance or code size?

@shemnon neither. The three vote types run the same loop; they just pull key/view/bucket off the message differently (prepare/commit bucket by hash, timeout by view). A struct of those extractors was the smallest way to share the loop without three new types.

philipsu522 (approved):

PR #4156, opened 2026-09-11
Add JSON output to evm-logical-digest and let inspect read a drain in flight

blindchaser:
## Summary

- add JSON output for both digest and inspect reports, with progress and warnings on stderr
- let inspect mode read the mid-migration composite EVM view and semantic memiavl replay, so operators can shard mismatched EVM state during a FlatKV drain
- keep main-only EVM balance handling intact; the v6.7 backport can resolve that release-branch difference

## Tests

- `GOPROXY=direct GOTOOLCHAIN=auto go test ./sei-db/tools/cmd/seidb/operations -count=1`
- `GOPROXY=direct GOTOOLCHAIN=auto make fmtcheck`


blindchaser:
Thanks — three reviewers independently found the composite census bug, and it reproduced exactly as described. Pushed 6e591a17d and 04e04227b.

### The composite census (codex P2, seidroid, Bugbot medium)

Confirmed all three legs: `consumeCompositeMemiavl` reached the row counters through `&d.census`, finalization passed `nil`, and the text guard `!= (evmZeroCensus{})` then let the block print with `zero_accounts=0` beside real row counts. seidroid's point that `&d.census` is not the fix on its own is also right — `mergeCompositeFlatKVAccount` never sets `codeHashRow`, so every FlatKV-sourced account would have landed in `LiveAccountsWithoutCodeHashRow`.

Rather than patch the two call sites, the census now lives in one field both levels read (`census *evmZeroCensus`), nil on the paths that take none. The two call sites pass the same field, so a half-measured census is unreachable instead of merely fixed here. Composite takes none, and the reason is recorded at the declaration. `TestCensusIsAllOrNothingAcrossBothCounterLevels` pins it.

### `zero_census` as a measured zero

Same change: the JSON field is now `*evmZeroCensus` with `omitempty`, and the prose guard is `!= nil`. An absent object and an absent prose block now mean the same thing, where before FlatKV and translator mode emitted six zeros in JSON and nothing in text — the drift this PR claims to prevent, so thank you for catching it on the PR that makes the claim. `TestCensusFreeDigestReadsAsUnmeasuredInBothForms` pins it.

### The `digestSink` godoc (codex P1)

Agreed, and rewritten from scratch per the convention rather than patched. The reason the narration is redirected rather than silenced moved to `enterJSONMode`.

### `TestJSONWithInspectBucketIsRefused` (seidroid)

Correct, and a good catch: the assertion would have stayed green off the `--memiavl-open-mode` refusal, which also names `--inspect-bucket`. Now asserts on `--json`.

### One I did not take: refusing `--json` when `SEI_LOG_OUTPUT` points at stdout

Warning on stderr instead. A healthy run does not need the variable — raising seilog to error level already clears every line such a run emits, measured as stdout being exactly the object either way — so refusing would convert an unset variable into a failed read at the moment someone is reading digests *because* something is wrong, including the runbook's own invocations. The warning names the condition and says that a report which fails to parse is this rather than a digest defect. Two tests pin it, one of them that the warning itself stays off stdout, since emitting it onto the report would be the fault it reports.

`make fmtcheck` clean; 54 tests pass in the package with no skips.

blindchaser:
Verified the `zero_census` fix on a live mainnet-sized store rather than only in tests.

**Pod** `eng-yiren/pacific1-flatkv-shadow-v67-migrating-0`, context `harbor`, `seidb` built from `04e04227b` as `CGO_ENABLED=0 GOOS=linux GOARCH=amd64`. Composite backend, `--memiavl-open-mode snapshot`, `--height 0`, over 707,473,660 logical EVM keys. Both readings exit 0, about 13 minutes each.

| case | exit | stdout lines | stdout bytes | `zero_census` |
|---|---|---|---|---|
| before the fix, 2026-09-01 | 0 | 1 | 1184 | present, six zeros |
| after the fix | 0 | 1 | **1065** | **absent** |

`grep -c zero_census` over the captured stdout returns `0`, and the prose run at the same settings carries no `Zero-value memiavl census` line. So the two forms now agree on what "not measured" looks like, which is the drift the shared `report()` value exists to prevent. The object is 119 bytes smaller.

The new `SEI_LOG_OUTPUT` warning landed as the first line of stderr while stdout stayed exactly one JSON object on one line — the property that warning protects.

Two honest limits on this run, both recorded in the evidence note:

Text and JSON were **not** shown here to carry the same numbers. Both used `--height 0`, which resolves to the current version, and the node committed 1,686 versions during the 13.5 minutes between the two starts, so the readings are of different states. Their structure and non-digest fields do agree — same `backend`, `mode`, `db_dir`, `normalization`, the same single `marker_adjustments` entry, no census on either side. The numeric equality is pinned deterministically by `TestDigestJSONReportCarriesTheSameNumbersAsTheProse`, which renders both forms from one `report()`, so it cannot drift.

The partial-census case this PR also closes is **unreachable on this node**. A composite digest scans memIAVL leaves only while `boundary.Status() != MigrationComplete`, and this node is past `H_done`, so the row counters are never raised and the half-measured census cannot appear. Covering that needs a node with a drain in flight.

PR #4157, opened 2026-09-11
Add runtime-selected 16-lane SHA-256 backend for Merkle hashing

masih:
Tendermint's Merkle hashing (`merkle.HashFromByteSlices`: tx hashes, part sets, commit signatures, results, validator sets) hashes every leaf and every tree level as independent SHA-256 calls, which is the batch shape a multi-lane kernel wants. `crypto/sha256` already uses single-lane SHA-NI, so the achievable win is smaller than for LtHash (#4151) and had to be measured rather than assumed.

This adds `sei-tendermint/crypto/tmhash` with the same runtime-selected backend pattern as #4151. `SumBatch(prefix, msgs, out)` is served by a default backend (a reused `sha256.New()`, always compiled) or, under `GOEXPERIMENT=simd` on a CPU with AVX-512F/VBMI/VBMI2, by a generated 16-lane `archsimd.Uint32x16` SHA-256 kernel that loads, prefixes, pads and transposes sixteen messages in-register and falls back to scalar for remainders and mixed lengths; `SEI_TMHASH_BACKEND=default` pins the portable path. `HashFromByteSlices` keeps its signature and, when a multi-lane backend is active and there are at least sixteen leaves, builds the tree level by level, pairing adjacent nodes and carrying an odd trailing node up, which is the RFC 6962 `getSplitPoint` shape. A differential test checks totals 1 to 130 against the recursive implementation, and the tmhash tests check every backend against `crypto/sha256` across block and padding boundaries, so the output is byte-identical.

Two Go 1.27 findings are handled here. The compiler never emits `VZEROUPPER` after `archsimd` code, so the legacy-SSE SHA-NI path that followed ran several times slower with dirty ZMM state; the SIMD backend calls a one-instruction assembly `vzeroupper` before handing off. Separately, with `GOEXPERIMENT=simd` on an AVX-512 machine the runtime's async preemption restores the ZMM registers without `VZEROUPPER`, which slowed all SHA-NI code in the same binary 2 to 4x; the CI job therefore takes the default column from a plain build and the SIMD column from the experiment build. That second effect applies to every legacy-SSE path in the process and should weigh on any decision to ship a `GOEXPERIMENT=simd` binary.

On an Intel Xeon Platinum 8559C (benchstat, n=8) the kernel is 2.0x faster than SHA-NI on 1024 inner nodes (124 µs to 61 µs), 1.6x on 256-byte leaves and 1.3x on 1 KiB leaves; the whole 1024 x 32-byte-leaf tree goes from 223 µs to 122 µs. This is a per-block cost of a few thousand hashes, so the node-level effect is modest. The `SIMD hash backends` workflow runs both packages' tests with and without the experiment and posts the benchstat table as a job summary and PR comment.


.github/workflows/simd-hash-bench.yml
```diff
@@ -1,15 +1,19 @@
-name: LtHash backends
+name: SIMD hash backends
```
wen-coding, line 1: The filename is still lthash-bench.yml

Do you need another file for SIMD hash?

.github/workflows/simd-hash-bench.yml
```diff
@@ -1,15 +1,19 @@
...
 on:
   workflow_dispatch:
   pull_request:
     paths:
       - 'sei-db/state_db/sc/flatkv/lthash/**'
+      - 'sei-tendermint/crypto/tmhash/**'
+      - 'sei-tendermint/crypto/merkle/**'
```
wen-coding, line 7: nit: do these need to be sorted?

sei-will (approved):

sei-tendermint/crypto/tmhash/backend_simd_amd64.go
```diff
@@ -0,0 +1,258 @@
...
+//go:generate go run gen_sha256x16.go
+
+// vzeroupper clears the upper halves of the vector registers.
+//
+//go:noescape
+func vzeroupper()
+
+const (
+	simdBackendName = "simd"
+	simdLanes       = 16
```
wen-coding: why 16?

wen-coding (approved):

PR #4160, opened 2026-09-14
Clamp Tendermint block pruning to evidence MaxAge bounds (CON-419)

wen-coding:
## Summary
- Cap `ResponseCommit.RetainHeight` against `EvidenceParams.MaxAgeNumBlocks` and `MaxAgeDuration` before Tendermint prunes block and state metadata.
- Steady-state path is two metadata lookups from store base; a binary search covers the first prune after a long gap.
- Local disk retention only; app hash is unchanged.

## Test plan
- [x] `go test ./sei-tendermint/internal/state/ -run 'TestPrunableHeight|TestPruneBlocks' -count=1`
- [x] `make fmtcheck`


sei-will (approved):

shemnon (approved):
LGTM

masih (approved):
Thanks for knocking this one out quickly Wen 🙌

PR #4161, opened 2026-09-14
Emit OCC stats from the Giga EVM-only executor (PLT-1273)

amir-deris:
## What

The Giga EVM-only executor already computed `OCCStats` per block and threw them away. This emits them on the global meter, so optimistic concurrency behavior is visible on a dashboard instead of only in a struct nobody reads.

Five instruments, under the node's `sei_chain` Prometheus namespace (`giga_occ_blocks_total` scrapes as `sei_chain_giga_occ_blocks_total`):

| Instrument | Labels | What it answers |
| --- | --- | --- |
| `giga_occ_blocks_total` | `outcome` = parallel \| fallback \| sequential | How often does optimistic execution pay off? |
| `giga_occ_fallbacks_total` | `reason` | Why did a block abandon it? |
| `giga_occ_reruns_total` | — | How much work is validation throwing away? |
| `giga_occ_conflicts_total` | `access`, `kind` | Which access and state kind is conflicting? |
| `giga_occ_rerun_depth` | — | Is one transaction reran deeply, or many reran once? |

Every block records an `outcome`, including blocks that never attempted optimistic execution, so the fallback rate has a full denominator.

## Cardinality

Label vocabularies are closed at build time. `occReason` maps a fallback reason onto a fixed set and collapses anything it does not recognize to `unknown`, so a reason added later cannot open a series nothing queries. Conflict addresses and slots are deliberately kept off the labels — they are per-contract and per-slot and would be unbounded; they stay in `OCCStats.ConflictSamples` for logs and spans.

## MaxIncarnation

`RerunCount` is block-wide and cannot distinguish one transaction rerun five times from five transactions rerun once — two very different signals. `occValidationResult` now tracks `maxIncarnation` alongside it, and `giga_occ_rerun_depth` reports the deepest incarnation a single transaction reached. Buckets are one per reachable incarnation, `0` through `occMaxTxIncarnations-1` (a rerun past that limit falls back instead), so blocks with no reruns are counted apart from blocks that reran a transaction once.

## Safety

`recordOCCStats` runs inside block execution, so it recovers from any telemetry panic rather than taking the block down with it.

## Testing

`giga/evmonly/metrics_test.go` reads instruments back through a manual reader and covers each outcome, that conflict labels carry no address or slot, that an unrecognized reason collapses to `unknown`, and that every attempted block records a rerun depth.


amir-deris:
@seidroid review

masih (approved):

Issue #4169, opened 2026-09-15
evmonly: the app keeps its committed height in memory, so any restart after block 1 crashloops forever


bdchatham:
## Summary

Give `evmOnlyApplication` a durable execution cursor so an `evmOnly` Autobahn node survives a restart. Today the cursor (`committedHeight`, `appHash`, `parentHash`) lives only in `evmOnlyState`, so every restart after block 1 reports `Info().LastBlockHeight == 0`, re-runs `InitChain`, and replays block 1 against a state store that already recovered to height N → `nonce too low` panic, forever.

Source: [sei-chain#4169](<https://github.com/sei-protocol/sei-chain/issues/4169>) (Linear mirror: SEI-10400). Blocks restart/chaos testing of Autobahn + Giga on harbor.

## Root cause (verified against `main` @ `9435e24`)

| Piece of state | Where it lives | Durable? |
| -- | -- | -- |
| EVM state (nonce/balance/code/storage) | flatkv via `StateDB.CommitStateChanges`, called **inside** `FinalizeBlock` (`giga/evmonly/giga_store.go`) | yes — WAL + `OpenDBWithRecovery` converges state/receipts/block store to one height on open |
| receipts | receipt store, same call | yes |
| `committedHeight`, `nextHeight` | `evmOnlyState` struct fields, advanced in `Commit` | **no** |
| `appHash` (chained sha256 of `prev ‖ height ‖ blockHash ‖ gasUsed ‖ changeset`) | `evmOnlyState.appHash` | **no** |
| `parentHash` (Autobahn header hash fed to the EVM as `ParentHash`) | `evmOnlyState.parentHash` | **no** |
| `gasLimit`, `executor` | set only in `InitChain` from `ConsensusParams.Block.MaxGas` | **no** — never rebuilt on restart |

`giga_router_common.go:runExecute` trusts `app.Info()`: `last == 0` → `InitChain` + start at `GenDoc.InitialHeight`. Its comment ("re-entering on restart is safe — nothing was committed") assumes `Info()` is durable, which holds for `BaseApp` (`cms.LastCommitID()`) but not here. The `seedInitialStateVersion` guard returns early when `initialHeight == 1`, so the "state already at height %d" refusal never fires at the default initial height.

[...]

PR #4170, opened 2026-09-15
fix(evmonly): admit on the price block validity charges

### LINKED ISSUE #4168: evmonly: one transaction with a low tip cap kills every validator, deterministically
On an `evmOnly` Autobahn chain, a single valid EIP-1559 transaction with a low priority fee is admitted by `CheckTx` and then panics every validator at once. Observed on a 4-validator harbor chain at `3cad9bf6`: all four died within 2 ms of each other on the same transaction, and the chain never recovered.

```
panic: critical task failed: name=giga, service=Node: execute: r.executeBlock(410):
app.FinalizeBlock(): execute tx 0 0xc3f3faa5…: insufficient gas price [recovered, repanicked]
```

The chain had committed 409 blocks before this.

## Mechanism

Two gates enforce the same policy constant against **different quantities**.

`evmOnlyMinGasPrice = 1_000_000_000` (`sei-tendermint/internal/evmonlyapp/app.go:25`) feeds both:

- **Admission** — `app.go:201` compares `tx.GasPrice()`. For a `DynamicFeeTx` go-ethereum returns the **fee cap**.
- **Execution** — `giga/evmonly/executor.go:254` compares `effectiveGasPrice(tx, baseFee)`, which for a dynamic-fee tx is `baseFee + tx.EffectiveGasTipValue(baseFee)` (`executor.go:442-450`). `evmonlyapp` always passes `BaseFee: new(big.Int)` — zero (`app.go:264`) — so this reduces to the **tip cap**.

So a transaction with `feeCap = 100 gwei` and `tipCap = 0` passes admission (100 ≥ 1) and fails block validity (0 < 1).

The transaction that did this was legitimate: a zero priority fee is valid EIP-1559, and sei-load's own `evmtransferfast` scenario sends one deliberately.

## Why it is fatal rather than a failed receipt

The executor is right that this is block-validity policy, and says so at `executor.go:251-252`:

> `MinGasPrice is block-validity policy; unlike EVM call failures, it does not produce a receipt for an otherwise invalid block.`

But `evmOnlyABCIResults` hardcodes `Code: abci.CodeTypeOK` for every transaction (`evmonlyapp/app.go:305-316`), so the app has no way to express a failed one. The error returns from `FinalizeBlock` as a plain Go error, through `giga_router_common.go:251` and `:464`, into `SpawnCritical` (`node/node.go:674`), which panics (`libs/service/service.go:182`).

The fail-stop is correct for safety — the error returns before `commitAppHashToVault` and before `Commit`, so no app hash is externalised and no divergence is possible. The cost is that the block is byte-identical on every validator, so the failure is perfectly correlated and no quorum survives.

`CheckTx` already carries the known gap at `app.go:163-164`:

> `// TODO(evmonly-production): close the gap between admission and block validity`
> `// before accepting arbitrary traffic; this test app assumes executable load-test transactions.`

A managed CRD (`SeiNetwork.spec.consensus.evmOnly`) now points real traffic at this app, so the assumption no longer holds.

## Reproduction

One transaction, no load generator. An `Autobahn` + `evmOnly: true` chain, any replica count, no `configValues` needed. Send an EIP-1559 transfer with `feeCap = 100 gwei`, `tipCap = 0`. `CheckTx` accepts it; the node panics on the block that includes it.

This is a unit test rather than a cluster test: drive `evmonlyapp.FinalizeBlock` with that transaction and assert the returned error.

## Suggested fix

Make admission enforce the predicate block validity actually uses, from **one** definition — the drift between two copies of the same policy is the root cause.

`giga/evmonly/executor.go:442`'s `effectiveGasPrice` is that predicate. Exporting it and calling it from `parseTx` with the same zero base fee the app passes at execution closes the gap by construction. It has 3 in-package call sites plus one test.

Note it is **not** interchangeable with `giga/deps/xevm/types/ethtx/utils.go:13`'s `EffectiveGasPrice`: the executor's version branches on transaction type and uses `EffectiveGasTipValue`.

A per-transaction failure channel in `evmOnlyABCIResults` would be the deeper fix, and would also cover the next admission/validity divergence. That is a larger behaviour change and worth its own decision.

## Related

The same incident left the chain unable to restart, which is a separate defect in the same file — filed alongside this one.

## Not the cause

The chain carried the Giga `configValues` recipe (`giga_executor.*`, `state-commit.sc-write-mode=test_only_dual_write`, `state-store.evm-ss-split=true`, `receipt-store.rs-backend=pebble`). All of it is inert on this path: `sei-db/config/giga_config.go:11-12` states `GigaStorageConfig` is not read from `app.toml`, and `giga/evmonly/storage_manager.go:12-20` builds it from `DefaultGigaStorageConfig` + `WithValidatorMode()`. Ruled out, so nobody re-investigates it.

bdchatham:
Fixes #4168. One transaction with a zero priority fee was admitted by `CheckTx` and then panicked all four validators of a harbor chain, deterministically.

**Not app-hash breaking.** `a.parseTx` has one caller, `CheckTx`; `FinalizeBlock` never touches it. The export is a rename, and `evmOnlyBaseFee()` returns `new(big.Int)` — bit-identical to the inline value. Same block, same app hash, no coordinated upgrade.

## The mismatch

`evmOnlyMinGasPrice` feeds both gates, which compared different quantities:

| | Compared | `feeCap 100 gwei, tipCap 0` |
|---|---|---|
| `parseTx` | `tx.GasPrice()` — the **fee cap** | 100 gwei → admitted |
| `executor.go` | effective gas price at `BaseFee: 0` — the **tip** | 0 → refused |

A zero priority fee is valid EIP-1559, so an ordinary client reaches this.

An executor refusal is not a failed receipt: `MinGasPrice` is block-validity policy, `evmOnlyABCIResults` hardcodes `CodeTypeOK`, and the error reaches `SpawnCritical` and panics. The block is byte-identical everywhere, so no quorum survives.

## What changed

Export the executor's own predicate rather than copy it — two copies of one fee policy was the root cause. It is **not** interchangeable with `ethtx.EffectiveGasPrice`: this one branches on transaction type and uses `EffectiveGasTipValue`. The base fee gets one definition for the same reason.

## Limits

This keeps a patched node's mempool from feeding one in. It does **not** make the executor tolerant, so an unpatched proposer can still include such a transaction and panic every node. The per-transaction failure channel in `evmOnlyABCIResults` is the real fix; it changes ABCI behaviour and deserves its own decision. #4169, the non-durable height, is untouched.

## Tests

`TestEVMOnlyApplicationRefusesATxTheExecutorWouldRefuse` fails on the old gate — verified by reverting the comparison. A second test guards the other direction, that dynamic-fee transactions clearing the minimum are still admitted.

`go test ./sei-tendermint/internal/evmonlyapp/` and `./giga/evmonly/...` pass; `go vet` clean.

If someone is already on #4168, close this in favour of theirs — the tests are worth keeping either way.


shemnon (approved):
LGTM

shemnon (approved):
LGTM

PR #4175, opened 2026-09-15
ci(giga-1): publish a seid image on every push to giga-1

bdchatham:
`nightly-ecr.yml` checks out a hardcoded `ref: main` — the step is even named
"Checkout main HEAD" — so it can never build `giga-1` however it is invoked, and
`ecr.yml` is branch-filtered to `main`/`release/**`. `giga-1` has no publishing
workflow at all. This adds one.

**Land this first.** It is the producer for
sei-protocol/platform#1707, which deploys these images to a new
4-validator `giga-testnet` chain on harbor. That PR's `spec.image` carries an
all-zero placeholder tag until a real `giga1-*` image exists.

## What it publishes

One vanilla image per push to `giga-1`, tagged
`giga1-<YYYYMMDDHHMMSS>-<sha7>`, to the existing
`189176372795.dkr.ecr.us-east-2.amazonaws.com/sei/sei-chain` repository. Same
repository, so the platform Pod Identity grant needs no terraform change.

## Decisions a reviewer should challenge

- **Push-triggered, not cron.** "Every new image deploys" means every commit,
  not one a day. `concurrency: cancel-in-progress: true` means a push during a
  build cancels it, so at most one build runs and the published tag is always
  the newest commit that finished.
- **Timestamp, not `<date>`.** Building on push puts more than one image in a
  day. The house `(?P<date>[0-9]{8})` extract would tie, and a numerical
  ImagePolicy cannot break a tie. 14 decimal digits stay inside float64's
  exact-integer range, so the ordering is exact.
- **`giga1-` prefix.** The three nightly ImagePolicy patterns are anchored on
  `nightly-`, `mock-nightly-` and
  `mock_chain_validation-mock_balances-nightly-`, and they carry no branch
  identity. A bare `nightly-<date>-<sha7>` tag built from `giga-1` would be
  selected by `sei-chain-nightly` and could bump the nightly release gate onto a
  `giga-1` build. This prefix cannot match any of them.
- **Vanilla only, no `mock_balances`.** `giga-testnet` submits no transactions,
  so no sender needs funding, and `mock_balances` would mask the state-machine
  regressions the chain exists to catch. One variant also keeps the loop fast
  and ECR growth low.
- **`load` then smoke then push.** The tag only exists once `seid version` has
  run inside the built image, so Flux cannot deploy a binary that never started.
  One build, so the bytes smoked are the bytes published. It does not prove the
  chain reaches consensus — that gate is named as a follow-up in the platform PR.
- **`paths-ignore` on markdown.** A docs-only commit produces an identical
  binary under a new `sha7`. The new tag would roll all four validators and stop
  block production for the restart window.
- **`workflow_dispatch` ref assertion.** `workflow_dispatch` takes no branch
  filter, so a dispatch from another ref would publish a `giga1-` tag built from
  that ref and Flux would deploy it. The first step fails loudly rather than
  skipping green.

## Verified

- `actionlint .github/workflows/giga1-ecr.yml` — clean.
- `evm.http_enabled`, `evm.ws_enabled` and `state-store.ss-enable` exist as
  dotted keys on this branch, and `SetEVMConfigByMode` /
  `TestTheRenderedConfigStillDiscardsTheStateStoreRules` confirm validator mode
  turns all three off. That is why the platform PR sets all three.
- `giga_executor.enabled` and `occ_enabled` both default to `true` in
  `giga/executor/config`, so the platform PR pins neither.

## Not verified

The workflow has never run. The build, the ECR push and the smoke step are
untested until this merges and `giga-1` takes a push, or someone dispatches it.


bdchatham:
Both findings were right. d307165d0 fixes them.

**Re-run ordering.** The wall-clock tag did break the invariant the design
rests on, and `cancel-in-progress: true` makes a cancelled older run the
likeliest re-run target. The tag now derives its timestamp from the committer
date, so it is a pure function of the commit.

One correction to the suggested snippet: `--date=format-utc:...` is not a git
date mode. `git 2.53` rejects it with "date format missing colon separator".
The working equivalent, which I checked, is:

```sh
TS=$(TZ=UTC0 git show -s --format=%cd --date=format-local:%Y%m%d%H%M%S HEAD)
```

It is UTC-normalised (cross-checked against `--date=iso-strict`, which reports
the committer's `+02:00`), identical across two invocations, and ordered
correctly against the parent commit. `format-local` plus `TZ=UTC0` is what does
the normalising; plain `format:` would use each committer's own recorded
offset, which two committers in different zones could invert.

A re-run now recomputes the same tag, and the repository is `IMMUTABLE`, so the
push refuses it and the re-run goes red. That is the right direction — a red
re-run rather than a chain rolled backwards — and
`clusters/harbor/giga-testnet/README.md` records it.

**Markdown filter.** Correct, and it mattered: `**` matches zero characters, so
`'**/*.md'` needs a leading slash and misses `AGENTS.md`, `CLAUDE.md` and
`README.md`, which is what this repo churns. Now `'**.md'`.

This workflow carries no comments by request, so both invariants live in the
consumer's README under "What the producer must keep true".

masih (approved):

philipsu522 (approved):

PR #4179, opened 2026-09-15
Instrument the autobahn producer tx ingest path

masih:
On the autobahn-e2e runs the validator receiving the load generator traffic sits at ~28 cores while its peers sit at 4-7, and its certify/execute stages fall hundreds of blocks behind receive. The ingest path (`producer.insertTx`) is the only part of that node doing work the peers do not, but nothing in it is measured, so we cannot tell whether the time goes to the app `CheckTx` (ecrecover), the `EvmNonce` state read taken under the mempool lock, the blocked-on-capacity wait, or the broadcast wake-ups that make every blocked inserter re-evaluate `IsFull`.

This adds a `tendermint_internal_autobahn_producer_*` metrics package, generated with `metricsgen` like the avail and data ones, and hooks it into `insertTx` without changing behaviour: histograms for `check_tx_latency`, `nonce_lookup_latency`, `capacity_wait_latency` and `admit_latency` (lock acquisition plus hold, minus the capacity wait), a `capacity_wait_wakeups` counter for the times a waiter woke up to a still-full mempool, an `in_flight{phase}` gauge over `check_tx`/`capacity_wait`/`admit`, and an `inserts{result}` counter over ok/rejected/too_large/full/not_producing/bad_nonce/error. The CheckTx call, nonce read and wait loop are lifted into named helpers so each measured step is one function; the wait loop body is otherwise unchanged. These give a before/after for the follow-up optimisations of the same path (single ecrecover per tx, nonce read outside the lock, FIFO admission, bounded ingest CPU).


philipsu522 (approved):

wen-coding (approved):

sei-tendermint/internal/autobahn/producer/mempool.go
```diff
@@ -244,32 +312,28 @@ func (s *State) insertTx(ctx context.Context, tx tmtypes.Tx, waitIfFull bool) (*
...
 		}
-		for m.IsFull() {
-			// mempool is constructed as a FIFO - we do not delay insertions of large txs (going over cap)
-			// in favor of waiting for smaller txs. This simple algorithm allows us to cap
-			// pending txs to size of a single block. We can refine this rule later if needed.
-			// NOTE: in case there are N concurrent InsertTx calls, this condition is reevaluated N times
-			// every time mempool is updated. Depending on proportion of N to the block size it might get too
-			// expensive.
-			if err := ctrl.Wait(ctx); err != nil {
+		if m.IsFull() {
```
wen-coding, line 329: why if instead of for?

PR #4182, opened 2026-09-15
Bound concurrent CheckTx on the producer ingest path

masih:
On autobahn-e2e load tests the validator that receives the load-generator traffic runs at ~28 of 32 Ps while its peers idle at 4-7, and its certify/execute stages fall hundreds of blocks behind receive. Every broadcast_tx RPC call runs the app CheckTx (including ecrecover) on its own goroutine via producer.State.insertTx, so under load there is nothing stopping ingest from taking every P and starving the consensus, data and execute loops.

This change bounds that work. producer.Config gains MaxConcurrentCheckTx (default max(1, GOMAXPROCS/2)), and State.insertTx now takes a permit from a utils.Semaphore of that size around CheckTxSafe, releasing it before the mempool Watch lock is taken. Waiting for a permit is ctx-cancellable, so cancelled waiters return without consuming a slot. Admission, ordering and nonce semantics are unchanged: the permit only covers the CheckTx call. The knob is exposed in the autobahn JSON config as max_concurrent_check_tx (absent means the default, 0 is rejected by Validate). The permit wait is exposed as its own insert phase (in_flight{phase="check_tx_wait"}) plus a check_tx_wait_latency histogram, so queue depth behind the cap is visible alongside the existing CheckTx metrics. BroadcastTxAsync still spawns its own goroutine per call; since the insert path now observes ctx cancellation while waiting for a permit, that goroutine passes context.WithoutCancel of the request ctx so an async broadcast is not dropped once the HTTP handler returns. Replacing the goroutine with a scope owned by the RPC environment remains out of scope here.

To validate on a brandon-autobahn run under the same load: pod CPU of the ingesting validator should drop from ~28 cores to roughly half of GOMAXPROCS plus consensus overhead, executed TPS should hold (or improve, since execute is no longer starved), and next_block{stage="certify"} - next_block{stage="receive"} should stay near 1 instead of drifting into the hundreds. If executed TPS drops with the lag fixed, raise max_concurrent_check_tx in the autobahn config to trade back some ingest parallelism.


philipsu522 (approved):

masih:
@seidroid review

PR #4183, opened 2026-09-15
Recover each EVM tx sender once on the ingesting validator

masih:
On the validator that receives load-generator traffic every EVM transaction is sender-recovered three times: once in `eth_sendRawTransaction` to pick a shard proxy, once in the evmonlyapp `CheckTx`, and once more in `Executor.PrepareBlock` when the block executes. ecrecover is the dominant per-transaction CPU cost on this path, and that node was pinned at ~28 cores while the others sat at 4-7.

This change makes the RPC layer recover the sender only when the backend can actually proxy: `Backend` and `GigaRouter` gain `EvmProxyEnabled()`, which is the validator's `EnableEvmProxy` flag (fullnodes always proxy), and `SendRawTransaction` goes straight to `BroadcastTx` when it is false. For execution, the evmonlyapp remembers the `hash -> sender` recovered by `CheckTx` in a bounded `utils.Mutex` map and, in `FinalizeBlock`, resolves them in one pass into `BlockRequest.Senders`, a `[]utils.Option[common.Address]` aligned with `Txs` (the hash of a raw transaction is the keccak of its bytes for every type, so this needs no decoding); `parsePreparedTx` uses the given sender only when the decoded transaction is protected and carries the block signer's chain ID, and otherwise recovers as before. Entries are removed from the map as they are handed to the executor, and the map is cleared if it grows past 2^18 admitted-but-never-executed entries. The trust argument is "same process, tx already verified by CheckTx": the cache is never fed from anything received over the network, so transactions that arrive through other lanes are still recovered in `PrepareBlock`, and consensus, admission and block validity are unchanged.

`BenchmarkParsePreparedTx` shows the saving per transaction: decode plus recovery is ~32 µs/op and 29 allocs, decode with a known sender is ~1.9 µs/op and 16 allocs, so for locally admitted transactions execution-side parsing drops by roughly 30 µs of CPU each, on top of the RPC-side recovery that is skipped entirely when proxying is off.


giga/evmonly/types.go
```diff
@@ -36,10 +36,15 @@ type ResultSink interface {
...
+// returns the sender of a transaction whose signature this process has already
+// verified against the executor's chain ID, keyed by transaction hash;
+// PrepareBlock uses it instead of recovering the sender for transactions bound
+// to that chain. Transactions it does not know are recovered as usual.
 type BlockRequest struct {
-	Context BlockContext
-	Txs     [][]byte
+	Context     BlockContext
+	Txs         [][]byte
+	KnownSender func(common.Hash) (common.Address, bool)
```
shemnon: every signature would be unique.  Would an array matching tx order work better?
unless we are expecting duplicate transactions from tip cut merge.

shemnon (approved):
LGTM
Fairly common optimization when it comes to ecdsa, so lower risk than might be expected.

giga/evmonly/types.go
masih: Apologies for the AI nosie; ty for pointing this out @shemnon

PR #4184, opened 2026-09-15
Admit blocked InsertTx calls in FIFO order and bound the waiter queue

masih:
When the producer mempool is full, every InsertTx call blocked in producer.State.insertTx waited on the mempool Watch, so each ctrl.Updated() (a sealed block, a prune, a new block starting) woke all blocked inserters, each of which re-took the lock and re-evaluated IsFull. Under load-generator traffic the ingesting validator accumulates thousands of such inserters, which makes every mempool update O(N) in lock contention and wake-ups and is consistent with the goroutine and CPU excess seen on that node.

Blocked inserters now take a ticket in a FIFO queue on mempoolInner and wait on a per-ticket AtomicSend. Only the head of the queue is signalled, and only when the mempool has capacity: after a prune, after the head itself is admitted or leaves, and on close. A waiter whose context is cancelled removes its ticket and passes the turn on, and closing the mempool releases every waiter with ErrNotProducing as before. A new producer.Config field MaxPendingInserts (default 4096, wired wherever producer.Config is built) caps the number of blocked inserters; past that InsertTx fails immediately with the mempool-full error so clients back off instead of piling up goroutines. TryInsertTx is unchanged.

New tests cover FIFO order under concurrent inserters, one admission per freed slot, a cancelled waiter not blocking the queue, release on close, and the bound. The producer package passes under -race, and the touched packages pass go vet, golangci-lint v2.13.2 and the formatters.


shemnon (approved):
LGTM

PR #4186, opened 2026-09-15
Skip race, upgrade and integration jobs in the giga-1 merge queue

masih:
The giga-1 merge queue re-ran the whole PR suite for every queued entry: four race-detection shards plus the upgrade tests (Go Test), the state_db race run (Sei-DB), RocksDB tests, and the full Docker integration matrix. Those had all already passed on the PR itself, so in the queue they only added wall-clock time and runner minutes, and with several PRs landing back-to-back the queue kept re-running them on each rebase.

Merge-queue entries whose base is giga-1 now skip all of those. The required check names are preserved so the queue still gets a green result: the Go Test `Race Detection` roll-up accepts `skipped` from the shards and upgrade tests only for the giga-1 queue, and the Sei-DB `Race Detection` job is skipped at the job level with the same predicate #4178 introduced, since a skipped job satisfies a required check. RocksDB tests are already skipped for giga-1 by #4178. Docker Integration Test drops giga-1 from its `merge_group` filter while still running on giga-1 pushes and pull requests. The giga-1 queue is left with lint, UCI Go lint and Buf; main and release/** are unchanged.


wen-coding (approved):

philipsu522 (approved):

masih:
@seidroid review

PR #3849, opened 2026-08-04
Refactor separating AppQC from CommitQC

pompon0:
Redesign of the AppQC flow:
* AppQCs are collected by data.State natively, rather than as AppProposals within CommitQCs
* avail.State anchor is persisted in data.State
* AppQCs are collected for every CommitQC, for the last block of the CommitQC. Local AppProposal (locally computed AppHashes) are persisted in BlockDB as well, because we currently have no access to historical apphashes.
* data.State can now only prune AFTER all: CommitQC,blocks,AppQC,(local) AppHash are collected for the given height. At least one (complete) row has to stay in DB.

PR #4189, opened 2026-09-15
Keep lane parent hash across pruning in ProduceLocalBlock

masih:
`ProduceLocalBlock` derived the parent hash from the lane queue, and emitted the zero hash whenever the queue had been fully pruned by the data Anchor (the lane idle, every produced block already certified and evicted). A replica whose Anchor had not advanced as far still held the previous block, so `PushBlock` compared the zero hash against the real one, logged `parent hash mismatch (producer equivocation)` and dropped the block. Later blocks of that lane then parked in `PushBlock`'s `WaitUntil` until the replica's own Anchor pruned past them, so the lane froze until the next commit round caught up. On brandon-autobahn-07 (running 3cad9bf) this fired on every producer lane and showed up as the global block number advancing in stall-then-burst steps, create-to-execute latency p50 of 60-115s, and a low blocks/sec on the autobahn-e2e dashboard, while commit-to-commit latency itself stayed at ~0.12s.

The lane block queue is now a `blockQueue` which records the hash of the last block pushed, and `ProduceLocalBlock` uses that instead of reading the previous entry, so pruning no longer changes the parent a producer puts on its next block. The receiver check is unchanged.

The same gap existed across a restart: `restoreInner` prunes to the Anchor before replaying the lane WAL, and `truncateForAnchor` could drop every block of an idle lane, so a node restarting with a fully certified lane produced its first block with a zero parent again. The lane WAL now always retains its last persisted block (`PruneBefore(min(first, nextBlockNum-1))`), and `restoreBlocks` seeds `lastHash` from the skipped WAL block at `q.next-1`. `LaneRange.LastHash()` from the Anchor is deliberately not used as a source, since it is zero for a lane the tipcut omitted.

Tested with `go test -race ./internal/autobahn/...` in sei-tendermint. `TestProduceLocalBlock_ParentHashSurvivesPrune`, `TestProduceLocalBlock_ParentHashSurvivesRestart` and `TestPrunePastAllKeepsLastBlock` each fail without the corresponding change and pass here.


wen-coding:
The in-process fix is right: lastHash on pushBack, ProduceLocalBlock reads that, prune does not clear it. But I think let's add a TODO to fix restart case:
Restart: lastHash is not restored when the queue is empty after prune (WAL entries below q.next are skipped). Do not seed it from the current QC’s LaneRange.LastHash(); that is zero when the tipcut omitted the lane even if the lane has history. Keep the last header (or lastHash) across prune/WAL truncate instead.

wen-coding:
Okay, thought about it further, three things:
1. The saved lastHash should be preserved with block number
2. If there is a parent b in local but you see b' in a QC, you should favor b' over b, and you should also discard whatever disagrees locally and refetch (this part we probably have never implemented)
3. If a lane has no blocks in a QC, we should copy over lastHash from last QC so we don't lose the lastHash

I feel there is a regression somewhere in [PR #3849](https://github.com/sei-protocol/sei-chain/pull/3849), I need to take a closer look tomorrow.

masih:
Closing in favour of Wen's PR

PR #4190, opened 2026-09-15
Persist the EVM-only execution cursor with each block's FlatKV state

### LINKED ISSUE #4169: evmonly: the app keeps its committed height in memory, so any restart after block 1 crashloops forever

masih:
Fixes #4169. The EVM-only Autobahn app made FlatKV state durable inside `FinalizeBlock` but kept its height, app hash and parent hash only in memory, advanced by `Commit`. After a restart `Info()` reported height 0, the Giga router took the fresh-genesis branch, and `InitChain` plus block 1 were replayed against state that already held those nonces, so the node crash-looped. The `initialHeight == 1` early return in the InitChain guard meant a non-empty store was never refused at the default initial height.

The executor gains an optional `BlockChangeSetEncoder` whose named changesets are appended to the block's state changesets in the same `CommitStateChanges` call. The app uses it to write an `evmonly`/`cursor` record (`height ‖ appHash ‖ parentHash ‖ gasLimit`) atomically with each block, and `NewEVMOnlyApplication` reads it back on construction so `Info()` reports the durable height and the router takes its restart branch. A crash between `FinalizeBlock(N)` and `Commit(N)` therefore restarts at N. `InitChain` is refused once an executor exists or the store is at any version other than 0 or `initialHeight-1`. The chained SHA-256 app hash is unchanged.


codchen (approved):

giga/evmonly/README.md
```diff
@@ -98,6 +98,15 @@ including for empty blocks. A receipt failure leaves state unchanged so the
 block can be retried. A state failure can leave receipts behind, but retrying
 the block overwrites them. `ResultSink` runs only after both stores succeed.
 
+Because state becomes durable inside `ExecuteBlock`, before any ABCI `Commit`,
```
cody-littley: Not entirely accurate. The crash durability point is when we flush the block to BlockDB. The storage layer is intentionally asynchronous, and so at any point in time we may have dozens of executed blocks that are not crash durable at the DB layer. This is ok though, since we can re-execute blocks from the BlockDB.

cody-littley: In theory we could make `ExecuteBlock` crash durable as this comment suggests. Cost of doing this is likely on the order of 50% to 80% reduction in max DB throughput.

cody-littley (approved):

PR #4198, opened 2026-09-16
Remove the CosmWasm pointer creation path

alexander-sei:
`MsgRegisterPointer` instantiated a CosmWasm wrapper (CW20/CW721/CW1155) for an ERC20/721/1155 token. That path has been closed for a while, from two directions:

- The `register_pointer_disabled` parameter has rejected registration since v6.1.0 ([#2186](https://github.com/sei-protocol/sei-chain/pull/2186), evm module migration 18).
- Mainnet [proposal 115](https://seistream.app/proposals/115) set `code_upload_access` to `Nobody`, which does not retroactively lock the already-stored wrapper code but does mean the remaining code-upload calls can no longer succeed. `StoreCWPointerCode` panics on the error it now gets, so any future upgrade wiring a new pointer-version migration would halt the chain at the upgrade height.

## What this removes

- The `MsgRegisterPointer` handler, which now returns `ErrRegisterPointerDeprecated`, following the existing `Associate` pattern. The proto RPC and message type stay so historical transactions still decode.
- `seid tx evm register-cw-pointer`.
- The embedded wrapper wasm (`x/evm/artifacts/erc{20,721,1155}`) and the checked-in copies under `contracts/wasm`.
- The genesis code upload in `InitGenesis`, `StoreCWPointerCode`, and the `MigrateCWERC*Pointers` helpers.

Migration boundaries 4, 6, 7, 9 and 15, whose only body was that work, stay registered but inert: a module's migration sequence has to be unbroken for a chain that upgrades across it.

## What deliberately stays

Everything on the read side, because the wrappers already registered on mainnet are live contracts: the registry lookups, the reverse registry, `MsgAssociateContractAddress` (the only way any wasm contract gets an EVM address, since contracts have no pubkey), and the `Pointer` / `PointerVersion` queries. `GetStoredPointerCodeID` stays with them — `PointerVersion` returns it as `cw_code_id`. The `erc20/721/1155` `CurrentVersion` constants stay too: pointer lookups scan versions downward from them, so changing one would make existing wrappers invisible to the chain. `register_pointer_disabled` is retained in `Params` for state compatibility but is no longer read.

The Rust sources under `example/cosmwasm/cw{20,721,1155}` are kept — they are the source of wrappers still running on mainnet.

## Consequence worth reviewing

A chain initialized from genesis no longer uploads the three wrapper wasm blobs, so **the first code a user stores gets code ID 1 rather than 4**, and the addresses derived from it move. Running chains are unaffected, since the code IDs they stored remain in state. This is why the pinned CW20 address in `evmrpc/tests` and `precompiles/wasmd` changes in this PR, and why `x/evm/module_test.go`'s exported-genesis fixture loses its three `serialized` entries.

Test coverage that exercised wrapper creation is gone with it: the three `CW*toERC*PointerTest.js` suites, the two CW-wrapper royalty cases in `x/evm/integration_test.go`, and the "pointer to the pointer" blocks in the surviving EVM-pointer suites. `TestAssociateContractAddress` and `DisableWasmTest.js` now instantiate a plain `cw20_base.wasm` instead of registering a pointer.

## Verification

`go build ./...`, `go vet ./...`, `make fmtcheck`, `golangci-lint run`, `make build`, and `scripts/ramtest.sh` over `x/evm`, `app`, `precompiles`, `evmrpc`, `occ_tests`, `wasmbinding` and `testutil`.

No CHANGELOG entry in this branch by request; the state-machine notes above are the part worth carrying into release notes.


alexander-sei:
@seidroid review

sei-will (approved):

masih (approved):

PR #4204, opened 2026-09-16
giga: make EVM-only transaction outcomes observable (PLT-1290)

amir-deris:
A giga run can today show that validators agreed on a block, but not what the
transactions in it actually did. This adds the signals that make transaction
outcomes observable, and a verifier that checks balances are conserved across a cluster, so a run can assert correctness rather than only quorum.

## What this adds

**`txs_executed_total`** (`giga/evmonly`) — EVM-only transactions per block
labelled `status` = `success` | `reverted` | `failed`, recorded from
`ExecutePreparedBlock`.

**`receipts_written_total`** (`sei-db/ledger_db/receipt`) — receipts the store
committed, on the same three-value `status` vocabulary, so the two counters
compare bucket for bucket and a gap between them localises to the write path.
The count is taken in a wrapper around the write, so a failed write is never
counted as a written one; the litt store counts inside its async writer over the
records it kept.

Both counters report **every** status on every block, including the ones that
block had none of. A counter series exists only once something records to it,
and a global OTel instrument drops measurements taken before the meter provider
is installed — so priming the label set from `init()` does not work, and a
status with no occurrences would otherwise be absent rather than zero.
Recording is wrapped in a recover: this runs inside block execution, and a
telemetry fault must not panic into the caller.

**Real ABCI tx results in evmonly** — each `ExecTxResult` now carries the EVM
failure reason in its `Log`. The code stays `CodeTypeOK` even for a revert: a
revert is a successfully executed transaction at this layer (it consumed its
nonce and gas, and the receipt status carries the failure), and a non-OK code
would put the hash in the mempool's failed set, which holds it for a second
chance instead of recording it as executed. Varying the log is safe because the
results hash covers only code, data and gas.

**`autobahn-e2e verify-conservation`** — queries `eth_getBalance` for a sample of
addresses on every validator in a cluster and requires all endpoints to agree.
An accounts file supplies the addresses and, optionally, `expected_balances` /
`expected_sum` to compare against a precomputed ledger; with no expected values
it still enforces endpoint agreement and the initial-balance ceiling. The file
is resolved into a ledger before any query, which is where it refuses a sample
limit that would truncate a summed file and rejects an `expected_balances` key
naming an unlisted address rather than silently skipping it.


amir-deris:
@seidroid review

masih (approved):

PR #4211, opened 2026-09-16
Refuse EVM-only state that holds blocks but no execution cursor

masih:
An EVM-only node whose FlatKV state was written before #4190 introduced the durable execution cursor holds block state at a nonzero version but no `evmonly/cursor` record. `loadEVMOnlyCursor` treated a missing cursor as "nothing to resume" and returned None, so `Info()` reported height 0, the giga router took the fresh-start path and called `InitChain`, and `seedInitialStateVersion` then refused because the store was already at height 364932. The node crash-looped with `App.InitChain(): EVM-only state is already at height 364932 before InitChain`, a message that reads like a stale-genesis or re-init problem when the actual cause is a store that cannot tell where execution stopped.

`NewEVMOnlyApplication` now takes the genesis initial height. A store without a cursor is still resumed through `InitChain` when it is empty or sits exactly at `initialHeight-1`, which is what a crash between `InitChain` and the first block leaves behind; at any other version it is refused at construction with `errEVMOnlyCursorMissing`, which names the state height, the chain's first height, and the remedy (reset the EVM-only state directory so the node resyncs from the block store). `prepareApplication` reads the genesis document to supply the height, so the same value drives both the startup check and `InitChain`. The pre-cursor stores on the testnet still need a one-time state reset; the change makes that failure explicit at startup instead of an ambiguous panic inside the router.


wen-coding:
Now that we are reverting #4190, why do we want to do this?

PR #4219, opened 2026-09-17
Improve EVM execution performance

yzang2019:

Raises `evmonly` execution throughput from ~48.7k to ~70k TPS on a 12-core dev box and 100k on ec2 12xlarge machine. Part of that is the workload rather than the executor — see the note under the account pool.

To reproduce it:
`./build/evmonly-loadtest --metrics-addr=0.0.0.0:9698 --report-interval=15s   --blocks=0 --accounts=200000 --txs-per-block=2000 --builders=12 --workers=1   --executor-workers=8 --queue-size=256`

**Improvements**

- **Commit off the critical path.** A block's state commit runs in the background while the next
  block executes. That block reads its predecessor's changes through a `pendingOverlay`, since a
  `StateView` never sees writes made after it was opened. One commit in flight at a time, so
  commits stay ordered.
- **One row read per account.** New optional `AccountReader` on `EVMStateView` returns balance,
  nonce and code hash together, instead of resolving the same row once per field.
- **Parallel prefetch in the OCC merge.** The merge is the largest serial phase, and most of it was
  per-address reads; they now resolve across the worker pool first.
- **Cheaper read path.** An unconditional `AccountExists` probe is short-circuited, and receipt
  encoding fans out across the OCC pool.
- **Account pool — the largest single factor, and a workload change.** `--accounts` adds a reusable
  sender pool with derived nonces, so `--blocks=0` streams blocks and runs until interrupted at flat
  memory instead of prebuilding the whole run. Reusing senders also means the view cache is actually
  warm: hit rate goes from ~70% to ~82% and throughput from 57.1k to 65.3k, ~14% on its own, with
  every read-heavy phase improving and the write phases flat. Minting a fresh account per
  transaction — the previous behaviour — is not what a real chain looks like, but a pooled number is
  still a ceiling relative to cold state, so it is not directly comparable to the old one.
- **Observability.** Block and pipeline stages publish OTel phase timers, plus a Grafana dashboard
  showing where the executor's wall time goes. `--gc-percent` exposes the Go GC target (opt-in:
  trades memory for throughput, needs `GOMEMLIMIT` on a memory-limited host).

**Risk.** The pipelined commit changes when a block's state becomes visible, so that is the part to
review. `ExecutePreparedBlock` returns before the commit lands; `AwaitCommits` is the sync point and
latches the first failure. `ExecuteBlock` still awaits internally, so the synchronous API is
unchanged. Outside `evmonly-loadtest`, nothing drives the pipelined path today.

- `go test -race ./giga/evmonly/...` and `golangci-lint run` on the touched packages, both clean.
- New tests pin the pipelined path: two blocks where the second spends what the first paid it,
  against a store whose view never advances; a commit failure reaching every waiter and staying
  latched; concurrent waiters all released; and a commit retired while a block opens its view, which
  fails without the ordering fix. Further tests cover the single-row account read and the prefetch.
- 4,000,000 transfers with zero failures and zero OCC fallbacks.
- Throughput measured with interleaved A/B runs, since the test machine drifts ~10% thermally.


yzang2019:
@seidroid review

yzang2019:
@seidroid review

yzang2019:
@seidroid review

yzang2019:
@seidroid renew

bdchatham (approved):

masih (approved):

PR #4238, opened 2026-09-17
Proposal: let an unappliable transaction fail instead of the block

bdchatham:
**This is a proposal, not a routine fix.** It changes what a validator agrees to, and it touches a behaviour @codchen wrote deliberately. I would like your read before it lands.

## What happened

`giga-testnet-2` halted at block 296796 and cannot be recovered.

```
executeBlock(296796) → app.FinalizeBlock():
execute tx 2 0xd6822e97c11492dcfdca26f566e4420f984601c591dd3dee3914f17fdbd1f6cb:
nonce too low: address 0x9a8e327a45A95AD9c7Fa3E1FFa19E03D86135992, tx: 178 state: 179
```

`core.ApplyMessage` returns `nonce too low`; the executor returns it as a block error; `FinalizeBlock` propagates it; `SpawnCritical` panics. Nine validators logged the identical block, transaction, address and nonce pair; the other 31 stalled behind them. No node disagrees: they all agree the block holds a transaction that cannot be applied, and the executor's answer is to die. Restarts and state rollbacks replay straight back into 296796. Second occurrence today.

## Why the current behaviour is now wrong, and why that isn't a criticism of it

`TestExecutorPreVMFailuresMatchGeth` and `TestExecutorValidationFailuresAbortBlock` assert the current behaviour on purpose, and **geth is right**: under Ethereum's rules a block containing a spent nonce is invalid, because the producer validated the nonce before including it.

The premise stopped holding. The executor landed in #3583 (2026-08-03) as a port of the sei-v3 execution shape; abort-on-pre-check-failure came with the port rather than being decided. `evmonlyapp`, the Autobahn path that feeds it pre-ordered *unvalidated* blocks, arrived five weeks later in #4088. "The producer already validated the nonce" was true of every caller when the code was written and silently stopped being true when Autobahn was wired in front of it.

## The change

`Config.RejectUnappliableTxs` records a transaction that fails `ApplyMessage`'s pre-checks as a failed receipt with zero gas and no state change. **Default off** — the parity suite and loadtest harness are live callers and keep geth semantics. The opt-in sits at the one call site whose ordering guarantees differ, `evmonlyapp`. **No existing test changed.**

```go
// executeTx
snapshot := stateDB.Snapshot()
poolGas  := gasPool.Gas()
execResult, err := core.ApplyMessage(evm, msg, gasPool)
if stateErr := stateDB.Error(); stateErr != nil { return ..., stateErr }   // node-local fault: still fails the block
if err != nil {
    if !e.cfg.RejectUnappliableTxs { return ..., err }
    stateDB.RevertToSnapshot(snapshot); stateDB.clearSnapshots()
    gasPool.SetGas(poolGas)
    return rejectedTx(p, block, txIndexUint, baseFee, err)              // failed receipt, GasUsed 0, Rejected: true
}
```

Three details matter:

- **The snapshot and pool restore are load-bearing.** geth's `preCheck` runs `buyGas` (debits the sender *and* `gasPool.SubGas`) before `ErrIntrinsicGas` / `ErrInsufficientFundsForTransfer` can fire, so a rejection can leave both the balance and the block's gas pool moved. Reverting both is what keeps the changeset and the remaining block gas consistent with the zero gas the receipt reports. `stateDB.Error()` is read before the revert because `RevertToSnapshot` restores the recorded error too.
- **OCC agrees with sequential.** `executeTxSpeculative` records `gasLimit: 0` for a rejected result; otherwise OCC validation charges the declared limit against remaining block gas, raises `ErrGasLimitReached` as a hard error, and fails the block on exactly the path `evmonlyapp` runs (it enables OCC). Without this a `GOMAXPROCS=1` node would accept a block its peers panic on.
- **Metrics.** `TxResult.Rejected` gives `giga_evmonly_txs_executed_total` a `rejected` label distinct from `reverted`, so a chain quietly dropping its traffic doesn't look like ordinary contract failure.

## Open decisions for @codchen

1. **Receipt or skip?** A receipt keeps the 1:1 alignment `evmOnlyABCIResults` needs between a block's transactions and its results; the cost is that a never-executed transaction is queryable through `eth_getTransactionReceipt`.
2. **`ErrGasLimitReached` is a statement about the block, not the transaction.** Every remaining transaction in a gas-exhausted block gets a failed receipt. Deterministic and identical on every node, but arguably the wrong answer.
3. **`ErrNonceTooHigh` is in the rejected set too.** A nonce gap is only final given this block's ordering, not the pre-state; if Autobahn can deliver one sender's nonces out of order, `n+1` burns a receipt for a transaction that would have applied a block later. I left it rejected because the alternative is the halt, and reordering belongs to ordering/admission, but it's a policy call.
4. **Is execution the right layer?** #4170 moved admission onto predicate block-validity charges — the same class of fix one layer earlier. If a spent nonce should be filtered at ordering or admission, this becomes a safety net rather than the fix, and I'd scope it that way.

## An accepted cost, stated explicitly

A rejected transaction consumes no gas and no nonce, so a sender can keep unappliable transactions flowing at zero cost while they occupy block space and receipt storage. Under Ethereum's rules this cannot happen, because such transactions never reach a block. Here they can.

I am accepting that rather than pricing it, on two grounds:

- The alternative is the halt this PR removes. A chain that dies on the first stale nonce has a worse problem than one carrying free receipts.
- Admission is where it closes, and `CheckTx` already carries the marker: `TODO(evmonly-production): close the gap between admission and block validity before accepting arbitrary traffic; this test app assumes executable load-test transactions` (`sei-tendermint/internal/evmonlyapp/app.go:262`).

This change does not widen that gap — it stops the gap from being fatal — but it does remove the pressure a halt put on closing it, which is exactly why it is worth writing down instead of letting it fall off. It needs to be settled before the flag goes near a chain taking untrusted traffic. On a load-test chain the senders are ours.

New tests in `giga/evmonly/reject_tx_test.go`, each run on both the sequential and the real OCC path (`OCCWorkers: 4`, asserting `OCCStats.Attempted && !Fallback`) where relevant:

- `TestExecutorRejectsSpentNonceWithoutFailingBlock` — reproduces the halt (`nonce too low: … tx: 0 state: 1`); fails without the change.
- `TestExecutorRejectsTxOverBlockGasWithoutFailingBlock` — 30k block, two 21k txs; failed under OCC with `gas limit reached` before the `gasLimit: 0` fix.
- `TestExecutorRejectionRestoresBlockGasPool` — 42k block, an `ErrIntrinsicGas` tx then two 21k transfers; the third tx is rejected without the `SetGas` restore.
- `TestExecutorRejectsUnderfundedTxWithoutFailingBlock`, `TestRejectedTxCountsAsItsOwnStatus`, `TestExecutorDefaultStillAbortsOnUnappliableTx` (pins the default).

`go test ./giga/evmonly/...`, `-race` on the new tests, `sei-tendermint/internal/evmonlyapp`, `go vet`, and `make fmtcheck` are clean locally.

I have not recreated the chain. `giga-testnet-2` stays down until this behaviour has an owner's sign-off.

Link to Devin session: https://app.devin.ai/sessions/5df181ca7c404288b8c351e7d1616c9d
Open in Devin Desktop: https://app.devin.ai/desktop/session/5df181ca7c404288b8c351e7d1616c9d?variant=devin
Requested by: @bdchatham


codchen (approved):

bdchatham:
Disposition of the two PR-level suggestions from @seidroid's review. Both are addressed; neither had an inline thread to resolve, so recording them here. The six inline threads are resolved with their dispositions.

**1. `evmOnlyABCIResults`' godoc — changed** (`7226cfaa2`). You were right that the stated reason no longer covered every result the function emits: "it consumed its nonce and gas" is false for a rejected transaction. It now reads that every result is OK — reverted or rejected — because the hash is recorded as executed rather than held in the mempool's failed set for a retry, and the receipt status carries the failure. The results-hash clause stays, since that part was still accurate.

**2. Zero-cost unappliable traffic — accepted, and now stated in the description** under *An accepted cost, stated explicitly*. A rejected transaction consumes no gas and no nonce, so a sender can keep them flowing for free while they occupy block space and receipt storage. Ethereum's answer is that such transactions never reach a block; here they can.

Accepted on two grounds. The alternative is the halt this PR removes, and a chain that dies on the first stale nonce has a worse problem than one carrying free receipts. And admission is where it closes — `CheckTx` already carries `TODO(evmonly-production): close the gap between admission and block validity before accepting arbitrary traffic` (`sei-tendermint/internal/evmonlyapp/app.go:262`).

Your framing of *why* it needs saying is the part I took: this change does not widen that gap, it stops the gap being fatal — but it does remove the pressure a halt put on closing it. That is exactly how a TODO like this quietly stops getting done, so it is written down rather than left implicit. It has to be settled before the flag goes near a chain taking untrusted traffic; on a load-test chain the senders are ours.

**On your inline `ErrNonceTooHigh` point** — left in the rejected set, and I want to flag that I am less comfortable with this than with the rest. Your distinction is the right one: a spent nonce, short funds and exhausted block gas are final given the block's pre-state, whereas a nonce gap is final only given this block's ordering. Burning a receipt for a transaction that would have applied one block later, whose only recourse is resubmitting identical bytes that already have a receipt, is a genuine wart. I left it rejected because the alternative is the halt, and out-of-order delivery is a question for ordering/admission rather than the executor — but it is listed as an open decision for @codchen rather than settled.

Also worth noting for the record: giga-1 moved while this was in review (#4219 landed and touches `occ.go`). Merged in; `occ.go`'s gas validation and `executeTxSpeculative` are untouched by it, and the full suite plus `-race` is clean on the new base.

bdchatham:
@seidroid review

PR #4240, opened 2026-09-17
Make mempool Update/Reap index maintenance incremental

masih:
Every block, `txStore.Update` called `compact`, which recomputes `inInclusionOrder` (two O(m log m) sorts over the whole mempool) and then throws away and rebuilds `byHash`/`byEvmHash`/`byNonce` by re-inserting every tx. `Reap(remove=true)` was worse: it computed the inclusion order to pick txs, then called `compact`, which computed it a second time. At the ~10k tx mempools seen in production this is tens of milliseconds of serialized work on the consensus path per block.

`Update` now removes executed/invalid/expired txs in place through a new `remove` helper and, when the store still fits in `softLimit`, calls `refresh` instead of `compact`: it clears and re-fetches account state, drops txs whose nonce fell below the account nonce, re-caches their metadata and re-derives readiness per account with `advanceReady` (the loop extracted from `insert`), all in O(m) with no sort and no index rebuild. `nextNonce` is still reset to `firstNonce` and rebuilt by walking `byNonce`, so readiness semantics are unchanged. `Reap(remove=true)` removes the reaped txs incrementally and re-derives readiness only for the affected accounts (`refreshReady`); it only falls back to `compact` when the store is still above `softLimit`, where eviction genuinely needs the full order. I did not reuse the pre-removal order for that fallback because reaping the head of an account demotes its successors from ready to pending, so the post-removal order is not a filter of the pre-removal one. Since the RPC snapshot is no longer a byproduct of `compact`, `RecentSnapshot` recomputes it lazily when `snapshotStale` is set.

`compactTotal`/`compactDurationSeconds` now only fire when `compact` actually runs (insert overflow, or when a store above `softLimit` needs eviction), so those series will drop in normal operation. Reviewers should look most closely at `refresh` and `refreshReady` against the previous `compact` re-insertion path. Existing mempool tests pass unchanged under `-race`; the new `tx_bench_test.go` shows `Update` going from ~20ms to ~9ms at 10k txs (32→9ms at 20k txs over 5k accounts, 121→49ms at 50k) and `Reap(remove=true)` from ~21ms to ~7ms (50→15ms at 20k), the remainder being the unavoidable per-account state re-fetch and the single inclusion-order sort.


sei-tendermint/internal/mempool/tx.go
```diff
@@ -353,35 +364,103 @@ func (inner *txStoreInner) shouldReject(txHash types.TxHash) bool {
...
+		wtx, ok := inner.byNonce[an]
+		if !ok {
+			break
+		}
+		requiredBalance := wtx.evm.OrPanic("non-evm tx").requiredBalance
+		if account.balance.Cmp(&requiredBalance) < 0 {
+			break
+		}
+		account.nextNonce += 1
+		state.ready.Inc(wtx.Size())
```
shemnon, line 408: Do we already know the state can hold all wrapped TXes at this point? So this is bookkeeping for what is being held?

If this is because advanceReady is the only caller, that should be called out in the comment so future edits don't break this invariant.

sei-tendermint/internal/mempool/tx.go
```diff
@@ -558,6 +618,46 @@ func (s *txStore) compact(inner *txStoreInner, clearAccounts bool) {
...
 }
 
+// O(m), re-evaluates account nonces and balances against the app and drops txs
+// which fell below their account nonce. Unlike compact, it keeps the indices
+// in place and never evicts, so the caller must ensure the store is within softLimit.
+func (s *txStore) refresh(inner *txStoreInner) {
+	state := txStoreState{}
+	inner.accounts = map[common.Address]*evmAccount{}
+	for txHash, wtx := range inner.byHash {
+		state.total.Inc(wtx.Size())
```
shemnon: Instead of inc-all and dec on eject wouldn't it be cleaner to go inc-only?  Add it in 631 for non-evm and then after the nonce check in 650?  

Also, are non-evm TXes going away in future forks?

shemnon (approved):

sei-tendermint/internal/mempool/tx.go
masih: Yes they will, but keep them for now

PR #4246, opened 2026-09-18
Recover each EVM tx sender once on the ingesting validator

masih:
On an EVM-only validator the sender of every locally submitted transaction was recovered from its signature three times: once in `eth_sendRawTransaction` to pick the shard proxy, once in `CheckTx`, and once more when the block executor parsed the transaction. Signature recovery is the dominant per-transaction CPU cost on the ingest path, so the repeated work caps throughput for no benefit.

This backports #4183 from giga-1. The RPC layer skips recovery entirely when the backend has no proxies configured, since CheckTx recovers the sender anyway. CheckTx remembers the recovered sender by transaction hash in a bounded map (cleared when it grows past 2^18 entries), and FinalizeBlock hands those senders to the executor through `BlockRequest.Senders`, removing them from the map as it goes. The parser only trusts a known sender when the transaction is protected and signed for the block signer's chain ID; anything else, including transactions that arrived via other lanes, is recovered as before, so the app hash is unaffected.

Compared with the giga-1 change the RPC hunk is re-homed into `server.go` (main never received the giga-1 `send.go` split) and the sender cache is added alongside main's single `state` mutex from #4204 rather than the older `executor`/`cursor` pair. The port is verified by `TestEVMOnlyApplicationExecutesCheckedTxLikeUncheckedTx`, which checks that a checked and an unchecked execution of the same block produce identical app hashes and gas, and by `TestSkipsShardLookupWithoutProxies` for the RPC path.

Part of PLT-1300.


shemnon (approved):

giga/evmonly/parser.go
```diff
@@ -60,25 +73,45 @@ func parseBlockTxs(ctx context.Context, txs [][]byte, signer ethtypes.Signer, wo
...
 }
 
-func parseTx(raw []byte, signer ethtypes.Signer) (*ethtypes.Transaction, common.Address, error) {
-	var tx ethtypes.Transaction
-	if err := tx.UnmarshalBinary(raw); err != nil {
-		return nil, common.Address{}, err
+// signerAccepts reports whether signer recovers senders for tx: the transaction
+// is bound to signer's chain and its type is enabled by signer's fork schedule.
+func signerAccepts(signer ethtypes.Signer, tx *ethtypes.Transaction) bool {
+	if !tx.Protected() || tx.ChainId().Cmp(signer.ChainID()) != 0 {
```
wen-coding: Can tx.ChainId() never be nil? How about a pre-EIP-155 ChainConfig plus a known sender?

giga/evmonly/rpc/server.go
```diff
@@ -52,13 +55,11 @@ func (api *sendAPI) SendRawTransaction(ctx context.Context, input hexutil.Bytes)
...
 	hash := tx.Hash()
 
-	if sender, err := ethtypes.Sender(ethtypes.LatestSignerForChainID(tx.ChainId()), tx); err == nil {
-		if client, ok := api.backend.EvmProxy(sender).Get(); ok {
-			if err := client.CallContext(ctx, &hash, "eth_sendRawTransaction", input); err != nil {
-				return hash, err
-			}
-			return hash, nil
+	if client, ok := api.shardProxy(tx).Get(); ok {
+		if err := client.CallContext(ctx, &hash, "eth_sendRawTransaction", input); err != nil {
```
wen-coding: @codchen Hmm I totally understand why you can't reuse evmrpc/, but giga/evmonly/rpc/server.go seems to be buried a bit too deep so people will naturally look for evmrpc implementations there and forget about this one.

I suggest we move the whole rpc/ directory here to some top level directory so we remember. Maybe giga/rpc ? What do you think?

masih: Plus one on this one from me FWIW

PR #4247, opened 2026-09-18
Add opt-in jemalloc 5.4.0 build for seid's cgo allocations

masih:
seid does a meaningful amount of allocation outside the Go heap: Pebble's block cache and memtables go through `internal/manual` (C `malloc`), DataDog/zstd buffers do the same, and libwasmvm carries its own Rust heap on top of the platform allocator. All of that is currently served by glibc malloc, whose per-arena fragmentation under a long-lived, high-churn cache is the reason CockroachDB links jemalloc for the very same Pebble code. Nothing in the tree today lets us try an alternative allocator for those paths short of `LD_PRELOAD`.

This adds an opt-in `jemalloc` build tag that links jemalloc into the binary, so every cgo allocation resolves to it while the Go heap is untouched. `utils/jemalloc` is the only cgo touchpoint: with the tag it carries the `-ljemalloc` directive and exposes `Version()` and `Allocated()` via `mallctl`; without it the package is a constant-`false` stub, so the default build, the static goreleaser build and the release images are byte-for-byte unaffected. `newApp` logs the linked version at start so a node's allocator is visible in its logs. `make build-jemalloc-lib` fetches, checksums and installs the pinned jemalloc 5.4.0 (Ubuntu ships 5.3.0), and `make build-jemalloc` / `make install-jemalloc` compile with the tag. A dynamic link was chosen over vendoring a static archive because the experiment is about measuring, not shipping; folding it into `scripts/build-static.sh` is a follow-up if the numbers justify it.

The new `jemalloc Build Check` workflow builds the library from source (cached), compiles seid with the tag, and verifies the result three ways: `ldd` shows `libjemalloc`, `seid version --long` reports the tag, and an unknown `MALLOC_CONF` key makes jemalloc complain on stderr, which only happens if it is the allocator actually loaded. The package test then asserts `stats.allocated` is non-zero, proving malloc is being intercepted rather than merely linked. There is no consensus, state or wire impact; the reviewer's attention is best spent on whether the opt-in surface is the right shape for a load-test comparison against the current allocator.

`BenchmarkPebbleCHeap` in `utils/jemalloc` is the before/after harness: it drives the real `pebbledb` engine (512 MiB block cache, zstd blocks) through a 200k x 1 KiB fill, random point reads and batched overwrites, and reports RSS alongside the timings so the same command run with and without `-tags jemalloc` diffs cleanly in benchstat. Ten interleaved runs of each build on a RAM disk (Xeon 8559C, 8 threads) show no throughput difference and a smaller footprint where memory is freed back:

```
                        │  libc.txt   │            jemalloc.txt            │
                        │   sec/op    │   sec/op     vs base               │
PebbleCHeap/fill-8        457.2m ± 1%   458.9m ± 3%       ~ (p=0.631 n=10)
PebbleCHeap/randread-8    2.564µ ± 4%   2.700µ ± 8%       ~ (p=0.050 n=10)
PebbleCHeap/overwrite-8   1.828m ± 8%   1.808m ± 9%       ~ (p=0.631 n=10)

                        │  libc.txt   │            jemalloc.txt            │
                        │   rss-MiB   │  rss-MiB    vs base                │
PebbleCHeap/fill-8        152.80 ± 7%   89.21 ± 5%  -41.62% (p=0.000 n=10)
PebbleCHeap/randread-8     338.5 ± 3%   306.5 ± 1%   -9.45% (p=0.000 n=10)
PebbleCHeap/overwrite-8    481.4 ± 5%   494.6 ± 1%        ~ (p=0.436 n=10)
```

The fill row measures RSS after the database is closed, so it is the allocator returning a freed 200 MiB working set to the OS; the read row is the steady-state block cache. This is a synthetic single-process workload, not a node under load, so it argues for running the tagged binary against a real archive node rather than for flipping the default.

`BenchmarkWasmVMCHeap` in `sei-wasmvm` covers the other large C-heap consumer: libwasmvm is a Rust library on the system allocator, so module compilation, instantiation and every host call allocate through whichever malloc the binary links. Unlike Pebble, whose buffers are pooled, this path is malloc-bound and jemalloc is consistently faster on it, same machine and method as above:

```
                          │  libc.txt   │             jemalloc.txt             │
                          │   sec/op    │    sec/op     vs base                │
WasmVMCHeap/compile-8       15.45m ± 3%   14.69m ±  5%   -4.93% (p=0.002 n=10)
WasmVMCHeap/instantiate-8   66.39µ ± 3%   59.57µ ±  9%  -10.27% (p=0.001 n=10)
WasmVMCHeap/execute-8       87.31µ ± 5%   78.86µ ±  2%   -9.68% (p=0.000 n=10)
WasmVMCHeap/query-8         56.36µ ± 2%   51.28µ ± 11%   -9.00% (p=0.004 n=10)
geomean                     266.5µ        243.9µ         -8.49%
```


masih:
Thankfully makes no significant difference for things we care about. Closing

PR #4248, opened 2026-09-18
Move OCC access sets out of the pooled state DB instead of cloning them

masih:
Every speculative OCC execution in `executeTxSpeculative` ends with `stateDB.accessSets()`, which deep-copied both the read set and the write set into fresh maps before the `nativeStateDB` was released back to the pool. The pooled object is reset on release, so nothing ever read those original maps again: the copy cost a map allocation plus one insert per distinct key touched, twice, for every transaction and again for every OCC re-execution. Storage-heavy contract calls pay this for thousands of slots.

`takeAccessSets` now transfers ownership instead: it returns the existing maps and nils out `s.readSet`/`s.writeSet`, remembering their final sizes. `enableAccessTracking` allocates fresh maps when they are nil, using those sizes as capacity hints so a recycled state DB does not rehash its way back up from zero, and `reset` guards nil, so the next acquisition from the pool behaves exactly as before, and `markRead`/`markWrite` are already no-ops on nil sets. `Copy()` keeps using `cloneAccessSet`, since a snapshot must stay independent of its source. The result maps hold the identical contents, so OCC validation is unchanged; the win is the removed 2N inserts per execution; the two map allocations remain but are pre-sized. Validated with `go test ./giga/evmonly/`.


cody-littley (approved):

giga/evmonly/state_db.go
```diff
@@ -741,8 +741,12 @@ func (s *nativeStateDB) enableAccessTracking() {
 	}
 }
 
+// accessSets hands the read and write sets to the caller and stops tracking;
+// enableAccessTracking must be called again before further accesses are recorded.
 func (s *nativeStateDB) accessSets() (map[stateAccessKey]struct{}, map[stateAccessKey]struct{}) {
```
codchen: would it be less surprising to callers to name this function something like `getAndResetAccessSets`?

masih: Removed the PR from merge queue to get this one done

giga/evmonly/state_db.go
```diff
@@ -741,8 +741,12 @@ func (s *nativeStateDB) enableAccessTracking() {
...
-	return cloneAccessSet(s.readSet), cloneAccessSet(s.writeSet)
+	readSet, writeSet := s.readSet, s.writeSet
+	s.readSet, s.writeSet = nil, nil
```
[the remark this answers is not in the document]
masih: Yup let's reflect on this one too

PR #4249, opened 2026-09-18
Memoize sorted field descriptors in hashable canonical marshaling

masih:
`builder.Message` in `sei-tendermint/internal/hashable` re-collects and re-sorts a message's field descriptors on every canonical marshal, and again for every nested message field via `builder.Value`. The sort order is a compile-time invariant of the message type, so this allocation and O(F log F) sort are pure repeated work on the hashing and signing path that `MarshalCanonical`/`ToHash` back for every vote, proposal, batch and certificate.

`sortedFields` now takes the `protoreflect.MessageDescriptor` and memoizes the sorted slice in a package-level `sync.Map` keyed by the descriptor. Descriptors are process-lifetime singletons, so the cache is bounded by the number of message types encoded, and `LoadOrStore` keeps concurrent first calls returning one shared slice. The output is unchanged; the existing hardcoded hashes in `TestMarshalCanonicalRoundTrip` pin that.

A new `BenchmarkMarshalCanonical` on the testonly message goes from ~4190 ns/op, 2584 B/op, 50 allocs/op to ~3640 ns/op, 1848 B/op, 41 allocs/op (about 14% faster, one fewer allocation per message and per nested message).


masih:
benchstat output of the benchmark:

```txt
│     old      │              new                    │
                   │    sec/op    │   sec/op     vs base                │
MarshalCanonical-8   4.753µ ± 5%   3.590µ ± 4%  -24.47% (p=0.000 n=10)

                   │     old      │              new                     │
                   │     B/op     │     B/op      vs base                │
MarshalCanonical-8   2.523Ki ± 0%   1.805Ki ± 0%  -28.48% (p=0.000 n=10)

                   │     old      │              new                    │
                   │  allocs/op   │ allocs/op   vs base                 │
MarshalCanonical-8   50.00 ± 0%    41.00 ± 0%  -18.00% (p=0.000 n=10)
```

sei-tendermint/internal/hashable/hashable.go
```diff
@@ -166,13 +166,25 @@ func isPackable(kind protoreflect.Kind) bool {
 	}
 }
 
-func sortedFields(fields protoreflect.FieldDescriptors) []protoreflect.FieldDescriptor {
+// sortedFieldsCache maps protoreflect.MessageDescriptor to its fields sorted
+// by field number. Descriptors are process-lifetime singletons, so the cache
+// is bounded by the number of message types encoded.
+var sortedFieldsCache sync.Map
```
wen-coding, line 172: Can an attacker make this cache arbitrary large?

wen-coding (approved):

PR #4251, opened 2026-09-18
Build flatkv physical keys into reusable buffers, not fresh slices

masih:
On the synchronous block-commit path, `classifyAndPrefix` (called from `CommitStore.ApplyChangeSets`) built each changed key's physical form with the allocating `ktype.EVMPhysicalKey` / `ktype.ModulePhysicalKey` constructors and then wrapped the result in `string(...)`, so every changed key cost two allocations: the intermediate `[]byte` and the string copy. The intermediate slice is dead as soon as the string exists.

This adds `ktype.AppendModulePhysicalKey` as the append-based twin of `ModulePhysicalKey` and has `classifyAndPrefix` append both EVM and module keys into one scratch slice reused across all pairs, paying only for the final `string` conversion. The produced key bytes are unchanged, which the new `TestAppendPhysicalKeysMatchAllocating` pins against the allocating constructors for every EVM key kind and for module keys. The `CommitStore.Get` helpers in `store_read.go` are deliberately left on the allocating constructors: the key is passed through the `view.View.Get` interface, so escape analysis moves a local `[physKeyBufLen]byte` buffer to the heap anyway (`go build -gcflags=-m` reports `moved to heap: buf`), and switching them would not remove an allocation.

`BenchmarkClassifyAndPrefix` (2,000 storage keys plus 500 module keys) goes from 4516 allocs/op and 510 KB/op to 2518 allocs/op and 382 KB/op, i.e. one allocation per changed key removed as expected. The flatkv package tests pass unchanged under `scripts/ramtest.sh`.


sei-db/state_db/sc/flatkv/store_read.go
```diff
@@ -201,7 +201,8 @@ func (s *CommitStore) getAccountData(keyBytes []byte) (*vtype.AccountData, error
 	if len(keyBytes) != ktype.AddressLen {
 		return nil, fmt.Errorf("accountDB: expected key length %d, got %d", ktype.AddressLen, len(keyBytes))
 	}
-	physKey := ktype.EVMPhysicalKey(ktype.EVMKeyAccount, keyBytes)
+	var buf [physKeyBufLen]byte
```
alrevuelta: mind checking this helps? i suspect go may store this in the heap, so not sure it will reduce allocs. afaik it cant be kept in the stack because you dont have any guarantee that .Get wont store it. and since the buf is dropped after the function returns that will lead to a dangling pointer (which cant happens, and using the heap is what prevents that)

sei-db/state_db/sc/flatkv/store_apply.go
```diff
@@ -328,6 +328,7 @@ func classifyAndPrefix(changeSets []*proto.NamedChangeSet) (map[keys.EVMKeyKind]
 		return m
 	}
 
+	var keyBuf []byte
```
alrevuelta: maybe
`keyBuf := make([]byte, 0, physKeyBufLen)`
?

so the scratch buffer is preallocated beforehand?

alrevuelta (approved):
lgtm but would like @cody-littley input.

cody-littley (approved):

PR #4253, opened 2026-09-18
evmonly/rpc: serve eth_subscribe("newHeads") from the Giga executed-block watch

bdchatham:

Follow-up to #4252 (WS transport). Adds `eth_subscribe("newHeads")` to the EVM-only RPC so sei-load's `OutcomeTracker` / `BlockCollector` (`ethclient.SubscribeNewHead`) work against an evmonly node unchanged. `eth_getBlockReceipts` is deliberately left for a separate PR.

Autobahn's `gigaRouterCommon.executeBlock` runs `FinalizeBlock`+`Commit` directly and never publishes `NewBlockHeader` on the tendermint EventBus, so the head source has to come from the Giga router itself:

```go
// sei-tendermint/internal/p2p  (same shape as the existing liveAddrVersion / data.State.Anchor() watches)
type GigaRouter interface {
    ...
    ExecutedBlocks() utils.AtomicRecv[atypes.ExecutedBlocks] // Store(Load().Push({b.GlobalNumber, finalizeBlockGasUsed(resp)})) after app.Commit in executeBlock
}
type atypes.ExecutedBlock  struct { Number GlobalBlockNumber; GasUsed uint64 }
type atypes.ExecutedBlocks struct { /* immutable window of the last 128 ExecutedBlock */ }  // Push / Latest / Get(n)

// sei-tendermint/internal/rpc/core/evm_heads.go — forwards the watch, ErrNewHeadsUnavailable without Giga
func (env *Environment) ExecutedBlocks() (utils.AtomicRecv[atypes.ExecutedBlocks], error)

// giga/evmonly/rpc: Backend gains ExecutedBlocks(); the stream is
for next := executed.Load().Latest().Number+1; ; next++ { w := executed.Wait(ctx, .Latest().Number >= next); Block(next) -> encodeHeader(w.Get(next).GasUsed) -> Notify }
```

`subscribeAPI.NewHeads` (registered under `eth`, so geth maps it to `eth_subscribe` kind `newHeads`) exits when `rpcSub.Err()` closes (eth_unsubscribe / connection drop). A failed block lookup logs and skips that height instead of ending the stream.

Headers are emitted at commit, with `gasUsed` taken from the FinalizeBlock `TxResults` the router already sums (`finalizeBlockGasUsed`) — the same source the full node's Autobahn newHeads uses (`evmrpc.encodeCommittedBlock`). There is no receipt-store dependency on the hot path, so a lagging store cannot stall the subscription. The watch value is an immutable window of the last 128 committed blocks, so a subscriber that falls behind still reads each height's commit total; only a height older than the window falls back to the receipt-store sum (0 if not yet landed). The stream goroutine sits outside go-ethereum's per-call recovery, so it recovers panics itself (logged, subscription ends) and skips a nil block rather than dereferencing it. The watch is seeded from `cfg.App.LastBlockHeight()` so a post-restart subscriber does not replay history.

`encodeBlock`'s header fields are extracted into `headerFields`/`encodeHeader` so the payload is exactly the header portion of `eth_getBlockByNumber`; `eth_getBlockBy*` behaviour is unchanged.

Not in scope: `logs` / `newPendingTransactions`; per-connection subscription caps (each subscription is one goroutine on a shared watch, no per-subscriber buffer — the WS connection cap from #4252 is the bound).

- `TestGigaRouter_FinalizeBlocks` (p2p): on the real 4-validator execution path, `ExecutedBlocks().Wait` reaches `LastCommittedBlockNumber()` and the height it yields is readable via `BlockByNumber`. `TestBuildDataStateStartsRecoveryAtAppTip` checks the watch is seeded at the app tip.
- `TestNewHeadsSubscriptionStreamsHeadersOverWebsocket`: real `ethclient.SubscribeNewHead` over the WS handler with an empty receipt store; decoded `types.Header` carries `Number=7`, `GasUsed=43500` (from the commit), `GasLimit`, `Time`.
- `TestNewHeadsCatchesUpFromExecutedWindow`: watch jumps 6→8 with an empty receipt store; blocks 7 and 8 both carry their commit `gasUsed`, delivered in order.
- `TestExecutedBlocks` (autobahn/types): window eviction and `Get` bounds.
- `TestNewHeadsRejectedOverPlainHTTP`: `ErrNotificationsUnsupported` over HTTP.
- `go test -race ./giga/evmonly/rpc/ ./sei-tendermint/internal/p2p/ -run FinalizeBlocks`, `go test ./sei-tendermint/internal/rpc/core/`, `go build ./giga/... ./sei-tendermint/...`, `go vet`, `make fmtcheck` clean.
- Not run against a live node.

Link to Devin session: https://app.devin.ai/sessions/977f51da5bcb49458bedca458921f125
Open in Devin Desktop: https://app.devin.ai/desktop/session/977f51da5bcb49458bedca458921f125?variant=devin
Requested by: @bdchatham


bdchatham:
@seidroid review

bdchatham:
@seidroid review

PR #4254, opened 2026-09-18
evmonlyapp: land block state commits behind FinalizeBlock (pipelined executor path)

bdchatham:

First production integration of the executor's pipelined path from #4219. Until now only `evmonly-loadtest` drove `PrepareBlock`/`ExecutePreparedBlock`; the node's `evmonlyapp.FinalizeBlock` called the synchronous `ExecuteBlock`, which waits for the block's state commit before returning. On giga-testnet-2 that wait is the ~26% "storage" slice of the Autobahn main loop (encode + commit_sc + flatkv apply).

Shape of the change in `sei-tendermint/internal/evmonlyapp/app.go`:

```go
// FinalizeBlock
-result, err := executor.ExecuteBlock(ctx, req)
+prepared, err := executor.PrepareBlock(ctx, req)
+result, err := executor.ExecutePreparedBlock(ctx, prepared)   // returns once commit(n) is *started*
```

The executor already guarantees ordering: `ExecutePreparedBlock(n+1)` lands commit(n) before starting commit(n+1), and executes n+1 against `pendingOverlay` so it sees n's writes before they are in the store. The app hash is a hash of the changeset, so it is available before the commit lands — `Commit` still only advances the in-memory cursor and does not wait.

What has to settle before touching the store (the new sync points, all via `AwaitCommits`):
- `EvmNonce` / `EvmBalance` (`openSettledView`) and `EvmCall` — CheckTx / RPC readers of committed state must include the last finalized block.
- `abandonPending` — decides whether a failed block's cursor is durable by reading the store version, so the in-flight commit must be final first.
- Node shutdown: `nodeImpl.closeGigaStorage` settles through a new optional `proxy.Proxy.AwaitCommits()` (implemented if the wrapped app exposes `AwaitCommits`) before closing the storage manager, so the store is not closed under a write.
- Restart: unchanged — `loadEVMOnlyCursor` reads the store's own version, so a block finalized-but-not-committed (or whose commit was cut off at shutdown) is simply replayed.

Failure surfacing: the executor latches the first failed commit; it is returned from the next `ExecutePreparedBlock`, so a failed background commit halts the node at the next `FinalizeBlock` (the same outcome as a synchronous commit failure, one block later), and from `EvmCall`/`abandonPending`/`AwaitCommits`.

Not in scope: the pipelined sender recovery from #4219 (a loadtest harness stage, not an executor feature) and any OCC/access-list scheduling.

Draft: intended for a benchmark A/B on the harbor giga-testnet before proposing to merge; I'd like the chain team's eyes on the `Commit`-does-not-wait decision and the `openSettledView` error handling (a failed commit is not reported to nonce/balance readers, only to the block loop).

- `scripts/ramtest.sh ./sei-tendermint/internal/evmonlyapp/... ./sei-tendermint/internal/proxy/... ./sei-tendermint/node/... -race -count=1` green; evmonlyapp additionally `-race -count=3`.
- New `TestEVMOnlyApplicationReadsSettleBehindFinalizeBlock`: four dependent-nonce blocks, nonce visible via `EvmNonce` right after each `FinalizeBlock` (before `Commit`), next block builds on the prior overlay, store version is 4 after settling. The existing finalized-but-not-committed restart test is unchanged and passes.
- Test fixture: storage now lives on a context that outlives the test body (the last commit is still landing when `t.Context()` is cancelled), and `closeEVMOnlyTestApp` settles before `storage.Close()`.
- `golangci-lint run` on the touched packages (0 issues), `make fmtcheck`, `go vet`, `go build ./...`.
- Not yet measured on a running chain.

Requested by: @bdchatham


bdchatham:
@seidroid review

bdchatham:
@seidroid review

bdchatham:
@seidroid review

PR #4257, opened 2026-09-18
Require a Giga storage manager for every Autobahn node

wen-coding:
Autobahn only opened a `GigaStorageManager` on the EVM-only path. Everyone else got a standalone BlockStore from `selectAutobahnBlockStore`, which could be in-memory when `persistent_state_dir` was unset, and Close had two owners. That split is the wrong place to hang AppHash and restart recovery; validators and fullnodes need the same durable ledger.

`prepareApplication` now opens storage whenever `AutobahnConfigFile` is set, via `openAutobahnStorageManager` and `AutobahnStorageConfig` (FlatKV, receipts, BlockDB; SS off). `createRouter` requires that manager and passes `manager.BlockStore()` into `buildGigaRouter`. The manager is the only closer, including if `wrapApplication` fails. EVM-only wrapping errors when storage is missing instead of falling through to the unwrapped app. Seed with Autobahn is refused (`errAutobahnSeed`). `gen-autobahn-config` refuses an empty `--persistent-state-dir`. CometBFT without an Autobahn file is unchanged (`None`).

This is node startup only: no wire or state-machine change, but Autobahn can no longer boot without `persistent_state_dir`. Node tests cover seed refusal, Autobahn-without-EVM-only opening storage, and EVM-only without storage; the layout helper is tested in `sei-db/config`.


codchen (approved):

sei-tendermint/cmd/tendermint/commands/gen_autobahn_config.go
```diff
@@ -93,12 +96,9 @@ Output is written to the file specified by --output.`,
...
 			// operator action. node/setup.go rootifies the relative path against
-			// cfg.RootDir at load time. Passing --persistent-state-dir= (empty)
-			// disables persistence and runs both consensus and data layers
-			// in-memory only.
-			if persistentStateDir != "" {
-				cfg.PersistentStateDir = utils.Some(persistentStateDir)
-			}
+			// cfg.RootDir at load time. An empty --persistent-state-dir is refused:
+			// Autobahn nodes require on-disk Giga storage.
+			cfg.PersistentStateDir = utils.Some(persistentStateDir)
```
shemnon: If it's undonditional shouldn't we move ito into the cfg decleration at lines 89-95?

sei-tendermint/node/setup.go
shemnon: If persistent state dir is now required, shouldn't we error here instead?

Meta question - should persistent state dir migrate out of an optional?

Or does this need to stay as optional for cases we don't enable giga/evmonly/autobahn?

sei-tendermint/cmd/tendermint/commands/gen_autobahn_config.go
wen-coding: We wish to have this option back some day. But you are right, when that happens we can make it an Option again, I'll change it.

sei-tendermint/node/setup.go
wen-coding: Yeah, we are still holding on to the thin hope that we can have a mem-only storage manager in the future. But it's fair, if we don't have it now, we should change it.

wen-coding: Done

sei-tendermint/cmd/tendermint/commands/gen_autobahn_config.go
wen-coding: Done

PR #4258, opened 2026-09-19
mempool/evmonlyapp: read nonce/balance through the pending overlay; fetch first-seen accounts outside the store lock

bdchatham:

Follow-up to #4254. After the pipelined commit rolled out to giga-testnet-2, producer admission and CheckTx wait were unchanged but nonce-lookup p99 went 0.3 ms → 4.5 ms: `EvmNonce`/`EvmBalance` for a first-seen account now `AwaitCommits` behind the in-flight commit, and `txStore.insert` made those calls while holding the `inner` write lock, so one cold account stalled every other mempool operation on the node.

Two independent fixes:

**1. `giga/evmonly`: `Executor.ReadLatestAccount(addr) (LatestAccount, error)`**
Reads balance+nonce through the same `pendingOverlay` block N+1 executes against, so the in-flight block's writes are served from memory and the read never waits on the store. Semantics:
- returns the latched failure if a commit has failed (never state that omits the failed block);
- pending is read *before* the view is opened, and re-checked after: if a newer block started its commit in between, the view may already hold its writes and replaying the older pending set over them would go backwards, so the read restarts (bounded: one commit in flight).

`evmonlyapp.EvmNonce/EvmBalance` → `latestAccount()`, which uses `ReadLatestAccount` when the executor exists and otherwise (pre-InitChain, or failed commit) falls back to the existing `openSettledView` path, so the failed-commit log/behaviour is unchanged. The funded-default for missing accounts is applied by the executor's `missingState`, same as execution.

**2. `mempool/tx.go`: first-seen account fetch moved out of the lock (gprusak's TODO)**
```
Insert(wtx):
  prefetched := prefetchAccount(wtx)     // RLock: known? -> none; else record accountsEpoch, unlock, app.EvmBalance/EvmNonce
  Lock:
    installPrefetched(inner, prefetched) // skip if epoch moved (Update/Clear reset accounts) or another inserter won
    insert(inner, wtx, true)             // unchanged; falls back to in-lock fetch only if the prefetch was discarded
```
`accountsEpoch` is bumped wherever `accounts` is reset (`compact(clearAccounts)`, `Clear`), so a value fetched for height H is never installed after `Update(H+1)`. `compact`/`Update` still use the in-lock `insert` path unchanged.

- `giga/evmonly`: `TestReadLatestAccountSeesTheBlockWhoseCommitIsInFlight` (store view never advances, read must come from the overlay), `TestReadLatestAccountReportsAFailedCommit`, `TestReadLatestAccountRestartsWhenABlockLandsUnderIt` (block 42 lands in the reader's OpenView; fails if the re-check is removed).
- `mempool`: `TestTxStore_InsertFetchesFirstSeenAccountOutsideTheLock` (gated app; `NextNonce` and another `Insert` complete while the cold fetch is blocked — deadlocks on the old code), `TestTxStore_InsertDiscardsAccountFetchedBeforeUpdate` (fails without the epoch check).
- `go test -race` on `giga/evmonly`, `internal/evmonlyapp` (ramtest), `internal/mempool`: green. golangci-lint v2.13.2 on the touched packages: 0 issues. `make fmtcheck` clean.
- Not yet measured on giga-testnet-2; after rollout the check is `nonce lookup` p99 returning to ~0.3 ms with admission/CheckTx unchanged.

Requested by: @bdchatham


bdchatham:
@seidroid review

PR #4260, opened 2026-09-19
autobahn/evmonlyapp: prepare block n+1 while block n executes

bdchatham:
Second half of the sender-recovery pipelining, after #4259's measurement. Every block a validator executes arrives as raw tx bytes that must be RLP-decoded and ECDSA-recovered before the EVM can run them; CheckTx caches the sender only for txs this node admitted (~1/40 at 40 validators), so the rest is recovered inline in `FinalizeBlock` on the single-threaded Autobahn main loop (#4259's `prepare` phase). That work depends only on the bytes, not on state, so it does not have to wait its turn in the serial loop.

The router's `runExecute` (`giga_router_common.go`) becomes a two-stage `scope.Run`: a fetcher task fetches `GlobalBlock(n)`, builds the state-independent part of the FinalizeBlock request (`finalizeRequest`: txs, hash, chain id, height, time), calls the new optional `app.PrepareBlock`, and hands the block over an unbuffered channel, so it is exactly one block ahead; the execute loop keeps `FinalizeBlock`, hash-vault commit, `Commit`, `PushAppHash` and prune in their existing order. `Header.ProposerAddress` is deliberately filled in by `executeBlock` (`proposerAddress()`) immediately before `FinalizeBlock`, because on the non-EVM-only path `GetValidators()` reads `BaseApp.checkState`, which `Commit` replaces; it must not be read from the fetcher. The proxy exposes `Proxy.PrepareBlock` through an optional `blockPreparer` interface (same pattern as `commitSettler`), a no-op for applications that do not implement it.

In `evmonlyapp`, `PrepareBlock` reads the executor from `settler` (no executor lock, so it runs concurrently with `FinalizeBlock`), peeks the CheckTx sender cache without consuming it, runs `evmonly.Executor.PrepareBlock` (context validation + parse, touches no state) and stores `{height, hash, txs}` in a one-slot `prepared` mutex. `FinalizeBlock` uses the slot only when height and hash match, then forgets the cached senders and goes straight to `ExecutePreparedBlock`; on a mismatch the slot is left alone and the existing `take_senders`/`prepare`/`execute` path runs. `PrepareBlock` returns nothing but `ctx.Err()`, so any error that would fail the block is reproduced by `FinalizeBlock` and the failure path (`abandonPending`, router error) is unchanged. Only `Number`/`Time` reach the decoded txs (via the signer); the parent-derived block context (`ParentHash`, `PrevRandao`, `GasLimit`) is still filled in by `FinalizeBlock`, so preparing ahead cannot bake in stale cursor state.

Invariants kept: block order, one commit in flight, background-commit failure surfacing on the next state-touching call, `PhaseTimer` touched only under the executor lock. Hits and misses are counted in `evmonly_finalize_prepared_blocks_total{prepared}`; the rollout check is that `prepared="true"` ≈ blocks/s and the main loop's `execution` share drops by roughly the former `prepare` phase; also watch the per-block `execution` phase itself, since block n+1's ECDSA recovery now competes for cores with block n's OCC workers and could offset part of the gain (the `consensus` slice now measures waiting on the fetcher, ~0 when the next block is already prepared). Validated with new `evmonlyapp` tests — prepared and unprepared blocks finalize to the same app hash and tx results, a wrong-hash preparation is bypassed and kept, an undecodable block prepared ahead is still reported by `FinalizeBlock`, preparing before InitChain is a no-op, and `PrepareBlock(n+1)` racing `FinalizeBlock(n)` over several blocks under `-race` matches a sequential run — plus the proxy and p2p suites, where `TestGigaRouter` now asserts the delivered blocks are contiguous and every header's `ProposerAddress` is a current validator, scoped golangci-lint and `make fmtcheck`.

Requested by: @bdchatham


bdchatham:
@seidroid review

bdchatham:
@seidroid review

PR #4261, opened 2026-09-19
evmonly: parallelize OCC validation and merge behind the serial acceptance barrier

bdchatham:

On giga-testnet-2, `occ_validate` + `occ_merge` are ~34% of main-loop wall time and ran entirely on the calling goroutine, while conflicts are ~0.85 per ~1,800-tx block. Nearly all of that time is bookkeeping for transactions that are accepted as they stand. This PR keeps acceptance strictly in block order (the same frontier, same rerun/incarnation/fallback rules) but moves the work behind it onto the existing `occWorkerPool`.

**Validation** (`validateBlockSTM`):
- The write index (`stateAccessIndex`) is now sharded by address (64 contiguous address ranges) and records a `txIndexSpan{first,last}` per key instead of a single index. All speculative results are indexed up front in parallel, each worker filling the shards it owns.
- Conflict checks go from `conflictsWithAfter(key, sourcePrefix)` to `conflictsWithin(key, sourcePrefix, txIndex)`: a write invalidates a result only if it sits in `[sourcePrefix, txIndex)`. With every result indexed, this is the same question the serial frontier asked, because when the frontier is at `i` every write below `i` is an accepted one. A rerun's writes are added; the previous incarnation's stay. Both the span and the stale entries can only produce false positives (an unnecessary rerun, which then validates trivially against its exact prefix), never a missed conflict.
- `acceptValidatedPrefix` runs a parallel pass over `[nextToValidate, n)`: workers evaluate `stmFrontierAccepts` (error, gas, read/write conflicts against the precomputed cumulative gas) and lower a shared `stop` to the first index that would not be accepted. Everything before `stop` is folded into the prefix with `applyRange` (each worker applies the shards it owns; per-address state lives entirely in one shard, so per-shard order equals block order). The serial frontier then handles only the transaction at `stop` — recording conflicts, scheduling the rerun, or surfacing the gas/execution error exactly as before.
- Each incarnation records the shards it touched as a bitmask (`occTxExecution.shards`, computed on the worker that produced it), so index/apply workers skip whole results that hold nothing in their shards instead of scanning every result's maps.
- Dependency chains that would make every parallel pass accept ~nothing switch to a serial stretch that doubles each time it happens (`serialUntil`), so worst-case blocks don't pay pool wake-ups per transaction. Blocks under 64 pending results skip the pool entirely.

**Merge** (`mergeOCCResults`):
- `blockSTMState` is sharded the same way; `changeSetIntoParallel` computes each shard's changeset (with its own `baseAccounts` row cache, so each account is still read once) on the pool into pooled per-shard fragments and concatenates shards in order, which is canonical address order — output is byte-identical to the serial `ChangeSetInto`. A prefix under `occMinParallelMergeKeys` (256) keys is merged serially straight into the pooled changeset. The tx/receipt loop (cumulative gas, log indexes) stays serial and unchanged.
- The previous `prefetchBaseAccounts` pass is subsumed: base rows are now read inside the per-shard merge.
- A closed pool during merge falls back to sequential execution like the other `errOCCWorkerPoolClosed` sites.

What does not change: transaction order, cumulative gas / block gas-limit / overflow checks, `occMaxTxIncarnations`, only-earliest-invalid-rerun, sequential fallback reasons, final changeset semantics. Conflict counts are still deterministic (a pure function of the block); they can be slightly higher than before due to span false positives.

Known shape: a hot contract's storage lands in one shard, so that shard's merge/apply runs on one worker. `BenchmarkExecuteTransferBlock` gained an `erc20_single_contract` case (1,000 transfers on one ERC-20) to keep that measured: 19.9 → 18.6 ms/op vs `giga-1`. Whole-block local benchmarks (8 cores, incl. parse — parse is ~50% of it): `conflict_free` 15.1 → 12.9 ms/op, `hot_recipient` and `same_sender_nonce_chain` unchanged within noise, allocations flat. The rollout check is `occ_validate` + `occ_merge` share of the main loop on testnet-2 and executed tx/s.

- New `TestExecutorOCCLargeMixedBlockMatchesSequential`: ~330-tx block (independent transfers, same-sender nonce chains, hot recipient, a late reader of the hot balance that must be rerun after the parallel pass) — asserts `Txs`, `Receipts`, `GasUsed` and the full `ChangeSet` equal the sequential executor's, and that a rerun actually happened.
- New `TestParallelMergeResolvesEveryTouchedAccountOnce` (replaces the prefetch tests): parallel merge output equals serial, canonical order, one row read per account.
- New `occ_shards_test.go`: `conflictsWithin` lower/upper bounds and span over-approximation, `occShardOf` monotonic and covering, `occShardsOwnedBy` partitioning, `cumulativeGasFrom` overflow truncation, `firstUnacceptedResult` returning the lowest rejection under concurrency (error, gas limit, conflict, source-prefix), `touchedShards`, and the serial-merge guard against a closed pool.
- Existing conflict-matrix / source-prefix / gas / incarnation tests updated only for the `txIndex` parameter.
- `go test -race ./giga/evmonly/...` and `./sei-tendermint/internal/evmonlyapp/...` green; `make fmtcheck` clean; `golangci-lint run ./giga/evmonly/...` 0 issues (v2.13.2, Go 1.27.1).

Requested by: @bdchatham


bdchatham:
@seidroid review

PR #4262, opened 2026-09-19
evmonly/evmonlyapp: size the next block's decode from its cost and the time it has; time PrepareBlock

bdchatham:

Since #4260, `PrepareBlock` decodes block n+1 (RLP decode + ecrecover for senders CheckTx did not see) on `GOMAXPROCS` workers while block n's OCC speculation holds a worker per processor. The `occ_speculate` share rose 0.19 → 0.36 s/s on that roll, so the prepare pool is competing with the block on the critical path for work that only has to finish *before block n does*.

Rather than a fixed share of the processors, the decode is sized per block from the work and the time it has:

```go
// giga/evmonly
func (e *Executor) PrepareBlockWithin(ctx, req BlockRequest, budget time.Duration) (PreparedBlock, error)
//   workers = clamp(ceil(len(txs) × perTxCost / budget), 1, min(ParseWorkers, len(txs)))
//   budget == 0, or no estimate yet  → every parse worker (unchanged behaviour); budget == 0 does not feed the estimate
//   perTxCost: elapsed × workers / txs over earlier budgeted decodes on this executor —
//              a decode above the estimate replaces it at once, one below decays it by 1/8
func (e *Executor) PrepareBlock(ctx, req) = PrepareBlockWithin(ctx, req, 0)

// evmonlyapp
budget = executeEstimate / 2   // executeEstimate: EWMA (1/8) of ExecutePreparedBlock wall time over prepared blocks
```

So on today's testnet-2 blocks (~1,850 txs, ~50 µs/tx ecrecover, ~20 ms execution) it decodes on ~9–13 workers of 32 (the measured cost includes contention, so the fixed point sits a little above the nominal count — the safe side); on a small block it uses 1–2; on a 5k-tx block it grows; and on a different box or tx mix the per-tx estimate re-fits itself. The asymmetry is deliberate: an undersized decode delays the block that needs it (the fetcher hands the prepared block over synchronously), an oversized one only spends processors, so the estimate jumps up on the first poorly-cached block and comes down slowly. The budget is smoothed for the same reason — a single short block dents it rather than zeroing it, which would hand the next decode every worker. The unprepared `FinalizeBlock` path (`executeBlockPipelined`) keeps `PrepareBlock` with no budget, so a decode that *is* on the critical path still uses every worker. `Config.ParseWorkers` remains the ceiling (still `GOMAXPROCS` in production wiring).

Also adds an `evmonly_prepare` phase timer (`parse`) around the decode so prepare wall time vs block time is visible after the roll; `evmonly_prepare_phase_duration_seconds_total{phase="parse"}`.

Measurement after roll: `occ_speculate` s/s and executed tx/s vs the #4261 baseline (~94k/validator), and `evmonly_prepare` parse wall staying well under `evmonly_finalize` execute wall. If parse wall approaches execute wall, a mid-decode re-check of the remaining budget is the next step.

- `giga/evmonly`: `parse_sizer_test.go` — every worker until the first estimate, fits the estimated cost in the budget (5 → 3 → 1 workers as the budget grows; capped at the ceiling / tx count), no budget uses every worker, estimate rises at once and decays gradually, ignores empty observations. `go test -race ./giga/evmonly/` green.
- `sei-tendermint/internal/evmonlyapp`: `TestPrepareBudgetIsAShareOfTheTypicalExecution`, `TestExecuteEstimateSmoothsOverBlocks`; `go test -race ./internal/evmonlyapp/` green (prepared-hit and unprepared-fallback paths both covered by the existing tests).
- `golangci-lint run` on both packages, `make fmtcheck` clean.

Requested by: @bdchatham


bdchatham:
@seidroid

PR #4264, opened 2026-09-19
evmonly: persist receipts and encode state changes behind the block; Commit advances the head once the receipts have landed

bdchatham:

Third of the three execution-side PRs (after #4262, #4263). Takes the rest of the executor's storage tail off the block loop, the way #4254 did for the state commit, while keeping "a block RPC serves has readable receipts".

**Before**, after execution the loop did, serially: encode state changesets → block encoder → encode receipts → `SetReceipts` → await previous commit → start this block's commit in the background. On testnet-2 the two encodes were ~0.08 s/s of loop time.

**After**, the loop does: [await previous commit] → block encoder (stays on the loop: what it stages is what the caller reports for the block) → await previous commit → start the pipeline → return. The pipeline goroutine then, per block, in order:

```
persistReceipts:   encode receipts (parallel) → SetReceipts → WaitForPendingWrites → check LatestVersion >= block
release the pooled BlockResult
commitStateChanges: encode state changesets (from a clone) → append block encoder's → CommitStateChanges
```

`NamedChangeSetEncoder` now runs after the previous block has landed and before this one starts, so an encoder that reads the store (the storage-clear expansion) sees every earlier block and none of this one; the `encodingReadsTheStore` special case that used to serialize such blocks is gone. It runs on a clone, so the pooled result is free as soon as receipts are encoded.

**Head advances after receipts land.** `evmOnlyApplication.Commit` calls the new `Executor.AwaitReceipts()` before advancing the cursor. The router publishes the block (and RPC `latest` follows it) only after `Commit`, so `latest` never points at a block whose receipts `eth_getTransactionReceipt` / `eth_getBlockReceipts` cannot read. Previously the litt store's own async writer meant receipts could trail the visible head by up to its queue depth; that window is closed, not widened.

To make "land" real rather than "enqueued", `littReceiptStore` now implements the existing `seidbtypes.PendingWriteWaiter` capability: each queued write carries a `landed` channel the writer closes after applying it (or skipping it after a latched failure), and `WaitForPendingWrites` waits on the last admitted one. The executor asserts `LatestVersion() >= block` afterwards so a write dropped after a latched failure is reported for *this* block rather than by the next `SetReceipts`.

**Failure semantics are unchanged in shape:** a receipt or state persistence failure latches (`pipelineFailure`), `Commit` returns it if the receipt stage failed, and the next `FinalizeBlock` returns it otherwise (`persist block: …`). Once latched the executor refuses further blocks. Cancelling the request that ran the block no longer abandons its persistence (`context.WithoutCancel`): `Close` waits for it, and the block's outcome is reported through the pipeline rather than by dropping it.

**Cost to watch after the roll.** `Commit` now waits for this block's receipt encode + litt write, which overlap the router's `vault_commit` (#4263) but no longer with the next block's execution. New `evmonly_pipeline` phase timer (`encode_receipts`, `write_receipts`, `await_receipts`, `encode_changesets`, `commit_state`) and the loop's `evmonly_block` timer (`encode_block_changesets`, `await_commit`, `start_commit`) show where the time went; if `await_receipts` is large the litt write itself is the next target.

- `giga/evmonly`: `TestAwaitReceiptsReturnsBeforeTheStateCommitLands` (receipts readable and pool lease returned while the state commit is still gated), `TestFailedReceiptWriteFailsTheBlockBeforeItsStateCommit` (no state commit after a failed receipt write; failure surfaces), `TestBackgroundEncoderSeesTheBlocksOwnChanges` (encoder gets a clone, not the recycled pooled result); existing store/pipeline/overlay/determinism tests unchanged except `TestExecutorReturnsReceiptStoreError`, updated for the write now landing behind the block and latching.
- `sei-db/ledger_db/receipt`: `TestLittIdxWaitForPendingWritesLandsEveryQueuedBlock`.
- `sei-tendermint/internal/evmonlyapp`: `TestEVMOnlyApplicationCommitReturnsWithTheBlocksReceiptsReadable` (receipt for height N readable from the store the moment `Commit` returns, N=1..3); existing failed-commit fallback and reopen tests unchanged.
- `scripts/ramtest.sh ./giga/evmonly/... -race`, `./sei-db/ledger_db/receipt/ -race`, `go test -race -count=8 ./internal/evmonlyapp/`, `go vet`, `golangci-lint run` on touched packages, `make fmtcheck` — all green.

Requested by: @bdchatham


bdchatham:
@seidroid review

PR #4265, opened 2026-09-20
evmonly: let a node run without a receipt store (autobahn `enable_receipt_store`)

bdchatham:
Since #4264, `Commit` on an EVM-only node waits for the block's receipts to land in the litt store before advancing the RPC-visible head. On testnet-2 that wait (`app_commit` ≈ 0.16 s/s, `await_receipts` ≈ 0.17 s/s) outweighed the loop time #4264 saved, and the 40 validators went from ~96k to ~87k tx/s. Validators serve no receipt reads, so both the write and the wait buy them nothing; `GigaStorageConfig.WithValidatorMode()` already turns the receipt DB off, but `NewValidatorStorageConfig` immediately forced it back on and the executor and RPC server refused to start without one.

`NewValidatorStorageConfig(homePath, receipts bool)` now takes the choice, and `node/setup.go` passes a new `AutobahnFileConfig.EnableReceiptStore` (`enable_receipt_store`, `utils.Option[bool]`, default true, so nothing changes for a node that does not set it). With it off, `bootstrap.GigaStorageManager.ReceiptDB()` is nil as its docs already allow; the executor drops `errMissingReceiptStore`, `startPipelineCommit` skips `persistReceipts` and closes the receipt gate at once, so `AwaitReceipts` returns immediately and the pipeline goes straight to `commitStateChanges`. Block results still carry receipts, they are just not persisted. `evmonlyrpc.newHandler` accepts a nil store: `eth_getBlockBy*` still serve, with `gasUsed` 0 as `blockGasUsed` already did outside the receipt range, and `eth_getTransactionReceipt`/`eth_getTransactionByHash` return a new `ErrNoReceiptStore` rather than `null`, which a client would read as "not mined". Registering a reduced eth namespace was the rejected alternative: sei-load and the proxy still need block, send and nonce methods on validators, and a missing method is harder to diagnose than an explicit error.

No consensus or on-disk format impact: the app hash and state commit are untouched, and a node that keeps receipts behaves exactly as before. Rollout is a config change per validator (`enable_receipt_store = false`), leaving RPC nodes on. Turning it off on a node whose receipt directory already holds blocks is refused at startup (`bootstrap.ErrDisabledReceiptStoreHoldsBlocks`, naming the directory): a disabled store would otherwise sit out recovery at its old head, and turning receipts back on later would make `findTargetRecoveryHeight` converge state back onto that stale head. With the directory removed, a later re-enable is the empty-store case recovery already handles and receipts fill from the current height — so an RPC node that ever ran without receipts has a gap, which is why the default stays true. Validated with new tests for the storage config, the executor without a receipt store (state still commits, pool result released, `AwaitReceipts` returns), the RPC handler over a nil store, the disable → advance → re-enable restart loop in `sei-db/bootstrap`, and `prepareApplication` producing a nil `ReceiptDB()` from the file config; race tests for `giga/evmonly`, `giga/evmonly/rpc`, `evmonlyapp` and `node` pass.

Requested by: @bdchatham


bdchatham:
@seidroid review

PR #4266, opened 2026-09-20
evmonly: start the receipt write as soon as execution returns; the state commit waits for it

bdchatham:

Since #4264 the receipt encode+write for block N is kicked off at the very end of `ExecutePreparedBlock` (`start_commit`), and `Commit` then waits for it to land before advancing the RPC-visible head. On testnet-2 that wait shows up as `app_commit` ≈ 0.16 s/s: the litt writer's ~3.5 ms/block used to trail off-loop and now sits on the critical path, because nothing overlaps it.

This PR starts the receipt write the moment execution returns, so it runs under the rest of the block's tail (block-changeset encoder, `await_commit` for block N-1, `start_commit`) and the router's tail (vault commit, PushAppHash) instead of after them. The head-advance invariant is unchanged: `Commit` still calls `AwaitReceipts` and the state commit still waits for the block's receipts before committing, so the store never holds a block whose receipts cannot be read.

```
ExecutePreparedBlock(N):
  execute
  receipts := startReceiptWrite(N)          // new: was inside startPipelineCommit
  [await_commit(N-1) | encode_block_changesets]   // receipt write overlaps these
  startPipelineCommit(N, receipts):          // goroutine: <-receipts.done; commitStateChanges
```

`startReceiptWrite`:
- runs one goroutine per block that first waits for the previous block's receipt write, so receipts still land in block order and the store's `LatestVersion` keeps meaning "every block up to here"; a failed previous write fails this one too (the pipeline is already latched at that point).
- records itself as `pipelineReceipts` immediately, so `AwaitReceipts` finds it even if the block later fails on the loop (block encoder error, ctx cancel). `Close` now also waits on it, since such a block has no commit to drain. Receipts for a block that failed after execution can therefore be persisted ahead of state — same class as the existing "state failure leaves receipts behind" case, healed by re-execution on restart; README updated.
- `startPipelineCommit` no longer takes a ctx or retains the result; the receipt goroutine owns the retention.

Metrics: receipt phases move to their own `PhaseTimer`, built from the same `evmonly_pipeline` factory with a `stage="receipts"` label (the state commit's timer gets `stage="state"`): the receipt write now overlaps block N-1's `commit_state`, so one timer is not goroutine-safe (the race detector caught this on the first cut) and the two record overlapping intervals — filter on `stage` for a share-of-clock panel. Phase names: `encode_receipts`/`write_receipts` unchanged; the litt-landing wait is now `await_store`; the old `await_receipts` is retired; `await_receipt_write` (`stage="state"`) is the state-commit goroutine's wait for the receipt write (i.e. the part of the receipt write that did *not* overlap). New `start_receipts` block phase (handoff only).

Independent of #4265 (receipt store optional); the two touch `startPipelineCommit` so whichever lands second gets a small rebase.

Expected effect: recovers the part of the 0.16 s/s `app_commit` wait that the ~2 ms loop+router tail can hide; RPC nodes benefit regardless of whether validators turn receipts off.

- New `TestReceiptWriteStartsBeforeThePreviousCommitLands`; `TestCloseWaitsForTheReceiptsOfABlockThatFailedAfterHandingThemOff` pins that `Close` drains the receipt write of a block that failed on the loop after the handoff (no state commit exists to drain it through): with a store-reading block encoder (loop must wait for N-1's commit) and N-1's commit held, block N's receipts are written while N's `ExecutePreparedBlock` is still blocked; N returns only after release; commits land in order.
- Existing pipeline tests unchanged and passing: receipt-failure fails the block before state commit, dropped-write detection, `AwaitReceipts` before commit lands, encoder sees own changes, `ReadLatestAccount` retiring cases.
- `go test -race -count=5 ./giga/evmonly/`, `go test -race ./giga/evmonly/... ./sei-tendermint/internal/evmonlyapp/...` green.
- `make fmtcheck`, `golangci-lint run ./giga/evmonly/...` (0 issues), `go build ./...`.

Requested by: @bdchatham


bdchatham:
@seidroid review

PR #4271, opened 2026-09-20
mempool/evmonlyapp: read nonce/balance through the pending overlay; fetch first-seen accounts outside the store lock

bdchatham:

Re-opens #4258 on top of #4270 (stack 2/4). Depends on #4270's background commit: with the commit landing behind `FinalizeBlock`, `CheckTx`'s nonce/balance reads would otherwise settle (wait for the commit) on every admission.

- `Executor.ReadLatestAccount(addr)` returns the account as of the last block the executor *ran*, laying the in-flight block's `pendingChanges` over a store view without waiting for its commit. A `pipelineGeneration` counter detects a commit that started between opening the view and reading, in which case the read restarts (the view may hold a later block, and replaying the older overlay over it would go backwards). It reports the first failed commit rather than state that lacks that block.
- `evmonlyapp.EvmNonce`/`EvmBalance` use it instead of `openSettledView`, so admission no longer settles the pipeline.
- `mempool/tx.go`: the first-seen account lookup (`accountReader.Account`) moves out of the inner mempool mutex — lookup, lock, recheck that nobody inserted the account meanwhile — so the store read no longer serializes all `CheckTx` callers.

- `scripts/ramtest.sh -race ./giga/evmonly/...` (overlay/pipelined-read tests: sees in-flight block, reports failed commit, restarts when a block lands or retires underneath).
- `go test -race ./sei-tendermint/internal/evmonlyapp/... ./sei-tendermint/internal/mempool/...`.
- `make fmtcheck`, `make lint`.

Requested by: @bdchatham


giga/evmonly/giga_store.go
```diff
@@ -201,12 +201,79 @@ func (e *Executor) AwaitCommits() error {
...
+// commit started after generation was read, since the view may then hold a later block's writes and
+// pending would replay older values over them; the caller reads again.
+func (e *Executor) readLatestAccount(snapshot gigatypes.EVMStateView, pending *pendingChanges, generation uint64, addr common.Address) (LatestAccount, bool) {
+	e.pipelineMu.Lock()
+	moved := e.pipelineGeneration != generation
+	e.pipelineMu.Unlock()
+	if moved {
+		return LatestAccount{}, false
+	}
+	reader := pending.overlay(gigaSnapshotStateReader{snapshot: snapshot, missingState: e.missingState})
```
codchen, line 252: is it possible that N's changes haven't finished committing when N+2 has begun execution? In that case missingState would carry changes from N+1 but not from N unless missingState itself is stacked

codchen (approved):

PR #4273, opened 2026-09-20
evmonly: parallelize OCC validation and merge behind the serial acceptance barrier

bdchatham:

Re-opens #4261 on top of #4272 (stack 4/4). With this PR applied the tree is byte-identical to today's `giga-1`.

OCC acceptance stays a serial barrier in block order, but the work behind it is spread across the existing worker pool (`occ_shards.go`):

- every incarnation's writes are indexed by tx index up front (`stateAccessIndex`);
- a parallel pass checks the pending run of transactions against that index (`conflictsWithin`) and reports the first one the frontier would not accept (`firstUnacceptedResult`); the frontier handles only that tx on the calling goroutine, then reruns from there (with a `serialBackoff` to avoid thrashing on hot contracts);
- the accepted run is folded into the prefix shard by shard (`occShardOf`: contiguous address ranges), each incarnation recording which shards it touched so workers skip results holding nothing of theirs;
- `mergeOCCResults` emits the changeset one shard at a time on the pool and concatenates shards in canonical address order, so the output is identical to the serial merge; a prefix with few keys is merged on the calling goroutine.

Determinism: acceptance decisions and the emitted changeset/receipts are the same as the serial path by construction (parity tests compare both).

- `scripts/ramtest.sh -race ./giga/evmonly/...` — parity vs sequential validation/merge, `conflictsWithin` bounds, `occShardOf`, `cumulativeGasFrom`, `firstUnacceptedResult`, boundary rejection at `results[0]`.
- `make fmtcheck`, `make lint`.

Requested by: @bdchatham


giga/evmonly/occ.go
```diff
@@ -677,99 +662,136 @@ func (e *Executor) mergeOCCResults(ctx context.Context, results []occTxExecution
...
+		shard.balances[change.Address] = cloneBig(change.Balance)
 	}
 	for _, change := range result.changeSet.Nonces {
-		s.nonces[change.Address] = change.Nonce
+		if owns.has(occShardOf(change.Address)) {
+			s.shard(change.Address).nonces[change.Address] = change.Nonce
+		}
 	}
 	for _, change := range result.changeSet.Code {
+		if !owns.has(occShardOf(change.Address)) {
```
shemnon: We are inconsistent with nested ifs (line 765) and guard block style ifs (this).  We should be consistent, at least within each function.

shemnon: We should consider a custom iterator instead of checking inside each loop, if the types are friendly to it.

shemnon (changes requested):

shemnon (approved):

PR #4276, opened 2026-09-21
evmonly: apply value transfers to codeless accounts without the EVM wrapper

bdchatham:

Adds an executor-level fast path in `giga/evmonly` for plain value transfers. The go-ethereum fork already skips the interpreter inside `evm.Call` when the recipient holds no code; what remained per transfer was the `core.StateTransition` wrapper around it (allocation, `Prepare` access-list warming, precompile lookup, snapshot/revert, refund bookkeeping). This PR performs the same transition directly against `nativeStateDB` for transactions that are provably eligible and leaves everything else on `core.ApplyMessage`. The fork is untouched. Same shape as Nethermind's "Fast Eth Transfer" (NethermindEth/nethermind#11804, differential tests in #12278).

Eligibility (`plainTransferCandidate` + `applyPlainTransfer` in `plain_transfer.go`):

```
to != nil && len(data) == 0 && len(accessList) == 0
&& authorizations == nil && blobHashes == nil && blobGasFeeCap == nil
&& !rules.IsEIP4762 && to ∉ activePrecompiles(rules, custom)
&& len(GetCode(to)) == 0          // a 7702 delegation designator is code -> falls through
```

`applyPlainTransfer` mirrors `StateTransition.Execute` step for step and returns the same `core` errors in the same order: nonce (too low/high/max, honoring `SkipNonceChecks`), sender EOA check (delegated senders allowed, as in the fork), London fee-cap/tip/base-fee checks, `BuyGas` balance check, block gas pool, `IntrinsicGas`, Prague floor-data gas, then debit gas → increment nonce → create-if-missing / transfer (`core.Transfer`) → refund → `returnGas` → Sei's coinbase credit including the base fee (`BalanceIncreaseRewardTransactionFee`, so it stays a commutative OCC write). All checks run before any mutation, so no snapshot is needed; a failing check leaves the StateDB untouched exactly as `ApplyMessage` would after rollback.

Plumbing: `blockExecEnv` caches chain config, block context, rules and precompile set per block; `txEVM` builds `vm.NewEVM` lazily so a fast-path transaction never constructs an EVM (previously the OCC path built one per tx). `executeTx`'s snapshot/`RejectUnappliableTxs` handling is unchanged for the `ApplyMessage` branch; a StateDB fault from either path comes back as `stateDBFault` and aborts the block before the `RejectUnappliableTxs` branch, as before. `Config.DisablePlainTransferFastPath` forces every transaction through `ApplyMessage` (used by the tests as the oracle; not read from any config file). Fast-path applications are reported per block as the OTel counter `giga_evmonly_plain_transfers_total`, which the tests also read to prove the path actually ran.

Not in scope: contract calls, precompile calls, 7702 recipients/authorizations, blob and create transactions — all unchanged.

`plain_transfer_test.go` runs every scenario twice (fast path on / forced `ApplyMessage`) on independent copies of the initial state and asserts identical `BlockResult` (tx results, receipts, gas, changeset), identical errors, and a non-vacuous `giga_evmonly_plain_transfers_total` count:

- happy path: existing/nonexistent recipient, zero value, legacy and dynamic fee with base fee, tip capped by fee cap − base fee, self-transfer, sender == coinbase, recipient == coinbase, exact-balance sender, chained transfers within a block
- rejections: one wei short, nonce too low/high (and `SkipNonceChecks`), gas below intrinsic, fee cap below base fee, tip above fee cap, block gas limit exhausted, sender with code, delegated sender, `RejectUnappliableTxs` on/off, a StateDB fault (balance over 256 bits) aborting the block under `RejectUnappliableTxs` on both paths
- fallthrough (counter stays 0, results still identical): code-bearing recipient, 7702-delegated recipient, precompile recipient, calldata, access list, authorization list, contract creation
- `TestPlainTransferAccessSetsMatchApplyMessage`: the OCC read/write sets recorded by `executeTxSpeculative` are equal on both paths (so conflict detection is unchanged)

Microbenchmark (`BenchmarkPlainTransfer`, same tx through `executeTx`, includes StateDB reset and receipt construction; `-benchtime=2s -count=3`):

```
fast-path        ~3.2 µs/op   3632 B/op    83 allocs/op
apply-message    ~7.2 µs/op   7424 B/op   113 allocs/op
```

Also: `scripts/ramtest.sh -race -count=1 ./giga/evmonly/...` green, `go build ./...`, `go vet`, `golangci-lint run ./giga/evmonly/...` 0 issues, `golangci-lint fmt --diff` clean.

Requested by: @bdchatham


bdchatham:
@seidroid review

giga/evmonly/plain_transfer.go
```diff
@@ -0,0 +1,249 @@
...
+// account: the same pre-checks, in the same order and with the same errors,
+// then gas purchase, nonce increment, value movement, gas return and fee
+// payment through the same StateDB calls.
+//
+// It returns applied=false, with the state and gas pool untouched, when the
+// recipient turns out to hold code; the caller then runs core.ApplyMessage,
+// which re-reads the same keys the pre-checks recorded in the access set. An
+// error is returned before any state is modified, so the caller needs no
+// snapshot.
+func (env *blockExecEnv) applyPlainTransfer(
```
codchen, line 44: should we also check the `to` is an EOA? We don't want to skip `receive()` of contracts

giga/evmonly/plain_transfer.go
```diff
@@ -0,0 +1,249 @@
...
+	value, overflow := uint256.FromBig(msg.Value)
+	if overflow {
+		return nil, true, fmt.Errorf("%w: address %v", core.ErrInsufficientFundsForTransfer, msg.From.Hex())
+	}
+	// The balance check above covers gasLimit*gasPrice + value, so the post-purchase
+	// CanTransfer check in StateTransition cannot fail here.
+
+	// The recipient must hold no code; a delegation designator counts as code.
+	to := *msg.To
+	if len(stateDB.GetCode(to)) != 0 {
```
shemnon: Performance: GetCodeSize(to) - it doesn't try to load the whole code.  Or code hash, but that isn't always loaded with the account tuple, it is?

giga/evmonly/plain_transfer.go
```diff
@@ -0,0 +1,249 @@
+package evmonly
```
shemnon, line 1: This file name needs to change, if it's going to become the general EVM with a delayed load. Name implies it only handles plain transfers when it's really a deferred setup variant.

shemnon (commented):
I'm concerned about over-fitting here.  Since all we have is a testnet where we are only running transfers the impact may be exciting.

Do we construct a new EVM for every block?  Every sub-block that is part of an OCC grouping?  Or once per TX?  If it's much higher than once per TX real world use cases will immediately crush this gain, and how often will we see large scale transfer only blocks in production?

bdchatham:
> I'm concerned about over-fitting here. Since all we have is a testnet where we are only running transfers the impact may be exciting.

@shemnon, definitely agree, we don't want to take a bad trade off where overhead goes to other transactions but not these simpler transfer cases. Especially not just to pat our backs for a cool benchmark. That said, what convinces me this is a worthwhile trade off is that geth already recognizes these as calls to a no-code account and skips the interpreter. The logical diff is it's lower in the stack, after the ApplyMessage wrapper has done its allocations. We're moving that same decision one layer up so the wrapper goes too (Nethermind shipped the same shape in NethermindEth/nethermind#11804).

We pay a relatively cheap work with no state reads and EVM construction cadence is unchanged for every other tx. On the other side, we're cutting out a whole category of allocations on what is a reasonably common tx type on any real network, not just our testnet.

Edit: there is a real cost of code added of course. It's not small, so want to make sure I'm acknowledging that too. It's not "just cheap checks"

shemnon:
Now that txEvm has been pulled out I can see that the plain_transfer function is fairly independent. The original structure obscured that and my first read was that getting rid of the evm init was the main speedup, vs. the per transfer call _into_ the evm.  St

shemnon (approved):
LGTM unless  @codchen has objections.

Two thinks to consider before mergeing:

(a) we will be adding EVM log events to transfers, this short cut will still allow us to handle building those events from a single function, right?  Just called from each path for value bearing transaction to code and non-code?

(b) high volumes of transfer only txes will not be the mayn tx typein high workloads, giga testnet is very synthetic in that regard.  We will be looking at mostly ERC-20 transfers and AMM code, not naked transfers.

PR #4290, opened 2026-09-22
Remove what the pointer creation removals left behind

alexander-sei:
[#4198](https://github.com/sei-protocol/sei-chain/pull/4198) and [#4230](https://github.com/sei-protocol/sei-chain/pull/4230) closed both pointer creation directions, but left behind test coverage and symbols that no longer reach anything.

## Test coverage for creation

- The two cw2981 royalty tests in `x/evm/integration_test.go` — the last callers of `UpsertERCCW{721,1155}Pointer` outside the legacy precompiles — and the two royalty wasm fixtures they read (868 KB).
- The `pointer` flag on `evmrpc/tests`' `cw20Initializer`.

**The `pointer` flag was never load-bearing**, which is the one claim here worth checking. `eth_*` deliberately does not surface CW-only synthetic receipts — `TestGetTransactionReceiptWithMixedLogs` already asserts exactly that, with `require.Nil` on the CW transaction's receipt and zero receipts for the CW-only block. So no test in the package could observe whether a pointer existed. Disabling registration outright leaves all five affected tests passing, so the flag is removed rather than re-pointed at a registry write nothing reads.

The flip side is worth stating plainly: **`evmrpc` has no synthetic-receipt coverage, and did not before this change either.** If `TestBlockBloom` looks like it guards against synthetic CW blooms leaking into the block bloom, it cannot — there is no synthetic bloom present for it to catch.

## Symbols orphaned by #4198

- `AttributeKeyPointerVersion`, referenced nowhere in the repo.
- `Keeper.GetRegisterPointerDisabled`, whose only remaining callers were three assertions in its own test.

The `register_pointer_disabled` **param stays wired** — field, `KeyRegisterPointerDisabled`, `DefaultRegisterPointerDisabled`, `ParamSetPair` and validator. The paramstore key table has to register every field, or `GetParams` would fail on chains carrying the param in state. Only the unused Go getter is gone.

## Dead fixtures from the same lineage

- `contracts/scripts/deploy.js`, which read a wasm path that no longer resolves and deployed `CW20ERC20Wrapper`, absent from `contracts/src` since the wrapper removal. Nothing referenced it; the `scripts/deploy.sh` hits in the Dockerfiles are a different file.
- The `WASM.CW721` and `WASM.CW1155` entries in `contracts/test/lib.js`. Only `WASM.CW20` is read, by `SeiSoloTest.js` and `DisableWasmTest.js`.

## What deliberately stays

Synthetic receipt and log coverage is untouched: `app/receipt_test.go` still covers all four CW pointee cases against a registry-only pointer, and the EVM→CW direction in `mixedLogTesterInitializer` is unchanged.

`UpsertERC*Pointer` and the `artifacts/cw{20,721,1155}` bytecode stay, because `precompiles/pointer/legacy/*` needs them to replay historical creation under `debug_trace*`, as `x/evm/AGENTS.md` already records.

## Verification

`go vet` over `./x/evm/...` and `./giga/...`, `make fmtcheck`, and `scripts/ramtest.sh` over `evmrpc/tests`, `x/evm/...` and `app/...` — all clean. The `x/evm/integration_test.go` change is a pure deletion with zero added lines, so the retained functions are byte-identical.

The JS suites need a running node, so `contracts/test/lib.js` was verified by `node --check` plus a reference sweep rather than execution.

## Not addressed

`giga/deps/xevm` carries its own identical copy of `GetRegisterPointerDisabled` with the same three test assertions, equally dead. Left alone because that tree is a hand-maintained fork with no sync script, so whether it tracks mainline cleanups is a call about the fork's purpose.


alexander-sei:
@seidroid review

sei-will (approved):

masih (approved):

philipsu522 (approved):

